PR #127, opened 2014-12-20
Bug in getReferencedTypes

JimiPepper:
**Message from last commit in bug branch :**
Adds one unit test that failed when you try to git a static reference type (in this case Encoding class which is a static class belonging to java.sound.sampled.AudioFormat). AudioFormat class takes several arguments to be instantiated including an Encoding instance. But Encoding class is a special case : It has public attributes which returns an Encoding object. But when you spoon an AudioFormat instantiation, it seems that Encoding dependency is not detected. The unit test I tried is in NoClasspathTest test file. It simply spoons a Bar class containing one method (doSomething) returning an AudioFormat object. The test get type references from the return statement to check if Encoding type is there.


monperrus:
Transformed as issue #174

PR #134, opened 2015-01-08
Remove uneeded toLowerCase() method call as the lower case version is

seintur:
Remove uneeded toLowerCase() method call as the lower case version is cached in the Enum type instances and returned by toString().


boeserwolf91:
The question isn't directly related to this pull request, but does it make sense to buffer the lower cased name? I would recommend removing the toString method and call mod.name().toLowerCase(Locale.US) if it's needed. But not storing the value. Is there any reason for buffering the value?

GerardPaligot:
That is a choice. The current solution doesn't shock me. When I call the toString method, I expect the correct output (according to the java syntax).

PR #151, opened 2015-02-05
Modifying  'equals' method from CtElementImpl. It has an if precondition...

martinezmatias:
... for checking the signature of the two elements under comparison: if at least one is empty the method throws an exception. This check produces that we can not compare empty literals

### DIFF
--- a/src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
+++ b/src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
@@ -127,8 +127,6 @@ public abstract class CtElementImpl implements CtElement, Serializable {
 			return false;
 		String current = getSignature();
 		String other = ((CtElement) o).getSignature();
-		if (current.length() <= 0 || other.length() <= 0)
-			throw new CtUncomparableException("Unable to compare elements");
 		return current.equals(other);
 	}
 
--- a/src/test/java/spoon/test/comparison/EqualTest.java
+++ b/src/test/java/spoon/test/comparison/EqualTest.java
@@ -1,0 +1,67 @@
+package spoon.test.comparison;
+
+import static org.junit.Assert.*;
+
+import org.junit.Test;
+
+import spoon.Launcher;
+import spoon.compiler.SpoonCompiler;
+import spoon.reflect.code.CtInvocation;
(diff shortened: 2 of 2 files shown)


src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
```diff
@@ -127,8 +127,8 @@ public boolean equals(Object o) {
 			return false;
 		String current = getSignature();
 		String other = ((CtElement) o).getSignature();
-		if (current.length() <= 0 || other.length() <= 0)
-			throw new CtUncomparableException("Unable to compare elements");
+		//if (current.length() <= 0 || other.length() <= 0)
+		//	throw new CtUncomparableException("Unable to compare elements");
```
GerardPaligot: Why you didn't remove these lines? You wrote a test case which will become red if we remove these comments.

GerardPaligot:
Congrats @martinezmatias. You are the first person who launched its PR in our Jenkins and all projects stayed green!

src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
martinezmatias: Yes, that is the point. I removed the "IF + throw"  because that check does not allow to compare elements with empty signature. As consequence, if we have an empty CtLiteral (imagine as example the: code "".equals(myString) ) , we can not compare this element. I wrote a test case that shows this problem: if you remove the comment (i.e., keep the precondition) the comparison of an empty ctelement will fail. So, the patch I propose is to remove the if and throw statement

GerardPaligot: Yes, yes but I would like to know why did you comment these lines and not remove them?

martinezmatias: Ah, ok, I can remove them:)

GerardPaligot: Yep, thanks. Because you wrote a test case which will fail if we uncomment these lines. So, they are no more necessary.

martinezmatias:
Commit updated: I removed the commented code.

PR #155, opened 2015-02-16
refactors the filtering API

monperrus:
1) uses runtime reflection instead of explicit and duplicate static typing
2) consequently, removes one of the oldest todos of Spoon
2) simplifies the API of filters (now only one method "matches")
3) adds test cases for the untested filters
4) cleans FieldAccesFilter, now a subtype of VariableAccessFilter

### DIFF
--- a/src/main/java/spoon/reflect/visitor/QueryVisitor.java
+++ b/src/main/java/spoon/reflect/visitor/QueryVisitor.java
@@ -49,12 +49,14 @@ public class QueryVisitor<T extends CtElement> extends CtScanner {
 	@SuppressWarnings("unchecked")
 	@Override
 	public void scan(CtElement element) {
-		if (element == null)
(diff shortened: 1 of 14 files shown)


src/main/java/spoon/reflect/visitor/QueryVisitor.java
```diff
@@ -49,12 +49,16 @@ public QueryVisitor(Filter<T> filter) {
...
 	@Override
 	public void scan(CtElement element) {
-		if (element == null)
+		if (element == null) 
 			return;
-		if (filter.getType().isAssignableFrom(element.getClass())) {
-			if (filter.matches((T) element)) {
-				result.add((T) element);
+		try {
+			{
```
GerardPaligot: This block doesn't seem necessary.

PR #161, opened 2015-03-02
move symbolic evaluator to a standalone project

petitpre:
I move the untested, unused and probably not fully-working symbolic evaluator from spoon to a standalone project : 
https://github.com/SpoonLabs/spoon-symbolic-evaluator


GerardPaligot:
+1 for this PR.

PR #171, opened 2015-03-19
introduces SpoonAPI

monperrus:
- cleans Launcher
- extracts its core API in an interface


src/main/java/spoon/Launcher.java
```diff
@@ -75,10 +75,12 @@
  * @see spoon.processing.ProcessingManager
  * @see spoon.processing.Processor
  */
-public class Launcher {
+	public class Launcher implements SpoonAPI {
```
GerardPaligot: Can you remove this tab?

src/main/java/spoon/Launcher.java
```diff
@@ -75,10 +75,12 @@
...
 
 	private Factory factory = createFactory();
 
+	private SpoonCompiler compiler = createCompiler();
```
GerardPaligot: Why do you deprecate `SpoonCompiler` but use it here?

src/main/java/spoon/Launcher.java
```diff
@@ -595,14 +599,8 @@ public SpoonCompiler createCompiler(Factory factory) {
...
-	/**
-	 * Creates a new Spoon Java compiler in order to process and compile Java
-	 * source code.
-	 *
-	 * @param factory
-	 *            the factory this compiler works on
-	 * @param inputSources
-	 *            the sources to be processed and/or compiled
+	/* (non-Javadoc)
+	 * @see spoon.SpoonAPI#createCompiler(spoon.reflect.factory.Factory, java.util.List)
```
GerardPaligot: You reference a method that doesn't exist.

src/main/java/spoon/Launcher.java
```diff
@@ -624,10 +622,8 @@ public SpoonCompiler createCompiler(Factory factory,
 		return c;
 	}
 
-	/**
-	 * Creates a new Spoon Java compiler with a default factory in order to
-	 * process and compile Java source code. The compiler's factory can be
-	 * accessed with the {@link SpoonCompiler#getFactory()}.
+	/* (non-Javadoc)
+	 * @see spoon.SpoonAPI#createCompiler()
```
GerardPaligot: You reference a method that doesn't exist.

src/main/java/spoon/support/compiler/jdt/JDTBasedSpoonCompiler.java
```diff
@@ -46,6 +46,7 @@
 
 import spoon.Launcher;
 import spoon.OutputType;
+import spoon.SpoonAPI;
```
GerardPaligot: This import is never used.

PR #172, opened 2015-03-19
Changes in Signature Printer and Test Cases

martinezmatias:
The signature printer loses information by printing invocation and literals.

### DIFF
--- a/src/main/java/spoon/support/visitor/SignaturePrinter.java
+++ b/src/main/java/spoon/support/visitor/SignaturePrinter.java
@@ -353,7 +353,15 @@ public class SignaturePrinter implements CtVisitor {
 	public <T> void visitCtInvocation(CtInvocation<T> invocation) {
 		write("(");
 		scan(invocation.getExecutable());
-		write("(...)");
+		write("(");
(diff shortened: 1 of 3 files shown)


GerardPaligot:
Thanks for your PR @martinezmatias but I would like to know something:

You test signature of static method invocations but never instance method invocations. What happens if we invocate instance method of 2 variables typed by 2 different classes?

```
public void wrap1() {
  Foo1 foo = new Foo1();
  foo.bar(); // Signature of this invocation?
}

public void wrap2() {
  Foo2 foo = new Foo2();
  foo.bar(); // Signature of this invocation?
}
```

martinezmatias:
Hi @GerardPaligot, good point. I created a new test that evaluate the cases when the arguments are local variables or parameters. I had to change the signature of those elements: it was not consistent with the signature of FieldsAccess.

martinezmatias:
These catches are not necessaries (They come from the old version of "visit field reference"  method), and I haven' t cleaned the code. I removed them and I will commit those changes within a new PR which targets another issue.

PR #173, opened 2015-03-24
CtPackage.addType sets the parent

monperrus:


src/test/java/spoon/test/parent/ParentTest.java
```diff
@@ -75,4 +76,18 @@ public void testParentSet() throws Exception {
...
 				
 	}
+	
+	@Test
+	public void testParentPackage() throws Exception {
+		// addType should set Parent
+		CtClass clazz = factory.Core().createClass();
+		clazz.setSimpleName("Foo");
+		CtPackage pack = factory.Core().createPackage(); 
+		//System.out.println(pack.getQualifiedName());
```
GerardPaligot: This comment is useless.

PR #175, opened 2015-03-25
Fix CtExecutableReference.getActualMethod() when method is inherited

petitpre:


GerardPaligot:
Where is your test case which was red before your bug fix? You have formatted the test class and it became horrible to read.

PR #182, opened 2015-03-30
fix: Writes a correct this or super invocation after a class rename.

### LINKED ISSUE #166: This invocation becomes super invocation after changing name of the enclosing class
We have a class `AClass` with 2 constructors and one of them referenced the other one with a `this` instruction:

```java
public class AClass {
  private final String string;

  public AClass() {
    this("");
  }

  public AClass(String string) {
    this.string = string;
  }
}
```

When we would like rename the name of the `CtClass`, `this("")` becomes `super("")`. How can be possible? Well, in the `DefaultJavaPrettyPrinter`, we print `this` or `super` according to these conditions:

```java
CtType<?> parentType = invocation.getParent(CtType.class);
if ((parentType != null) 
    && (parentType.getQualifiedName() != null)
    && parentType.getQualifiedName().equals(invocation.getExecutable().getDeclaringType().getQualifiedName())) {
  write("this");
} else {
  if (invocation.getTarget() != null) {
    write(invocation.getTarget().getSignature() + '.');
  }
  write("super");
}
```

The problem is that `invocation.getExecutable().getDeclaringType().getQualifiedName()` returns a `CtTypeReference` with the older name of the `CtClass`. So, the condition returns false and we print a `super` instruction.

GerardPaligot:
When a class is rename, references of the class concerned aren't renamed.
So we don't calculate no more the type of the invocation in the
DefaultJavaPrettyPrinter but when we build the model, in the JDTTreeBuilder.

Closes #166

### DIFF
--- a/src/main/java/spoon/reflect/code/CtInvocation.java
+++ b/src/main/java/spoon/reflect/code/CtInvocation.java
@@ -17,14 +17,31 @@
 
 package spoon.reflect.code;
 
-
 /**
  * This code element defines a concrete invocation.
- * 
+ *
  * @param <T>
- *            Return type of this invocation
+ * 		Return type of this invocation
  */
 public interface CtInvocation<T> extends CtAbstractInvocation<T>, CtStatement,
 		CtTargetedExpression<T, CtExpression<?>> {
+	/**
+	 * Checks if the invocation is a this invocation.
+	 */
+	boolean isThisInvocation();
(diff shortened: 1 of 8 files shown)


monperrus:
After discussion in our Spoon committee today, a better solution will be to put this in a Refactorings.changeTypeName(CtType t)

PR #184, opened 2015-03-30
refactors CtTypeReference

monperrus:
CtTypeReference is where, for historical reasons, many methods where put,
breaking the initial vision and the clear responsibility. This refactoring
makes it coming back to its roots. This refactoring should neither break API
nor behavioral compatibility.

It consists of three main steps:

1) introduce CtTypeInformation which bridges information that can be both get
at compile-time and at runtime. This introduces good regularity in the API.
(This was implicit and half-done before)

2) move compile-time methods to the CtElement hierarchy (e.g.
getAllExecutables)

3) move important runtime methods in RtHelper (an old class as well)

4) remove box/unox that are only used once and are too specific for being in the interface

There still remains the problems of the hacky getPackage and getDeclaringType.
They have their counter-part in the CtElement hierarchy, but in one case they
return a reference and in the other an element. Hard to understand for the new
comer, and impossible to unify. Still, it seems we may be able to simply removing them.

@Nicolas: your CtInheritanceScannerTest caught me! Good job! :-)


src/main/java/spoon/reflect/declaration/CtSimpleType.java
```diff
@@ -31,7 +32,7 @@
  * @param <T>
  *            the actual runtime type
  */
-public interface CtSimpleType<T> extends CtNamedElement, CtTypeMember {
+public interface CtSimpleType<T> extends CtNamedElement, CtTypeInformation, CtTypeMember {
```
GerardPaligot: Mh. I'm not sure it is really a problem but now, a `CtSimpleType` can access to its super class, super interfaces and can calculate if it is a sub type of another type.

GerardPaligot:
This PR can be closed. A new PR has been opened https://github.com/INRIA/spoon/pull/195.

PR #187, opened 2015-03-31
fix: Adds all references via TypeReferenceScanner.

### LINKED ISSUE #174: Bug in getReferencedTypes
it seems that AudioFormat.Encoding dependency is not detected by getReferencedTypes

```
import javax.sound.sampled.AudioFormat;
import javax.sound.sampled.AudioFormat.Encoding;
import java.util.HashMap;

public class Bar {
    public AudioFormat doSomething() {
        return new AudioFormat(Encoding.ALAW, (float)1.0, 8, 2, 1, (float)1.0, true, new HashMap<String, Object>());
    }
}
```

GerardPaligot:
Before, in the TypeReferenceScanner, if the enclosing reference
of the current reference is already in the set of references,
we didn't add it in the set. Now, we add all in set and the Set
class remove all duplicate items.

Closes #174

### DIFF
--- a/src/test/java/spoon/test/api/NoClasspathTest.java
+++ b/src/test/java/spoon/test/api/NoClasspathTest.java
@@ -7,7 +7,9 @@ import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
 import java.io.File;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 import org.junit.Test;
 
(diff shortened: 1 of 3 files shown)


src/test/java/spoon/test/api/NoClasspathTest.java
```diff
@@ -108,5 +113,32 @@ public void testBug20141021() {
...
+		Launcher spoon = new Launcher();
+		final Factory factory = spoon.getFactory();
+		factory.getEnvironment().setAutoImports(false);
+		spoon.addInputResource("./src/test/java/spoon/test/api/testclasses/");
+		spoon.run();
+
+		CtTypeReference expectedType = factory.Type().createReference(javax.sound.sampled.AudioFormat.Encoding.class);
+		CtClass<?> clazz = factory.Class().get(Bar.class);
+
+		try {
```
monperrus: remove try block

src/test/java/spoon/test/api/NoClasspathTest.java
```diff
@@ -108,5 +113,32 @@ public void testBug20141021() {
...
+			CtMethod<?> method = clazz.getMethodsByName("doSomething").get(0);
+			CtReturn ctReturn = method.getElements(new TypeFilter<CtReturn>(CtReturn.class)).get(0);
+
+			Set<CtTypeReference> references = new HashSet<CtTypeReference>();
+
+			for (CtTypeReference<?> type : ctReturn.getReferencedTypes()) {
+				references.add(type);
+			}
+
+			assertEquals(true, references.contains(expectedType));
```
monperrus: why not simply ctReturn.getReferencedTypes().contains(...)?

PR #194, opened 2015-04-04
Use Java 8 functional interfaces for filtering model elements.

seintur:


boeserwolf91:
:-1: for this.For me this change is not worth to set the language level to java 8. I prefer the java version 6.

pschichtel:
:-1: from me too. I'd say that's better done in an extra java8 support module

GerardPaligot:
I agree with @boeserwolf91 and @pschichtel. It seems overkill to upgrade java version support from 1.6 to 1.8 just to have a beautiful syntax with filters in spoon.

seintur:
No problem guys, Time is on my side ;-)

PR #200, opened 2015-04-11
Redefine the getDeclaration() method with a more specialized return type.

seintur:
Subject says it all.

### DIFF
--- a/src/main/java/spoon/reflect/reference/CtTypeReference.java
+++ b/src/main/java/spoon/reflect/reference/CtTypeReference.java
@@ -17,6 +17,7 @@
 
 package spoon.reflect.reference;
 
+import spoon.reflect.declaration.CtType;
 import spoon.reflect.declaration.CtTypeInformation;
 
 /**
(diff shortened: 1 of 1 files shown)


src/main/java/spoon/reflect/reference/CtTypeReference.java
```diff
@@ -42,6 +43,13 @@
 	Class<T> getActualClass();
 
 	/**
+	 * Tries to get the declaration that corresponds to the referenced element.
```
monperrus: Returns the CtElement, a CtType, that corresponds to the reference.

src/main/java/spoon/reflect/reference/CtTypeReference.java
```diff
@@ -42,6 +43,13 @@
...
+	 * 
+	 * @return referenced element or null if element does not exist
```
monperrus: or null if the type declaration is not in the analyzed source files.

PR #201, opened 2015-04-12
Use addConstructor method instead of get().add()

seintur:
get().add() throws UnsupportedOperationException because the returned collection is non
mutable.


petitpre:
Could you add some tests to check the factory and the constructors unmutable ?

PR #205, opened 2015-04-17
Throws an exception when we try to insert statement before super invocation.

GerardPaligot:
All in the title.


monperrus:
change to SpoonException?

GerardPaligot:
Ready to merge!

PR #210, opened 2015-04-19
Bug fix for CtTypeImpl#getAllMethods.

seintur:
CtTypeReference cannot be cast to CtType and CtTypeReference does not provide getAllMethods.
Associated test for CtAnnotationTypeImpl.
getAllMethods is redefined in other subtypes of CtType, i.e. in CtClassImpl and in CtInterfaceImpl.

### DIFF
--- a/src/test/java/spoon/test/annotation/AnnotationTest.java
+++ b/src/test/java/spoon/test/annotation/AnnotationTest.java
@@ -79,6 +79,7 @@ public class AnnotationTest {
 		CtType<?> type = this.factory.Type().get("spoon.test.annotation.testclasses.Bound");
 		assertEquals("Bound", type.getSimpleName());
 		assertEquals(1, type.getAnnotations().size());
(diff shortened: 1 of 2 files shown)


monperrus:
Why is it bug? 
The contract in the super interface says that super classes are taken into account.

src/test/java/spoon/test/annotation/AnnotationTest.java
```diff
@@ -79,6 +79,7 @@ public void testModelBuildingAnnotationBound() throws Exception {
 		CtType<?> type = this.factory.Type().get("spoon.test.annotation.testclasses.Bound");
 		assertEquals("Bound", type.getSimpleName());
 		assertEquals(1, type.getAnnotations().size());
+		assertEquals(0, type.getAllMethods().size());
```
GerardPaligot: Your assert was red before your modification?

seintur:
LIve discussion with @monperrus

PR #219, opened 2015-05-09
Fix a copy-paste bug in CtTypeReferenceImpl.unbox()

leventov:


src/test/java/spoon/test/reference/TypeReferenceTest.java
```diff
@@ -166,4 +166,11 @@ public void testNullReferenceSubtype() throws Exception {
 		assertFalse(nullRef.isSubtypeOf(ref));
 
 	}
+	
+    @Test
+    public void unboxTest() {
+        Factory factory = new Launcher().createFactory();
+        CtTypeReference<Boolean> boxedBoolean = factory.Class().createReference(Boolean.class);
+        assertEquals(boxedBoolean.unbox().getActualClass(), boolean.class);
```
monperrus: boolean.class should be the first argument (in JUnit, the order is (expected, actual))

PR #246, opened 2015-06-02
Refactor CtSuperAccess, CtThisAccess and CtFieldAccess

### LINKED ISSUE #228: Deprecate CtTargetedAccess?
- I doubt CtTargetedAccess is ever a useful abstraction, because `this` and `super` are not proper variables.

GerardPaligot:
- Changes super class of CtSuperAccess and CtThisAccess for CtTargetedExpression.
- Merges CtFieldAccess and CtTargetedAccess.

Closes #228 and #243

So, I propose this new AST:

targeted_access


src/main/java/spoon/reflect/code/CtThisAccess.java
```diff
@@ -19,20 +19,9 @@
...
-	 * Tells if this access must be qualified by the class it refers to.
-	 */
-	boolean isQualified();
-
-	/**
-	 * Sets this access to be qualified by the class it refers to.
-	 */
-	void setQualified(boolean qualified);
-
+public interface CtThisAccess<T> extends CtTargetedExpression<T, CtExpression<?>> {
```
monperrus: extend CtTargetedAccess?

GerardPaligot: Why? `CtSuperAccess` extends `CtTargetedAccess` to stay compatible with older version of Spoon but that isn't the case for `CtThisAccess`.

PR #251, opened 2015-06-10
replace generic method CtElement.replace() by more specifics methods …

petitpre:
…in CtStatement, CtExpression, etc

### DIFF
--- a/src/main/java/spoon/reflect/code/CtExpression.java
+++ b/src/main/java/spoon/reflect/code/CtExpression.java
@@ -46,5 +46,10 @@ public interface CtExpression<T> extends CtCodeElement, CtTypedElement<T>,
 	 * Adds a type cast.
 	 */
 	void addTypeCast(CtTypeReference<?> type);
+
+	/**
+	 * Replaces this element by another one.
+	 */
+	<E extends T> void replace(CtExpression<E> element);
 	
 }
--- a/src/main/java/spoon/reflect/code/CtStatement.java
+++ b/src/main/java/spoon/reflect/code/CtStatement.java
@@ -64,4 +64,8 @@ public interface CtStatement extends CtCodeElement {
 	 */
 	void setLabel(String label);
 
+	/**
+	 * Replaces this element by another one.
+	 */
+	void replace(CtStatement element);
 }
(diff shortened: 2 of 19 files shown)


src/main/java/spoon/reflect/declaration/CtField.java
```diff
@@ -35,4 +35,9 @@
 	 * @see spoon.reflect.declaration.CtNamedElement#getReference()
 	 */
 	CtFieldReference<T> getReference();
+
+	/**
+	 * Replaces this element by another one.
+	 */
+	<R extends T> void replace(CtField<R> element);
```
GerardPaligot: I'm not sure it is relevant to replace a field.

src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
```diff
@@ -328,14 +328,14 @@ private boolean compare(Object o1, Object o2) {
...
-		}
-	}
+//	public void replace(Filter<? extends CtElement> replacementPoints,
+//			CtElement element) {
+//		List<? extends CtElement> l = Query
+//				.getElements(this, replacementPoints);
+//		for (CtElement e : l) {
+//			e.replace(element);
+//		}
+//	}
```
GerardPaligot: If it is no more used, remove this method. We have the git history if we would like to retrieve something.

src/main/java/spoon/reflect/declaration/CtField.java
monperrus: It may be useful in some cases. We may add a precondition on the name and/or the type.

monperrus: To be consistent, we should add CtMethod.replace

src/main/java/spoon/reflect/code/CtExpression.java
```diff
@@ -46,5 +46,10 @@
 	 * Adds a type cast.
 	 */
 	void addTypeCast(CtTypeReference<?> type);
+
+	/**
+	 * Replaces this element by another one.
+	 */
+	<E extends T> void replace(CtExpression<E> element);
```
monperrus: add a specific test case for CtExpression.replace?

src/main/java/spoon/reflect/code/CtStatement.java
```diff
@@ -64,4 +64,8 @@ void insertBefore(CtStatementList statements)
 	 */
 	void setLabel(String label);
 
+	/**
+	 * Replaces this element by another one.
+	 */
+	void replace(CtStatement element);
```
monperrus: add a specific test case for CtStatement.replace?

PR #256, opened 2015-06-14
add SpoonModelTree.expandAll

argius:
Hello, this is my first PR in spoon.

I implement a new feature "Expand all" to GUI.
The feature, in SpoonModelTree, expands all child nodes of a selected node recursively.

Thanks for your review.

Regards.

### DIFF
--- a/src/main/java/spoon/support/gui/SpoonModelTree.java
+++ b/src/main/java/spoon/support/gui/SpoonModelTree.java
@@ -317,15 +317,7 @@ public class SpoonModelTree extends JFrame implements KeyListener,
 					q.add(node);
 					while (!q.isEmpty()) {
 						final DefaultMutableTreeNode n = q.poll();
-						SwingUtilities.invokeLater(new Runnable() {
-							public void run() {
-								TreePath path = new TreePath(n.getPath());
-								if (!jTree.isExpanded(path)) {
-									jTree.expandPath(path);
-									jTree.updateUI();
-								}
-							}
-						});
+						expand(n);
 						@SuppressWarnings("unchecked")
 						Enumeration<DefaultMutableTreeNode> children = n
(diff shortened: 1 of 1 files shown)


GerardPaligot:
Hello @argius,

First, thanks for your first PR on Spoon! :)

I tested your PR and it works well. I'll make a code review later today but I already have two remarks:

- The depth of your method is really big. It is maybe a good idea to reduce it.
- Your method returns always `null`.

argius:
Hello Gérard,

Thanks for your review and pointing out some problems in my code.
I will consider those points.

argius:
I revised `expandAll` method as follows:

- The method returns the top of the expanded node. If not expanded, it returns `null`.
- Removed some unnecessary codes; cast of `DefaultMutableTreeNode`, catch block.
- Uses `SwingUtilities.invokeLater` instead `invokeAndWait`.

Thanks.

src/main/java/spoon/support/gui/SpoonModelTree.java
```diff
@@ -284,6 +305,46 @@ public DefaultMutableTreeNode search() {
...
+			return null;
+		}
+		final ExecutorService executor = Executors.newSingleThreadExecutor();
+		executor.execute(new Runnable() {
+			public void run() {
+				try {
+					Queue<DefaultMutableTreeNode> q = new LinkedList<DefaultMutableTreeNode>();
+					q.add(node);
+					while (!q.isEmpty()) {
+						final DefaultMutableTreeNode n = q.poll();
```
GerardPaligot: Creates a method `expand(node)` seems a good idea to reduce the depth of this method.

argius: I got it.

Issue #249, opened 2015-06-06
Problems with setTarget() assigning


leventov:
Now source calls `setParent(this)` in setter methods more consistently, but it leads to NullPointerExceptions in some cases.

E. g. `ctTargetedAccess.setTarget(null)` -- seems to be a valid way to construct AST sometimes? Or I should always know on which `this` (it might be not very simple, if there are nested inner classes) I call each expression and set `this` implicit?

Otherwise source of many setters should look more like
```
setTarget(E target) {
    if (target != null) // this condition is absent currently
        target.setParent(this);
    this.target = target;
}
```

monperrus:
Can you write a failing test case?

leventov:
Ok, see the pull request. The test case might look artificial, but I do actually need this. Imagine I want to replace delegation:

```
class Foo {
  Bar bar;
  void foo() {
    bar.bar();
  }
}
```

with inheritance/merged classes. In this case I have some code like
```
fooAst.filter(CtTargetedAccess.class).forEach(access -> {
  if (access.getTarget().getReference().getClass() == Bar.class) {
    access.setTarget(null);
  }
})
```

to remove `bar.` parts.

As I said, replacing `bar.` with `this.` (might be implicit) is possible, but not convenient, in case of nested classes etc., and also less efficient (why create `CtThisAccess` instances for each occurance, if it might be just `null`).

PR #259, opened 2015-06-15
Allow to assign null AST childs after non-null

leventov:
See #249

### DIFF
--- a/src/main/java/spoon/support/reflect/code/CtAssertImpl.java
+++ b/src/main/java/spoon/support/reflect/code/CtAssertImpl.java
@@ -46,7 +46,8 @@ public class CtAssertImpl<T> extends CtStatementImpl implements CtAssert<T> {
 	}
 
 	public void setExpression(CtExpression<T> value) {
-		value.setParent(this);
+		if (value != null)
+			value.setParent(this);
 		this.value = value;
 	}
 
--- a/src/test/java/spoon/test/targeted/TargetedExpressionTest.java
+++ b/src/test/java/spoon/test/targeted/TargetedExpressionTest.java
@@ -5,13 +5,16 @@ import spoon.reflect.code.CtFieldAccess;
 import spoon.reflect.code.CtSuperAccess;
 import spoon.reflect.declaration.CtClass;
 import spoon.reflect.declaration.CtConstructor;
+import spoon.reflect.declaration.CtField;
 import spoon.reflect.declaration.CtMethod;
 import spoon.reflect.declaration.CtType;
 import spoon.reflect.factory.Factory;
+import spoon.reflect.reference.CtFieldReference;
 import spoon.reflect.visitor.filter.AbstractFilter;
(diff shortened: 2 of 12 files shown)


src/test/java/spoon/test/targeted/TargetedExpressionTest.java
```diff
@@ -74,4 +77,21 @@ public void testTargetOfFieldAccess() throws Exception {
 		assertEquals("Target is CtThisAccessImpl if there is a 'this' explicit.", CtThisAccessImpl.class, elements.get(0).getTarget().getClass());
 		assertNull("Targets is null if there isn't a 'this' explicit.", elements.get(1).getTarget());
 	}
+
+	@Test
+	public void testNotTargetedExpression() throws Exception {
+		Factory factory = build(Foo.class);
+		CtClass<Object> fooClass = factory.Class().get(Foo.class);
+		CtField<?> iField = fooClass.getField("i");
+		CtVariableAccess<?> iRead =
```
GerardPaligot: unused.

GerardPaligot:
This feature seems necessary. We shouldn't have a NPE when we set a target with null. But my first reaction was that it is awful to check everywhere if the parameter is null. After reflection, if we want to keep the setters of parents and/or target on the AST, we don't have any choices. So, I'm ok with this PR.

leventov:
@GerardPaligot fixed

src/main/java/spoon/support/reflect/code/CtAssertImpl.java
```diff
@@ -46,7 +46,8 @@ public void accept(CtVisitor visitor) {
 	}
 
 	public void setExpression(CtExpression<T> value) {
-		value.setParent(this);
+		if (value != null)
+			value.setParent(this);
```
GerardPaligot: An expression of an assertion can be null? For me, an assertion must have an expression so I propose to change for something like:

```java
public void setExpression(CtExpression<T> value) {
    if (value == null) {
        throw new SpoonException("An assertion must have an expression.");
    }
    value.setParent(this);
    this.value = value;
}
```

What do you think?

GerardPaligot:
For now, you have only one test in your PR to test a target null but there aren't tests for other checks. It could be interesting to add new tests for all others cases. We know now why we add these check but not necessarily in a few years.

So, can you add tests?

src/main/java/spoon/support/reflect/code/CtAssertImpl.java
leventov: There is an "assertionExpression" and just "expression". Null expression corresponds to the statement `assert cond;`

leventov:
Ok

leventov:
@GerardPaligot please look why CI build is failing. Locally it is ok.

GerardPaligot:
I hate bugs I can't reproduce in local. I'll check what is the problem.

PR #268, opened 2015-06-25
improves static typing so as to support diamonds with TypeFilter

monperrus:


GerardPaligot:
I'm ok with this PR but I'm not a fan of the test case.

PR #269, opened 2015-06-25
purifies the contract of CtInheritanceScanner, documents it and improves its error message

monperrus:
contract is now:
if a method visitX exists, it must be invoked
if a method scanX exists, it must be invoked

(change: we don't force to write empty visitX and scanX methods for new metamodel elements)

for example, the failure message is now:
visitCtCodeSnippetExpression does not call scanCtExpression

### DIFF
--- a/src/test/java/spoon/reflect/visitor/CtInheritanceScannerTest.java
+++ b/src/test/java/spoon/reflect/visitor/CtInheritanceScannerTest.java
@@ -1,5 +1,6 @@
 package spoon.reflect.visitor;
 
+import static org.junit.Assert.fail;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.verify;
 
@@ -15,6 +16,7 @@ import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
 import org.mockito.Mockito;
+
 import spoon.reflect.factory.CoreFactory;
 import spoon.reflect.factory.Factory;
 import spoon.test.TestUtils;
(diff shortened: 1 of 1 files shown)


src/test/java/spoon/reflect/visitor/CtInheritanceScannerTest.java
```diff
@@ -65,23 +73,29 @@
...
+			Method mth=null;
+			
+			// if a method visitX exists, it must be invoked
+			try {				
 				mth = CtInheritanceScanner.class.getDeclaredMethod("visit" + intf.getSimpleName(), intf);
-
-				// we should not have a scan method
-				try {
-					CtInheritanceScanner.class.getDeclaredMethod("scan" + intf.getSimpleName(), intf);
-					Assert.fail("should not have both visit" + intf.getSimpleName() + " and scan" + intf.getSimpleName()
```
GerardPaligot: Why did you remove this assert?

GerardPaligot: You didn't check no more if we have both visit and scan methods for one interface.

petitpre:
I'm not agree with the contract you propose. For me:

- a visitXXX method should be define in CtVisitor for all concrete elements in model.
- a scanXXX method should be define in CtInheritanceScanner for all abstract concept in model.

so, for any new element in spoon model, you SHOULD have one visitXXX method CtVisitor if the element could be instanced, OR a scanXXX method in CtInheritanceScanner if it's an abstract concept.

monperrus:
I don't propose a new contract, I've purified the existing one.

> you SHOULD have one visitXXX method CtVisitor if the element could be
> instanced, OR a scanXXX method in CtInheritanceScanner if it's an
> abstract concept.
I do agree with this one, and this sentence can be put verbatim in the
test case.

Can you propose a PR on my PR to add it?

PR #278, opened 2015-07-07
Fix Type package with incomplete sources

diorcety:
I triggered an issue with packaging and incomplete sources.
For example if a class is missing (generated or something like that).
The file add a import

package com.abcd.project2
import com.abcd.project1.MyClass

void aa() {
  new MyClass();
}

The MyClass object will be reported as com.abcd.project2.MyClass. Even if the class is not present we explicitly have an import. Shouldn't we use the import?


monperrus:
It looks interesting, could you add a test case?

diorcety:
I will try

diorcety:
I added the test case. Without the fix i have

org.junit.ComparisonFailure: 
Expected :fr.inria.internal
Actual   :fr.inria

GerardPaligot:
Nice bug. Thanks for your PR. Unfortunately, as you can see, some checks have errors in Travis. Can you fix them or you need some help?

diorcety:
i will try. Thanks

diorcety:
I can't understand the final remaining issue...

GerardPaligot:
Hi @diorcety,

PR done on your private repository. It fixes error here.

monperrus:
Thanks for the nice contribution!

PR #280, opened 2015-07-07
Add quiet switch

diorcety:
Add quiet switch when output messages (incomplet sources for example) are too verbose

### DIFF
--- a/src/main/java/spoon/Launcher.java
+++ b/src/main/java/spoon/Launcher.java
@@ -186,6 +186,14 @@ public class Launcher implements SpoonAPI {
 			sw1.setHelp("Output messages about what the compiler is doing.");
 			jsap.registerParameter(sw1);
 
+			// Quiet
+			sw1 = new Switch("quiet");
+			sw1.setShortFlag('q');
+			sw1.setLongFlag("quiet");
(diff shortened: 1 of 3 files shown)


GerardPaligot:
We'll talk about this new feature at the next merge session to know if you want this feature or not.

Thanks for you PR!

monperrus:
Hi @diorcety, 

instead of having multiple boolean switches (verbose, quiet), whose interactions are often hard to understand if not counter intuitive, I propose transform the "verbose" item into an enum with for instance:
QUIET: absolutely nothing on standard output, even in case of errors
ERROR: outputs something only in case of non recoverable errors
INFO: displays information about the Spoon process
WARNING: displays warnings
DEBUG: maximal level of output for debug

In command-line, this would read:

   spoon -v QUIET -i ...
   spoon -v INFO,WARNING  -i ...

what do you think?

diorcety:
i agree

GerardPaligot:
@diorcety Do you want make it yourself or you prefer a PR on your branch?

diorcety:
I will off for one week. So please go for it with a new PR i will close mine at the right time.

PR #286, opened 2015-07-23
Set Type arguments in reference

diorcety:
The reference doesn't copy the CtMethod ActualTypeArguments

### DIFF
--- a/src/main/java/spoon/reflect/factory/ExecutableFactory.java
+++ b/src/main/java/spoon/reflect/factory/ExecutableFactory.java
@@ -106,8 +106,10 @@ public class ExecutableFactory extends SubFactory {
 			refs[i++] = param.getType();
 		}
 		if (e instanceof CtMethod) {
-			return createReference(((CtMethod<T>) e).getDeclaringType().getReference(),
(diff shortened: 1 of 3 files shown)


GerardPaligot:
Can you add a test case for this PR? 

Thanks in advance!

monperrus:
Hi @diorcety,

Thanks for the proposal. However, we don't fully see your problem.

getActualTypeArguments() almost never returns the formal type parameters.

Let's consider the following example:
```
public  <E> List<E> get() {...}

class Foo<F> {
main() {
  <F>get();
  <String>get(); 
}
```

for `<F>get();` getActualTypeArguments() returns [F]
for `<String>get();` getActualTypeArguments() returns [String]

I don't see yet a reason to have a default link between  formal type parameters and actual type arguments.

what's your initial problem?

diorcety:
The test show my issue. From a ExecutableReference i can't get any information about TypeArguments (formal in this case). Maybe the fix is not good, but there is a lack about the reference provided data

monperrus:
So what you want is `execRef.getDeclaration().getFormalTypeParameters()`?

this one should contain 'E'.

diorcety:
Oh yes maybe. I didn't use this code anymore. Sorry

PR #287, opened 2015-07-24
Fix NPE with properties directory is empty

diorcety:
getXmlRootFolder().listFiles() can return null
which will cause an NPE in the for loop


GerardPaligot:
Can you add a test case for this PR?

Thanks in advance!

diorcety:
Done. Sorry i'm a lazy guy :)

diorcety:
it's not empty. But non existing directory

monperrus:

PR #289, opened 2015-07-29
Added correct handling for enum's listed in an enum definition (CtEnum)

adamjryan:
- before the toString() method was simply used which does not take
account of the compilation unit's context.
* Removed the compute imports statement from CtElement's toString()
method as it has no context and therefore it cannot assume the
corresponding import is present that would allow it to be shown in
the shortened version.
* Added test to catch some edge cases in the shortening of identifiers
that may produce ambiguous/invalid code and updated ones that relied on
the toString() method to calculate the shorthand names.
* Made some minor modifications to remove some warnings in the code
as flagged by the IDE.

### DIFF
--- a/src/test/java/spoon/test/prettyprinter/DefaultPrettyPrinterTest.java
+++ b/src/test/java/spoon/test/prettyprinter/DefaultPrettyPrinterTest.java
@@ -1,6 +1,13 @@
 package spoon.test.prettyprinter;
 
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+import java.io.File;
+import java.util.List;
+
 import org.junit.Test;
+
 import spoon.Launcher;
 import spoon.compiler.SpoonCompiler;
 import spoon.compiler.SpoonResource;
@@ -9,25 +16,17 @@ import spoon.reflect.code.CtConstructorCall;
 import spoon.reflect.code.CtInvocation;
 import spoon.reflect.declaration.CtClass;
 import spoon.reflect.declaration.CtType;
-import spoon.reflect.declaration.ParentNotInitializedException;
 import spoon.reflect.factory.Factory;
 import spoon.reflect.reference.CtImplicitTypeReference;
 import spoon.reflect.reference.CtTypeReference;
+import spoon.reflect.visitor.DefaultJavaPrettyPrinter;
 import spoon.reflect.visitor.Query;
 import spoon.reflect.visitor.filter.TypeFilter;
-import spoon.support.reflect.declaration.CtClassImpl;
-import spoon.support.reflect.declaration.CtElementImpl;
 import spoon.test.TestUtils;
 import spoon.test.prettyprinter.testclasses.AClass;
 
-import java.io.File;
-import java.util.List;
-
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
-
 public class DefaultPrettyPrinterTest {
+	private static final String nl = System.lineSeparator();
 
 	@Test
 	public void printerCanPrintInvocationWithoutException() throws Exception {
@@ -49,26 +48,26 @@ public class DefaultPrettyPrinterTest {
 		CtInvocation<?> mathAbsInvocation = elements.get(1);
 		assertEquals("java.lang.Math.abs(message.length())", mathAbsInvocation.toString());
 	}
-	
+
 	@Test
 	public void superInvocationWithEnclosingInstance() throws Exception {
-		
+
 		/**
 		 * To extend a nested class an enclosing instance must be provided
 		 * to call the super constructor.
 		 */
-		
+
 		String sourcePath = "./src/test/resources/spoon/test/prettyprinter/NestedSuperCall.java";
 		List<SpoonResource> files = SpoonResourceHelper.resources(sourcePath);
 		assertEquals(1, files.size());
-		
+
 		SpoonCompiler comp = new Launcher().createCompiler();
 		comp.addInputSources(files);
 		comp.build();
-		
+
 		Factory factory = comp.getFactory();
 		CtType<?> theClass = factory.Type().get("spoon.test.prettyprinter.NestedSuperCall");
-		
+
 		assertTrue(theClass.toString().contains("nc.super(\"a\")"));
 	}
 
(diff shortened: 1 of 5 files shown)


monperrus:
Hi @adamjryan ,

Thanks for your proposal.

Your change in DefaultJavaPrettyPrinter looks interesting, it is not good to rely on toString there:
```
-					write(ctexpr.toString());
+					scan(ctexpr);
```

But it is somehow lost in other changes.

Could you write the smallest possible failing test case and PR for it?

Best,

--Martin

PS:
- could you put the interesting "minor modifications to remove some warnings in the code as flagged by the IDE" in a separate commit and PR?
- note that on lines 70-80 of DefaultPrettyPrinterTest, you remove the specification of CtElement.toString which is not desirable.

adamjryan:
The test was updated as the toString() method is also changed to not shorten type identifiers. This is because there are no import statements in the toString() return value and may be ambigous if the types contained a user defined 'Integer' type for example.
In order to get a string result that does factor in imports a Pretty Printer is used and a calcImports() is run on the object as seen in the test, where the user has chosen what imports are in the context of the result.
The previous failure can be seen by reverting the changes in CtElementImpl and DefaultJavaPrettyPrinter and running or debugging the new test.
I found the problem when it was prevously shortening the type identifier and as it clashed with a type with the same name defined locally it produced invalid code that needed the fully qualified type.

monperrus:
Hi @adamjryan ,

I start to see what you mean.

1) Can you update test autoImportUsesFullyQualifiedNameWhenImportedNameAlreadyPresent with strong explicit assertions so that we see the collision in the failing assertion?

2) I still don't see why you change testPrintAClassWithImports and testPrintAMethodWithImports, the specification is good here.

3) I understand there is a bug, but I'm not sure the solution is to remove the call to computeImports from ToString (the answer to 1) and 3) will help to find the best solutiojn)

--Martin

adamjryan:
For point 2), the computeImports() call is made outside the toString method because it is no longer performed by toString but the test is on the output produced when all the types shown in the output have been imported, so it still needs to be performed as part of the test.
The reason computeImports() is no longer in toString() is because the output it produces has no context and could convey something that may be incorrect to the user. For example the output of a  method could show:
public Integer getInteger(){
 return new Integer();
}
however fully typed it may be:
public my.notaninteger.Integer getInteger(){
 return new my.notaninteger.Integer();
}
If the user is in control of what is imported then they will be in control of what is shortened so if there is a type clash such as:
new my.notaninteger.Integer(java.lang.Integer.MAX_VALUE)
they can choose to shorten the standard library type if preferable.

The test will output a meaningful error as it outputs the exception from any compile error, I thought this might be more useful and specific in comparison to a string match verification although it does rely on the integrity of the compiler. Would a string comparision (similar to the other tests) be preferable?

monperrus:
- about computeImports() in toString(): not that even if we call computeImports(), toString() still outputs a fully qualified version

- about the tests, what is preferable is a strong and explicit assertion in an easily understandable test. sometimes it means a string comparison. could you rewrite yours?

adamjryan:
I've updated the tests.
The last assertion of the autoImportUsesFullyQualifiedNameWhenImportedNameAlreadyPresent test highlights the failure that I originally found. If the changes in DefaultJavaPrettyPrinter and CtElementImpl are reverted the following string will be produced in that test:
<pre>
    public enum ENUM {
    E1(<b>TypeIdentifierCollision.globalField</b>,TypeIdentifierCollision.ENUM.E1);
        final int NUM;
        final Enum<?> e;
        private ENUM(int num ,Enum<?> e) {
            NUM = num;
            this.e = e;
        }}
</pre>

This is incorrect because within the context of the TypeIdentifierCollision the reference to the field contained in another class also named TypeIdentifierCollision should be fully qualified.

I've put the related assertions in that one test which makes it quite long, I could put them into seperate tests otherwise I hope this is what your after!

monperrus:
Looks goods, thanks!

Can you remove the changes in testPrintAClassWithImports(), testPrintAMethodWithImports()? 

(In case you think those tests specify an incorrect behavior, please only change the expected value).

PS: PR #315  might solve your bug also.

adamjryan:
Those changes to those tests are neccessary otherwise they will fail. The imports are now not calculated in toString(). They are testing that the identifiers get correctly shortened when imports are caclulated so the expected value is correct however it requires the prettyPrinter.computeImports() call in order that the toString() of the pretty printer shows the correct result taking into account the imports that have been added.

adamjryan:
I've added another test that should demonstrate the above and also catch any regression.

monperrus:
> Those changes to those tests are neccessary otherwise they will fail.
can you only change the "expected" values in the assertions of those
tests then to see whether we agree with the specification change?

otherwise, it may mean that your fix is not the optimal one.

thanks for the long-running work on this PR,

--Martin

adamjryan:
Their expected values are correct.

monperrus:
OK. So you cannot change those tests, and we have to find a slightly
different version of the fix.

My gut feeling is that removing toString() in DJPP is good, but that
removing computeImpots in toString is not appropriate.

Any other idea?

--Martin

adamjryan:
I suppose what I'm saying is that computeImports in toString is wrong and that it shouldn't be there (as specified in the last test I just added), so... up to you :-).

GerardPaligot:
@adamjryan I made a [PR](https://github.com/adamjryan/spoon/pull/1) on our fork for this PR.

adamjryan:
Ok, thanks Gerard. There's a few conflicts with the latest revision, I think for the sake of getting the bug sorted I'll just remove the changes to CtElementImpl and update this when I've got a free moment.

adamjryan:
Made the updates requested, should be good to go now.

GerardPaligot:
@adamjryan I make a PR on your repo to remove all usages of the DefaultJavaPrettyPrinter class in the DefaultJavaPrettyPrinterTest class and there was a bug in `JDTTreeBuilder`, so I fixed it too.

adamjryan:
All seems good, thanks.

src/test/java/spoon/test/prettyprinter/DefaultPrettyPrinterTest.java
```diff
@@ -150,4 +152,87 @@ public void testPrintAMethodWithGeneric() throws Exception {
...
+		;
+		String computed = aClass.getMethodsByName("setFieldUsingExternallyDefinedEnumWithSameNameAsLocal").get(0).toString();
+		assertEquals( "the externally defined enum should be fully qualified to avoid a name clash with the local enum of the same name", expected, computed );
+
+		expected = //This is what is expected
+			"public void setFieldUsingLocallyDefinedEnum() {" +nl+
+			"    localField = ENUM.E1.ordinal();" +nl+
+			"}"
+		;
+		expected = //This is correct however it could be more concise.
```
monperrus: we should only keep one "expected"

src/test/java/spoon/test/prettyprinter/DefaultPrettyPrinterTest.java
```diff
@@ -150,4 +152,87 @@ public void testPrintAMethodWithGeneric() throws Exception {
...
+		computed = aClass.getMethodsByName("setFieldOfClassWithSameNameAsTheCompilationUnitClass").get(0).toString();
+		assertEquals( "The static field of an external type with the same identifier as the compilation unit should be fully qualified", expected, computed );
+
+		expected = //This is what is expected
+				"public void referToTwoInnerClassesWithTheSameName() {" +nl+
+				"    ClassA.VAR0 = ClassA.getNum();" +nl+
+				"    Class1.ClassA.VAR1 = Class1.ClassA.getNum();" +nl+
+				"}"
+			;
+		expected = //This is correct however it could be more concise.
```
monperrus: idem

monperrus:
Thanks for the PR.

PR #292, opened 2015-08-09
bug with generic invocation due to JDT (waiting for JDT fix)

boeserwolf91:
Type mismatch: cannot convert from Integer to double at D:\Programmieren\CubeEngine\spoon\src\test\java\spoon\test\casts\Castings.java:12

However this could be a jdt bug too. I didn't investigate it yet.

Don't merge it accidentally. :smile: The PR just shows a bug and doesn't fix it.

### DIFF
--- a/src/test/java/spoon/test/casts/CastTest.java
+++ b/src/test/java/spoon/test/casts/CastTest.java
@@ -1,12 +1,14 @@
 package spoon.test.casts;
 
 import static org.junit.Assert.assertEquals;
+import static spoon.test.TestUtils.build;
 
 import org.junit.Test;
 
 import spoon.reflect.code.CtVariableRead;
 import spoon.reflect.declaration.CtClass;
 import spoon.reflect.declaration.CtMethod;
+import spoon.reflect.declaration.CtType;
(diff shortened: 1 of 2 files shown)


boeserwolf91:
Maybe I should mention that the bug just occurs with this generic invocation.

invoking the test method with ```test(Integer.valueOf(3))``` or ```test(new Integer(3))``` works fine.

GerardPaligot:
Unfortunately, the bug don't come from Spoon but in JDT and the [bug](https://bugs.eclipse.org/bugs/show_bug.cgi?id=464496) is already reported on their bugtracker.

boeserwolf91:
okay thanks @GerardPaligot 
So this is going to be fixed in a few years (maybe) :smile:

GerardPaligot:
I hope no ... I [up their issue in their bugtracker](https://bugs.eclipse.org/bugs/show_bug.cgi?id=464496#c2).

GerardPaligot:
I create the [issue of this bug](https://github.com/INRIA/spoon/issues/328). This PR can be closed.

PR #297, opened 2015-08-17
feat: No more ParentNotInitializedException when we print elements.

GerardPaligot:


monperrus:
what's the underlying problem?

PR #301, opened 2015-08-18
Added imports and noclasspath to Ant Task

alcides:
Updates the Ant Task with two new features that are in the Launcher: with-imports and no-classpath.


monperrus:
Thanks.

PR #302, opened 2015-08-18
No innerclass import

alcides:
When "with-imports" option is turned on, the generated output code will add:

import pkg.Klass$InnerKlass

with code defined as
```java
package pkg
public class Klass {
public class InnerKlass {}
}
```

This pull requests prevents those imports from being generated (as in my case, I have the import of the Klass class). However, I believe that it should be replaced by pkg.Klass.InnerKlass, although I am not sure how this works with non-static inner classes.

### DIFF
--- a/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
+++ b/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
@@ -913,6 +913,7 @@ public class DefaultJavaPrettyPrinter implements CtVisitor, PrettyPrinter {
 	}
 
 	private void writeEnumField(CtField<?> f) {
+		visitCtNamedElement(f);
 		write(f.getSimpleName());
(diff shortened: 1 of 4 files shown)


GerardPaligot:
You have a merge commit in your PR. Can you remove it and add a test case for your PR?

monperrus:
This contribution looks valuable, thanks.

Could you rebase, remove the merge commit from the PR, and add a test case please? (so that the diff view of the PR is clear)

Best,

--Martin

GerardPaligot:
I [rebased your branch](https://github.com/GerardPaligot/spoon/tree/QA_no_innerclass_import) and removed the commit of @petitpre but impossible to reproduce your bug!

Do you have a sample? You can give it in this issue, I'll develop the test case.

monperrus:
cleaning the old pending PRs.

if the bug is still there, feel free to open an issue

PR #310, opened 2015-08-24
fixes a test which failed on windows.............

boeserwolf91:


GerardPaligot:
My bad, it's my fault. We should find a solution for windows users to avoid this kind of PR in the future.

boeserwolf91:
What about using Jenkins on a windows machine too? So you could trigger the tests on Linux and on Windows and ensure that the core works on two different OS.

GerardPaligot:
> What about using Jenkins on a windows machine too? So you could trigger the tests on Linux and on Windows and ensure that the core works on two different OS.

I didn't forget this idea. I created an issue [here](https://github.com/INRIA/spoon/issues/324).

PR #312, opened 2015-08-25
feat: Adds the possibility to set level for all loggers of Spoon.

GerardPaligot:
Closes #280


monperrus:
deprecate verbose/debug, in default command line doc

boeserwolf91:
What is the reason for deprecating them?

GerardPaligot:
@boeserwolf91 With the level argument, you can specify a level "WARN" to exclude "INFO" and "DEBUG" messages. That seems no more necessary to keep verbose and debug arguments.

src/main/java/spoon/processing/Severity.java
```diff
@@ -17,25 +17,43 @@
...
  * @see Environment#report(Processor, Severity,
- *      spoon.reflect.declaration.CtElement, String)
+ * spoon.reflect.declaration.CtElement, String)
  */
 public enum Severity {
 	/**
 	 * Error-level severity.
 	 */
-	ERROR,
+	ERROR {
```
boeserwolf91: btw, I don't like this enum. why overriding the toLevel method for every enum value? You could write ```ERROR(Level.ERROR)``` and add a constructor taking the Level as an argument. This way you don't have to override the method 3 times. you can write it once and use it for as many as enum values as you want. 

```java
private final Level level;

Severity(Level level) {
    this.level = level;
}

public Level toLevel() {
   return this.level;
}
```

boeserwolf91:
As long as you just "deprecate" the doc, its okay. But I would like to have the possibilty to specify debug as well.

src/main/java/spoon/processing/Severity.java
GerardPaligot: Yes, that's better. I'll change that.

GerardPaligot:
Unfortunately, deprecated features will be removed in a next release.

PR #313, opened 2015-08-25
feat: Prints always an element with its toString method.

### LINKED ISSUE #309: CtLement.toString should never throw an exception
CtLement.toString should never throw an exception. (bad for debug and robust Spoon applications)

Print the current partial toString instead and the stack trace after instead

GerardPaligot:
Closes #309

### DIFF
--- a/src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
+++ b/src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
@@ -58,8 +58,8 @@ import spoon.support.visitor.TypeReferenceScanner;
  */
 public abstract class CtElementImpl implements CtElement, Serializable , Comparable<CtElement>{
 
-	protected static final Logger logger = Logger
-			.getLogger(CtElementImpl.class);
+	protected static final Logger logger = Logger.getLogger(CtElementImpl.class);
+	public static final String ERROR_MESSAGE_TO_STRING = "Error in printing the node. One parent isn't initialized!";
 
 	// we don't use Collections.unmodifiableList and Collections.unmodifiableSet 
 	// because we need clear() for all set* methods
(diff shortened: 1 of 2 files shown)


src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
```diff
@@ -369,9 +369,13 @@ public void enter(CtElement e) {
...
-		DefaultJavaPrettyPrinter printer = new DefaultJavaPrettyPrinter(getFactory().getEnvironment());
-		printer.computeImports(this);
-		printer.scan(this);
+		DefaultJavaPrettyPrinter printer = new DefaultJavaPrettyPrinter(
+				getFactory().getEnvironment());
+		try {
+			printer.computeImports(this);
+			printer.scan(this);
+		} catch (ParentNotInitializedException ignore) {
+		}
```
monperrus: add something in the result string: <parent error to print the rest if the element>

GerardPaligot: Something like "Error occurred. Element not well printed."?

src/test/java/spoon/test/prettyprinter/DefaultPrettyPrinterTest.java
```diff
@@ -97,4 +100,15 @@ public void testPrintAMethodWithImports() throws Exception {
 				+ "}";
 		assertEquals(expected, aClass.getMethodsByName("aMethod").get(0).toString());
 	}
+
+	@Test
+	public void testParentNotInitializedInToString() throws Exception {
+		final Factory factory = TestUtils.createFactory();
+		final CtClass<Object> aClass = factory.Core().createClass();
+		try {
+			aClass.toString();
```
monperrus: add an assert on the exepcted string

PR #318, opened 2015-08-26
feat: Adds addProcessor(Processor) method in SpoonAPI.

### LINKED ISSUE #300: add addProcessor(Processor) in SpoonAPI
see http://lists.gforge.inria.fr/pipermail/spoon-discuss/2015-August/001794.html

GerardPaligot:
Closes #300

### DIFF
--- a/src/main/java/spoon/Launcher.java
+++ b/src/main/java/spoon/Launcher.java
@@ -40,7 +40,9 @@ import spoon.compiler.SpoonFile;
 import spoon.compiler.SpoonFolder;
 import spoon.compiler.SpoonResource;
 import spoon.compiler.SpoonResourceHelper;
+import spoon.processing.Processor;
 import spoon.processing.Severity;
+import spoon.reflect.declaration.CtElement;
 import spoon.reflect.factory.Factory;
 import spoon.reflect.factory.FactoryImpl;
 import spoon.reflect.visitor.DefaultJavaPrettyPrinter;
@@ -87,7 +89,8 @@ public class Launcher implements SpoonAPI {
 	private static JSAP jsapSpec;	
 	protected JSAPResult jsapActualArgs;
 
-	private List<String> processors = new ArrayList<String>();
+	private List<String> processorTypes = new ArrayList<String>();
+	private List<Processor<? extends CtElement>> processors = new ArrayList<Processor<? extends CtElement>>();
 
 	/**
 	 * A default program entry point (instantiates a launcher with the given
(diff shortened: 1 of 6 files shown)


src/main/java/spoon/Launcher.java
```diff
@@ -88,6 +90,7 @@
 	protected JSAPResult jsapActualArgs;
 
 	private List<String> processors = new ArrayList<String>();
+	private List<Processor<? extends CtElement>> processorsInstance = new ArrayList<Processor<? extends CtElement>>();
```
boeserwolf91: I do NOT like the name. Why not just call it "processors" without the word "Instance"?

src/main/java/spoon/Launcher.java
```diff
@@ -485,6 +493,14 @@ protected void processArguments() {
 	}
 
 	/**
+	 * Gets the list of processors instance to be initially applied during the
+	 * processing.
+	 */
+	protected List<Processor<? extends CtElement>> getProcessorsInstance() {
```
boeserwolf91: Why not call it getProcessors()?...
Is the access modifier with protected the right one? You could make it public and either return a unmodifiable list or an array.

src/main/java/spoon/Launcher.java
GerardPaligot: Me too but `processors` is declared above this line.

src/main/java/spoon/Launcher.java
GerardPaligot: `getProcessors()` already exists and these methods aren't exposed in the API.

src/main/java/spoon/Launcher.java
boeserwolf91: oh I see. I'm blind :smile: 
Anyway the name is horrible and sounds wrong. "processorInstances" would be the right way to go. But it's not much better.
Why not making it clearer in some way? Call processors -> processorNames and the processorInstances -> processors (the names aren't real processors, just the names)

src/main/java/spoon/Launcher.java
boeserwolf91: I can't see a getProcessors method. "just" getProcessorTypes()

src/main/java/spoon/Launcher.java
boeserwolf91: oh call the processors field processorTypes to math the getter method instead of processorNames

GerardPaligot: Good idea!

src/main/java/spoon/Launcher.java
GerardPaligot: Oh yes, that's `getProcessorTypes()`. I change it!

boeserwolf91: great. thank you. it looks better :+1:

src/main/java/spoon/SpoonAPI.java
```diff
@@ -22,7 +24,12 @@
 	 * Adds a processor (fully qualified name).
 	 */
 	public void addProcessor(String name);
-	
+
+	/**
+	 * Adds a instance of a processor.
```
monperrus: "an instance' the user is responsible for keeping a pointer to it for later retrieving some processing information

PR #341, opened 2015-09-16
Fixed array[index] on RHS of assignment being considered ArrayWrite

alcides:
Introduces test to validate if s[0] = s[1] has one write and one read.
Fixes the problem of considering all Ct{Variable, Field, Array}Access with an CtAssignment as a parent as Writes.

This is fixed by creating a context variable assigned that is turned on on the LHS of the assignment only.

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilder.java
+++ b/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilder.java
@@ -266,6 +266,8 @@ public class JDTTreeBuilder extends ASTVisitor {
 
 		boolean forupdate = false;
 
+		boolean assigned = false;
+
 		Stack<String> label = new Stack<String>();
 
 		boolean selector = false;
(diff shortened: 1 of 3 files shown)


src/main/java/spoon/support/compiler/jdt/JDTTreeBuilder.java
```diff
@@ -1851,7 +1862,17 @@ public boolean visit(CompoundAssignment compoundAssignment, BlockScope scope) {
...
+        context.assigned = true;
+        if ((compoundAssignment.lhs) != null)
+            compoundAssignment.lhs.traverse(this, scope);
+        
+        context.assigned = false;
+        if ((compoundAssignment.expression) != null)
+            compoundAssignment.expression.traverse(this, scope);
+        
+        context.arguments.pop();
+        return false;
```
GerardPaligot: Can you indent these lines?

GerardPaligot:
I posted a comment but that sounds good for me.

monperrus:
Thanks for the contribution

PR #367, opened 2015-10-08
chore: Updates travis to the new architecture and change mvn cmd.

GerardPaligot:
Now, Travis is using containers to build projects. To use this
new architecture, we must add "sudo: false". That's all.
And we change the mvn command from test to verify to apply
checkstyle plugin in travis builds.


GerardPaligot:
@monperrus Travis ok.

PR #373, opened 2015-10-14
Fix anonymous classname and CtClass.getConstructor

tdurieux:
The anonymous classname was not correct when there are more than 9 anonymous per class.

The method CtClass.getConstructor did not return the correct constructor when there are serveral contructors with array/matrix parameter.

### DIFF
--- a/src/main/java/spoon/support/reflect/declaration/CtClassImpl.java
+++ b/src/main/java/spoon/support/reflect/declaration/CtClassImpl.java
@@ -69,8 +69,7 @@ public class CtClassImpl<T extends Object> extends CtTypeImpl<T> implements CtCl
 		for (CtConstructor<T> c : constructors) {
 			boolean cont = c.getParameters().size() == parameterTypes.length;
 			for (int i = 0; cont && (i < c.getParameters().size()) && (i < parameterTypes.length); i++) {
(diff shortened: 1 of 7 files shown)


src/main/java/spoon/support/reflect/declaration/CtClassImpl.java
```diff
@@ -69,8 +69,7 @@ public void accept(CtVisitor v) {
 		for (CtConstructor<T> c : constructors) {
 			boolean cont = c.getParameters().size() == parameterTypes.length;
 			for (int i = 0; cont && (i < c.getParameters().size()) && (i < parameterTypes.length); i++) {
-				if (!c.getParameters().get(i).getType().getQualifiedName()
-						.equals(parameterTypes[i].getQualifiedName())) {
+				if (!parameterTypes[i].equals(c.getParameters().get(i).getType())) {
```
GerardPaligot: It is necessary to get constructor in an anonymous class?

GerardPaligot: I read your PR and you fix 2 bugs in your PR, you should create another PR for this bug.

GerardPaligot:
Very nice PR, thanks @tdurieux!

Issue #368, opened 2015-10-11
unable to spoon RxJava 


JakeWharton:
```
Caused by: java.lang.NullPointerException
	at spoon.support.compiler.jdt.JDTTreeBuilder.createType(JDTTreeBuilder.java:946)
	at spoon.support.compiler.jdt.JDTTreeBuilder.visit(JDTTreeBuilder.java:2972)
	at org.eclipse.jdt.internal.compiler.ast.TypeDeclaration.traverse(TypeDeclaration.java:1374)
	at spoon.support.compiler.jdt.JDTTreeBuilder.visit(JDTTreeBuilder.java:2511)
	at org.eclipse.jdt.internal.compiler.ast.QualifiedAllocationExpression.traverse(QualifiedAllocationExpression.java:604)
	at spoon.support.compiler.jdt.JDTTreeBuilder.visit(JDTTreeBuilder.java:2315)
	at org.eclipse.jdt.internal.compiler.ast.MessageSend.traverse(MessageSend.java:1031)
	at org.eclipse.jdt.internal.compiler.ast.Block.traverse(Block.java:139)
	at spoon.support.compiler.jdt.JDTTreeBuilder.visit(JDTTreeBuilder.java:1602)
	at org.eclipse.jdt.internal.compiler.ast.LambdaExpression.traverse(LambdaExpression.java:678)
	at spoon.support.compiler.jdt.JDTTreeBuilder.visit(JDTTreeBuilder.java:2315)
	at org.eclipse.jdt.internal.compiler.ast.MessageSend.traverse(MessageSend.java:1031)
	at spoon.support.compiler.jdt.JDTTreeBuilder.visit(JDTTreeBuilder.java:2219)
	at org.eclipse.jdt.internal.compiler.ast.LocalDeclaration.traverse(LocalDeclaration.java:301)
	at spoon.support.compiler.jdt.JDTTreeBuilder.visit(JDTTreeBuilder.java:2396)
	at org.eclipse.jdt.internal.compiler.ast.MethodDeclaration.traverse(MethodDeclaration.java:327)
	at spoon.support.compiler.jdt.JDTTreeBuilder.visit(JDTTreeBuilder.java:3053)
	at org.eclipse.jdt.internal.compiler.ast.TypeDeclaration.traverse(TypeDeclaration.java:1319)
	at org.eclipse.jdt.internal.compiler.ast.CompilationUnitDeclaration.traverse(CompilationUnitDeclaration.java:748)
[...]

PR #374, opened 2015-10-15
fix: Circular references with generic types.

GerardPaligot:
When you have an invocation to a method with this
kind of reference : <T extends Comparable<? extends T>>,
you got a circular reference and a StackOverflowException.
With this PR, we add a cache which detects circular references
in JDTTreeBuilder and uses this information in
CtScanner and DefaultJavaPrettyPrinter.

Closes #368

PR #378, opened 2015-10-16
fix: Parameters in a lambda expression.

GerardPaligot:
From now, we don't re-print type of parameters of lambda
if original sources don't specify it. We print the name
of these parameters and we save the expected type in the
AST.

Issue #368

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilder.java
+++ b/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilder.java
@@ -276,6 +276,8 @@ public class JDTTreeBuilder extends ASTVisitor {
 
 		boolean isGenericTypeExplicit = true;
 
+		boolean isLambdaParameterImplicitlyTyped = true;
+
 		/**
 		 * Stack of all parents elements
 		 */
@@ -370,6 +372,8 @@ public class JDTTreeBuilder extends ASTVisitor {
 
 		boolean bounds = false;
 
+		boolean isImplicit = false;
+
 		public CtTypeReference<?> getBoundedTypeReference(TypeBinding binding) {
 			bounds = true;
 			CtTypeReference<?> ref = getTypeReference(binding);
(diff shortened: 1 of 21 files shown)


src/main/java/spoon/support/reflect/reference/CtImplicitTypeReferenceImpl.java
```diff
@@ -2,12 +2,18 @@
 
 import spoon.reflect.reference.CtImplicitTypeReference;
 import spoon.reflect.reference.CtReference;
+import spoon.reflect.visitor.CtVisitor;
 
 public class CtImplicitTypeReferenceImpl<R> extends CtTypeReferenceImpl<R>
```
GerardPaligot: Moves to internal package when https://github.com/INRIA/spoon/pull/374 is merged.

src/main/java/spoon/support/compiler/jdt/JDTTreeBuilder.java
```diff
@@ -276,6 +276,8 @@ public String toString() {
 
 		boolean isGenericTypeExplicit = true;
 
+		boolean isWithType = true;
```
GerardPaligot: isLambdaParameterImplicitType

src/test/java/spoon/test/generics/GenericsTest.java
```diff
@@ -111,8 +111,8 @@ public void testDiamond1() {
 		CtField<?> f = clazz.getFields().get(0);
 		CtConstructorCall<?> val = (CtConstructorCall<?>) f.getDefaultExpression();
 
-		// the diamond is resolved to String
-		assertEquals("java.lang.String", val.getType().getActualTypeArguments().get(0).toString());
+		// the diamond is resolved to String but we don't print it, so we use the fully qualified name.
+		assertEquals("java.lang.String", val.getType().getActualTypeArguments().get(0).getQualifiedName());
```
GerardPaligot: Add a new assertion for "" -> toString

JakeWharton:
Failing test case:
```java
private static <T1, T2, R> Function<Object[], R> toFunction(BiFunction<? super T1, ? super T2, ? extends R> biFunction) {
    Objects.requireNonNull(biFunction);
    return a -> {
        if (a.length != 2) {
            throw new IllegalArgumentException("Array of size 2 expected but got " + a.length);
        }
        return ((BiFunction<Object, Object, R>)biFunction).apply(a[0], a[1]);
    };
}
```
Becomes:
```java
private static <T1, T2, R>java.util.function.Function<java.lang.Object[], R> toFunction(java.util.function.BiFunction<? super T1, ? super T2, ? extends R> biFunction) {
    java.util.Objects.requireNonNull(biFunction);
    return ([] a) -> {
        if ((a.length) != 2) {
            throw new java.lang.IllegalArgumentException(("Array of size 2 expected but got " + (a.length)));
        } 
        return ((java.util.function.BiFunction<java.lang.Object, java.lang.Object, R>)(biFunction)).apply(a[0], a[1]);
    };
}
```
(Note the erroneous `([] a)` declaration of the lambda param).

GerardPaligot:
Yep but this is due to another bug. PR coming very soon about that! :)

JakeWharton:
Ah, ok. Sorry. I should just let you do your thing! I thought it was related to omitting casts due to its placement.

PR #381, opened 2015-10-16
fix: CtTypeParameterReference doesn't print always full generic.

GerardPaligot:
Issue #368


GerardPaligot:
@monperrus if you're not yet on vacation, this PR is ok.

PR #385, opened 2015-10-20
Infered generic type

tdurieux:
The  ```<T> ``` at https://github.com/apache/commons-lang/blob/740c0f95fbd99cb7c07bcf7c54bc077c3ab27bd1/src/main/java/org/apache/commons/lang3/SerializationUtils.java#L267 is automatically inferred to  ```Object ``` by JTD. 

Thus the JDT method ```messageSend.genericTypeArguments()``` return the class  ```Object ``` and not  ```<T> ```. This implies that the Spoon AST is not correct.

This PR fix the Spoon AST but I cannot reproduce this error with a test.

Fill free to complete this PR with a test.

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilder.java
+++ b/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilder.java
@@ -2346,9 +2346,9 @@ public class JDTTreeBuilder extends ASTVisitor {
 					e.traverse(this, scope);
 				}
 			}
-			if (messageSend.genericTypeArguments() != null) {
-				for (TypeBinding typeBinding : messageSend.genericTypeArguments()) {
-					inv.getExecutable().addActualTypeArgument(references.getTypeReference(typeBinding));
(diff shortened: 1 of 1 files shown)


GerardPaligot:
Well, I can't reproduce the bug too. I imported the full class in a test case, I applied spoon on it and everything works fine. You had to meet a very specific case ...

But the modification isn't stupid because JDT use `typeArguments` and not `genericTypeArguments()` when [it prints invocations with generic types](http://grepcode.com/file/repo1.maven.org/maven2/org.eclipse.jdt.core.compiler/ecj/4.2.1/org/eclipse/jdt/internal/compiler/ast/MessageSend.java#368).

What do you think @monperrus?

GerardPaligot:
Well, after more test with @tdurieux, it seems that there is a new bug in JDT and linked at the environment. When @tdurieux and me apply the same version of spoon on commons-lang project, he gets the error and not me. Thanks Eclipse ... :)

PR #386, opened 2015-10-20
Fix super constructor call of the parameter

tdurieux:


src/test/java/spoon/test/constructor/ConstructorTest.java
```diff
@@ -51,4 +52,11 @@ public void testTransformationOnConstructorWithInsertBefore() throws Exception {
 		assertEquals(1, ctConstructor.getBody().getStatements().size());
 		assertEquals("super()", ctConstructor.getBody().getStatement(0).toString());
 	}
+
+	@Test
+	public void callParamConstructor() throws Exception {
+		CtClass<Object> aClass = factory.Class().get(AClass.class);
+		CtConstructor<Object> constructor = aClass.getConstructors().iterator().next();
+		assertEquals("{\n    enclosingInstance.super();\n}", constructor.getBody().toString());
```
GerardPaligot: Can you change '\n' character for `System.lineSeparator()` invocation? Think windows users! :)

tdurieux: Done

PR #389, opened 2015-10-20
Print  the full qualified name of the variable type when the parent of the type has generics

tdurieux:


src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
```diff
@@ -1891,7 +1891,8 @@ public void visitCtCircularTypeReference(CtCircularTypeReference reference) {
 
 	private <T> boolean hasDeclaringTypeWithGenerics(CtTypeReference<T> reference) {
 		// If current reference use generic types, we don't need this hack.
-		if (reference.getActualTypeArguments().size() != 0) {
+		if (reference.getActualTypeArguments().size() != 0
+				&& reference.getDeclaringType().getActualTypeArguments().size() == 0) {
```
GerardPaligot: Can you move this check below this [check](https://github.com/INRIA/spoon/pull/389/files#diff-5f33bf5d1f38230de815e4c60a67a215R1906)?

tdurieux: The complete if or just my additional condition?

GerardPaligot: Your additional condition. Potentially, `reference.getDeclaringType()` can be null.

tdurieux: No, I cannot. but I can move the complete if

GerardPaligot: Move the complete if.

PR #390, opened 2015-10-21
nested class NPE

revaultch:
Added the test class  src/test/java/spoon/test/SampleNestedClassExtending.java that crashes spoon with a NPE. 
Tentative fix included, however not sure about the impact.
Regards,

### DIFF
--- a/src/test/java/spoon/test/imports/ImportTest.java
+++ b/src/test/java/spoon/test/imports/ImportTest.java
@@ -71,7 +71,9 @@ public class ImportTest {
 		String expected = "visibility.YamlRepresenter.RepresentConfigurationSection";
 		assertEquals(expected, innerClass.getReference().toString());
 
-		expected = "org.yaml.snakeyaml.representer.Representer.RepresentMap";
+		// was org.yaml.snakeyaml.representer.SafeRepresenter.RepresentMap
+		// but RepresentMap is defined in SafeRepresenter
+		expected = "org.yaml.snakeyaml.representer.SafeRepresenter.RepresentMap";
 		assertEquals(expected, innerClass.getSuperclass().toString());
 	}
 
--- a/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilder.java
+++ b/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilder.java
@@ -955,7 +955,7 @@ public class JDTTreeBuilder extends ASTVisitor {
 							(ReferenceBinding) typeDeclaration.enclosingType.superclass.resolvedType);
 				} else if (resolvedType instanceof BinaryTypeBinding && resolvedType.enclosingType() != null && !getModifiers(resolvedType.enclosingType().modifiers).containsAll(modifiers)) {
 					typeDeclaration.superclass.resolvedType = new SpoonReferenceBinding(typeDeclaration.superclass.resolvedType.sourceName(),
-							(ReferenceBinding) typeDeclaration.enclosingType.superclass.resolvedType);
+							(ReferenceBinding) typeDeclaration.superclass.resolvedType.enclosingType());
 				}
 			}
 			if (typeDeclaration.superclass != null) {
--- a/src/test/java/spoon/test/SampleNestedClassExtending.java
+++ b/src/test/java/spoon/test/SampleNestedClassExtending.java
@@ -1,0 +1,12 @@
+package spoon.test;
+
+public class SampleNestedClassExtending {
+
+    static class TextAttribute extends java.text.AttributedCharacterIterator.Attribute {
+
+        protected TextAttribute(String name) {
+            super(name);
+        }
+    }
+
+}


src/test/java/spoon/test/imports/ImportTest.java
```diff
@@ -71,7 +71,9 @@ public void testImportOfAnInnerClassInASuperClassAvailableInLibrary() throws Exc
 		String expected = "visibility.YamlRepresenter.RepresentConfigurationSection";
 		assertEquals(expected, innerClass.getReference().toString());
 
-		expected = "org.yaml.snakeyaml.representer.Representer.RepresentMap";
+		// was org.yaml.snakeyaml.representer.SafeRepresenter.RepresentMap
+		// but RepresentMap is defined in SafeRepresenter
```
GerardPaligot: Yes but `SafeRepresenter` isn't public. We cannot print `org.yaml.snakeyaml.representer.SafeRepresenter.RepresentMap` because the Java source code wont be valid.

GerardPaligot:
Hi @revaultch and thanks for your first PR!

I commented your PR because you changed a good test case but I suppose there is a real bug behind your PR. Maybe you can write a test case where you expose your bug. In your PR, I see the class `SampleNestedClassExtending` but I didn't see any new test cases and/or new assertions.

revaultch:
If you add SampleNestedClassExtending.java the test `spoon.test.main.MainTest` will fail. I haven't added a separate test case. If you just retrieve this file and do a mvn install you will have test failures. Concerning the test I modified, the following command:

    unzip -l src/test/resources/visibility/snakeyaml-1.9.jar|grep RepresentMap
       1207  08-12-11 05:36   org/yaml/snakeyaml/representer/SafeRepresenter$RepresentMap.class

shows that RepresentMap is defined in SafeRepresenter. To me its qualified name is therefore org.yaml.snakeyaml.representer.SafeRepresenter.RepresentMap and not rg.yaml.snakeyaml.representer.Representer.RepresentMap.

revaultch:
Sorry for adding another commit to this, made a bad manipulation on Github, should be two pulls.

GerardPaligot:
> shows that `RepresentMap` is defined in `SafeRepresenter`. To me its qualified name is therefore `org.yaml.snakeyaml.representer.SafeRepresenter.RepresentMap` and not `org.yaml.snakeyaml.representer.Representer.RepresentMap`.

Well, you're right but there is a biggest bug behind this. If we have these classes:

```java
class SuperClassNotPublic {
  protected class InnerClassFromSuper {
  }
}
```

```java
public class ChildClassPublic extends SuperClassNotPublic{
}
```

```java
public class Client extends ChildClassPublic {
  public class InnerClass extends ChildClassPublic.InnerClassFromSuper {
  }
}
```

Where `SuperClassNotPublic` and `ChildClassPublic` are in a different package (and/or in a jar file). In this specific case (exposed in the test cases ImportTest#testImportOfAnInnerClassInASuperClassPackage and ImportTest#testImportOfAnInnerClassInASuperClassAvailableInLibrary), these classes reprinted by spoon don't compile anymore.

So, what you said is right but the bug behind is tricky! :)

> Sorry for adding another commit to this, made a bad manipulation on Github, should be two pulls.

No problem but can you cherry pick your commit and squash this PR?

revaultch:
OK. I see your problem now, how do you test this? I mean mvn install passed for me. I guess not qualifying the class could be an option if they share a common ancestry. I guess there is a similar problem with static fields accessed through a child class.

GerardPaligot:
It lacks an assertion. I changed the test case to add assertion `TestUtils.canBeBuild(target, 8);` at the end of the test and to check that the source code printed by spoon compiles or not.

```java
@Test
public void testImportOfAnInnerClassInASuperClassAvailableInLibrary() throws Exception {
	final String target = "./target/visibility-import-jar/";
	launcher.getEnvironment().setSourceClasspath(new String[] {"./src/test/resources/visibility/snakeyaml-1.9.jar"});
	launcher.addInputResource("./src/test/resources/visibility/YamlRepresenter.java");
	launcher.setSourceOutputDirectory(target);
	launcher.run();

	CtType<?> theClass = factory.Type().get("visibility.YamlRepresenter");

	final CtClass<?> innerClass = theClass.getNestedType("RepresentConfigurationSection");
	String expected = "visibility.YamlRepresenter.RepresentConfigurationSection";
	assertEquals(expected, innerClass.getReference().toString());

	// was org.yaml.snakeyaml.representer.SafeRepresenter.RepresentMap
	// but RepresentMap is defined in SafeRepresenter
	expected = "org.yaml.snakeyaml.representer.SafeRepresenter.RepresentMap";
	assertEquals(expected, innerClass.getSuperclass().toString());

	TestUtils.canBeBuild(target, 8);
}
```

Unfortunately, the source code doesn't compile.

GerardPaligot:
There is a good solution for this PR but it isn't this one. We should save a new information in the AST like "accessing element". If we create this new information, we can use it when we rewrite the source code and we can access at the right value in `getSuperClass()`.

This work seems a little bit to hard for a new contributor. I'll make a PR as soon as possible. See https://github.com/INRIA/spoon/issues/398 for the progression of this work.

I close this PR.

PR #397, opened 2015-11-02
first revival of TemplateMatcher

monperrus:
TemplateMatcher is an old and great feature of Spoon, that needed to be cleaned and tested. This is the beginning, see test case.


GerardPaligot:
That should be nice to have a piece of documentation about this "new" feature. :)

PR #402, opened 2015-11-05
refactor(api): Deprecates process method for a new one.

GerardPaligot:
In SpoonModelBuilder, we had two methods very semilar in their
signature. To avoid confusing these methods, we deprecates one of
them with a method name more explicit.

### DIFF
--- a/src/main/java/spoon/SpoonModelBuilder.java
+++ b/src/main/java/spoon/SpoonModelBuilder.java
@@ -17,15 +17,15 @@
 
 package spoon;
 
-import java.io.File;
-import java.util.Collection;
-import java.util.List;
-import java.util.Set;
-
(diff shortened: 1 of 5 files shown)


src/main/java/spoon/SpoonModelBuilder.java
```diff
@@ -146,12 +146,19 @@
...
 	 * Processes the Java model with the given processors.
+	 * @see #processType(List)
 	 */
+	@Deprecated
 	void process(List<String> processorTypes);
 
 	/**
 	 * Processes the Java model with the given processors.
 	 */
+	void processType(List<String> processors);
```
monperrus: instantiateAndProcess

GerardPaligot:
@monperrus Everything is ok here.

PR #414, opened 2015-11-18
Fix a bug in ImportScannerImpl.visitCtInvocation

leventov:
shouldn't import invoked method declaration classes

### DIFF
--- a/src/main/java/spoon/reflect/visitor/ImportScannerImpl.java
+++ b/src/main/java/spoon/reflect/visitor/ImportScannerImpl.java
@@ -63,12 +63,10 @@ public class ImportScannerImpl extends CtScanner implements ImportScanner {
 
 	@Override
 	public <T> void visitCtInvocation(CtInvocation<T> invocation) {
-		// For a ctinvocation, we don't have to import declaring type
 		enter(invocation);
 		scan(invocation.getAnnotations());
 		scanReferences(invocation.getTypeCasts());
 		scan(invocation.getTarget());
-		scan(invocation.getExecutable());
 		scan(invocation.getArguments());
 		exit(invocation);
 	}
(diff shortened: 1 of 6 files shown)


src/main/java/spoon/reflect/visitor/ImportScannerImpl.java
```diff
@@ -63,12 +63,10 @@
 
 	@Override
 	public <T> void visitCtInvocation(CtInvocation<T> invocation) {
-		// For a ctinvocation, we don't have to import declaring type
 		enter(invocation);
 		scan(invocation.getAnnotations());
 		scanReferences(invocation.getTypeCasts());
 		scan(invocation.getTarget());
-		scan(invocation.getExecutable());
```
tdurieux: I already try to remove this line but some time this line is required.
I don't remember in which case (may be in static call).

GerardPaligot:
@leventov Can you add a test case where it is useful to avoid scanning executable of an invocation?

leventov:
In my case this tries to import package-private class, if the called method is originally defined in it, e. g.

```java
package foo;
class Foo {
    public void foo() {}
}

public class Bar extends Foo {
}
```

```java
package other.package;

bar.foo();
```

When spoon compiles imports for such, it includes `import foo.Foo;` that is illegal.

src/main/java/spoon/reflect/visitor/ImportScannerImpl.java
leventov: @tdurieux static call target (class name) should be processed by the prev line `scan(invocation.getTarget())`, isn't it?

GerardPaligot: yes, `getTarget()` is used for static invocation.

GerardPaligot:
@leventov I made a PR on your repository. Can you accept it to merge your PR here?

leventov:
@GerardPaligot thank for making that ground work instead of me :)

GerardPaligot:
No problem @leventov :)

@monperrus I contributed on this PR. I let you merge.

PR #421, opened 2015-11-23
Fix the position in the spoon model

tdurieux:


src/test/java/spoon/test/position/PositionTest.java
```diff
@@ -0,0 +1,225 @@
...
+				content+=line + System.lineSeparator();
+			}
+		} catch (IOException x) {
+			System.err.format("IOException: %s%n", x);
+		}
+
+		return content;
+	}
+
+	private String substring(String content, SourcePosition position) {
```
monperrus: rename in contentAtPosition?

PR #423, opened 2015-11-23
feat(delete): Allows delete any element in a Spoon AST.

### LINKED ISSUE #403: add CtStatement.delete()
It's been a while that I encounter transformations where we need to remove some code. This is only valid for statements and not for expressions. So one solution is to have a CtStatement.delete()

GerardPaligot:
Closes #403

### DIFF
--- a/src/main/java/spoon/reflect/declaration/CtElement.java
+++ b/src/main/java/spoon/reflect/declaration/CtElement.java
@@ -218,4 +218,9 @@ public interface CtElement extends FactoryAccessor, CtVisitable {
 	 * be called to check and fix parents after manipulating the model.
 	 */
 	void updateAllParentsBelow();
+
+	/*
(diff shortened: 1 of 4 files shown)


src/main/java/spoon/reflect/declaration/CtElement.java
```diff
@@ -224,4 +224,8 @@
 	 */
 	<E extends CtElement> E setAnnotations(List<CtAnnotation<? extends Annotation>> annotation);
 
+	/**
+	 * Deletes the statement in its block.
```
monperrus: Deletes the element. For instance, delete a statement from its containing block. Warning: it may result in an incorrect AST, use at your own risk.

GerardPaligot:
@monperrus PR okay here.

GerardPaligot:
@monperrus PR ok.

PR #428, opened 2015-12-01
fix(noclasspath): Supports noclasspath for parameters.

GerardPaligot:


src/test/java/spoon/test/parameters/ParameterTest.java
```diff
@@ -0,0 +1,29 @@
...
+	@Test
+	public void testParameterInNoClasspath() throws Exception {
+		final Launcher launcher = new Launcher();
+		launcher.addInputResource("./src/test/resources/parameter");
+		launcher.setSourceOutputDirectory("./target/parameter");
+		launcher.getEnvironment().setNoClasspath(true);
+		launcher.run();
+
+		final CtClass<Object> aClass = launcher.getFactory().Class().get("org.eclipse.draw2d.text.FlowUtilities");
+		final CtParameter parameter = aClass.getElements(new TypeFilter<>(CtParameter.class)).get(2);
```
monperrus: why is it get(2)? it seems that it should be get(3)? (another bug?)

should we replace with a NameFilter or a custom filter?

GerardPaligot:
@monperrus PR ok.

PR #434, opened 2015-12-02
fix(noclasspath): superclass of enclosing class of static class was null

GerardPaligot:


GerardPaligot:
@monperrus if ok for you, PR ok.

PR #440, opened 2015-12-04
fix(scanner): Changes all enter/exitReferences inv to enter/exit.

### LINKED ISSUE #437: Changes enter/exitReference to enter/exit in CtScanner for references

GerardPaligot:
Closes #437

### DIFF
--- a/src/main/java/spoon/reflect/visitor/CtScanner.java
+++ b/src/main/java/spoon/reflect/visitor/CtScanner.java
@@ -116,24 +116,10 @@ public abstract class CtScanner implements CtVisitor {
 	}
 
 	/**
-	 * This method is upcalled by the scanner when entering a scanned element
-	 * reference. To be overridden to implement specific scanners.
-	 */
-	protected void enterReference(CtReference e) {
-	}
-
-	/**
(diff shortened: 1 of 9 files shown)


src/test/java/spoon/test/template/TemplateTest.java
```diff
@@ -107,7 +107,8 @@ public void testTemplateC1() throws Exception {
 		assertEquals("return \"testparam2\"", m2.getBody().getStatement(0)
 				.toString());
 
-		new ModelConsistencyChecker(factory.getEnvironment(), false, true).scan(c1);
+		// fixInconsistencies is set to true because references don't have a null parent.
```
monperrus: > because references don't have a null parent.
This is surprising.

All references now have a parent right?

GerardPaligot: Yes, they have a parent. Before the application of the scanner `ModelConsistencyChecker`, the parent of references are elements of the template. The scanner set the parent after the substitution of templates and put the target parent in references.

GerardPaligot:
@monperrus PR ok.

PR #441, opened 2015-12-04
[fix #431] Change the line number of all CtNamedElements

### LINKED ISSUE #431: bug in line numbers
line numbers are now wrong, because of a corner case of 4b3aa3a83a2457ffa9d3affc3046c8d9cf2bdcf8

tdurieux:
The line number of a ```CtNamedElement``` is now the line where the developer declares the name.


monperrus:
seems fine to me. thanks.  @GerardPaligot, if you also validate, you can proceed with merge

GerardPaligot:
Can you add more testcase? You create a test case for methods with javadoc but you didn't without javadoc, with class and field.

tdurieux:
This case is already tested by the test suite.

PR #443, opened 2015-12-05
[fix #304] Refactor Snippet helper

### LINKED ISSUE #304: createCodeSnippetExpression/Statement can only be used once.
```factory.Code().createCodeSnippetExpression(...).compile()``` and its Statement counterpart can only be used once, even in different Processors.

I want to be able to generate two or more snippets in order to build a new AST to modify the original program. These are small things that are not enough to justify a template, but take 10-20 lines to create manually.

tdurieux:
close #304, need a test case.


GerardPaligot:
@monperrus PR ok.

PR #444, opened 2015-12-07
Create a code factory for the constructor calls

tdurieux:


src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
```diff
@@ -1604,7 +1607,12 @@ public void visitCtNamedElement(CtNamedElement e) {
 	}
```
monperrus: add contract when the top level does not exist

src/main/java/spoon/reflect/factory/CodeFactory.java
```diff
@@ -103,6 +105,33 @@ public CodeFactory(Factory factory) {
 	}
 
 	/**
+	 * Creates a constructor call.
+	 *
```
monperrus: add note about inference of correct constructor based on parameters

Issue #393, opened 2015-10-23
Updates SpoonFormatterPrefs.xml for the new formatting


GerardPaligot:
`SpoonFormatterPrefs.xml` wasn't maintained when we integrated checkstyle in the project.

PR #450, opened 2015-12-08
Fix bug related to "insertBefore()" and "getSignature()"

maxcleme:
There is the same bug on insertAfter, but i didn't succeed to fix it.

### DIFF
--- a/src/main/java/spoon/support/reflect/code/CtStatementImpl.java
+++ b/src/main/java/spoon/support/reflect/code/CtStatementImpl.java
@@ -183,7 +183,13 @@ public abstract class CtStatementImpl extends CtCodeElementImpl implements CtSta
 
 			@Override
 			void insertFromFirstStatement(CtBlock<?> block, CtStatement target, CtStatementList statementsToBeInserted) {
-				int indexOfTargetElement = block.getStatements().indexOf(target);
+			  int indexOfTargetElement = -1;
+        for (int i = 0; i < block.getStatements().size(); i++) {
+          if (block.getStatements().get(i) == target) {
+            indexOfTargetElement = i;
+            break;
+          }
+        }
 				for (CtStatement s : statementsToBeInserted) {
 					s.setParent(block);
 					block.getStatements().add(indexOfTargetElement++, s);
--- a/src/main/java/spoon/support/visitor/SignaturePrinter.java
+++ b/src/main/java/spoon/support/visitor/SignaturePrinter.java
@@ -230,6 +230,9 @@ public class SignaturePrinter implements CtVisitor {
 	public <E> void visitCtCase(CtCase<E> caseStatement) {
 		write("case (");
 		scan(caseStatement.getCaseExpression());
+		for (CtStatement statement : caseStatement.getStatements()) {
+      scan(statement);
+    }
 		write(")");
 	}
 
--- a/src/test/java/spoon/test/processing/ProcessingTest.java
+++ b/src/test/java/spoon/test/processing/ProcessingTest.java
@@ -3,11 +3,14 @@ package spoon.test.processing;
 import org.junit.Test;
 import spoon.Launcher;
 import spoon.SpoonException;
+import spoon.reflect.code.CtSwitch;
 import spoon.reflect.declaration.CtClass;
 import spoon.reflect.declaration.CtConstructor;
 import spoon.reflect.declaration.CtMethod;
+import spoon.reflect.visitor.filter.TypeFilter;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.fail;
 import static spoon.test.TestUtils.build;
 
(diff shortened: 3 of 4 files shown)


src/main/java/spoon/support/reflect/code/CtStatementImpl.java
```diff
@@ -183,7 +183,13 @@ void insert(CtBlock<?> block, CtStatementList statementsToBeInserted) {
 
 			@Override
 			void insertFromFirstStatement(CtBlock<?> block, CtStatement target, CtStatementList statementsToBeInserted) {
-				int indexOfTargetElement = block.getStatements().indexOf(target);
+			  int indexOfTargetElement = -1;
```
monperrus: oops? indentation with tabs? (this is the cause of travis' failure)

src/main/java/spoon/support/visitor/SignaturePrinter.java
```diff
@@ -230,6 +230,9 @@ public void visitCtBreak(CtBreak breakStatement) {
 	public <E> void visitCtCase(CtCase<E> caseStatement) {
 		write("case (");
 		scan(caseStatement.getCaseExpression());
+		for (CtStatement statement : caseStatement.getStatements()) {
```
GerardPaligot: Change to `scan(caseStatement.getStatements())`

src/test/java/spoon/test/processing/ProcessingTest.java
```diff
@@ -41,6 +44,20 @@ public void testInsertBegin() throws Exception {
...
 		}
+			
+		CtConstructor<?> constructor = type.getConstructor(type.getFactory().Type().INTEGER_PRIMITIVE);
+    String myBeforeStatementAsString = "int before";
+    for (CtSwitch<?> ctSwitch : constructor.getElements(new TypeFilter<CtSwitch<?>>(CtSwitch.class))) {
+      ctSwitch.insertBefore(type.getFactory().Code()
+        .createCodeSnippetStatement(myBeforeStatementAsString));
+    }
+    assertEquals("insert has not been done at the right position", myBeforeStatementAsString, constructor.getBody().getStatement(3).toString());
+    assertEquals("insert has not been done at the right position", myBeforeStatementAsString, constructor.getBody().getStatement(5).toString());
```
GerardPaligot: `constructor.getBody().getStatement(5)` doesn't seem to be a switch at this position. Why there is an insertion here?

src/main/java/spoon/support/reflect/code/CtStatementImpl.java
maxcleme: Hi,

I just tried to use the formatter included with the spoon' repository ( SpoonFormatterPrefs.xml ), however when I apply it to the project, 458 files are modified. Do you really use this formatter ?

src/test/java/spoon/test/processing/ProcessingTest.java
```diff
@@ -41,6 +44,20 @@ public void testInsertBegin() throws Exception {
...
+    assertEquals("insert has not been done at the right position", myBeforeStatementAsString, constructor.getBody().getStatement(7).toString());
+
+    assertFalse("switch should not be the same", constructor.getBody().getStatement(6).equals(constructor.getBody().getStatement(8)));
```
GerardPaligot: This assertion was red before your fix?

src/main/java/spoon/support/reflect/code/CtStatementImpl.java
GerardPaligot: Nope: https://github.com/INRIA/spoon/issues/393

src/test/java/spoon/test/processing/ProcessingTest.java
maxcleme: Yes

src/test/java/spoon/test/processing/ProcessingTest.java
maxcleme: It is, I add them to the SampleForInsertBefore.java file.

src/main/java/spoon/support/visitor/SignaturePrinter.java
maxcleme: There is no scan method for List<CtStatement>

src/test/java/spoon/test/processing/ProcessingTest.java
GerardPaligot: If a copy paste the constructor here:

```java
	public SampleForInsertBefore(int j) {
		this(j, 0);
		new Thread() {
		};
		switch (j) {
      default:
        break;
    }
    switch (j) {
      default:
        break;
    }
    switch (j) {
      default: {
        break;
      }
    }
```

We can see that the position 5 is a `break`.

src/main/java/spoon/support/visitor/SignaturePrinter.java
monperrus: > Change to |scan(caseStatement.getStatements())|
>
|it's not equivalent, the scan does not work, we trued it first :-)|

src/test/java/spoon/test/processing/ProcessingTest.java
tdurieux: No, the statement 5 is not a break, because getStatements only returns the statements which have the method block as parent.

tdurieux:
Too clarify this fix:
The error was in this line ```int indexOfTargetElement = block.getStatements().indexOf(target);``` because ```indexOf``` uses the equality between two objects  but in this case we are looking for the reference equality (that was the case before https://github.com/INRIA/spoon/commit/6703c2e77025eddca0fd94fe7b5600f2f68f08bc#diff-0e647d5fcf8e2eea99bd6087436a08baL190).
The bug is probably also present in other locations of this class.

monperrus:
> that was the case before
> 6703c2e#diff-0e647d5fcf8e2eea99bd6087436a08baL190
> <https://github.com/INRIA/spoon/commit/6703c2e77025eddca0fd94fe7b5600f2f68f08bc#diff-0e647d5fcf8e2eea99bd6087436a08baL190>
fun! a test was clearly missing :-)

tdurieux:
I fixed the insertEnd and insertAfter in this branch https://github.com/tdurieux/spoon/tree/fix-insert

monperrus:
after discussion with @maxcleme , he lets us resolve the formatting (grrr #393) and add the missing case.
@GerardPaligot , do you handle this?

GerardPaligot:
@tdurieux already did it this morning: https://github.com/maxcleme/spoon/pull/1

monperrus:
could you close this PR and check and accept the other one?

GerardPaligot:
The PR of @tdurieux isn't on the spoon project. It is a PR on this PR. @maxcleme must accept it and this PR will be automatically updated.

PR #451, opened 2015-12-09
test(scanner): Check scan and visit methods in CtInheritanceScanner.

### LINKED ISSUE #276: add implementation of CtInheritanceScanner contract
follow up of #269 

>  you SHOULD have one visitXXX method CtVisitor if the element could be instanced, OR a scanXXX method in CtInheritanceScanner if it's an abstract concept.

GerardPaligot:
Closes #276

### DIFF
--- a/src/test/java/spoon/reflect/visitor/CtVisitorTest.java
+++ b/src/test/java/spoon/reflect/visitor/CtVisitorTest.java
@@ -21,13 +21,13 @@ import org.junit.Test;
 import spoon.Launcher;
 import spoon.reflect.visitor.processors.CheckVisitorProcessor;
 
-public class CtInheritanceScannerMethodsTest {
+public class CtVisitorTest {
 	@Test
-	public void testMethodsInInheritanceScanner() throws Exception {
-		// contract: CtInheritanceScanner must declare all scanner and visitor methods.
+	public void testMethodsInVisitor() throws Exception {
+		// contract: CtVisitor must declare all visit methods.
 		final Launcher launcher = new Launcher();
 		launcher.getEnvironment().setNoClasspath(true);
-		launcher.addProcessor(new CheckVisitorProcessor(CtInheritanceScanner.class).withScanners().withVisitors());
+		launcher.addProcessor(new CheckVisitorProcessor(CtVisitor.class).withVisitors());
 		launcher.setSourceOutputDirectory("./target/trash");
 		// interfaces.
 		launcher.addInputResource("./src/main/java/spoon/reflect/code");
@@ -37,7 +37,7 @@ public class CtInheritanceScannerMethodsTest {
 		launcher.addInputResource("./src/main/java/spoon/support/reflect/code");
 		launcher.addInputResource("./src/main/java/spoon/support/reflect/declaration");
 		launcher.addInputResource("./src/main/java/spoon/support/reflect/reference");
-		launcher.addInputResource("./src/main/java/spoon/reflect/visitor/CtInheritanceScanner.java");
+		launcher.addInputResource("./src/main/java/spoon/reflect/visitor/CtVisitor.java");
 		launcher.run();
 
 		// All assertions are in the processor.
--- a/src/test/java/spoon/reflect/visitor/processors/CheckModelProcessor.java
+++ b/src/test/java/spoon/reflect/visitor/processors/CheckModelProcessor.java
@@ -1,0 +1,76 @@
+/*
+ * Copyright (C) 2006-2015 INRIA and contributors
+ * Spoon - http://spoon.gforge.inria.fr/
+ *
+ * This software is governed by the CeCILL-C License under French law and
+ * abiding by the rules of distribution of free software. You can use, modify
+ * and/or redistribute the software under the terms of the CeCILL-C license as
+ * circulated by CEA, CNRS and INRIA at http://www.cecill.info.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the CeCILL-C License for more details.
+ *
+ * The fact that you are presently reading this means that you have had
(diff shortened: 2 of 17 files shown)


GerardPaligot:
@monperrus PR ok.

monperrus:
very interesting PR

now we need a clear contract for CtVisitor, because it's in the hierarchy.
Should it be
declares write* methods for and only for factory instantiable metamodel classes?
declares write* methods for and only for factory non abstract metamodel classes?
(should those two contracts be equivalent?)

we also need a clear contract for CtScanner and AbstractVisitor, but this is independent of this issue and PR.
...
unless, CtInheritanceScanner should extend AbstractVisitor and not directly implement CtVisitor. This may be the case since they both declare scan(CtElement) and the like.

what do you think?

GerardPaligot:
> now we need a clear contract for CtVisitor, because it's in the hierarchy.
> Should it be
> declares write* methods for and only for factory instantiable metamodel classes?
> declares write* methods for and only for factory non abstract metamodel classes?
> (should those two contracts be equivalent?)

I don't understand. What is `write*` methods? Do you mean `visit*`? Why do you talk about factory?

> unless, CtInheritanceScanner should extend AbstractVisitor and not directly implement CtVisitor. This may be the case since they both declare scan(CtElement) and the like.

Why not.

monperrus:
> What is |write*| methods? Do you mean |visit*|? 
yes

Do you agree with the contract of CtVisitor:
'declares visit* methods for and only for non abstract metamodel classes'

--Martin

GerardPaligot:
> Do you agree with the contract of CtVisitor: 'declares visit* methods for and only for non abstract metamodel classes'

Yes but this test is already done with `visit*` methods on `CtInheritanceScanner` class.

monperrus:
it seems that this test is done at the level of CtInheritanceScanner not
at the level of CtVisitor.

If it is done at the CtVisitor level, we can remove the check from
CtInheritanceScanner since the compiler enforces it.

GerardPaligot:
To resume, I remove all checks about `visit*` methods on `CtInheritanceScanner` and I create a new test case for `CtVisitor` with these checks. We are okay?

monperrus:
> To resume, I remove all checks about |visit*| methods on
> |CtInheritanceScanner| and I create a new test case for |CtVisitor|
> with these checks. We are okay?
>
remove all checks about about the presence of
but keep the checks about the call of those methods.

--Martin

GerardPaligot:
Do you talk about my test case or the test case of Nicolas with `CtInheritanceScannerTest`? Because, in my test case, I don't have checks about the call of `visit*` methods.

monperrus:
That's right, it's another test. So only I guess it's only splitting your test in two, one for CtVisitor and one for  CtInheritanceScanner

By the way, could you add the contract in CtVisitor? and a couple of method Javadoc is missing there.

GerardPaligot:
> and a couple of method Javadoc is missing there.

Where?

monperrus:
just open the file :-)

    <T> void visitCtAnnotationFieldAccess(CtAnnotationFieldAccess<T>
annotationFieldAccess);
    <T> void visitCtSuperAccess(CtSuperAccess<T> f);

GerardPaligot:
@monperrus PR ok.

src/test/java/spoon/reflect/visitor/processors/CheckModelProcessor.java
```diff
@@ -0,0 +1,99 @@
...
+
+		if (hasScanners) {
+			checkScan(visitor, theInterface, element.getModifiers().contains(ModifierKind.ABSTRACT));
+		}
+		if (hasVisitors) {
+			checkVisit(visitor, theInterface, element.getModifiers().contains(ModifierKind.ABSTRACT));
+		}
+	}
+
+	public CheckModelProcessor hasScanners() {
```
monperrus: method hasSomething returns a boolean by convention

src/test/java/spoon/reflect/visitor/processors/CheckModelProcessor.java
```diff
@@ -0,0 +1,99 @@
...
+import spoon.reflect.declaration.CtClass;
+import spoon.reflect.declaration.CtMethod;
+import spoon.reflect.declaration.CtType;
+import spoon.reflect.declaration.ModifierKind;
+import spoon.reflect.visitor.CtVisitor;
+
+import java.util.Collections;
+import java.util.List;
+
+public class CheckModelProcessor<T extends CtVisitor> extends AbstractProcessor<CtClass<?>> {
```
monperrus: javadoc?

src/test/java/spoon/reflect/visitor/processors/CheckModelProcessor.java
```diff
@@ -0,0 +1,99 @@
...
+		hasScanners = true;
+		return this;
+	}
+
+	public CheckModelProcessor hasVisitors() {
+		hasVisitors = true;
+		return this;
+	}
+
+	private void checkScan(CtType<CtVisitor> scanner, CtType<Object> theInterface, boolean isAbstract) {
```
monperrus: rename in checkPresenceofScanMethods? meaning of parameters?

src/test/java/spoon/reflect/visitor/CtVisitorTest.java
```diff
@@ -0,0 +1,46 @@
...
+import org.junit.Test;
+import spoon.Launcher;
+import spoon.reflect.visitor.processors.CheckModelProcessor;
+
+public class CtVisitorTest {
+	@Test
+	public void testMethodsInVisitor() throws Exception {
+		final Launcher launcher = new Launcher();
+		launcher.getEnvironment().setNoClasspath(true);
+		launcher.addProcessor(new CheckModelProcessor(CtVisitor.class, message -> {
```
monperrus: what about launcher.addProcessor(new CheckThat(CtVisitor.class).hasRequiredScanners()) simply 

(the default strategy could be directly in the processor

GerardPaligot:
@monperrus PR ok.

GerardPaligot:
@monperrus PR ok.

PR #452, opened 2015-12-09
Overriding/OverriddenMethodFilter

### LINKED ISSUE #274: add support for override analysis
We have a transformation that requires a method that gives the overridden method.

```
CtMethod {
  // returns the overridden method, null if none 
  CtMethod getOverridenMethod()

}
```

GerardPaligot:
Closes #274

### DIFF
--- a/src/main/java/spoon/reflect/visitor/filter/OverriddenMethodFilter.java
+++ b/src/main/java/spoon/reflect/visitor/filter/OverriddenMethodFilter.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright (C) 2006-2015 INRIA and contributors
+ * Spoon - http://spoon.gforge.inria.fr/
+ *
+ * This software is governed by the CeCILL-C License under French law and
+ * abiding by the rules of distribution of free software. You can use, modify
+ * and/or redistribute the software under the terms of the CeCILL-C license as
+ * circulated by CEA, CNRS and INRIA at http://www.cecill.info.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the CeCILL-C License for more details.
+ *
+ * The fact that you are presently reading this means that you have had
+ * knowledge of the CeCILL-C license and that you accept its terms.
+ */
+package spoon.reflect.visitor.filter;
+
+import spoon.reflect.declaration.CtExecutable;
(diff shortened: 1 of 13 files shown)


src/main/java/spoon/reflect/visitor/filter/OverridingMethodFilter.java
```diff
@@ -0,0 +1,47 @@
...
+import spoon.reflect.reference.CtExecutableReference;
+import spoon.reflect.visitor.Filter;
+
+/**
+ * Gets all overriding method from the method given.
+ */
+public class OverridingMethodFilter implements Filter<CtMethod<?>> {
+	private final CtExecutableReference<?> executableReference;
+
+	public OverridingMethodFilter(CtExecutableReference<?> executableReference) {
```
monperrus: Take as parameter a CtMethod (and not a reference)?

avoid the getReference in usage (cf test) with encapsulatio

src/main/java/spoon/reflect/visitor/filter/OverriddenMethodFilter.java
```diff
@@ -0,0 +1,47 @@
...
+import spoon.reflect.reference.CtExecutableReference;
+import spoon.reflect.visitor.Filter;
+
+/**
+ * Gets all overridden method from the method given.
+ */
+public class OverriddenMethodFilter implements Filter<CtMethod<?>> {
+	private final CtExecutableReference<?> executableReference;
+
+	public OverriddenMethodFilter(CtExecutableReference<?> executableReference) {
```
monperrus: idem

src/main/java/spoon/reflect/visitor/filter/OverridingMethodFilter.java
GerardPaligot: I adopted the same logic of `InvocationFilter` for this: https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/reflect/visitor/filter/InvocationFilter.java

monperrus: good idea to have a regular design. but I would rather update
CtInvocationFilter with a new constructor with a CtMethod.

GerardPaligot: We don't need a `CtMethod` in `InvocationFilter`. Why we should have a constructor useless?

monperrus:
> We don't need a CtMethod in InvocationFilter

We need it for regaularity in the API. And this will be useful for users in the future.

GerardPaligot:
@monperrus PR ok.

PR #453, opened 2015-12-10
adds new tests for TemplateMatcher

monperrus:
everything works now, incredible!

a side-effect on the recent work on references?


GerardPaligot:
> a side-effect on the recent work on references?

if only all side effects can be like this one. :)

PR #458, opened 2015-12-16
Use CtTypeAccess for CtFieldRead of .class and instanceof

GerardPaligot:


src/main/java/spoon/reflect/code/CtTypeAccess.java
```diff
@@ -25,4 +25,7 @@
  * 		type in the expression.
  */
 public interface CtTypeAccess<T> extends CtExpression<T> {
+	boolean isClassType();
```
monperrus: to be discussed. I'm not sure this have to explicit in the metamodel, it may be inferred from the context where it is used

GerardPaligot: I'm agree that should be discussed. Me, I prefer come back to `CtFieldRead`.

src/test/java/spoon/test/type/TypeTest.java
```diff
@@ -0,0 +1,115 @@
...
+						+ "  boolean x= (klass == short.class);" //
+						+ "}};").compile();
+		CtMethod<?> foo = (CtMethod<?>) clazz.getMethods().toArray()[0];
+
+		CtBlock<?> body = foo.getBody();
+		CtLocalVariable<?> ass = body.getStatement(1);
+		CtBinaryOperator<?> op = (CtBinaryOperator<?>) ass.getDefaultExpression();
+		assertEquals("Class", op.getLeftHandOperand().getType().getSimpleName());
+		assertFalse(op.getLeftHandOperand().getType().isPrimitive());
+		assertEquals("short", op.getRightHandOperand().getType().getSimpleName());
```
GerardPaligot: Should return `Class` and `op.getLeftHandOperand().getTarget().getSimpleName()` return `short`.

GerardPaligot:
@monperrus PR ok for me. If ok for you, you can merge.

PR #460, opened 2015-12-16
feat(field): Adds a field at a position given.

### LINKED ISSUE #455: CtClass#addField(CtField) need to add the CtField at the beginning of the class
The CtClass#addField(CtField), applies to a class, adds a field at the end of the class. The drawback is for the static fields : a static variable cannot use another that is already defined after. A solution is to add this field at the beginning of the class.

GerardPaligot:
Closes #455

### DIFF
--- a/src/main/java/spoon/reflect/declaration/CtType.java
+++ b/src/main/java/spoon/reflect/declaration/CtType.java
@@ -16,10 +16,10 @@
  */
 package spoon.reflect.declaration;
 
+import spoon.reflect.reference.CtTypeReference;
+
 import java.util.List;
 import java.util.Set;
-
-import spoon.reflect.reference.CtTypeReference;
 
 /**
  * This abstract element defines a super-type for classes and interfaces, which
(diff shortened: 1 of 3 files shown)


monperrus:
the use case is slightly different.

the missing method is addStaticField(Field f) (contract f is static)

such that the field is added before all static blocks such that it can be used in a static block

GerardPaligot:
No, the problem reported by @sofianekabylino is that the code below isn't valid:

```java
class Tacos {
  static int CONSTANT_2 = CONSTANT_1 + 1;
  static int CONSTANT_1 = 1;
}
```

But this code is valid:

```java
class Tacos {
  int variable = CONSTANT + 1;
  static int CONSTANT = 1;
}
```

sofianekabylino:
No, in my case it's like this example:
```java
public class GeneratedClass {
	private static int oldConstant = generatedConstant;
        /*
         ...
        */
	private static int generatedConstant = 42;
}
```
generatedConstant is the the injected code.

PR #462, opened 2015-12-16
clarifies the dumb contract of CtAbstractVisitor

monperrus:


src/main/java/spoon/reflect/visitor/CtAbstractVisitor.java
```diff
@@ -79,36 +79,10 @@
 import java.lang.annotation.Annotation;
 import java.util.Collection;
 
+/** Provides an empty implementation of CtVIsitor. 
+ *  See {@link CtScanner} for a much more powerful implementation.
```
GerardPaligot: Why do you talk about `CtScanner` here? It implements directly `CtVisitor`.

PR #464, opened 2015-12-20
[WIP] Spoon Template using byte code decompilation

tdurieux:
# WIP
It's a POC of Spoon template using byte code decompilation of the template instead of spooning the source code of the template.

But I have an issue with this type of expression: ```this.PARAM2.S()``` which is not interpreted. 
I did not found the class that handle these expressions, can you guide me?

Thanks

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -171,6 +171,10 @@
       <name>Maven Repository for JDT Compiler release</name>
       <url>https://repo.eclipse.org/content/repositories/eclipse-staging/</url>
     </repository>
+    <repository>
+      <id>clojars.org</id>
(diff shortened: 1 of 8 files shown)


monperrus:
looks really cool!

Since it introduces fancy dependencies, it might be more appropriate for [https://github.com/SpoonLabs/spoon-incubator](spoon-incubator) first. It seems it is easy to extend ExtensionTemplate. Could you create the PR there?

This may help us to identify missing extension points.

tdurieux:
I agree. I posted this PR in order to have some help to identify my issue. 
I don't completely understand the behavior of templates.

PR #472, opened 2016-01-05
feat(type): Partial support of type intersection.

### LINKED ISSUE #425: Problem with type intersections
When migrating our project to Java 8 we came across issues with Spoon's Java 8 support. `IntersectionCastTypeBinding` and `PolyTypeBinding` were not being used by `JDTTreeBuilder` and `createExecutableReferenceExpression` was causing problems with a specific method reference (see: https://github.com/SpongePowered/SpongeAPI/commit/55abbe5f7804570e22fcf0eaf40985fa1d49849c). I hacked around the problems in a fork here: https://github.com/SpongePowered/spoon/commit/0443dbeb5eaf582dd63a80acbe65a8b91c1ccf74. You can find some more details in the comments, but I want to put the emphasis on:

> One big issue is the new TypeIntersection which is a subtype of Type. Since it represents multiple types, most CtTypeReference methods, which expect a single type, can't properly be implemented by CtTypeIntersectionReference.

We also had to fix the source code generator. It isn't related to Java 8, but you can find the commit here: https://github.com/SpongePowered/spoon/commit/ae40d42b1ef451b878282dc70c0e68fd4ab052fb

GerardPaligot:
Fixes test case MainTest#testTest in IDE.

Closes #425

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilder.java
+++ b/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilder.java
@@ -121,6 +121,7 @@ import org.eclipse.jdt.internal.compiler.lookup.ClassScope;
 import org.eclipse.jdt.internal.compiler.lookup.CompilationUnitScope;
 import org.eclipse.jdt.internal.compiler.lookup.ElementValuePair;
 import org.eclipse.jdt.internal.compiler.lookup.FieldBinding;
+import org.eclipse.jdt.internal.compiler.lookup.IntersectionTypeBinding18;
 import org.eclipse.jdt.internal.compiler.lookup.LocalTypeBinding;
 import org.eclipse.jdt.internal.compiler.lookup.LocalVariableBinding;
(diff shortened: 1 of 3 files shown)


monperrus:
could you add a test case as well?

GerardPaligot:
@monperrus PR ok here for me.

monperrus:
Any idea why getTypeCasts() returns a list?

GerardPaligot:
Because `(String) ((Object) var))` is valid.

monperrus:
interesting. I would have encoded this as different elements, each one
with a single cast.

Could you add a test case to check that getTypeCasts() indeed returns
two elements for |"(String) ((Object) var))"|? What is the convention
for the order?

GerardPaligot:
For now, I'm on a different issue. If you want it, I'll write the test case when I'll develop the complete support of intersection types but [yes, I'm sure](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java#L256).

PR #477, opened 2016-01-08
fix(noclasspath): Gets package of types in the class declaration.

### LINKED ISSUE #465: Problem with qualified names of CtTypeReference when using setNoClasspath(true)
When settings setNoClasspath(true) on the environment, unresolved type references don't have the correct qualified name. `getQualifiedName()` returns the simple name (without the package prefix) instead.

I discovered that behaviour with this processor:

```Java

class TypeReferenceProcessor extends AbstractProcessor<CtType<?>> {

    @Override
    public void process(CtType<?> spType) {
        for (CtTypeReference<?> spRef : spType.getReferencedTypes()) {
            System.out.println(spRef.getQualifiedName());
        }
    }
}

```

GerardPaligot:
Closes #465

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilder.java
+++ b/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilder.java
@@ -732,6 +732,10 @@ public class JDTTreeBuilder extends ASTVisitor {
 					ref = factory.Core().createTypeReference();
 				}
 				ref.setSimpleName(new String(binding.readableName()));
+				final CtReference declaring = references.getDeclaringReferenceFromImports(binding.sourceName());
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -159,7 +159,7 @@
   </scm>
 
   <properties>
-    <java.src.version>1.6</java.src.version>
+    <java.src.version>1.8</java.src.version>
```
pschichtel: don't forget to revert this :)

GerardPaligot: Don't worry, Travis tell us this kind of error. :)

GerardPaligot:
@monperrus PR okay here for me.

PR #496, opened 2016-01-19
fix(anonymous): getDeclaration throws a NPE in anonymous class.

### LINKED ISSUE #494: NPE in getDeclaration type with anonymous classes
A NPE is threw in the getDeclaration() method when the variable is in an Anonymous class which is in an Anonymous class.
I creates a quick fix in this branch: https://github.com/tdurieux/spoon/tree/fix-get-type, fills free to clean it.

GerardPaligot:
Closes #494


GerardPaligot:
@monperrus PR ok.

PR #498, opened 2016-01-20
WIP fix(targeted): A class can be declared in a lambda expression.

GerardPaligot:


src/main/java/spoon/reflect/factory/TypeFactory.java
```diff
@@ -216,14 +214,7 @@ public boolean matches(CtNewClass element) {
...
 		if (pack == null) {
-			// The fully qualified name doesn't contain always the package. i.e. a class
-			// declared in a method. In this case, try to get the class from its name
-			// in the AST contained in the factory.
-			final List<CtType<T>> elements = Query.getElements(factory, new NameFilter<CtType<T>>(qualifiedName));
-			if (elements.size() == 0) {
-				return null;
-			}
-			return elements.get(0);
+			return null;
```
GerardPaligot: @monperrus This modification seems useful for some users (phd students at Rennes).

monperrus: ??

GerardPaligot: The goal of this PR is the suppression of this query to come back at the previous code

But you don't like the implementation. We should talk about it because I don't find another solution.

GerardPaligot:
Change the status of the PR to WP. Don't merge it.

PR #515, opened 2016-02-01
Added Metadata slots. Fixes #422.

### LINKED ISSUE #422: Feature Request: Extra Data Field
### Use case scenario:

I am performing static analysis of the AST. I want to save information related to each node, and then I want to perform operations on the AST, while maintaining the extra information of each node. (I am doing this for automatic parallelization, cost-model analysis and so on).

### How am I doing this and what the issue is:

Right now I am using an HashMap from each ASTNode to an Object I created. However, whenever I perform modifications on the AST, the replacement methods in Spoon change the AST nodes (specially parents of the node I am replacing), leading to new nodes that I have no information of in my hashmap. This was very troublesome to fix, and I believe I still have bugs on my code because of issues like this.

## How I propose to fix this:

Adding an Object field to CtElement. A Generic type would be useful, but would add a lot of extra code to everyone else's code. Being an Object, one could easily (and unsafely) cast it to our own data type.

I know it is possible to create another factory for my own Nodes. The problem is that I don't want to create 30-something classes just for adding a field. This and they won't be compatible with all the code that uses the existing factory or factory elements.

alcides:
Uses an Internal HashMap inside CtElement to store objects as metadata slots.

### DIFF
--- a/src/main/java/spoon/reflect/declaration/CtElement.java
+++ b/src/main/java/spoon/reflect/declaration/CtElement.java
@@ -233,4 +233,19 @@ public interface CtElement extends FactoryAccessor, CtVisitable {
 	 * Deletes the element. For instance, delete a statement from its containing block. Warning: it may result in an incorrect AST, use at your own risk.
 	 */
 	void delete();
+
+	/*
+	 * Saves metadata inside an Element.
+	 */
+	<E extends CtElement> E putMetadata(String key, Object val);
+
+	/*
+	 * Retrieves metadata stored in an element. Returns null if it does not exist.
+	 */
+	Object getMetadata(String key);
+
+	/*
+	 * Returns the metadata keys stored in an element.
+	 */
+	Set<String> getMetadataKeys();
 }
--- a/src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
+++ b/src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
@@ -44,7 +44,9 @@ import java.lang.reflect.Field;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 
 import static spoon.reflect.ModelElementContainerDefaultCapacities.ANNOTATIONS_CONTAINER_DEFAULT_CAPACITY;
@@ -149,6 +151,8 @@ public abstract class CtElementImpl implements CtElement, Serializable, Comparab
 	String docComment;
 
 	SourcePosition position;
+
+	Map<String, Object> metadata = new HashMap<String, Object>();
 
 	public CtElementImpl() {
 		super();
@@ -441,5 +445,21 @@ public abstract class CtElementImpl implements CtElement, Serializable, Comparab
 
 	private boolean compare(Object o1, Object o2) {
 		return o1 == o2;
+	}
+
+	@Override
+	public <E extends CtElement> E putMetadata(String key, Object val) {
+		metadata.put(key, val);
+		return (E) this;
+	}
+
+	@Override
+	public Object getMetadata(String key) {
+		return metadata.get(key);
+	}
+
+	@Override
+	public Set<String> getMetadataKeys() {
+		return metadata.keySet();
 	}
 }
--- a/src/test/java/spoon/test/ctElement/MetadataTest.java
+++ b/src/test/java/spoon/test/ctElement/MetadataTest.java
@@ -1,0 +1,38 @@
+package spoon.test.ctElement;
+
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertEquals;
+
+import org.junit.Test;
+
+import spoon.reflect.code.CtReturn;
+import spoon.reflect.declaration.CtClass;
+import spoon.reflect.declaration.CtMethod;
+import spoon.reflect.factory.Factory;
+import spoon.test.TestUtils;
+import spoon.test.ctElement.testclasses.Returner;
+
+public class MetadataTest {
+
+
+	@Test
+	public void testMetadata() throws Exception {
+
+		final Factory factory = TestUtils.build(Returner.class);
+		final CtClass<Returner> returnerClass = factory.Class().get(Returner.class);
+		final CtMethod<?> staticMethod = returnerClass.getMethodsByName("get").get(0);
+		final CtReturn<Integer> ret = staticMethod.getBody().getLastStatement();
+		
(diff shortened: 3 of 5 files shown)


src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
```diff
@@ -420,17 +425,17 @@ protected void replace(CtElement element) {
...
 				} else if (compare(tmp, toReplace)) {
-					f.set(parent, getReplacement(replacement, parent));
+					f.set(parent, getReplacement(toReplace, replacement, parent));
 				}
 			}
 		}
 	}
 
-	private <T extends FactoryAccessor> T getReplacement(T replacement, Object parent) {
+	private <T extends FactoryAccessor> T getReplacement(Object toReplace, T replacement, Object parent) {
```
GerardPaligot: This new parameter is unused.

src/test/java/spoon/test/ctElement/MetadataTest.java
```diff
@@ -0,0 +1,35 @@
...
+import spoon.reflect.declaration.CtMethod;
+import spoon.reflect.factory.Factory;
+import spoon.test.TestUtils;
+import spoon.test.ctElement.testclasses.Returner;
+
+public class MetadataTest {
+
+
+	@Test
+	public void testMetadata() throws Exception {
```
GerardPaligot: you should add an assert or a test case about the replace.

alcides:
I have undone the modifications of replace. After a second thought, it may be the wrong semantic. The replacement should come with its own metadata, instead of inheriting the metadata from the existing element, despite this being useful for my project.

So the replace test makes no sense now.

tdurieux:
I think inheriting the metadata from the existing element can be really useful but not for everyone.
I good solution is to parameterize this behavior.
What do you thing?

alcides:
I've thought about it. But I already have a wrapper around replace where I
copy the Metadata, but it might be useful to be parametrized, sure.

GerardPaligot:
Hi @alcides,

I don't know if your test case pass or not but first, you should correct checkstyle errors:

```console
[INFO] Starting audit...
/home/travis/build/INRIA/spoon/src/main/java/spoon/reflect/declaration/CtElement.java:231: error: Line has trailing spaces.
/home/travis/build/INRIA/spoon/src/main/java/spoon/reflect/declaration/CtElement.java:236: error: Line has trailing spaces.
/home/travis/build/INRIA/spoon/src/main/java/spoon/reflect/declaration/CtElement.java:241: error: Line has trailing spaces.
/home/travis/build/INRIA/spoon/src/main/java/spoon/support/reflect/declaration/CtElementImpl.java:155: error: Line has trailing spaces.
/home/travis/build/INRIA/spoon/src/main/java/spoon/support/reflect/declaration/CtElementImpl.java:449: error: Line has trailing spaces.
/home/travis/build/INRIA/spoon/src/main/java/spoon/support/reflect/declaration/CtElementImpl.java:454: error: Line has trailing spaces.
/home/travis/build/INRIA/spoon/src/main/java/spoon/support/reflect/declaration/CtElementImpl.java:459: error: Line has trailing spaces.
Audit done.
```

alcides:
Should be fixed now. The verify maven commands yields some errors related to missing headers, which I have not touched.

alcides:
Regarding the replace behavior, I believe allowing to attach custom hooks to replace would be a more general way of doing this. One may use the hook to copy metadata from replaced to replacement, but also for out-of-band updates on other databases (just like I am doing right now).

src/test/java/spoon/test/intercession/IntercessionTest.java
```diff
@@ -166,6 +163,8 @@ public boolean matches(CtMethod<?> element) {
 				});
 
 		for (CtMethod<?> setter : setters) {
+			if (setter.getSimpleName().toLowerCase().contains("metadata")) continue;
```
monperrus: why skip this?

use equals instead of contains?

alcides: I will change it to equals to setMetadata.

This test was failing because setMetadata return type is not parametrized. Which does not make much sense. I guess it is some test you might have written to have guarantees of a well-typed model api. But setMetadata is not a method from the AST API (as in setBody or setTarget).

GerardPaligot: @alcides Have you time for this or do you want a PR on your branch?

alcides: On Wed, Feb 10, 2016 at 10:06 AM, Gérard Paligot <notifications@github.com>
wrote:

> > +			if (setter.getSimpleName().toLowerCase().contains("metadata")) continue;
>
>
I'll accept the PR, but I don't understand why the test fails if I remove
that guard.

GerardPaligot: PR here: https://github.com/alcides/spoon/pull/1

GerardPaligot:
@alcides For information, you should squash your commits and rebase from upstream/master. Otherwise, Travis will fail.

GerardPaligot:
@monperrus PR ok here.
@alcides Thanks for the rebase.

src/test/java/spoon/test/ctElement/MetadataTest.java
```diff
@@ -0,0 +1,35 @@
...
+		final CtMethod<?> staticMethod = returnerClass.getMethodsByName("get").get(0);
+		final CtReturn<Integer> ret = staticMethod.getBody().getLastStatement();
+		
+		final CtMethod<?> staticMethod2 = returnerClass.getMethodsByName("get2").get(0);
+		final CtReturn<Integer> ret2 = staticMethod2.getBody().getLastStatement();
+		
+		ret.setMetadata("foo", "bar");
+		ret.setMetadata("fiz", 1);
+		
+		assertNotNull(ret.getMetadata("fiz"));
```
monperrus: can you add two assertEquals on getMetadata("foo") on getMetadata("fiz")

src/main/java/spoon/reflect/declaration/CtElement.java
```diff
@@ -233,4 +233,19 @@
 	 * Deletes the element. For instance, delete a statement from its containing block. Warning: it may result in an incorrect AST, use at your own risk.
 	 */
 	void delete();
+
+	/*
+	 * Saves metadata inside an Element.
+	 */
+	<E extends CtElement> E setMetadata(String key, Object val);
```
monperrus: rename as putMetadata

monperrus:
thanks for keeping this advancing. we still have todos before merging.

alcides:
Fixed the two comments @monperrus pointed out.

GerardPaligot:
@alcides Last PR on your fork for this PR: https://github.com/alcides/spoon/pull/2. After your merge, we merge this one.

PR #518, opened 2016-02-04
refactor(enum): Uses CtEnumValue for field values.

### LINKED ISSUE #474: Creates CtEnumValue and its access.
The values of enums are listed with the fields, this is completely counterintuitive because an enum value does not have: type, visibility and default expression.

The only solution to differentiate classic fields and enum values is to check that the type is null.

GerardPaligot:
Closes #474

### DIFF
--- a/src/main/java/spoon/reflect/declaration/CtEnumValue.java
+++ b/src/main/java/spoon/reflect/declaration/CtEnumValue.java
@@ -1,0 +1,34 @@
+/**
+ * Copyright (C) 2006-2015 INRIA and contributors
+ * Spoon - http://spoon.gforge.inria.fr/
+ *
+ * This software is governed by the CeCILL-C License under French law and
+ * abiding by the rules of distribution of free software. You can use, modify
+ * and/or redistribute the software under the terms of the CeCILL-C license as
+ * circulated by CEA, CNRS and INRIA at http://www.cecill.info.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the CeCILL-C License for more details.
+ *
+ * The fact that you are presently reading this means that you have had
+ * knowledge of the CeCILL-C license and that you accept its terms.
+ */
+package spoon.reflect.declaration;
+
+/**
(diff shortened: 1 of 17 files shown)


src/test/java/spoon/test/annotation/AnnotationTest.java
```diff
@@ -296,7 +297,7 @@ public void testAnnotatedElementTypes() throws Exception {
 		assertTrue(annotations.get(0).getAnnotatedElement().equals(enumeration));
 		assertEquals(CtAnnotatedElementType.TYPE, annotations.get(0).getAnnotatedElementType());
 
-		List<CtField<?>> fields = enumeration.getValues();
```
monperrus: if we are backward compatible only with deprecation, we should keep the previous getValues and assertions

GerardPaligot: It is weird you say that because, since the beginning of my contribution to this project, you want to remove all usage of deprecated methods in source and test code.

src/main/java/spoon/reflect/declaration/CtEnumValue.java
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the CeCILL-C License for more details.
+ *
+ * The fact that you are presently reading this means that you have had
+ * knowledge of the CeCILL-C license and that you accept its terms.
+ */
+package spoon.reflect.declaration;
+
```
GerardPaligot: documentation

GerardPaligot:
@monperrus PR ok for me here.

src/main/java/spoon/reflect/declaration/CtEnumValue.java
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Corresponds to one of enum value specified in an enumeration.
+ * The default expression of an enum value is always a CtConstructorCall
+ * where we have zero, one or more arguments.
```
GerardPaligot: Corresponds to one enum value specified in an enumeration.
If the enum value implicitly calls a constructor (see example below), it is stored in the default expression of the field as CtConstructorCall,
              
T: the type of the enum, hence equal to the type of getParent()

PR #519, opened 2016-02-04
chore(license): Updates configuration of license-maven-plugin.

### LINKED ISSUE #516: travis does not fail with headers issues
I've also encountered what @alcides describes on #515 about the missing headers.

it seems that Travis does not catch this.

GerardPaligot:
Should fixed #516

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -547,6 +547,9 @@
                 <exclude>src/test/**</exclude>
                 <exclude>src/changes/**</exclude>
                 <exclude>src/site/**</exclude>
+                <exclude>target/**</exclude>
+                <exclude>bin/**</exclude>
+                <exclude>build/**</exclude>
               </excludes>
             </configuration>
             <executions>


pom.xml
```diff
@@ -534,20 +534,9 @@
...
-                <exclude>doc/**</exclude>
-                <exclude>spooned/**</exclude>
-                <exclude>src/main/resources/**</exclude>
-                <exclude>src/main/assembly/**</exclude>
-                <exclude>src/test/**</exclude>
-                <exclude>src/changes/**</exclude>
-                <exclude>src/site/**</exclude>
-              </excludes>
+              <includes>
+                <include>src/main/java/**</include>
```
monperrus: we should exclude the packages.html and maybe some other resources

GerardPaligot: `package.html` files already have license headers (e.g., [package.html](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/package.html)). So, why not?

GerardPaligot:
@monperrus PR ok for me here.

PR #532, opened 2016-02-15
refactors the three responsibilities of equal/hashcode/signature

monperrus:
Before:
The three responsibilities of equals/hashCode/signature were handled by the same code in SignaturePrinter,
with two major problems:
- no deep equality for package, class and method
- low performance of hashCode

Now:
The three responsibilities are handled by three different visitors:
- SignaturePrinter for signatures
- EqualVisitor for equal
- HashcodeVisitor for equal

### DIFF
--- a/src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
+++ b/src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
@@ -16,7 +16,21 @@
  */
 package spoon.support.reflect.declaration;
 
+import static spoon.reflect.ModelElementContainerDefaultCapacities.ANNOTATIONS_CONTAINER_DEFAULT_CAPACITY;
+
+import java.io.Serializable;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Field;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
 import org.apache.log4j.Logger;
+
 import spoon.Launcher;
 import spoon.processing.FactoryAccessor;
 import spoon.reflect.cu.SourcePosition;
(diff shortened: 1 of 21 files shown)


GerardPaligot:
PR launched on our Jenkins. Results soon.

monperrus:
fingers crossed :-)

monperrus:
we also have to find a solution for the Log.isDebug in equal.

GerardPaligot:
Faster solution but worst: isRunningTests()
Lower solution but better: Specific logger for this check. By default, disabled but we can enable it for some tests like `MainTest`.

GerardPaligot:
Jenkins ok.

GerardPaligot:
Jenkins not ok with Google Guice: http://193.51.236.32:8080/job/Google%20Guice/76/

src/main/java/spoon/support/visitor/EqualVisitor.java
```diff
@@ -0,0 +1,700 @@
...
+			for (CtTypeReference<?> b : ref.getBounds()) {
+				scan(b);
+				write(", ");
+			}
+		}
+	}
+
+	@Override
+	public <T> void visitCtTypeReference(CtTypeReference<T> reference) {
+		write(reference.getSimpleName());
```
GerardPaligot: This should be `reference.getQualifiedName()` (fix Google Guice).

src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
```diff
@@ -158,23 +161,39 @@ public CtElementImpl() {
...
-		String other = o.getSignature();
+		String current = getDeepRepresentation(this);
+		String other = getDeepRepresentation(o);
 		if (current.length() <= 0 || other.length() <= 0) {
 			throw new ClassCastException("Unable to compare elements");
 		}
 		return current.compareTo(other);
 	}
 
+	public String getDeepRepresentation(CtElement elem) {
```
GerardPaligot: This doesn't work. e.g., when you'll build the deep representation of a method like this one:

```java
public void toProvider(Provider<? extends T> provider) {
}
```

The result of `getDeepRepresentation` will be: `void toProvider(javax.inject.Provider,)voidprovider{}`.

monperrus:
thanks for the debug. updated.

monperrus:
> Lower solution but better: Specific logger for this check. By default,
> disabled but we can enable it for some tests like |MainTest|.
I like this solution. But, it should not be for MainTest but for all tests.

Can we have two different log configurations, for testing and for deploying?

If yes, we have the solution!

GerardPaligot:
Ok with Jenkins.

Issue #550, opened 2016-03-07
NPE when initially processing sources on large GWT project


sjd78:
I'm using spoon to help do some static code transform against a large GWT project (ovirt-engine - git://gerrit.ovirt.org/ovirt-engine).  Before my code runs, the Launcher throws an NPE:

```
Exception in thread "main" java.lang.NullPointerException
	at spoon.support.compiler.jdt.JDTBasedSpoonCompiler.buildSources(JDTBasedSpoonCompiler.java:228)
	at spoon.support.compiler.jdt.JDTBasedSpoonCompiler.build(JDTBasedSpoonCompiler.java:481)
	at spoon.Launcher.buildModel(Launcher.java:682)
	at spoon.Launcher.run(Launcher.java:634)
```

I tracked down the issue, and the CompilationUnitDeclaration[] returned from TreeBuilderCompiler contains a large number of consecutive null elements at the end of the array.  If I change TreeBuilderCompiler.buildUnits() to return an array bound in length by "this.totalUnits" everything is well.

I plan on submitting a pull request against the stable branch with this simple fix.

GerardPaligot:
Can you give us an example to know how to reproduce your bug? I tried to apply Spoon due to the Spoon Maven Plugin. Even if I can't apply Spoon on all modules of your project because you have some private dependencies, everything seems to work fine.

Maybe, you can create a test case in your PR to highlight the bug?

sjd78:
I'll take another look.  I'm using the Launcher directly in code instead of the maven plugin, but I've encountered the issue both ways.  Hopefully I can get it to reliably break.

The tool I'm building is ding a one time transform of the codebase, so running it continulily with the maven plugin is a bit overkill.

I'll report back if I can get it breaking

sjd78:
I can only consistently recreate the compile NPE problem running spoon 5.0.2 under eclipse's m2e maven exec:exec launcher.  Standard spoon 5.0.2 works with maven exec:exec from the command line.  The only difference is the classpath.  The eclipse launcher resolves my projects to the eclipse workspace while the command line call resolves my projects to their installed location in my local .m2 repository.  With my patch applied, spoon works in both instances.

I'll post an example the best I can soon.

GerardPaligot:
Ok, thanks to the eclipse maven plugin... :)

I commented your PR.

GerardPaligot:
PR merged. Thanks.

PR #551, opened 2016-03-07
fix bug: NPE when compilation units array contains null elements

sjd78:
Fix a bug in TreeBuilderCompiler.buildUnits() where a retured array is
longer then expected and the extra spaces are null. The method is updated
to use the **totalUnits** instance variable as the true element count in
the **unitsToProcess** array instead of the length of the array itself.

This bug broke processing of classes using method JDTBatchCompiler.getUnits() and
expecting every element in the returned array to be non-null.  The problem came up
when trying to process the source for a large GWT project.

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/TreeBuilderCompiler.java
+++ b/src/main/java/spoon/support/compiler/jdt/TreeBuilderCompiler.java
@@ -17,6 +17,7 @@
 package spoon.support.compiler.jdt;
 
 import java.io.PrintWriter;
+import java.util.ArrayList;
 
 import org.eclipse.jdt.core.compiler.CompilationProgress;
 import org.eclipse.jdt.internal.compiler.ICompilerRequestor;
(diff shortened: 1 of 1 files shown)


sjd78:
This pull request addresses the problem described in issue #550.

src/main/java/spoon/support/compiler/jdt/TreeBuilderCompiler.java
```diff
@@ -72,6 +72,8 @@
 			requestor.acceptResult(unit.compilationResult);
 		}
 
-		return this.unitsToProcess;
+		CompilationUnitDeclaration[] unitsToReturn = new CompilationUnitDeclaration[this.totalUnits];
+		System.arraycopy(this.unitsToProcess, 0, unitsToReturn, 0, this.totalUnits);
```
GerardPaligot: That seems more smart to iterate on `unitsToProcess` to check which items are `null` and returns this list.

sjd78: I checked the content of the `unitsToProcess` array and it always has non-null elements in index 0 to `this.totalUnits`.  The remainder of the elements in the array are `null`.  I can change it to scan the array and  only copy non-null values, but it won't be as efficient as the `System.arraycopy`.

monperrus: > I can change it to scan the array and only copy non-null values
That would be better for future maintenance, thanks.

You can also document this contract in the Javadoc (returns an array of
non-null elements).

--Martin

sjd78:
I've updated this PR to check every element in the returning array and only return non-null elements.

GerardPaligot:
Thanks @sjd78.

Issue #557, opened 2016-03-15
How to visit a node before and after processing?


retuxx:
I need to visit a node *before* its children are processed and *after*. Currently, spoon only provides the strategies [PRE_ORDER](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/processing/TraversalStrategy.java#L33) and [POST_ORDER](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/processing/TraversalStrategy.java#L40) allowing me to either visit nodes before or after. Is there a method to combine these strategies?

GerardPaligot:
Hi @retuxx,

For now, it isn't possible but that could be interesting to integrate this kind of feature in Spoon. PR welcome or I'll work on it soon.

retuxx:
I think in order to add this feature the [Processor](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/processing/Processor.java) interface needs to be extended because the method [process](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/processing/Processor.java#L69) does not know whether it is called pre or post order. Do you have any suggestions what to do? I could try to add this feature and making a pr if you are busy at the moment.

GerardPaligot:
I think that the better way is to add a new state in [TraversalStrategy](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/processing/TraversalStrategy.java#L24) and update the logical in [ProcessingVisitor](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/visitor/ProcessingVisitor.java#L81-L94). That seems pretty easy. :)

retuxx:
Yes, but how should I know whether [process](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/processing/Processor.java#L69) (which gets called by [scan](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/visitor/ProcessingVisitor.java#L81-L94)) is called in pre- or in post-order step?

GerardPaligot:
Ok, I see your issue.

In the [Processor](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/processing/Processor.java#L30) interface, you can add 2 methods named `preProcess(CtElement)` and `postProcess(CtElement)` with an empty default implementation in `AbstractProcessor`.

[...]

PR #559, opened 2016-03-15
Pre post order

retuxx:
This pr is just a suggestion to handle the feature requested of #557. The general idea is to extend the Processor and CtVisitor class in order to provide a state that can be used to determine if the process/visit method has been called from pre- or from post-order step. The state attribute ensures backward compatibility as the usage of it is optional. Unfortunately, I had to create some code duplications (the setter of the state). Maybe you should enhance the class hierarchy in order to prevent such duplications. Nonetheless, this should not (and actually is not) part of this pr. I'm unsure if calling process twice leads to errors. If yes, we could just remove it as the visit methods are the one I want to get executed twice, before and after. However, running the tests results to an error in spoon.test.intercession.IntercessionTest.testSettersAreAllGood line 202. Maybe you know whats going on here. All other tests pass.

### DIFF
--- a/src/main/java/spoon/processing/AbstractManualProcessor.java
+++ b/src/main/java/spoon/processing/AbstractManualProcessor.java
@@ -31,6 +31,8 @@ public abstract class AbstractManualProcessor implements Processor<CtElement> {
 
 	Factory factory;
 
+	private TraversalStrategy traversalState = TraversalStrategy.POST_ORDER;
+
 	/**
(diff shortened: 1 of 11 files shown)


retuxx:
Oh, I didn't see that you are using checkstyle. I will fix the errors if you think that this pr is worthwhile to be merged.

retuxx:
This pr contains an error on design level. I will close it for the moment until I fixed it.

PR #560, opened 2016-03-16
Added an example using the feature requested in #557

retuxx:


GerardPaligot:
I don't see any reason to refuse this PR. Thanks @retuxx for this example.

Issue #575, opened 2016-04-06
add and remove a method to a class and compile it


mehdijk:
Hi
I have a ctClass object and add a ctMethod to its compilation. It works well but when I remove this method from class again and try to compile, it keeps the ctMethod. Actually it shows that it has been removed successfully but when I compile, it is already there and makes the same compilation error.
I have another question. How can I get the compilation output message when I call compile() from ModelBuilder? It only returns a boolean value but I need detail of the compilation output which is shown in console. 
This is a my code:
`ctClazz.addMethod(ctMethod);

launcher.getModelBuilder().compile();
ctClazz.removeMethod(ctMethod);
launcher.getModelBuilder().compile();`

tdurieux:
Hi,

I can reproduce you problem, it appears when you try to compile more than one time a class (cache).

Unfortunately there is no good workaround (you can use reflection see below).

You can get the compilation error via:
```Java
SpoonModelBuilder modelBuilder = launcher.getModelBuilder();
if (modelBuilder instanceof JDTBasedSpoonCompiler) {
	((JDTBasedSpoonCompiler) modelBuilder).getProblems();
}
```
But there is no way to get the warning messages (the method that handles the output is ```JDTBatchCompiler.getBatchRequestor()```)


## Clear compilation cache manually
```Java
Field loadedContent = modelBuilder.getClass().getDeclaredField("loadedContent");
loadedContent.setAccessible(true);
Object loadedContentObj = loadedContent.get(modelBuilder);
Method clear = loadedContentObj.getClass().getDeclaredMethod("clear");
clear.invoke(loadedContentObj);
```

mehdijk:
Hi
Thank you for response.
I clear compilation cache manually according to that code. The compilation message is expected but at the end of the message, an  spoon.compiler.ModelBuildingException occurs again. Actually this situation is happened when I do add and remove like this:
`ctClazz.addMethod(ctMethod1);
ctClazz.removeMethod(ctMethod2);
launcher.getModelBuilder().compile();
//clear the cache
ctClazz.addMethod(ctMethod2);
ctClazz.removeMethod(ctMethod1);
launcher.getModelBuilder().compile();`

tdurieux:
What is the message of ModelBuildingException?

mehdijk:
After first compile this is the part or output:
....
**3 problems (1 error, 2 warnings)**
[...]

PR #577, opened 2016-04-07
fix(cache): Invalidate cache of the compiler.

GerardPaligot:
Issue #575

### DIFF
--- a/src/test/java/spoon/test/api/APITest.java
+++ b/src/test/java/spoon/test/api/APITest.java
@@ -7,6 +7,7 @@ import spoon.Launcher;
 import spoon.SpoonAPI;
 import spoon.compiler.Environment;
 import spoon.reflect.declaration.CtClass;
+import spoon.reflect.declaration.CtMethod;
 import spoon.reflect.declaration.CtPackage;
 import spoon.reflect.declaration.CtType;
 import spoon.reflect.factory.Factory;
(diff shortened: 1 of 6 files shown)


src/test/java/spoon/test/api/APITest.java
```diff
@@ -233,4 +235,30 @@ public void testPrintNotAllSourcesInCommandLine() throws Exception {
...
+		assertTrue(spoon.getModelBuilder().compile());
+
+		final CtClass<Bar> aClass = spoon.getFactory().Class().get(Bar.class);
+		aClass.addMethod(aMethod);
+		spoon.getModelBuilder().setShouldInvalidateCache(true);
+		assertFalse(spoon.getModelBuilder().compile());
+
+		aClass.removeMethod(aMethod);
+
+		spoon.getModelBuilder().setShouldInvalidateCache(true);
```
monperrus: why is it required to call twice `setShouldInvalidateCache(true)`?

monperrus:
what if we don't have an option and we always invalidate the cache?

src/test/java/spoon/test/api/APITest.java
GerardPaligot: A compiler is created at each build.

GerardPaligot:
> what if we don't have an option and we always invalidate the cache?

Why have we a cache?

tdurieux:
> Why have we a cache?
Because JDT asks 3 times for the content of the file during one compilation.

PR #582, opened 2016-04-14
Refactoring in CtElementImpl.replaceIn()

leventov:


GerardPaligot:
Can you add a test case where this refactoring is necessary?

Btw, I'm working on a new implementation for this feature. You can see the working in progress on this branch: https://github.com/GerardPaligot/spoon/tree/feat_replace

leventov:
I don't think it is necessary anywhere. If you are preparing a shiny new replacement implementation, I think applying this patch is pointless.

PR #590, opened 2016-04-18
fix(doc): Handles comments in CtNewArray and in CtConditional

### LINKED ISSUE #588: Parent isn't initialized for a specific comment
When you have this kind of comment, Spoon doesn't initialize its parent:

```java
return sample ? (n == 1) ? null    // NULL (not NaN) is correct in this case
                                 : new Double(vk / (double) (n - 1))
                      : new Double(vk / (double) (n));
```

tdurieux:
Fix #588


GerardPaligot:
Thanks @tdurieux for this fix.

PR #591, opened 2016-04-18
Optimize CtElement.equals()

leventov:
In my application this single line change reduced run time by 45%.


GerardPaligot:
Thanks @leventov!

PR #596, opened 2016-04-19
Fix bugs in ImportScannerImpl

leventov:


GerardPaligot:
Can you add a test case?

leventov:
Ok, will do.

.. While doing this I found another bug, in the meantime, could you please explain why you added `scan(reference.getType())` here: https://github.com/INRIA/spoon/commit/7fbc5fd3622330830a8fe3572fb65d1fd0f37859#diff-d6d96ce9408866e5f08058c0be741b93R60

GerardPaligot:
`CtExecutableReference` is saved in `CtInvocation` (for example) and we delegate the type of this last node at its executable reference.

leventov:
@GerardPaligot I don't understand your explaination, please check this PR again, I think the test case shows why this call shouldn't be there.

src/main/java/spoon/reflect/visitor/ImportScannerImpl.java
```diff
@@ -72,19 +72,17 @@
...
 
 	@Override
 	public <T> void visitCtExecutableReference(
 			CtExecutableReference<T> reference) {
 		enter(reference);
-		if (reference.getDeclaringType() != null
-				&& reference.getDeclaringType().getDeclaringType() == null) {
-			addImport(reference.getDeclaringType());
+		// constructor call
+		if (reference.getSimpleName().equals("<init>")) {
```
GerardPaligot: Why?

src/main/java/spoon/reflect/visitor/ImportScannerImpl.java
```diff
@@ -72,19 +72,17 @@
...
+			scan(reference.getDeclaringType());
 		}
-		scan(reference.getType());
```
GerardPaligot: If you remove this, we can't compute the import needed for a `ConstructorCall` (for example).

src/main/java/spoon/reflect/visitor/ImportScannerImpl.java
```diff
@@ -95,6 +93,7 @@
 		scan(invocation.getAnnotations());
 		scan(invocation.getTypeCasts());
 		scan(invocation.getTarget());
+		scan(invocation.getExecutable());
```
GerardPaligot: And if you add this, you'll add an unnecessary import with https://github.com/INRIA/spoon/pull/596/files#diff-d6d96ce9408866e5f08058c0be741b93L87

src/main/java/spoon/reflect/visitor/ImportScannerImpl.java
leventov: If you add this back, it will make the added test to fail (it will import `internal3.Bar` however it shouldn't.)

src/main/java/spoon/reflect/visitor/ImportScannerImpl.java
leventov: This is needed to import actual type arguments, as in the added test -- without this line `java.io.File` won't be imported.

src/main/java/spoon/reflect/visitor/ImportScannerImpl.java
GerardPaligot: What happens when I want compute this kind of invocation:

```java
Type.staticMethod()
```

src/main/java/spoon/reflect/visitor/ImportScannerImpl.java
GerardPaligot: Ok, so we need to scan the type of the invocation here.

GerardPaligot: Good point.

leventov:
Ok  I finally made this PR to pass all tests, please review again

PR #601, opened 2016-04-19
Add missing scanning of CtExecutableReference's parameters in CtScanner

leventov:
Sorry I'm lazy to add a test for this because this is obviously a missing AST scanning branch, the test is going to be trivial from the source: "yes, we actually didn't visit execRef's parameters".

### DIFF
--- a/src/main/java/spoon/reflect/visitor/CtScanner.java
+++ b/src/main/java/spoon/reflect/visitor/CtScanner.java
@@ -379,6 +379,7 @@ public abstract class CtScanner implements CtVisitor {
 		enter(reference);
 		scan(reference.getDeclaringType());
 		scan(reference.getType());
+		scan(reference.getParameters());
 		scan(reference.getActualTypeArguments());
 		scan(reference.getAnnotations());
 		scan(reference.getComments());


monperrus:
It's indeed obvious. However, a test is not for now, it's for the future. If somebody in one year makes a change (in this class or elsewhere) that breaks this expected behavior, s/he will notice it.

so I kindly ask you to add a test :-)

leventov:
Down the rabbit hole...

To write a test I decided to improve hashCodeVisitor, because currently it is way too simple, emitting just `1` for a lot of elements (all which don't have names). Here is my version:
```java
/**
 * Copyright (C) 2006-2015 INRIA and contributors
 * Spoon - http://spoon.gforge.inria.fr/
 *
 * This software is governed by the CeCILL-C License under French law and
 * abiding by the rules of distribution of free software. You can use, modify
 * and/or redistribute the software under the terms of the CeCILL-C license as
 * circulated by CEA, CNRS and INRIA at http://www.cecill.info.
 *
 * This program is distributed in the hope that it will be useful, but WITHOUT
 * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
 * FITNESS FOR A PARTICULAR PURPOSE. See the CeCILL-C License for more details.
 *
 * The fact that you are presently reading this means that you have had
 * knowledge of the CeCILL-C license and that you accept its terms.
 */
package spoon.support.visitor;

import spoon.reflect.declaration.CtNamedElement;
import spoon.reflect.visitor.CtScanner;

import java.util.Collection;
import java.util.Iterator;
import java.util.List;
import java.util.Set;

/** Responsible for computing CtElement.hashCode().
 * Version that is fast and compatible with EqualVisitor
 */
public class HashcodeVisitor extends CtScanner {

	/**
	 * This prime is used in google/auto
	 */
	private static final int HASH_CODE_X = 1000003;

	private boolean goDown = true;
	private int hashCode = 1;

	@Override
	protected void enter(CtElement e) {
		goDown = false;
	}

	@Override
	protected void exit(CtElement e) {
		goDown = true;
	}

	@Override
		hashCode *= HASH_CODE_X;
		if (element instanceof CtNamedElement) {
			hashCode += ((CtNamedElement) element).getSimpleName().hashCode();
		} else {
			hashCode += 1;
		}
		if (goDown) {
			super.scan(element);
		}
	}

	@Override
	public void scan(Object o) {
		if (o instanceof CtElement) {
			scan((CtElement) o);
		}
		if (o instanceof Collection<?>) {
			Iterator<?> it = ((Collection) o).iterator();
			if (it.hasNext() && it.next() instanceof CtElement) {
				scan((Collection<? extends CtElement>) o);
			}
		}
	}

	@Override
	public void scan(Collection<? extends CtElement> elements) {
		hashCode *= HASH_CODE_X;
		int result = 1;
		if (elements instanceof List) {
			for (CtElement element : elements) {
				result *= HASH_CODE_X;
				if (element instanceof CtNamedElement) {
					result += ((CtNamedElement) element).getSimpleName().hashCode();
				} else {
					result += 1;
				}
			}
		} else if (elements instanceof Set) {
			for (CtElement element : elements) {
				if (element instanceof CtNamedElement) {
					result ^= ((CtNamedElement) element).getSimpleName().hashCode();
				} else {
					result ^= 1;
				}
			}
		}
		// if elements is not List and not Set, result variable stays 1 because we cannot
		// compute hash code of unknown type of collection
		hashCode += result;
	}

	public int getHasCode() {
		return hashCode;
	}
}

```

But when I tried to test it, it turned out that equals is very often considers elements equal while they actually are not...

PR #612, opened 2016-04-22
Fix a bug in CodeFactory.createThisAccess() (target was not set)

leventov:


GerardPaligot:
PR done on your repo.

PR #616, opened 2016-04-22
Fix a bug in CtElementImpl.replace(), shouldn't re-add null into a co…

leventov:
…llection


monperrus:
we can merge this, but this will be removed with the new much faster implementation of replace that is in preparation. without a test, the bug may re-appear at that time. could you add a test?

leventov:
Ok will do.

monperrus:
closing this one, we now have a brand-new implementation of replace. don't hesitate to re-open if the bug is still there.

PR #635, opened 2016-04-27
Throws a SpoonException when we set ? in CtTypeRef.

### LINKED ISSUE #631: Ternary Operator in Constructors Fails.
The following example does not work with spoon 5.1.0 (The code itself is not useful, it's for demonstration purpose only):

```java
import java.util.Date;

public class TernaryOperator {

    private Date currentDate;

    public TernaryOperator(Date date) {
        currentDate = new Date(date == null ? null : date);
    }
}
```

If I build the model with `Launcher#buildModel()` the following exception is thrown:

```java
java.lang.RuntimeException: arg
```

I found that spoon does not correctly handle ternary operators used in constructors. The exception is thrown in `CtReferenceImpl#setSimpleName(String simplename)`.

GerardPaligot:
Closes #631

### DIFF
--- a/src/main/java/spoon/support/reflect/reference/CtReferenceImpl.java
+++ b/src/main/java/spoon/support/reflect/reference/CtReferenceImpl.java
@@ -16,6 +16,7 @@
  */
 package spoon.support.reflect.reference;
 
+import spoon.SpoonException;
 import spoon.reflect.factory.Factory;
 import spoon.reflect.factory.FactoryImpl;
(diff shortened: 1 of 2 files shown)


tdurieux:
I think it is not a good idea to specify this behavior because this can break some build model in noclasspath.

For example:
You have a project that has a method call of a method created in version 2 of a library
You spoon this project in noclasspath with version 1 of the library 
At the creation of the method reference, this exception will be threw.

PR #638, opened 2016-04-28
fix(getparent): getParent(Filter) does not return the current element when it matches the filter

### LINKED ISSUE #637: getParent(Filter) return this when the element match the filter

tdurieux:
close #637


src/test/java/spoon/test/parent/ParentTest.java
```diff
@@ -42,6 +33,14 @@
 import spoon.reflect.visitor.filter.TypeFilter;
 import spoon.test.replace.testclasses.Tacos;
 
+import java.util.Stack;
```
GerardPaligot: Seems unused.

PR #645, opened 2016-05-04
Fix many small bugs with double usage of AST nodes in different subtrees, ...

leventov:
…,

a couple of bugs in CtScanner and add checker which always checks AST correctness after model build.

### DIFF
--- a/src/main/java/spoon/reflect/factory/CodeFactory.java
+++ b/src/main/java/spoon/reflect/factory/CodeFactory.java
@@ -139,7 +139,7 @@ public class CodeFactory extends SubFactory {
 		fieldReference.setDeclaringType(type);
 
 		CtFieldRead<Class<T>> fieldRead = factory.Core().createFieldRead();
-		fieldRead.setType(classType);
+		fieldRead.setType(factory.Core().clone(classType));
 		fieldRead.setVariable(fieldReference);
 		fieldRead.setTarget(typeAccess);
 		return fieldRead;
@@ -159,7 +159,7 @@ public class CodeFactory extends SubFactory {
 		CtExecutableReference<T> executableReference = factory.Core()
 				.createExecutableReference();
 		executableReference.setType(type);
-		executableReference.setDeclaringType(type);
+		executableReference.setDeclaringType(factory.Core().clone(type));
 		executableReference.setSimpleName(CtExecutableReference.CONSTRUCTOR_NAME);
 		List<CtTypeReference<?>> typeReferences = new ArrayList<CtTypeReference<?>>();
 		for (int i = 0; i < parameters.length; i++) {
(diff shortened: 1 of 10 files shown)


leventov:
BTW I had really hard time figuring out that after altering CtScanner I also should change ` b/src/main/java/spoon/support/visitor/replace/ReplacementVisitor.java` in order to make test passing..

src/main/java/spoon/reflect/visitor/CtScanner.java
```diff
@@ -573,10 +574,10 @@ public void visitCtIf(final CtIf ifElement) {
 	public <T> void visitCtNewClass(final CtNewClass<T> newClass) {
 		enter(newClass);
 		scan(newClass.getAnnotations());
-		scan(newClass.getType());
```
monperrus: why this?

monperrus:
interesting PR

src/main/java/spoon/reflect/visitor/CtScanner.java
leventov: @monperrus because ctNewClass.getType() is not a "true" property of a ctNewClass, it delegates to getExecutable().getType(). In this case scanner shouldn't visit same node twice. Also see that CtNewClass extends CtConstructorCall, which have the same logic for getType(), but doesn't have scan(getType()) in it's visit method in CtScanner.

monperrus:
@GerardPaligot what do you think?

src/main/java/spoon/reflect/factory/CodeFactory.java
```diff
@@ -139,7 +139,7 @@ public CodeFactory(Factory factory) {
 		fieldReference.setDeclaringType(type);
 
 		CtFieldRead<Class<T>> fieldRead = factory.Core().createFieldRead();
-		fieldRead.setType(classType);
+		fieldRead.setType(factory.Core().clone(classType));
```
GerardPaligot: Do you have an use case where the type of the field read isn't equals of the field reference?

src/main/java/spoon/support/visitor/replace/ReplacementVisitor.java
```diff
@@ -1486,16 +1486,17 @@ public void visitCtIf(final spoon.reflect.code.CtIf ifElement) {
 		replaceInListIfExist(ctConstructorCall.getTypeCasts(), new spoon.support.visitor.replace.ReplacementVisitor.CtExpressionTypeCastsReplaceListener(ctConstructorCall));
 		replaceElementIfExist(ctConstructorCall.getExecutable(), new spoon.support.visitor.replace.ReplacementVisitor.CtAbstractInvocationExecutableReplaceListener(ctConstructorCall));
 		replaceElementIfExist(ctConstructorCall.getTarget(), new spoon.support.visitor.replace.ReplacementVisitor.CtTargetedExpressionTargetReplaceListener(ctConstructorCall));
+		replaceInListIfExist(ctConstructorCall.getActualTypeArguments(), new spoon.support.visitor.replace.ReplacementVisitor.CtGenericElementReferenceActualTypeArgumentsReplaceListener(ctConstructorCall));
```
GerardPaligot: `ReplacementVisitor` is generated by a processor in `spoon.generating`. A test case should notify you about that. I'll check that.

src/main/java/spoon/reflect/factory/CodeFactory.java
leventov: @GerardPaligot it's not the matter of equality. Any _object_ cannot appear in ast more than once. Otherwise very subtle bugs with element.replace() appear.

GerardPaligot:
PR on your PR today.

GerardPaligot:
My mistake, PR not necessary. I merge it.

PR #646, opened 2016-05-04
Replace CtClass.constructors model from Set to List

leventov:
`// TODO: CHANGE SETS TO LIST TO AVOID HAVING TO DO THIS` says for itself, and I also encountered this issue in my use cases.

Generally, it seems hardly ever sensible to have any AST children node collections as Sets. Because finally it's AST user responsibility to generate correct code.  I think, framework shouldn't be "too smart" trying to do something for us in terms of Java source code semantics, i. e. making constructors/methods unique. It will be a java compiler error, finally, if we have two identical methods/constructors in a class body. On the other hand, pattern clone method -> insert into method list -> update list of arguments, body.. is pretty reasonable, but it is ruled out by the current Set<CtMethod> (and constructors, regaring this PR).

Though all collections should be Lists (IMO), even for lists we should probably check that some element (java object) is not inserted twice into any list. Because it is definitely a programmic error (like addressed by #645 ) and could lead to hard to find bugs, if replace/clone/traverse takes place.

### DIFF
--- a/src/main/java/spoon/reflect/ModelElementContainerDefaultCapacities.java
+++ b/src/main/java/spoon/reflect/ModelElementContainerDefaultCapacities.java
@@ -37,6 +37,9 @@ public final class ModelElementContainerDefaultCapacities {
      * than ArrayList's default of 10.
      */
 
+	// JDK 7 average is 1.467
+	public static final int CONSTRUCTORS_DEFAULT_CAPACITY = 2;
+
 	// JDK 7 average is 1.063 (methods), 1.207 (constructors)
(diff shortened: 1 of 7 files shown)


monperrus:
I'm not against such change, although we would make it carefully, since it breaks client code.

Below a generic discussion on what frameworks should do, what is the concrete problem with having a Set in constructors?

monperrus:
Closing this one. In Spoon, we take a great care of balancing improvement and backward compatibility.  Don't hesitate to reopen it if necessary.

Issue #636, opened 2016-04-28
Performance improvement hascode and compareTo


tdurieux:
If you store CtClass in a set, map, ... (for example 256 ctClass) the performance are really bad.

This is due to:
* [x] to much hash code collision 
* [x] bad performance of the [EqualVisitor](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/visitor/EqualVisitor.java) (string comparison)  [#674]
* [ ] comparaTo in CtElement

leventov:
I resolved this issue locally (written a better EqualsVisitor and HashCodeVisitor, the latter even already showed up -- https://github.com/INRIA/spoon/pull/601#issuecomment-212035743)

But I'm severely overhelmed... I hope I will find time to push changes over a week.

tdurieux:
@GerardPaligot has almost finished a new EqualsVisitor (without reflection) that resolves the equals issue.

leventov:
@tdurieux  In this case I will just post my version -- @GerardPaligot might incorporate something into this version.

leventov:
`BiScanner.java`

```java
/**
 * Copyright (C) 2006-2015 INRIA and contributors
 * Spoon - http://spoon.gforge.inria.fr/
 *
 * This software is governed by the CeCILL-C License under French law and
 * abiding by the rules of distribution of free software. You can use, modify
 * and/or redistribute the software under the terms of the CeCILL-C license as
 * circulated by CEA, CNRS and INRIA at http://www.cecill.info.
 *
 * This program is distributed in the hope that it will be useful, but WITHOUT
 * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
 * FITNESS FOR A PARTICULAR PURPOSE. See the CeCILL-C License for more details.
 *
 * The fact that you are presently reading this means that you have had
 * knowledge of the CeCILL-C license and that you accept its terms.
 */
package spoon.reflect.visitor;

import spoon.reflect.declaration.CtAnnotation;
import spoon.reflect.declaration.CtElement;
import spoon.reflect.reference.CtTypeParameterReference;
import spoon.reflect.reference.CtTypeReference;

import java.lang.annotation.Annotation;
import java.util.ArrayList;
import java.util.Collection;
import java.util.Collections;
import java.util.Iterator;
import java.util.List;

import static spoon.reflect.visitor.BiScanner.ScanResult.CONTINUE;
[...]

PR #647, opened 2016-05-04
Improvements to collections: Make CtElementImpl.emptyList() and emptySet() DIFFERENT objects

leventov:
... because otherwise they cannot conform both java.util.List and java.util.Set equals()/hashCode() contract; never use Collections.unmodifiableCollection() (if wrapping list or set) because it narrows possible use on the client side without actual need; try to avoid creating unmodifiableList/Set(children) wrapper when the wrapped children is already an empty collection


leventov:
I discovered this CtElementImpl.emptySet() == CtElementImpl.emptyList() issue while testing this class: https://github.com/INRIA/spoon/issues/636#issuecomment-216770436

PR #654, opened 2016-05-09
refactor(api): Uses actual types of executable ref in constructor call.

### LINKED ISSUE #599: ActualTypeArguments are duplicated in CtConstructorCall and CtExecutableReference
In the same situation `CtExecutableReferenceExpression` doesn't have "own" actualTypeArguments in addition to type args incapsulated in it's `CtExecutableReference`.

GerardPaligot:
An interesting enhancement can be to add `CtGenericElementReference` on `CtInvocation` to write something like this : `ctInvocation.getActualTypeArguments()`. For now, you must  pass by the `CtExecutableReference`: `ctInvocation.getExecutable().getActualTypeArguments()`. It isn't really easy to use.

What do you think @monperrus?

Closes #599


monperrus:
Good idea, but we should rename `CtGenericElementReference` into `CtActualTypeContainer`

PR #655, opened 2016-05-09
doc(executable): Specify executable returns by getAllExecutables.

### LINKED ISSUE #621: TypeInformation.getAllExecutables() is underspecified and unusable
Specification says
```java
	/**
	 * Gets the executables declared by this type and by all its supertypes if
	 * applicable.
	 */
```

It doesn't say about destinction between static and instance executables, constructors, and what happens when overriding occurs.

`CtTypeImpl.getAllExecutables()` just puts everything into a hashSet, (relying on executable.equals(other), that is *not* the same as executable.overrides(other)), also it doesn't take into account default methods defined in super interfaces.

I would deprecate this method (or remove, depends on your compatibility rules).

GerardPaligot:
Closes #621

### DIFF
--- a/src/main/java/spoon/reflect/declaration/CtTypeInformation.java
+++ b/src/main/java/spoon/reflect/declaration/CtTypeInformation.java
@@ -122,7 +122,16 @@ public interface CtTypeInformation {
 
 	/**
 	 * Gets the executables declared by this type and by all its supertypes if
-	 * applicable.
+	 * applicable. This method returns:
+	 *
+	 * <ul>
+	 *     <li>static, instance and default executables</li>
+	 *     <li>Overridden methods</li>
+	 *     <li>constructors</li>
+	 * </ul>
+	 *
+	 * If a method is overridden twice in the hierarchy, it counts for two different elements.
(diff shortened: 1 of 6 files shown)


src/main/java/spoon/reflect/declaration/CtTypeInformation.java
```diff
@@ -122,7 +122,12 @@
 
 	/**
 	 * Gets the executables declared by this type and by all its supertypes if
-	 * applicable.
+	 * applicable. This method returns:
+	 * <ul>
+	 *     <li>static, instance and default executables</li>
+	 *     <li>Overridden methods</li>
```
monperrus: if it's overridden twice in the hierarchy, does it count for two different elements?

GerardPaligot: Yes, see the test case: https://github.com/INRIA/spoon/pull/655/files#diff-1d756de7906a25260ce271613f43b677R114

monperrus: could you update the doc accordingly?

thanks.

GerardPaligot: "Overridden methods" isn't enough?

PR #656, opened 2016-05-09
Rename CtGenericElement(Reference) and add CtActualTypeContainer on CtInvocation

GerardPaligot:


src/main/java/spoon/reflect/reference/CtActualTypeContainer.java
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * The fact that you are presently reading this means that you have had
+ * knowledge of the CeCILL-C license and that you accept its terms.
+ */
+package spoon.reflect.reference;
+
+import java.util.List;
+
+/**
+ * This interface defines a reference to {@link spoon.reflect.declaration.CtFormalTypeDeclarer}.
```
monperrus: This interface defines the capability related to binding generics (aka type parameters).

monperrus:
Not a single change in tests. I'm fine with this refactoring with much better names. OK for merging?

GerardPaligot:
Yes

PR #659, opened 2016-05-10
refactor(api): Changes bounds from List to Set.

### LINKED ISSUE #624: CtTypeParameterReference.bounds should be Set, not List
In Java language bounds play much more like a Set rather than a List:

```java
class Foo {
	<T extends Iterable & Comparable> T foo() { return null; }
}

class Bar extends Foo {
	// permutations allowed
	@Override
	<T extends Comparable & Iterable> T foo() { return null; }
	// Repeats are not allowed, compilation error
	<T extends Comparable & Iterable & Comparable> T bar() { return null; }
}
```

GerardPaligot:
Closes #624

### DIFF
--- a/src/test/java/spoon/test/type/TypeTest.java
+++ b/src/test/java/spoon/test/type/TypeTest.java
@@ -40,6 +40,7 @@ import spoon.test.type.testclasses.Pozole;
 
 import java.io.Serializable;
 import java.util.List;
+import java.util.stream.Collectors;
 
 import static junit.framework.TestCase.assertFalse;
 import static org.junit.Assert.assertEquals;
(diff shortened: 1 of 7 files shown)


src/test/java/spoon/test/type/TypeTest.java
```diff
@@ -174,8 +175,8 @@ public void testIntersectionTypeReferenceInGenericsAndCasts() throws Exception {
 		assertTrue(generic.getBoundingType() instanceof CtIntersectionTypeReference);
 		assertEquals("java.lang.Runnable & java.io.Serializable", generic.getBoundingType().toString());
 		final CtIntersectionTypeReference<?> superType = generic.getBoundingType().asCtIntersectionTypeReference();
-		assertEquals(aPozole.getFactory().Type().createReference(Runnable.class), superType.getBounds().get(0));
-		assertEquals(aPozole.getFactory().Type().createReference(Serializable.class), superType.getBounds().get(1));
+		assertEquals(aPozole.getFactory().Type().createReference(Runnable.class), superType.getBounds().stream().collect(Collectors.toList()).get(0));
```
monperrus: should be replaced by a contains, the order of a set may change

src/test/java/spoon/test/type/TypeTest.java
```diff
@@ -235,8 +236,8 @@ public void testIntersectionTypeOnTopLevelType() throws Exception {
 		assertNotNull(ref.getBoundingType());
 		assertTrue(ref.getBoundingType() instanceof CtIntersectionTypeReference);
 		assertEquals(2, ref.getBoundingType().asCtIntersectionTypeReference().getBounds().size());
-		assertEquals(Number.class, ref.getBoundingType().asCtIntersectionTypeReference().getBounds().get(0).getActualClass());
-		assertEquals(Comparable.class, ref.getBoundingType().asCtIntersectionTypeReference().getBounds().get(1).getActualClass());
+		assertEquals(Number.class, ref.getBoundingType().asCtIntersectionTypeReference().getBounds().stream().collect(Collectors.toList()).get(0).getActualClass());
```
monperrus: idem

src/test/java/spoon/test/type/TypeTest.java
GerardPaligot: A comparator is used based on the `SourcePosition`: https://github.com/INRIA/spoon/pull/659/files#diff-9cd2517bedc2b7b0aba6d67ef583fee4R70

PR #694, opened 2016-06-08
Always allow null parameter for setters.

GerardPaligot:


monperrus:
still failing?

GerardPaligot:
Think to add a test case which call the setter with null argument.

GerardPaligot:
@monperrus PR ok for me.

PR #696, opened 2016-06-09
fix: doesn't adjust position for implicit source code

petitpre:


monperrus:
do you have a test case? Thanks!

petitpre:
No, but you can contribute :p

This modification only remove warning when the pretty printer cannot adjust position of implicit elements...

monperrus:
We don't have a single assertion about the line-preserving mode, it
would be great to have some.

If you use this mode, you should be able to write a nice test :-)

GerardPaligot:
Thanks @petitpre for this PR.

PR #705, opened 2016-06-16
Fix a number of bugs related to handling of unnamed packages

leventov:


monperrus:
very interesting PR! can you add failing assertions that demonstrate the bug?

leventov:
I don't understand why tests are so unstable. Why they pass locally and fail in CI. Local results from-ide and from console via mvn test are also different.

tdurieux:
I maybe because your pull request does not contains all modification present in master.
Can you rebase your PR?

GerardPaligot:
I pull our branch and got same errors than Travis, Jenkins or Appveyor.

PR on your PR done here: https://github.com/leventov/spoon/pull/5

leventov:
@GerardPaligot please wait for a minute. I will try to put everything together.

leventov:
@monperrus for unnamed package issues it would be difficult for me.

For java.lang importing issues, demonstrating case is easy: declare a class clashing with some class in java.lang, i. e. com.mycompany.String, and use it from another class. It should stay fully-qualified when pretty-printed. Before this PR spoon imports it and uses unqualifed, and output is not compilable because java compiler doesn't know which class to use, com.mycompany.String or java.lang.String.

src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
```diff
@@ -2027,6 +2027,23 @@ public void visitCtTypeParameterReference(CtTypeParameterReference ref) {
 		}
 	}
 
+	private boolean printQualified(CtTypeReference<?> ref) {
+		if (importsContext.isImported(ref)) {
+			// If java.lang.Something is imported, but we are in the context of a class which is
+			// also called "Something", we should still use qualified version java.lang.Something
+			if (ref.getPackage() != null && ref.getPackage().getSimpleName().equals("java.lang")) {
```
leventov: As an afterthought, this shouldn't probably be restricted with java.lang, but for any imported classes.

src/main/java/spoon/reflect/visitor/ImportScannerImpl.java
```diff
@@ -218,7 +226,44 @@ private boolean addImport(CtTypeReference<?> ref) {
 		if (imports.containsKey(ref.getSimpleName())) {
 			return isImported(ref);
 		}
+		// don't import unnamed package elements
+		if (ref.getPackage() == null || ref.getPackage().isUnnamedPackage()) {
+			return false;
+		}
+		if (!ref.getPackage().getSimpleName().equals("java.lang")) {
+			if (classNamePresentInJavaLang(ref)) {
```
monperrus: those two consecutive ifs look like an oxymoron. can we rewrite them in a more maintainable way?

src/main/java/spoon/reflect/visitor/ImportScannerImpl.java
```diff
@@ -218,7 +226,44 @@ private boolean addImport(CtTypeReference<?> ref) {
...
+			// not produces "false positives" if the analyzed source complianceLevel is > 6.
+			// For example, it reports that FunctionalInterface is present in java.lang even
+			// for compliance levels 6, 7. But this is not considered a bad thing, in opposite,
+			// it makes generated code a little more compatible with future versions of Java.
+			if (namesPresentInJavaLang8.contains(ref.getSimpleName())
+					|| namesPresentInJavaLang9.contains(ref.getSimpleName())) {
+				presentInJavaLang = true;
+			} else {
+				// Assuming Spoon's own runtime environment is Java 7+
+				try {
```
monperrus: could we do only this try catch in all cases?

src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
monperrus: what do you mean?

leventov: Probably this should be checked for all classes, i. e. in terms of code, just remove `ref.getPackage().getSimpleName().equals("java.lang")` condition

src/main/java/spoon/reflect/visitor/ImportScannerImpl.java
leventov: first line is about that this class is not `java.lang`. Second line is about it's name is in java lang, i. e. com.mycompany.String.

src/main/java/spoon/reflect/visitor/ImportScannerImpl.java
leventov: If spoon's runtime is Java 7 and analyzed code is Java 8 it won't skip e. g. `FunctionalInterface`.

src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
monperrus: could you update this PR?

src/main/java/spoon/reflect/visitor/ImportScannerImpl.java
monperrus: got it.

src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
leventov: ok

monperrus:
Thanks. @GerardPaligot if you're also OK, do you merge?

GerardPaligot:
Thanks for this contribution @leventov!

PR #713, opened 2016-06-17
Refactor AnnotationFactory's factory methods

leventov:
So that they don't try to get annotation class in Spoon's runtime, because the annotation type shouldn't be in the spoon's runtime

### DIFF
--- a/src/main/java/spoon/reflect/factory/AnnotationFactory.java
+++ b/src/main/java/spoon/reflect/factory/AnnotationFactory.java
@@ -89,9 +89,9 @@ public class AnnotationFactory extends TypeFactory {
 	 * 		the value of the annotation element
 	 * @return the created/updated annotation
 	 */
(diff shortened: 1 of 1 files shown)


GerardPaligot:
That should be nice to deprecate rather than change methods declaration.

leventov:
@GerardPaligot  the set of methods is not changed, I just switched method signatures and bodies.

GerardPaligot:
> I just switched method signatures

Yep. If a user uses `getFactory().Annotation().annotate(...)`, its source code will be broken with your changes.

leventov:
@GerardPaligot I don't see, how. As far as I can see AnnotationFactory's API is not changed by this commit.

tdurieux:
@GerardPaligot Leventov is right the API does not change,

GerardPaligot:
Ok, I got it.

PR #723, opened 2016-06-24
fix(comment): Allows usage of the new flag only.

### LINKED ISSUE #721: Comments not included in pretty print unless both generateJavadoc and commentsEnabled are set
I had to set the option `generateJavadoc` (which is deprecated), not just `commentsEnabled`, to have all comments printed. (I think the problem was only triggered if there was a class-level javadoc.)

I think the boolean logic in the `getComment`/`printComment` methods of `DefaultJavaPrettyPrinter` like this:

    if (!env.isGenerateJavadoc() || !env.isCommentsEnabled() || e == null)

should probably be something like this:

    if (!(env.isGenerateJavadoc() || env.isCommentsEnabled()) || e == null)

SNAPSHOT version: spoon-core-5.2.0-20160622.235104-91

GerardPaligot:
Closes #721

### DIFF
--- a/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
+++ b/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
@@ -1126,7 +1126,7 @@ public class DefaultJavaPrettyPrinter implements CtVisitor, PrettyPrinter {
 
 	@Override
 	public void visitCtComment(CtComment comment) {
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -165,7 +165,7 @@
   </scm>
 
   <properties>
-    <java.src.version>1.7</java.src.version>
```
monperrus: why this in this PR?

GerardPaligot: Mistake.

PR #727, opened 2016-06-28
Copy collections in CloneBuilder.

### LINKED ISSUE #726: Bug with setVisibility when cloning a method
Dear spooners,

When cloning a method and setting the visibility of the clone, the visibility is also modified on the original method. It looks like the original and the clone are entangled and share the same visibility.

The spoon.test.method.MethodTest#testClone test in the branch https://github.com/seintur/spoon/tree/ctmethodclone_bug illustrates the issue.

Lionel.

GerardPaligot:
Closes #726


monperrus:
do we have the same bug for type modifiers?

GerardPaligot:
The visibility is stored in modifiers. So yes. :)

PR #731, opened 2016-06-29
fix(clone): Clone type when create a field reference.

### LINKED ISSUE #722: After the execution of checkShadow the AST is inconsistent
I try to had a test that verifies positions of elements in the AST.
But after the execution of the test checkShadow the AST is inconsistent. 
Some parents does not point anymore to the correct element.

It's probably a bug in the shadow implementation (a missing clone?)

I create this test that founds 170 elements that does not have the good parent.
```java
private void checkParentConsistency(CtPackage pack) {
		Set<CtElement> inconsistentParents = new HashSet<>();
		new CtScanner() {
			private Deque<CtElement> previous = new ArrayDeque();
			@Override
			protected void enter(CtElement e) {
				if (e != null) {
					if (!previous.isEmpty()) {
						try {
							if (e.getParent() != previous.getLast()) {
								inconsistentParents.add(e);
							}
						} catch (ParentNotInitializedException ignore) {
							inconsistentParents.add(e);
						}
					}
					previous.add(e);
				}
				super.enter(e);
			}

			@Override
			protected void exit(CtElement e) {
				if (e == null) {
					return;
				}
				if (e.equals(previous.getLast())) {
					previous.removeLast();
				} else {
					throw new RuntimeException("Inconsistent stack");
				}
				super.exit(e);
			}
		}.visitCtPackage(pack);
		assertEquals("All parents have to be consistent", 0, inconsistentParents.size());
	}
```

GerardPaligot:
Closes #722

### DIFF
--- a/src/main/java/spoon/reflect/factory/FieldFactory.java
+++ b/src/main/java/spoon/reflect/factory/FieldFactory.java
@@ -107,7 +107,10 @@ public class FieldFactory extends SubFactory {
 	 * Creates a field reference from an existing field.
 	 */
 	public <T> CtFieldReference<T> createReference(CtField<T> field) {
-		return createReference(factory.Type().createReference(field.getDeclaringType()), field.getType(), field.getSimpleName());
+		final CtFieldReference<T> reference = createReference(factory.Type().createReference(field.getDeclaringType()), field.getType().clone(), field.getSimpleName());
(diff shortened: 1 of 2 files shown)


GerardPaligot:
@monperrus PR ok.

PR #741, opened 2016-07-05
WIP: Element Builder proposal

tdurieux:
I created this PR in order to discuss the architecture of a possible new Element Builder in spoon.

The current factories are neither easy to use nor explicit. This new builder aims to solve these issues.
But I don't know yet the best solution.

This PR contains a POC of the builder. I would like to have your opinion.

# Syntaxe

## BinaryOperator
```java
Builder B = factory.Builder();
CtBinaryOperator binary = B.Binary(
		B.Literal(2))
		.plus(B.Literal(1))
		.lower(
				B.Binary(B.Literal(10))
						.minus(B.Literal(5))
		).build();
Assert.assertEquals("(2 + 1) < (10 - 5)", binary.toString());
```

## Try

```java
Builder B = factory.Builder();
CtTry aTry = B.Try()
		.inBody(B.Increment(x.Literal(1)))
		.createCatch()
			.parameter("e", IllegalArgumentException.class)
			.inBody(x.Increment(x.Literal(1)))
			.close()
		.createCatch()
			.parameter("e", Exception.class)
			.close()
		.inFinally(x.Decrement(x.Literal(1)))
.build();
Assert.assertEquals("try {\n"
		+ "    1++;\n"
		+ "} catch (java.lang.IllegalArgumentException e) {\n"
		+ "    1++;\n"
		+ "} catch (java.lang.Exception e) {\n"
		+ "} finally {\n"
		+ "    1--;\n"
		+ "}", aTry.toString());
```
## Method

```java
Builder x = factory.Builder();

		CtReturn<Object> aReturn = factory.Core().createReturn();
		aReturn.setReturnedExpression(x.Literal(1).build());
		CtMethod m = x.Method("m")
				.type(int.class)
				.setPrivate()
				.setStatic()
				.addParam(int.class, "x")
				.inBody(aReturn).build();
		Assert.assertEquals("private static int m(int x) {\n"
				+ "    return 1;\n"
				+ "}", m.toString());
```
# Limitations
1. The builder requires a method to indicate the end of the element (here `CatchBuilder.close()`) in order to get back the parent Builder (here `TryBuilder`). It is not intuitive... 

2. This architecture will introduce a lot of duplicate codes. Example: all elements that have a body will have the method `inBody`. Unfortunately the inheritance will not solve this issue (in this case we need multi inheritance)

### DIFF
--- a/src/test/java/spoon/reflect/builder/BinaryBuilderTest.java
+++ b/src/test/java/spoon/reflect/builder/BinaryBuilderTest.java
@@ -1,0 +1,61 @@
+/**
+ * Copyright (C) 2006-2015 INRIA and contributors
+ * Spoon - http://spoon.gforge.inria.fr/
+ *
+ * This software is governed by the CeCILL-C License under French law and
+ * abiding by the rules of distribution of free software. You can use, modify
+ * and/or redistribute the software under the terms of the CeCILL-C license as
+ * circulated by CEA, CNRS and INRIA at http://www.cecill.info.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the CeCILL-C License for more details.
+ *
+ * The fact that you are presently reading this means that you have had
+ * knowledge of the CeCILL-C license and that you accept its terms.
+ */
+package spoon.reflect.builder;
+
+import org.junit.Assert;
+import org.junit.Test;
+import spoon.Launcher;
+import spoon.reflect.code.CtBinaryOperator;
+import spoon.reflect.code.CtUnaryOperator;
+import spoon.reflect.factory.Factory;
+
+public class BinaryBuilderTest {
+
+	@Test
+	public void multiBuilderTest() {
+		Factory factory = new Launcher().getFactory();
+		Builder B = factory.Builder();
+		CtUnaryOperator not = B.Not(B.Binary(B.Literal(2)).lower(B.Literal(1))).build();
+		Assert.assertEquals("!(2 < 1)", not.toString());
+	}
+
+	@Test
+	public void simpleBinaryTest() {
+		Factory factory = new Launcher().getFactory();
+		Builder B = factory.Builder();
+		CtBinaryOperator binary = B.Binary(
+				B.Literal(2))
+				.multiplication(B.Literal(1))
+				.equals(B.Literal(2))
+				.build();
+		Assert.assertEquals("(2 * 1) == 2", binary.toString());
+	}
+
+	@Test
+	public void multiBinaryTest() {
+		Factory factory = new Launcher().getFactory();
+		Builder B = factory.Builder();
+		CtBinaryOperator binary = B.Binary(
+				B.Literal(2))
+				.plus(B.Literal(1))
+				.lower(
+						B.Binary(B.Literal(10))
+								.minus(B.Literal(5))
+				).build();
+		Assert.assertEquals("(2 + 1) < (10 - 5)", binary.toString());
+	}
+}
--- a/src/test/java/spoon/reflect/builder/TryBuilderTest.java
+++ b/src/test/java/spoon/reflect/builder/TryBuilderTest.java
@@ -1,0 +1,54 @@
+/**
+ * Copyright (C) 2006-2015 INRIA and contributors
+ * Spoon - http://spoon.gforge.inria.fr/
+ *
+ * This software is governed by the CeCILL-C License under French law and
+ * abiding by the rules of distribution of free software. You can use, modify
+ * and/or redistribute the software under the terms of the CeCILL-C license as
+ * circulated by CEA, CNRS and INRIA at http://www.cecill.info.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the CeCILL-C License for more details.
+ *
+ * The fact that you are presently reading this means that you have had
+ * knowledge of the CeCILL-C license and that you accept its terms.
+ */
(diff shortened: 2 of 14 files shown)


alcides:
Love it! I really need this!

monperrus:
looks like javapoet for Spoon?

tdurieux:
I don't know

GerardPaligot:
@monperrus Yes but better, with love and for Spoon! <3

src/test/java/spoon/reflect/builder/BinaryBuilderTest.java
```diff
@@ -0,0 +1,61 @@
...
+		Builder B = factory.Builder();
+		CtUnaryOperator not = B.Not(B.Binary(B.Literal(2)).lower(B.Literal(1))).build();
+		Assert.assertEquals("!(2 < 1)", not.toString());
+	}
+
+	@Test
+	public void simpleBinaryTest() {
+		Factory factory = new Launcher().getFactory();
+		Builder B = factory.Builder();
+		CtBinaryOperator binary = B.Binary(
```
monperrus: infix notation? `Lit(2).times(Lit(3))`

src/test/java/spoon/reflect/builder/TryBuilderTest.java
```diff
@@ -0,0 +1,54 @@
...
+public class TryBuilderTest {
+
+	@Test
+	public void test() {
+		Factory factory = new Launcher().getFactory();
+		Builder x = factory.Builder();
+
+CtTry aTry = x.Try()
+		.inBody(x.Increment(x.Literal(1)))
+		.createCatch()
```
monperrus: createCatch("e", IllegalArgumentException.class)

monperrus: .body(...)

monperrus:
I love this idea. It will be an excellent alternative for intercession in addition to the core intercession API, snippets and templates.

monperrus:
generate the builder with https://github.com/google/FreeBuilder?

pvojtechovsky:
>The builder requires a method to indicate the end of the element (here CatchBuilder.close()) in order to get back the parent Builder (here TryBuilder). It is not intuitive...

What about API like this?
```java
Builder B = factory.Builder();
CtTry aTry = B.Try()
        .inBody(B.Increment(x.Literal(1)))
        .createCatch(B.Catch("e", IllegalArgumentException.class)
            .inBody(x.Increment(x.Literal(1)))
        ).createCatch(B.Catch("e", Exception.class)
        ).inFinally(x.Decrement(x.Literal(1)))
.build();
```
the close() can be avoided by building whole catch content as a parameter of createCatch(...)

tdurieux:
Why not but there are a lot of parenthesis and call to the factory (that I was trying to avoid)

pvojtechovsky:
```java
CtTry aTry = B.Try()
        .inBody(B.Increment(x.Literal(1)))
        .createCatch()
            .parameter("e", IllegalArgumentException.class)
            .inBody(x.Increment(x.Literal(1)))
            .close()
        .createCatch()
            .parameter("e", Exception.class)
            .close()
        .inFinally(x.Decrement(x.Literal(1)))
.build();

**Comment by tdurieux** [comment]
I did not understand your message sorry

**Comment by tdurieux** [comment]
> I do not prefer any yet. Just giving ideas into this nice concept, to help to move it forward 

Yes, it is very hard to create a good design for this builder.

> or just rename of methods?

I prefer this solution

Or use the build method

**Comment by pvojtechovsky** [comment]
> there are a lot of parenthesis and call to the factory (that I was trying to avoid)

May be we can avoid other calls of factory too
```java
CtTry aTry = B.Try()
        .beginTry()
            .Increment(x.Literal(1)))
        .endTry()
        .beginCatch()...
```
the `endTry()`, cannot be replaced by `end()`, because `endTry()` returns CtTryBuilder instance. The consistency of begin/end pairs would be assured in runtime by builder.

But I agree that this idea is really wild. So parenthesis and use of factory is better in this case, because we are building another spoon model element.

So the rule would be:
* building of each spoon model element starts on factory
* fluent API is used to fill all attributes of that one spoon model element

tdurieux:
I would said:

CtTry aTry = B.Try()
        .beginBody()
            .Increment(x.Literal(1)))
        .endBody()
        .beginCatch()...
```

**Comment by pvojtechovsky** [comment]
> This architecture will introduce a lot of duplicate codes.

it is no problem, if code of builder will be generated.

**Comment by pvojtechovsky** [comment]
```java
CtTry aTry = B.Try()
.beginBody()
   .Increment(x.Literal(1)))
.endBody()
.beginCatch()
    .parameter("e", IllegalArgumentException.class)
    .beginBody()
       .Increment(x.Literal(1)))
     .endBody()
.endCatch()
```
The problem of this API is "What is the return type of `endBody() method"? Will be the same endBody method used in building of CtMethod?
A) If `endBody` returns generic `SpoonModelBuilder` class which is model element independent, then it there will be a lot of methods and API client will be not guided.
B) If `endBody` returns CtTry specific `CtTryBuilder` class, then how will be the `Increment(..)` method implemented to be able to return CtTryBuilder ??
Therefore I suggested `endTry` ... but I do not think that it is good way at all.

tdurieux:
You are right, I did not think enough

pvojtechovsky:
One more idea ... to use lambda expression to generate nested elements. Like this:
```java
CtTry aTry = factory.Builder().Try()
        .inBody(body ->
             body.Increment(x.Literal(1)))
        .createCatch(IllegalArgumentException.class, e, body->
             body.statement(st -> st.Increment(x.Literal(1)))
                     .statement(st -> st.invoke("java.lang.System.out.println()"))
        ).createCatch(Exception.class, "e")
        .inFinally(body -> 
             body.statemenet(st -> st.Decrement(x.Literal(1)))
.build();
```
* autocompletion will work on each level
* builder code has no duplicities

tdurieux:
I can be a solution but it is only nice for Java 8.
In Java 7 it is not practicable but I don't know if it is a problem

PR #758, opened 2016-07-22
Resolve Literal#constant if required. Fixes #755.

### LINKED ISSUE #755: NullPointerException with DoubleLiteral
Hi,

I get a `NullPointerException` when creating a spoon model for the following code:

```java
import java.util.HashMap;
import java.util.Map;

public class DoubleLiteralBug {

	public static void main(final String[] args) {
		Map<String, UknownClass> map = new HashMap<>();

		map.forEach((key, value) -> {
			double d = 1d;
		});
	}
}
```

The model is created as follows:
(with `projectPath` being the path to the `DoubleLiteralBug.java` file)

```java
Launcher runLaunch = new Launcher();
runLaunch.getEnvironment().setNoClasspath(true);
runLaunch.addInputResource(projectPath);
runLaunch.buildModel();
```

I noticed that the error can be removed by either replacing `UknownClass` with a class available in classpath, or by moving `double d = 1d;` out of the `forEach` lambda.

retuxx:
In some cases (an example is given in #755) Literal#constant may
be `null` and, thus, causes a NullPointerException when calling
`localDeclaration.initialization.traverse(this, scope);`. This
patch fixes the issue by resolving Literal#constant if required.

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilder.java
+++ b/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilder.java
@@ -61,6 +61,7 @@ import org.eclipse.jdt.internal.compiler.ast.InstanceOfExpression;
 import org.eclipse.jdt.internal.compiler.ast.IntLiteral;
 import org.eclipse.jdt.internal.compiler.ast.LabeledStatement;
 import org.eclipse.jdt.internal.compiler.ast.LambdaExpression;
+import org.eclipse.jdt.internal.compiler.ast.Literal;
 import org.eclipse.jdt.internal.compiler.ast.LocalDeclaration;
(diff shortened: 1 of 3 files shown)


monperrus:
Excellent, thanks! Would you add a test case showing that the bug has disappeared?

GerardPaligot:
AppVayor seems to fail for bad reasons here. I relaunched it.

@retuxx do you find time to add a test case on your PR?

retuxx:
Working on it.

GerardPaligot:
Thanks! :)

If you need help, come back to us.

retuxx:
Test case has been added. Fixed another issue that occurs if a Literal is a StringLiteral and scope is null.

retuxx:
Oh sorry, looks like I forgot to check style guide.

PR #760, opened 2016-07-22
Skip `findImport` and `createPackage` if necessary. Fixes #759.

### LINKED ISSUE #759: ArrayIndexOutOfBoundsException with wildcard import and package name in super class
The following example causes an `ArrayIndexOutOfBoundsException` while create a spoon model:

```java
import antlr.*;

public class FindImportBug extends antlr.LLkParser {
    public static void main(final String[] args) {}
}
```

The model is created as follows:
(with projectPath being the path to the FindImportBug.java file)

```java
Launcher runLaunch = new Launcher();
runLaunch.getEnvironment().setNoClasspath(true);
runLaunch.addInputResource(projectPath);
runLaunch.buildModel();
```

I think this issue could be solved by [checking the length `chars`](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java#L252). That is,  `findImport(chars, false, false);` is skipped if `chars.length == 0` .

retuxx:
`findImport` and `createPackage` require an array with a minimum
length of 1. This patch skips the method calls if necessary and
properly redirects to the fall back.


monperrus:
Thanks a lot. Would you add a test case with an assert showing that the bug has disappeared?

GerardPaligot:
Same here, AppVayor relaunched.

Issue #756, opened 2016-07-22
Bug with variable declaration resolving after clone


fedorov-s-n:
# Problem
After cloning AST subtree variable references are resolved to wrong declaration that makes it difficult to perform further modifications.

# Example
Suppose there's a code:

    public class A {

        int field;

        public void b(int param) {
            IntUnaryOperator f1 = x -> x;
            IntBinaryOperator f2 = (a, b) -> a + b;
            try {
                System.out.println(f1.applyAsInt(f2.applyAsInt(field, param)));
            } catch (RuntimeException e) {
                throw e;
            }
        }
    }

Goal is to clone class `A`, rename copy to `B` and perform some modifications with `B`.

# Problems with example

  1. Local variables references `f1`, `f2` in `B.b()` still refer to declarations in `A.b()`. The same is true about Catch variable reference `e`
  2. Method parameter reference and and class field reference are resolved to declarations in `A.b()` and `A` correspondingly because they resolves using type reference that points to `A`
  3. Resolving variables using plain field reference (local and catch variables) and dynamic lookup (field and parameter variables) are different strategies, and that's not clear that different variable types use different strategies. Developers who use Spoon API should look to sources to understand it.
  4. In scenario that extensively uses resolving declaration by reference lookups took longer than field read.

# Proposed solution

  1. remove `setDeclaringType` and `setDeclaringExecutable` methods, references are not declared! (fields and parameters are)
  2. add `setDeclaration` method to `CtVariableReference`
  3. cache declaration in reference if it was already resolved, allow to invalidate cache
  4. set unified resolving strategy for all variable types

# Prototype code:

    public interface CtVariableReference<T, V extends CtVariable<T>> extends CtReference {

        @Override
        public V getDeclaration();

        public void setDeclaration(V declaration);
    }

    public abstract class CtVariableReferenceImpl<T, V extends CtVariable<T>> extends CtReferenceImpl implements CtVariableReference<T, V> {

        private V declaration = null;

        @Override
        public V getDeclaration() {
            if (declaration == null) {
                declaration = lookupDeclaration();
            }
            return declaration;
        }

        /**
         *
         * @param declaration if not null, set declaration of this variable
         * reference; if null, set declaration as unknown and lookup it on next call
         * to `getDeclaration()`
         *
         */
        @Override
        public void setDeclaration(V declaration) {
            this.declaration = declaration;
        }

        protected abstract V lookupDeclaration();
    }

[...]

PR #764, opened 2016-07-26
Lookup dynamically the declaration of all variable references

GerardPaligot:
This PR is a proposal for the issue #756.

On the issue, the solution proposed @fedorov-s-n was good but have 2 problems (specified by him): change the API and can throw an exception `ParentNotInitializeException` if the model isn't complete.

Me, I propose a solution widely inspired by @fedorov-s-n (big thanks to him) but I remove disadvantages: I haven't any change in the API and I handle the exception.

What do you think about it?

### DIFF
--- a/src/main/java/spoon/generating/replace/CtListener.java
+++ b/src/main/java/spoon/generating/replace/CtListener.java
@@ -19,7 +19,7 @@ package spoon.generating.replace;
 import spoon.reflect.declaration.CtElement;
 
 class CtListener implements ReplaceListener<CtElement> {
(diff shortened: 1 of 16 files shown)


monperrus:
I like this fully backward compatible proposal.
TODO: 

* add asserts on field and method param reference
* add the test of @fedorov-s-n

GerardPaligot:
@monperrus Test integrated.

monperrus:
thanks a lot for the joint work @fedorov-s-n @GerardPaligot

PR #767, opened 2016-07-27
fix(annotation): Supports enum fields with Annotation#annotate() method.

### LINKED ISSUE #750: Enum Value as annotation type value not possible
I am trying to annotate a field via the annotation factory, and would like to set the value of an annotation element to an enum value, e.g.

`annoFactory.annotate(field, BoundNumber.class, "byteOrder", ByteOrder.LittleEndian);`

where ByteOrder.LittleEndian is the enum value. Running the annoation processor, I get the error message

> "Please, submit a valid value."

In CtAnnotationImpl.java, starting at line 88, the annotation values passed in are converted to CtExpression values, but I can not find any clause that would match en enum value. What is the proper way to convert an enum value as shown above to the according CtExpression instance?

GerardPaligot:
Closes #750

### DIFF
--- a/src/test/java/spoon/test/annotation/AnnotationValuesTest.java
+++ b/src/test/java/spoon/test/annotation/AnnotationValuesTest.java
@@ -4,18 +4,22 @@ import org.junit.Test;
 import spoon.reflect.code.CtConstructorCall;
 import spoon.reflect.code.CtExpression;
 import spoon.reflect.code.CtFieldAccess;
+import spoon.reflect.code.CtFieldRead;
 import spoon.reflect.code.CtLiteral;
 import spoon.reflect.code.CtNewArray;
 import spoon.reflect.declaration.CtAnnotation;
 import spoon.reflect.declaration.CtClass;
 import spoon.reflect.declaration.CtElement;
+import spoon.reflect.declaration.CtField;
 import spoon.reflect.declaration.CtType;
 import spoon.reflect.factory.Factory;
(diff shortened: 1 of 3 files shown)


src/test/java/spoon/test/annotation/AnnotationValuesTest.java
```diff
@@ -102,6 +105,18 @@ public void testAnnotationFactory() throws Exception {
...
 
+	@Test
+	public void testAnnotateWithEnum() throws Exception {
+		final Factory factory = createFactory();
+		final CtClass<Object> target = factory.Class().create("org.example.Tacos");
+		final CtField<String> field = factory.Field().create(target, new HashSet<>(), factory.Type().STRING, "field");
+		target.addField(field);
+
+		final CtAnnotation<BoundNumber> byteOrder = factory.Annotation().annotate(field, BoundNumber.class, "byteOrder", BoundNumber.ByteOrder.LittleEndian);
+		on(field).giveMeAnnotation(BoundNumber.class);
```
monperrus: assertEquals with byteOrder

src/test/java/spoon/test/annotation/AnnotationValuesTest.java
```diff
@@ -102,6 +105,18 @@ public void testAnnotationFactory() throws Exception {
...
+		on(byteOrder).giveMeAnnotationValue("byteOrder");
```
monperrus: add assert on result

GerardPaligot:
@monperrus PR ready.

PR #770, opened 2016-07-29
doc: improve CtLiteral doc

monperrus:


src/main/java/spoon/reflect/code/CtLiteral.java
```diff
@@ -34,6 +36,7 @@
 	 */
 	<C extends CtLiteral<T>> C setValue(T value);
 
+	/** Overriding return type, a clone of a CtLiteral returns a CtLiteral */
```
GerardPaligot: That seems obvious, no? :)

monperrus: IMHO explicitness is always better. non commented interface methods are
not good.

PR #771, opened 2016-07-29
fix: getAllMethods returns all methods, incl. those of libraries and java.lang.Object

monperrus:
thanks to the new and great runtime reflection API, we can have at last a correct implementation of getAllMethods


retuxx:
Does this include methods of libraries not available in classpath?

GerardPaligot:
@retuxx Yes. `CtTypeReference#getTypeDeclaration` uses the Java API reflexion to get all structural elements of external classes.

tdurieux:
@GerardPaligot  you are wrong if the class is not in the classpath the reflexion does not work...

PR #778, opened 2016-08-09
Added a caution about the java -jar command and also linked to examples on github

lukemerrick:
I'm not even sure if the original example command that uses the `-jar` flag is even correct, but in any case I added a cautionary note about the usage of the `-jar` flag when needing to specify a classpath in the command line (the `-jar` flag causes Java to ignore such classpath information), added an example of running Spoon without the `-jar` flag, and added a link to some example bash scripts I recently put up on GitHub in case anyone needs to look through a few more examples or wants to use a script to make things a little more streamlined.


monperrus:
Thanks a lot for this contribution. The doc problem has already been addressed in #777.

You scripts look interesting, and it would be good to have them in the main repo instead of a side one. 

Would you propose them in a pull request?

lukemerrick:
I'd be happy to, but I'm not 100% sure on where the best spot for usage examples is in the project. Is there possibly an examples folder buried somewhere in doc or src that I'm overlooking?

PR #787, opened 2016-08-15
adds bash script usage example

lukemerrick:
I apologize if my inexperience with GitHub collaboration and the layout of the Spoon project causes a bit of a headache here. I coudln't really find a "perfect spot" to drop the script within the project so I just suggested it at top-level since it thematically fits better with README type files than documentation or source code. Anyhow, this pull request simply proposes to add the bash script example from pull request #778 in a slightly cleaner way; I hope this is closer to what you were looking for.


monperrus:
thank a lot for the contribution.
could you put the scripts in folder `/doc/example_scripts/`?

monperrus:
Hi Luke, do you plan to move the scripts in `/doc/example_scripts/`? 
We can then proceeding with merging. Thanks.

lukemerrick:
Sorry for the delay. In the process of changing 13 timezones and moving back into college it had completely slipped my mind to fix this, my apologies.

doc/example_scripts/example_usage_script.sh
```diff
@@ -13,10 +13,13 @@ SPOON_JAR="lib/spoon.jar"
 ### 1) the fully qualified name of the processor to use (ex: com.example.spoonprocessors.MyProcessor)
 ### 2) the location of the source code to be spooned (ex: tester_project/src)
 ### 3) the destination for the spooned output source code (ex: spooned)
+### 4) [OPTIONAL AND ONLY USED IN OSX CONFIGURATION] the path to the processor source code (ex: src/com/example/spoonprocessors/MyProcessor.java)
```
GerardPaligot: Why "OPTIONAL AND ONLY USED IN OSX CONFIGURATION"?

doc/command_line.md
```diff
@@ -23,6 +23,8 @@ The basic usage of Spoon consists in defining the original source location and t
 $ java -classpath /path/to/binary/of/your/processor.jar:spoon-core-4.2.0-jar-with-dependencies.jar spoon.Launcher -i /path/to/src/of/your/project -p fr.inria.gforge.spoon.processors.CatchProcessor
 ```
 
+If you plan to repeatedly run Spoon from the command line, it may be a good idea to combine all of your commands into a single bash script. An example of this can be found [here](https://github.com/INRIA/spoon/blob/master/example_usage_script.sh).
```
monperrus: update link?

doc/example_scripts/example_usage_script.sh
lukemerrick: OSX bash isn't quite the same as Linux, so this line simply provides a fix

monperrus:
thanks for your contribution.

Issue #600, opened 2016-04-19
It should be impossible to configure actual type arguments for CtTypeParameterReference


leventov:
methods like `addActualTypeArgument()` should throw exception, because this is illegal things to do. But they have some logic currently in `CtTypeParameterReferenceImpl`.

monperrus:
I'm not in favor of throwing an exception because it would require client code to add ugly try/catch.

however, you're right, and there should not be be code in  CtTypeParameterReference.addActualTypeArgument. We'll try to remove it and replace it by an appropriate comment.

Issue #597, opened 2016-04-19
support for analyzing bound vs unbound type references (roadmap point)


monperrus:
solution may be instanceof CtTypeParameterReference

contract in mainTest:
if ref instanceof CtTypeParameterReference, then 
assert ref.getSimpleName declaredas formal in the scope (in one of the parents)

GerardPaligot:
Make this one first: https://github.com/INRIA/spoon/issues/606

PR #799, opened 2016-08-24
Fix 573 597 600

### LINKED ISSUE #573: getMethod(String, CtTypeReference...) doesn't work with generics
With this snippet:

```java
public <T> void method1(T t) {
}

public <T extends String> void method1(T t) {
}
```

If I call `getMethod("method1", int.class)` on a `CtType`, it returns the wrong method.

GerardPaligot:
Fix #573 #597 #600


src/test/java/spoon/test/main/MainTest.java
```diff
@@ -91,6 +92,21 @@ public void checkGenericContracts(CtPackage pack) {
...
+		// type parameter reference.
+		checkBoundAndUnboundTypeReference(pack);
+	}
+
+	private void checkBoundAndUnboundTypeReference(CtPackage pack) {
+		new CtScanner() {
+			@Override
+			public void visitCtTypeParameterReference(CtTypeParameterReference ref) {
+				if (!(ref.getParent() instanceof CtReference)) {
+					assertEquals(ref.getSimpleName(), ref.getDeclaration().getSimpleName());
```
monperrus: only in Spoon mode:  (ref.getDeclaration!=null)

monperrus:
CtTypeInformation {

   /** returns true if this element is a generics (eg "T") and false if it is an actual type (eg 'Book" or "String") */
   isGenerics() {
     ....
  }

}

GerardPaligot:
@monperrus PR ok.

monperrus:
could you add somewhere in a test a specification of isGenerics assertTrue(isGenerics) and assertFalse(isGenerics)?

PR #800, opened 2016-08-24
fix: Used types in a snippet statement.

### LINKED ISSUE #650: Bug in CtType#getUsedTypes for types in root package
CtType#getUsedTypes fails with a NPE when invoked for a type declared in the root package (i.e. no package).

The following branch contains a test that exhibits the issue: https://github.com/seintur/spoon/tree/getUsedTypes_rootPackage_bug

The reason why I'm still wondering about a potential fix is that I'd like the contract of CtType#getPackage() to be clarified.

For types in the root package, should CtType#getPackage() return:
* null
* or CtRootPackage (defined in CtModelImpl)?

Same question, asked differently: What is the condition to check to assess whether a type is defined in the root package or not?

Cheers,
Lionel.

GerardPaligot:
Closes #650


GerardPaligot:
@monperrus PR ok.

PR #809, opened 2016-08-30
Avoid NullPointerException in CtExecutableReferenceImpl.

### LINKED ISSUE #808: CtExecutableReferenceImpl may throw a NullPointerException in getDeclaration
I found that `CtExecutableReferenceImpl#getDeclaration` may throw a `NullPointerException` in noclasspath mode. Unfortunately, I can not provide a short example but it definitely occurs. The following snippet shows it's implementation:

```java
@Override
public CtExecutable<T> getDeclaration() {
	return getCtExecutable(getDeclaringType().getDeclaration());
}
```

As we can see, `getDeclaration` calls `getDeclaringType()` which, according to it's JavaDoc, may return `null`. By calling `getDeclaringType().getDeclaration()` a `NullPointerException` is throws if `getDeclaringType()` is `null`.

retuxx:
According to JavaDoc, `getDeclaringType()` may return null. Until now,
`getDeclaration`, does not handle the null case and, therefore, may
throw a NullPointerException. This exception is thrown most likely in
noclasspath mode.  Fixes #808.

### DIFF
--- a/src/main/java/spoon/support/reflect/reference/CtExecutableReferenceImpl.java
+++ b/src/main/java/spoon/support/reflect/reference/CtExecutableReferenceImpl.java
@@ -139,7 +139,8 @@ public class CtExecutableReferenceImpl<T> extends CtReferenceImpl implements CtE
 	@Override
 	@SuppressWarnings("unchecked")
 	public CtExecutable<T> getDeclaration() {
-		return getCtExecutable(getDeclaringType().getDeclaration());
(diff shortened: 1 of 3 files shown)


retuxx:
As you can see in the test case, the unknown method has to be called from an Array. Otherwise, the current implementation returns `null` instead of throwing a `NullPointerException`. Maybe the AST generator has to be fixed too, since it looks like it is missing to set `declaringType` in case of Arrays.

monperrus:
thanks

PR #811, opened 2016-08-31
Create CtTypeAccess/CtTypeReference for ArrayReferences in noclasspath mode and static context.

retuxx:


retuxx:
Once  #809 is merged I can extend the provided example checking if `getDeclaringType != null`

alcides:
I believe github now allows pull requests against other branches than master. I don't know if this might help here.

src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
```diff
@@ -418,11 +421,20 @@ CtReference getDeclaringReferenceFromImports(char[] expectedName) {
...
+					typeAccess = helper.createTypeAccessNoClasspath((SingleNameReference) messageSend.receiver);
+				} else if (messageSend.receiver instanceof QualifiedNameReference) {
+					typeAccess = helper.createTypeAccessNoClasspath((QualifiedNameReference) messageSend.receiver);
+				} else if (messageSend.receiver instanceof ArrayReference) {
+					typeAccess = helper.createTypeAccessNoClasspath((ArrayReference) messageSend.receiver);
+				} else {
+					throw new SpoonException();
+				}
+				ref.setDeclaringType(typeAccess.getAccessedType());
+			} catch (final SpoonException e) { /* ignore */ }
```
monperrus: really all SpoonException?

monperrus:
do you have a test case?

src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
retuxx: Since it is noclasspath mode any of these function may throw a `SpoonException` if they were not able to create a type access. Do you have a better idea?

retuxx:
Is comming

retuxx:
Added an example that shows that even multi dimensional arrays are processed.

src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
retuxx: I could replace the `SpoonException` with a `null` return. It's up to you.

monperrus: An ignored exception is always a threat to debugging. Is it necessary? What if you don't catch anything in this method?

retuxx: Ok, I will remove the exception handling.

monperrus: ok, thanks.

GerardPaligot:
Big thanks for this PR @retuxx! I'm happy to see that someone else can contribute to these new JDT builders. (:

GerardPaligot:
This PR need a rebase.

retuxx:
I still need to remove the catch block and fix some other issues. I will try to fix it the next days, currently I'm in a rush. Sorry for that.

monperrus:
@retuxx , should we close this old one?

retuxx:
I'm still working on it but it is hard to test. If I remove the catch blocks I get another error which only appears when using `mvn test`.

monperrus:
#884 may help the mvn test problem. will be merged soon.

retuxx:
> #884 may help the mvn test problem. will be merged soon.

The problem is, that when selecting one of the tests failing with `mvn test` and running them by their own, they do not fail any longer. That's very annoying...

monperrus:
> The problem is, that when selecting one of the tests failing with |mvn
> test| and running them by their own
with "mvn test -Dtest=pack.myTest" or in the IDE?

retuxx:
I tried to run a particular test from withing the IDE in order to have breakpoints.

monperrus:
OK, so the difference is between MVN and IDE, not between all-tests and
single test.

Did you rebase against master recently?

retuxx:
> OK, so the difference is between MVN and IDE

I don't thing so. I'm using IntelliJ which simply runs maven without further processing. Running `mvn test` from withing IntelliJ fails as well.

> Did you rebase against master recently?

Not yet.

monperrus:
closing this old inactive PR. Don't hesitate to re-open it if required.

Issue #813, opened 2016-09-01
Suspicious calculation of `sourceEnd` in `JDTTreeBuilderHelper#createTypeAccessNoClasspath`


retuxx:
In [`JDTTreeBuilderHelper#createTypeAccessNoClasspath`](https://github.com/INRIA/spoon/blob/0ba75bff9af07c87f1737b6a5dfd6bab34440101/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java#L426) the following code is used to calculate `sourceEnd`:

```java
int sourceEnd = (int) (positions[qualifiedNameReference.indexOfFirstFieldBinding - 1] >>> 32) - 2;
```

I'm wondering why you don't use:

```java
int sourceEnd = qualifiedNameReference.sourceEnd();
```
instead? I have a case in which `qualifiedNameReference.indexOfFirstFieldBinding` is `0` and, therefore, an `ArrayIndexOutOfBoundsException` is thrown.

PR #814, opened 2016-09-02
Fix access lookup in lambda for noclasspath mode.

retuxx:
This is a work in progress pull request showing my current work on an enhanced variable lookup for `SingleNameReference` tokens in lambda expressions (noclasspath mode). It took me some time to understand `JDTTreeBuilder` and all its helper classes but I think this is going into right direction, though.

As you might already have noticed, there are two test cases that fail. The first one is a good example for #813 and can be fixed by using `int sourceEnd = qualifiedNameReference.sourceEnd();`. The second test case fails because [`methodJDT.binding`](https://github.com/retuxx/spoon/commit/634ee11d4e57adaaa8a513d9df57c809ef97a6b4#diff-a490aa7d6a5ea006caa9168294d3d9f4R215) is `null` for anonymous classes`. Since I'm not sure how to create a type reference to an  anonymous class (yet), I was not able to properly handle this case.

Furthermore, [`getVariableDeclaration`](https://github.com/retuxx/spoon/commit/634ee11d4e57adaaa8a513d9df57c809ef97a6b4#diff-05a638c1f9cf369088d14c6b1e4ed0e7R156) is not finished yet as there might be a variable declared in a super class/interface. This case can be handled by iterating through all super classes/interfaces of a `TypeBinding` and checking if the potential variable has been declared in one of them. Unfortunately, I did not managed to create a `TypeDeclaration` from a `TypeBinding` yet. Maybe you have an idea how to do that.

However, in it's current state this pr produces much more precise AST in noclasspath mode as a `SingleNameReference` in a lambda expression is not always mapped to a `CtLocalVariableReference`. In addition, `getVariableDeclaration` reduces the amount of cloned code and ensures that a variable is visible in current context.


retuxx:
The recent commit fixes the anonymous class issue by creating an 'empty' `CtTypeReference` if `methodJDT.binding == null`.

retuxx:
I fixed the second test case as well by replacing:

```java
int sourceEnd = (int) (positions[qualifiedNameReference.indexOfFirstFieldBinding - 1] >>> 32) - 2;
```

with:

```java
int sourceEnd = qualifiedNameReference.sourceEnd();
```

as described in #813.

Two things are missing:
1) properly handle fields of super classes/interfaces and statically imported fields
2) add a test case showing that a type access is mapped correctly (not mapped to a variable access)
3) handle catch variables

monperrus:
Thanks, it looks interesting. However, "enhance" is vague, a PR is either a "bug fix" or a "new feature". What's this one? Could you add the associated test cases? Thanks.

retuxx:
> Could you add the associated test cases?

Working on it.

> What's this one?

It is a bug fix.

monperrus:
ok, I've renamed the PR.

retuxx:
The recent commit adds the first lines to search for variables imported statically or declared by super classes/interfaces. Anyhow, the creation of the CtField is still missing since I'm not entirely sure how to properly create one.

retuxx:
I have added a small test showing that a TypeAccess (Strings in our example) is mapped to a local variable access. Further tests with different variable access types, for instance, field access, statically imported variables and so on are coming soon. However, this simple test should give you an idea of what I'm trying to fix.

retuxx:
I just added another test with a field access. In current master this test fails to compile the code due to #813.

retuxx:
Recent commit should properly find variables in super classes/interfaces. I will add some test cases tomorrow.

src/main/java/spoon/support/compiler/jdt/ContextBuilder.java
```diff
@@ -112,55 +125,153 @@ void exit(ASTNode node) {
...
+		final InterfaceFactory interfaceFactory = jdtTreeBuilder.getFactory().Interface();
+		final ReferenceBuilder referenceBuilder = jdtTreeBuilder.getReferencesBuilder();
+		// there is some extra work to do if we are looking for CtFields (and subclasses)
+		final boolean lookingForFields = clazz == null
+				|| coreFactory.createField().getClass().isAssignableFrom(clazz);
+
+		// try to find the variable on stack beginning with the most recent element
+		for (final ASTPair astPair : stack) {
+			// the variable may have been declared directly by one of these elements
+			final List<U> variables = astPair.element.getElements(new AbstractFilter<U>() {
```
tdurieux: Why do you use an AbstractFilter instead of EarlyTerminatingScanner?

retuxx: I think there was a reason but let me have a look if an EarlyTerminatingScanner is feasible.

retuxx:
Recent commit uses EarlyTerminatingScanner instead of AbstractFilter making it faster in finding variables.

retuxx:
I think the PR should be ready now. I will write a summary later depicting what I improved/fixed.

retuxx:
So, here is the summary:

- I found that JDTTreeBuilder in noclasspath mode always creates a variable access for `SingleNameReference`s without proper binding ([link](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilder.java#L1336)), though, a `SingleNameReference` may point to a type access as well. (I've added a small example [here](https://github.com/INRIA/spoon/pull/814/files#diff-f9f8d8a41c9d4c82605534a4280d75e4R1)). It is easy to see that 'something' imported non-statically must be a type access because you can not import non-static variables. Therefore, I added [simple if statement](https://github.com/INRIA/spoon/pull/814/files#diff-072d4b8ceab9c4e090f675ba9342b6f9R1338) to JDTTreeBuilder checking if a `singleNameReference` points to variable or not. If it doesn't a type access is created instead (yes, it is some kind of heuristic but I will discuss this later).

- Until now, `JDTTreeBuilderHelper#createVariableAccessNoClasspath` creates a reference to a [lambda parameter](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java#L167) or reference to a [local variable](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java#L170) only, though, a `SingleNameReference` may point to a field, a catch variable, and so on, as well. Thus, I added some lines to this method determining the actual type of a (potential) variable. In order to retrieve the actual type another method (getVariableDeclaration, which will be explained later) is called. If the given name does not point to a variable ([`variable == null`](https://github.com/INRIA/spoon/pull/814/files#diff-a490aa7d6a5ea006caa9168294d3d9f4R177)) `null` is returned indicating that it must be something else. This return value is used by the JDTTreeBuilder as described above.

- Since  `JDTTreeBuilderHelper#createVariableAccessNoClasspath` allows to create a reference to a (potential) variable of arbitrary type, we need a function to determine the type of `SingleNameReference`s pointing to a variable. Hence, I was looking for a function providing this features and found something similar in `ContextBuilder`, more precisely, [getLocalVariableDeclaration](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/compiler/jdt/ContextBuilder.java#L115) and [getCatchVariableDeclaration](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/compiler/jdt/ContextBuilder.java#L142). Unfortunately, they only support local variables and catch variables. By looking closely to the implementation, one can see that one of the two methods is a clone of the other. Thus, I decided to generalize them by adding a method for arbitrary variables types (hey, something we were looking for) which is used them ([link](https://github.com/INRIA/spoon/pull/814/files#diff-05a638c1f9cf369088d14c6b1e4ed0e7R174)).

- There is a small but very significant [TODO](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/compiler/jdt/ContextBuilder.java#L117) in `getLocalVariableDeclaration` saying that we need to validate the actual scope of the found variable. I will give you an example:

```java
public class A {
    private final String title;
    ...
    public void aMethod(final String title) {
    ....
    }

    public void anotherMethod() {
        System.out.println(title); // <------
    }
}
```
In method `anotherMethod`, `title` obviously points to the field of `A` not the parameter of  `aMethod`. Now, the current implementation may return the parameter which definitely is a bug! So, `getVariableDeclaration` not only generalizes `getLocalVariableDeclaration` and `getCatchVariableDeclaration`, but also makes sure that the found variable is in current scope! Furthermore, neither `getCatchVariableDeclaration`, nor `getLocalVariableDeclaration` examined super classes/interfaces and static imports for potential variables with matching name. I added a [test case](https://github.com/INRIA/spoon/pull/814/files#diff-2113bfde9ebf2debe401c7847fb583a7R125) showing that this feature is necessary though.

- From `getVariableDeclaration`'s point of view, there is a situation in which we can not strictly decide whether a token points to a variable or a function call. It occurs if a static import of an unknown source for the given token is present. However, for noclasspath mode I added a heuristic checking if the name of the import consists of upper case characters only as, by Java Language Specification, this pattern is used for constant variables. I think it is the best we can do in this situation. Anyway, this heuristic does not affect [`JDTTreeBuilder#visit(SingleNameReference singleNameReference, BlockScope scope)`](https://github.com/INRIA/spoon/pull/814/files#diff-072d4b8ceab9c4e090f675ba9342b6f9R1323) as a function calls get visited by [`JDTTreeBuilder#visit(MessageSend messageSend, BlockScope scope)`](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilder.java#L1184)

- As a side effect of the fixes mentioned above, one of the existing test cases fails due to #813. The reason why it didn't failed before is, that without this pr the variable access was not mapped properly for this test and, therefore, (luckily) the execution path was omitted. I fixed it as well by applying the patch described in #813.

- Finally, I added a few test cases for all kind of variable accesses (hopefully I didn't miss one) with given source showing that my pr does not affect the regular case ([link](https://github.com/INRIA/spoon/pull/814/files#diff-2113bfde9ebf2debe401c7847fb583a7R99)).

So, I would be appreciated if you could review this commit the next days (and may merge it into master) because I we really need the changes for our study we want to conduct the next days :). I know that this pr contains a lot of changes but I hope I could explain why it is necessary.

src/test/java/spoon/test/lambda/LambdaTest.java
```diff
@@ -77,6 +80,63 @@ public void testLambdaExpressionWithExpressionBodyAndWithoutParameter() throws E
...
+		runLaunch.addInputResource("./src/test/resources/noclasspath/lambdas/TypeAccessInLambda.java");
+		runLaunch.buildModel();
+
+		assertEquals("The token 'Strings' has not been parsed as CtTypeAccess",
+				runLaunch.getModel().getElements(new Filter<CtTypeAccess>() {
+			@Override
+			public boolean matches(final CtTypeAccess element) {
+				return element.getAccessedType().getSimpleName().equals("Strings");
+			}
+		}).size(), 1);
```
tdurieux: The expected parameter must be the second parameter of `assertEquals`

src/test/java/spoon/test/lambda/LambdaTest.java
```diff
@@ -77,6 +80,63 @@ public void testLambdaExpressionWithExpressionBodyAndWithoutParameter() throws E
...
+						|| name.equals("pathSeparator")
+						|| name.equals("fieldInSeparateInterface")
+						|| name.equals("fieldInClassBase")
+						|| name.equals("fieldInClass")
+						|| name.equals("fieldInInterfaceBase")
+						|| name.equals("fieldInInterface")
+						|| name.equals("iAmToLazyForAnotherFieldName");
+			}
+		});
+		assertEquals(fieldAccesses.size(), 8);
```
tdurieux: he expected parameter must be the first parameter of assertEquals

src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java
```diff
@@ -147,32 +154,105 @@ static String createQualifiedTypeName(char[][] typeName) {
 	}
 
 	/**
-	 * In this case, we are in no classpath so we don't know if the access is a variable, a field or a type.
-	 * By default, we assume that when we don't have any information, we create a variable access.
+	 * Analyzes if {@code singleNameReference} points to a {@link CtVariable} visible in current
+	 * scope and, if existent, returns it's corresponding {@link CtVariableAccess}. Returns
```
tdurieux: it's -> its

src/test/java/spoon/test/lambda/LambdaTest.java
retuxx: done.

src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java
retuxx: done

tdurieux: Thanks

monperrus:
I think this is a very nice PR, thanks a lot. It improves much the noclasspath mode. Since it's a large PR, I wait until tomorrow for merging, to let the other contributors have a look at it.

monperrus:
Thanks a lot for your hard work on this touchy issue.

PR #818, opened 2016-09-06
fix(executable): fix the method ExecutableFactory#createReference

### LINKED ISSUE #817: API isStatic() under CtExecutableReference does not return true.  Under Spoon 5.2.0.
The API `isStatic()` under `CtExecutableReference` does not return the staticness.  Under Spoon 5.2.0.
Not sure whether this is a dropped feature or a bug. 

The input source code 
```
package temp;
public class A {
    public static int getInt1() { return 1; }
    public final static int getInt2() { return 2; }
    public final int getInt3() { return 3; }	
    public int getInt4() { return 4; }
}
```
The Spoon processor
```
package spoonbug;
public class ExecRefProcessor extends AbstractProcessor<CtClass> {
	@Override
	public void process(CtClass ctClass) {
        CtTypeReference typeRef = ctClass.getReference();
		Collection<CtExecutableReference<?>> collectExecRef = typeRef.getDeclaredExecutables(				
        for(CtExecutableReference execRef: collectExecRef)
		{
			System.out.println("" + execRef.getSimpleName() + " # isStatic()? " + execRef.isStatic() );
			System.out.println("" + execRef.getSimpleName() + " # isFinal()? " + execRef.isFinal() );
		}
	}
}
```

Output is as follows. Then `isStatic()` does not return true for `getInt1()` and `getInt2()`

```
getInt1 # isStatic()? false
getInt1 # isFinal()? false
getInt2 # isStatic()? false
getInt2 # isFinal()? true
getInt3 # isStatic()? false
getInt3 # isFinal()? true
getInt4 # isStatic()? false
getInt4 # isFinal()? false
<init> # isStatic()? false
<init> # isFinal()? false
```

tdurieux:
The method `ExecutableFactory#createReference` now supports static method.

Fix #817

### DIFF
--- a/src/test/java/spoon/test/executable/ExecutableTest.java
+++ b/src/test/java/spoon/test/executable/ExecutableTest.java
@@ -5,8 +5,10 @@ import spoon.Launcher;
 import spoon.reflect.code.CtBlock;
 import spoon.reflect.declaration.CtAnonymousExecutable;
 import spoon.reflect.declaration.CtType;
+import spoon.reflect.reference.CtExecutableReference;
 import spoon.reflect.visitor.Query;
 import spoon.reflect.visitor.filter.TypeFilter;
+import spoon.test.executable.testclasses.A;
 import spoon.test.executable.testclasses.Pozole;
 import spoon.testing.utils.ModelUtils;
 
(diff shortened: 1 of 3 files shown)


src/test/java/spoon/test/executable/ExecutableTest.java
```diff
@@ -40,4 +42,22 @@ public void testBlockInExecutable() throws Exception {
 		final CtType<Pozole> aPozole = ModelUtils.buildClass(Pozole.class);
 		assertTrue(aPozole.getMethod("m").getBody().getStatement(1) instanceof CtBlock);
 	}
+
+	@Test
+	public void testGetReference() throws Exception {
+		final CtType<A> aClass = ModelUtils.buildClass(A.class);
+
+		testMethodReference(aClass, "getInt1", true, false);
```
monperrus: could you please inline the assertions? it would be much more readable.

PR #842, opened 2016-09-22
Adds c2spoon in the Ecosystem of Spoon

tdurieux:


retuxx:
@tdurieux Is c2spoon in a usable state?

tdurieux:
It is a POC..., I created it to transform this benchmark: https://github.com/ProgramRepair/IntroClass to a java benchmark. 
It works well but this benchmark contains very simple C code.
c2spoon uses srcml which parses the C code into xml and then c2spoon transform the xml into a spoon AST, there is no compilation thus no type inference, no relation between invocation and declaration, ...

The repository c2spoon does not currently contains all the processors that transform  C methods into java methods. c2spoon it is only useful to statically analyse C with a java API.

retuxx:
> c2spoon uses srcml which parses the C code into xml and then c2spoon transforms the xml into a spoon AST. there is no compilation thus no type inference, no relation between invocation and declaration, ...

Are there any chances that c2spoon supports references etc. It sounds very interesting for toolchain (code smell detection).

tdurieux:
In fact I created two new spoon elements: 
* [CPointerElement](https://github.com/SpoonLabs/c2spoon/blob/master/src/main/java/fr/inria/spirals/c2spoon/spoon/CPointerElement.java)
* [CReferenceElement](https://github.com/SpoonLabs/c2spoon/blob/master/src/main/java/fr/inria/spirals/c2spoon/spoon/CReferenceElement.java)

to detect C references. The biggest limitation, the tool does not support structures, it is not difficult to support it but I did not need it.
I fact c2spoon use [srcml ](http://www.srcml.org/) that can parse C, C++, C# and java.
I can thus easily improve it to support more languages.

retuxx:
> In fact I created two new spoon elements:
    CPointerElement
    CReferenceElement

So, how do you get the declaration of a pointer if srcml does not give you this information?

tdurieux:
I have a system that binds method with declared method and binds variable accesses to local variable declarations.
I do not handled binding between elements from different files (because I did not needed) but it is probably possible to do it with the processor.

retuxx:
It would be really awesome if you could add this feature as it would simplify our multi language code smell detection tool.

tdurieux:
To be honest, the project is not a priority (not related to my PhD).
But the PR are welcome (as always (: )

PR #863, opened 2016-09-28
Allow Launcher to set Factory. Fixes #862.

### LINKED ISSUE #862: Allow Launcher To Set Factory
I'm working on #752 (currently it is for my project only but maybe I can make a PR of it later) and found that creating a second, partial model is not that easy due to the fact that both models (and all created elements) use different factories which, if I am right, yields to errors while resolving references. It would be nice if one could set the actual factory of a launcher before building the model. This way, I could reuse the factory of my main model and ensure that both, my old and new types, share the same factory.

retuxx:


monperrus:
Immutable objects are better. 

Could you do what you want to do with a new constructor "Launcher(Factory f)"

retuxx:
> Immutable objects are better. 

Indeed.

> Could you do what you want to do with a new constructor "Launcher(Factory f)"

Yes. Will you add this constructor or should I add another PR?

monperrus:
The best thing is to force-push a new commit on this PR/branch.

retuxx:
> The best thing is to force-push a new commit on this PR/branch.

done.

src/main/java/spoon/Launcher.java
```diff
@@ -125,12 +125,28 @@ public void printUsage() {
...
+	 * @param pFactory
+	 * 		The {@link Factory} that will be utilized in {@link #buildModel()}.
+	 * @throws IllegalArgumentException
+	 * 		If {@code pFactory == null}.
+	 */
+	public Launcher(final Factory pFactory) {
+		if (pFactory == null) {
+			throw new IllegalArgumentException("unable to create launcher with null factory");
+		}
+		factory = pFactory;
```
retuxx: Oops!

retuxx: Fixed.

tdurieux: You can also call this constructor from the previous one ```this(createFactory())``` in order to reduce code duplication.

retuxx: I tried but Java (for good reasons) does not allow to call member methods before `this` has been called. The actual compiler message is:

```bash
cannot reference this before supertype constructor has been called
```

tdurieux: This constructor will work.
```java
    public Launcher() {
		this(createFactory());
	}
```

retuxx: For me it doesn't. I tried both, OracleJDK (8) and OpenJDK (8).

tdurieux: My bad, createFactory is not static...

This one works but it is hugly.
```java
public Launcher() {
   this(new FactoryImpl(new DefaultCoreFactory(),  new StandardEnvironment()));
}
```

retuxx: I could add another static function to Launcher, or leave it as is (or use your approach). Either one is fine.

tdurieux: It is ok for my to keep it this way.

PR #873, opened 2016-09-30
getDeclaration of CtLocalVariable 

### LINKED ISSUE #868: getDeclaration of CtLocalVariable does not find declarations in for loops
I have the following (stripped version) code:

```java
package net.sf.jabref.collab;

import java.io.*;
import java.util.*;
import javax.swing.*;
import javax.swing.tree.DefaultMutableTreeNode;
import net.sf.jabref.*;
import net.sf.jabref.groups.*;
import net.sf.jabref.imports.*;


public class ChangeScanner extends Thread {
    
    
    /**
     * Finds the entry in neu best fitting the specified entry in old. If no entries get a score
     * above zero, an entry is still returned.
     * @param old EntrySorter
     * @param neu EntrySorter
     * @param index int
     * @return BibtexEntry
     */
    private BibtexEntry bestFit(EntrySorter old, EntrySorter neu, int index) {
        double comp = -1;
        int found = 0;
        loop: for (int i=0; i<neu.getEntryCount(); i++) {
            double res = Util.compareEntriesStrictly(old.getEntryAt(index),
            neu.getEntryAt(i));
            if (res > comp) {
                comp = res;
                found = i;
            }
            if (comp > 1)
                break loop;
        }
        return neu.getEntryAt(found);
    }
    
}
```

As you can see, `i` is a for loop variable that is referenced multiple times, for example by `neu.getEntryAt(i)`. I'm using a `CtScanner` to find declarations that are not referenced, also known as unused code. Nevertheless, if my visitor visits any variable reference referencing `i`, `getDeclaration()` is not able to `i's` declaration. The reason for this error is obvious, [CtLocalVariableReferenceImpl](https://github.com/INRIA/spoon/blob/f1baa5b9da2c7657f958ec73446db5412801e0d2/src/main/java/spoon/support/reflect/reference/CtLocalVariableReferenceImpl.java#L46) traverses all parent blocks one after another and [checks](https://github.com/INRIA/spoon/blob/f1baa5b9da2c7657f958ec73446db5412801e0d2/src/main/java/spoon/support/reflect/reference/CtLocalVariableReferenceImpl.java#L51) whether a statement of the current examined block is a CtLocalVariable and, if true, whether the variable is the one we are looking for. In case of a CtForLoop (let's call it `loop`), this code is missing `loop`'s declarations (`i` in our example). One solution to fix this error is to check whether a statement is a for loop (or any other statements allowing to declare variables in a new scope-lambdas maybe). However, you may have a better idea that is less brittle.

retuxx:
This should handle all kinds of elements, such as loops, lambdas and so on. Fixes #868.

### DIFF
--- a/src/test/java/spoon/test/reference/VariableAccessTest.java
+++ b/src/test/java/spoon/test/reference/VariableAccessTest.java
@@ -3,6 +3,7 @@ package spoon.test.reference;
 import org.junit.Test;
 import spoon.Launcher;
 import spoon.reflect.code.CtArrayWrite;
+import spoon.reflect.code.CtLiteral;
 import spoon.reflect.code.CtLocalVariable;
 import spoon.reflect.code.CtVariableAccess;
 import spoon.reflect.declaration.CtClass;
@@ -165,6 +166,29 @@ public class VariableAccessTest {
 		final Launcher launcher = new Launcher();
 		launcher.getEnvironment().setNoClasspath(true);
 		launcher.addInputResource("src/test/resources/reference-test/ChangeScanner.java");
+		launcher.buildModel();
+		new CtLocalVariableReferenceScanner().scan(launcher.getModel().getRootPackage());
+	}
+
+	@Test
+	public void testMultipleDeclarationsOfLocalVariable() {
+		final class CtLocalVariableReferenceScanner extends CtScanner {
+			@Override
+			public <T> void visitCtLocalVariableReference(
+					final CtLocalVariableReference<T> reference) {
+				assertNotNull(reference.getDeclaration());
+				final CtLocalVariable decl = reference.getDeclaration();
+				assertEquals(decl.getPosition().getLine(), 7);
+				assertTrue(decl.getDefaultExpression() instanceof CtLiteral);
+				final CtLiteral literal = (CtLiteral) decl.getDefaultExpression();
+				assertEquals(literal.getValue(), 42);
+				super.visitCtLocalVariableReference(reference);
+			}
+		}
+
+		final Launcher launcher = new Launcher();
+		launcher.getEnvironment().setNoClasspath(true);
+		launcher.addInputResource("src/test/resources/reference-test/MultipleDeclarationsOfLocalVariable.java");
 		launcher.buildModel();
 		new CtLocalVariableReferenceScanner().scan(launcher.getModel().getRootPackage());
 	}
--- a/src/test/java/spoon/test/reference/VariableAccessTest.java
+++ b/src/test/java/spoon/test/reference/VariableAccessTest.java
@@ -13,6 +13,7 @@ import spoon.reflect.declaration.CtType;
 import spoon.reflect.reference.CtLocalVariableReference;
 import spoon.reflect.reference.CtParameterReference;
 import spoon.reflect.reference.CtVariableReference;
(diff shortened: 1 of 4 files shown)


src/test/java/spoon/test/reference/VariableAccessTest.java
```diff
@@ -146,6 +147,24 @@ public void testReferences() throws Exception {
 		System.out.println(make);
 	}
 
+	@Test
+	public void testReferenceToLocalVariableDeclaredInLoop() {
+		final class CtLocalVariableReferenceScanner extends CtScanner {
+			@Override
+			public <T> void visitCtLocalVariableReference(
+					final CtLocalVariableReference<T> reference) {
+				assertNotNull(reference.getDeclaration());
```
monperrus: add an assert on reference.getDeclaration().getSimpleName()? and maybe on getType()?

src/test/resources/reference-test/ChangeScanner.java
```diff
@@ -0,0 +1,39 @@
...
+import net.sf.jabref.*;
+import net.sf.jabref.groups.*;
+import net.sf.jabref.imports.*;
+
+
+public class ChangeScanner extends Thread {
+    
+    
+    /**
+     * Finds the entry in neu best fitting the specified entry in old. If no entries get a score
```
seintur: neu?

retuxx: This is a stripped version of a single Java file in JabRef :). I think `neu` has been chosen to avoid new, which is a keyword.

src/test/java/spoon/test/reference/VariableAccessTest.java
retuxx: done

retuxx:
Oh wait, I think iteration should not stop a package but on class level. Will fix this asap.

retuxx:
A proper fix should be given in newest commit (see doc for more information).

monperrus:
I'm not sure about the algorithm. Consider

```
boolean i;
{
  int j;
}
int j;
i=j; <- this reference will point to the wrong i, and the test may fail
````
because of the order getElements()
is that correct?

retuxx:
> this reference will point to the wrong i, and the test may fail

Do you mean `j` instead of `i` because `i` has been declared only once. If yes, the implementation should find the second `j` because the parent of the first `j` is a block and not a parent of `i=j;` whereas the parent of the second `j` is a method (or something similar) and must be a parent of `i=j;`.

monperrus:
> Because the parent of the first j is a block

However, `parent.getElements(filter)` is a recursive descent and will match the first int j.

retuxx:
> However, parent.getElements(filter) is a recursive descent and will match the first int j.

That's why I added an additional check validating that the parent of a matched element must be the parent of the reference as well: https://github.com/INRIA/spoon/pull/873/files#diff-3fee2e106cbe6edb419441990c9f5febR77

retuxx:
I added another test showing you that my approach respects currently visible scope.

tdurieux:
I have something very similar in my code base, that I think is faster.
It uses a CtInheritanceScanner and loops on the parent until it find the variable declaration.
Currently I do not have the time to propose my solution.
We can merge this one and I can propose a nicer solution in the future.

monperrus:
Thanks a lot for the PR!

PR #879, opened 2016-10-04
fix(compilation): always uses the content of the ast to compile the ast

### LINKED ISSUE #878: Launcher.compile() does not compile the AST changes
Launcher.compile() only compiles the content of the original file...
The interest of this method is thus very limited.

tdurieux:
fixes #878

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/CompilationUnitWrapper.java
+++ b/src/main/java/spoon/support/compiler/jdt/CompilationUnitWrapper.java
@@ -16,57 +16,42 @@
  */
 package spoon.support.compiler.jdt;
 
-import org.apache.commons.io.IOUtils;
 import org.eclipse.jdt.internal.compiler.batch.CompilationUnit;
-import spoon.Launcher;
+import spoon.reflect.declaration.CtType;
+import spoon.reflect.visitor.DefaultJavaPrettyPrinter;
 
-import java.io.FileInputStream;
-import java.io.FileNotFoundException;
-import java.io.InputStream;
+import java.util.ArrayList;
+import java.util.List;
 
 class CompilationUnitWrapper extends CompilationUnit {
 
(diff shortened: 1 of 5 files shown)


src/test/java/spoon/test/compilation/CompilationTest.java
```diff
@@ -0,0 +1,91 @@
...
+						this.value = m.invoke(aClass.newInstance());
+					} catch (Exception e) {
+						Assert.fail(e.getMessage());
+					}
+					notify();
+				}
+			}
+		}
+
+		MethodRunner methodRunner = new MethodRunner();
```
tdurieux: I do not need a asynchronous call but a need a thread the change the classloader.
And this is the only way that I know to wait the end of the execution of a thread.

monperrus: Without thread: new URLClassLoader(urls).loadClass(...).newInstance()

tdurieux: Done

src/main/java/spoon/support/compiler/jdt/CompilationUnitWrapper.java
```diff
@@ -43,30 +44,14 @@
 			return jdtCompiler.loadedContent.get(s);
```
monperrus: does this cache mean that one cannot modify - compile and then modify again and compile again?

tdurieux: No, it does not.
The cache is cleared before each compilation (see JDTBatchCompiler constructor)

monperrus:
any idea about the exception in the build log: https://travis-ci.org/INRIA/spoon/builds/165035633#L3829?

tdurieux:
The cache was not correctly handled

PR #880, opened 2016-10-04
Added a method to know whether a given method is part of a CtType

arnobl:
Hey all,
As suggested on the mailing list, here is a PR to add a method in CtType to know whether a given type has a method (inherited or not, abstract or implemented). Look at the Javadoc for more details. Let me know if you want me to change the behaviour of the added method (replacing PR 870).

### DIFF
--- a/src/main/java/spoon/reflect/declaration/CtType.java
+++ b/src/main/java/spoon/reflect/declaration/CtType.java
@@ -216,6 +216,14 @@ public interface CtType<T> extends CtNamedElement, CtTypeInformation, CtTypeMemb
 	List<CtMethod<?>> getMethodsByName(String name);
 
 	/**
+	 * Searches in the type for the given method.
+	 * Super classes and implemented interfaces are considered.
+	 * @param method The method to search for in the class.
+	 * @return True: the type has the given method. False otherwise.
+	 */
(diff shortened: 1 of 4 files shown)


monperrus:
This method is quite handy to know whether one can call a method on typed variable.

I wonder if using a signature-based test would be better than an object-based test.

Ie replacing `method.getParent() == this` by
```
final String sig = method.getSignature();
for(CtMethod<?> m : getMethods()) {
	if(m.getSignature().equals(sig)) {
		return true;
	}
}
```
Obviously, this behavior supersedes the current one.

What do you think?

arnobl:
Yes, the object-based test is not mandatory but just an optimisation. I can remove it. Tell me.

monperrus:
let's go for a signature-based test then

src/main/java/spoon/reflect/declaration/CtType.java
```diff
@@ -216,6 +216,14 @@
 	List<CtMethod<?>> getMethodsByName(String name);
 
 	/**
+	 * Searches in the type for the given method.
+	 * Super classes and implemented interfaces are considered.
```
monperrus: The matching criterion is that the signatures are identical.

monperrus:
thanks a lot for this contribution

PR #883, opened 2016-10-05
fix(if): prints the else body when the body is not a block

tdurieux:


monperrus:
a test case with that?

tdurieux:
@monperrus Done

PR #884, opened 2016-10-05
feature(runtime): allow to instantiate any class at any time

monperrus:
```java
		// contract: a ctclass can be instantiated, and each modification results in a new valid object
		CtClass c = factory.Code().createCodeSnippetStatement(
							"class X implements spoon.test.compilation.Ifoo { public int foo() {int i=0; return i;} }").compile();
		c.addModifier(ModifierKind.PUBLIC); // required otherwise java.lang.IllegalAccessException at runtime when instantiating

		CtBlock body = c.getElements(new TypeFilter<>(CtBlock.class)).get(1);
		Object o = c.newInstance();
		assertEquals(0, ((Ifoo)o).foo());
		for (int i=1; i<=10; i++) {
			body.getStatement(0).replace(factory.Code().createCodeSnippetStatement("int i = "+i+";"));
			o = c.newInstance();
			// each time the behavior has changed!
			assertEquals(i, ((Ifoo)o).foo());
		}

```

### DIFF
--- a/src/main/java/spoon/SpoonModelBuilder.java
+++ b/src/main/java/spoon/SpoonModelBuilder.java
@@ -76,21 +76,29 @@ public interface SpoonModelBuilder {
 	 */
 	boolean build(JDTBuilder builder);
 
+	/** The types of compilable elements */
+	enum InputType {
+		FILES, CTTYPES
+	}
+
+
 	/**
 	 * Generates the bytecode associated to the classes stored in this
 	 * compiler's factory. The bytecode is generated in the directory given by
 	 * {@link #getBinaryOutputDirectory()}.
 	 *
-	 * @see #getSourceClasspath()
-	 */
-	boolean compile();
-
-	/**
-	 * Generates the bytecode by compiling the input sources. The bytecode is
-	 * generated in the directory given by {@link #getBinaryOutputDirectory()}.
+	 * The array of types must be of size 0 or 1. If it's empty,
+	 * the types of the factory are compiled.
+	 * If it's InputType.FILES, the files given as input are compiled.
+	 *
+	 *Note that the varargs ... enables this version to be backward compatible for callers.
 	 *
 	 * @see #getSourceClasspath()
 	 */
+	boolean compile(InputType... types);
+
+	/** replaced by {@link #compile(InputType...)} */
+	@Deprecated
 	boolean compileInputSources();
 
 	/**
@@ -233,10 +241,6 @@ public interface SpoonModelBuilder {
 
 	/**
 	 * Gets the classpath that is used to build/compile the input sources.
-	 *
-	 * @see #compileInputSources()
-	 * @see #build()
-	 * @see #compile()
 	 */
 	String[] getSourceClasspath();
 
--- a/src/main/java/spoon/support/compiler/SnippetCompilationHelper.java
+++ b/src/main/java/spoon/support/compiler/SnippetCompilationHelper.java
@@ -81,6 +81,12 @@ public class SnippetCompilationHelper {
 		// Clean up
 		c.getPackage().getTypes().remove(c);
 
+		if (ret instanceof CtClass) {
+			CtClass klass = (CtClass) ret;
+			klass.setSimpleName(klass.getSimpleName().replaceAll("^[0-9]*", ""));
+			klass.setParent(ret.getFactory().Package().getRootPackage());
+			ret.getFactory().Package().getRootPackage().addType(klass);
+		}
 		return ret;
 	}
 
--- a/src/main/java/spoon/support/compiler/jdt/CompilationUnitWrapper.java
+++ b/src/main/java/spoon/support/compiler/jdt/CompilationUnitWrapper.java
@@ -17,6 +17,7 @@
 package spoon.support.compiler.jdt;
 
 import org.eclipse.jdt.internal.compiler.batch.CompilationUnit;
+
 import spoon.reflect.declaration.CtType;
 import spoon.reflect.visitor.DefaultJavaPrettyPrinter;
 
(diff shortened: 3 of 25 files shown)


src/main/java/spoon/support/compiler/jdt/JDTBatchCompiler.java
```diff
@@ -70,10 +71,15 @@ public ICompilerRequestor getBatchRequestor() {
 			public void acceptResult(CompilationResult compilationResult) {
 				if (compilationResult.hasErrors()) {
 					for (CategorizedProblem problem:compilationResult.problems) {
-						JDTBatchCompiler.this.jdtCompiler.reportProblem(problem);
+						if (JDTBatchCompiler.this.jdtCompiler != null) {
+							JDTBatchCompiler.this.jdtCompiler.reportProblem(problem);
+						} else {
+							new Exception().printStackTrace();
+							System.err.println(problem + " " + new String(problem.getOriginatingFileName()));
```
tdurieux: We have a logger please use it.

src/main/java/spoon/support/reflect/code/CtStatementImpl.java
```diff
@@ -67,10 +67,14 @@ public static void insertBefore(CtStatement target, CtStatementList statementsTo
...
-			if (target instanceof CtInvocation && ((CtInvocation<?>) target).getExecutable().getSimpleName().startsWith("<init>")) {
-				throw new SpoonException("cannot insert a statement before a super or this invocation.");
+		try {
+			if (target.getParent(CtConstructor.class) != null) {
+				if (target instanceof CtInvocation && ((CtInvocation<?>) target).getExecutable().getSimpleName().startsWith("<init>")) {
+					throw new SpoonException("cannot insert a statement before a super or this invocation.");
+				}
 			}
+		} catch (ParentNotInitializedException ignore) {
+			// no constructor in parent
```
tdurieux: You have this exception only because you set the parent of the CtClass to null when you use the snippet to create the class.

src/main/java/spoon/support/compiler/SnippetCompilationHelper.java
```diff
@@ -81,6 +81,11 @@ private static CtStatement internalCompileStatement(CtElement st) {
 		// Clean up
 		c.getPackage().getTypes().remove(c);
 
+		if (ret instanceof CtClass) {
+			CtClass klass = (CtClass) ret;
+			klass.setSimpleName(klass.getSimpleName().replaceAll("^[0-9]*", ""));
+			klass.setParent(null);
```
tdurieux: Use the rootPackage instead of null.

src/main/java/spoon/support/reflect/declaration/CtClassImpl.java
```diff
@@ -240,4 +254,35 @@ public void replace(CtStatement element) {
...
+					.annotationProcessingOptions(new AnnotationProcessingOptions().compileProcessors()) //
+					.sources(new SourceOptions().sources()) //
+					.build();
+			compiler.compile(args);
+			Class<?> klass = new URLClassLoader(new URL[] { new File("./").toURL() }).loadClass(getQualifiedName());
+
+			// cleaning afterwards
+			new File(getFactory().getEnvironment().getBinaryOutputDirectory() + "/" + getQualifiedName().replace('.', '/') + ".class").delete();
+
+			return klass.newInstance();
```
tdurieux: I think it is a good idea to force the compilation the complete project because if the user did not set shouldCompile to true, all reference to other classes of the project will produce compilation error.
Same error if the klass contains a reference to a new class created with spoon.

src/main/java/spoon/SpoonModelBuilder.java
```diff
@@ -76,22 +76,24 @@
...
+
+
 	/**
 	 * Generates the bytecode associated to the classes stored in this
 	 * compiler's factory. The bytecode is generated in the directory given by
 	 * {@link #getBinaryOutputDirectory()}.
 	 *
-	 * @see #getSourceClasspath()
-	 */
-	boolean compile();
```
GerardPaligot: `SpoonModelBuilder` is an interface (and very exposed). Think to deprecate methods rather than remove them like this.

src/main/java/spoon/SpoonModelBuilder.java
```diff
@@ -76,22 +76,24 @@
...
-	/**
-	 * Generates the bytecode by compiling the input sources. The bytecode is
-	 * generated in the directory given by {@link #getBinaryOutputDirectory()}.
+	 * The array of types must be of size 0 or 1. If it's empty,
+	 * the types of the factory are compiled.
+	 * If it's InputType.FILES, the files given as input are compiled.
 	 *
 	 * @see #getSourceClasspath()
 	 */
-	boolean compileInputSources();
```
GerardPaligot: Same here.

src/main/java/spoon/support/compiler/jdt/CompilationUnitWrapper.java
```diff
@@ -17,41 +17,38 @@
...
 import org.eclipse.jdt.internal.compiler.batch.CompilationUnit;
+
 import spoon.reflect.declaration.CtType;
 import spoon.reflect.visitor.DefaultJavaPrettyPrinter;
 
 import java.util.ArrayList;
 import java.util.List;
 
-class CompilationUnitWrapper extends CompilationUnit {
+public class CompilationUnitWrapper extends CompilationUnit {
```
GerardPaligot: This is really necessary to expose this class?

src/main/java/spoon/support/reflect/code/CtStatementImpl.java
```diff
@@ -67,10 +67,14 @@ public static void insertBefore(CtStatement target, CtStatementList statementsTo
...
-			if (target instanceof CtInvocation && ((CtInvocation<?>) target).getExecutable().getSimpleName().startsWith("<init>")) {
-				throw new SpoonException("cannot insert a statement before a super or this invocation.");
+		try {
+			if (target.getParent(CtConstructor.class) != null) {
+				if (target instanceof CtInvocation && ((CtInvocation<?>) target).getExecutable().getSimpleName().startsWith("<init>")) {
+					throw new SpoonException("cannot insert a statement before a super or this invocation.");
+				}
 			}
+		} catch (ParentNotInitializedException ignore) {
+			// no parent set somewhere
```
GerardPaligot: That smells bad...

PR #887, opened 2016-10-07
Fixes potential NullPointerException in JDTTreeBuilderQuery.

retuxx:
Unfortunately, I can't give you an example as I didn't manage to create a minimal version exploiting this error. However, it should be obvious that `currentPackage` may be null.


monperrus:
thanks.

PR #891, opened 2016-10-12
fix(literal): fix literal type

tdurieux:


monperrus:
see also
http://stackoverflow.com/questions/12206819/what-are-the-java-primitive-data-type-modifiers

PR #909, opened 2016-10-28
fix(newinstance): forces the load of the new version of the class

tdurieux:


monperrus:
Following our discussion, see https://github.com/tdurieux/spoon/pull/3

Issue #911, opened 2016-10-28
bug in printing of new Array with comment


pvojtechovsky:
The code like
``
String[] AN_ARRAY = new String[] {"a","b"/*, "c"*/};
``
is printed by DefaultJavaPrettyPrinter.visitCtNewArray as
``
String[] AN_ARRAY = new String[] {"a" , "b" , /* , "c" * };
``
bug: There is invalid comma after last array item
bug: The last character "/" of comment is deleted.

PR #912, opened 2016-10-28
fixes bug in printing of new Array with comment #911

pvojtechovsky:
In case of comment the last character is the last character of the
comment and not the ",".


monperrus:
Hi Pavel,
Thanks for the contribution. Would you add a test case to avoid a regression in the future?
Thanks, --Martin

pvojtechovsky:
Hi Martin,
The test was added and it comply with formatting rules now too. 
Best regards,
Pavel

monperrus:
now it breaks a test, could you make it pass? thanks.

pvojtechovsky:
I modified the fix by way it keeps formatting of annotations too.
Now all seems to be ok. I just actually do not know how to squash all commits into one, without creating new branch and loosing this PR.... but I will hopefully found it soon...

monperrus:
excellent, thanks. We can squash at merge time.

PR #915, opened 2016-10-28
improves performance of VirtualFolder.getAllFiles()

pvojtechovsky:
Contains
1. test which measures performance of VirtualFolder.getAllFiles() 
2. test which measures performance of VirtualFolder.getSubFolders() 
3. code which improves that performance.

The performance improvement consists of these changes:
1. FileSystemFile calls slow File.getCanonicalFile() only once in constructor. It is not called later. I hope it breaks nothing. Please check it.
2. FileSystemFolder calls slow File.getCanonicalFile() only once in constructor. It is not called later. I hope it breaks nothing. Please check it.
3.  VirtualFolder getAllFiles() and getSubFolders() uses helper HashSet to collect distinct set of files

Note: Before the test needed 80s on 900 files. After performance fix it needs 0.2s
Note: The usage of helper HashSet would be enough to fix performance of VirtualFolder getAllFiles() and getSubFolders(), but FileSystemFile.toFile() and FileSystemFolder.toFile() and getPath() is probably called on other places in Spoon too, therefore I suggest to take both changes.

### DIFF
--- a/src/main/java/spoon/support/compiler/FileSystemFile.java
+++ b/src/main/java/spoon/support/compiler/FileSystemFile.java
@@ -35,7 +35,11 @@ public class FileSystemFile implements SpoonFile {
 
 	public FileSystemFile(File file) {
 		super();
-		this.file = file;
+		try {
+			this.file = file.getCanonicalFile();
+		} catch (IOException e) {
+			throw new SpoonException(e);
+		}
 	}
 
 	public InputStream getContent() {
@@ -94,11 +98,7 @@ public class FileSystemFile implements SpoonFile {
 
 	@Override
 	public File toFile() {
-		try {
-		return file.getCanonicalFile();
-		} catch (IOException e) {
-			throw new SpoonException(e);
-		}
+		return file;
 	}
 
 	@Override
--- a/src/main/java/spoon/support/compiler/FileSystemFolder.java
+++ b/src/main/java/spoon/support/compiler/FileSystemFolder.java
@@ -18,7 +18,6 @@ package spoon.support.compiler;
 
 import java.io.File;
 import java.io.FileNotFoundException;
-import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 
@@ -37,7 +36,11 @@ public class FileSystemFolder implements SpoonFolder {
 		if (!file.isDirectory()) {
 			throw new SpoonException("Not a directory " + file);
 		}
-		this.file = file;
+		try {
+			this.file = file.getCanonicalFile();
+		} catch (Exception e) {
+			throw new SpoonException(e);
+		}
 	}
 
 	public FileSystemFolder(String path) {
@@ -114,13 +117,8 @@ public class FileSystemFolder implements SpoonFolder {
 	}
 
 	public String getPath() {
-		try {
-			return file.getCanonicalPath();
-		} catch (Exception e) {
-			Launcher.LOGGER.error(e.getMessage(), e);
 			return file.getPath();
 		}
-	}
 
 	@Override
 	public boolean isArchive() {
@@ -134,11 +132,7 @@ public class FileSystemFolder implements SpoonFolder {
 
 	@Override
 	public File toFile() {
-		try {
-			return file.getCanonicalFile();
-		} catch (IOException e) {
-			throw new SpoonException(e);
-		}
+		return file;
 	}
 
 	@Override


src/main/java/spoon/support/compiler/VirtualFolder.java
```diff
@@ -46,10 +46,10 @@ public void addFolder(SpoonFolder o) {
 	@Override
 	public List<SpoonFile> getAllFiles() {
 		List<SpoonFile> result = new ArrayList<>();
-
+		Set<SpoonFile> distinctSet = new HashSet<>();
 		for (SpoonFile f : getFiles()) {
 			// we take care not to add a file that was already found in a folder
-			if (!result.contains(f)) {
```
monperrus: are your sure hashCode/equals of SpoonFile are correct? I'm not sure they are well tested.

src/main/java/spoon/support/compiler/VirtualFolder.java
```diff
@@ -87,9 +87,10 @@ public SpoonFolder getParent() {
 	@Override
 	public List<SpoonFolder> getSubFolders() {
 		List<SpoonFolder> result = new ArrayList<>();
+		Set<SpoonFolder> distinctSet = new HashSet<>();
 		for (SpoonFile f : getAllFiles()) {
 			SpoonFolder folder = f.getParent();
-			if (!result.contains(folder)) {
```
monperrus: are your sure hashCode/equals of SpoonFolder are correct? I'm not sure they are well tested.

src/test/java/spoon/test/support/ResourceTest.java
```diff
@@ -90,4 +90,31 @@ public void testFilteringFolder() throws Exception {
...
+		String dir = "src";
+		FileSystemFolder fileSystemFolder = new FileSystemFolder(new File(dir));
+
+		SpoonFolder folder = new VirtualFolder();
+		folder.addFolder(fileSystemFolder);
+
+		//Measure how long it needs to return all files
+		long startTime = System.currentTimeMillis();
+		folder.getAllFiles();
+		assertTrue("The VirtualFolder.getAllFiles() is too slow", (System.currentTimeMillis() - startTime)<1000);
```
monperrus: In Spoon, we try not to make assertions based on currentTimeMillis, they are too dependent on the platform. To me, you can remove the tests. Everybody agrees?

pvojtechovsky: Do you check the times needed by execution of each test? I do not know your testing environment, but some tools reports a warning/error even when test passes, but needs much longer time then previous run of the same test. If you have such execution time comparator, then it would make sense to keep the test there - I would just remove assert which checks absolute time.

src/main/java/spoon/support/compiler/VirtualFolder.java
pvojtechovsky: The VirtualFile and ZipFile really miss hashCode, so they would not work correctly with the HashSet. Do you agree to add hashCode method to them?

src/main/java/spoon/support/compiler/VirtualFolder.java
pvojtechovsky: The VirtualFolder really miss hashCode, so it would not work correctly with the HashSet. Do you agree to add hashCode method to it?

src/main/java/spoon/support/compiler/VirtualFolder.java
monperrus: I would prefer not to have equals/hashcode in those classes, in order to not to maintain them in the future. And consequently, it seems betetr to avoid a hashset here.

pvojtechovsky: The not using HashSet means a performance problems with bigger number of files. The MS Windows file.equals() method is much faster then File.getCannonicalFile, but it is still quite slow. So recursive calling of it in List.contains needs time too. I will measure it tomorrow and give you exact number.

pvojtechovsky:
I have measured the performance of solution with and without hashSet and the results are similar. So I have rolled back changes in VirtualFolder so the hashSet is not used, so the hashCode methods does not have to be added.
The remaining changes makes sense. They really improves performance 60s->0.2s on 900 files on MS Windows 10. So it makes sense to accept it.
I have removed platform specific test, which failed on bad performance.
I have squashed all commits into one. 
Is there anything else needed to accept this PR?

monperrus:
What if we simply remove all usages of getCanonicalFile? Do we get a performance improvement as well?

pvojtechovsky:
Yes, it improves performance too, but then two files, which are same will may be return false on File.equals() when they are using different path ... 
/root/aDir/../backInRoot/x.txt
/root/backInRoot/x.txt

monperrus:
we can keep the canonical version. However, I'd remove the isCanonical boolean, if not canonicalizable, we throw a SpoonException, for sake of least astonishment.

PR #917, opened 2016-10-28
improves performance of FindCommentParentScanner

pvojtechovsky:
The collecting of comments seems to be slow in method 

``` java
JDTCommentBuilder$FindCommentParentScanner.scan(CtElement) 
...
element.getClass().getMethod("getBody")
```

Reason: Java reflection is slow. Throwing exception is slow. 

Solution: Cache the getBody method of classes in HashMap
I do not know how to make a test for that. But I see the visible improvement in building time of the model.

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/JDTCommentBuilder.java
+++ b/src/main/java/spoon/support/compiler/jdt/JDTCommentBuilder.java
@@ -23,19 +23,23 @@ import org.eclipse.jdt.internal.compiler.env.ICompilationUnit;
 import spoon.reflect.code.CtBinaryOperator;
 import spoon.reflect.code.CtBlock;
 import spoon.reflect.code.CtCase;
+import spoon.reflect.code.CtCatch;
 import spoon.reflect.code.CtComment;
 import spoon.reflect.code.CtConditional;
 import spoon.reflect.code.CtIf;
+import spoon.reflect.code.CtLoop;
 import spoon.reflect.code.CtNewArray;
 import spoon.reflect.code.CtStatement;
 import spoon.reflect.code.CtStatementList;
 import spoon.reflect.code.CtSwitch;
+import spoon.reflect.code.CtTry;
 import spoon.reflect.cu.CompilationUnit;
 import spoon.reflect.cu.SourcePosition;
 import spoon.reflect.declaration.CtAnonymousExecutable;
 import spoon.reflect.declaration.CtClass;
 import spoon.reflect.declaration.CtConstructor;
 import spoon.reflect.declaration.CtElement;
+import spoon.reflect.declaration.CtExecutable;
 import spoon.reflect.declaration.CtField;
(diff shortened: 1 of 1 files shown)


monperrus:
Did your measure the performance improvement?

monperrus:
What about calling directly getBody() with no reflection: eg `((CtExecutable)element).getBody()`? 
@tdurieux what do you think?

pvojtechovsky:
I measured (5 times) the performance of building model with comments ON.
Origin code needed: 117621, 112918 [ms]
Improved code needs: 51140, 49180, 56217 [ms]
So it really helps!

tdurieux:
> What about calling directly getBody() with no reflection: eg ((CtExecutable)element).getBody()? 
It's not possible because getBody is exposed in several different interface (CtExecutable, Ctoop, CtIf, CtTry, ...).
We also can use CtInheritanceScanner.

monperrus:
@tdurieux what do you think of this caching solution?

tdurieux:
I think it will be better to do something like this.
No need of reflection...
```java
if (e instanceof CtLoop) {
	return ((CtLoop) e).getBody();
} else if (e instanceof CtExecutable) {
	return ((CtExecutable) e).getBody();
} else if (e instanceof CtTry) {
	return ((CtTry) e).getBody();
} else if (e instanceof CtCatch) {
	return ((CtCatch) e).getBody();
}
return null;
```

pvojtechovsky:
Now the code looks nice. I have added new method 
static CtElement getBody(CtElement e) {...}
into JDTCommentBuilder
Idea: What about moving this method into some helper class so others can use it too.

monperrus:
> What about moving this method into some helper class so others can use it too.

good idea, but I don't see a class where this method would be easily discoverable. what do you propose?

pvojtechovsky:
I do not know spoon enough. No idea yet.

pvojtechovsky:
What about adding new interface with CtStatement getBody() method and let CtLoop, CtExecutable, CtTry and CtCatch extend it?
I just do not know how to name this interface
- CtStatementContainer
- CtBodyHolder
- CtStatementHolder
...

pvojtechovsky:
I vote for moving of the idea to have getBody method available for others into another PR. 
It is not related to origin performance problem... so you can merge my PR in.

monperrus:
ok for creating `CtBodyHolder`. Could you do it in this PR?

pvojtechovsky:
I failed to do it in simple way and I am not sure if the solution I have found is correct. So I suggest to merge this PR as it is. Then I will make another PR where we can discuss if the solution is acceptable or not.

The problem is that interface `CtExecutable` has method
```java
<B extends R> CtBlock<B> getBody();
```
which cannot be compiled with any inherited `getBody` method. The only solution I have found is to rollback the declaration of getBody method to 
```java
CtBlock<R> getBody();
```
which is compilable ... after some other changes in all implementations of the `CtExecutable`

PR #918, opened 2016-10-28
improves performance of CtElementImpl.equals()

pvojtechovsky:
This method is called very often so it makes sense to add fast checks
for null and check that classes are equal, otherwise the ClassCastException is thrown and caught internally and it is slow.

I do not know how to make a test for that, but during build of model with 6000 files it spent 43 seconds in CtElementImpl.equals(), while after fix it measured 0s in java VisualVM sampler

### DIFF
--- a/src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
+++ b/src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
@@ -109,6 +109,12 @@ public abstract class CtElementImpl implements CtElement, Serializable {
 		if (this == o) {
 			return true;
 		}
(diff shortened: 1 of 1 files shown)


monperrus:
Did your measure the performance improvement?

pvojtechovsky:
As I mentioned above, building of my source model needs 43 seconds less with the improvement. What measure would you like?

pvojtechovsky:
I did another performance measurements and now I see no difference. Please do not accept this PR until I found what is happening there... I am sorry.

pvojtechovsky:
I do not know what I have seen before ... I am closing this PR

PR #919, opened 2016-10-30
fix LauncherTest, which fails on MS Windows

pvojtechovsky:
MS Windows uses backslash in file system path therefore this:
assertTrue(inputSources.get(0).getPath().contains("src/main/java"));
always fails on MS Windows


monperrus:
thanks!

PR #924, opened 2016-11-02
simplify code of JDTCommentBuilder - same behavior

pvojtechovsky:
Hi Martin,
do you like PR like this too? It just simplifies the code. No change in behavior.
Best regards
Pavel


tdurieux:
I do not remember why I wrote such ugly code :'(

pvojtechovsky:
May be it was a deep night ... like today. ;-)))
I have no problems with comments feature. It seems to be working. It is hard to say whether some comments are lost somewhere, I have not checked it yet. Till now I had performance problems, so I tuned them little bit. And current problem, I have, is the duplicate annotations - probably because sources are compiled and added to model twice. Generally I like spoon a lot. I hope it will help me to clean some mess in my project.

tdurieux:
Thanks @pvojtechovsky

PR #926, opened 2016-11-03
fix: filter out some JDT errors on noclasspath

### LINKED ISSUE #788: Spoon Error logs are printed when called from a gradle plugin.
When using a spoon project with logLevel to _OFF_ directly from java, no logs are shown, as expected.
But when this project is integrated in a gradle plugin, error logs are shown during the build process.

This project environment uses `noclasspath`, thus those errors are not relevant.

```
$ ./gradlew compileDebugJava
Parallel execution is an incubating feature.
Incremental java compilation is an incubating feature.
:dartJavaTask
:dartApiTask
error: Dependency cannot be resolved to a type at /project/build/generated/source/generation/debug/package/GenClass.java:12
[...] // Similar errors
######################################################
Spoon logger level: OFF
Gradle logger level (debug?): false // Spoon debug log aren't shown even if gradle log is to debug
Gradle logger level (error?): true

[...]

BUILD SUCCESSFUL
```

aveuiller:
Errors of type `UndefinedType`, `UndefinedName` and `ImportNotFound` are irrelevant for noclasspath mode, they are not logged anymore.

Closes #788

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/JDTBasedSpoonCompiler.java
+++ b/src/main/java/spoon/support/compiler/jdt/JDTBasedSpoonCompiler.java
@@ -621,7 +621,12 @@ public class JDTBasedSpoonCompiler implements SpoonCompiler {
 				throw new ModelBuildingException(message);
 			} else {
 				// in noclasspath mode, errors are only reported
-				environment.report(null, problem.isError() ? Level.ERROR : Level.WARN, message);
+				// but undefined import, type, and name errors are irrelevant
+				int problemId = problem.getID();
+				if (problemId != IProblem.UndefinedType && problemId != IProblem.UndefinedName
+						&& problemId != IProblem.ImportNotFound) {
+					environment.report(null, Level.WARN, message);
+				}
 			}
 		}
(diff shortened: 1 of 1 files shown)


src/main/java/spoon/support/compiler/jdt/JDTBasedSpoonCompiler.java
```diff
@@ -621,7 +621,12 @@ protected void report(Environment environment, CategorizedProblem problem) {
 				throw new ModelBuildingException(message);
 			} else {
 				// in noclasspath mode, errors are only reported
-				environment.report(null, problem.isError() ? Level.ERROR : Level.WARN, message);
+				// but undefined import, type, and name errors are irrelevant
+				int problemId = problem.getID();
+				if (problemId != IProblem.UndefinedType && problemId != IProblem.UndefinedName
+						&& problemId != IProblem.ImportNotFound) {
+					environment.report(null, Level.ERROR, message);
```
monperrus: what about only using Level.WARN in noclasspath mode?

PR #927, opened 2016-11-03
fixed(prettyprint): parentheses are missing when pretty-pretting some ternary expressions

arnobl:
The current prettyprint of ternary expressions (CtConditional) adds parentheses around the condition only when the parent is an assignment. The added test show that parentheses are also required on variable declaration.

The test shows that the mandatory parentheses that originally surround the condition are removed when pretty-printing the spooned code. This leads to ambiguous code that does not compile.
The fix adds the parentheses when the parent is a variable declaration.

Maybe the parentheses should be always added.

### DIFF
--- a/src/test/java/spoon/test/prettyprinter/DefaultPrettyPrinterTest.java
+++ b/src/test/java/spoon/test/prettyprinter/DefaultPrettyPrinterTest.java
@@ -288,16 +288,14 @@ public class DefaultPrettyPrinterTest {
 	}
 
 	@Test
-	public void testTernaryParenthesesOnLocalVariable() throws SnippetCompilationError {
+	public void testTernaryParenthesesOnLocalVariable() {
 		// Spooning the code snippet
 		Launcher launcher = new Launcher();
 		CtCodeSnippetStatement snippet = launcher.getFactory().Code().createCodeSnippetStatement(
 			"final int foo = (new Object() instanceof Object ? new Object().equals(null) : new Object().equals(new Object())) ? 0 : new Object().hashCode();");
 		CtStatement compile = snippet.compile();
-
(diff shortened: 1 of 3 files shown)


src/test/java/spoon/test/prettyprinter/DefaultPrettyPrinterTest.java
```diff
@@ -285,4 +286,19 @@ public void importsFromMultipleTypesSupported() {
...
+	@Test
+	public void testTernaryParenthesesOnLocalVariable() throws SnippetCompilationError {
+		// Spooning the code snippet
+		Launcher launcher = new Launcher();
+		CtCodeSnippetStatement snippet = launcher.getFactory().Code().createCodeSnippetStatement(
+			"final int foo = (new Object() instanceof Object ? new Object().equals(null) : new Object().equals(new Object())) ? 0 : new Object().hashCode();");
+		CtStatement compile = snippet.compile();
+
+		// Pretty-printing the Spooned code snippet and compiling the resulting code.
+		launcher = new Launcher();
```
monperrus: do you need a new Launcher?

src/test/java/spoon/test/prettyprinter/DefaultPrettyPrinterTest.java
```diff
@@ -285,4 +286,19 @@ public void importsFromMultipleTypesSupported() {
...
+		snippet = launcher.getFactory().Code().createCodeSnippetStatement(compile.toString());
+		// If it throws an exception, the printed code is not valid.
+		snippet.compile();
```
monperrus: add an assert `assertEquals(compile, snippet.compile())`?

monperrus:
thanks!

PR #928, opened 2016-11-03
fix bug: NPE when compilation units array contains null elements

sjd78:
Fix a bug in TreeBuilderCompiler.buildUnits() where a retured array is
longer then expected and the extra spaces are null. The method is updated
to use the _totalUnits_ instance variable as the true element count in
the _unitsToProcess_ array instead of the length of the array itself.

This broke processing of classes using JDTBatchCompiler.getUnits() and
expecting every element in the array to be non-null.  The problem came up
when trying to process the source for a large GWT project.

This applies the fix for issue #550 to master.

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/TreeBuilderCompiler.java
+++ b/src/main/java/spoon/support/compiler/jdt/TreeBuilderCompiler.java
@@ -17,6 +17,7 @@
 package spoon.support.compiler.jdt;
 
 import java.io.PrintWriter;
+import java.util.ArrayList;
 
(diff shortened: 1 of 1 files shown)


monperrus:
I see that #550  was merged in stable. And this was never ported to master?

sjd78:
Correct.  I happened to come back to my project using spoon and hit the error again. Reapplied the PR and things started working again.

I just checked the code in the current stable branch and I don't see the change their either.

monperrus:
sorry for that. we have a bug in the process to be fixed.

sjd78:
No problem.  Thanks for the quick merge!

Issue #921, opened 2016-10-30
Duplicity in package-info.java when templates are used


pvojtechovsky:
The package-info.java with this content

``` java
/**
 * This is test
 * JavaDoc.
 */
@Deprecated
package spoon.test.pkg;
```

prints this not compilable output

``` java
/**
 * This is test
 * JavaDoc.
 *//**
 * This is test
 * JavaDoc.
 */@java.lang.Deprecated
@java.lang.Deprecated
package spoon.test.pkg;
```

when templates are added to the Launcher

See this [branch with test](https://github.com/pvojtechovsky/spoon/tree/duplicitiesInPackageInfo) to reproduce the problem.

What way do you prefer to contribute new test? 
A) branch like this?
B) branch like this + PR?

monperrus:
If you have a PR with a fix, it's always better :-)

pvojtechovsky:
It looks like FileCompiler.getCompilationUnits() always compiles sources and templates
```java
	public CompilationUnit[] getCompilationUnits() {
		List<SpoonFile> files = new ArrayList<>();
		files.addAll(jdtCompiler.sources.getAllJavaFiles());
		files.addAll(jdtCompiler.templates.getAllJavaFiles());
```
but may be it should compile only the required java *files* - these which are sent as input to 
```java
JDTBatchCompiler.getUnits(List<SpoonFile> files)
```
I mean stacktrace like this
```
	at spoon.support.compiler.jdt.FileCompiler.getCompilationUnits(FileCompiler.java:54)
	at spoon.support.compiler.jdt.JDTBatchCompiler.getUnits(JDTBatchCompiler.java:117)
	at spoon.support.compiler.jdt.FileCompiler.getUnits(FileCompiler.java:1)
	at spoon.support.compiler.jdt.JDTBasedSpoonCompiler.****buildSources OR buildTemplates****(JDTBasedSpoonCompiler.java:380)
	at spoon.support.compiler.jdt.JDTBasedSpoonCompiler.build(JDTBasedSpoonCompiler.java:116)
	at spoon.support.compiler.jdt.JDTBasedSpoonCompiler.build(JDTBasedSpoonCompiler.java:99)
	at spoon.Launcher.buildModel(Launcher.java:712)
	at spoon.test.pkg.PackageTest.testAnnotationInPackageInfoWhenTemplatesCompiled(PackageTest.java:120)
```

monperrus:
[...]

PR #929, opened 2016-11-03
Concept for compiler refactoring and bug fixing

pvojtechovsky:
Hi Martin, 

thanks for the answer in #921. It really helped me to proceed. The new compilation works for me. We will see whether other tests will pass and I am mainly looking forward to your feedback, I hope I made no fatal mistake in my understanding and it was not wasting of my time :-)

Please check mainly new method `JDTBasedSpoonCompiler.buildUnits()` and my implementation of `JDT bug HACK`

The code is not style-checked yet, I will do it at the end before you agrees that it is on the good way.

I like fast and positive reactions of contributors of Spoon team. Thanks for Your time!

### DIFF
--- a/src/main/java/spoon/compiler/builder/ClasspathOptions.java
+++ b/src/main/java/spoon/compiler/builder/ClasspathOptions.java
@@ -55,7 +55,7 @@ public class ClasspathOptions<T extends ClasspathOptions<T>> extends Options<T> 
 		if (currentClassLoader instanceof URLClassLoader) {
 			final URL[] urls = ((URLClassLoader) currentClassLoader).getURLs();
 			if (urls != null && urls.length > 0) {
-				List<String> classpaths = new ArrayList<>(urls.length+1);
+				List<String> classpaths = new ArrayList<>(urls.length + 1);
 				classpaths.add(".");
 				for (URL url : urls) {
 					classpaths.add(url.getFile());
--- a/src/main/java/spoon/compiler/builder/ClasspathOptions.java
+++ b/src/main/java/spoon/compiler/builder/ClasspathOptions.java
@@ -19,6 +19,8 @@ package spoon.compiler.builder;
 import java.io.File;
 import java.net.URL;
 import java.net.URLClassLoader;
+import java.util.ArrayList;
+import java.util.List;
 
 public class ClasspathOptions<T extends ClasspathOptions<T>> extends Options<T> {
 	public ClasspathOptions() {
@@ -42,22 +44,26 @@ public class ClasspathOptions<T extends ClasspathOptions<T>> extends Options<T> 
 	}
 
 	public T classpathFromListOrClassLoader(String... classpaths) {
-		return (classpaths != null && classpaths.length > 0) ? classpath(classpaths) : classpathFromCurrentClassLoader();
+		return classpath(getClasspathFromListOrCurrentClassloader(classpaths));
 	}
 
-	public T classpathFromCurrentClassLoader() {
+	public static String[] getClasspathFromListOrCurrentClassloader(String... classpaths) {
+		return (classpaths != null && classpaths.length > 0) ? classpaths : getClasspathFromCurrentClassloader();
+	}
+	public static String[] getClasspathFromCurrentClassloader() {
 		ClassLoader currentClassLoader = Thread.currentThread().getContextClassLoader();
 		if (currentClassLoader instanceof URLClassLoader) {
 			final URL[] urls = ((URLClassLoader) currentClassLoader).getURLs();
 			if (urls != null && urls.length > 0) {
-				String classpath = ".";
+				List<String> classpaths = new ArrayList<>(urls.length+1);
+				classpaths.add(".");
 				for (URL url : urls) {
-					classpath += File.pathSeparator + url.getFile();
+					classpaths.add(url.getFile());
 				}
-				classpath(classpath);
+				return classpaths.toArray(new String[classpaths.size()]);
 			}
 		}
-		return myself;
+		return null;
 	}
 
 	public T bootclasspath(String bootclasspath) {
--- a/src/main/java/spoon/support/compiler/jdt/FileCompiler.java
+++ b/src/main/java/spoon/support/compiler/jdt/FileCompiler.java
@@ -27,7 +27,7 @@ import spoon.compiler.SpoonFile;
 import spoon.compiler.SpoonFolder;
 
 public class FileCompiler extends JDTBatchCompiler {
-	
+
 	protected List<SpoonFile> inputFiles;
 
 	public FileCompiler(JDTBasedSpoonCompiler jdtCompiler) {
@@ -44,10 +44,10 @@ public class FileCompiler extends JDTBatchCompiler {
 	 */
 	@Override
 	public CompilationUnit[] getCompilationUnits() {
-		if(inputFiles==null) {
+		if (inputFiles == null) {
 			setInpuFiles(jdtCompiler.sources, jdtCompiler.templates);
 		}
-		
+
 		List<CompilationUnit> culist = new ArrayList<>();
 		for (SpoonFile f : inputFiles) {
 			if (filesToBeIgnored.contains(f.getPath())) {
@@ -76,11 +76,10 @@ public class FileCompiler extends JDTBatchCompiler {
 	public void setInpuFiles(List<SpoonFile> p_inpuFiles) {
 		inputFiles = p_inpuFiles;
 	}
-	
+
 	public void setInpuFiles(SpoonFolder... folders) {
 		inputFiles = new ArrayList<>();
-		for (SpoonFolder folder : folders)
-		{
+		for (SpoonFolder folder : folders) {
 			inputFiles.addAll(folder.getAllJavaFiles());
 		}
 	}
--- a/src/main/java/spoon/support/compiler/jdt/FileCompiler.java
+++ b/src/main/java/spoon/support/compiler/jdt/FileCompiler.java
@@ -24,11 +24,19 @@ import org.eclipse.jdt.internal.compiler.batch.CompilationUnit;
 
 import spoon.SpoonException;
 import spoon.compiler.SpoonFile;
+import spoon.compiler.SpoonFolder;
 
 public class FileCompiler extends JDTBatchCompiler {
+	
+	protected List<SpoonFile> inputFiles;
 
 	public FileCompiler(JDTBasedSpoonCompiler jdtCompiler) {
 		super(jdtCompiler);
+	}
+
+	public FileCompiler(JDTBasedSpoonCompiler jdtCompiler, List<SpoonFile> files) {
+		super(jdtCompiler);
+		inputFiles = files;
 	}
 
 	/**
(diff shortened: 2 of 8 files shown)


src/main/java/spoon/compiler/builder/ClasspathOptions.java
```diff
@@ -42,22 +44,26 @@ public T classpath(String... classpaths) {
 	}
 
 	public T classpathFromListOrClassLoader(String... classpaths) {
-		return (classpaths != null && classpaths.length > 0) ? classpath(classpaths) : classpathFromCurrentClassLoader();
```
monperrus: required for this bug fix?

src/main/java/spoon/support/compiler/jdt/FileCompiler.java
```diff
@@ -24,23 +24,32 @@
...
 
 public class FileCompiler extends JDTBatchCompiler {
 
+	protected List<SpoonFile> inputFiles;
+
 	public FileCompiler(JDTBasedSpoonCompiler jdtCompiler) {
 		super(jdtCompiler);
 	}
 
+	public FileCompiler(JDTBasedSpoonCompiler jdtCompiler, List<SpoonFile> files) {
```
monperrus: what passing the files to the constructor to have an immutable object?

src/main/java/spoon/support/compiler/jdt/JDTBasedSpoonCompiler.java
```diff
@@ -401,55 +380,90 @@ protected JDTBatchCompiler createBatchCompiler(InputType... types) {
...
+	private static final CompilationUnitDeclaration[] EMPTY_RESULT = new CompilationUnitDeclaration[0];
+
+	protected CompilationUnitDeclaration[] buildUnits(JDTBuilder jdtBuilder, SpoonFolder sourcesFolder, String[] classpath, String debugMessagePrefix, boolean buildOnlyOutdatedFiles) {
+		List<SpoonFile> sourceFiles = sourcesFolder.getAllJavaFiles();
+		if (sourceFiles.isEmpty()) {
+			return EMPTY_RESULT;
+		}
+		FileCompiler batchCompiler = (FileCompiler) createBatchCompiler(InputType.FILES);
+		//tell the compiler which files it should build
+		batchCompiler.setInpuFiles(sourceFiles);
```
pvojtechovsky: Here is the main bug fix, but I think my changes fixes more bugs. I am sorry that I did so many changes at once, but I needed these changes to simplify code to understand it. I believe I can simplify it even more, but I preffer to have your feedback in the middle of way, to not produce something what is not acceptable for you.

src/main/java/spoon/compiler/builder/ClasspathOptions.java
pvojtechovsky: required for simplification

src/main/java/spoon/support/compiler/jdt/FileCompiler.java
pvojtechovsky: this constructor is not needed. I will remove it

pvojtechovsky:
I will try to describe the ideas behind my changes. I will give them identifier (Ix), so we can discuss them separately.

I1: `JDTBasedSpoonCompiler.buildSources` and `JDTBasedSpoonCompiler.buildTemplates` had two different implementations, but I have feeling like they should be nearly same, because they do same think. They build java files and fills the CtModel. Just input is different, so I extracted common code into `JDTBasedSpoonCompiler.buildUnits`

I2: I created `JDTBasedSpoonCompiler.buildModel(...)`, which contains code which was copied on 3 places. In case of snipped compiling it did not processed comments. I guess it was a bug, but if it is feature, then I suggest to add parameter `boolean processComments`

I3: `JDTBasedSpoonCompiler.buildUnits(...)` `JDT bug HACK`. Was buggy. It expected that `createTmpJavaFile()` will produce Tmp.java file, and that if that file already exists that it does not creates new one. But `File.createTempFile()`, behaves different. It assures that it always creates new unique file like Tmp5875465467435.java. Second problem of the origin code was that if there would be more folders, then only last tmp file will be deleted. All others will stay there till JVM correctly exits. Anyway I have no idea whether my change is correct, so author of origin code should check it.

I4: `JDTBasedSpoonCompiler.buildUnits(...)` is used for building of java files defined by parameter `SpoonFolder sourcesFolder`, so the `JDTBatchCompiler` must get this list here to be able to produce correct array of `CompilationUnit`. See `batchCompiler.setInpuFiles(sourceFiles);` which is the main fix of this issue

I5: `JDTBatchCompiler.getUnits()` algorithm created JDT CompilationUnits more then it was needed. If you agree that it should build only the provided list of java files and not also another set of files which comes from somewhere else (like origin code), then it is clear now that we can reuse already made CompilationUnits and get them here `unit.compilationResult.compilationUnit`

**Remaining bugs**
B1: if the batch compiler args produced by `.sources(new SourceOptions().sources(sourceFiles))` are used (I guess they are not) and 'buildOnlyOutdatedFiles==true', then we should give it only outdated source files. So the outdated filtering code should be moved up.

B2: `JDTBatchCompiler.getUnits(sourceFiles)` has actually unused parameter 'sourceFiles'. We should fix it.
> A) to remove that parameter, but then we change a public API ... which did not worked before anyway. So it should be OK to change it. I vote for this.
> B) to refactor it even more so the `sourceFiles` are used and send them to JDTBatchCompiler by call of `setInpuFiles(...)`

**Other ideas - not yet implemented**
I10: JDTBatchCompiler should implement `getCompilationUnits()` and `setCompilationUnits(...)` and should remember array of compilation units in it. And may be we should delete `FactoryCompiler` and `FileCompiler` and just keep their functions to build list of CompilationUnits. Then the API of 'JDTBatchCompiler' will be quite clear ... it will compile the units which are provided by previous call of setCompilationUnits(...)
Note: I see that FileCompiler is used in `CompilationTest` for filtering of files and folders. It seems to me like quite tricky way to filer source input. Anyway if there is no other way, then we have to keep it.

I11: I would like to move shared code of `JDTSnippetCompiler.buildSources` to `JDTBasedSpoonCompiler.buildUnits` (or something similar) to minimize copies of code.

Note: I am highly motivated to have spoon running. At my work I am responsible for some old, long, copy pasted and dark code ... much darker then JDT integration of Spoon :-) ... so I am get used to analyze and refactor such things. And I hope that your nice Spoon might help me with that.

src/main/java/spoon/compiler/builder/ClasspathOptions.java
```diff
@@ -42,22 +44,26 @@ public T classpath(String... classpaths) {
...
+		return classpath(getClasspathFromListOrCurrentClassloader(classpaths));
 	}
 
-	public T classpathFromCurrentClassLoader() {
+	public static String[] getClasspathFromListOrCurrentClassloader(String... classpaths) {
+		return (classpaths != null && classpaths.length > 0) ? classpaths : getClasspathFromCurrentClassloader();
+	}
+	public static String[] getClasspathFromCurrentClassloader() {
```
tdurieux: Is this method need to be public?

tdurieux:
> I1: JDTBasedSpoonCompiler.buildSources and JDTBasedSpoonCompiler.buildTemplates

I think the classpath of two builds are different.

> I2: I created JDTBasedSpoonCompiler.buildModel(...)

Ok for me

> I3: JDTBasedSpoonCompiler.buildUnits(...) JDT bug HACK.  so author of origin code should check it.

The original author does not work anymore on Spoon :(
Personally, I do not know well the compilation part of Spoon...


> I4: JDTBasedSpoonCompiler.buildUnits(...) 

Ok

> I5: JDTBatchCompiler.getUnits() algorithm created JDT CompilationUnits more then it was needed

I have no idea why need investigation

Remaining bugs
> B1: OutdatedFiles

I did not know that this feature was implemented.
This is probably not well tested and buggy.

> B2: JDTBatchCompiler.getUnits(sourceFiles) has actually unused parameter 'sourceFiles'. We should fix it.
> A) to remove that parameter, but then we change a public API ... which did not worked before anyway. So it should be OK to change it. I vote for this.

This API is probablly never used by third party

Other ideas - not yet implemented
> I10: JDTBatchCompiler should implement getCompilationUnits() and setCompilationUnits(...)  it.

I like the idea of a stateless JDTBatchCompiler.

> I11: I would like to move shared code of JDTSnippetCompiler.buildSources to JDTBasedSpoonCompiler.buildUnits (or something similar) to minimize copies of code.

Good idea

monperrus:
Thanks a lot for the detailed analysis. 

Basically, the best way is to split the tasks in different PRs, so as to simplify the review process and so that we merge fast each of them. One PR per bug fix + test (*), one PR per refactoring. The process may seem slower but eventually it's faster.  So let's start by fixing this bug with the minimal diff/change (basically setInputList) and then we proceed with refactoring (or the other way around).  

Thanks for you great past and upcoming contributions.

(*) for instance the no-comment snippet compilation should be fixed and also tested.

pvojtechovsky:
> One PR per bug fix + test
I agree. I will check my changes and extract the smaller parts. I will not produce 20 PR today. I make one or two and after they are merged I will make next. This should minimize merging conflicts and extra effort. I suggest to rename this PR to something like Ideas for compiler improvements - so the analyze and feedback are kept. We can discuss here next steps too. (This my first github project, so let me know if I can do something better)

@tdurieux thanks for feedbacks, I will handle them soon

src/main/java/spoon/support/compiler/jdt/JDTBasedSpoonCompiler.java
```diff
@@ -401,55 +380,90 @@ protected JDTBatchCompiler createBatchCompiler(InputType... types) {
...
+	}
+
+	private static final CompilationUnitDeclaration[] EMPTY_RESULT = new CompilationUnitDeclaration[0];
+
+	protected CompilationUnitDeclaration[] buildUnits(JDTBuilder jdtBuilder, SpoonFolder sourcesFolder, String[] classpath, String debugMessagePrefix, boolean buildOnlyOutdatedFiles) {
+		List<SpoonFile> sourceFiles = sourcesFolder.getAllJavaFiles();
+		if (sourceFiles.isEmpty()) {
+			return EMPTY_RESULT;
+		}
+		FileCompiler batchCompiler = (FileCompiler) createBatchCompiler(InputType.FILES);
```
pvojtechovsky: I do not like it too, because it forces implementations to return only that type. I have problem with createBatchCompiler method. It internaly creates a compiler which is already initialized with set of files which has to be compiled, ... but in this context we need a different set of files. I do not like it. I would prefer to split creation of compiler and initialization of set of files. But how? This method is overwritten in some tests - as example how to filter files. This stops me. 

I would prefer to create here simply the JDTBatchComplier (no FileCompiler) and then to give it the files (or directly units) which has to be compiled. But this direct aproach needs changes in the tests... @monperrus How much do you persist on FileCompiler and FactoryCompiler which comes with built-in set of files? What do you suggest?

src/main/java/spoon/compiler/builder/ClasspathOptions.java
pvojtechovsky: Good point! It makes the usage of ClasspathOptions more complicated. I will move these static methods to place where I need them, so they do not make ClasspathOptions more compex.

monperrus:
> I would prefer to create here simply the JDTBatchComplier (no
> FileCompiler) and then to give it the files (or directly units) which
> has to be compiled.
That's a good point.


> But this direct aproach needs changes in the tests...

It's OK to change those tests.

monperrus:
> I make one or two and after they are merged I will make next. This should minimize merging conflicts and extra effort.

OK, closing this one then

PR #930, opened 2016-11-05
support of comments in snippet

pvojtechovsky:
The comments in snippet was always ignored. This PR includes this:

- Test case which fails when comments are lost in snippet
- Implementation of comments in snippet
- new method `CompilationUnit JDTSnippetCompiler.getSnippetCompilationUnit()`, which returns CompilationUnit of the compilet snippet. Note: I had to remember the CompilationUnit of the snippet in the CompilationUnit cache, otherwise JDTCommentBuilder was not able to add comments. And at the end I removed that CU from the cache ... so it is now possible to access that CU from the JDTSnippetCompiler too.

It is the first chunk of changes suggested by PR #929

### DIFF
--- a/src/main/java/spoon/reflect/factory/CompilationUnitFactory.java
+++ b/src/main/java/spoon/reflect/factory/CompilationUnitFactory.java
@@ -25,6 +25,7 @@ import spoon.reflect.cu.Import;
 import spoon.reflect.reference.CtFieldReference;
 import spoon.reflect.reference.CtPackageReference;
 import spoon.reflect.reference.CtTypeReference;
+import spoon.support.compiler.jdt.JDTSnippetCompiler;
 import spoon.support.reflect.cu.ImportImpl;
 
 /**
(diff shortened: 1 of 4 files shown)


src/main/java/spoon/reflect/factory/CompilationUnitFactory.java
```diff
@@ -76,6 +79,16 @@ public CompilationUnit create(String filePath) {
 	}
 
 	/**
+	 * Removes compilation unit from the cache and returns it
+	 * Used by JDTSnippetCompiler to remove processed snippet from the cache
+	 * @param filePath
+	 * @return a cached compilation unit or null
+	 */
+	public CompilationUnit remove(String filePath) {
```
monperrus: rename method to removeFromCache?

monperrus: package-visible instead of public?

pvojtechovsky: > removeFromCache

good idea

> package-visible

It is different package ... or I do not know this java feature...

pvojtechovsky:
I have committed the rename of method remove->removeFromCache. I do not know how to implement package visibility if the caller is in different package. I think it is not possible. But Travis does not want to check my last commit. OR am I wrong?

monperrus:
thanks!

PR #932, opened 2016-11-05
fix bug #921 Duplicity in package-info.java when templates are used

pvojtechovsky:
This is next part of #929. It assures that `JDTBasedSpoonCompiler.build()` compiles sources and templates only once. Before it was compiled twice, therefore the annotations were duplicated.

I had to add new `InputType.UNITS` to tell the `createBatchCompiler` method that it should create JDTBatchCompiler which let's us configure to be compiled files.

I do not like this solution, because public `InputType.UNITS` is there just for internal purposes and it makes no sense for `JDTBasedSpoonCompiler.compile()` ... so it is confusing... But I was not able to found beter way without changing of contract of createBatchCompiler method, which is may be used by clients to filter files and directories.

@monperrus if you agree that `InputType.UNITS` is wrong, then please refactor `createBatchCompiler` and related tests by some way, so it can be used in simpler way.

I vote for removing support of filtering by folders and files (test `testFilterResourcesDir` and `testFilterResourcesFile()`)by overwriting of protected method ... this is not good because protect function becomes a "public API"

### DIFF
--- a/src/main/java/spoon/SpoonModelBuilder.java
+++ b/src/main/java/spoon/SpoonModelBuilder.java
@@ -76,9 +76,13 @@ public interface SpoonModelBuilder {
 	 */
 	boolean build(JDTBuilder builder);
 
-	/** The types of compilable elements */
+	/** The types of compilable elements
+	 * FILES - compiles the java files from the file system, which were registered by {@see SpoonModelBuilder#addInputSource()} and {@see SpoonModelBuilder#addTemplateSource(File)}
+	 * CTTYPES - compiles virtual java files, which are dynamically generated from the all top level classes of the CtModel by {@see DefaultJavaPrettyPrinter}
+	 * UNITS - compiles the java files, which are specified later by JDTBasedSpoonCompiler.build. This InputType cannot be used by compile.
+	 */
 	enum InputType {
-		FILES, CTTYPES
+		FILES, CTTYPES, UNITS
 	}
 
 
--- a/src/main/java/spoon/support/compiler/jdt/ConfigurableCompiler.java
+++ b/src/main/java/spoon/support/compiler/jdt/ConfigurableCompiler.java
@@ -1,3 +1,19 @@
+/**
+ * Copyright (C) 2006-2016 INRIA and contributors
+ * Spoon - http://spoon.gforge.inria.fr/
+ *
+ * This software is governed by the CeCILL-C License under French law and
+ * abiding by the rules of distribution of free software. You can use, modify
+ * and/or redistribute the software under the terms of the CeCILL-C license as
+ * circulated by CEA, CNRS and INRIA at http://www.cecill.info.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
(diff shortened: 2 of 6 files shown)


monperrus:
I see the idea. Really good work to be backward compatible. 

What about creating a ConfigurableCompiler extends JDTBatchCompiler which has setInputFiles, and consequently JDTBatchCompiler remains unchanged with its abstract method?

pvojtechovsky:
> ConfigurableCompiler 

Yes, it is the option too. But I would prefer to get rid of FileCompiler and FactoryCompiler in future. Why?

- Why we need to create new instance of compiler when we just want to compile some special set of files? Isn't it more clear if standard compiler is made and the appropriate set of files is set as input?

But I am not sure. May be you are right and FileCompiler and FactoryCompiler  are good ideas. Then ConfigurableCompiler is good idea too. I do not know the spoon enough. 

... but I am sure that `InputType.UNITS` is not nice. It will confuse spoon clients...

I finish today. Thanks for fast merging other changes! Good night Martin. ;-)

pvojtechovsky:
I agree that in current implementation the `ConfigurableCompiler` is good solution. So I will do it today. May be later we remove it again, but that is OK.

pvojtechovsky:
I remeber now why I did not made ConfigurableCompiler at the beginning. It is not backward compatible. Anyway I am glad it is on this way. I suggest to change tests because code is better from my point of view. OK?

pvojtechovsky:
I did two things now

1)  `ConfigurableCompiler`
2) new method `JDTBatchCompiler createBatchCompiler(List<SpoonFile> files)` which internally uses old method - for backward compatibility.

monperrus:
Thanks.

Would you create another pull request with the alternative solution (adding setInputFiles, removing InputType.UNITS, FileCompiler, FactoryCompiler and changing the 3 tests)?

Than we can compare and make an informed decision.

pvojtechovsky:
Note: The file and folder filtering test must be changed anyway, because even in previous implementation where they passed, they caused usage of wrong files during compiling - so if somebody would real use this filtering approach together with templates, the compiler would compile all the files twice and would reproduce bug #921.

To the alternative PR - I will try it and let you know. May be it is better to merge this PR first and then optionally apply other improvements - because other is based on this PR

pvojtechovsky:
@tdurieux You wrote:
> I like the idea of the stateless JDTBAtchCompiler

could you plese explain it in more details? I would like to understand you before I make next PR. Thanks!

tdurieux:
I don't know if it is possible or it is currently the case but a compiler that does not have state is less likely to produce unexpected behavior when several compilations are launched on the same spoon instance.
It is not very important.

pvojtechovsky:
> the idea of the stateless JDTBatchCompiler

I understand, but actually JDTBatchCompiler extends from JDT...Main compiler which has internal states, so it cannot be used between more threads. But we are creating one instance of such compiler for each compilation thread, so conflict should not occure ... if JDT Main compiler does not uses not synchronized shared static resoruces.

tdurieux:
Oups, I forgot this state...

PR #934, opened 2016-11-06
alternative fix bug #921 Duplicity in package-info.java when templates are used

pvojtechovsky:
@monperrus Hi Martin, here is the alternative PR for #932 as you suggested.

It is variant, which uses non abstract `JDTBatchCompiler`. `FactoryCompiler` and `FileCompiler` does not inherit from `JDTBatchCompiler` and they were renamed to `FactoryCompilerConfig` and `FileCompilerConfig`, because they are now used to **configure** the `JDTBatchCompiler`.

`CompilationTest` file and folder filtering tests are commented for now. There is no support for filtering by overwriting of `createBatchCompiler`. I suggest to discuss how to implement it later.

`ExtendedStringLiteralTest` was updated. It should be OK.

`InputType` was changed from `enum` to `interface`. The change is compiler compatible so source code of the clients does not have to be changed. Just recompiled. Implementation of the `InputType` can configure `JDTBatchCompiler` before the compilation is started. 

```java
	interface InputType {
		InputType FILES = FileCompilerConfig.INSTANCE;
		InputType CTTYPES = FactoryCompilerConfig.INSTANCE;

		void initializeCompiler(JDTBatchCompiler compiler);
	}
```

So we have standard defaults, but now they can extend and use different configurations if needed. I use it internally instead of InputType.UNIT, I am using 
```java
... createBatchCompiler(new FileCompilerConfig(sources));
```


src/main/java/spoon/SpoonModelBuilder.java
```diff
@@ -76,9 +79,15 @@
...
-		FILES, CTTYPES
+	/** The types of compilable elements
+	 * FILES - compiles the java files from the file system, which were registered by {@see SpoonModelBuilder#addInputSource()} and {@see SpoonModelBuilder#addTemplateSource(File)}
+	 * CTTYPES - compiles virtual java files, which are dynamically generated from the all top level classes of the CtModel by {@see DefaultJavaPrettyPrinter}
+	 */
+	interface InputType {
+		InputType FILES = FileCompilerConfig.INSTANCE;
+		InputType CTTYPES = FactoryCompilerConfig.INSTANCE;
+
+		void initializeCompiler(JDTBatchCompiler compiler);
```
monperrus: add javadoc:
/** responsible for setting the parameters of JDTBatchCompiler, must call setCompilationUnits() */

src/main/java/spoon/support/compiler/jdt/JDTBatchCompiler.java
```diff
@@ -63,9 +66,36 @@
...
-	public abstract CompilationUnit[] getCompilationUnits();
+	public CompilationUnit[] getCompilationUnits() {
+		return compilationUnits;
+	}
+
+	public void setCompilationUnits(CompilationUnit[] compilationUnits) {
+		this.compilationUnits = compilationUnits;
+	}
+
+	public void setInputFiles(List<SpoonFile> files) {
```
monperrus: I propose that this method moves to FileCompilerConfig. And FileCompilerConfig would be responsible for calling setCompilationUnits().

src/test/java/spoon/test/compilation/CompilationTest.java
```diff
@@ -154,7 +146,7 @@ public void testNewInstance() throws Exception {
 		}
 
 	}
-
+/*
```
tdurieux: It is better to remove instead of commenting.

pvojtechovsky:
I have added comment and moved the method as suggested. I have also rebased to master

src/test/java/spoon/test/compilation/CompilationTest.java
pvojtechovsky: I will give it a try, but I would prefer if you merge my changes first, because then I can create another PRs too, related to compiler this evening. Otherwise I can fix only test and cannot do more until it is merged. Because of conflicts ...

PR #940, opened 2016-11-07
update CompilationTest to match latest compiler changes

pvojtechovsky:
Here are uncommented and fixed CompilationTest tests


monperrus:
thanks!

Issue #683, opened 2016-05-31
new feature: Statement always in a block


tdurieux:
Always put a block in loop and condition body.

In order to handle loops and conditions without block, we can set the block as implicit and then not print it.

PR #943, opened 2016-11-07
add interface CtBodyHolder

pvojtechovsky:
Note! I had to rollback `CtExecutable.getBody` declaration made by 

> GerardPaligot <gerard.paligot@gmail.com> on 7.8.15 14:20

to make it compilable with new `CtBodyHolder`. I have not found a way how to declare `CtBodyHolder` to have it compilable with origin `CtExecutable.getBody` declaration.

I have no idea whether my changes are correct. It might have some impact to Spoon templates!

### DIFF
--- a/src/main/java/spoon/reflect/code/CtBodyHolder.java
+++ b/src/main/java/spoon/reflect/code/CtBodyHolder.java
@@ -16,33 +16,15 @@
  */
 package spoon.reflect.code;
 
+import spoon.reflect.declaration.CtElement;
+
 /**
- * This code element defines a <code>catch</code> of a <code>try</code>.
- *
- * @see spoon.reflect.code.CtTry
+ * This abstract code element defines an element, which contains a body
  */
-public interface CtCatch extends CtCodeElement, CtBodyHolder {
+public interface CtBodyHolder extends CtElement {
 
 	/**
-	 * Gets the catch's parameter (a throwable).
+	 * Gets the body of this element
 	 */
-	CtCatchVariable<? extends Throwable> getParameter();
-
-	/**
-	 * Sets the catch's parameter (a throwable).
-	 */
-	<T extends CtCatch> T setParameter(CtCatchVariable<? extends Throwable> parameter);
-
-	/**
-	 * Gets the catch's body.
-	 */
-	CtBlock<?> getBody();
-
-	/**
-	 * Sets the catch's body.
-	 */
-	<T extends CtCatch> T setBody(CtBlock<?> body);
-
-	@Override
-	CtCatch clone();
+	CtStatement getBody();
 }
--- a/src/main/java/spoon/reflect/code/CtBodyHolder.java
+++ b/src/main/java/spoon/reflect/code/CtBodyHolder.java
@@ -27,4 +27,9 @@ public interface CtBodyHolder extends CtElement {
 	 * Gets the body of this element
 	 */
 	CtStatement getBody();
+
+	/**
+	 * Sets the body of this element.
+	 */
+	<T extends CtBodyHolder> T setBody(CtStatement body);
 }
--- a/src/main/java/spoon/reflect/code/CtBodyHolder.java
+++ b/src/main/java/spoon/reflect/code/CtBodyHolder.java
@@ -30,6 +30,7 @@ public interface CtBodyHolder extends CtElement {
 
 	/**
 	 * Sets the body of this element.
+	 * If body is not a block, it is wrapped in a CtBlock which is semantically equivalent and eases transformation afterwards if required.
 	 */
 	<T extends CtBodyHolder> T setBody(CtStatement body);
 }
--- a/src/main/java/spoon/reflect/code/CtCatch.java
+++ b/src/main/java/spoon/reflect/code/CtCatch.java
@@ -36,12 +36,14 @@ public interface CtCatch extends CtCodeElement, CtBodyHolder {
 	/**
 	 * Gets the catch's body.
 	 */
+	@Override
 	CtBlock<?> getBody();
 
 	/**
-	 * Sets the catch's body.
+	 * Sets the catch's body. The instance of CtBlock makes sense too
 	 */
-	<T extends CtCatch> T setBody(CtBlock<?> body);
+	@Override
+	<T extends CtBodyHolder> T setBody(CtStatement body);
 
 	@Override
 	CtCatch clone();
--- a/src/main/java/spoon/reflect/code/CtCatch.java
+++ b/src/main/java/spoon/reflect/code/CtCatch.java
@@ -39,12 +39,6 @@ public interface CtCatch extends CtCodeElement, CtBodyHolder {
 	@Override
 	CtBlock<?> getBody();
 
-	/**
-	 * Sets the catch's body. The instance of CtBlock makes sense too
-	 */
-	@Override
-	<T extends CtBodyHolder> T setBody(CtStatement body);
-
 	@Override
 	CtCatch clone();
 }
--- a/src/main/java/spoon/reflect/code/CtLoop.java
+++ b/src/main/java/spoon/reflect/code/CtLoop.java
@@ -26,12 +26,14 @@ public interface CtLoop extends CtStatement, TemplateParameter<Void>, CtBodyHold
 	/**
 	 * Gets the body of this loop.
 	 */
+	@Override
 	CtStatement getBody();
 
 	/**
-	 * Sets the body of this loop.
+	 * Sets the body of this loop. The instance of CtBlock makes sense too
 	 */
-	<T extends CtLoop> T setBody(CtStatement body);
+	@Override
+	<T extends CtBodyHolder> T setBody(CtStatement body);
 
 	@Override
 	CtLoop clone();
--- a/src/main/java/spoon/reflect/code/CtLoop.java
+++ b/src/main/java/spoon/reflect/code/CtLoop.java
@@ -29,12 +29,6 @@ public interface CtLoop extends CtStatement, TemplateParameter<Void>, CtBodyHold
 	@Override
 	CtStatement getBody();
 
-	/**
-	 * Sets the body of this loop. The instance of CtBlock makes sense too
-	 */
-	@Override
-	<T extends CtBodyHolder> T setBody(CtStatement body);
-
 	@Override
 	CtLoop clone();
 }
--- a/src/main/java/spoon/reflect/code/CtLoop.java
+++ b/src/main/java/spoon/reflect/code/CtLoop.java
@@ -21,7 +21,7 @@ import spoon.template.TemplateParameter;
 /**
  * This abstract code element defines a loop.
  */
-public interface CtLoop extends CtStatement, TemplateParameter<Void> {
+public interface CtLoop extends CtStatement, TemplateParameter<Void>, CtBodyHolder {
 
 	/**
 	 * Gets the body of this loop.
--- a/src/main/java/spoon/reflect/code/CtTry.java
+++ b/src/main/java/spoon/reflect/code/CtTry.java
@@ -55,12 +55,14 @@ public interface CtTry extends CtStatement, TemplateParameter<Void>, CtBodyHolde
 	/**
 	 * Sets the tried body.
 	 */
+	@Override
 	CtBlock<?> getBody();
 
 	/**
-	 * Sets the tried body.
+	 * Sets the tried body. The instance of CtBlock makes sense too
 	 */
-	<T extends CtTry> T setBody(CtBlock<?> body);
+	@Override
+	<T extends CtBodyHolder> T setBody(CtStatement body);
 
 	/**
 	 * Gets the <i>finalizer</i> block of this <code>try</code> (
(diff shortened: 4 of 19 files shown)


monperrus:
Hi,
It's OK to change the return of `CtExecutable.getBody()`.

We also need a method `setBody()` in `CtBodyHolder`.

pvojtechovsky:
Note that some children of CtBodyHolder are using
`setBody(CtStatement)`
and some 
`setBody(CtBlock)`

These which expects `setBody(CtBlock)` have this new method
```java
	public <T extends CtBodyHolder> T setBody(CtStatement statement) {
		CtBlock<?> block;
		if (statement instanceof CtBlock<?>) {
			block = (CtBlock<?>) statement;
		} else {
			block = getFactory().Code().createCtBlock(statement);
		}
		return setBody(block);
	}
```
Another solution would be to throw exception that CtBlock is expected. But I vote for automatic creation of CtBlock, because it makes live of clients more comfortable.

monperrus:
> But I vote for automatic creation of CtBlock, because it makes live of
> clients more comfortable.
Yes, that's fine, that's along the line of
47bd0f2030cb07b225b8a575c3ee2fa779c0c0c7

pvojtechovsky:
ParentContractTest failed, but I do not understand what is that. Could you fix it or explain it?

monperrus:
For different reasons, first because having both `setBody(CtStatement)` and `setBody(CtBlock)` relies on correct dispatching on arguments, which is done at compile time in cryptic ways. 
If you change:
```
	@Override
	public <T extends CtBodyHolder> T setBody(CtStatement statement) {
		CtBlock<?> block;
		if (statement instanceof CtBlock<?>) {
			block = (CtBlock<?>) statement;
		} else {
			block = getFactory().Code().createCtBlock(statement);
		}
		return setBody(block);
	}
```
by
```
	@Override
	public <E extends CtBodyHolder> E setBody(CtStatement statement) {
		CtBlock<? extends T> block;
		if (statement instanceof CtBlock<?>) {
			block = (CtBlock<? extends T>) statement;
		} else {
			block = (CtBlock<? extends T>) getFactory().Code().createCtBlock(statement);
		}
		return setBody(block);
	}
```
 
then it works.

But then, other contracts are violated, and for good reasons. I would go for removing the automatic block creation in `setBody(CtStatement statement)`. This can be done in a helper method for those who need this.

pvojtechovsky:
> I would go for removing the automatic block creation in setBody(CtStatement statement).

So it should throw exception if client does not sent CtBody?

The I do not understand how will look the **generic** code which can call CtBodyHolder.setBody() 

A) client does not know whether CtBlock or CtStatement has to be used, because it does not know real instance which is behind. And if it uses the wrong value then we throw an exception?
B) client has to know what it is calling... then we do not need CtBodyHolder and they can directly call e.g. CtCatch.setBody.

monperrus:
if we all agree on #956, it should solve the parent problem.

pvojtechovsky:
So should I change something in that PR? I do not understand the problem of ParentContractTest so I do not know what to do

pvojtechovsky:
Now I see that `CtExecutableImpl.setBody(CtStatement)` called recursively `setBody(CtStatement)`, but it should call `setBody(CtBlock)` of course. Fixed

monperrus:
> I do not understand the problem of ParentContractTest so I do not know
> what to do
I suggest to first rebase.

pvojtechovsky:
Now the `spoon.processing.CtGenerationTest` fails, but it succeeds in my local workspace. I have no idea what is the problem.

monperrus:
> I have no idea what is the problem.

Because since you have a change the signature of a metamodel method, you also have to regenerate the ReplacementVisitor. It's generated by `CtGenerationTest` in `target` what you have to do after your local test is:

    cp ./target/generated/spoon/support/visitor/replace/ReplacementVisitor.java ./src/main/java/spoon/support/visitor/replace/ReplacementVisitor.java

and to add it in the commit.

pvojtechovsky:
What about changing all setBody methods to have the same signature setBody(statement)? And convert it block if needed internaly.
Client can send both statement or block and internally we would convert it or throw exception if input is mot compatible with target element

monperrus:
It would be less intuitive for clients who just rely on the type system for documentation. I prefer to keep it this way.

monperrus:
the tests pass hurray!

could you push ReplacementVisitor with Unix newlines (the diff will be much smaller)?

also could you add tests for the new setBody() methods?

monperrus:
Afterthought about this. Instead of adding 3x `setBody(CtStatement)`. What about using `setBody(CtBlock)` as follows and only adding `setBody(CtStatement)` in CtLoop for `backward-compatibity`?


```
public interface CtBodyHolder extends CtElement {
	/**
	 * Gets the body of this element
	 */
	CtBlock getBody();

	/*** Sets the body of this element.
	 */
	<T extends CtBodyHolder> T setBody(CtStatement CtBlock);
}
```

pvojtechovsky:
> the tests pass hurray!

;-) ... no, this PR is never ending story :( 
After you merge #968, the tests will fail again, because actually this API contract tests simply checks nothing.

>  What about using setBody(CtBlock) as follows and only adding setBody(CtStatement) in CtLoop for backward-compatibity?

I would prefer opposite. In suggest to use `setBody(CtStatement)` in all 4 cases. The getBody would stay like it is -some have CtStatement and some have CtBlock.
**pros**
- client of spoon has to write less code when calling setBody
  - in case s/he wants to have only single statement in the body, then s/he simply calls setBody(CtStatement). S/he does not have to create wrapping CtBlock. The wrappnig CtBlock would be made automatically internally for CtElements which expects body of type CtBlock
  - in case s/he wants to have block of statements, then s/he simply calls the same setBody(CtStatement), but this time with CtBlock instance. It will pass too.
- it is the case with minimum code for the maintenance

**cons**
- It would be less intuitive for clients who just rely on the type system for documentation. You are right, that if a developer starts to work with spoon first day, then s/he will need longer time to found that s/he has to create CtBlock to be able to add more statements into CtExecutable. Because s/he does not know spoon model and has no idea that there is something like CtBlock. But everybody who works with spoon longer will understand that CtBlock is a set of CtSatements and will like that they do not have to create CtBlock manually if they just want to add one statement somewhere.

I understand, that from technical point of view (java specification), the CtExecutable.body must be a `block` and compiler would fail if there would be just a `statement`. 
**But we all love spoon, because it abstracts from technical details.** 
So again an argument why CtStatement is OK, because from abstract point of view it is correct when CtExecutable has only one statement.

And if setBody(CtBlock) is always used, then everybody has to always create CtBlock even if s/he wants to add one statement only. What is advantage in that?

I see, ... hard choice. ;-) 
I am using spoon just for short time, and I have never called setBody... yet (I have not wrote tests for that PR yet :-))). So it is quite good chance that Your opinion is the correct one. But I actually do not understand why.

src/main/java/spoon/support/reflect/code/CtLoopImpl.java
```diff
@@ -31,8 +32,9 @@ public CtStatement getBody() {
 		return body;
 	}
 
+	@SuppressWarnings("unchecked")
 	@Override
-	public <T extends CtLoop> T setBody(CtStatement body) {
+	public <T extends CtBodyHolder> T setBody(CtStatement body) {
 		if (body != null) {
```
monperrus: to be consistent, we should also create a block if the argument is not a block

monperrus:
@tdurieux agree with you, and your reasoning makes sense.

however, we should also modify CtLoopImpl.setBody to be consistent.

pvojtechovsky:
> @tdurieux agree with you, and your reasoning makes sense.

Do I understand well, that you will accept the PR, which has **only one method `setBody` in all children of CtBodyHolder and this one method has one parameter `CtStatement`.**

It of course means, that all methods setBody(CtBlock) will be removed.

Yes, I like that solution. But I do not like to rewrite it again and again ... so I would prefer to have your agreement about that concept, before I spent hours of my free time for that. ;-)

monperrus:
> But I do not like to rewrite it again and again ... so I would prefer to have your agreement about that concept, before I spent hours of my free time for that. ;-)

I understand :-)

>  that all methods setBody(CtBlock) will be removed.

Yes but it will be backward compatible both statically (`CtStatement` is an upper class) and dynamically (same behavior)

And we'll keep the return type of `getBody` to be `CtBlock` (we can override the return type in subinterfaces of `CtBodyHolder`) for those for which it was already the case (backward compatible, dynamically enforced by the setter).

This solution seems to be 1) fully backward compatible and 2) more powerful because it abstracts away details and simplifies the job of client code.

monperrus:
You'll also have to rebase due to the change in `ReplacementVisitor` in sorry.

src/main/java/spoon/support/reflect/code/CtLoopImpl.java
pvojtechovsky: I do not understand why. CtLoop accepts statement, so we should not create CtBlock. If you speak about change of spoon model, then I would prefer to do it in another PR ... PLEASE!!! :-)))

src/main/java/spoon/reflect/code/CtCatch.java
```diff
@@ -36,12 +36,14 @@
 	/**
 	 * Gets the catch's body.
 	 */
+	@Override
 	CtBlock<?> getBody();
 
 	/**
-	 * Sets the catch's body.
+	 * Sets the catch's body. The instance of CtBlock makes sense too
```
monperrus: we don't need to add it again in CtCatch if we don't change the return type or the contract.

src/main/java/spoon/reflect/code/CtBodyHolder.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+public interface CtBodyHolder extends CtElement {
+
+	/**
+	 * Gets the body of this element
+	 */
+	CtStatement getBody();
+
+	/**
+	 * Sets the body of this element.
```
monperrus: here we should document the automated block creation.

"If body is not a block, it is wrapped in a CtBlock which is semantically equivalent and eases transformation afterwards if required."

src/main/java/spoon/reflect/code/CtLoop.java
```diff
@@ -21,17 +21,19 @@
...
 	 */
+	@Override
 	CtStatement getBody();
 
 	/**
-	 * Sets the body of this loop.
+	 * Sets the body of this loop. The instance of CtBlock makes sense too
 	 */
-	<T extends CtLoop> T setBody(CtStatement body);
+	@Override
```
monperrus: same

src/main/java/spoon/reflect/code/CtTry.java
```diff
@@ -55,12 +55,14 @@
 	/**
 	 * Sets the tried body.
 	 */
+	@Override
 	CtBlock<?> getBody();
 
 	/**
-	 * Sets the tried body.
+	 * Sets the tried body. The instance of CtBlock makes sense too
```
monperrus: same

src/main/java/spoon/support/reflect/code/CtLoopImpl.java
monperrus: for two reasons:
1) because we want a 100% consistent behavior over all subinterfaces of CtBodyHolder
2) because it's 100% along the line of the new AlwaysInBlock contract introduced in #683

src/main/java/spoon/reflect/factory/CodeFactory.java
```diff
@@ -508,6 +508,25 @@ public CodeFactory(Factory factory) {
...
 	/**
+	 * Accepts instance of CtStatement or CtBlock.
+	 * If element is CtStatement, then it creates wrapping CtBlock, which contains the element
+	 * If element is CtBlock, then it directly returns that element
+	 * If element is null, then it returns null.
+	 * note: It must not create empty CtBlock - as expected in CtCatch, CtExecutable, CtLoop and CtTry setBody implementations
+	 * @param element
+	 * @return CtBlock instance
+	 */
+	public <T extends CtStatement> CtBlock<?> getOrCreateCtBlock(T element) {
```
monperrus: excellent to factorize the code!

pvojtechovsky: Thanks!
What about setImplicit? it is mentioned in #683
```java
return this.createCtBlock(element).setImplicit(true);
```

pvojtechovsky:
I have forgot to add tests. I will do it tommorrow

monperrus:
No, setImplicit is only for JDTTReeBuilder. As soon as you use an
intercession method, all modified and created blocks must be explicit.

src/main/java/spoon/reflect/code/CtTry.java
```diff
@@ -53,16 +53,12 @@
 	boolean removeCatcher(CtCatch catcher);
 
 	/**
-	 * Sets the tried body.
+	 * Gets the tried body.
```
monperrus: Thanks for the first type, one remains "try body" (and not tried)

monperrus:
looks like we're almost done :-)

pvojtechovsky:
I had to fix some failing tests too. They were creating CtElement instances without factory. I guess it was a bug. Or should spoon work well when CtElement.factory == null ??

pvojtechovsky:
> looks like we're almost done :-)

It will be like a X-mass for me :-) But tomorrow. The test is missing

monperrus:
> Or should spoon work well when CtElement.factory == null ??
No. Setting the factory is always necessary.

monperrus:
many thanks Pavel.

PR #949, opened 2016-11-10
feat(CoreFactory): add create(Class) so as to create elements by reflection

monperrus:


monperrus:
Indeed it is, it is required otherwise the test also considers the new
create method.

tdurieux:
The PR does not contains test related to the added code.

monperrus:
it does: see SpoonArchitectureEnforcer.java
<https://github.com/INRIA/spoon/pull/949/files#diff-77ab0c75ea41f38f415582afc6510857>

tdurieux:
Your test is not a metamodelPackageRule

PR #951, opened 2016-11-10
fix(isAssignableFrom): handle primitive in isAssignableFrom

tdurieux:
In response to #948


monperrus:
What do you mean?

tdurieux:
I created a useful isAssignableFrom method that does not have the same behavior of isSubtypeOf

Issue #959, opened 2016-11-10
CtClass.getSuperClass() returns invalid CtTypeReference


pvojtechovsky:
The `spoon.test.imports.ImportTest.testSpoonWithImports` fails with PR #913, because it now calls `CtTypeReference.getActualClass()` on invalid `CtTypeReference` to child class 

```java
spoon.test.imports.testclasses.internal.SuperClass$InnerClassProtected
```
which has wrong qualified name
```java
spoon.test.imports.testclasses.internal.ChildClass$InnerClassProtected
```
I will provide PR with simplified failing test.

pvojtechovsky:
The wrong declaringType of that inner class is caused by this code in `JDTTreeBuilderHelper` method `createType(TypeDeclaration typeDeclaration)`
```java
				// Sorry for this hack but see the test case ImportTest#testImportOfAnInnerClassInASuperClassPackage.
				// JDT isn't smart enough to return me a super class available. So, I modify their AST when
				// superclasses aren't in the same package and when their visibilities are "default".
				List<ModifierKind> modifiers = Arrays.asList(ModifierKind.PUBLIC, ModifierKind.PROTECTED);
				final TypeBinding resolvedType = typeDeclaration.superclass.resolvedType;
				if ((resolvedType instanceof MemberTypeBinding || resolvedType instanceof BinaryTypeBinding)//
						&& resolvedType.enclosingType() != null && typeDeclaration.enclosingType.superclass != null//
						&& Collections.disjoint(modifiers, getModifiers(resolvedType.enclosingType().modifiers))) {
					typeDeclaration.superclass.resolvedType = jdtTreeBuilder.new SpoonReferenceBinding(typeDeclaration.superclass.resolvedType.sourceName(),
							(ReferenceBinding) typeDeclaration.enclosingType.superclass.resolvedType);
				}
```
@GerardPaligot please have a look. As author you can fix it probably the fastest. Thank You!

@monperrus By the way, how do you manage assignment of issues to contributors? May be somebody is already fixing it and I was just wasting my time ;-)

monperrus:
> By the way, how do you manage assignment of issues to contributors?

I don't assign anything. If somebody starts working on an issue, she/he can simply drop a comment on it to let the others know.

tdurieux:
I don't know how to fix this.
With the current implementation we use the qualified name as an access path not the path to the declaration.
In the current metamodel of spoon we cannot store the these informations.
[...]

PR #960, opened 2016-11-10
test bug CtClass.getSuperClass() returns invalid CtTypeReference

pvojtechovsky:
Added test  `spoon.test.imports.ImportTest.testAccessToNestedClass` which reproduces #959

### DIFF
--- a/src/main/java/spoon/reflect/declaration/CtTypeInformation.java
+++ b/src/main/java/spoon/reflect/declaration/CtTypeInformation.java
@@ -96,6 +96,12 @@ public interface CtTypeInformation {
 	boolean isGenerics();
 
 	/**
+	 * @param type
+	 * @return true if this Type is visible in class type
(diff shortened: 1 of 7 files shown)


pvojtechovsky:
I have tried another solution, which may be fixes #959:

1) keep correct qualified name of the classes (the JDT hack removed)
2) import only classes which are visible
3) Use the class identifier only if it is visible

monperrus:
just discussed with @tdurieux 

the behavior specified in ImportTest is correct. What is tricky here is that the super class in Java is actually an access path not a fully-qualified name. This happens in very very rare cases. And the access path depends on the visibility, that's why this behavior is correct. We have updated ImportTest accordingly, with to versions (fully-qualified and autoimport), see PR #971.

Consequently, this fix is incorrect, and you have to find a workaround in #913

PR #955, opened 2016-11-10
SerializationModelStreamer does not work

tdurieux:
I dont know how to fix it

PR #964, opened 2016-11-11
fix part of serialization problem

pvojtechovsky:
After this PR is applied the #955 will pass the tests.


monperrus:
wy do you say "part of" in the title? it looks like a complete fix for #955

pvojtechovsky:
It is complete fix for #955 if it is merged together with #955. This PR without #955 will not solve the origin serialization problem.

monperrus:
OK. @tdurieux can merge your fix in his PR.

monperrus:
thanks a lot for the fix!

PR #967, opened 2016-11-12
add CtTypeInformation.getDeclaredOrInheritedField(fieldName)

pvojtechovsky:
Here you are. New PR just for CtTypeInformation.getDeclaredOrInheritedField(fieldName). Including some tests.

### DIFF
--- a/src/main/java/spoon/reflect/declaration/CtTypeInformation.java
+++ b/src/main/java/spoon/reflect/declaration/CtTypeInformation.java
@@ -132,6 +132,20 @@ public interface CtTypeInformation {
 	Collection<CtFieldReference<?>> getAllFields();
 
 	/**
+	 * Gets a field from its name.
+	 *
+	 * @return null if does not exit
+	 */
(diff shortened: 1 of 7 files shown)


monperrus:
now that we have shadow classes, I'd propose to put those two methods in CtType and the methods would return CtField (instead of CtFieldReference)

monperrus:
well, no, because it would not be aligned with the rest of the design. ideally, now that we have shadow classes, we could get rid of CtTypeInformation that returns references.

pvojtechovsky:
What do you mean by `shadow classes`? 

In case of PrettyPrinting the references to fields are better, Because their creation is less expensive. PrettyPrinting needs only field name so reference is good.

Anyway I am not sure what exactly would you expect. I do not feel need for that change.

monperrus:
that's fine this way.

> What do you mean by shadow classes? 

CtClasses built by runtime reflection (hence with no  method body)

PR #973, opened 2016-11-14
fix(getTypeDeclaration): getTypeDeclaration('?') should point to Object

monperrus:


monperrus:
@tdurieux discussed change done

PR #975, opened 2016-11-14
fix(test): much stronger contracts for setter spec and tests

monperrus:
New behavior: all setters are fail-safe: if they don't make sense on a certain type, they do nothing (instead of throwing an exception).

This is better because:

* this is visible from the API doc thanks to annotation `@UnsettableAnnotation`
* the client code is not polluted by defensive try/catch blocks

New testing strategy: we test all setters:

* incl. those that take a list as argument (and we put one element in that list). 
* incl those that take a reference as argument

By doing this, we fix a couple of bugs a parent of references.

and also removed the duplicate `ParentTest` class.


src/main/java/spoon/support/reflect/code/CtConstructorCallImpl.java
```diff
@@ -173,8 +177,9 @@ public boolean removeActualTypeArgument(CtTypeReference<?> actualTypeArgument) {
 
 	@Override
 	public <C extends CtTypedElement> C setType(CtTypeReference<T> type) {
-		if (getExecutable() != null) {
+		if (getExecutable() != null && type != null) {
 			getExecutable().setType(type);
+			type.setParent(this);
```
tdurieux: null check after

src/main/java/spoon/support/reflect/code/CtInvocationImpl.java
```diff
@@ -150,8 +154,9 @@ public void replace(CtStatement element) {
 
 	@Override
 	public <C extends CtTypedElement> C setType(CtTypeReference<T> type) {
-		if (getExecutable() != null) {
+		if (getExecutable() != null && type != null) {
 			getExecutable().setType(type);
+			type.setParent(this);
```
tdurieux: null check after

PR #976, opened 2016-11-15
check that Template is on source path of builder

pvojtechovsky:
Template engine should throw an exception when template is not part of model - when it is created using java reflection - without method bodies.

Without that exception, the clients might wonder why templates does not work for them ... like me :-)

If you agree, it is good idea, then I will add the check to all other templates too.


monperrus:
thanks!

PR #979, opened 2016-11-16
refactor: check everywhere that Template is on source path of builder

pvojtechovsky:


src/main/java/spoon/template/Substitution.java
```diff
@@ -533,4 +537,40 @@ public static void redirectTypeReferences(CtElement element, CtTypeReference<?>
 		}
 	}
 
+	static <T> CtClass<T> getTemplateCtClass(CtType<?> targetType, Template<?> template) {
```
monperrus: javadoc?

src/main/java/spoon/template/Substitution.java
```diff
@@ -533,4 +537,40 @@ public static void redirectTypeReferences(CtElement element, CtTypeReference<?>
...
+			// if it's template with reference replacement
+			factory = targetType.getFactory();
+		} else {
+			// else we have at least one template parameter with a factory
+			factory = getFactory(template);
+		}
+		return getTemplateCtClass(factory, template);
+	}
+
+	static <T> CtClass<T> getTemplateCtClass(Factory factory, Template<?> template) {
```
monperrus: javadoc?

monperrus:
It seems that you like static methods very much. Why?

pvojtechovsky:
Do you mean new static `getTemplateCtClass` in Substitution class? I do not like static. I like when algorithm is implemented only once and there are minimized dependencies. And it is the case now.

monperrus:
> Do you mean new static |getTemplateCtClass| in Substitution class? I
> do not like static. I
OK, we're on the same page. I also think that static methods are bad
future unanticipated reuse with polymorphism and with respect to discovery.

PR #984, opened 2016-11-17
feat(position): improve the element positions

tdurieux:


src/main/java/spoon/reflect/cu/position/SourcePosition.java
```diff
@@ -14,16 +14,67 @@
  * The fact that you are presently reading this means that you have had
  * knowledge of the CeCILL-C license and that you accept its terms.
  */
-package spoon.reflect.cu;
```
monperrus: keep in same package

monperrus:
some doc in /doc

src/main/java/spoon/reflect/cu/SourcePosition.java
```diff
@@ -16,14 +16,64 @@
...
+import java.io.File;
+import java.io.Serializable;
+
 /**
  * This interface represents the position of a program element in a source file.
  */
-public interface SourcePosition {
+public interface SourcePosition extends Serializable {
+
+	SourcePosition NOPOSITION = new SourcePosition() {
```
monperrus: singleton outside, declaration here

PR #986, opened 2016-11-18
[WIP] Consider comments in EqualsVisitor as well.

retuxx:
Do I have to copy the file EqualsVisitor.java from target/generated/spoon/support/visitor/equals/EqualsVisitor.java to src/main/java/spoon/support/visitor/equals/EqualsVisitor.java manually? The test case CtGenerationTest.testGenerateEqualsVisitor fails because the generated and commit files differ.

### DIFF
--- a/src/main/java/spoon/support/visitor/equals/EqualsChecker.java
+++ b/src/main/java/spoon/support/visitor/equals/EqualsChecker.java
@@ -18,6 +18,7 @@ package spoon.support.visitor.equals;
 
 import spoon.reflect.code.CtBinaryOperator;
 import spoon.reflect.code.CtBreak;
+import spoon.reflect.code.CtComment;
 import spoon.reflect.code.CtContinue;
 import spoon.reflect.code.CtLiteral;
 import spoon.reflect.code.CtOperatorAssignment;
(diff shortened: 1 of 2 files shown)


src/main/java/spoon/support/visitor/equals/EqualsChecker.java
```diff
@@ -254,4 +255,33 @@ public void visitCtContinue(CtContinue e) {
 		}
 		super.visitCtLiteral(e);
 	}
+
+	@Override
+	public void visitCtComment(CtComment e) {
```
tdurieux: There is an implemented equals in CtComment, why do you not use it?

retuxx: I don't see any equals method in [CtComment](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/reflect/code/CtComment.java)

tdurieux: It is available in the implementation: https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/reflect/code/CtCommentImpl.java#L61

retuxx: But `visitCtComment` takes a `CtComment`. Should I really depend on the `CtCommentImpl` implementation?

PR #990, opened 2016-11-18
performance: cache URLs of sourceClasspath

pvojtechovsky:
This PR improves performance of MainTest.testMain() test on MS Windows 10 to 50% of origin time. From 45s to 22s.

### DIFF
--- a/src/main/java/spoon/support/StandardEnvironment.java
+++ b/src/main/java/spoon/support/StandardEnvironment.java
@@ -333,6 +333,12 @@ public class StandardEnvironment implements Serializable, Environment {
 	}
 
 	private ClassLoader classloader;
+	/*
+	 * cache class loader which loads classes from source class path
+	 * we must cache it to make all the loaded classes compatible
+	 * The cache is reset when setSourceClasspath(...) is called
+	 */
+	private ClassLoader inputClassloader;
 
 	@Override
 	public void setInputClassLoader(ClassLoader aClassLoader) {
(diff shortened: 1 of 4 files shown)


pvojtechovsky:
It looks like build engine (Travis) is ill...

monperrus:
what's the reason, the call to `toURI().toURL()`?

pvojtechovsky:
toURI().toURL() is not my code. It was there before. So I do not know. I know that in java 4,5,6 theref was different behavior in file to url conversion. May be author solved that problem...

monperrus:
what about removing field `String[] sourceClasspath` so as to avoid duplicating the state? (and keeping a backward compatible API with on-the-fly translation)

pvojtechovsky:
> removing field String[] sourceClasspath so as to avoid duplicating the state?

good idea. I will try it.

I have noticed that instance of ClassLoader returned by 
```java
ClassLoader getInputClassLoader()
```
is not cached too, Normaly it is not ok to create new class loader fot each java class, but I have not analyzed yet, how it is used in spoon. May be it is intentional here. I measured no performance change with cached class loader.

tdurieux:
If I correctly remember ```getInputClassLoader()``` is not used by Spoon.

pvojtechovsky:
I have checked how `getInputClassLoader()` is used. It is used for loading of classes whenever java reflection is used by spoon to gather information about classes.
So I changed the implementation of this PR. It no more caches urlClasspaths, but it caches inputClassLoader instead. 

> The input classloaded must be cached, because otherwise two classes loaded by two different (not cached) input classloaders **are not compatible**. 

See the new `CompilationTest.testSingleClassLoader()` for details.
As a side effect of this change the `urlClasspath()` is called only once during creation of inputClassLoader, so the performance improvement is same like with origin solution of this PR.

The call of setSourceClasspath resets the inputClassLoader cache so the new class loader is created when needed first time.

monperrus:
thanks!

Issue #826, opened 2016-09-17
DefaultJavaPrettyPrinter.visitCtThisAccess


leventov:
I think it shouldn't rely on getTarget().isImplicit() when printing this qualifier, because classes could be moved around, and Implicit flag is usually not modified.

Such implementation work better for me, probably it should be adapted for DefaultJavaPrettyPrinter itself (+ corner cases, which I omitted in my code because not the case for me):

``` java
    class MyPrinter extends DefaultJavaPrettyPrinter {

        Deque<CtClass<?>> currentThis = new ArrayDeque<>();

        public MyPrinter(Environment env) {
            super(env);
        }

        @Override
        public <T> void visitCtClass(CtClass<T> ctClass) {
            currentThis.push(ctClass);
            super.visitCtClass(ctClass);
            currentThis.pop();
        }

        @Override
        public <T> void visitCtThisAccess(CtThisAccess<T> thisAccess) {
            enterCtExpression(thisAccess);
            if (!thisAccess.isImplicit()) {
                if (thisAccess.getType().getDeclaration() != currentThis.peek()) {
                    visitCtTypeReferenceWithoutGenerics(thisAccess.getType());
                    write(".");
                }
                write("this");
            }
            exitCtExpression(thisAccess);
        }
    }
```

monperrus:
if it's a bug, could you provide us with a failing test case? thanks.

monperrus:
Hi @leventov , any news on this? is this really a bug?

leventov:
@monperrus probably I can pay some time for this, as well as other issues I've reported, next weekend.

monperrus:
OK, thanks.

monperrus:
ping @leventov 
do you work on this? 
(to know whether we wait for your issues for the next release)

leventov:
@monperrus please don't wait for issues opened by me.

PR #991, opened 2016-11-18
test for thisAccess like this.field and not SomeClass.this.field

pvojtechovsky:
I guess the issue #826 created by @leventov is about the problem, which is reproducable by 2 new tests of this PR

The problem, which I feel too is that in the source class like
```java
package spoon.test.fieldaccesses.testclasses;
public class Pizza {
	int size;
	void setSize(int size) {
		this.size = size;
	}
}
```

The spoon produces field access `this.size` in these two wrong forms 

if `factory.getEnvironment().setAutoImports(true)` then it is
> Pizza.this.size = size

if `factory.getEnvironment().setAutoImports(false)` then it is
> spoon.test.fieldaccesses.testclasses.Pizza.this.size = size

But I would prefer to print `this.size` in all cases.

### DIFF
--- a/src/test/java/spoon/test/fieldaccesses/FieldAccessTest.java
+++ b/src/test/java/spoon/test/fieldaccesses/FieldAccessTest.java
@@ -23,6 +23,7 @@ import spoon.reflect.declaration.CtParameter;
 import spoon.reflect.declaration.CtType;
 import spoon.reflect.factory.Factory;
 import spoon.reflect.reference.CtFieldReference;
+import spoon.reflect.reference.CtParameterReference;
 import spoon.reflect.visitor.CtScanner;
 import spoon.reflect.visitor.DefaultJavaPrettyPrinter;
 import spoon.reflect.visitor.Query;
@@ -413,5 +414,19 @@ public class FieldAccessTest {
 		factory.getEnvironment().setAutoImports(true);
 
 		assertEquals("this.size = size", pizza.getElements(new TypeFilter<>(CtAssignment.class)).get(0).toString());
+	}
+
+	@Test
+	public void testThisDotFieldAccessAutoNotImplicit() throws Exception {
+		Factory factory = build(Pizza.class);
+		CtClass<?> pizza = factory.Class().get(Pizza.class);
+		factory.getEnvironment().setAutoImports(true);
+
+		CtMethod<?> m = pizza.getMethod("addSize", factory.Type().INTEGER_PRIMITIVE);
+		CtParameter<?> param = m.getParameters().get(0);
+		param.setSimpleName("size");
+		m.getElements((CtParameterReference<?> e)-> "plus".equals(e.getSimpleName())).forEach((CtParameterReference<?> e)->e.setSimpleName("size"));
+		//the printer detects the name conflict between field name and parameter name and should use explicit thisAccess automatically  
+		assertEquals("this.size = (this.size) + size", m.getElements(new TypeFilter<>(CtAssignment.class)).get(0).toString());
 	}
 }
--- a/src/test/java/spoon/test/fieldaccesses/FieldAccessTest.java
+++ b/src/test/java/spoon/test/fieldaccesses/FieldAccessTest.java
@@ -3,6 +3,7 @@ package spoon.test.fieldaccesses;
 import org.junit.Test;
 import spoon.Launcher;
 import spoon.reflect.code.CtArrayWrite;
+import spoon.reflect.code.CtAssignment;
 import spoon.reflect.code.CtExpression;
 import spoon.reflect.code.CtFieldAccess;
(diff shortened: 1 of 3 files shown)


pvojtechovsky:
And I have added another test, then one which should automatically use thisAccess even if the field was implicit before the refactoring.

Idea of the test. Let's have method `addSize` in class like:
```java
public class Pizza {
	int size;
	void addSize(int plus) {
		size = size + plus;
	}
}
```
The field `size` uses implicit access - OK.
But then if I apply refactoring - rename parameter "plus" to "size" ... and spoon prints invalid result
```java
	void addSize(int size) {
		size = size + size;
	}
```
but it should produce `this.size = this.size + size;'

So I agree with @leventov, that printer should not care about CtFieldAccess.isImplicit, but should print optimized value.
... I see that it will be not simple task, because printer needs to have correct information about context where field access is printed.

tdurieux:
Hi, I wrote a possible solution here: https://github.com/tdurieux/spoon/commit/52d75a7a4518490ada9b4239cfc760d78b78ce96
I don't know if I handled every case (probably no).

pvojtechovsky:
I tried your solution on my project and it produce compilation errors in following situation
```java
public abstract class A {
  public void  methodOnA() {}
}
public class B extends A {
  public void methodOnB() {
    methodOnA();
  }
}
```
because it produced
```java
public void methodOnB() {
  A.this.methodOnA();
}

```

pvojtechovsky:
Anyway thank You for investigating this problem! I will need solution for that probably next year - in the case, when all pass and I let spoon refactor sources of my project. More I understand spoon more I see how much work is in it! ... and how complicated task it is to analyze/refactor sources.

tdurieux:
Yes I know, it does not work at all. It's much more complex that anticipated.

pvojtechovsky:
Looking at `AccessibleVariablesFinder`, I always wonder how powerful are these visitors. I am not get used to use visitors in my work. I wonder that it is quite fast too, even if it looks for variable for each this access...

If I would have time to do it, I would try opposite approach. To make PrintingContext more clever then it is now. The PrintingContext is called (push/pop) with each CtElement. So it knows the context the best. I guess the algorithm, which would resolve accessible variables, should be based on the PrintingContext and helper data collected there. But it is just my feeling, I have not tried it yet or thought about that deeper.

tdurieux:
> I guess the algorithm, which would resolve accessible variables, should be based on the PrintingContext and helper data collected there. But it is just my feeling, I have not tried it yet or thought about that deeper.

It does not work when you extends classes from external libraries.

Example
```java
// class from external library
public External() {
  int a = 0;
}


public class Parent () {
  int a = 0;
  private class Child() extends External {
    public void m () {
       Parent.this.a++; // with your technique the qualified name will be removed and java will access a from External not from Parent
    }
  }
}
```

I update my code, now I succeed to compile Spoon https://github.com/tdurieux/spoon/commit/e6aed5c412f5b681641b10e86d4971e062a32b1d

pvojtechovsky:
It is really a challenge  :-) 
... I have never wrote such code, but I see that spoon must handle all such tricks...

tdurieux:
my code is not compatible with the no-classpath mode.
Nevertheless is only a POC, but seams possible when you have a valid classpath.

pvojtechovsky:
I just checked the ClassNotFoundExceptions and the problem is not in the class path. I am not using noClassPath mode.

That new algorithm fails on each nested class like
```java
package a.b.c;
public class A {
protected static class B {}
}
```
because it tries to load class with name "B", but it should load class "a.b.c.A$B"

tdurieux:
If I have some time I will try to fix to fix.

Bye the way I updated my branch and now it supports invocation.

tdurieux:
> I just checked the ClassNotFoundExceptions and the problem is not in the class path. I am not using noClassPath mode.

I did not succeed to reproduce your issue. The qualified name of your example is correct.

pvojtechovsky:
I have committed here the test which reproduces the ClassNotFoundException on invalid qualified name of nested type.

Run the FieldAccessTest.testFieldAccessDeclaredInADefaultClass() in debug and put breakpoint into 
CtTypeReferenceImpl.isSubtypeOf ... 
Launcher.LOGGER.error("cannot determine runtime types for 

The type.getQualifiedName() is "Fails", but should be spoon.test.fieldaccesses.testclasses.internal.Foo$Fails

tdurieux:
Ok, thanks.

With my current code I cannot reproduce the bug because I remove the isSubtypeOf call from the jdttreebuilder.
Maybe the qualified name was incorrect because the tree was not complete.
I will investigate.

monperrus:
fixed by #1031?

pvojtechovsky:
Yes, #1031 is fixing this issue

monperrus:
closed by #1031

PR #997, opened 2016-11-22
fix(equal): actual type refs must be part of equality

monperrus:
this fixes a very old and deep bug, which was enforced in certain existing assertions.

but now we have the expected:

```
CtLocalVariable var = factory.Code().createCodeSnippetStatement("java.util.List<String> l ").compile();
CtLocalVariable var2 = factory.Code().createCodeSnippetStatement("java.util.List<Object> l ").compile();
assertNotEquals(var2, var); // the generics are taken into account
```
(this holds for any actual type parameter).


monperrus:
@tdurieux ready for merging.

PR #999, opened 2016-11-22
refactor(printerContext): assure correct restore of state

pvojtechovsky:
The new PrintingContext assures that boolean states like [noTypeDecl, ignoreGenerics, ...] are always set to origin value.


monperrus:
Very nice to use try with resources.

OK for me. 

Since we're at the core of Spoon, @tdurieux, what 's your opinion ?

tdurieux:
It is nice, but I am not sure if that increases the readability of the class.

pvojtechovsky:
I agree that readability is worst then 
```java
context.flag=true;
...
context.flag=false;
```
but, my intention was not make it more readable but to make it more reliable. With the new code (if reader understands it) you can be sure that origin state is correctly restored. It was not the case before. There was a place for a mistake ... and may be in some wild class/nested class/lambda/... combination it would even fail actually... nobody know.

monperrus:
I agree, there is a trade-off between readability and reliability.

PR #1004, opened 2016-11-24
add CtTypeMember.getTopLevelType()

pvojtechovsky:
I have added new method 
```java
interface CtTypeMember
	/**
	 * Returns top level declaring type of this. 
         * If this is already a TopLevelType, then it returns this
	 */
	<T> CtType<T> getTopLevelType();
```
I need it for next PR, which brings some new Refactoring methods for changing of the name of the CtVariable based elements.

Please give me feedback, if such method is acceptable. If yes, then I will add test case too.

Note: I also moved some getDeclaringType to more common place in the hierarchy of CtExecutableImpl, so I did not had to copy the implementation of getTopLevelType 3 times.

### DIFF
--- a/src/main/java/spoon/reflect/declaration/CtTypeMember.java
+++ b/src/main/java/spoon/reflect/declaration/CtTypeMember.java
@@ -30,6 +30,11 @@ public interface CtTypeMember extends CtModifiable {
 	 * @return declaring class
 	 */
 	@DerivedProperty
-	CtType<?> getDeclaringType();
+	<T> CtType<T> getDeclaringType();
 
+	/**
+	 * Returns top level declaring type of this. If this is already a TopLevelType, then it returns this
+	 */
+	@DerivedProperty
+	<T> CtType<T> getTopLevelType();
 }
--- a/src/main/java/spoon/reflect/declaration/CtTypeMember.java
+++ b/src/main/java/spoon/reflect/declaration/CtTypeMember.java
@@ -30,10 +30,11 @@ public interface CtTypeMember extends CtModifiable {
 	 * @return declaring class
 	 */
 	@DerivedProperty
-	<T> CtType<T> getDeclaringType();
+	CtType<?> getDeclaringType();
 
 	/**
-	 * Returns top level declaring type of this. If this is already a TopLevelType, then it returns this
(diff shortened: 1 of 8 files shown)


src/main/java/spoon/reflect/declaration/CtTypeMember.java
```diff
@@ -30,6 +30,11 @@
 	 * @return declaring class
 	 */
 	@DerivedProperty
-	CtType<?> getDeclaringType();
+	<T> CtType<T> getDeclaringType();
```
monperrus: not backward-compatible. do you really need to add generics?

src/main/java/spoon/reflect/declaration/CtTypeMember.java
```diff
@@ -30,6 +30,11 @@
...
 
+	/**
+	 * Returns top level declaring type of this. If this is already a TopLevelType, then it returns this
```
monperrus: Returns the top level type declaring this type if an inner type. If this is already a top-level type, then returns itself.

src/main/java/spoon/reflect/declaration/CtTypeMember.java
```diff
@@ -30,6 +30,11 @@
...
+	 */
+	@DerivedProperty
+	<T> CtType<T> getTopLevelType();
```
monperrus: makes a lot of sense. it is OK to me.

src/main/java/spoon/reflect/declaration/CtTypeMember.java
pvojtechovsky: I needed that to solve another compilation problem in CtConstructorImpl ... but I will try different solution

retuxx:
I'm looking for exactly the same feature.

monperrus:
excellent. Could you add tests?

retuxx:
Is it possible to add a similar method that returns the parent type instead of the top level type?

pvojtechovsky:
> Is it possible to add a similar method that returns the parent type instead of the top level type?

there is `getDeclaringType()` for that, is not?

tdurieux:
or `getParent(CtType.Class)`

pvojtechovsky:
Please merge it if it pass. I need this method in #1007

PR #1005, opened 2016-11-24
review 1: Change local variable name refactoring

pvojtechovsky:
This code is not tested and may be even wrong somewhere. I will work on it, because I will need such refactoring in my project.

Just have a short look at class `Refactoring`, please. Whether you want such methods in Spoon or it is out of it's scope.

What about checking of **validity** of such refactoring requests? For example if I have 
> class A and B in package P and I decide to rename class A to B

then the class B will be twice in the model. Do you have some preferences how to report such invalid requests?
A) to throw exception
B) to call some listener and silently skip the operation
C) ?
I am just sure that in all cases the validity should be checked before the model is modified.


src/main/java/spoon/refactoring/Refactoring.java
```diff
@@ -36,15 +49,115 @@
...
+		changeSimpleName(type, name, null, new TypeFilter<CtTypeReference<?>>(CtTypeReference.class) {
 			@Override
 			public boolean matches(CtTypeReference<?> reference) {
-				return type.getQualifiedName().equals(reference.getQualifiedName());
+				return oldQualifiedName.equals(reference.getQualifiedName());
 			}
 		});
+	}
+
+	public static void changeVariableName(CtVariable<?> variable, String name) {
```
monperrus: javadoc

monperrus:
> Just have a short look at class Refactoring, please. 

that's a very useful contribution

> Do you have some preferences how to report such invalid requests?

an exception.

we also need strong tests.

pvojtechovsky:
> we also need strong tests.

I like the idea of usage of complex thirdparty library (like guava), with good jUnit tests, as a test environment.

The test would look like this:
1) rename all local variables to v1, v2, ... vN. 
2) rename all method parameters to v1,v2,...vN
3) rename all fields to to v1,v2,...vN
4) rename all classes to v1,v2...vN
The renaming should detect conflict if name vX is already used in current context and throw exception, then test will try name v(X+1) as long as v(X+Y) is usable name.
Then we print the refactored model and run the refactored tests on it too. If it pass, then it is OK
... but may be it is too complex task, so it will never pass :-) ... we will see

I came with this idea, hoping that I can avoid writing small jUnit test for each refactoring feature. But now I see these small tests are useful when detecting where is the problem in the code.  

So the best will be combination of small tests at beginning and at the end that thirparty library refactoring test, which founds the most tricky things.

monperrus:
> So the best will be combination of small tests at beginning and at the
> end that thirparty library refactoring test, which founds the most
> tricky things.
>
I agree. For the complex ones, we run them through `build.sh`.

pvojtechovsky:
The refactoring request validity checks are missing.

These checks can be quite tricky. For example rename of local variable should check
* if there is another local variable with same name defined in visibility scope of that local variable
* if there is method parameter with same name in current method
* if there is catch parameter with same name in current catch block

Do you think that we should continue with refactoring functions, which are implemented as static methods of one class? I have feeling like it would be better to have one class for each refactoring. The API might be like this:
```java
interface CtRefactoring {
  boolean isValid();
  void refactor() throws SpoonInvalidRequest;
}
```
And sometime there are probably some refactorings which cannot be done without temporary inconsistency in the model, so then we should be able to switch OFF the validation check.

I have implemented nothing yet, so I welcome your suggestions. It is easy to accept any design if there is no code yet. ;-)

monperrus:
> Do you think that we should continue with refactoring functions, which are implemented as static
> methods of one class?
Spoon's design has always been very much OO, So I really support the introduction of OO for refactoring.

pvojtechovsky:
It is POC of Refactoring class, which does rename of local variable.
Usage:
```java
CtLocalVariable target = ...
ChangeLocalVariableName rename = new ChangeLocalVariableName();
rename.setTarget(target);
rename.setNewName("someName");
if(rename.getIssues().size()>0) {
 ... handle conflicts, etc...
}
rename.refactor();
```
I will add fluent API or a constructor of course. But please give me feedback mainly to
* `Refactor` interface
* `Issue` interface
* `VariableScopeQuery` and `VariableReferenceQuery` - the nearly finished queries, which can be used for many purposes - they will be used by other future refactorings too ...
* anything else, what seems to be relevant from your point of view

monperrus:
Hi Pavel, 

This is very interesting but this already contains a lot of different things to be discussed, they may be split in several PRs (hence in several separate discussions, one after the other).

According to your long-term plan, what do you prefer to discuss and merge first? maybe `VariableScopeQuery`?

pvojtechovsky:
Hi Martin,

I am aware that there is a lot of things, many baby steps, some of them may be in wrong direction... I develop it this way as a prototype to see what code/structures I need to implement refactoring algorithms in short, efficient and maintainable way. And note, I do not need VariableScopeQuery or VariableReferenceQuery. I need refactorings, which are based on these queries. So I started with refactorings and as a side effect there are VariableScopeQuery and VariableReferenceQuery.

> first topic for discussion: `VariableScopeQuery`?

It is good topic for discussion, because it is one of the core components of that refactorings. So yes, let's discuss it first. I am just not sure if it makes sense to make a PR only for that, because VariableScopeQuery is not much usable on it's own. The better is combination with VariableReferenceQuery. These two are tightly coupled. If you agree then we can move discussion to #1114, which contains only these two queries.

monperrus:
>  If you agree then we can move discussion to #1114, which contains only these two queries.

OK

pvojtechovsky:
During implementation of local variable rename refactoring, class `ChangeLocalVariableName`, I have found that I need to extend CtQuery. I need:
* to know when query `enter`/`exit` an element during filterChildren on any query step level 
* to be able to skip processing of children of current element 
* to be able to skip processing of siblings and children of current elements

I will prepare another PR for that soon. Do you have any preference/ideas how to do it? You can influence the solution before it is implemented - so less effort is needed.

If you want to see context where I need it, then have a look at `ChangeLocalVariableName#detectNameConflicts` of this PR. There are comments which explains the situation.
This PR depends on PR #1154, so look there for classes, which are missing here.

monperrus:
Sounds interesting.


> Do you have any preference/ideas how to do it?
not really. I look forward to your solution for this.

pvojtechovsky:
I am testing ChangeLocalVariableName refactoring method using ChangeVariableNameTest. OK, there are still some todos, but I stuck on one thing. The testing algorithm is like this:
1. take one method from `VariableRename` class and search for `TryRename` annotation. 
2. For each found annotation try to rename assigned variable to all new names defined in `TryRename` annotation. The prefix "-" in the annotation, means that rename to this name should be not possible.
3. If the rename is possible and passes then the test does following extra steps (see `ChangeVariableNameTest#printModelAndTestConsistency`)
4. print the refactored model
5. compile refactored model
6. create new class loader and load the class from the refactored model
7. create instance of that refactored class = call constructor, whose implementation will run all the methods of that class and will check that all assertions are still passing.

The problem is that step 7) does not work. I have no idea why class loader loads some other implementation of the class ...
... it must be so, because otherwise the assertions should fail, because the rename of "var1" to "var2" in this method
```java
	void nestedClassMethodWithShadowVar() {
		@TryRename({"var2", "var3"})
		int var1 = 2; //<--------- the "var1" is renamed to "var2"
		new Runnable() {
			@TryRename({"var1", "var3"})
			int var2 = 3;
			@Override
			public void run() {
				@TryRename({"-var1", "var2"})
				int var3 = 1;
				assertTrue(var1 == 2); //<-----------this reference is renamed too
				//this var1 shadows above defined var1. It can be renamed
				@TryRename({"var2", "-var3"})
				int var1 = 4;
				assertTrue(var1 == 4);
				assertTrue(var2 == 3);
				assertTrue(var3 == 1);
			}
		}.run();
		assertTrue(var1 == 2);
	}
```
by mistake passes (and it should not pass), and pretty printer produces this valid java code:
```java
    void nestedClassMethodWithShadowVar() {
        @spoon.test.refactoring.testclasses.TryRename(value = { "var2" , "var3" })
        int var2 = 2; //<!----------- is renamed to "var2" - OK
        new java.lang.Runnable() {
            @spoon.test.refactoring.testclasses.TryRename(value = { "var1" , "var3" })
            int var2 = 3;

            @java.lang.Override
            public void run() {
                @spoon.test.refactoring.testclasses.TryRename(value = { "-var1" , "var2" })
                int var3 = 1;
                org.junit.Assert.assertTrue((var2 == 2)); //<-------- is renamed to var2, 
//but the value should be 3, because it is no more local variable reference, but it is now the reference to the class field, 
//which shadows origin local variable! So it should fail here!!!
                @spoon.test.refactoring.testclasses.TryRename(value = { "var2" , "-var3" })
                int var1 = 4;
                org.junit.Assert.assertTrue((var1 == 4));
                org.junit.Assert.assertTrue(((var2) == 3));
                org.junit.Assert.assertTrue((var3 == 1));
            }
        }.run();
        org.junit.Assert.assertTrue((var2 == 2));
    }
```
but execution of that code does not fail on AssertionError, but it should fail :-(

Any idea why my test class loader does not load classes compiled from refactored code?

tdurieux:
I could be wrong (I did not read all your changes).
It seams that you compile the original files not the model : https://github.com/INRIA/spoon/pull/1005/files#diff-617f2b7a84783b10a9bd2ed77f0fe739R135

pvojtechovsky:
@tdurieux thanks! It really compiled origin not refactored sources. Now the test fails as expected :-)

This PR is still not finished. There are several bugs ... working on it.

pvojtechovsky:
Looks like my local variable rename tests passed. Now we have to pass through all the changes to have it in acceptable form in main branch.
I will make some smaller PRs from this one later ...

pvojtechovsky:
This is compilable after rebase on #1180, #1193, #1194. It means that these changes are needed, but are not part of this PR, so it is easier to review this new feature.

This PR contains working code, which can safely refactor name of local variables. It detects name conflicts and shadowing of variables. In such case the refactoring throws exception, before invalid change would be done, so the model stays consistent.

I would like to discuss here these topics:
* name refactoring object: `ChangeLocalVariableName` ... it needs better name
* before the refactoring is processed it detects whether requested change is possible (will not cause change in behavior or compilation errors). If problem is found then it's description is added to collection of issues. `List<Issue> Refactor#getIssues()`. The question is whether we will collect such issues and what issue types we will support?
* any other improvements You see.

monperrus:
rebase to have a cleaner diff?

pvojtechovsky:
Here you are :-)

src/test/java/spoon/test/refactoring/ChangeVariableNameTest.java
```diff
@@ -0,0 +1,244 @@
...
+		launcher.buildModel();
+		factory = launcher.getFactory();
+		varRenameClass = factory.Class().get(VariableRename.class);
+		tryRename = factory.createCtTypeReference(TryRename.class);
+		local1Var = varRenameClass.filterChildren((CtLocalVariable<?> var)->var.getSimpleName().equals("local1")).first();
+	}
+
+	@Test
+	public void testModelConsistency() throws Throwable {
+		new VariableRename();
```
monperrus: does this have side effects?

src/test/java/spoon/test/refactoring/ChangeVariableNameTest.java
```diff
@@ -0,0 +1,244 @@
...
+			return parentMethod.getParent(CtType.class).getSimpleName()+"#"+parentMethod.getSimpleName();
+		} else {
+			return ele.getParent(CtType.class).getSimpleName()+"#annonymous block";
+		}
+	}
+	
+
+	@Test
+	public void testRefactorWithoutTarget() throws Exception {
+
```
monperrus: contract: a name refactoring where the given variable name does not exist results in a SpoonException

src/test/java/spoon/test/refactoring/ChangeVariableNameTest.java
```diff
@@ -0,0 +1,244 @@
...
+		
+		ChangeLocalVariableName refactor = new ChangeLocalVariableName();
+		refactor.setTarget(local1Var);
+		refactor.setNewName("local1");
+		refactor.refactor();
+		assertEquals("local1", local1Var.getSimpleName());
+	}
+	
+	@Test
+	public void testRenameLocalVariableToInvalidName() throws Exception {
```
monperrus: merge all exception tests (incl testRefactorWithoutTarget) in one single test

src/test/java/spoon/test/refactoring/testclasses/VariableRename.java
```diff
@@ -0,0 +1,244 @@
...
+
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+import java.util.function.Consumer;
+import java.util.function.Function;
+
+import static org.junit.Assert.*;
+
+public class VariableRename
```
monperrus: better name? such as `RenameTestSubject`?

src/test/java/spoon/test/refactoring/testclasses/VariableRename.java
```diff
@@ -0,0 +1,244 @@
...
+{
+	public VariableRename() throws Throwable
+	{
+		int local1 = 0;
+		//call all not private methods of this class automatically, to check assertions
```
monperrus: interesting. would be event better in a separate clear method. 
new RenameTestSubject().callAllMethods()

src/test/java/spoon/test/refactoring/testclasses/TryRename.java
```diff
@@ -0,0 +1,16 @@
+package spoon.test.refactoring.testclasses;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.LOCAL_VARIABLE, ElementType.PARAMETER, ElementType.FIELD})
+public @interface TryRename {
```
monperrus: I really like the idea of having an annotation for testing, this is beautifully declarative.

may we should have `Test` in its name for more clarity.

src/test/java/spoon/test/refactoring/ChangeVariableNameTest.java
```diff
@@ -0,0 +1,244 @@
...
+		} catch(SpoonException e) {
+			
+		}
+	}
+
+	@Test
+	public void testRenameLocalVariableToSameName() throws Exception {
+		
+		ChangeLocalVariableName refactor = new ChangeLocalVariableName();
+		refactor.setTarget(local1Var);
```
monperrus: `local1var`: what is this?

tests that are self-contained  and that can be understood by just looking at the method (without scrolling) are easier to understand an maintain in the long term.

src/test/java/spoon/test/refactoring/ChangeVariableNameTest.java
```diff
@@ -0,0 +1,244 @@
...
+	@Test
+	public void testModelConsistency() throws Throwable {
+		new VariableRename();
+	}
+	
+	String[] DEBUG = new String[]{/*"nestedClassMethodWithoutRefs", "var3", "var1"*/};
+
+	@Test
+	public void testRenameLocalVariableToUsedName() throws Exception {
+
```
monperrus: what's the contract tested here?

src/main/java/spoon/refactoring/Issue.java
```diff
@@ -0,0 +1,21 @@
...
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the CeCILL-C License for more details.
+ *
+ * The fact that you are presently reading this means that you have had
+ * knowledge of the CeCILL-C license and that you accept its terms.
+ */
+package spoon.refactoring;
+
+public interface Issue {
```
monperrus: rename to `RefactoringIssue`? If it's an issue, what about using a `RefactoringException` directly?

src/main/java/spoon/refactoring/Refactor.java
```diff
@@ -0,0 +1,27 @@
...
+ * knowledge of the CeCILL-C license and that you accept its terms.
+ */
+package spoon.refactoring;
+
+import java.util.List;
+
+/**
+ *
+ */
+public interface Refactor {
```
monperrus: interface and methods must be documented. Can you give a usage snippet? I don't really see here what can be done.

src/main/java/spoon/refactoring/AbstractRenameRefactor.java
```diff
@@ -0,0 +1,111 @@
...
+		forEachReference(new CtConsumer<CtReference>() {
+			@Override
+			public void accept(CtReference t) {
+				t.setSimpleName(AbstractRenameRefactor.this.newName);
+			}
+		});
+		target.setSimpleName(newName);
+	}
+
+	protected abstract void forEachReference(CtConsumer<CtReference> consumer);
```
monperrus: javadoc? contract for the implementors?

src/test/java/spoon/test/refactoring/ChangeVariableNameTest.java
pvojtechovsky: It has no side effects, but it throws an exception if the tested model is inconsistent. It is similar to `VariableReferencesModelTest`, where I first tested model consistency by call of constructor, which internally called all other methods of model to check if they contain consistent assertions. Then you came with good idea to switch it to jUnit test class too. So `VariableRename` is very similar to that, but in this case I cannot easily switch it to jUnit test class, because I am testing model consistency before and after each model refactoring test action.

src/test/java/spoon/test/refactoring/ChangeVariableNameTest.java
pvojtechovsky: it is the main test method (...bad named...), which uses spoon model of `VariableRename` class. It looks for each CtVariable and it's CtAnnotation and tries to rename that variable to the name defined by annotation. 
If the annotation name is prefixed with "-", then that refactoring should fail.
If the annotation name is not prefixed, then that refactoring should pass.

src/main/java/spoon/refactoring/Issue.java
pvojtechovsky: Do you mean to throw exception directly instead of collecting of issues and then throw an exception? 
I would like to support use case where list of issues is better then one exception on the first issue.
The idea is like this:
1) create an refactoring
2) call Refactoring#getIssues()
3) if there are some issues, then solve them automatically, by applying other refactorings
4) call Refactoring#getIssues() again
5) if there are still some issues then skip this refactoring
6) if there are no issues, then process refactoring.

pvojtechovsky: But it is no problem to remove issues for now and add them later with some mature inheritance hierarchy of issues. So KISS wins here.

src/test/java/spoon/test/refactoring/testclasses/VariableRename.java
pvojtechovsky: done

pvojtechovsky:
Your ides/change suggestions were implemented. Please review

src/main/java/spoon/refactoring/Refactor.java
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Defines basic contract of all refactoring implementations.<br>
+ * Usage:<br>
+ * <pre>
+ * SomeRefactoring r = new SomeRefactoring();
+ * //configure refactoring by calling setters on `r`
+ * r.refactor();
+ * </pre>
+ */
+public interface Refactor {
```
monperrus: rename to Refactoring?

add method `setTarget(CtElement)` to really show in the interface that one always refactors something?

src/main/java/spoon/refactoring/AbstractRenameRefactor.java
```diff
@@ -0,0 +1,87 @@
...
+ * knowledge of the CeCILL-C license and that you accept its terms.
+ */
+package spoon.refactoring;
+
+import java.util.regex.Pattern;
+
+import spoon.SpoonException;
+import spoon.reflect.declaration.CtNamedElement;
+
+public abstract class AbstractRenameRefactor<T extends CtNamedElement> implements Refactor {
```
monperrus: currently we have only one subclass. what other subclasses do you envision?

src/test/java/spoon/test/refactoring/RenameLocalVariableRefactorTest.java
```diff
@@ -0,0 +1,255 @@
...
+		//rollback changes
+		refactor.setNewName(originName);
+		try {
+			refactor.refactor();
+		} catch(SpoonException e) {
+			throw new AssertionError(getParentMethodName(targetVariable)+" Rename of \""+originName+"\" to \""+newName+"\" passed, but rename back to \""+originName+"\" failed", e);
+		}
+	}
+
+	private void printModelAndTestConsistency(String refactoringDescription) {
```
monperrus: is an assert-like method, it should be named accordingly `assert*`, eg `assertCorrectModel`

src/test/java/spoon/test/refactoring/RenameLocalVariableRefactorTest.java
```diff
@@ -0,0 +1,255 @@
...
+
+	/**
+	 * The {@link RenameLocalVariableRefactorTestSubject} class is loaded as spoon model. Then:
+	 * - It looks for each CtVariable and it's CtAnnotation and tries to rename that variable to the name defined by annotation.
+	 * - If the annotation name is prefixed with "-", then that refactoring should fail.
+	 * - If the annotation name is not prefixed, then that refactoring should pass.
+	 * If it behaves different then expected, then this test fails
+	 */
+	@Test
+	public void testRenameAllLocalVariablesOfRenameTestSubject() throws Exception {
```
monperrus: this test is super powerful, it's beautiful.

However, I would suggest to add before a simpler test, which does one simple refactoring, it would show a small and clear snippet on how to use  `RenameLocalVariableRefactor`, and it would help to understand this one.

monperrus: by the way, it's in essence a parametrized test, where RenameLocalVariableRefactorTestSubject contains the test parameters. Really clever. It would be even more explicit and maintainable written as  [`@Parametrized`](https://github.com/junit-team/junit4/wiki/parameterized-tests).

src/main/java/spoon/refactoring/RenameLocalVariableRefactor.java
```diff
@@ -0,0 +1,266 @@
...
+ * refactor.setTarget(anLocalVariable);
+ * refactor.setNewName("someNewName");
+ * try {
+ *   refactor.refactor();
+ * } catch (RefactoringException e) {
+ *   //handle name conflict or name shadowing problem
+ * }
+ * </pre>
+ */
+public class RenameLocalVariableRefactor extends AbstractRenameRefactor<CtLocalVariable<?>> {
```
monperrus: add a static method in Refactoring (eg `Refactoring#renameLocalVariable(CtNamedElement target, String newName)`) to help discoverability and usage of this class?

src/main/java/spoon/refactoring/Refactor.java
pvojtechovsky: > rename to Refactoring

but there is already class `spoon.refactoring.Refactoring`, so ?

> add method setTarget(CtElement)

Does all possible refactorings accepts exactly one CtElement? Does all possible refactoring have input called `target`? I am not sure. I vote to let `setTarget` method defined on level of more specific refactoring class

src/main/java/spoon/refactoring/AbstractRenameRefactor.java
pvojtechovsky: `RenameTypeRefactor`, `RenameParameterRefactor`, `RenameFieldRefactor`, `RenameMethodRefactor`, ...

src/test/java/spoon/test/refactoring/RenameLocalVariableRefactorTest.java
pvojtechovsky: Do you have experience with parametrized test? Me not. I can learn it, but if you would have time to rewrite this test to parametrized, I might focus on resolving of type parameters. ;-)

src/main/java/spoon/refactoring/RenameLocalVariableRefactor.java
pvojtechovsky: good idea. I will do so tomorrow evening.

monperrus:
> but there is already class |spoon.refactoring.Refactoring|, so ?
>
right. Then "CtRefactoring"?

monperrus:
> Does all possible refactorings accepts exactly one CtElement?
correct,  setTarget(CtElement...) may be a better option.

> I vote to let |setTarget| method defined on level of more specific refactoring class
The problem with that is that the interface cannot be understood without knowing concrete
implementations, which is an issue

monperrus:
> Do you have experience with parametrized test? Me not. I can learn it, but if you would have time
> to rewrite this test to parametrized
No problem, we keep it like this. I wanted to point to an interesting thing...

pvojtechovsky:
> CtRefactor#setTarget

I would prefer to keep CtRefactor as it is. I can imagine refactoring methods, which has no `target` at all. And how would rename refactoring work on array of target elements?

> The problem with that is that the interface cannot be understood without knowing concrete
implementations, which is an issue

But `CtRefactor` is very abstract. It is more like marker interface, which collects all the future refactoring implementations. I guess the only think which can be "understood" here is that "client can run refactoring  after s/he configures it by refactoring specific methods"

I understand that Spoon is interface based, so I have added new interface, where setTarget fits better. What about this one?
```java
public interface CtRenameRefactoring<T extends CtNamedElement> extends CtRefactoring {
	CtRenameRefactoring<T> setTarget(T target);
	CtRenameRefactoring<T> setNewName(String newName);
}
```

> add Refactoring#renameLocalVariable ...

I have added 
```java
public static void changeLocalVariableName(CtLocalVariable<?> localVariable, String name)
```
whose name and parameters fits more to the existing method
```java
public static void changeTypeName(final CtType<?> type, String name)
```

What about rename of `RenameLocalVariableRefactor` to 
A) `CtRenameLocalVariableRefactor`
B) `CtRenameLocalVariableRefactoring'
because it is based on `CtRefactoring` interface now.

src/main/java/spoon/refactoring/Refactoring.java
```diff
@@ -52,4 +53,17 @@ public boolean matches(CtTypeReference<?> reference) {
...
+	 * Changes name of a {@link CtLocalVariable}.
+	 *
+	 * @param localVariable
+	 * 		to be renamed {@link CtLocalVariable} in the AST.
+	 * @param name
+	 * 		New name of the element.
+	 * @throws RefactoringException when rename to newName would cause model inconsistency, like ambiguity, shadowing of other variables, etc.
+	 */
+	public static void changeLocalVariableName(CtLocalVariable<?> localVariable, String name) throws RefactoringException {
+		new RenameLocalVariableRefactor().setTarget(localVariable).setNewName(name).refactor();
```
monperrus: beautiful line

monperrus:
> I have added
>
> public static void changeLocalVariableName(CtLocalVariable<?> localVariable, String name)
thanks, rename parameter name to newName?

> I guess the only think which can be "understood" here is that "client can run refactoring after
> s/he configures it by refactoring specific methods"
> I would prefer to keep CtRefactoring as it is. 
I'm not convinced, I prefer strong and consistent interfaces, but it's no reason to argue about this
for now :-)

> What about rename of |RenameLocalVariableRefactor| to
> A) |CtRenameLocalVariableRefactor|
> B) |CtRenameLocalVariableRefactoring|

|Good idea, ||CtRenameLocalVariableRefactoring is perfect. ||Let's move |AbstractRenameRefactor
<https://github.com/INRIA/spoon/pull/1005/files#diff-4ae10ecea7772cfbf5caccd21ff3b831>to
AbstractRenameRefactoring also

monperrus:
thanks!

pvojtechovsky:
You are welcome ;-)

PR #1007, opened 2016-11-25
fix typed parameters in access path

### LINKED ISSUE #1000: regression on fully-qualified names
When one spoon https://github.com/spoon-bot/guava with the latest snapshot, `AbstractMapBasedMultimap.java` is not correctly rewritten.

     /builds/workspace/Guava/guava/target/generated-sources/spoon/com/google/common/collect/AbstractMapBasedMultimap.java:[656,16] error: AbstractMapBasedMultimap.WrappedList.WrappedListIterator is not abstract and does not override abstract method next() in ListIterator

@pvojtechovsky it's maybe due to your work on access path.

pvojtechovsky:
fixes #1000

### DIFF
--- a/src/main/java/spoon/reflect/reference/CtTypeReference.java
+++ b/src/main/java/spoon/reflect/reference/CtTypeReference.java
@@ -153,7 +153,8 @@ public interface CtTypeReference<T> extends CtReference, CtActualTypeContainer, 
 	CtTypeReference<?> getTopLevelType();
 
 	/**
-	 * Computes nearest access path parent from contextType to this type.
+	 * Computes nearest access path parent to this type from the context of this type reference.
+	 * The context is defined by this.getParent(CtType.class).
 	 *
 	 * Normally the declaring type can be used as access path. For example in this class hierarchy
 	 * <pre>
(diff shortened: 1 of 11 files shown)


src/main/java/spoon/reflect/factory/TypeFactory.java
```diff
@@ -324,6 +324,9 @@ public TypeFactory() {
 		}
 
 		ref.setSimpleName(type.getSimpleName());
+		for (CtTypeParameter typeParam : type.getFormalCtTypeParameters()) {
```
monperrus: why is this change required?

pvojtechovsky: I have already found, that this change is wrong. I will rollback it and do it different.

pvojtechovsky:
This implementation is still not 100% correct, but 
- it is fitting more to origin concept of @tdurieux, because CtTypeReference.toString() now returns access path and not qualified name - same like in his origin concept! I restored some test cases to the state before my accessPath implementation.
- all tests pass and guava seems to compile now too.

The remaining problem is how to correctly map actual type arguments. See the comment in `CtTypeReferenceImpl.applyActualTypeArguments`.
I do not want to implement that algorithm now. I vote for merging it as it is and to fix it after somebody writes code which will really need that. ... the origin problem was very rare and the case when this implementation fails should be even more rare :-)

monperrus:
it looks good to me, thanks. @tdurieux OK for you?

> CtTypeReference.toString() now returns access path and not qualified name - same like in his origin concept!

this is fine, in general `toString` on references is not as important as `toString` on code elements, it's more for debug purposes, so the specification is more malleable.

tdurieux:
@monperrus it is good for me

PR #1008, opened 2016-11-26
fix(ThisAccess): the type access associated to a this must be the type containing the this

### LINKED ISSUE #1006: Issue when cloning methods that contain invocations to j.l.Object methods
Dear all,

When cloning methods that contain invocations to j.l.Object methods such as:

public void methodUsingjlObjectMethods() {
    notify();
}

Spoons returns a cloned method that, when pretty-printed reads like:

public void methodUsingjlObjectMethods() {
    java.lang.Object.this.notify();
}

The issue is that this code does not compile, and outputs the error "No enclosing instance of type Object is accessible in scope" for "java.lang.Object.this".

This issue is illustrated with the following (quickly written) test:

https://github.com/seintur/spoon/blob/42c597dade6795a6b48c99aa41a200b3058a7c69/src/test/java/spoon/test/method/MethodTest.java#L58

The issue is certainly related to recent commits since this is a new one that just appeared recently in the past days.

Cheers,
Lionel.

monperrus:
fixes #1006.

Note that #1006 was neither in clone, nor in pretty-printer, but in the model.

This is the problem when implicit=true, the model can be incorrect and one does not see it.

### DIFF
--- a/src/main/java/spoon/reflect/factory/CodeFactory.java
+++ b/src/main/java/spoon/reflect/factory/CodeFactory.java
@@ -107,7 +107,13 @@ public class CodeFactory extends SubFactory {
 	 * @return a accessed type expression.
 	 */
 	public <T> CtTypeAccess<T> createTypeAccess(CtTypeReference<T> accessedType) {
-		return createTypeAccessWithoutCloningReference(accessedType == null ? null : accessedType.clone());
+		if (accessedType == null) {
+			return factory.Core().createTypeAccess();
+		}
+		CtTypeReference<T> access = accessedType.clone();
+		// a type access doesn't contain actual type parameters
+		access.setActualTypeArguments(null);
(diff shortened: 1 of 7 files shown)


src/test/java/spoon/test/method/MethodTest.java
```diff
@@ -85,4 +90,26 @@ public void testGetAllMethods() throws Exception {
 		assertEquals(1, methods.stream().filter(method -> "foo".equals(method.getSimpleName())).count());
 	}
 
+	@Test
+	public void testCloneMethodUsingjlObjectMethods() throws Exception {
```
tdurieux: What does  the **jl** mean in this test name? testCloneMethodUsing**jl**ObjectMethods

seintur: Correct :-)

monperrus:
fixed test name

PR #1013, opened 2016-11-28
deprecate reference filtering

pvojtechovsky:
I implemented all ReferenceFilters using Filters and marked them as deprecated. The code is shorter now and it is clear that ReferenceFilter is obsolete.


monperrus:
looks OK to me. nobody against this deprecation?

src/main/java/spoon/reflect/visitor/filter/AbstractFilter.java
```diff
@@ -31,6 +31,7 @@
 	/**
 	 * Creates a filter with the type of the potentially matching elements.
 	 */
+	// TODO: INFER TYPE BY INTROSPECTION
```
retuxx: Will this todo be fixed soon?

pvojtechovsky: I just copied that, but I would also like to have to simpler. Could somebody add a comment which will show how the calling code will look like after it is fixed?

PR #1015, opened 2016-11-29
Type of filter by reflection

pvojtechovsky:
The type of the Filter does not have to be specified in constructor, but is computed by reflection from the matches parameter type.

This PR is based on #1013, so I have to rebase it after you merge #1013. Just have a look if code like this OK, or we can make it even simpler?
```java
List<CtNewClass<?>> elements = foo.getElements(new AbstractFilter<CtNewClass<?>>() {
	@Override
	public boolean matches(CtNewClass<?> element) {
		return true;
	}
});
```

### DIFF
--- a/src/main/java/spoon/reflect/visitor/QueryVisitor.java
+++ b/src/main/java/spoon/reflect/visitor/QueryVisitor.java
@@ -35,14 +35,15 @@ public class QueryVisitor<T extends CtElement> extends CtScanner {
 	/**
 	 * Constructs a query visitor with a given filter.
 	 */
+	@SuppressWarnings("unchecked")
 	public QueryVisitor(Filter<T> filter) {
 		super();
 		this.filter = filter;
 		if (filter instanceof AbstractFilter) {
-			filteredType =  ((AbstractFilter) filter).getType();
+			filteredType =  ((AbstractFilter<T>) filter).getType();
 		} else {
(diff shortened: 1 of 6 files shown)


monperrus:
you can go ahead with rebasing!

pvojtechovsky:
You can already write it using java 8 lambda like this:
```java
List<CtExpression<?>> expressions2 = foo.getElements((CtExpression<?> e)->true);

expressions2.forEach(it->{
	assertTrue(it instanceof CtExpression);
});
```
but it will not infer type correctly. It will fail with ClassCastException in call of matches. But it is expected and handled as not matches. I guess this ClassCastException has a performance impact...

May be in Java 9 the parameter type of lambda can be get by reflection.

monperrus:
interesting. could you add tests to check that the inferred `filteredType` is correct?

pvojtechovsky:
Can I replace all the calls of `new AbstractFilter(CLASS)`, by `new AbstractFilter()` in all tests and all spoon code? It would be good test that it works. And it would shorten code again...

Should I will keep `new AbstractFilter(CLASS)` for compatibility?

pvojtechovsky:
OK, I will add these tests probably tomorrow in the evening. Good night ;)

pvojtechovsky:
Tests were added

PR #1017, opened 2016-11-29
functional ElementConsumer in queries

pvojtechovsky:
Support for queries like this:
```java
anElement.forEachElement(anFilter, (e)->...do something with filtered element e...)
```
I believe that such filtering approach is more memory efficient and less code has to be written, because **most of the code** which uses
```java 
List result = anElement.getElements(filter)
```
is followed by a `for(...){...}` cycle. 

But again it can be used in nice way only if java 8 is used by client code


pvojtechovsky:
I am closing this PR. The part of this PR will be added to PR #1018, which will contain new QueryBuilder with support of ElementConsumer

PR #1018, opened 2016-11-30
feature: new query architecture with support for Java8 queries

pvojtechovsky:
Requirements for a new query architecture:

* [x] map functions with arbitrary return types (incl. that are not CtElement)
* [x] support for queries with Java 8 lambdas
* [x] chained queries ( "chaining" means writing `filterChildren(x).filterChildren(y)...` or `map(z).map(t)...` or even `filterChildren(p).map(q)...`)
* [x] lazy evaluation

----------------------------------------------------

With this PR you can write code like
```java
CtMethod<?> method = ...
List<CtMethod<?>> overridenMethods = Query.getElements(new OverriddenMethodFilter(method));
```
Which makes usage of complex filters more reliable, because these filters can automatically define correct searching context. For example 
- `OverriddenMethodFilter` will search in root package
- `local variable reference` filter will search in scope of its CtBock element
- `member variable reference` filter will search in scope of its Class and depending on visibility in all inherited class, in current package or in root package
- ...

or you can write
```java
CtMethod<?> method = ...
List<CtMethod<?>> overridenMethods = method.getElements(new OverriddenMethodFilter())
```
which 1) is shortest code and 2) which will let us chain filters (later) into complex filters chains, which are able to browse spoon model in sequence of searching steps. I will probably need something like this for refactoring methods, I am preparing.


tdurieux:
Nice. :+1: 

One day we will able to write modern code that uses spoon :D

monperrus:
Excellent indeed! Could you add a presentation of this new feature in `doc/filter.md`?

pvojtechovsky:
> documentation

ok, but I would first like to finish real chaining of filters using a new QueryBuilder and PR #1017. The combination of these two things will start to be interesting. I will then document it together.

pvojtechovsky:
Documentation added.

tdurieux:
Do you know that you can use java 8 features in the tests?

pvojtechovsky:
> java 8 features in the tests

yes, I found it when I added spoon into Eclipse. The spoon code is java 7 and tests are java 8 compatible

pvojtechovsky:
I have found one problem of this PR. It is not backward compatible with existing filters. For example
this code of `FilterTest.testFieldAccessFilter()`:
```java
CtFieldReference<?> ref = ...
List<CtFieldAccess<?>> expressions = someClass.getElements(new FieldAccessFilter(ref));
```
will behave different when we improve FieldAccessFilter by way it implements `ScopeAwareFilter` and `ChainableFilter`. Because then the correct call is:
```java
CtFieldReference<?> ref = ...
List<CtFieldAccess<?>> expressions = ref.getElements(new FieldAccessFilter());
```

How to solve that?
A) to not use CtElement#getElements, but new one like `CtElement#query` or `searchFor` or ...?
B) to keep old filters as they are and to implement new filters ... 👎 
C) ? any ideas ?

By the way, Travis failed. Should I change something?

pvojtechovsky:
I have rolled back CtElement#getElements, because it is not backward compatible. 

I would actually prefer solution like this:
```java
CtFieldReference<?> ref = ...
List<CtFieldAccess<?>> expressions = ref.search(new FieldAccessFilter()).list();
//or optionally this should be supported too
ref.search(new FieldAccessFilter()).forEach(CtFieldAccess<?>fa->{...});
```
Note that `CtElement#search(Filter)` method would return new QueryBuilder object, which can be used to build even more complex queries using fluent API. When query is built, then there are two ways how to start the query
A) `QueryBuilder#list()` which processes query and returns List of found elements
B) `QueryBuilder#forEach(ElementConsumer consumer)` which processes query and calls consumer for each found element.

If you like this approach, then it is question how to make a PR for that?
A) add QueryBuilder into this PR
B) you will merge this PR as it is and I will make new PR with QueryBuilder

pvojtechovsky:
I have added QueryBuilder to this PR. The QueryBuilder is now used in Query API too, so it is tested by all test code which uses Queries.

The documentation does not contain QueryBuilder yet. I will update it probably tomorrow.

This old code
```java
CtFieldReference<?> ref = ...
List<CtFieldAccess<?>> expressions = someClass.getElements(new FieldAccessFilter(ref));
```
can be written this way:
```java
CtFieldReference<?> ref = ...
List<CtFieldAccess<?>> expressions = ref.query(new FieldAccessFilter()).list();
```
after `FieldAccessFilter` will implement `ChainableFilter` in future PRs.

pvojtechovsky:
This new code needs further refactoring and comments, test, etc. But the idea is like this:
Query builder connects ElementPipes into chain, which can be used process several queries on the model and to collect several data from these nodes. See this example
```java
	class Context {
		CtMethod method;
		CtMethod overridenMethod;
	}
		
	Context context = new Context();

	launcher.getFactory().Package().getRootPackage().query(new TypeFilter(CtMethod.class))
		.then(new ElementPipeImpl<CtMethod,CtMethod>(){
			@Override
			protected void onAccept(CtMethod element) {
				context.method = element;
			}
		}).then(new OverriddenMethodFilter())
		.then(new ElementPipeImpl<CtMethod,CtMethod>(){
			@Override
			protected void onAccept(CtMethod element) {
				context.overridenMethod = element;
                                ...
			}
		}).list();
```
it searches for all methods and then for all overriden methods of found methods. The Context class is used just as example how client can collect data from the middle of the query processing.

pvojtechovsky:
This PR now contains support for query step chain. I implemented it again since beginning, so please forget olders concepts and check if they are still used in this PR or not.

See the first draft of documentation in doc/Filter.md.

The code is not finished yet.
- it needs to be commented
- I don't like `CtElement#query()` but how to do it better?
- write jUnit test - the code is really not tested yet

Please review the names of the new Classes, packages and methods. Let me know if something is not understandable or confusing. Do you have a better name for AsyncFunction?
Are the `QueryStep` build methods `then`, `scan` and `matches` understandable?
Any other feedback is welcome too.

pvojtechovsky:
I am still not finished, but query like this works now. Do you know what it does??
```java
		launcher.getFactory().Package().getRootPackage().query().scan((CtClass<?> c)->{return true;})
			.then((CtClass<?> c)->c.getSuperInterfaces())
			.then((CtTypeReference<?> iface)->iface.getTypeDeclaration())
			.then((CtType<?> iface)->iface.getAllMethods())
			.then((CtMethod<?> method)->method.getSimpleName().equals("make"))
			.then((CtMethod<?> m)->m.getType())
			.then((CtTypeReference<?> t)->t.getTypeDeclaration())
			.forEach((CtInterface<?> c)->{
				assertEquals("ITostada", c.getSimpleName());
			});
```
It scans AST from root package, accepts CtClasses, then it navigates to all super interfaces of found classes and searches there for each method whose name is equal to "make". If an method is found then it takes its return type and checks that it has found expected value.

I have removed QueryStep#matches. It actually uses `then(Function)` and the QueryStep detects that returned value is Boolean and in such case behaves as predicate.

src/main/java/spoon/reflect/visitor/chain/AsyncFunction.java
```diff
@@ -0,0 +1,25 @@
...
+ * The fact that you are presently reading this means that you have had
+ * knowledge of the CeCILL-C license and that you accept its terms.
+ */
+package spoon.reflect.visitor.chain;
+
+/**
+ * @param <T> the type of the input to the function
+ * @param <R> the type of the result of the function
+ */
+public interface AsyncFunction<T, R> {
```
monperrus: name and javadoc not clear enough to convey the core design intuition and contract

monperrus: why async? executed asynchronously? or do you mean deferred?

src/main/java/spoon/reflect/visitor/chain/AsyncFunctionQueryStep.java
```diff
@@ -0,0 +1,39 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE. See the CeCILL-C License for more details.
+ *
+ * The fact that you are presently reading this means that you have had
+ * knowledge of the CeCILL-C license and that you accept its terms.
+ */
+package spoon.reflect.visitor.chain;
+
+import spoon.support.util.SafeInvoker;
+
+public class AsyncFunctionQueryStep<O> extends QueryStep<O> {
```
monperrus: javadoc?

src/main/java/spoon/reflect/visitor/chain/Function.java
```diff
@@ -0,0 +1,25 @@
...
+ * The fact that you are presently reading this means that you have had
+ * knowledge of the CeCILL-C license and that you accept its terms.
+ */
+package spoon.reflect.visitor.chain;
+
+/**
+ * @param <T> the type of the input to the function
+ * @param <R> the type of the result of the function
+ */
+public interface Function<T, R> {
```
monperrus: clear

src/main/java/spoon/reflect/visitor/chain/Predicate.java
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * The fact that you are presently reading this means that you have had
+ * knowledge of the CeCILL-C license and that you accept its terms.
+ */
+package spoon.reflect.visitor.chain;
+
+/**
+ * @param <T> the type of the input to the function
+ */
+public interface Predicate<T> {
```
monperrus: the same as Filter? (but the generic T?)

src/main/java/spoon/reflect/visitor/chain/QueryStep.java
```diff
@@ -0,0 +1,167 @@
...
+ * A) scan(Filter) scan all child elements of input element and send to output only these elements, which matches the filter
+ * B) then(AsyncFnc) initialize filter with input element and then scan all children of the start element returned by filter and send to output only these elements, which matches the filter
+ * C) scan all children of the start element returned by filter and send to output only these elements, which matches the filter - ignore input
+ * D) matches(Predicate) - send input to output if it matches filter
+ *
+ *
+ * @param <I> - input type
+ * @param <O> - output type
+ */
+public abstract class QueryStep<O> implements Consumer<Object> {
```
monperrus: what about having an interface `QueryStep`? (in Spoon, the design is very interface-oriented: most key concepts and key methods are defined in interfaces). The method `then` seems indeed essential for the design.

src/main/java/spoon/reflect/visitor/filter/Scann.java
```diff
@@ -0,0 +1,42 @@
...
+ * knowledge of the CeCILL-C license and that you accept its terms.
+ */
+package spoon.reflect.visitor.filter;
+
+import spoon.reflect.declaration.CtElement;
+import spoon.reflect.visitor.CtScanner;
+import spoon.reflect.visitor.chain.AsyncFunction;
+import spoon.reflect.visitor.chain.Consumer;
+
+public class Scann extends CtScanner implements AsyncFunction<CtElement, CtElement> {
```
monperrus: what does `Scann` mean?

src/main/java/spoon/support/util/RtHelper.java
```diff
@@ -192,15 +192,15 @@ private static void getAllIMethods(Class<?> c, List<Method> methods) {
 	 * @param clazz
 	 * @param methodName
 	 * @param numParams
-	 * @return the types of the parameters of such method or null if not found
+	 * @return the found method or null
 	 */
-	public static Class<?>[] getMethodParameterTypes(Class<?> clazz, String methodName, int numParams) {
+	public static Method getMethod(Class<?> clazz, String methodName, int numParams) {
```
monperrus: non backward-compatible change on public method

src/main/java/spoon/support/util/SafeInvoker.java
```diff
@@ -0,0 +1,151 @@
...
+package spoon.support.util;
+
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+
+import spoon.Launcher;
+import spoon.SpoonException;
+
+/**
+ * @param <T> the Type of the delegate object, which is target of method invocation
```
monperrus: contract? responsibility?

src/main/java/spoon/reflect/visitor/chain/AsyncFunction.java
pvojtechovsky: I know that Async is not good name for that. Deferred has a time meaning too, but it is better then Async. The purpose of this function is to produce zero, one or more elements from one input element. May be names like
- ChainableFunction
- CallbackFunction
- ProductFunction
- ...
I like no one :-(

src/main/java/spoon/reflect/visitor/chain/Predicate.java
pvojtechovsky: Yes generic.
1) QueryStep chain can be used to process elements which are not inherited from CtElement. I am not sure whether we needs that.
2) Spoon `Filter` has already old meaning like Filtering of children produced by CtScanner, I wanted something more generic, what can be used to decide whether input element should be sent to output or not.

src/main/java/spoon/reflect/visitor/filter/Scann.java
pvojtechovsky: It can be used in the QueryStep chain to visit all the children of chain input element.
It is used automatically In combination with filter in method `QuertStep#scan(Filter)`, but I guess it can be use independent using ...query().then(new Scann())....

src/main/java/spoon/support/util/RtHelper.java
monperrus: OK if it's a recent one.

src/main/java/spoon/reflect/visitor/chain/Predicate.java
monperrus: OK, could you put this explanation in the Javadoc?

src/main/java/spoon/reflect/visitor/chain/AsyncFunction.java
monperrus: ChainableFunction is better IMHO

src/test/java/spoon/test/filters/FilterTest.java
```diff
@@ -332,6 +335,8 @@ public void testOverriddenMethodFromAbstractClass() throws Exception {
 		final CtClass<AbstractTostada> aClass = launcher.getFactory().Class().get(AbstractTostada.class);
 
 		assertEquals(0, Query.getElements(launcher.getFactory(), new OverriddenMethodFilter(aClass.getMethodsByName("prepare").get(0))).size());
+		assertEquals(0, Query.query().then(new OverriddenMethodFilter()).list(aClass.getMethodsByName("prepare").get(0)).size());
```
monperrus: hard to understand to start with a call to "then" with no element.

src/test/java/spoon/test/filters/FilterTest.java
```diff
@@ -466,4 +475,78 @@ public boolean matches(CtInvocation<?> element) {
...
+		
+		class Context {
+			int counter = 0;
+		}
+		Context context = new Context();
+		
+		QueryStep<CtClass<?>> l_qv = launcher.getFactory().getModel().getRootPackage().query().scan(new TypeFilter<>(CtClass.class));
+		
+		assertEquals(0, context.counter);
+		l_qv.forEach(cls->{
```
monperrus: I like the `foreach` very much

src/test/java/spoon/test/filters/FilterTest.java
```diff
@@ -466,4 +475,78 @@ public boolean matches(CtInvocation<?> element) {
...
+		class Context {
+			CtMethod<?> method;
+			int count = 0;
+		}
+		
+		Context context = new Context();
+
+		launcher.getFactory().Package().getRootPackage().query().scan(new TypeFilter<CtMethod<?>>(CtMethod.class))
+		.thenConsume((CtMethod<?> method) -> {context.method = method;})
+		.then(new OverriddenMethodFilter())
```
monperrus: what you call `then` is a functional `map` (https://en.wikipedia.org/wiki/Map_(higher-order_function))?

src/test/java/spoon/test/filters/FilterTest.java
```diff
@@ -466,4 +475,78 @@ public boolean matches(CtInvocation<?> element) {
...
+		
+		class Context {
+			CtMethod<?> method;
+			int count = 0;
+		}
+		
+		Context context = new Context();
+
+		launcher.getFactory().Package().getRootPackage().query().scan(new TypeFilter<CtMethod<?>>(CtMethod.class))
+		.thenConsume((CtMethod<?> method) -> {context.method = method;})
```
monperrus: what's the difference between `then` and `thenConsume`?

src/test/java/spoon/test/filters/FilterTest.java
pvojtechovsky: > Query.query().then(...)

yes, I do not like it too. But it was minimum code at the beginning. If you have better idea, then tell me please. Otherwise I suggest to wait little bit until I (or somebody else) start to use it and come with real experience and needs... I am also looking for a better way ...

src/test/java/spoon/test/filters/FilterTest.java
pvojtechovsky: I would like to see `foreach` more often in Spoon API - even in core model ... I will make other PR. But later ...

src/test/java/spoon/test/filters/FilterTest.java
pvojtechovsky: The problem is that we must visibly distinguish between 
A) then(...a function which produces elements for next step...)
B) thenConsume(... a function which just eats elements and produces nothing....)

We cannot let this distinction on the java compiler, because it would be confusing. Example:
`(element)->element.doSomething()`
Is that Consumer, which produces nothing or is it a function which produces result? Sometime the client wants to call function which returns values as consumer too.

If you write code like 
`...query().then(X).then(Y).then(Z)`
then it processes X and the result of X is sent to Y and result of **Y** is sent to Z

But if you write code like 
`...query().then(X).thenConsume(Y).then(Z)`
then it processes X and the result of X is sent to Y, which produces nothing and result of **X** is sent to Z

I am again open to change function name to anything better. May be I should learn some RxJava API and take over the names from that. Is there anybody who knows that API and can suggest names which are generally used and understandable?

src/test/java/spoon/test/filters/FilterTest.java
pvojtechovsky: Thanks! I need suggestions and links like that! I do not have theoretic training in these things

monperrus:
> A) then(...a function which produces elements for next step...)
> B) thenConsume(... a function which just eats elements and produces
> nothing....)
A) map and B) forEach (or "iter")?

> But if you write code like
> |...query().then(X).thenConsume(Y).then(Z)|
> then it processes X and the result of X is sent to Y, which produces
> nothing and result of *X* is sent to Z
That's quite counter-intuitive.to me. Do we need/want this behavior?

pvojtechovsky:
I will replace `then` by `map`.

> thenConsume(Consumer) ... That's quite counter-intuitive.to me. Do we need/want this behavior?

I wanted that to be able to collect the intermediate results of the query (model browsing). But it can be replaced by 
```java
query().then((element)->{
   ...do required consuming...
  return element;
})
```
so we have one query "command" (`thenConsume`) less => simpler API and it is easy to get same behavior with existing `then` (`map`) command. I will remove it. Thanks for feedback! I want to have it simple and intuitive too.

> thenConsume ... forEach (or "iter")?

`thenConsume(Consumer)` was intended for use in the middle of the query chain. The `QueryStep#forEach(Consumer)` is already there. The difference to `thenConsume(Consumer)` is that `forEach` **executes** the query, while `thenConsume` only **add** a QueryStep, but do not execute the query.

> forEach or iter

I prefer `forEach`, because there is java 8 `Iterable#forEach(Consumer)`. So I think this name will be intuitive for Java 8 developers. Or do you think it is something else and we should use different name?

pvojtechovsky:
... still not ready for merge. I have pushed it here to show the current state and to see whether it pass the tests.

monperrus:
OK

pvojtechovsky:
I have commented all the new code and did final simplifications and renames. I am satisfied with that version, but if you have any suggestions, or I forgot something then let me know please.

The test cases are still missing.  Please make a review of new API so I can finally document it in Filter.md. I will do that probably on Saturday evening.

src/main/java/spoon/reflect/declaration/CtElement.java
```diff
@@ -35,7 +36,7 @@
  * element).
  */
 @Root
-public interface CtElement extends FactoryAccessor, CtVisitable, Cloneable {
+public interface CtElement extends FactoryAccessor, CtVisitable, Cloneable, QueryComposer {
```
monperrus: why this?

src/main/java/spoon/reflect/visitor/chain/QueryComposer.java
```diff
@@ -0,0 +1,76 @@
...
+
+import spoon.reflect.declaration.CtElement;
+import spoon.reflect.visitor.Filter;
+
+/**
+ * QueryComposer contains methods, which can be used to create/compose a {@link QueryStep}
+ * It is implemented 1) by {@link CtElement} to allow creation new query and its first step
+ * 2) by {@link QueryStep} to allow creation of next query step
+ */
+public interface QueryComposer {
```
monperrus: I'm not sure about the name, what do you think of `QueryResult`?

src/main/java/spoon/reflect/visitor/chain/QueryStep.java
```diff
@@ -0,0 +1,86 @@
...
+ * </ul>
+ * The query can be used several times.<br>
+ * But QueryStep is not thread safe. So you must create new query for each thread.<br>
+ * Usually the new query is created each time when you need to query something.
+ * The reusing of QueryStep instance makes sense when the same query has to be evaluated
+ * several times in the loop.
+ *
+ * @param <O> the type of the element produced by this QueryStep
+ */
+public interface QueryStep<O> extends QueryComposer, Consumer<Object> {
```
monperrus: The role and methods of QueryStep and QueryComposer seem close. Does it make sense to merge them in a single interface?

src/main/java/spoon/reflect/declaration/CtElement.java
pvojtechovsky: QueryComposer contains all the methods, which can begin with creation of some query. It is the replacement for `CtElement#query()`. With QueryComposer you can easily start to query spoon model starting from any element. 
Note: QueryComposer is implemented by QueryStep too.

src/main/java/spoon/reflect/visitor/chain/QueryStep.java
pvojtechovsky: I think no. QueryComposer has only methods, which can **create** query step. But QueryStep has methods which can **evaluate** query too. It makes no sense to evaluate query on CtElement (which implements QueryComposer) directly, because there is no query yet. First you have to create one or more query steps (using QueryComposer) and then you can evaluate the query using methods, which are only on QueryStep only.

src/main/java/spoon/reflect/visitor/chain/ChainableFunction.java
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Functional interface used to map elements using a function which accepts one input parameter and calls output.accept(out) for each produced output element
+ *
+ * It is used for example by {@link QueryStep#map(ChainableFunction))}
+ *
+ * @param <T> the type of the input to the function
+ * @param <R> the type of the result produced by this function
+ */
+public interface ChainableFunction<T, R> {
+	void apply(T input, Consumer<R> output);
```
pvojtechovsky: It is not possible. It is different contract. Have a look at new class
src/main/java/spoon/reflect/visitor/filter/Scann.java
It implements ChainableFunction. How you would implement such thing using `Consumer<R> apply(T input)` ?

src/main/java/spoon/reflect/visitor/chain/QueryComposer.java
```diff
@@ -0,0 +1,76 @@
...
+
+	/**
+	 * Appends queryStep to the current (last) step of the query
+	 * and returns queryStep, which was entered. It is now the last query step
+	 *
+	 * @param queryStep
+	 * @return the entered queryStep, which is now the last step of the query
+	 */
+	<R> QueryStep<R> map(QueryStep<R> queryStep);
```
monperrus: all map functions are really close (same name, same return type). It suggests a common abstraction (and the proximity of Function and ChainableFunction goes in the same direction).

Is it possible to have a single map function with an appropriate common interface as parameter?

monperrus:
I'd propose to first converge on the design of the core new interface `QueryComposer`

src/main/java/spoon/reflect/visitor/chain/QueryComposer.java
```diff
@@ -0,0 +1,76 @@
...
+	/**
+	 * scan all child elements of an input element.
+	 * The child element is sent to next step only if filter.matches(element)==true
+	 *
+	 * Note: the input element is also checked for match and if true it is sent to output too.
+	 *
+	 * @param filter used to filter scanned children elements of AST tree
+	 * @return the created QueryStep, which is now the last step of the query
+	 */
+	<T extends CtElement> QueryStep<T> scan(Filter<T> filter);
```
pvojtechovsky: Yes these two codes does exactly the same
```java
List filteredChildren = anElement.getElements(anFilter)
```
```java
List filteredChildren = anElement.scan(anFilter).list()
```
The `getElements` is needed just for backward compatibility. We can mark it as Deprecated.

> would an alternative be to move getElements in QueryStartingPoint

I do not understand. there is no `QueryStartingPoint`. Did you mean `StartQueryStep`? If yes, how that would help with backward compatibility?

src/main/java/spoon/reflect/visitor/chain/QueryComposer.java
pvojtechovsky: > Is it possible to have a single map function with an appropriate common interface as parameter?

I do not know how it would look like. Please describe your idea in more details.

Why three functions are problem?

I need all three functions for these cases:

1) map(QueryStep) - it let's us to combine query chains together

2) map(Function) - it let's us to make a query based on standard functions of CtModel elements. It simply gets the input element and returned value is sent to next step. See javadoc description for details about 4 possible usages.

3) map(ChainableFunction) it let's us implement clever Filters, which automatically computes the scanning scope. For example FieldReferencesFilter, which searches for references to an field depending on field visibility.

Note: we cannot merge map(Function) and map(ChainableFunction) into one. Because Function#apply has 1 input parameter, while ChainableFunction#apply has 2 parameters. And if I would make one Interface, which would contains both `apply` functions, then it would be not possible to implement it using java 8 lambda expression, because it expects only one non default method in the interface.

src/main/java/spoon/reflect/visitor/chain/QueryComposer.java
pvojtechovsky: It is there to compose first or next queryStep. It is really not about result/evaluation of the query.

If you do not like QueryComposer, then alternatives might be
- QueryCreator
- QueryStepBuilder
- QueryFactory
- ... ?

pvojtechovsky:
I agree, QueryComposer is the core. It is easy to make it better now. So I am open to all ideas, which makes it more understandable and usable :-)

monperrus:
> I agree, QueryComposer is the core

Excellent. 

* [name] As new name, What do you think of "QueryStartingPoint" or "Queryable" or "CtQueryable"? (because a ctelement is a query starting point). Core contract: allow to query and filter from an element or a list of elements.
 * [scan] AFAI, scan is a `getElements()` which returns a queryable element (as opposed to a list). If correct, I agree we need a new method (and we keep  `getElements()` non-deprecated), but what about returning a queryable type instead, this would be more intuitive and readable?
* [map] interesting, and see comment above

src/main/java/spoon/reflect/visitor/chain/ChainableFunction.java
monperrus: Looking at Scann, basically Consumer is the implicit Function. What if Consumer extends Function (accept becomes apply) and then you only have one method map?

pvojtechovsky:
I will rename QueryComposer to CtQueryable

> getElements(Filter) might return QueryStep

I like this idea, but it is not backward compatible. May be it is no problem, because it will be easy to find and fix that incompatibility by appending .list(). Should I do it like this?

pvojtechovsky:
I have removed `CtQueryable#map(QueryStep)` so we have now only two map functions
1) map(Function) - Function has one input and one output parameter
2) map(ChainableFunction) - ChainableFunction has two input parameters and no output

This change was possible because `QueryStep extends ChainableFunction` now.

We can also remove QueryVisitor, which is not used in Spoon. I have marked it as Deprecated for now.

TODOs:
- to found better name for CtQueryable#scan(Filter)
- to update Filter.md after API is stable
- to convince Martin that 2 `map` functions are now good design and enough elegant  ;-)

> What if Consumer extends Function (accept becomes apply) 

I tried this change. It includes change of `void` with lower case "v"
```java 
void accept(O)
```
to `Void` with upper case "V"
```java
Void apply(O)
```
which causes that java compiler required `return null;` at the end of each consuming lambda expression. So that change forces clients of spoon to write longer useless code. I think you agree that is really not elegant.

I personally like these 2 map functions, but if you are still looking for something different then I see these (not so elegant) ways:

1) to get rid of `map(Function)` and to map everything using `map(ChainableFunction)` for example like this:
```java
anElement.map((CtClass clazz, Consumer out)->out.accept(clazz.getSuperClass()))
```

2) to kill performance and elegance by collecting all the elements, which would be produced by ChainableFunction into a List and return that list as result from Function

3) to replace ChainableFunction#apply(input, Consumer output)` by new interface `OutputAwareFunction` with two methods. The code would look like this
```java
class OutputAwareFunction<I,O> extends Consumer<I> {
  @override
  void apply(I input);
  /**
  * contract: the client of ChainableFunction MUST always use code like this:
  * ChainableFunction cf = ...
  * cf.setOutputConsumer(consumer);
  * cf.apply(input);
  */
  void setOutputConsumer(Consumer<O> consumer);
}
```
This solution has these drawbacks:
1) we still need 2 `map` methods, because if we extend ChainableFunction from Function<O,Void>, then client must write always `return null;`
2) Such interface cannot be implemented by Lambda expression, because it contains two functions. But may be it is not a problem, because algorithms, which needs a ChainableFunction seems to be usually complex so they will be not implemented by Lambda expression, but by a normal class.
3) The implementations of the `OutputAwareFunction` will be more tricky then implementation of `ChainableFunction`. For example try to rewrite `OverriddenMethodFilter#apply` which implements `ChainableFunction` 
```java
public void apply(CtMethod<?> input, Consumer<CtMethod<?>> output) {
	method = input;
	method.getFactory().Package().getRootPackage().scan(this).forEach(output);
}
```
to `OutputAwareFunction`...
... please not this way ...

Note: After this PR is merged, I will implement more filters similar to OverriddenMethodFilter, which help me write the refactoring methods in an elegant way. So note that example of `OverriddenMethodFilter` is not just minor case, but it is the MAIN reason why I need this PR.

pvojtechovsky:
May be we should also rename 
- Function to CtFunction
- ChainableFunction to CtChainableFunction
- Consumer to CtConsumer

otherwise spoon Consumer and Function might be by developers misunderstood as java 8 Consumer and Function.
Or optionally we should upgrade to java 8 and replace spoon Consumer and Function by java 8 Consumer and Function. I am not sure if it would be good step even if we would already use java 8.

pvojtechovsky:
The SafeInvoker has bad performance. I will replace it by `try{}catch(ClassCastException)`
I will also try idea how to simplify QueryStepImpl.
But it will have no influence to methods of `CtQueryable`. I need them all.

src/main/java/spoon/reflect/declaration/CtElement.java
```diff
@@ -35,7 +36,7 @@
  * element).
  */
 @Root
-public interface CtElement extends FactoryAccessor, CtVisitable, Cloneable {
+public interface CtElement extends FactoryAccessor, CtVisitable, Cloneable, CtQueryable {
```
monperrus: looks good

src/main/java/spoon/reflect/visitor/chain/CtQueryable.java
```diff
@@ -0,0 +1,67 @@
...
+
+import spoon.reflect.declaration.CtElement;
+import spoon.reflect.visitor.Filter;
+
+/**
+ * QueryComposer contains methods, which can be used to create/compose a {@link QueryStep}
+ * It is implemented 1) by {@link CtElement} to allow creation new query and its first step
+ * 2) by {@link QueryStep} to allow creation of next query step
+ */
+public interface CtQueryable {
```
monperrus: we now have a good name!

src/main/java/spoon/reflect/visitor/chain/CtQueryable.java
```diff
@@ -0,0 +1,67 @@
...
+	/**
+	 * scan all child elements of an input element.
+	 * The child element is sent to next step only if filter.matches(element)==true
+	 *
+	 * Note: the input element is also checked for match and if true it is sent to output too.
+	 *
+	 * @param filter used to filter scanned children elements of AST tree
+	 * @return the created QueryStep, which is now the last step of the query
+	 */
+	<T extends CtElement> QueryStep<T> scan(Filter<T> filter);
```
monperrus: let's call it "query"?

src/main/java/spoon/reflect/visitor/chain/CtQueryable.java
```diff
@@ -0,0 +1,67 @@
...
+	 * <tr><td>{@link Boolean}<td>Sends input to the next step if returned value is true
+	 * <tr><td>{@link Iterable}<td>Sends each item of Iterable to the next step
+	 * <tr><td>{@link Object[]}<td>Sends each item of Array to the next step
+	 * <tr><td>? extends {@link Object}<td>Sends returned value to the next step
+	 * </table><br>
+	 *
+	 * @param code a Function with one parameter of type I returning value of type R
+	 * @return the create QueryStep, which is now the last step of the query
+	 */
+	<I, R> QueryStep<R> map(Function<I, R> code);
```
monperrus: Looks good. and I agree let's go for `CtFunction`

pvojtechovsky:
> I'd propose to first converge and merge a first design without chainable/outputaware functions

I would do it, if it would be possible. But it is not possible without killing the MAIN PURPOSE of this PR: `OverriddenMethodFilter` which automatically selects the start scanning point. I do not persist on OverriddenMethodFilter`, but `OverriddenMethodFilter` is the good (very simple) example of other clever filters, which I need.

And even if I would rollback changes in `OverriddenMethodFilter` together with removing `ChainableFunction`, I would have to implement QueryStep#scan in more complicated and not nice way.

Please copy this branch and show me how to implement `OverriddenMethodFilter` without `ChainableFunction`. I am out of ideas.

pvojtechovsky:
>  ChainableFunction ... it seems an additional feature. 

It is not additional feature. It is core feature. The `CtFunction` is the additional feature.

What about renaming like this. 
- QueryStep -> CtQuery
- QueryStepImpl -> CtQueryImpl
- ChainableFunction#apply -> CtQueryStep#query
- CtQueryable#scan -> CtQueryable#filterChildren

With the names above it is more visible that ChainableFunction/CtQueryStep is the core contract of the CtQuery.

monperrus:
> What about renaming like this.
>
>   * QueryStep -> CtQuery
>   * QueryStepImpl -> CtQueryImpl
>   * ChainableFunction#apply -> CtQueryStep#query
>   * CtQueryable#scan -> CtQueryable#filterChildren
>
excellent!

> The |CtFunction| is the additional feature.
and remove this one, to better understand the core design and contract.

--Martin

pvojtechovsky:
I did these changes:
ChainableFunction#apply renamed to CtQueryStep#process

Removed class `Scann`. The children scanning is implemented directly in `CtQueryImpl$ChildrenFilteringWrapper`

By default ClassCastException is thrown when query step parameters are not assignable. It simplifies detection of mistakes in query building. There is `CtQuery#failOnCCE(boolean)`, which can be used to enable/disable CCE in current query. Note: The scan/filterChildren method never throws CCE.

I kept `CtFunction` in this PR, because:
- it's remove would need changes in tests
- there are only these three places where `CtFunction` is used 
  1) CtElementImpl#map(CtFunction)
  2) CtQueryImpl#map(CtFunction)
  3) CtQueryImpl$FunctionWrapper
  so it should be possible to understand the core contract, by ignoring these 3 places.

logging of query process was improved. There is possible to name each query step and to see it's name in the log. If name is not provided then each step automatically gets the name which is equal to order number of this step.

monperrus:
Thanks for the huge simplicity effort, with now six new interfaces Consumer, CtFunction, CtQuery, CtQueryStep, CtQueryable with few methods each.

* interface Consumer: rename to CtConsumer
* failOnCCE(boolean) -> failurePolicy(enum QueryFailurePolicy)
* logging() -> not elegant in an interface, if really necessary, let's put it in the implementation only, and the client code which wants to log would cast to the implementation class?
* CtQueryStep::process(T input, Consumer<R> output). Process if for processor, let's call it forEach? and switch the two parameters as follows forEach(Consumer<R> output, T input) (for a clean relation with the other forEach?)

pvojtechovsky:
> interface Consumer: rename to CtConsumer

done
> failOnCCE(boolean) -> failurePolicy(enum QueryFailurePolicy)

done

> logging() -> not elegant in an interface, let's put it in the implementation only,

done
> CtQueryStep::process(T input, Consumer output). Process if for processor, let's call it forEach? and switch the two parameters as follows forEach(Consumer output, T input) (for a clean relation with the other forEach?)

done

pvojtechovsky:
Hi Martin, anything else needed to merge it?

monperrus:
Hi Pavel,

Thanks again for the great progress. I've just made a PR on your PR about the doc.

One or two more things about the design:

* it seems that the essence of the design is [lazy evaluation](https://en.wikipedia.org/wiki/Lazy_evaluation). Do you agree?

* we have two forEach methods (in CtQueryStep and in CtQuery), which is not nice. We have to find the common and reusable concept for this. Do we really need CtQueryStep? Would CtConsumer only suffice? 

* related: `CtQueryable.map(CtQueryStep)` is not a map function. A better name would be `append`.

Finally, there are a couple of [untested lines and cases in CtQueryImpl](https://coveralls.io/builds/9286191/source?filename=src%2Fmain%2Fjava%2Fspoon%2Freflect%2Fvisitor%2Fchain%2FCtQueryImpl.java)

pvojtechovsky:
>  Do we really need CtQueryStep?
yes. Try to reimplement OverridenMethodFilter in this PR, without using of CtQueryStep. And show me that. I guess it will be clear then whether it it is possible or not.

> 2*forEach not good
I agree. Any suggestions?

monperrus:
> Try to reimplement OverridenMethodFilter in this PR, without using of CtQueryStep.

No problem, I trust you.

I'm trying to find a design that is understandable and appropriately documented for the other users. If I don't get myself the gist of a design in Spoon, it's a bad smell. In this case, it's probably a naming issue, let me think about it.

monperrus:
In the PR on your changes (https://github.com/pvojtechovsky/spoon/pull/4), there is new commit https://github.com/pvojtechovsky/spoon/pull/4/commits/7d6b8f4ab1a08dd7aa6503db3773463d8e614814

Changelog:

* CtQueryStep has been deleted, replaced by CtConsumer.
* Consequently, there is only one function map in CtQueryable

Open question:

* All tests pass but one, but it seems a minor problem (not a conceptual one), I've not yet completely grasped your step architecture.

WDYT?

monperrus:
added the requirements for this new feature at the top of this thread.

pvojtechovsky:
Thanks for Your PR! It helped me to understand Your thinking. Now I can try to clarify the misunderstanding, which is there. You are trying to implement different feature then me.
My PR contains these features:
F1 possibility to implement CQS like FieldReferencesQuery. I need that for another PR about refactoring.
F2 chaining of query steps
F3 the query steps based on CtFunction
Your PR contains only F2 and F3. It is not sufficient for me. OK, I know that you preffer small steps and my PR was too big step. 

So I will comment your PR now and then merge it.

pvojtechovsky:
Please see my comments in your PR, but do not fix anything in your PR. I will takeover it as it is and then, I will simplify it even more. So we will have first small step on which we both can agree on.

What I will do?
I like the idea of FieldReferenceFilter as Query. So we will have no query steps. Just queries which can be composed from other queries.
I will try to give a chance the idea of replacing Consumer by CtFunction<T,Void>
I will remove CtQueryable from the CtElement. The problem of that approach is that Query has predefined input (source element) since beginning and it makes things more complicated then is needed. So after that change we will evaluate query on an `inputElement`this way:
```java
factory.Query().map(new FieldReferenceQuery()).forEach(resultConsumingFunction, inputElement
//OR
List list=factory.Query().map(new FieldReferenceQuery()).list(inputElement)
//OR
factory.Query().map(new FieldReferenceQuery()).map(resultConsumingFunction).list(inputElement)
```
This approach is not so nice like current one, but it is simplest code and the most powerful and flexible way. Later we can use that low level code and to write some nicer facade about that, so the usability will improve too.
I am also not sure about factory.Query, forEach and list method name... etc.

One more idea: If we will not merge code of OverridenMethodFilter with OverridenMethodQuery, then we can even merge `map` function with `filterChildren`, because the algoritm and client will know what to do by looking at implemented interface or class name. But may be `filterChildren` is good name, which intuitivelly explains what such query step will do, so we can keep it.

monperrus:
> I will takeover it as it is and then, I will simplify it even more. So we will have first small step on which we both can agree on.

OK, excellent.

>  Just queries which can be composed from other queries.

Sounds a cool idea

> But may be `filterChildren` is good name, which intuitivelly explains what such query step will do, so we can keep it. 

The big advantage of `filterChildren` right now, is that it's a natural explanation and transition step from old-style filters no new-style queries.

> I will remove CtQueryable from the CtElement. 

Are you sure? It's a really meaningful, concise and coherent interface, perfectly fitting into `CtElement`.

pvojtechovsky:
> > I will remove CtQueryable from the CtElement.
>
> Are you sure? It's a really meaningful, concise and coherent interface, perfectly fitting into CtElement.

I agree with you, but it is kind of facade over the internal query engine. And I have learned from this PR, that too much steps at once causes misunderstanding. So I want to have first low level query engine merged in Spoon and then we can discuss how to integrate it into CtElement.
Why that? The interface like CtQueryStep#forEach(Consumer out, CtElement input) is the low level and I NEED it to build other features. But if I directly implement CtElement facade, then it looks like this interface is not needed and You tends to delete it. ;-)

monperrus:
> And I have learned from this PR, that too much steps at once causes
> misunderstanding. 

I learned that too. And also that writing requirements upfront is useful.

> So I want to have first low level query engine merged in Spoon and
> then we can discuss how to integrate it into CtElement.
OK.

pvojtechovsky:
I am not able to implement next version, because each idea ends with the progress killing thought: "Martin doesn't like CtConsumer, CtQueryStep, forEach".... 
... and I do not like `list()`, which will need a lot of memory for queries in my project where I need to process 6000 classes * 20 methods * 5 parameter * ?? elements * in queries with about 5 to 10 query steps. 
... and which produces not nice code like:
```java
List list = ...query.list();
for(item:list) {
 ... do something with each item
}
```
spoon is already full of such code. Just search for usages of CtElement#getElements()
So I am giving up. Any idea how to proceed?

pvojtechovsky:
Here is the exact definition what avoids future progress from my side:
Your PR on my PR deletes OverriddenMethodFilter#forEach(CtConsumer<CtMethod<?>> output, CtMethod<?> input), which is the **example** of usage of MAIN feature of my PR. With code of your PR it is impossible for me to implement such method effectively. Show me how to implement it using Your approach. Then 
A) I will understand a way, which is evident for You, but not for me.
B) You will understand that CtQueryStep#forEach(Consumer, input) and CtConsumer are good and will not try to remove it from the design.

monperrus:
Let's focus on requirements and API.

It seems we now agree on a basic set of requirements (listed here) and
API (in https://github.com/pvojtechovsky/spoon/pull/4).

Why about merging it first? What prevents this to your opinion?

For implementing scope-based query, we will probably either add new
methods in the query interfaces or new interfaces, but again, let's
first agree on the core design here.

pvojtechovsky:
I merged your PR. 

> What prevents this to your opinion?

Query composition is not possible now with your PR. 

> For implementing scope-based query, we will probably either add new
methods in the query interfaces or new interfaces, 

ok, I suggest that You continue on future query development, until it gets at least to state, where it is possible to implement scope based queries. 

> but again, let's first agree on the core design here.

this core design is quite limited now.  I agree with anything what will cause a progress with this PR. But I will spent my effort on another Spoon features now, which are based on the state of this PR before your PR was merged.

monperrus:
> Query composition is not possible now with your PR.
>
renamed the title.

do you agree that the four listed requirements are well supported with
this PR?

pvojtechovsky:
> map functions with arbitrary return types (incl. that are not CtElement)

yes

> chained queries

partially. You can chain implementations of CtFunction, but you cannot chain queries.

> lazy evaluation

partially. The CtFunction is evaluated lazily for each input element, but it is not possible the apply lazy evaluation to output of that function, because function must return results using `return` statement and cannot call next query step by Consumer.apply(element). Therefore usage of such queries is limited to smaller result sets.

> support for queries with Java 8 lambdas

yes

But I have no problem if you merge it like it is and then continue with scope-based query

monperrus:
> But I have no problem if you merge it like it

The idea is to be 100% on the same wavelength before merging



>> chained queries
> partially. You can chain implementations of CtFunction, but you cannot chain queries.

to me, "chaining" means writing `filterChildren(x).filterChildren(y)...` or `map(z).map(t)...` or even `filterChildren(p).map(q)...`

This is completely supported, right?

pvojtechovsky:
yes this is supported

monperrus:
On lazy evaluationn you say

> The CtFunction is evaluated lazily for each input element, but it is not possible the apply lazy evaluation to output of that function, because function must return results using return statement and cannot call next query step by Consumer.apply(element). Therefore usage of such queries is limited to smaller result sets.

I don't understand. Could you elaborate? 

To me, in the current implementation, all queries are only performed by calling `list()` on the last query method. Correct?

pvojtechovsky:
>> it is not possible the apply lazy evaluation to output of that function

> I don't understand. Could you elaborate?

The implementation of ChildrenFilteringWrapper is correct, because it calls 
```java
if (matches) {
	//send input to output, because Fitler.matches returned true
	next.accept(element);
} 
```
So it will work effectively for large amount of elements too.

The implementation of FunctionWrapper is correct too, because CtFunction is intended for query steps, which will return relatively small number of results.

So what is wrong? There is no way how to extend Query by way it works well for functions, which produces large result set. You mentioned it will be next step. OK. But as I mentioned until this feature is back again, I cannot continue, so I will continue on my old branch in the mean time.

Anyway thanks for Your effort Martin, to try to make it/keep it understandable!

monperrus:
@surli this PR is now ready for your review. there is no need to read and understand the whole discussion, the PR itself (doc, javadoc, tests) should now be self-contained.

src/main/java/spoon/reflect/visitor/QueryVisitor.java
```diff
@@ -42,8 +46,8 @@ public QueryVisitor(Filter<T> filter) {
 		if (filter instanceof AbstractFilter) {
 			filteredType =  ((AbstractFilter<T>) filter).getType();
 		} else {
-			Class<?>[] params = RtHelper.getMethodParameterTypes(filter.getClass(), "matches", 1);
-			filteredType = (Class<T>) params[0];
+			Method method = RtHelper.getMethod(filter.getClass(), "matches", 1);
```
surli: You did not check here that `method` can be null as you did in `AbstractFilter`: it's on purpose because you assume it can never happens or is it missing?

src/main/java/spoon/reflect/visitor/chain/CtQueryImpl.java
```diff
@@ -0,0 +1,452 @@
...
+			return false;
+		}
+	}
+
+	/**
+	 * Interface used to query and manipulate elements in a functional manner.
+	 *
+	 * @param &lt;T> - the type of accepted elements
+	 */
+	private interface CtConsumer<T> {
```
surli: What's the purpose exactly to distinguish this interface from Step? From my point of view, each time you use a CtConsumer it's in fact a Step you're using, am I wrong?

src/main/java/spoon/reflect/visitor/QueryVisitor.java
pvojtechovsky: I cannot imagine situation when "matches" method is missing, because Filter interface has such method.
But I can add check there. Thank you

src/test/java/spoon/test/filters/FilterTest.java
```diff
@@ -466,4 +472,106 @@ public boolean matches(CtInvocation<?> element) {
...
+			.map((CtFieldReference<?> fr)->{
+				if("myfield".equals(fr.getSimpleName())){
+					context.count++;
+				}
+				context.totalCount++;
+				return true;
+			}).list();
+		assertTrue(context.count==1);
+		assertTrue(context.totalCount>1);
+
```
pvojtechovsky: You are right this test brings nothing new, The testFunctionQueryStep already tests the same. I will remove this test

src/main/java/spoon/reflect/visitor/chain/CtQueryImpl.java
pvojtechovsky: I need CtConsumer as public interface later. But Martin made it private in this first step, because after his simplifications/deleting of functionality this interface is not needed.

src/main/java/spoon/reflect/visitor/QueryVisitor.java
pvojtechovsky: fixed

src/test/java/spoon/test/filters/FilterTest.java
pvojtechovsky: I have deleted the test

Issue #980, opened 2016-11-16
bug no fully qualified name should be written if top package is shadowed by a local variable


monperrus:
See test case in https://github.com/monperrus/spoon/commit/9694ff929b8980948e618cc83304bcbc5420e5a1

PR #1021, opened 2016-12-05
Proposal to fix #972

### LINKED ISSUE #972: test APITest.testSetterInNodes() does not work at all
The API contract which should be enforced by the APITest.testSetterInNodes() IS NOT CHECKED, because this test is actually doing NOTHING.

Put the breakpoint on line 359
```java
assertTrue("Check the method " + statement.getParent(CtMethod.class).getSignature() ...
```
and you will see it NEVER enters the loop.

I have made a PR #968, which fixes that bug.... 

Yes Travis reports the errors, but it is not bug in the PR #968, it is report about broken API contracts!

I am not going to fix PR #968, because I have no idea how to do it. But I guess it is good idea to fix it to avoid merging more and more invalid APIs.

surli:
fix #972
Change test condition for testing setter: originally an OR was used where an AND should be used IMHO.

### DIFF
--- a/src/main/java/spoon/reflect/factory/CodeFactory.java
+++ b/src/main/java/spoon/reflect/factory/CodeFactory.java
@@ -107,7 +107,13 @@ public class CodeFactory extends SubFactory {
 	 * @return a accessed type expression.
 	 */
 	public <T> CtTypeAccess<T> createTypeAccess(CtTypeReference<T> accessedType) {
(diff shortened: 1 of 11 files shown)


monperrus:
#972 is not about shadowed variables. do you mean #980?

surli:
Hm no my fix was about #972 but just after I pushed a bit fast some code about #980, I close the PR I will move the fix on another branch and resend it.

monperrus:
OK, thanks.

PR #1022, opened 2016-12-05
fix #972: test APITest.testSetterInNodes()

### LINKED ISSUE #972: test APITest.testSetterInNodes() does not work at all

surli:
fix #972
Change test condition for testing setter: originally an OR was used where an AND should be used IMHO.


monperrus:
remove noise + fix bug (if any)

surli:
Add a condition in the test to ignore annotation UnsettableProperties.
Some setter had to be fixed so choices have been made: 
  - in CtFieldImpl and CtLocalVariableImpl an UnsettableProperty has been added to setAssignment as it does not refer to a property but call setDefaultExpression
  - in CtLoopImpl, CtTryImpl, and CtExecutableImpl setBody has been changed to add a conditional test on statement attribute : it's redundant with the test on body, but I was not sure if I needed to change the test to ignore some setters or the method here...
  - in CtVariableAccessImpl I added the type.setParent as expected in the contract of the test, but should it contain one ?

monperrus:
looks almost ready to be merged. @pvojtechovsky you created this issue, OK for you?

src/test/java/spoon/test/api/APITest.java
```diff
@@ -355,11 +371,13 @@ public void accept(CtVisitor visitor) {
 
 		final List<CtMethod<?>> setters = Query.getElements(launcher.getFactory(), new SetterMethodWithoutCollectionsFilter(launcher.getFactory()));
```
pvojtechovsky: I suggest to check the size of the setters array. At least with the code like
```java
assertTrue(setters.size()>0)
```
actually there is 69 methods and at the time of the bug the collection size was 0.

pvojtechovsky:
I have added two comments. It is up to You if you like it and implement it. The code is OK even without these changes.
Thanks to @surli for fixing!

src/main/java/spoon/support/reflect/code/CtVariableAccessImpl.java
```diff
@@ -58,6 +58,9 @@
 	@Override
 	public <C extends CtTypedElement> C setType(CtTypeReference<T> type) {
 		if (type != null) {
+			type.setParent(this);
```
surli: Then the contract should not be respected for this specific setter. I could simply ignore it, but I imagine in the future it won't be the only one. 
So should I create a specific annotation to mark setters which does not respect this contract ? Or do you have another idea ?

monperrus: yes, this is why the getter in the interface is @DerivedProperty.

surli: Then I can ignore all setters with @DerivedProperty?

monperrus:
2 solutions:

- ignore all setters for which the getter is @DerivedProperty

- put @UnsettableProperty on the setter.

what's the best to your opinon?

surli:
Put @UnsettableProperty on the setter does not look good, as the property is set, but not in the current object. 

The second option seems cleaner, I'll take a look on it.

monperrus:
looks good. ready to merge for you?

surli:
Ok for me.

monperrus:
first merged PR :-) for @surli. welcome to the wonderful world of Spoon!

PR #1031, opened 2016-12-07
fix: shorter this access

### LINKED ISSUE #826: DefaultJavaPrettyPrinter.visitCtThisAccess

monperrus:
This is an important change in the way we print this access.

The current strategy of writing qualified this `v.l.pack.n.Foo.this.bar()` yields unreadable code even in readable with-imports mode.

This PR fixes this.

In the mid-term, I would like to completely get rid of checks on isImplicit for this accesses, but this requires some additional work on the model. As discussed recently, there are incorrect implicit targets, which nobody detects and reports because they are not printed. Only advanced transformers suffer from them.

The pretty-printing of this accesses was incredibly strongly specified (kudos to @GerardPaligot ), hence the number of changed tests, please review carefully.

closes #826


src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
```diff
@@ -719,45 +716,50 @@ public void visitCtDo(CtDo doLoop) {
...
-			if (!(parent instanceof CtFieldWrite) || !thisAccess.equals(((CtFieldWrite) parent).getTarget()) || thisAccess.getParent(CtConstructor.class) == null) {
-				return false;
+			// TODO find a way to remove this check
+			// ideally we sould not rely on implicitness
+			// for correct pretty-printing, because this is fragile
+			// wrt to bugs in the model and when refactoring
+			if (thisTarget.isImplicit()) {
+				// write nothing, "this" is implicit and we unfortunately cannot always know
+				// what the good target is in JDTTreeBuilder
+				return;
```
monperrus: no because the caller checks for implicitness as well (see ctVisitInvocation for instance).

however, this is an interesting question: who is responsible for printing the "."? probably the parent element, as done today (but consistently?).

however instead of 
```
if (target.isImplicit()) {
  scan(target);
  print(".")
}
```
I would prefer 
```
scan(target);
if (somethingHasBeenWritten()) {
  print(".")
}
```
This would be less fragile.

(outside the scope of this PR but I'd like to have your opinion on this)
```

**Comment by monperrus** [comment]
well done. then a test is missing, I'll add it.

**Comment by monperrus** [comment]
simple this compute getTarget() on the fly

File: src/test/java/spoon/test/prettyprinter/QualifiedThisRefTest.java [hunk 2]
```diff
@@ -55,7 +56,8 @@ public void testQualifiedThisRef() {
 		ctTypes.add(ctClass);
 		printer.getElementPrinterHelper().writeHeader(ctTypes, imports);
 		printer.scan(ctClass);
-		assertTrue(printer.getResult().contains("Object o = QualifiedThisRef.Sub.this"));
+		Assert.assertTrue(printer.getResult().contains("Object o = this"));
+		Assert.assertTrue(printer.getResult().contains("Object o = QualifiedThisRef.this"));
```
tdurieux: contains is weak...

pvojtechovsky:
I tried the code of this PR with my project and printed version compiles well. I also do not see problem in code ... but there are part which I do not understand, or I am lazy to understand ... :-)

src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
```diff
@@ -682,8 +681,9 @@ public void visitCtDo(CtDo doLoop) {
 			}
 			if (f.getTarget() != null) {
 				if (!isInitializeStaticFinalField(f.getTarget())) {
+					printer.snapshotLength();
 					scan(f.getTarget());
-					if (!f.getTarget().isImplicit()) {
+					if (printer.hasNewContent()) {
```
monperrus: you're right, replaced by a queue.

pvojtechovsky: what about this code?
```java
printer.getSnapshotLength();
{
...some complex code with many ifs, loops 
... and evil
          return;
... added by somebody else after some years, will bring him hot time ...
}
if (printer.hasNewContent()) {...}
```
this 
```java
int snapshot = printer.getSnapshotLength()
...
if(print.hasNewContent(snapshot)) {...}
```
is little bit more code, but is really safe

monperrus: yes, but looks less elegant. let's wait for the devil to make ugly things :-).

monperrus:
I wanted to go first for shorter and more explicit this accesses. I only keep shorter for now because I don't want to break too much test cases from client code.

For handling explicitness during refactoring, we have to remove implicit in the appropriate setters and refactoring methods.

src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
pvojtechovsky: ;-))

monperrus:
@tdurieux you can merge

Issue #970, opened 2016-11-13
bug: some annotations are not well-printed 


monperrus:
if you spoon `https://github.com/spoon-bot/dropwizard/blob/master/dropwizard-metrics-graphite/src/main/java/io/dropwizard/metrics/graphite/GraphiteReporterFactory.java` twice (spoon(spoonx)): 

```
    @JsonProperty
    public int getPort() {
        return port;
}
```
is rewritten as:
```
   @com.fasterxml.jackson.annotation.JsonProperty
    public @io.dropwizard.validation.PortRange
    @io.dropwizard.validation.PortRange
    int getPort() {
        return port;
    }
```

monperrus:
To reproduce:
```
git clone https://github.com/spoon-bot/dropwizard.git
cd dropwizard/
git checkout .
curl http://spoon.gforge.inria.fr/jenkins/build.sh | stdbuf -i 50k bash
```

PR #1032, opened 2016-12-08
Proposal to fix issue with annotations by adding a test in substituteAnnotation

surli:
Fix the issue #970 by adding a test in substituteAnnotation in ParentExiter: instead of just adding the annotation to the returnType of the method, it test before is the annotation is already there or not.


surli:
And dropwizard is well building with this fix.

Issue #1025, opened 2016-12-06
ClassCastException when using TemplateMatcher with a string literal parameter


monperrus:
see http://stackoverflow.com/questions/40964338/classcastexception-when-using-templatematcher-with-a-string-literal-parameter

surli:
No answer concerning the fix proposed, neither here or on stackoverflow: do we close the ticket for now?

surli:
I close the issue for now, don't hesitate to reopen it if the problem persists.

PR #1034, opened 2016-12-09
Propose a fix to #1033 and add a new test for TemplateMatcher

### LINKED ISSUE #1033: Template matcher also find itself when matching
See https://github.com/surli/spoon/commit/1dd645d8907fbe6a228d2bcba7fd61a44557ba6c to reproduce the bug: currently it detects both `BTServiceImpl` and `SecurityCheckerTemplate`

surli:
Fix #1033 by looking if the TemplateMatcher has been detected in the find and remove it if it the case. Moreover this PR add a new dynamic contract by checking before doing any scan, if the TemplateMatcher is able to match itself.

However I still wonder why the `find` method in TemplateMatcher does return a `List` and not a `Set` or at least a `Collection` ?


src/main/java/spoon/template/TemplateMatcher.java
```diff
@@ -185,7 +186,20 @@ public void scan(CtElement element) {
 				}
 				super.scan(element);
 			}
-		}.scan(targetRoot);
+		};
+
+		scanner.scan(templateRoot);
+		if (!finds.contains(templateRoot)) {
+			throw new SpoonException("TemplateMatcher was unable to find itself, it certainly express a problem somewhere in the template.");
```
monperrus: TemplateMatcher was unable to find itself, it certainly indicates a bug. Please revise your template or report an issue.

monperrus:
>  However I still wonder why the `find` method in TemplateMatcher does return a `List` and not a `Set` or at least a `Collection` ? 

what do you mean?

monperrus:
so is there a bug behind #1025 ?

src/main/java/spoon/template/TemplateMatcher.java
surli: I'll change that.

surli:
>> However I still wonder why the find method in TemplateMatcher does return a List and not a Set or at least a Collection ?

> what do you mean?

The signature of the method `find` in the TemplateMatcher is the following:
`public List<CtElement> find(final CtElement targetRoot)`

But I don't really see the point of returning a `List`: is it possible to have multiple times the same element in the findings? Or should each finding be an independent element (as in a `Set`)?

> so is there a bug behind #1025 ?

Frankly I don't know: I tried to show the bug by taking elements from [the stackoverflow question](http://stackoverflow.com/questions/40964338/classcastexception-when-using-templatematcher-with-a-string-literal-parameter) to create a test but apparently it works fine: I ask the guy to have a look on my test and to try the last Spoon version to check if it's ok for him... 
Maybe we can close the issue and reopen it or create a new one depending on his answer?

monperrus:
>  Maybe we can close the issue and reopen it or create a new one depending on his answer?

good idea

PR #1037, opened 2016-12-12
Interface to Exclude Compilation Units From CtModel.

### LINKED ISSUE #985: [SpoonModelBuilder] Create a Model for Subset of Input
Creating a Spoon model from large size software systems takes several minutes due to the amount of analyses performed while parsing all input compilation units. Sometimes - for example when individual files have been updated (#752) - I want to create model describing a subset of the input sources only. I could add the desired files only but using noclasspath mode having a more complete classpath a very useful since, otherwise, existing references may be missing. Therefore, I'm looking for a feature that allows me to create a model for a subset of my input sources. 

My suggestion is to add a method to `SpoonModelBuilder` allowing me specify a filter that is applied **after** creation of the compilation units (in order to have more complete classpath) but **before** the traversal of a particular unit in order to minimize the required time to create a model.

retuxx:
fixes #985


src/main/java/spoon/support/compiler/jdt/CompilationUnitFilter.java
```diff
@@ -0,0 +1,45 @@
...
+	 * @param cud
+	 *      The {@link CompilationUnitDeclaration} that may or may not be
+	 *      excluded.
+	 * @param path
+	 *      {@code cud}'s file path (see
+	 *      {@link CompilationUnitDeclaration#getFileName()})
+	 * @return {@code true} if and only if {@code cud} should be excluded,
+	 *         {@code false} otherwise.
+	 */
+	boolean exclude(final CompilationUnitDeclaration cud, final String path);
```
monperrus: this is a JDT interface. Ideally, all Spoon interfaces only depend on other Spoon interfaces (good encapsulation). What about changing to `CompilationUnit`?

retuxx: That's why I stored this file in `support/compiler/jdt` rather than `support/compiler/`. Changing the parameter `cud` from `CompilationUnitDeclaration` to `CompilationUnit` should be feasible, though.

retuxx: On the other hand, doesn't `JDTBasedSpoonCompiler.java` have to traverse a compilation unit in order to produce a `CompilationUnit`. That's exactly what I want to avoid here.

retuxx: I think removing the first parameter might be the best solution here.

src/main/java/spoon/support/compiler/jdt/JDTBasedSpoonCompiler.java
```diff
@@ -413,7 +414,15 @@ protected boolean buildUnitsAndModel(JDTBuilder jdtBuilder, SpoonFolder sourcesF
 
 	protected void buildModel(CompilationUnitDeclaration[] units) {
 		JDTTreeBuilder builder = new JDTTreeBuilder(factory);
+		units:
```
tdurieux: Can you change the name of the label because it is the same as the parameter. It is confusing.

Ex: unitLoop

retuxx: Done.

retuxx:
Is this PR ready to merge or do I miss something?

retuxx:
Thank you for merging.

monperrus:
Thanks for your contribution.

PR #1040, opened 2016-12-13
New feature: allow spoon to do static imports for methods and fields

surli:
Improve `ImportScannerImpl` to allow Spoon to do static imports. 
Change the interface of `ImportScanner` to return list of CtReference when doing computeImports. 

Lots of tests had to be changed as with autoImports it automatically imports all static elements.


src/main/java/spoon/reflect/visitor/ImportScannerImpl.java
```diff
@@ -132,134 +146,305 @@ public void scan(CtElement element) {
...
+		listallImports.addAll(this.classImports.values());
+		listallImports.addAll(this.fieldImports.values());
+		listallImports.addAll(this.methodImports.values());
+		return listallImports;
 	}
 
 	@Override
 	public void computeImports(CtElement element) {
-		imports.clear();
+		classImports.clear();
```
tdurieux: You have to clear the 3 lists

surli: Agreed. I push the change.

monperrus:
cool!

in Spoon, we aim at maximum backward compatibility and not to change the signature public methods, esp. in interfaces. 

to this extent, I propose to keep computeImports as is and to introduce a new method (eg computeAllImports).

this will greatly reduce the impact in the tests (and in client code also).

(fortunately isImported(CtReference ref) is backward compatible)

surli:
> to this extent, I propose to keep computeImports as is and to introduce a new method (eg computeAllImports).
> this will greatly reduce the impact in the tests (and in client code also).

Ok but then DefaultJavaPrettyPrinter will use the computeAllImports, right?

surli:
Ok to merge this now @monperrus ?

PR #1048, opened 2016-12-16
Fix some issues with static imports and inner classes or enums

surli:
Some issues have been detected using spoon on open source projects (see https://ci.inria.fr/sos/job/EveryNightRegressionTesting/516/) du to failing static import on inner classes or enum.
Fix the detected errors by changing the way fully qualified name are printed in imports for inner classes (for now a $ was printed before the inner class name) and by detecting in the MinimalImportScanner if the simpleName of the currentRef is creating a collision: in that case, we should not import the element as it could create bug by autoreference.


src/main/java/spoon/reflect/visitor/printer/ElementPrinterHelper.java
```diff
@@ -271,18 +271,22 @@ public void writeHeader(List<CtType<?>> types, Collection<CtReference> imports)
...
-					printer.write("import static " + fieldRef.getDeclaringType().getQualifiedName() + "." + fieldRef.getSimpleName() + ";").writeln().writeTabs();
+					printer.write("import static " + this.removeInnerTypeSeparator(fieldRef.getDeclaringType().getQualifiedName()) + "." + fieldRef.getSimpleName() + ";").writeln().writeTabs();
 				}
 
 			}
 			printer.writeln().writeTabs();
 		}
 	}
 
+	private String removeInnerTypeSeparator(String FQN) {
```
monperrus: Name `FQN` does not comply with Java conventions. Our xxxx-checkstyle doesn't detect this?

src/main/java/spoon/reflect/visitor/printer/ElementPrinterHelper.java
```diff
@@ -271,18 +271,22 @@ public void writeHeader(List<CtType<?>> types, Collection<CtReference> imports)
...
+		return FQN.replace('$', '.');
```
monperrus: we have a TYPE_SEPARATOR representing `$` somewhere.

src/main/java/spoon/reflect/visitor/printer/ElementPrinterHelper.java
surli: Apparently not... I'll open an issue to fix/simplify the checkstyle!

Issue #1045, opened 2016-12-16
regression on fully-qualified names for constructor


monperrus:
see generated code in #1039 

regression spotted by the eagle eye of @tdurieux

surli:
@monperrus could you just check if the fix proposed in #1048 changes anything for this issue?
And could you give me info how you call the generators exactly?
Thanks!

monperrus:
>  And could you give me info how you call the generators exactly?

By simply running `CtGenerationTest`

surli:
Wrong issue closed. #1048 was intended to fix #1044. It seems this bug is still open.

PR #1052, opened 2016-12-19
Fix FQN regression in FQN mode

### LINKED ISSUE #1041: Improvement: setAutoimport and DefaultJavaPrettyPrinter
With the new implementation of ImportScanner, MinimalImportScanner, we must change the way/the moment when DefaultJavaPrettyPrinter create the instance of ImportScanner, because if we call setAutoImport after the instance creation, it will never change in DefaultJavaPrettyPrinter resulting in bad behaviour.

surli:
The previously implemented `MinimalImportScanner` was only comparing FQN against list of previously explored variables, which means all variables were considered even those which were not on the scope of the current variable. It fixes issue #1045.
The proposed implementation now only check the current scope to search name collisions.
Moreover, now the ImportScanner is recreated in DefaultJavaPrettyPrinter when `calculate` method is called: it avoids the scanner to test collision names against another class variable names, but it should also fix #1041 (to be tested).


src/main/java/spoon/reflect/visitor/ImportScannerImpl.java
```diff
@@ -64,6 +64,20 @@
 	protected CtTypeReference<?> targetType;
 	private Map<String, Boolean> namesPresentInJavaLang = new HashMap<>();
 
+	/**
+	 * Determine if the ImportScanner should work in a FQN mode or in auto import mode.
+	 */
+	private boolean fqnMode;
```
surli: handle with polymorphism

PR #1058, opened 2016-12-20
Refactor CtReference in order to inherit from CtNamedElement

### LINKED ISSUE #1046: Change the org.eclipse.jdt.core dependency to use dependency from Maven Central
See if we can use the dependency from Tycho project : https://mvnrepository.com/artifact/org.eclipse.tycho/org.eclipse.jdt.core

### LINKED ISSUE #1047: refactoring: CtReference extend CtNamedElement
This is a general question about the metamodel.
It is explicitely specified that `CtReference` is a named element:

> This is the root interface for named program element references.

However it extends `CtElement` and specify the methods `getSimpleName` and `setSimpleName` which are defined in `CtNamedElement`. 
I assume this is because `CtNamedElement` also define the method `getReference` which we do not want in `CtReference`. But why not restrict `CtNamedElement` to the two previously mentionned methods and create another interface `CtElementWithReference` for `getReference`, then create another interface which inherits from `CtNamedElement` and `CtElementWithReference`?

surli:
Refactor `CtReference` which was declaring the same method as CtNamedElement to make it inherit from `CtNamedElement`. Change the implementation and the CtInheritanceScanner accordingly. Also make a change in `FilterName` to not consider the CtReference elements when matching. Fix #1047.

Also change the dependency for eclipse.jdt.core in pom.xml to use maven central. Fix #1046.


src/main/java/spoon/support/reflect/reference/CtWildcardReferenceImpl.java
```diff
@@ -32,11 +31,6 @@ public CtWildcardReferenceImpl() {
 	}
 
 	@Override
-	public <T extends CtReference> T setSimpleName(String simplename) {
```
surli: Keep it

src/main/java/spoon/reflect/visitor/filter/NameFilter.java
```diff
@@ -39,7 +40,11 @@ public NameFilter(String name) {
 
 	public boolean matches(T element) {
 		try {
-			return name.equals(element.getSimpleName());
+			if (!(element instanceof CtReference)) {
```
surli: Document the change

PR #1059, opened 2016-12-20
Add a new test to check if TemplateMatcher can match two snippets

### LINKED ISSUE #1035: Test: add a new test to check if TemplateMatcher can match twice the same snippet
We should add a new test to check if TemplateMatcher can match twice the same snippet at two different locations.

surli:
Fix #1035


src/test/java/spoon/test/template/testclasses/TwoSnippets.java
```diff
@@ -0,0 +1,37 @@
...
+
+    public void hello() {
+        if(!contextHelper.hasPermission("c")) {
+            throw new SecurityException();
+        }
+        bDao.toString();
+    }
+
+    public void toto() {
+        if(!contextHelper.hasPermission("b")) {
```
surli: We want to see if we match both if they are exactly the same.

Issue #1055, opened 2016-12-20
This is not correctly printed in case of multiple inner class


surli:
In the following structure: 
```
public class Rule {
    public static final class Phoneme implements PhonemeExpr {
        public static final Comparator<Phoneme> COMPARATOR = new Comparator<Phoneme>() {[...]};

        private final StringBuilder phonemeText;
        private final Language language;

        public Phoneme(final CharSequence phonemeText, final Language language) {
            this.phonemeText = new StringBuilder(phonemeText);
            this.language = language;
        }
```

The last `this` are printed as following: `Rule.Phoneme.this` leading to a compilation error because the fields are `final`.

PR #1061, opened 2016-12-20
Fix a print bug with 'this'

surli:
Another bug has been spotted when writing this access in inner classes for final field assignment. It has also been fixed: #1055.


monperrus:
bug is probably related to context this pile not being correct

Issue #1073, opened 2016-12-24
ParentNotInitializedException when processing comments with -c 


giphon:
I am aware that comments can be processed using the command "-c"  since Spoon 5.2.0. Thanks for this feature.  

When using Spoon to process comments in either code snippet below, ParentNotInitializedException will be thrown. An empty processor can be used to reproduce this issue. This happens in both Spoon 5.2 and 5.4.

public class Comment_1 {

	public void code_1()
	{
		try {		}
		// A 
		// B
		catch (Exception ex)
		{		}
	}
}

public class Comment2 {

	// C
 	@interface Code_2{} 	
}

Details of the exception are as follows.

spoon.reflect.declaration.ParentNotInitializedException: parent not initialized for class spoon.support.reflect.code.CtCommentImpl (/home/jifeng/workspace/Temp4/src/main/java/example/Comment1.java:17)
	at spoon.support.reflect.declaration.CtElementImpl.getParent(CtElementImpl.java:284)
	at spoon.support.compiler.jdt.JDTCommentBuilder.insertCommentInAST(JDTCommentBuilder.java:377)
	at spoon.support.compiler.jdt.JDTCommentBuilder.buildComment(JDTCommentBuilder.java:131)
	at spoon.support.compiler.jdt.JDTCommentBuilder.build(JDTCommentBuilder.java:96)
	at spoon.support.compiler.jdt.JDTBasedSpoonCompiler.buildSources(JDTBasedSpoonCompiler.java:387)
	at spoon.support.compiler.jdt.JDTBasedSpoonCompiler.build(JDTBasedSpoonCompiler.java:116)
	at spoon.support.compiler.jdt.JDTBasedSpoonCompiler.build(JDTBasedSpoonCompiler.java:99)
	at spoon.Launcher.buildModel(Launcher.java:712)
	at spoon.Launcher.run(Launcher.java:663)
	at spoon.Launcher.run(Launcher.java:106)
	at spoon.Launcher.main(Launcher.java:99)

surli:
Thank you for the feedback, I just reproduce your bug in https://github.com/surli/spoon/commit/7082dcf803faf0f2ca0dc512f7610459fbf17c0d I will investigate for a fix. 
In the meantime, you can pass the checks in Spoon using `-a` argument in command-line.

giphon:
Thanks a lot. Here is another example for incorrect comments for -c  . 
When processing 
```
	public void code_3()
	{
		int[] myArray = {
[...]

PR #1074, opened 2016-12-24
Fix for a bug concerning multiples comments before a catch block

surli:
Multiple comments placed before a catch block were ignored, see #1073. This PR contains test to spot the bug and a fix.

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/JDTCommentBuilder.java
+++ b/src/main/java/spoon/support/compiler/jdt/JDTCommentBuilder.java
@@ -24,6 +24,7 @@ import spoon.reflect.code.CtBinaryOperator;
 import spoon.reflect.code.CtBlock;
 import spoon.reflect.code.CtBodyHolder;
 import spoon.reflect.code.CtCase;
+import spoon.reflect.code.CtCatch;
 import spoon.reflect.code.CtComment;
 import spoon.reflect.code.CtConditional;
 import spoon.reflect.code.CtIf;
@@ -187,6 +188,7 @@ class JDTCommentBuilder {
 		// visitor that inserts the comment in the element
 		CtInheritanceScanner insertionVisitor = new CtInheritanceScanner() {
 			private boolean isScanned = false;
+
 			@Override
 			public void scan(CtElement e) {
 				if (e == null) {
@@ -372,6 +374,14 @@ class JDTCommentBuilder {
 			@Override
 			public <T> void visitCtParameter(CtParameter<T> e) {
 				e.addComment(comment);
+			}
+
+			@Override
+			public void visitCtCatch(CtCatch e) {
+				if (comment.getPosition().getLine() <= e.getPosition().getLine()) {
+					e.addComment(comment);
+					return;
+				}
 			}
 		};
 		insertionVisitor.scan(commentParent);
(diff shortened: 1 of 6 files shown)


monperrus:
OK for me. @tdurieux (expert of the comment mode), OK for you?

surli:
Don't do the merge right now apparently there is more for issue #1073. I'll
check on it tomorrow.


Le dim. 25 déc. 2016 à 11:05, Martin Monperrus <notifications@github.com> a
écrit :

monperrus:
this is already two fixes here, right?.

we can still do another PR for the rest of #1073 (one PR = one fix).

surli:
Ok to do another PR for the rest of #1073 then.


Le dim. 25 déc. 2016 à 12:38, Martin Monperrus <notifications@github.com> a
écrit :

src/main/java/spoon/support/compiler/jdt/JDTCommentBuilder.java
```diff
@@ -373,6 +375,14 @@ public void visitCtAnonymousExecutable(CtAnonymousExecutable e) {
 			public <T> void visitCtParameter(CtParameter<T> e) {
 				e.addComment(comment);
 			}
+
+			@Override
+			public void visitCtCatch(CtCatch e) {
+				if (comment.getPosition().getLine() <= e.getPosition().getLine()) {
```
tdurieux: use sourceStart() instead of getLine

src/main/java/spoon/support/compiler/jdt/JDTCommentBuilder.java
```diff
@@ -373,6 +375,14 @@ public void visitCtAnonymousExecutable(CtAnonymousExecutable e) {
...
+					e.addComment(comment);
+					return;
+				}
```
tdurieux: call super.visitCtCatch()

src/main/java/spoon/support/compiler/jdt/JDTCommentBuilder.java
surli: Actually I used the same `getLine` which is used in `visitCtClass` which manage properly multiple inline comments. If I use here `sourceStart()` one comment is missing when writing something like:
```
try {}
// one comment
// two comment
catch {}
```

src/main/java/spoon/support/compiler/jdt/JDTCommentBuilder.java
surli: This super call is directly implemented in the `scan` override, do we really need it here? It is missing in all the other overriding `visitCtXX` methods...

src/main/java/spoon/support/compiler/jdt/JDTCommentBuilder.java
tdurieux: It is strange because getLine uses sourceStart() to perform the line number

surli: What I understand is that `sourceStart` starts at the same position than the first comment, which means the equality is respected and the comment is included (it was the previous implementation). But then the sourceStart of the second comment is greater than this sourcePosition and so it's not considered. 
Now with getLine either it considers all are on the same line, either it computes well the different lines using the line separator position, which is the case I think.

Issue #1070, opened 2016-12-22
feature: all you want for queries


pvojtechovsky:
Requirements/goals for new features of queries:

1. support of lazy foreach (better performance for large queries because no creation of useless intermediate lists)
    * <s>`CtQuery extends Iterator` would an elegant solution  but not possible because a Scanner is not iterable </s>
1. support for chaining of foreach
    * what should be the elements given for the next steps?
1. support for lazy evaluation in map for functions that returns an iterator (already works for functions returning Iterable).
1. support for queries reusable over multiple inputs (`CtQuery query = new CtBaseQueryImpl().map(...).map(...)` )
1. support for sending of results of one step directly to next step, without necessity to build helper lists.
```java
interface CtQuery { // already exists
  /** Performs an action on each element of the query results.
    * Implemented efficiently for CtScanner based queries */
  CtQuery forEach(CtQueryStep);
}

interface CtQueryStep {
  /** 
   * evaluates query step and applies mapping of one input element to many output elements.
   * The output elements are `returned` by call of `outputConsumer.accept(outputElement)`, 
   * which is called once for each outputElement
   */
   void evaluate(Object input, CtConsumer<Object> outputConsumer);
}

/**This interface is equivavlent to java 8 Consumer. 
We can remove it after we upgrade to java 8 */
interface CtConsumer<T> {
 void accept(T item);
}
```
The method `CtQuery#map(CtQueryStep)` is semantically equal to the existing method `CtQuery#map(CtFunction)`. The only difference is the way how result of mapping is delivered to the next query step.
* Implementation of CtFunction - returns result by `return objectBooleanOrList;`. This is optimal for mapping functions which returns 1 or few elements.
* Implementation of CtQueryStep- returns result by `outputConsumer.apply(item)`, which is called for each spoon model element, which has to be returned by this step. This is optimal for mapping functions, which returns many elements.

--------------------------------------------

I need a spoon query, which correctly returns all the references to an private or protected class field. It means that query must search for references in correct scope(s).

There was an idea how to implement it during development of PR #1018, but then we removed that code for simplicity reasons. 

@monperrus does it makes sense to create a new PR, which will reuse the same ideas again or do you know some better way?

monperrus:
> @monperrus <https://github.com/monperrus> does it makes sense to
> create a new P
Yes, go ahead.

We will first extensively discuss the design and contracts (javadoc) of:

* the new methods in existing interfaces

* the new interfaces (if any)

--Martin

[...]

PR #1076, opened 2016-12-26
Query improvements - CtQuery and CtBaseQuery

pvojtechovsky:
This PR is a proposed solution for the #1070.

It actually does not 100% fits to solution suggested by Martin. I suggest to have a look at this to make further discussion less abstract. So we can now discuss exact code lines and avoid other misunderstandings.


pvojtechovsky:
> But it's really important that we're along the same page, so that we could maintain your code, and so that we can write together high-quality documentation

@monperrus , I suggest this way to understand each other and to discuss future steps
1. to understand conceptual model and to agree on names of the concepts. 
  It should minimize misunderstandings.
2. to understand how these concepts are mapped to this PR
3. to discuss problems of this PR and Martin's suggestions how to do it different.

# Conceptual model
Concept model

Concept | Description
----------|--------------
MappingFunction | maps one input element to zero one or more output elements 
BaseQuery | is kind of MappingFunction, because same like MappingFunction it maps one input element to zero or more output elements. The difference is that this mapping is done by calling of other MappingFunctions, which are organized in a chain of mapping **steps**. The input of query is sent as input of first MappingFunction and all the outputs of that MappingFunction are sent as input of next MappingFunction ... recursively. Note that `BaseQuery extends MappingFunction` so we can use one BaseQuery as a step of another BaseQuery.
BoundQuery | is a wrapper around BaseQuery, which allows to bind the query to one input element and later evaluate this query. It is a syntactic API sugar, which makes possible to create queries from the spoon model elements. Note that it makes no sense to use BoundQuery as a step of another query, because such query is bound to constant input element.

# Mapping of Concepts to java classes
I suggest to postpone the discussion of the content of this table after we have agreed the **Conceptual model** and we converge on the names for these concepts. The table below is there just to make definition of concepts more exact in case of some ambiguity... you can have a look into java code of this PR.

Concept | Interface | Implementations
----------|--------------|-----------
MappingFunction | CtLazyFunction | CtFunction, OverridenMethodQuery, CtFieldReferenceQuery, ...
BaseQuery | CtBaseQuery | CtBaseQueryImpl
BoundQuery | CtQuery | CtQueryImpl

pvojtechovsky:
> IntercessionTest.testSettersAreAllGood:215 Your setter setInput in CtQuery don't have a generic type for its return type.

```java
public interface CtQuery extends CtQueryable.Step<CtQuery> {
	CtQuery setInput(Object... input);
```
How to fix it?

pvojtechovsky:
Was decided to continue with #1090. Closing this one

PR #1078, opened 2016-12-27
Fixes a typo in doc/gradle.md

dima74:


surli:
Thanks for spotting that one :)

PR #1090, opened 2017-01-05
feature: reusable and efficient CtQueries

pvojtechovsky:
This is the clone of #1076 where CtBaseQuery are merged in CtQuery.

These solutions are nearly compatible from client's point of view. 

Note: I have not added `CtQueryImpl#evaluate(Object, Consumer)` into `CtQuery`, because it is not needed on that high level of coding. The method is public and available in CtQueryImpl.

I would choose #1076, because the code is little bit cleaner (shorted classes = easier maintenance), but it is up to you. Both versions are acceptable for me.

The javadoc has to be read and fixed. There are till some wrong descriptions, which does not fit to latest design. But first choose which PR is adept for merge.


pvojtechovsky:
Finally we should modify Spoon factory and add there some method which creates CtQueryImpl. I suggest this:

factory.createQuery() - creates unbound query
factory.createQuery(Object input) - creates bound query
actually both methods will return CtQueryImpl instance. Please give me feedback or suggestion for factory method which you like. I do not care how that will exactly look, but I think the factory method should be there.

I will make another PR for that

pvojtechovsky:
I am finished with code changes. Please check the documentation. Then we can merge it.
Have a look at #1093 If you like it, I can add it as part of this PR too - as you wan ;-)

monperrus:
Really cool. I've just made a PR on your PR on the tests. If we're along the same line, I will proceed with doc.

> I can add it as part of this PR too - as you wan ;-)

No please don't, this PR already contains 3 baby steps, :-) (more on this later)

pvojtechovsky:
I am finished with documentation fixes. It is ready for merge from my point of view

monperrus:
I am finished too. I'm very happy that we've converged, thanks.

@surli could you review this?

I'd recommend to start with documentation (incl Javadoc), then tests,
then implementation.

monperrus:
Note that we'll merge after tomorrow's release.

doc/filter.md
```diff
@@ -39,35 +39,41 @@ list3 = rootPackage.filterChildren(
...
+```
+
+**Compatibility with existing filters** `CtQueryable#filterChildren(Filter)` is a filtering query that can be chained:
 
 ```java
 // collecting all methods of deprecated classes
 list2 = rootPackage
-    .filterChildren(new AnnotationFilter(Deprecated.class))
-    .filterChildren(new TypeFilter(CtMethod.class)).list();
+    .filterChildren(new AnnotationFilter(Deprecated.class)).list()
```
surli: Why don't you keep an example witch chaining queries here?

pvojtechovsky: I guess @monperrus removed it.

src/main/java/spoon/reflect/visitor/chain/CtQuery.java
```diff
@@ -16,56 +16,124 @@
...
  *
- * @param <O> the type of the element produced by this query
  */
-public interface CtQuery<O> extends CtQueryable {
+public interface CtQuery extends CtQueryable {
+
+	/**
+	 * Recursively scans all children elements of an input element.
+	 * The matched child element for which (filter.matches(element)==true) are sent to the next query step.
+	 * Essentially the same as {@link CtElement#getElements(Filter)} but more powerful, because it
```
surli: Just to be sure here: when you say filtered elements are sent to the next query step, it's not mandatory: you can just get the list of filtered elements through `list()` right? The meaning of your comment here is that the filterChildren can be used in the chain as it returns a CtQuery?

src/main/java/spoon/reflect/visitor/chain/CtQuery.java
```diff
@@ -16,56 +16,124 @@
...
+	 * @return the list of elements collected by the query.
+	 * @see #forEach(CtConsumer) for an efficient way of manipulating the elements without creating an intermediate list.
+	 */
+	<R extends Object> List<R> list();
 
 	/**
-	 * actually evaluates the query and returns all the produced elements collected in a List
+	 * Same as {@link CtQuery#list()}, but with static typing on the return type
+	 * and the final filtering, which matches only results, which are assignable from that return type.
+	 *
```
surli: So classes which are not assignable for that return type are just ignored and you cannot have a ClassCastException?

src/main/java/spoon/reflect/visitor/chain/CtQueryImpl.java
```diff
@@ -28,42 +28,21 @@
 import spoon.reflect.visitor.Filter;
 
 /**
- * Contains the default implementation of the generic {@link CtQuery} methods
+ * The facade of {@link CtBaseQuery} which represents a query bound to the {@link CtElement},
+ * which is the constant input of this query.
+ * It is used by {@link CtElement} implementations of {@link CtQueryable}.
+ *
+ * @param <O> - the type of element which is produced by the last step of the query
```
surli: I think you forgot to remove reference to param O.

src/main/java/spoon/reflect/visitor/chain/CtQueryImpl.java
```diff
@@ -90,126 +66,122 @@ public CtQueryImpl() {
 	 * @param input
 	 * @return this to support fluent API
 	 */
-	public CtQuery<O> addInput(O input) {
+	public CtQueryImpl addInput(Object... input) {
```
surli: Why do you return the impl here and not the interface?

src/main/java/spoon/reflect/visitor/chain/CtQueryImpl.java
```diff
@@ -218,127 +190,184 @@ public int getLength() {
...
+		@Override
+		public void accept(Object input) {
+			if (input == null) {
+				return;
+			}
+			stepIdx++;
+			try {
+				if (stepIdx <= steps.size()) {
+					//process next intermediate step
+					AbstractStep step = steps.get(stepIdx - 1);
```
surli: you defined a function `getStep()` for that.

src/main/java/spoon/reflect/visitor/chain/CtQueryImpl.java
```diff
@@ -218,127 +190,184 @@ public int getLength() {
...
-	private interface Step extends CtConsumer<Object> {
-		/**
-		 * @return next {@link CtConsumer}
-		 */
-		CtConsumer<Object> getNext();
-		/**
-		 * sets next {@link CtConsumer}
-		 */
-		void setNext(CtConsumer<Object> next);
+	private class CurrentStep implements CtConsumer<Object> {
```
surli: Found the name CurrentStep confusing as it's not a step (and does not extends AbstractStep) but more the CurrentConsumer.

src/main/java/spoon/reflect/visitor/chain/CtQueryImpl.java
```diff
@@ -28,42 +28,21 @@
 import spoon.reflect.visitor.Filter;
 
 /**
- * Contains the default implementation of the generic {@link CtQuery} methods
+ * The facade of {@link CtBaseQuery} which represents a query bound to the {@link CtElement},
```
surli: What is `CtBaseQuery` I assume you mean `CtQuery`?

src/main/java/spoon/reflect/visitor/chain/CtQueryImpl.java
```diff
@@ -218,127 +190,184 @@ public int getLength() {
...
+			} catch (ClassCastException e) {
+				onClassCastException(outputConsumer, e, input);
+				return;
 			}
 		}
 	}
 
 	/**
-	 * a step which calls Function. Implements contract of {@link CtQuery#map(CtFunction)}
+	 * a step which calls Function. Implements contract of {@link CtBaseQuery#map(CtFunction)}
```
surli: `CtBaseQuery` here again

src/main/java/spoon/reflect/visitor/chain/CtQueryImpl.java
surli: Ok it took me a while to really understand how the whole process work and I think you should improve the doc on this class as it is central to process the different steps. 
As far as I can see, this class plays as a kind of an orchestrator to move the step cursor forward, get the step, apply it and finally to call the output consumer. It could be great to add those information in the javadoc.

src/main/java/spoon/reflect/visitor/chain/CtQuery.java
pvojtechovsky: > the filterChildren can be used in the chain as it returns a CtQuery?

I do not understand

src/main/java/spoon/reflect/visitor/chain/CtQuery.java
pvojtechovsky: yes, exactly

src/main/java/spoon/reflect/visitor/chain/CtQueryImpl.java
pvojtechovsky: Because there are some interesting methods (evaluate, ...) which are available on CtQueryImpl only and if I cast query to CtQueryImpl once then I want to have CtQueryImpl after call of addInput too. It is correct implementation of fluent API

src/main/java/spoon/reflect/visitor/chain/CtQueryImpl.java
pvojtechovsky: you are right. I removed it now

src/main/java/spoon/reflect/visitor/chain/CtQueryImpl.java
pvojtechovsky: ok, I change the code so the getStep() is used

src/main/java/spoon/reflect/visitor/chain/CtQueryImpl.java
pvojtechovsky: javadoc added

src/main/java/spoon/reflect/visitor/chain/CtQueryImpl.java
pvojtechovsky: yes, CtQuery is correct. DONE

src/main/java/spoon/reflect/visitor/chain/CtQueryImpl.java
pvojtechovsky: yes, CtQuery is correct. DONE

surli:
Ok this is ready to merge for me. @pvojtechovsky could I close #1070 too or is it a pending discussion?

pvojtechovsky:
I have closed #1070.

surli:
Ok! Then let's merge and thanks for your work 👍

monperrus:
Thanks @surli for the final review.

PR #1093, opened 2017-01-08
Query factory

pvojtechovsky:
Implements
```java
launcher.getFactory().Query().createQuery().map...
launcher.getFactory().Query().createQuery(anInput).map...
```
This PR is based on #1090. I will rebase it after #1090 is finished and merged.

### DIFF
--- a/doc/filter.md
+++ b/doc/filter.md
@@ -87,11 +87,12 @@ rootPackage
 
 Finally, if the CtFunction returns and Iterable or an Array then each item of the collection/array is sent to next query step or result.
 
-**Query reuse**. Method `setInput` allows you to reuse the same query over multiple inputs.
+**Query reuse**. Method `setInput` allows you to reuse the same query over multiple inputs. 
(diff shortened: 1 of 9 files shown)


monperrus:
rebase?

monperrus:
Interesting. By the way: what does an bound and empty query return by default? nothing or all children?

pvojtechovsky:
> what does an bound and empty query return by default?

I have added tests for this contract.
* Bound query returns bound element
* Unbound or null bound query returns no element

monperrus:
Super.
Would you also add one or two sentences and remove the remaining `new CtQueryImpl()` in `filters.md`?

monperrus:
> [ERROR] /home/travis/build/INRIA/spoon/src/main/java/spoon/template/TemplateMatcher.java:455: error: @param name not found
> [ERROR] * @param templateRoot

OK now we catch doc errors again in Travis, but it's a nightmare to locate them.

pvojtechovsky:
Seems to be done from my side

PR #1095, opened 2017-01-09
fix(reference): fix reference ctconstructor

### LINKED ISSUE #1094: bug: constructor parameter references are not equal
In the model code like:
```java
public class ParamRefs {
	public ParamRefs(int param) {
		if(param == 0){}
	}
}
```
following spoon code
```java
CtClass<Object> aClass = launcher.getFactory().Class().get(ParamRefs.class);
CtParameter<?> parameter = aClass.getElements(new NameFilter<CtParameter<?>>("param")).get(0);
CtParameterReference<?> parameterRef1 = parameter.getReference();
CtParameterReference<?> parameterRef2 = aClass.getElements((CtParameterReference<?> ref)->ref.getSimpleName().equals("param")).get(0);
//this is cause of failure
assertEquals(parameterRef1.getDeclaringExecutable().getType(), parameterRef2.getDeclaringExecutable().getType());
//and therefore this fails too
assertEquals(parameterRef1, parameterRef2);

```
fails becuase the `CtParameterReference`s are not equal. I think it is bug. Could you please check it?
There is a [commit](https://github.com/pvojtechovsky/spoon/commit/4752741047d592479e63a6a09b42480d810156af) with failing test.
It is hard for me to fix it, because I do not know how it should behave.

tdurieux:
Fix #1094

Small PR to begin the week

### DIFF
--- a/src/main/java/spoon/reflect/factory/ExecutableFactory.java
+++ b/src/main/java/spoon/reflect/factory/ExecutableFactory.java
@@ -114,6 +114,7 @@ public class ExecutableFactory extends SubFactory {
 		} else if (e instanceof CtAnonymousExecutable) {
 			return createReference(((CtAnonymousExecutable) e).getDeclaringType().getReference(), e.getType().clone(), executableName);
 		}
+		// constructor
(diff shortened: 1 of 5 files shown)


pvojtechovsky:
Thank You @tdurieux!

PR #1098, opened 2017-01-10
Fix NullPointerException resulting from variable accesses in ctors.

retuxx:
`createVariableAccessNoClasspath` throws a NullPointerException if a
variable access is resolved for parameters in constructors as the parent
of constructor node is not available at this point.


retuxx:
Do you need tests for this PR?

surli:
Thanks for the PR.

> Do you need tests for this PR?

If you can do it quickly it's always better, at least to avoid regression.

retuxx:
Tests will be added in a minute.

retuxx:
Is this test suitable?

Issue #1099, opened 2017-01-10
element.toString() crashes with a NP in DefaultJavaPrettyPrinter.visitCtTypeReference()


arturbosch:
When analyzing elasticsearch we get a NP when calling element.toString().
As this has something to do with type references, I can't give you a concrete source file with this problem, but rather a part of the project as tar archive. (
[elasticsearch.tar.gz](https://github.com/INRIA/spoon/files/695853/elasticsearch.tar.gz)
) 

```
2017-01-10 11:37:50,692 ERROR  [de.unibremen.st.ismell.config.DetectorStarter] : Exception while running: GodClassDetector 
java.lang.NullPointerException
	at spoon.reflect.visitor.DefaultJavaPrettyPrinter.visitCtTypeReference(DefaultJavaPrettyPrinter.java:1609)
	at spoon.reflect.visitor.DefaultJavaPrettyPrinter.visitCtTypeReference(DefaultJavaPrettyPrinter.java:1578)
	at spoon.support.reflect.reference.CtTypeReferenceImpl.accept(CtTypeReferenceImpl.java:68)
	at spoon.reflect.visitor.DefaultJavaPrettyPrinter.scan(DefaultJavaPrettyPrinter.java:265)
	at spoon.reflect.visitor.DefaultJavaPrettyPrinter.visitCtLocalVariable(DefaultJavaPrettyPrinter.java:1101)
	at spoon.support.reflect.code.CtLocalVariableImpl.accept(CtLocalVariableImpl.java:49)
	at spoon.reflect.visitor.DefaultJavaPrettyPrinter.scan(DefaultJavaPrettyPrinter.java:254)
	at spoon.reflect.visitor.printer.ElementPrinterHelper.writeStatement(ElementPrinterHelper.java:140)
	at spoon.reflect.visitor.DefaultJavaPrettyPrinter.visitCtBlock(DefaultJavaPrettyPrinter.java:401)
	at spoon.support.reflect.code.CtBlockImpl.accept(CtBlockImpl.java:45)
	at spoon.reflect.visitor.DefaultJavaPrettyPrinter.scan(DefaultJavaPrettyPrinter.java:254)
	at spoon.reflect.visitor.DefaultJavaPrettyPrinter.visitCtMethod(DefaultJavaPrettyPrinter.java:1153)
	at spoon.support.reflect.declaration.CtMethodImpl.accept(CtMethodImpl.java:62)
[...]

PR #1102, opened 2017-01-10
fix of ElasticSearch access path problem #1099

pvojtechovsky:
Here is the test which can be used to reproduce the problem in #1099 

I will check how to fix it today or tomorrow

### DIFF
--- a/src/test/java/spoon/test/imports/ImportTest.java
+++ b/src/test/java/spoon/test/imports/ImportTest.java
@@ -403,8 +403,11 @@ public class ImportTest {
 		}
 		CtClass<?> mm = launcher.getFactory().Class().get("spoon.test.imports.testclasses2.AbstractMapBasedMultimap");
 		CtClass<?> mmwli = launcher.getFactory().Class().get("spoon.test.imports.testclasses2.AbstractMapBasedMultimap$WrappedList$WrappedListIterator");
-		assertTrue(mmwli.toString().indexOf("AbstractMapBasedMultimap<K, V>.WrappedCollection.WrappedIterator")>=0);
+		assertEquals("private class WrappedListIterator extends spoon.test.imports.testclasses2.AbstractMapBasedMultimap<K, V>.WrappedCollection.WrappedIterator {}",mmwli.toString());
 		assertTrue(mm.toString().indexOf("AbstractMapBasedMultimap<K, V>.WrappedCollection.WrappedIterator")>=0);
+
+		CtClass<?> mmwliother = launcher.getFactory().Class().get("spoon.test.imports.testclasses2.AbstractMapBasedMultimap$OtherWrappedList$WrappedListIterator");
+		assertEquals("private class WrappedListIterator extends spoon.test.imports.testclasses2.AbstractMapBasedMultimap<K, V>.OtherWrappedList.WrappedIterator {}",mmwliother.toString());
 		 								  
 	}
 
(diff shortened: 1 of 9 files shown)


pvojtechovsky:
There was problem with detection of visibility of nested private classes. Now it correctly decides that all private classes in scope of one top level class can access each other. 
This fix slightly changes behavior of some other tests, therefore I updated these existing tests too, so they are passing now.

src/test/java/spoon/test/imports/ImportTest.java
```diff
@@ -403,8 +403,8 @@ public void testNestedAccessPathWithTypedParameter() throws Exception {
 		}
 		CtClass<?> mm = launcher.getFactory().Class().get("spoon.test.imports.testclasses2.AbstractMapBasedMultimap");
 		CtClass<?> mmwli = launcher.getFactory().Class().get("spoon.test.imports.testclasses2.AbstractMapBasedMultimap$WrappedList$WrappedListIterator");
-		assertTrue(mmwli.toString().indexOf("AbstractMapBasedMultimap<K, V>.WrappedList.WrappedIterator")>=0);
-		assertTrue(mm.toString().indexOf("AbstractMapBasedMultimap<K, V>.WrappedList.WrappedIterator")>=0);
+		assertTrue(mmwli.toString().indexOf("AbstractMapBasedMultimap<K, V>.WrappedCollection.WrappedIterator")>=0);
+		assertTrue(mm.toString().indexOf("AbstractMapBasedMultimap<K, V>.WrappedCollection.WrappedIterator")>=0);
```
surli: Ok I think there's a behaviour change here, but maybe it's better, I'll try to clarify it. 
What we had before was:
```
abstract class AbstractMapBasedMultimap<K, V> {
    private class WrappedCollection {
        class WrappedIterator {        }
    }

    private class WrappedList extends AbstractMapBasedMultimap<K, V>.WrappedCollection {
        private class WrappedListIterator extends AbstractMapBasedMultimap<K, V>.WrappedList.WrappedIterator {        }
    }
}
```

What we have now is:
```
abstract class AbstractMapBasedMultimap<K, V> {
    private class WrappedCollection {
        class WrappedIterator {        }
    }

    private class WrappedList extends AbstractMapBasedMultimap<K, V>.WrappedCollection {
        private class WrappedListIterator extends AbstractMapBasedMultimap<K, V>.WrappedCollection.WrappedIterator {        }
    }
}
```
which means if now you redefine class `WrappedIterator` in `WrappedList` it won't be take into account. 
But the original source code is:
```
abstract class AbstractMapBasedMultimap<K, V> {
	private class WrappedCollection {
		class WrappedIterator {
		}
	}
	private class WrappedList extends WrappedCollection {
		private class WrappedListIterator extends WrappedIterator {
		}
	}
}
```
And I don't think the WrappedIterator here point to WrappedList.WrappedIterator... WDYT @tdurieux ?

tdurieux: It is hard to say what is the best solution.
I cannot find any behavior change.

pvojtechovsky: > which means if now you redefine class WrappedIterator in WrappedList it won't be take into account.

We can make a test for that. As long as there is no ambiguous name, the algorithm can use arbitrary access path. But once there are two WrappedIterator instances in different scope, then we will see whether it behaves correctly

surli: It could be great to add the test: we check with @tdurieux and it will work as defined in the test. However if you build the model and then you put dynamically a new class `WrappedIterator` in `WrappedList` the `WrappedListIterator` continue to extend the old `WrappedIterator` and not the newly one.
There is a real discussion here on what should be the behaviour, but I think it's out of the scope of this PR.

Issue #1104, opened 2017-01-11
Bug: javadoc not building using JDK 8


surli:
See the following stackoverflow post: http://stackoverflow.com/questions/15886209/maven-is-not-working-in-java-8-when-javadoc-tags-are-incomplete

surli:
Fixed in #1105

PR #1105, opened 2017-01-11
doc: Fix issues related to javadoc in java 8 #1104

surli:
Fix all errors encountered while launching ```mvn javadoc:javadoc``` on Spoon. Lot of warnings are still remaining but don't block the goal.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -241,6 +241,11 @@
       <artifactId>commons-io</artifactId>
       <version>1.4</version>
     </dependency>
+    <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-lang3</artifactId>
+      <version>3.5</version>
+    </dependency>
   </dependencies>
 
   <build>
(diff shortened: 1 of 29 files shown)


pom.xml
```diff
@@ -241,6 +241,11 @@
       <artifactId>commons-io</artifactId>
       <version>1.4</version>
     </dependency>
+    <dependency>
```
tdurieux: it is a test dependency.
Please specify the scope of the dependency.

tdurieux:
Maybe it is a good idea that spoon encodes and decodes the comments.
What do you think?

surli:
> Maybe it is a good idea that spoon encodes and decodes the comments.

Encodes when building model and decodes when pretty-printing?
I'm not sure about the benefits of doing that, but still it should be in another issue/PR.

pom.xml
surli: Done.

PR #1107, opened 2017-01-11
travis: Remove the -Xdoclint parameter in pom.xml fix #1106

### LINKED ISSUE #1106: Travis don't fail on javadoc errors in Java 8
To remain consistent with the release process, Travis should fail when java 8 related javadoc issues occured.

surli:
Remove the -Xdoclint parameter in pom.xml to show errors related to javadoc using java 8 rules. Travis should fail when those errors occured. Fix #1106

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -567,7 +567,6 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-javadoc-plugin</artifactId>
         <configuration>
-          <additionalparam>-Xdoclint:-missing</additionalparam>
           <quiet>true</quiet>
         </configuration>
       </plugin>


monperrus:
> Travis should fail when those errors occured. 

are you sure, did you try? it seems this is only an output setting. what about the "quiet"?

surli:
> are you sure, did you try? it seems this is only an output setting. what about the "quiet"?
Actually I tried: when I proposed this PR, #1105 wasn't merged yet and I obtained a failed build for this PR: I launched it back after the merge to be sure it has the correct behaviour. 

I assume the quiet is here to display only few messages, but it does show the warnings.

surli:
@monperrus as you can see it works: #1090 introduces some javadoc JDK8 errors ;)

monperrus:
well-done :-)

so we need a new PR on the doc, before being able to merge this one, right?

surli:
> so we need a new PR on the doc, before being able to merge this one, right?
I'm doing it. I push it in 5 minutes.

PR #1108, opened 2017-01-11
Set the type of a parameter declared in a lambda to Object instead of ommitting them when unknown.

retuxx:
This PR also contains a fix for a minor bug introduced in #1098.


retuxx:
Any chances that this PR will be in 5.5.0?

retuxx:
Thank you!

Issue #1080, opened 2016-12-29
feature: template parameters collecting TemplateMatcher


pvojtechovsky:
TemplateMatcher actually returns the list of model elements, which matches the Template T1.
It would be nice if TemplateMatcher would not return only list of matching elements, but for each matched element it might return Template parameters, which can be used to build that element using Template T1.
I am speaking about inverse process to the Template T1.apply(templateParameters TP1) creates an element E1_1.

What it would be good for? Replacing of one design pattern implementation by another implementation.

Example:
code like
```java
for(Iterator iter=collection.iterator();iter.hasNext();) {
ItemType item = iter.next();
...bodyOfLooop...
}
```
might be automatically replaced by modern `for(ItemType item : cllection){...bodyOfLoop...)`

Algoritm:
1) search for all accurences of oldLoop using TemplateMatcher and `templateOldLoop`
2) for each occurence check if moodern loop is compatible
3) for each compatible `oldLoop` create model element `newLoop` of modern loop using `templateModernLoop`+actual template parameters
4) replace oldLoop by newLoop.

And not only that ;-)

monperrus:
It's a very interesting improvement of template matchers.

monperrus:
actually, there a natural way of getting the matched expression: taking the template parameter name:
```
public TemplateParameter<Collection<?>> _col_;
...
if (_col_.S().size() > 10)
```

then, one gets the value of match as follows:
...
interface TemplateMatcher{
CtElement getMatchedElement(String name)
}

aMatcher.getMatchedElement("_col_") ->the matched expression
```

**Comment by pvojtechovsky** [comment]
I searched through existing code and have found no `getMatchedElement`, so you probably speak about future API design. Right?

In Future I would like to use `TemplateMatcher` as query `Filter`. It would be change like this:
```java
class TemplateMatcher implements Filter {
  public boolean match(CtElement element) {
    return this.match(element, this.templateRoot);
  }
... current code of TemplateMatcher ...
}
```

So it can be used in query to search for elements, which matches the template. In my case I expect that TemplateMatcher will usually match more then one element, so I need TemplateMatcher API, which will give me access to actual template parameters for each element.

S1) May be the code based on your `TemplateMatcher.getMatchedElement` might look like this:
```java
TemplateMatcher matcher = new TemplateMatcher(template);
rootPackage.filterChildren(matcher).map(match->{
  //this code is called immediately after `matcher` matches the `match` element
  //so may be we can access matcher's current state like this
[...]

PR #1110, opened 2017-01-11
Template matcher implemented as Filter

pvojtechovsky:
As discussed in #1080, there is `TemplateMatcher` implemented as `Filter`

### DIFF
--- a/doc/matcher.md
+++ b/doc/matcher.md
@@ -30,7 +30,7 @@ To define a template matcher one must:
 
 1. specify the "holes" of the template matcher 
 1. write the matcher in a dedicated method
-1. instantiate TemplateMatcher and call method `find`.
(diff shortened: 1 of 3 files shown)


monperrus:
cool feature, and incredible simplification of the code.

even if the new feature is indirectly tested, would you add an explicit test case with "contract: the query API supports template matcher acting as a filter"? thanks.

pvojtechovsky:
Good idea. See modified `TemplateTest#testTemplateMatcherMatchTwoSnippets()`.

monperrus:
and a line of documentation in `doc/matcher.md` would also be great. (new feature = new doc). Thanks.

monperrus:
Thanks a lot.

PR #1119, opened 2017-01-13
[WIP] Add method to determine whether a type overrides a method.

retuxx:
I will add some more tests later.


pvojtechovsky:
I do not understand the use case. Why do you need that?

Why do you care about @Overrides annotation? Why the behavior changes depending on presence/absence of this annotation?

Why do you check only one level of super type or super interfaces? I guess it should search recursivelly through all super classes/interfaces. Is it intentional?

Suggestion: What about returning CtMethod instead of boolean? Such function would be more useful from my point of view.

retuxx:
I found the method `CtExecutable#getOverridingExecutable` which is more handy for my case. Thus, I'm closing this pr.

PR #1128, opened 2017-01-18
Fixes #1125.

### LINKED ISSUE #1125: CtTypeReferenceImpl#getSuperClass() fails in noClasspath mode
The method has the following implementation:

```java
@Override
public CtTypeReference<?> getSuperclass() {
	CtType<T> t = getDeclaration();
	if (t != null) {
		return t.getSuperclass();
	} else {
		Class<T> c = getActualClass();
		Class<?> sc = c.getSuperclass();
		if (sc == null) {
			return null;
		}
		return getFactory().Type().createReference(sc);
	}
}
```

As you can see, it doesn't check for noClasspath and, thus, may fail when calling `getActualClass()`.

retuxx:


src/main/java/spoon/support/reflect/reference/CtTypeReferenceImpl.java
```diff
@@ -465,7 +465,14 @@ private void handleParentNotFound(SpoonClassNotFoundException cnfe) {
 	public Collection<CtExecutableReference<?>> getDeclaredExecutables() {
 		CtType<T> t = getDeclaration();
 		if (t == null) {
-			return RtHelper.getAllExecutables(getActualClass(), getFactory());
+			try {
+				return RtHelper.getAllExecutables(getActualClass(), getFactory());
+			} catch (final SpoonClassNotFoundException e) {
+				if (getFactory().getEnvironment().getNoClasspath()) {
+					return Collections.emptyList();
```
retuxx: I was not sure if I should return an empty list or null. When changing this to null, there is another method to fix as it uses a for each loop without checking the type returned by this method. However, I used null since CtTypeImpl instantiates `typeMembers` with an empty list indicating that an unknown type has no methods.

surli: Returning EmptyList as you did here is better.

Issue #1120, opened 2017-01-14
feature: Find first element using query


pvojtechovsky:
The problems like #1119 might be implemented using queries, but we need two extra features in queries:
* to tell the query evalutor to stop evaluation early - when searched element is found
* (optionaly) support for recursive mapping. It means for example query step N can send some values back to input of query step N-1.

monperrus:
> to tell the query evalutor to stop evaluation early - when searched element is found

That would indeed be a special and interesting kind of query. Not sure how we would implement this. Maybe only with a new factory method `createFindFirstQuery`

pvojtechovsky:
What about `T CtQuery#first()` and code like
```java
CtElement firstResult = element.map(...).filterChildren(...)...first();
```

surli:
Could you elaborate more on the semantic of your proposed feature: 
  * is it supposed to stop each step whenever one element is found satisfying the step query?
  * or is it supposed to stop only intermediate steps whenever one element is found and execute entirely the final step?

The signature you propose:
> What about T CtQuery#first() and code like
> ```CtElement firstResult = element.map(...).filterChildren(...)...first();```

seems to only operate to the second semantic.

> (optionaly) support for recursive mapping. It means for example query step N can send some values back to input of query step N-1.

I have the feeling this can be a really error prone feature which can easily create infinite loop... Do you think of many usecases for this feature?

pvojtechovsky:
Originally it is supposed to stop evaluation of any other query step immediately after first result is produced by last query step.
The behavior of this new code
```java
CtElement firstResult = element.map(...).filterChildren(...)...first();
```
is equal to this old code
```java
List allResults = element.map(...).filterChildren(...)...list();
CtElement firstResult = allResults.isEmpty()?null:allResults.get(0);
```
but faster and shorter, because query knows that other results are not needed so it stops the evaluation and returns the first found element.
So it was my origin intention. I already found a use case for that. The refactoring validation algorithm has to report problem when there exists e.g. a conflict in the name. Such query has to search for first conflict and then can finish. In this case I even do not need any results, just to know whether query returns any element or none.


But you mentioned 
> is it supposed to stop each step whenever one element is found satisfying the step query?

It sounds interesting too, but I do not know yet what would be the use case behind.

>> support for recursive mapping
>
> this can be a really error prone feature 

[...]

PR #1138, opened 2017-01-19
feature: early terminating query

pvojtechovsky:
Solution for #1120


monperrus:
it looks good, thanks. 

do we need a public interface for CtQueryContext? Would a private class be OK instead?

pvojtechovsky:
> public interface for CtQueryContext? Would a private class be OK instead?

I agree, that for now it can be private. Later I have some plans with it, but it is not mature enough, so it is better to hide it. 
I will do it.

pvojtechovsky:
Looks like this is done from my point of view

monperrus:
thanks a lot for this series of PRs.

pvojtechovsky:
Thanks a lot for managing this nice project and for Your time spent on reviews!

PR #1140, opened 2017-01-20
feature CtQuery#filterChildren(null)

pvojtechovsky:
Support for calling filterChildren without filter. It returns all visited elements.

### DIFF
--- a/src/main/java/spoon/reflect/visitor/chain/CtQuery.java
+++ b/src/main/java/spoon/reflect/visitor/chain/CtQuery.java
@@ -47,7 +47,8 @@ public interface CtQuery extends CtQueryable {
 	 * The elements which throw {@link ClassCastException} during {@link Filter#matches(CtElement)}
 	 * are considered as **not matching**, ie. are excluded.
 	 *
-	 * @param filter used to filter scanned children elements of the AST tree
+	 * @param filter used to filter scanned children elements of the AST tree.
+	 * 	If null then all children elements pass to next step.
 	 * @return this to support fluent API
 	 */
 	@Override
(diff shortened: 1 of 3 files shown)


pvojtechovsky:
What is the problem? I do not know how to fix it.

pvojtechovsky:
>> Filter filter = new ...;
>> element.map(new ParameterScopeQuery).map(e->filter.matches(e))...
>
> Martin wrote: I like it, esp with lambdas, it's clear and explicit. Let's first aim at this simplest version.

There is one more problem when Filter is used in existing `CtQuery#map(CtFunction)`. It is that query engine will throw ClassCastException, when Filter input **type** does not matches. 

So you have to write
```java
Filter filter = new ...;
element.map(new ParameterScopeQuery)
  .map(e->filter.matches(e)).failurePolicy(QueryFailurePolicy.IGNORE)...
```

monperrus:
super atomic PR, thanks.

Would you add one line of Javadoc in   filterChildren to document the contract?

monperrus:
> It is that query engine will throw ClassCastException, when Filter input type does not matches.

which is the default behavior we want, because it's very explicit at runtime, right?

pvojtechovsky:
>> It is that query engine will throw ClassCastException, when Filter input type does not matches.
> 
> which is the default behavior we want, because it's very explicit at runtime, right?

right. We want to throw CCE on standard mapping functions, because otherwise it is very hard to found "Why my query does not work and returns nothing..."

pvojtechovsky:
Done from my point of view

monperrus:
rebase?

pvojtechovsky:
Done.

PR #1141, opened 2017-01-20
review FieldReferenceFunction, FieldScopeFunction, SubtypeFilter

pvojtechovsky:
This PR contains FieldReferenceFunction, FieldScopeFunction, SubtypeFilter.
The main purpose is FieldReferenceFunction, which returns all the references to the field. Other two are helper classes which might be useful for others so they are public.

It depends on these PRs
* CtQuery#filter(Filter) #1142
* CtQuery#filterChildren(null) #1140

### DIFF
--- a/src/main/java/spoon/reflect/visitor/filter/FieldScopeFunction.java
+++ b/src/main/java/spoon/reflect/visitor/filter/FieldScopeFunction.java
@@ -56,13 +56,13 @@ public class FieldScopeFunction implements CtConsumableFunction<CtField<?>> {
 		}
 	}
 	protected void searchForPrivateField(CtField<?> field, CtConsumer<Object> outputConsumer) {
-		//private field can be referred from the scope of current top level type only
+		//private field can be referred from the scope of current top level type only and children
 		field.getTopLevelType()
 			.filterChildren(null)
 			.forEach(outputConsumer);
 	}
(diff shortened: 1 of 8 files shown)


pvojtechovsky:
Only this PR might need more tests, which would check package and public visible class fields in different packages.

src/main/java/spoon/reflect/visitor/filter/FieldScopeFunction.java
```diff
@@ -0,0 +1,87 @@
...
+		}
+	}
+	protected void searchForPrivateField(CtField<?> field, CtConsumer<Object> outputConsumer) {
+		//private field can be referred from the scope of current top level type only
+		field.getTopLevelType()
+			.filterChildren(null)
+			.forEach(outputConsumer);
+	}
+	protected void searchForProtectedField(CtField<?> field, CtConsumer<Object> outputConsumer) {
+		//protected field can be referred from the scope of current top level type only
```
surli: > from the scope of current top level type only

and children

surli:
Seems good to me. I'm waiting for the new tests before merging.

pvojtechovsky:
The tests of field visibility in other classes and packages were added. It is finished from my point of view.

pvojtechovsky:
Looks like Travis is ill.

PR #1142, opened 2017-01-21
feature CtQuery#select(Filter)

pvojtechovsky:
Actually Filter can be used as mapping function this way:
```java
Filter filter = new SomeFilter();
element.map(new CtFunction<R, Boolean>() {
	@Override
	public Boolean apply(R input) {
		return filter.matches(input);
	}
}).failurePolicy(QueryFailurePolicy.IGNORE);
```
There are these problems:
* I needed 2 minutes to write it and 1h to found that failurePolicy must be added too - and as a author I know the queries the best - so current way how legacy filters can be used is ugly and error prone.
* the failurePolicy is set to whole query, but it should be set only to the query step which uses the filter. Internally there is already a `CtQueryImpl#stepFailurePolicy`, which should be therefore called instead, but then the code would be even more ugly.

So this PR allows to write:
```java
element.filter(new SomeFilter());
```

### DIFF
--- a/src/main/java/spoon/reflect/visitor/chain/CtQuery.java
+++ b/src/main/java/spoon/reflect/visitor/chain/CtQuery.java
@@ -53,6 +53,16 @@ public interface CtQuery extends CtQueryable {
 	@Override
 	<R extends CtElement> CtQuery filterChildren(Filter<R> filter);
 
+	/**
+	 * The matched element for which (filter.matches(element)==true) is sent to the next query step.
+	 *
+	 * The elements which throw {@link ClassCastException} during {@link Filter#matches(CtElement)}
+	 * are considered as **not matching**, ie. are excluded.
+	 *
+	 * @param filter used to detect if input element can pass to next query step
+	 * @return this to support fluent API
+	 */
+	<R extends CtElement> CtQuery select(Filter<R> filter);
 
 	/**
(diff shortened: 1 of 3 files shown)


src/main/java/spoon/reflect/visitor/chain/CtQueryable.java
```diff
@@ -38,6 +38,11 @@
 	<R extends CtElement> CtQuery filterChildren(Filter<R> filter);
 
 	/**
+	 * @see CtQuery#filter(Filter)
+	 */
+	<R extends CtElement> CtQuery filter(Filter<R> filter);
```
monperrus: do we need it in CtQueryable?

monperrus:
you convinced me. 

However, I would name the new method `select` (as in some other languages) to avoid any ambiguity with `filterChildren`.

Would you add an explanation in `filter.md`?

src/main/java/spoon/reflect/visitor/chain/CtQueryable.java
pvojtechovsky: you are right. We do not need it. Thanks!

pvojtechovsky:
Done

> documentation in filter.md

Yes, I will do it, but I suggest to add it later in another PR. Because 
* there is conflict with PR `early terminating query` #1138, which already changes filter.md
* there are other PRs which are waiting for this PR.

PR #1144, opened 2017-01-21
review LocalVariableReferenceFunction, LocalVariableScopeFunction

pvojtechovsky:
This PR contains LocalVariableReferenceFunction, LocalVariableScopeFunction.
The main purpose is LocalVariableReferenceFunction, which returns all the references to the local variable. The second on is helper classes which might be useful for others so it is public.

It depends on PR `CtQuery#filterChildren(null)` #1140 

### DIFF
--- a/src/main/java/spoon/reflect/visitor/filter/LocalVariableScopeFunction.java
+++ b/src/main/java/spoon/reflect/visitor/filter/LocalVariableScopeFunction.java
@@ -52,7 +52,7 @@ public class LocalVariableScopeFunction implements CtConsumableFunction<CtLocalV
 	public void apply(CtLocalVariable<?> localVariable, CtConsumer<Object> outputConsumer) {
 		CtStatementList statements = localVariable.getParent(CtStatementList.class);
 		if (statements == null) {
-			//cannot search for parameter references of parameter which has no executable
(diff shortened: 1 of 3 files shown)


src/main/java/spoon/reflect/visitor/filter/LocalVariableScopeFunction.java
```diff
@@ -0,0 +1,71 @@
...
+public class LocalVariableScopeFunction implements CtConsumableFunction<CtLocalVariable<?>> {
+
+	public LocalVariableScopeFunction() {
+	}
+
+	@Override
+	public void apply(CtLocalVariable<?> localVariable, CtConsumer<Object> outputConsumer) {
+		CtStatementList statements = localVariable.getParent(CtStatementList.class);
+		if (statements == null) {
+			//cannot search for parameter references of parameter which has no executable
```
surli: Wrong comment I think. But don't you want to tests emptyness of the statementList here too?

pvojtechovsky: comment was wrong. Thanks. The check for statement.isEmpty() is not needed because it later searches in the statements list and throws exception in case of empty collection too. So this case is handled too.

PR #1145, opened 2017-01-21
review CatchVariableReferenceFunction, CatchVariableScopeFunction

pvojtechovsky:
This PR contains CatchVariableReferenceFunction, CatchVariableScopeFunction.
The main purpose is CatchVariableReferenceFunction, which returns all the references to the catch variable. The second on is helper classes which might be useful for others so it is public.

It depends on PR CtQuery#filterChildren(null) #1140


surli:
> The tests will be added later, because they share a lot with tests of #1136

It seems you already have your tests, in the similar way than #1136. Do you want to add more tests or is it only conflict solving?

pvojtechovsky:
I do not need more tests for this PR. It was conflict solving.

PR #1153, opened 2017-01-30
review: feature: ParentFunction

pvojtechovsky:
Implements new mapping functions, which visits all the parents of current element. 
It is needed by to LocalVariableScopeFunction to detect scope of local variable, by searching for a parent which matches some rules... will be used in next PR, which fixes LocalVariableScopeFunction for variables used in CtLoop expressions and CtTryWithResource.

Note: I have one question to the scanning of all parents. It is "Which parent is the top parent?"
If I visit parents until the getParent()==null, then I get to helper annonymous class based on CtElementImpl, which is declared in CtModelImpl$CtRootPackage$1. I guess it is not correct ... Your opinion please?
The test actually fails because the scanning does not end on RootPackage, but one level below...

### DIFF
--- a/src/main/java/spoon/reflect/CtModelImpl.java
+++ b/src/main/java/spoon/reflect/CtModelImpl.java
@@ -51,7 +51,12 @@ public class CtModelImpl implements CtModel {
 				public CtElement getParent() throws ParentNotInitializedException {
 					return null;
 				}
-							});
+
+				@Override
+				public Factory getFactory() {
+					return CtRootPackage.this.getFactory();
+				}
+			});
 		}
 
 		@Override
--- a/src/main/java/spoon/reflect/visitor/filter/ParentFunction.java
+++ b/src/main/java/spoon/reflect/visitor/filter/ParentFunction.java
@@ -1,0 +1,56 @@
+/**
+ * Copyright (C) 2006-2017 INRIA and contributors
+ * Spoon - http://spoon.gforge.inria.fr/
+ *
+ * This software is governed by the CeCILL-C License under French law and
+ * abiding by the rules of distribution of free software. You can use, modify
+ * and/or redistribute the software under the terms of the CeCILL-C license as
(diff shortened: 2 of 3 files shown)


surli:
> If I visit parents until the getParent()==null, 

why don't you use `getParent() == factory.getModel().getRootPackage()` as condition to stop?

pvojtechovsky:
I know several ways how to do it different/better. But I did not know what is the correct behavior from the spoon design point of view. So I waited for a feedback of an experienced spooner.
So, please confirm that parent visiting should stop on factory.getModel().getRootPackage() - including that node. And I will do so. Or discuss it and let me know what is correct. Thank You!

tdurieux:
> Note: I have one question to the scanning of all parents. It is "Which parent is the top parent?"
If I visit parents until the getParent()==null, then I get to helper annonymous class based on CtElementImpl, which is declared in CtModelImpl$CtRootPackage$1. I guess it is not correct ... Your opinion please?
The test actually fails because the scanning does not end on RootPackage, but one level below...

If I correctly remember, we only created the CtRootPackage to "store" all elements of a model. It does not represent any Java concept.
It is probably better to include CtRootPackage in order to match the behavior of `getParent(Filter<E> filter)`.
Even though there is no interest  to visit it.

msteinbeck:
> If I correctly remember, we only created the CtRootPackage to "store" all elements of a model. It does not represent any Java concept.

Isn't it the package of files without any package declaration?

tdurieux:
> Isn't it the package of files without any package declaration?

Yes, and also for the parent of declared packages.
Before CtRootPackage, we identified the default package by comparing its name to `TOP_LEVEL_PACKAGE_NAME`

pvojtechovsky:
Thanks to everybody for good feedback. I modified the code and test so it expects that last visited parent element is CtModelImpl$CtRootPackage.

surli:
Ok it seems good to me.

surli:
Test revapi.

Issue #1159, opened 2017-02-01
bug: Lambda parameter references are not equal


pvojtechovsky:
The code:
```java
Consumer<Integer> c = (field)->{
	field=1;
};
```
contains CtParameter field in Lambda Expression and CtParameterReference in body of lambda expression. If I try to check whether these parameter references are equal then they are not
```java
CtParameter<?> param = (CtParameter<?>)lambda.getParameters().get(0);
CtParameterReference paramRef = lambda.filterChildren(new TypeFilter<>(CtParameterReference.class)).first();
//it fails here
assertTrue(param.getReference().equals(paramRef));
```

See failing test here https://github.com/pvojtechovsky/spoon/commit/b2f771622ed24d69165bcb51b521e5f30cd9b2b7

Issue #1163, opened 2017-02-02
bug: CtMethod#isDefaultMethod() returns false for default methods


pvojtechovsky:
This code fails, even if `Consumer#andThen` method is default.
```
assertTrue(factory.Class().get(java.util.function.Consumer.class)
   .getMethodsByName("andThen").get(0).isDefaultMethod());
```

pvojtechovsky:
Fixed by #1165

PR #1164, opened 2017-02-02
feature: CtLambda#getMethod() and fix of #1159

pvojtechovsky:
This PR fixes issue #1159. The issue #1159 is caused by wrong return type of CtExecutableReference created for lambda. Therefore this code failed too:
```java
assertTrue(lambdaRef.getDeclaringExecutable().getType()
     .equals(paramRef.getDeclaringExecutable().getType()));
``` 

Now there is a question whether solution of this PR is correct.

In current spoon model the interface `CtLambda` represents both:
* `CtExpression`, which returns an instance, which implements a type defined by result of `CtLambda.getType()`
* `CtExecutable`, which represents an executable element. It means it has parameters, body, thrownTypes ... and **return type** 

The problem was that **return type** of executable element was not accessible. Therefore during creation of CtExecutableReference of lambda expression, there was used wrong return type of that executable (see change in `ExecutableFactory`).

Now the  issue #1159 is fixed - the parameter references are equal.
... but I have found another issue #1163, which causes that `CtLambdaImpl#getExecutableType` actually does not work correctly.

### DIFF
--- a/src/main/java/spoon/reflect/code/CtLambda.java
+++ b/src/main/java/spoon/reflect/code/CtLambda.java
@@ -66,7 +66,7 @@ public interface CtLambda<T> extends CtExpression<T>, CtExecutable<T> {
 	 * Must be defined as a non-default method in an interface, e.g. Consumer.accept().
 	 */
 	@DerivedProperty
-	<R> CtMethod<R> getMethod();
+	<R> CtMethod<R> getOverriddenMethod();
 
 	/**
 	 * Sets the expression in the body of the lambda. Nothing will change
--- a/src/main/java/spoon/reflect/code/CtLambda.java
+++ b/src/main/java/spoon/reflect/code/CtLambda.java
@@ -17,7 +17,9 @@
 package spoon.reflect.code;
 
 import spoon.reflect.declaration.CtExecutable;
+import spoon.reflect.declaration.CtMethod;
 import spoon.reflect.reference.CtTypeReference;
+import spoon.support.DerivedProperty;
 import spoon.support.UnsettableProperty;
 
 import java.util.Set;
@@ -58,6 +60,12 @@ public interface CtLambda<T> extends CtExpression<T>, CtExecutable<T> {
 	 * of statements.
 	 */
 	CtExpression<T> getExpression();
+
+	/**
+	 * @return a method which is implemented by this lambda expression
+	 */
+	@DerivedProperty
+	<R> CtMethod<R> getMethod();
 
 	/**
 	 * Sets the expression in the body of the lambda. Nothing will change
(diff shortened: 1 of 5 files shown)


monperrus:
I'd say there is a solution without introducing a new method.

What about setting the type to the correct value when it's available or simply overriding `getType` in `CtLambdaImpl`.

pvojtechovsky:
CtLambdaImpl#getType is already used for different purpose and makes sense too. It cannot be overriden.

> What about setting the type to the correct value when it's available

I do not understand. Do you mean to have not derived property CtLambdaImpl#executableType, which is set during compilation of model?

pvojtechovsky:
an idea: to replace `CtLambda#getExecutableType()`, by
A) ctLambda.getExecutable().getType()
B) ctLambda.getExecutableMethod().getType()
C) ctLambda.getMethod().getType()
Note: all these methods are derived From the result of ctLambda.getType().map(new SAMMethodFilter()).list() ...

I suggest that, because the new method getExecutableType is too special. More generic method which returns CtMethod, which is implemented by result of lambda expression, is better.

pvojtechovsky:
I replaced `CtLambda#getExecutableType()` by `CtLambda#getMethod()`, which is more useful.

If somebody needs the getExecutableType, then s/he can call ctLambda.getMethod().getType(). It is the same.

I think this code makes sense. The client needs to know, which method is called by lambda expression. 

What about the method name `CtLambda#getMethod()`. Is it good name, or you have better one?

Note: the tests fails because fix of #1163 is not part of this PR (to keep PR simple). After #1165 is merged this PR will pass tests too.

src/main/java/spoon/reflect/code/CtLambda.java
```diff
@@ -60,6 +62,12 @@
 	CtExpression<T> getExpression();
 
 	/**
+	 * @return a method which is implemented by this lambda expression
```
monperrus: @return the method that this lambda expression implements. Must be defined as a non-default method in an interface, eg Consumer.accept().

monperrus:
One question: Is `lambda.getMethod()` equivalent to `lambda.getRefarence().getOverridingExecutable().getDeclaration()`?

pvojtechovsky:
>  Is lambda.getMethod() equivalent to lambda.getReference().getOverridingExecutable().getDeclaration()?

I made a test and `lambda.getReference().getOverridingExecutable()` is actually null.

I do not know whether it should be equivalent or not. Probably be yes. 

I just know that the expression `lambda.getReference().getOverridingExecutable().getDeclaration()` cannot be used instead of `lambda.getMethod()`, because `lambda.getMethod()` is used internally in `lambda.getReference()`, so it would end with stack overflow.

monperrus:
OK. This really makes sense, and I'm fine with adding this method to lambda, but with a more specific name.

Do you prefer `getOverridenMethod`or `getImplementedMethod` or `getImplementedInterfaceMethod`or something similar?

pvojtechovsky:
I like both getOverridenMethod or getImplementedMethod. 
getOverridenMethod is probably better, because it's meaning is same like CtExecutable#getOverridenMethod. I will rename it

monperrus:
OK, good.

I did a spelling mistake: getOverriddenMethod  with double r and double d

PR #1165, opened 2017-02-02
fix: Consumer#andThen isDefaultMethod()

pvojtechovsky:
This PR contains failing test, which reproduces issue #1163


pvojtechovsky:
The `java.lang.reflect.Method#isDefault` is not known by java 7 runtime. How do you compile java 8 features?

pvojtechovsky:
The java 8 `Method#isDefault()` is called by java reflection now. So it will work on both Java 7 and 8.
This PR is done from my point of view.

PR #1178, opened 2017-02-09
fix: bug regression with lambda expression when extending Function

surli:
Ignore static modifiers in `CtLambda:getOverriddenMethod()`


monperrus:
the test case does not contain static method, is that correct?

surli:
> the test case does not contain static methods, is that correct?

Yes, I reproduced the bug obtained with RxJava: the static method is inherited from Function.

monperrus:
OK. ready for merge?

surli:
For me it's ready yes.

pvojtechovsky:
This PR is ok for me too

monperrus:
thanks!

PR #1180, opened 2017-02-09
review: introduction of CtScannerFunction

pvojtechovsky:
As I mentioned in #1005 it is sometime needed to 
* R1 to know when query step of type filterChildren enter/exit an element on one selected query step.
* R2 to be able to skip processing of children of current element 
* R3 to be able to skip processing of siblings and children of current elements
* R4 to be able to make mapping functions aware of query processing state (actually mainly whether query is terminated)

There is already a Query Execution Context (QEC) impelemented by class `CtQueryImpl$CurrentStep`. 

There are needed these two things to implement this PR:
1.  to get access to the QEC, or to provide own extension of that context
2.  to define API of that QEC, which can be used to
    * drive execution of query 
    * or to get query processing state.
      * the list of all Functions of the query
      * the index of actually processed Function
      * the depth of current node in the tree of nodes visited by this query

## Access to QEC
The client might get access to QEC, by these ways:
**Access1) CtQuery#createQueryContext()**
By creation of QEC before query is executed. Then that instance can be used in functions of query to drive query
```java
//1) create instance of QEC
 CtQueryContext cc = anElement.map(...).createQueryContext();
//2) use QEC to process query and to drive query, etc.
cc.forEach(...)
```

**Access2) CtQuery#setQueryContext()**
By creating an own instance of QEC and setting it to query
```java
//1) create instance of QEC
 CtQueryContext cc = new QEC();
//assign that instance to query
anElement.map(...).setQueryContext(cc).forEach(...)
```

**Access3) (CtQueryContext)outputConsumer**
By conversion of outputConsumer of `CtConsumableFunction(Object,CtConsumer)` to QEC.
```java
anElement.map((input, output)->{
CtQueryContext cc = (CtQueryContext)output;
//use cc to drive query
}).forEach((output)->{
CtQueryContext cc = (CtQueryContext)output;
//use the same cc instance like above to drive query
});
```
**Access4) CtQuery extends CtQueryContext**
CtQuery extends CtQueryContext, so all the methods are available through query instance.

## API of QEC
There might be these methods on QEC:
* setOutputConsumer(CtConsumer) - sets the CtConsumer which will get results of this query execution
* accept(Object input) - executes the query with input.
* setTerminated(boolean) - terminates whole query
* boolean isTerminated() - returns true if query is terminated
* void onEnter(CtQueryListener<T>) - register listener which is called before element is sent to mapping function
* void onEnter(Filter, CtQueryListener<T>) - some like onEnter, but calls listener only for elements which match the filter.
* void onExit(CtQueryListener<T>) - register listener which is called after element is sent to mapping function
* void onExit(Filter, CtQueryListener<T>) - same like onExit, but but calls listener only for elements which match the filter.
* void onExitOf(Object, CtQueryListener<T>) - same like onExit, but calls listener only when element is same like provided object.
* void skipElement(CtElement); - skips element and all children. The mapping function is not called for such element and it's children
* void setSkipMode(SkipMode); - skips elements depending mode until next call of setSkipMode or end of query.

```java
interface CtQueryListener<T> {
   void onElement(T element, CtQueryContext QEC);
}
```

Example
```java
//create query listener which skips all children of method
CtQueryListener ql = new CtQueryListener() {
  boolean skipping = false;
   //return false to skip current element
   boolean onElement(CtElement e, CtQueryContext context) {
       if(skipping) return false;
       if(e instanceof CtMethod) {
          //if client wants to be called on exit of this element, then he can register listener for that
          context.onExitOf(e, (e, context)->{
             skipping = false;
          });
          return false;
       }
       return true;
   }
}
anElement.map(...)..onEnter(ql).forEach(...);
```
...to be continued... 
Feedbacks and suggestions are welcome

### DIFF
--- a/src/main/java/spoon/reflect/visitor/chain/CtQuery.java
+++ b/src/main/java/spoon/reflect/visitor/chain/CtQuery.java
@@ -166,4 +166,16 @@ public interface CtQuery extends CtQueryable {
 	 */
 	@Override
 	<I> CtQuery map(CtConsumableFunction<I> queryStep);
+
+	/**
+	 * terminates current query evaluation.
+	 * This method returns normally. It does not throw exception.
+	 * But it causes that query evaluation engine terminates
+	 * and returns all the till now collected results.
+	 */
+	void terminate();
+	/**
+	 * @return true if evaluation has to be/was terminated
+	 */
+	boolean isTerminated();
 }
--- a/src/main/java/spoon/reflect/visitor/chain/CtQueryAware.java
+++ b/src/main/java/spoon/reflect/visitor/chain/CtQueryAware.java
@@ -1,0 +1,30 @@
+/**
+ * Copyright (C) 2006-2017 INRIA and contributors
+ * Spoon - http://spoon.gforge.inria.fr/
+ *
+ * This software is governed by the CeCILL-C License under French law and
+ * abiding by the rules of distribution of free software. You can use, modify
+ * and/or redistribute the software under the terms of the CeCILL-C license as
+ * circulated by CEA, CNRS and INRIA at http://www.cecill.info.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the CeCILL-C License for more details.
+ *
+ * The fact that you are presently reading this means that you have had
+ * knowledge of the CeCILL-C license and that you accept its terms.
+ */
+package spoon.reflect.visitor.chain;
+
+/**
+ * The mapping function can implement this interface,
+ * when it needs the context of the {@link CtQuery} instance,
+ * where this mapping function is going to be evaluated
+ */
+public interface CtQueryAware {
+	/**
+	 * This method is called once when mapping function is added to the {@link CtQuery}
+	 * @param query the instance of bound {@link CtQuery}
+	 */
+	void setQuery(CtQuery query);
+}
--- a/src/main/java/spoon/reflect/visitor/chain/CtQueryAware.java
+++ b/src/main/java/spoon/reflect/visitor/chain/CtQueryAware.java
@@ -21,7 +21,8 @@ package spoon.reflect.visitor.chain;
  * that need to access the state of the top-level {@link CtQuery} instance
  * containing the function to be evaluated.
  *
- * Not meant to be implemented directly, only in conjunction with {@link CtFunction} or {@link spoon.reflect.visitor.Filter}.
+ * Not meant to be implemented directly, only in conjunction with
+ * {@link CtConsumableFunction}, {@link CtFunction} or {@link spoon.reflect.visitor.Filter}.
  */
 public interface CtQueryAware {
 	/**
--- a/src/main/java/spoon/reflect/visitor/chain/CtQueryImpl.java
+++ b/src/main/java/spoon/reflect/visitor/chain/CtQueryImpl.java
@@ -24,8 +24,8 @@ import java.util.List;
 import spoon.Launcher;
 import spoon.SpoonException;
 import spoon.reflect.declaration.CtElement;
-import spoon.reflect.visitor.EarlyTerminatingScanner;
 import spoon.reflect.visitor.Filter;
+import spoon.reflect.visitor.filter.CtScannerFunction;
 
 /**
  * The facade of {@link CtQuery} which represents a query bound to the {@link CtElement},
@@ -38,6 +38,12 @@ public class CtQueryImpl implements CtQuery {
 	 * All the constant inputs of this query.
 	 */
 	private List<Object> inputs;
+
+	private OutputFunctionWrapper outputStep = new OutputFunctionWrapper();
+	private AbstractStep lastStep = outputStep;
+	private AbstractStep firstStep = lastStep;
+
+	private boolean terminated = false;
 
 	public CtQueryImpl(Object... input) {
 		setInput(input);
@@ -76,65 +82,11 @@ public class CtQueryImpl implements CtQuery {
 		return this;
 	}
 
-	/**
-	 * The evaluation context of the CtQuery. Can be used to bind the query the the output {@link CtConsumer}
-	 * using {@link CtQueryContext#outputConsumer(CtConsumer)} and then
-	 * <ul>
-	 * <li>to evaluate the query on provided input using {@link CtQueryContext#accept(Object)}
-	 * <li>to terminate the query evaluation at any phase of query execution using {@link CtQueryContext#terminate()}
-	 * <li>to check if query is terminated at any phase of query execution using {@link CtQueryContext#isTerminated()}
-	 * and to stop an expensive query evaluating process
-	 * </ul>
-	 */
-	private interface CtQueryContext extends CtConsumer<Object> {
-		/**
-		 * @return the {@link CtConsumer} used to deliver results of the query evaluation
-		 */
-		CtConsumer<?> getOutputConsumer();
-		/**
-		 * @param outputConsumer the {@link CtConsumer} used to deliver results of the query evaluation
-		 * @return this to support fluent API
-		 */
-		CtQueryContext outputConsumer(CtConsumer<?> outputConsumer);
-
-		/**
-		 * terminates current query evaluation.
-		 * This method returns normally. It does not throw exception.
-		 * But it causes that query evaluation engine terminates
-		 * and returns all the till now collected results.
-		 */
-		void terminate();
-		/**
-		 * @return true if evaluation has to be/was terminated
-		 */
-		boolean isTerminated();
-	}
-
-	/**
-	 * Creates CtQueryContext, which can be used to evaluate or terminate the query.
-	 * Usage:<br>
-	 * <pre>
-	 * {@code
-	 * CtQueryContext cc = factory.createQuery().map(...).createQueryContext();
-	 * cc.setOutputConsumer(e->{
-	 *   //... process returned elements
-	 *   //... or optionally terminate the query by
-	 *   cc.terminate();
-	 * });
-	 * //evaluate the query with `input`. The results will be delivered to `resultConsumer`
-	 * cc.accept(input);
-	 * }
-	 * </pre>
-	 * @return new instance of CtQueryContext of this query
-	 */
-	private <R> CtQueryContext createQueryContext() {
-		return new CurrentStep();
-	}
-
+	@Override
 	public <R> void forEach(CtConsumer<R> consumer) {
-		CtQueryContext cc = createQueryContext().outputConsumer(consumer);
+		outputStep.setNext(consumer);
 		for (Object input : inputs) {
-			cc.accept(input);
+			firstStep.accept(input);
 		}
 	}
 
@@ -166,44 +118,46 @@ public class CtQueryImpl implements CtQuery {
 	@SuppressWarnings("unchecked")
 	@Override
 	public <R> R first(final Class<R> itemClass) {
-		final CtQueryContext cc = createQueryContext();
 		final Object[] result = new Object[1];
-		cc.outputConsumer(new CtConsumer<R>() {
+		outputStep.setNext(new CtConsumer<R>() {
 			@Override
 			public void accept(R out) {
 				if (out != null && itemClass.isAssignableFrom(out.getClass())) {
 					result[0] = out;
-					cc.terminate();
+					terminate();
 				}
 			}
 		});
 		for (Object input : inputs) {
-			cc.accept(input);
+			firstStep.accept(input);
+			if (isTerminated()) {
+				break;
+			}
 		}
 		return (R) result[0];
 	}
-
-	private List<AbstractStep> steps = new ArrayList<>();
 
 	private boolean logging = false;
 	private QueryFailurePolicy failurePolicy = QueryFailurePolicy.FAIL;
 
 	@Override
 	public <I> CtQueryImpl map(CtConsumableFunction<I> code) {
-		steps.add(new LazyFunctionWrapper(code));
+		addStep(new LazyFunctionWrapper(code));
 		return this;
 	}
 
 	@Override
 	public <I, R> CtQueryImpl map(CtFunction<I, R> function) {
-		steps.add(new FunctionWrapper(function));
+		addStep(new FunctionWrapper(function));
 		return this;
 	}
 
 	@Override
 	public <R extends CtElement> CtQueryImpl filterChildren(Filter<R> filter) {
-		map(new ChildrenFilteringFunction(filter));
-		stepFailurePolicy(QueryFailurePolicy.IGNORE);
+		map(new CtScannerFunction());
+		if (filter != null) {
+			select(filter);
+		}
 		return this;
 	}
 
@@ -219,6 +173,15 @@ public class CtQueryImpl implements CtQuery {
 		return this;
 	}
 
+	@Override
+	public boolean isTerminated() {
+		return terminated;
+	}
+	@Override
+	public void terminate() {
+		terminated = true;
+	}
+
 	/**
 	 * Evaluates this query, ignoring bound input - if any
 	 *
@@ -226,12 +189,13 @@ public class CtQueryImpl implements CtQuery {
 	 * @param outputConsumer method accept of the outputConsumer is called for each element produced by last mapping function of this query
 	 */
 	public <I, R> void evaluate(I input, CtConsumer<R> outputConsumer) {
-		createQueryContext().outputConsumer(outputConsumer).accept(input);
+		outputStep.setNext(outputConsumer);
+		firstStep.accept(input);
 	}
 
 	@Override
 	public CtQueryImpl name(String name) {
-		getLastStep().setName(name);
+		lastStep.setName(name);
 		return this;
 	}
 
@@ -242,7 +206,7 @@ public class CtQueryImpl implements CtQuery {
 	}
 
 	public CtQueryImpl stepFailurePolicy(QueryFailurePolicy policy) {
-		getLastStep().setLocalFailurePolicy(policy);
+		lastStep.setLocalFailurePolicy(policy);
 		return this;
 	}
 	/**
@@ -257,18 +221,33 @@ public class CtQueryImpl implements CtQuery {
 		return this;
 	}
 
-	private AbstractStep getStep(int stepIdx) {
-		if (stepIdx >= steps.size()) {
-			return null;
+	protected void handleListenerSetQuery(Object target) {
+		if (target instanceof CtQueryAware) {
+			((CtQueryAware) target).setQuery(this);
 		}
-		return steps.get(stepIdx);
 	}
 
-	private AbstractStep getLastStep() {
-		if (steps.isEmpty()) {
-			throw new SpoonException("There is no step in the query");
+	private void addStep(AbstractStep step) {
+		step.nextStep = outputStep;
+		lastStep.nextStep = step;
+		lastStep = step;
+		if (firstStep == outputStep) {
+			firstStep = step;
 		}
-		return getStep(steps.size() - 1);
+		step.setName(String.valueOf(getStepIndex(step) + 1));
+	}
+
+	private int getStepIndex(AbstractStep step) {
+		int idx = 0;
+		AbstractStep s = firstStep;
+		while (s != outputStep) {
+			if (s == step) {
+				return idx;
+			}
+			s = (AbstractStep) s.nextStep;
+			idx++;
+		}
+		return -1;
 	}
 
 	private boolean isLogging() {
@@ -281,7 +260,7 @@ public class CtQueryImpl implements CtQuery {
 	 * @param e
 	 * @param parameters
 	 */
-	private void onClassCastException(CurrentStep step, ClassCastException e, Object... parameters) {
+	private void onClassCastException(AbstractStep step, ClassCastException e, Object... parameters) {
 		if (step.isFailOnCCE()) {
 			throw new SpoonException(getStepDescription(step, e.getMessage(), parameters), e);
 		} else if (Launcher.LOGGER.isTraceEnabled()) {
@@ -291,13 +270,13 @@ public class CtQueryImpl implements CtQuery {
 		log(step, e.getMessage(), parameters);
 	}
 
-	private void log(CurrentStep step, String message, Object... parameters) {
+	private void log(AbstractStep step, String message, Object... parameters) {
 		if (isLogging() && Launcher.LOGGER.isInfoEnabled()) {
 			Launcher.LOGGER.info(getStepDescription(step, message, parameters));
 		}
 	}
 
-	private String getStepDescription(CurrentStep step, String message, Object... parameters) {
+	private String getStepDescription(AbstractStep step, String message, Object... parameters) {
 		StringBuilder sb = new StringBuilder("Step ");
 		sb.append(step.getName()).append(") ");
 		sb.append(message);
@@ -313,112 +292,12 @@ public class CtQueryImpl implements CtQuery {
 	}
 
 	/**
-	 * The thread local implementation of CtConsumer,
-	 * which knows index of actually processed step
-	 * and handles response of current step and sends it to next step.
-	 *
-	 * This class plays a role of an orchestrator to move the step cursor forward,
-	 * get the step, apply it and finally to call the output consumer.
-	 */
-	private class CurrentStep implements CtQueryContext {
-		private CtConsumer<?> outputConsumer;
-		private int stepIdx = 0;
-		private boolean terminated = false;
-
-		CurrentStep() {
-		}
-
-		@Override
-		public CtConsumer<?> getOutputConsumer() {
-			return outputConsumer;
-		}
-
-		@Override
-		public CtQueryContext outputConsumer(CtConsumer<?> outputConsumer) {
-			this.outputConsumer = outputConsumer;
-			return this;
-		}
-
-		@SuppressWarnings("unchecked")
-		@Override
-		public void accept(Object input) {
-			if (input == null || isTerminated()) {
-				return;
-			}
-			stepIdx++;
-			try {
-				if (stepIdx <= steps.size()) {
-					//process next intermediate step
-					AbstractStep step = getStep();
-					log(this, "received", input);
-					try {
-						step.apply(input, this);
-					} catch (ClassCastException e) {
-						onClassCastException(this, e, input);
-					}
-				} else {
-					//send element to outputConsumer, it means return one value of the query
-					log(this, "returning", input);
-					if (outputConsumer != null) {
-						try {
-							((CtConsumer<Object>) outputConsumer).accept(input);
-						} catch (ClassCastException e) {
-							onClassCastException(this, e, input);
-						}
-					}
-				}
-			} finally {
-				stepIdx--;
-			}
-		}
-
-		private String getName() {
-			AbstractStep stepFunction = getStep();
-			if (stepFunction == null) {
-				return "outputConsumer";
-			}
-			String name = stepFunction.getName();
-			if (name == null) {
-				name = String.valueOf(stepIdx);
-			}
-			return name;
-		}
-
-		private boolean isFailOnCCE() {
-			AbstractStep step = getStep();
-			if (step == null) {
-				//it is final consumer. Never throw CCE on final forEach consumer
-				return false;
-			}
-			return step.isFailOnCCE();
-		}
-
-		private AbstractStep getStep() {
-			return CtQueryImpl.this.getStep(stepIdx - 1);
-		}
-
-		@Override
-		public String toString() {
-			return "Step " + getName();
-		}
-
-		@Override
-		public void terminate() {
-			terminated = true;
-		}
-
-		@Override
-		public boolean isTerminated() {
-			return terminated;
-		}
-	}
-
-	/**
 	 * Holds optional name and local QueryFailurePolicy of each step
 	 */
-	private abstract class AbstractStep {
+	private abstract class AbstractStep implements CtConsumer<Object> {
 		String name;
 		QueryFailurePolicy localFailurePolicy = null;
+		CtConsumer<Object> nextStep;
 
 		/**
 		 * @return name of this Step - for debugging purposes
@@ -446,8 +325,40 @@ public class CtQueryImpl implements CtQuery {
 		private void setLocalFailurePolicy(QueryFailurePolicy localFailurePolicy) {
 			this.localFailurePolicy = localFailurePolicy;
 		}
+	}
 
-		abstract void apply(Object input, CurrentStep outputConsumer);
+	/**
+	 * Wrapper around terminal {@link CtConsumer}, which accepts output of this query
+	 */
+	private class OutputFunctionWrapper extends AbstractStep {
+		@Override
+		public void accept(Object element) {
+			if (element == null || isTerminated()) {
+				return;
+			}
+			try {
+				nextStep.accept(element);
+			} catch (ClassCastException e) {
+				if (Launcher.LOGGER.isTraceEnabled()) {
+					//log expected CCE ... there might be some unexpected too!
+					Launcher.LOGGER.trace(e);
+				}
+			}
+		}
+		@SuppressWarnings({ "unchecked", "rawtypes" })
+		<R> void setNext(CtConsumer<R> out) {
+			//we are preparing new query execution.
+			reset();
+			nextStep = (CtConsumer) out;
+			handleListenerSetQuery(nextStep);
+		}
+	}
+
+	/**
+	 * Called before query is evaluated again
+	 */
+	protected void reset() {
+		terminated = false;
 	}
 
 	private class LazyFunctionWrapper extends AbstractStep {
@@ -457,14 +368,18 @@ public class CtQueryImpl implements CtQuery {
 		LazyFunctionWrapper(CtConsumableFunction<?> fnc) {
 			super();
 			this.fnc = (CtConsumableFunction<Object>) fnc;
+			handleListenerSetQuery(this.fnc);
 		}
 
 		@Override
-		public void apply(Object input, CurrentStep outputConsumer) {
+		public void accept(Object input) {
+			if (input == null || isTerminated()) {
+				return;
+			}
 			try {
-				fnc.apply(input, outputConsumer);
+				fnc.apply(input, nextStep);
 			} catch (ClassCastException e) {
-				onClassCastException(outputConsumer, e, input);
+				onClassCastException(this, e, input);
 				return;
 			}
 		}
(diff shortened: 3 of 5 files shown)


pvojtechovsky:
It is still not finished, but API will be like this.
```java
public interface CtQueryListener {
	/**
	 * Called when new object enters query step, before the object is processed by mapping function of that step.
	 * <br>
	 * The returning false or throwing an exception causes that processing of input object and all it's children is skipped.
	 * The returning true causes that input object is normally processed, and exit is called for that object.
	 * @param element the processed element
	 * @param context context the query execution context
	 * @return true to continue processing this step and it's children,
	 * false to skip this step and it's children.
	 */
	boolean enter(Object element, CtQueryContext context);
	/**
	 * The exit is called always only after `enter` returns true and input element and all children are processed or they throw an exception.
	 * @param element the processed element
	 * @param context the query execution context
	 * @param exception null or an exception if the processing of the mapping function failed
	 */
	void exit(Object element, CtQueryContext context, Throwable exception);
}
```
and the listener can be assigned to the CtQuery using new methods `CtQuery#setListener(CtQueryListener)`

Please note that there is also new interface `CtQueryContext`, which is sent to both `CtQueryListener` methods. The `CtQueryContext` can be actually used to
* terminate the query
* to send the new input to the query - recursive execution of query (I do not know if it makes sense)
* to change the output of the query - (I do not know if it makes sense)

In future it might be used to get access to the:
* the depth of current step in query execution
* the instance of actually executed CtQuery
* the instance of current mapping function

The one query step and processing of all it's children can be terminated by returning false from `CtQueryListener#enter` method too, so it actually looks like that whole CtQueryContext is not needed and may stay private.

monperrus:
Hi Pavel,
That's an interesting design. However, it feels much more complex than the previous one.

The previous one had two new protected methods in an implementation class "enter" and "exit", and I liked it very much.

Here we have two new interfaces CtQuetyListener and CtQueryContext. I'm afraid that this more complex design will be more difficult to document, maintain and understand for new users.

WDYT?

pvojtechovsky:
Hi Martin, thanks for feedback. I will try to make `CtQueryContext` private again.

> the previous design.

It was just in the middle of work. It was no real solution. There were protected methods on private class and there was no way how use that extended class, even if it would be protected. But I will think about that solution too.

I will experiment with current design from client's point of view. I will try to use it in refactoring code and get some usability experience.

pvojtechovsky:
Do we need multi-threading support in CtQuery?

I mean to create one instance of CtQuery and then evaluate it parallel in multiple threads? Actually it is partially possible (depending on used mapping function). 

But that support, brings some extra complexity, which influences design of this PR too.

So do you agree to make CtQuery and it's future API design single threading?

pvojtechovsky:
> So do you agree to make CtQuery and it's future API design single threading?

I know you like KISS, so your answer is 99.9% yes.

So now the CtQueryImpl is refactored and the query evaluation context is directly part of the CtQuery. 
It means:
1. new methods:
   * `CtQuery#isTerminated()`
   * `CtQuery#setTerminated(boolean)`
2. we do not need CtQueryContext, but we can directly use `CtQuery` instead. So there is:
```java
public interface CtQueryListener {
	boolean enter(Object element, CtQuery context);
	void exit(Object element, CtQuery context, Throwable exception);
}
```
I have added CtQueryImpl$QueryConsumer here too, which is used in this PR by `ParentFunction`. It is optionally second baby step in this PR. Let's discuss it later.

monperrus:
>     So do you agree to make CtQuery and it's future API design single threading?
>
> I know you like KISS, so your answer is 99.9% yes.
>
Correct :-) I'm fundamentally KISS.

src/main/java/spoon/reflect/visitor/chain/CtQueryListener.java
```diff
@@ -0,0 +1,41 @@
...
+	 * @param context context the query execution context
+	 * @return true to continue processing this step and it's children,
+	 * false to skip this step and it's children.
+	 */
+	boolean enter(Object element, CtQuery context);
+	/**
+	 * The exit is called always only after `enter` returns true and input element and all children are processed or they throw an exception.
+	 * @param element the processed element
+	 * @param context the query execution context
+	 * @param exception null or an exception if the processing of the mapping function failed
```
monperrus: What's parameter "exception" used for?.

pvojtechovsky: to detect whether processing of the children passed correctly or failed. I do not need it now, because my code does not throw exceptions, but it is API so it should be generic enough

pvojtechovsky: actually it is used like this:
```java
listener.enter(item, query);
try {
 ... process children
} catch (Throwable e) {
  listener.exit(item, query, e);
  throw e;
}
listener.exit(item, query, null);
```
so the exit is called always, and you can detect in exit whether it failed or not by checking exception parameter.

Optionally we might implement it like this
```java
listener.enter(item, query);
try {
 ... process children
} catch (Throwable e) {
  listener.failureExit(item, query, e);
  throw e;
}
listener.exit(item, query);
```
where `failureExit` method might added to the API later, after somebody needs it.
In such case we should provide an default implementation of listener, so the clients can extend that default implementation and will be not forced to change their code later, after we add method failureExit into API.

monperrus:
> I do not need it now
Then, I propose to remove it. Unchecked exceptions can still be thrown if necessary.

pvojtechovsky:
During experimenting with `CtQueryListener`, I have found I do not want to listen for query steps. I need to listen to enter/exit of AST nodes during `CtQuery#filterChildren`. Therefore I have completely redesigned this PR. It is no more about query context, but about scanning context.

What I did?

BS 1. The `CtQuery` (represents query context too) now holds state of the query execution. The new methods `CtQuery#isTerminated()` and `CtQuery#setTerminated(boolean)`. The `CtQueryImpl` was refactored to hold query context directly.

BS 2. The `CtQuery` can be passed now to any mapping function or Filter. The mapping function or Filter, which needs that information has to implement `CtQueryAware` and `CtQueryAware#setQuery(CtQuery)` will be called once before the mapping function is executed. See `ParentFunction` as an example of usage.

BS 3. The existing private class `CtQueryImpl$ChildrenFilteringFunction` was moved to own file and named `FilterChildrenFunction`.

BS 4. The new `FilterChildrenFunction` mapping function accepts the `CtScanningListener` (it is the renamed `CtQueryListener`). This listener can influence processing/skipping children elements during AST scanning process.

(The BS means Baby Step ;-) ... I know there are 4, but may be they are acceptable in this one PR...)

pvojtechovsky:
> IntercessionTest.testSettersAreAllGood:215 Your setter setTerminated in CtQuery don't have a generic type for its return type.

@monperrus Is it problem of my implementation or problem of IntercessionTest?

monperrus:
Thanks for the proposal. The BS you propose seem reasonable, even if 4 BS may hinder a quick understanding and converging. Now, I feel a bit lost now. Before discussing further, I'd like to understand the relation between the BS and the requirements of this PR (at the top of this thread):

> 1. to know when query `enter`/`exit` an element during filterChildren on any query step level 
> 1. to be able to skip processing of children of current element 
> 1. to be able to skip processing of siblings and children of current elements

Do you still address them all? Do you address new ones?

pvojtechovsky:
I still address them. With this little change
> old: to know when query enter/exit an element during filterChildren on any query step level
> new: to know when query enter/exit an element during chosen filterChildren mapping

They are covered by BS4, which needs BS3, which needs BS2 and BS1

I address new:
4. to be able to make mapping functions aware of query processing state (actually mainly whether query is terminated) - covered by BS1, BS2

monperrus:
OK, let's discuss on BS1 and BS2 since since the others depend on it.

BS1: OK.

BS2: I'm not convinced because: 1) the responsibility of CtQueryAware is trivial, it's not worth introducing a new interface for this. 2) nobody expects a CtQueryAware (I didn't find it) 3) it introduces some state in function, which is essentially a stateless concept (and stateless is better). Am I missing something?

pvojtechovsky:
> 1) the responsibility of CtQueryAware is trivial, 

yes, it is trivial. I learned that concept by Spring framework. It is an easy way, how to declare, that some object needs a knowledge about something outside.

> it's not worth introducing a new interface for this. 

I am different opinion :-). But do not care. The question is: Do you have an alternative solution for: How to deliver CtQuery into instance of `Filter`, `CtFunction` or `CtConsumableFunction`?

> 2) nobody expects a CtQueryAware (I didn't find it) 

I do not understand this note. Do you mean nobody needs that? It is used by `ParentFunction` (just little optimization) and it is needed by `FilterChildrenFunction` which was introduced in this PR. And I have feeling like many more complex mapping functions will need to be able to early terminate processing, so they actually need to know CtQuery.

> 3) it introduces some state in function, which is essentially a stateless concept (and stateless is better). 

I also wanted stateless at the beginning. But that costs something ... higher complexity ... less KISS.
And note: the CtScanner, which is used in core of most of mapping functions, is already NOT stateless.

By this question:
> So do you agree to make CtQuery and it's future API design single threading?

I asked for the agreement to leave stateless concept. 
I like stateless too, and if you want to invest effort and pay in form of more complex client's API, and more complex implementation of mapping functions (new CtScanner, etc.), then let's discuss **What advantage stateless brings?** Do we really need that?

pvojtechovsky:
> The question is: Do you have an alternative solution for: How to deliver CtQuery into instance of Filter, CtFunction or CtConsumableFunction?

I was wrong. The context is needed in CConsumableFunction only. Other two mapping functions are not called when query is terminated. And CtConsumableFunction gets second parameter CtConsumer, which might be change or cast to something what can return the query.

But it is true that current implementation of mapping functions is usually no stateless.

pvojtechovsky:
another change of my opinion :-)

> The question is: Do you have an alternative solution for: How to deliver CtQuery into instance of Filter, CtFunction or CtConsumableFunction?

This is correct question, because CtFunction and Filter sometime might need context (CtQuery) to be able to terminate the query. I already found this use case in implementation of PR #1005

pvojtechovsky:
This PR is finished from my side. 

Or do you have a better - stateless - way how to deliver `CtQuery` to `Filter`, `CtFunction` or `CtConsumableFunction`? 
Or do you agree that stateless is too expensive in this case and Spoon mapping functions will be stateful? 
Most of mapping functions which I made till now are stateful anyway.

surli:
>> IntercessionTest.testSettersAreAllGood:215 Your setter setTerminated in CtQuery don't have a generic type for its return type.

> @monperrus Is it problem of my implementation or problem of IntercessionTest?

The contract of this intercessionTest is that all setters contained in Ct* classes must return a generic type which extends a Ct* class. I think you can easily change your implementation of setTerminated to also return the CtQuery.

pvojtechovsky:
I have use this signature
```java
CtQuery setTerminated(boolean terminated)
```
But test still fails. Now with
> java.lang.AssertionError: Your setter setTerminated in CtQuery don't have a generic type for its return type.

If you think that IntercessionTest is OK, then I can use this signature
```java
void terminated(boolean terminated)
```
Test passes then. What do you think?

And then the test complains on `CtQueryAware#setQuery(CtQuery)`
Solution
A) rename CtQueryAware to QueryAware
B) `CtQueryAware#setQuery(CtQuery)` to `CtQueryAware#query(CtQuery)`
C) ?

What is the meaning of `Ct` prefix? May be the IntercessionTest should be applied only to classes from selected Spoon packages?

tdurieux:
Personally I  think that the scope of the test `testSettersAreAllGood` is too large (the complete source code of Spoon). It will probably better to focus only on the model and not on the visitors or queries...

surli:
> It will probably better to focus only on the model and not on the visitors or queries...

Agreed with @tdurieux. I thought it was only a simple change on the setTerminated, but if it's more complicated we should improve this test to ignore some packages.

tdurieux:
@pvojtechovsky your PR is passing now

pvojtechovsky:
@tdurieux thanks! :-)

pvojtechovsky:
This PR is finished from my side. Please review it.
I use CtQueryAware in most of mapping functions I am working on. It would be easier to continue with them if this PR would be merged. If you think there are problems, then let's discuss them. I am open to change it. It should not be too much work to do it different. 
Later we do change it, more code in my workspace will depend on it and more effort will this change cost me. Thank You. I can wait week more no problem ...

pvojtechovsky:
I can imagine one stateless solution, which I would like. 
It changes `CtConsumableFunction` this: way.
```java
public interface CtConsumableFunction<T> {

//	void apply(T input, CtConsumer<Object> outputConsumer);
//would be replaced by
	void apply(T input, CtQueryStep queryStep);
}
```
Where new `CtQueryStep` interface would be defined like this:
```java
interface CtQueryStep extends CtConsumer<Object>{
    CtQuery getQuery();
}
```
Pros:
* it is stateless
* it does not need any extra code, because outputConsumer is already implemented as query step internally.
*  implementation of mapping functions is simpler (it does not have to implement CtQueryAware (which will not exist) and store `query` instance as a field.

Cons:
* it does not cover `CtFunction` and `Filter`. But may be it is not a real problem, because if an developer wants to implement a mapping function, which needs to know CtQuery context, then s/he can always use `CtConsumableFunction` instead of `CtFunction` or `Filter`.
* it is not backward compatible.

monperrus:
Hi Pavel, I'm back, we'll resume the discussion. --Martin

pvojtechovsky:
Hi Martin, Nice to see you ;-)

monperrus:
Let's discuss BS2.

I mean that no single variable, parameter or field is typed by CtQueryAware, and there is no instanceof of it. 
Why do we need it? 
What about a simpler solution where the constructor FilterChildrenFunction simply takes a query as parameter?

pvojtechovsky:
> solution where the constructor FilterChildrenFunction simply takes a query as parameter?

```java
CtElement ele = ...
ele.map(new FilterChildrenFunction(...queryDoesn't exist yet or at all...)...
```

Please write the java code, which would fit to that solution. 
Even if you would found some trick, it would make it much complicated for Spoon clients to use such filters. The CtQueryAware interface is invisible for spoon clients. They can simply create mapping functions as before and the query context is automatically injected into mapping function at query execution time.

pvojtechovsky:
> I mean that no single variable, parameter or field is typed by CtQueryAware, and there is no instanceof of it. Why do we need it?

It is used by this `CtQueryImpl` method, which is then called from several places
```java
	public static void handleListenerSetQuery(Object target, CtQuery query) {
		if (query != null && target instanceof CtQueryAware) {
			((CtQueryAware) target).setQuery(query);
		}
	}
```

monperrus:
> It is used by this CtQueryImpl method

OK, I was caught again by the stupid partial diff of Github (you have to click one by one on "load diff", grrrr)

monperrus:
In `FilterChildrenFunction`, the goal of knowing the query in only to test for `query.isTerminated()` or is there another responsibility?

pvojtechovsky:
> is there another responsibility?
actually I am using in my implementations of 'CtConsumableFunction'
* query.isTerminated()
* query.setTerminated(boolean)
* and may be In one case I needed to get state of one mapping function X to mapping function Y. So it might be done by future function `query.getQueryStepMappingFunction(String name)`, but in my case I was able to solve that by giving the reference to X directly to code of Y using final local variable (reference was passed during construction of Y). And it is not so bad solution, so I actually do not need `getQueryStepMappingFunction`

monperrus:
>>     is there another responsibility?
>     actually I am using in my implementations of 'CtConsumableFunction'
>
do you mean in your own classes that are not in this PR?

pvojtechovsky:
yes, I mean class which are not part of this PR, but which are part of other PRs.

monperrus:
I find the responsibility of CtQueryAware weak. Also the contract of when `handleListenerSetQuery` is called is unclear, sometimes it's in a subclass of AbstractStep sometimes in a `Function` directly (as in `FilterChildrenFunction`).

I'm not sure about the solution. Handling `handleListenerSetQuery` directly in AbstractStep, and having it private would give good encapsulation. What do you think?

pvojtechovsky:
>  Handling handleListenerSetQuery directly in AbstractStep, and having it private would give good encapsulation. 

This method is used from different packages, therefore I had to make it public, but I agree that it is an internal API, so it will be better to make it private and copy the code of this method to all (2) places, where it is needed. I will do it today.

pvojtechovsky:
handleListenerSetQuery is private now. What next? ;-)

monperrus:
> I agree that it is an internal API

Super.

> What next? 

Several things still in BS2 :-) Let's do them one by one, it's slow but easier.
  
Now, we can remove handleListenerSetQuery from FilterChildrenFunction (definition and call), because it's the responsibility of the query framework to call it. Correct?

pvojtechovsky:
> remove handleListenerSetQuery from FilterChildrenFunction

No we cannot because query framework has no idea that FilterChildrenFunction has an listener

monperrus:
do you mean the call of `handleListenerSetQuery` in `setListener`? who calls `setListener`?

pvojtechovsky:
yes

pvojtechovsky:
and this call `handleListenerSetQuery(listener, this.query);` in setQuery too.

monperrus:
Q1: who calls setListener?

Q2: in setQuery: it seems impossible that the CtScannerListener is a CtQueryAware so why having `handleListenerSetQuery(listener, this.query);`?

(I'm trying to really understand your idea and the responsibility of CtQueryAware)

pvojtechovsky:
> Q1: who calls setListener?

The [client](https://github.com/INRIA/spoon/pull/1193/files#diff-3eb793be3943b0dcf74b4255d8428a9dR98)

> Q2: in setQuery: it seems impossible that the CtScannerListener is a CtQueryAware so why having handleListenerSetQuery(listener, this.query);?

Each mapping function, Filter or CtScannerListener can optionally implement also CtQueryAware. The developer of such code can do that when he needs to know state (isTerminated/setTerminted) of query .

pvojtechovsky:
may be you might checkout this [branch](https://github.com/pvojtechovsky/spoon/tree/AllTypeMembersFunctionCompilable) , which contains compilable version of all my PRs. If you would have this branch in your development environment, then it would be easier for you to see how CtQueryAware and listeners etc. are used.

pvojtechovsky:
The contract of `CtQueryAware` is easy:
**If any mapping function** (CtFunction, CtConsumableFunction, Filter) or CtScannerListener or any future mapping function listener **needs to know instance of CtQuery, then it simply implements CtQueryAware** and the spoon code assures that CtQueryAware#setQuery is called before the mapping function or listener is used in query.

monperrus:
it seems we're replaying the movie of #1070: there are too many intricated requirements and responsibilities.. 

One solution is to come back to real baby steps, in separate PRs.

Consequently, I've named the requirements at the top of this tread. I would like that to come back to the initial idea of this thread and to converge to a simple solution for requirement R1 only, first.

I would say it only requires to attach something similar to CtScannerListener to a step, ie adding one single interface and one single method  `CtQuery#attach(CtScannerListener)`, plus a test with a fake CtScannerListener.

WDYT?

Thanks for your patience :-)

pvojtechovsky:
I understand. I will split it into smaller PRs

monperrus:
Thanks.

pvojtechovsky:
> I would say it only requires to attach something similar to CtScannerListener to a step, ie adding one single interface and one single method CtQuery#attach(CtScannerListener), plus a test with a fake CtScannerListener.

This solution will not work, because CtScannerListener is not related to CtQuery, but is related to "query step" of type scanner - the internal code of FilterChildrenFunction. Therefore it makes no sense to add listener to query and to add an API there. OK, I might delegate CtQuery#attach to FilterChildrenFunction#setListener, but 
1. it just makes thinks less visible
2. I already need another listeners for another mapping functions (not scanner based). Will we add again query delegate method for that? I am sure it is not good approach.

So now I see that it is not about splitting into smaller PRs, but about misunderstanding on main contract of CtScannerListener.

This misunderstanding is probably caused by this old concept:
> R1 to know when **query** enter/exit an element during filterChildren on **any query step level**

which is no more correct! Actually it sounds like this:
> R1 to know when **query step** of type filterChildren enter/exit an element on **one selected query step**.

I have found that it makes no sense to listen for events in whole Query. I always need it only for one selected query step.

pvojtechovsky:
I can imagine following small steps:

S1) add new CtQuery methods setTerminated/isTerminated. 

So clients can check or influence the query state. 
Note: that this needs refactoring of CtQueryImpl, because actually instance of CtQueryImpl has no access to query state, which is independent.

S2) implement a way how can mapping function (implementation of `CtConsumableFunction`) get access to query state (CtQuery#isTerminated/CtQuery#setTerminated()). -> CtQueryAware

It is needed to be able to stop producing next results and terminate instead.

S3) move private CtQueryImpl$FilterChildrenFunction to public FilterChildrenFunction. 

It is needed to get access to FilterChildrenFunction instance, to be able to configure it later. 

S4) modify FilterChildrenFunction by way it accepts and calls CtScannerListener on enter/exit of each AST node.

Note that S2 needs S1, S3 needs S2 +S1, S4 needs S3.
Note that S1 makes no sense without S2. So PR which contains only S1, cannot be tested.
Note that S3 makes no sense without S4. So PR which contains only S3 is useless.

If you agree with this order and concepts/features, then I can start to create smaller PRs.

pvojtechovsky:
> One thing I don't understand is the presence of FilterChildrenFunction in the PR. I would say it would be easier to understand if we put the changes in the query engine (CtQueryImpl) in one first PR

I cannot register CtScannerListener in FilterChildrenFunction if it is private in CtQueryImpl. I can of course add new method CtQuery#attachListener, which would then call private FilterChildrenFunction.setListener, but I think it is not good approach, because
* CtQuery does not have to know about listeners of FilterChildrenFunction
* Client might try to call `attachListener` after other mapping functions too - should it fail in that case?
* CtQuery#attachListener is confusing, because it looks like listener is attached to query and not to scanner.
* I already have another mapping function, which is not based on scanner and which needs listener too. Will we register that listener using CtQuery too? Or will we do it different in each case?

So I think easiest and most consistent approach is to make FilterChildrenFunction public.
If client just needs to filter children then, s/he can use old CtQuery.filterChildren
If client needs more then, s/he can use CtQuery.map(new FilterChildrenFunction()...configure this instance or even create a new class out of it...)

monperrus:
Hi Pavel,

Your plan makes sense, and it clarifies certain things, thanks for it. I agree with your good
testability reflects good design, and consequently a PR should be an atomic and testable feature.

Instead of low-level coding actions (S1, S2), I'd propose to reason in terms of features, client
code and and dependencies between features.

According to all your comments, "FilterChildrenFunction" is a core concept for what you do. In
CtQueryImpl we already have "ChildrenFilteringFunction".

From a naming perspective, it is confusing, and that's maybe one cause of this discussion.

What's the responsibilities of "FilterChildrenFunction" and "ChildrenFilteringFunction"? What's the
differences between them?

pvojtechovsky:
> What's the responsibilities of "FilterChildrenFunction" and "ChildrenFilteringFunction"? What's the
differences between them?

there is no difference. I moved all responsibility of ChildrenFilteringFunction to FilterChildrenFunction.

See changed implementation of CtQueryImpl#filterChildren of this PR:
```java
public <R extends CtElement> CtQueryImpl filterChildren(Filter<R> filter) {
	map(new FilterChildrenFunction(filter));
	return this;
}
```

So at the beginning they are exactly same. Later in this PR I added possibility to register CtScannerListener. 
So the difference is that
* FilterChildrenFunction is public
* FilterChildrenFunction can register CtScannerListener and calls it's enter/exit methods

monperrus:
> there is no difference. I moved all responsibility of ChildrenFilteringFunction to
> FilterChildrenFunction.
>
OK, to some extent, you made ChildrenFilteringFunction public. That's a first candidate well-scoped
refactoring PR.

> class ChildrenFilteringFunction extends EarlyTerminatingScanner<Void> implements
> CtConsumableFunction<CtElement>
>
Because of the naming again, I'm not sure of the use of inheritance versus composition.

Is ChildrenFilteringFunction more a function or a scanner to your opinion?

pvojtechovsky:
> OK, to some extent, you made ChildrenFilteringFunction public. 

yes, exactly.

> That's a first candidate well-scoped refactoring PR.

yes, but as I described in S1-S4. It is the S3, which needs S1 and S2 and makes no sense with S4 :-)

> Is ChildrenFilteringFunction more a function or a scanner to your opinion?
both:
- it is mapping function, which is using scanner
- it is scanner which can be used as a mapping function.
But if anybody will use it, then s/he will use it as mapping function (1), which is doing scanning.

monperrus:
I'd say it's more a function.

For sake of understandability, what do you think of replacing inheritance with composition, such
that ChildrenFilteringFunction is composed of an

EarlyTerminatingScanner (or composed of private subclass of EarlyTerminatingScanner)?

pvojtechovsky:
ok, no problem. Just one critical question: 
Do you speak about no more existing ChildrenFilteringFunction or about new FilterChildrenFunction ?

pvojtechovsky:
> replacing inheritance with composition
note: inheritance LITTLE bit faster, because less objects has to be created. And the FilterChildrenFunction is used EVERYWHERE.

monperrus:
> Do you speak about no more existing ChildrenFilteringFunction or about new FilterChildrenFunction ?
the new FilterChildrenFunction, it's simpler for your rebase.

pvojtechovsky:
> the new FilterChildrenFunction

Correct answer. I wanted to hear that 👍 
I will do it. But on Saturday. I have no time before.

monperrus:
excellent :-)

pvojtechovsky:
FilterChildrenFunction now does not extends EarlyTerminatingScanner, but in contains it as private class.

Anything else has to be changed?

pvojtechovsky:
What about name and contract of `CtScannerListener`?
* What is the meaning of Ct prefix?
* `CtScannerListener` is actually used as listener for `FilterChildrenFunction`, which internally uses CtScanner. But may be it makes sense to listen to enter/exit events on some other mapping functions too. For example `SuperHierarchyFunction` [#1195](https://github.com/INRIA/spoon/pull/1195/files#diff-cd0d32ba8d0dedeeb332293a9d947bb3) is not based on `CtScanner`, but it visits hierarchy of classes and interfaces and it makes sense to listen on enter/exit events, to provide access to context of actually visited `CtType`.

monperrus:
Hi Pavel,

This looks clearer, thanks. 

To focus the discussion, I continue with `FilterChildrenFunction` (and I've tentatively renamed the PR accordingly):

First question, does it make sense that `CtQueryAware` becomes `CtQueryAwareFunction extends CtFunction`? (or `extends CtConsumableFunction`)?

Second question, the private Scanner does three things:
   1. uses a filter
   1. extends termination handling using a query
   1. declares a listener

It seems a lot. What is the key responsibility of `FilterChildrenFunction`: declaring a listener? or handling query termination?

pvojtechovsky:
> does it make sense that CtQueryAware becomes CtQueryAwareFunction extends CtFunction? (or extends CtConsumableFunction)?

no, because all implementations of CtFunction, CtFunction, Filter and CtScannerListener may OPTIONALLY implement CtQueryAware  to get access to the query, which processes them.

> What is the key responsibility of FilterChildrenFunction: declaring a listener? or handling query termination?

> or handling query termination?

It looks like a misunderstanding. The query termination (the optimal implementaion) must be handled on more places. 
1) if CtQueryImpl detect, that it is terminated, then it will not call other mapping functions and finishes ASAP.
2) if any mapping function (e.g. FilterChildrenFunction, but there will be others too) detects that query is terminated then it should not produce another results, because they will be ignore by query anyway.

So FilterChildrenFunction must do all what it does. I see no way/reason why to remove some responsibility..

monperrus:
> The query termination (the optimal implementaion) must be handled on more places. 
> 1) if CtQueryImpl detect, that it is terminated, then it will not call other mapping functions and finishes ASAP.
> 2) if any mapping function (e.g. FilterChildrenFunction, but there will be others too) detects that query is terminated then it should not produce another results, because they will be ignore by query anyway.

This is very clear and interesting, thanks. For sake of simplification (documentation and maintenance), would the first point only be sufficient?

pvojtechovsky:
It can bring performance problems (scanner will produce million of nodes, even if query is already terminated) and if we remove that feature (as you suggest), then it safes only few lines of code. I do not thing that this "simplification" is good idea.

pvojtechovsky:
That simplification would remove already merged feature #1138

monperrus:
> It can bring performance problems

OK, that's fine to me.

I start to understand the design intent.

I have two concerns:

1. Concern 1: CtScanner and CtScannerListener are obviously conceptually related but this is not visible in the current design. It would make sense to have a  CtScannerListener outside the realm of queries. What about a `CtListenableScanner extends CtScanner` that is independent of the query architecture? (and then FilterChildrenFunction would be based on it.)
2. Concern 2: "all implementations of CtFunction, Filter and CtScannerListener may OPTIONALLY implement CtQueryAware" while CtFunction and CtFilter are conceptually related (they ground a query), CtScannerListener is of different nature. So it's confusing to put them all under the same umbrella. This is technically possible thanks to typing by `Object` and of the genericity of a marker interface, but this is not really satisfying. In what cases do you need right that a CtScannerListener is CtQueryAware?

pvojtechovsky:
>> It can bring performance problems

> OK, that's fine to me.

What is fine? A) performance problems B) that EarlyTerminatingScanner is used :-)
May be looks like stupid question, but in concept of CtListenableScanner you probably expect that EarlyTerminatingScanner is not used?

> CtListenableScanner 

If somebody will still use CtScanner without query in future, then CtListenableScanner might be useful. But it is problem in case of FilterChildrenFunction, because the scanner which is used there is
1) early terminating
2) listenable
and java does not support multiple inheritance.

> concern 2

CtScannerListener and CtQueryAware are not related. But the implementation of CtScannerListener CAN OPTIONALLY implement CtQueryAware, if developer needs to get access to query stat in the scanner implementation. I did not needed that till now, but I was near to that solution several times. So I can imagine situation where it really makes sense.
What is cost of this till now not needed feature? 1 line of code
`handleListenerSetQuery(scanner.listener, scanner.query);`
I guess it is acceptable....

monperrus:
> What is fine?

To have the optimization regarding termination spread in two different places.

> If somebody will still use CtScanner without query in future, then CtListenableScanner might be useful.

We may say that it's the responsibility of `EarlyTerminatingScanner` to do this. It means we would move the code related to the listener from `FilterChildrenFunction.Scanner` to `EarlyTerminatingScanner`. In this case, we would be able to use `CtScannerListener` outside the query context. What do you think?

pvojtechovsky:
> move the code related to the listener from FilterChildrenFunction.Scanner to EarlyTerminatingScanner

I agree with that. 

What do you say about reuse of `CtScannerListener` in other contexts? Actually it is used by CtScanner only, but I plan to use it for class hierarchy scanning too. Then the name  `CtScannerListener` is not good. Other things are OK.

monperrus:
>> move the code related to the listener from FilterChildrenFunction.Scanner to EarlyTerminatingScanner

>I agree with that. 

Super

> What do you say about reuse of `CtScannerListener` in other contexts?

If we don't need it now, let's leave this aside.

>  What is cost of this till now not needed feature? 1 line of code

I disagree. It's also having static typing with `Object` which is hard to document, maintain, and error-prone. That's why I would prefer to focus the `CtQueryAware` capability on Function.

pvojtechovsky:
> It's also having static typing with Object

I do not understand. Could you explain it?

monperrus:
```
handleListenerSetQuery(Object target, CtQuery query)
```

Explicit typing would be
```
/** register a query in the given target which must be an instance of CtQueryAware */ 
handleListenerSetQuery(CtFunction target, CtQuery query)
```

pvojtechovsky:
I understand. Thanks!

But `handleListenerSetQuery(Object target, CtQuery query)` will need Object in future too, because CtFunction, CtConsumableFunction and Filter does not have common parent anyway. So removing CtScannerLister from this list does not help. Anyway I do not persist on the CtScannerListener beeing CtQueryAware, so we can remove that for now. No problem

pvojtechovsky:
I did two changes
1) The listener of FilterChildrenFunction.Scanner was moved to EarlyTerminatingScanner. See separate PR #1210.
2) The CtScannerLister cannot be CtQueryAware now (I removed calls of `handleListenerSetQuery(listener, query)`)

What next?

monperrus:
> CtFunction, CtConsumableFunction and Filter does not have common parent anyway

unless we go for:
```
interface CtQueryElement{} // marker interface
CtFunction extends CtQueryElement
Filter extends CtQueryElement
handleListenerSetQuery(CtQueryElement target, CtQuery query)
```
what do you think?

> What next?

First, a couple of minor design questions, then some renaming of classes and methods, and finally documentation, and tests.

Let's start with the design questions:

- does it make sense to restart a query? can we simplify the API with by replacing `setTerminated(boolean)` by `terminate()`?
- can we remove ParentFunction, since the scope of this PR is `FilterChildrenFunction`?

pvojtechovsky:
> CtQueryElement

I understand, but I do not know if it is good idea. It brings nothing from my point of view. But if you like, I have no problem to add it. I do not care in this case. I am waiting for your decision here...

> does it make sense to restart a query? 

There is a case where I needed to temporarily terminate query which was processing CtScanner for some nodes only and then continue with scanning of further nodes, but later I solved this situation by returning false from `CtScannerListener#enter()`, which skips processing node and children. So actually I have no use case where I would need to continue with the scanning or query processing after temporary termination applied only to some nodes.

But it definitely makes sense to run query several times, even after it was terminated. In such case the `terminated` flag should be automatically reset before query is executed again.

> can we simplify the API with by replacing setTerminated(boolean) by terminate()?

I can actually live with `terminate()` method. I do not know whether setTerminated(false) will be needed in future.

> can we remove ParentFunction, since the scope of this PR is FilterChildrenFunction?

sure, I will remove it. It was just an example, how to use `CtQueryAware` on other places then in `FilterChildrenFunction`

monperrus:
>>  CtQueryElement

> I understand, but I do not know if it is good idea. It brings nothing from my point of view. 

If you don't see a major advantage, let's KISS and keep it this way.

> I can actually live with terminate() method.

Let's do it then

> But it definitely makes sense to run query several times, even after it was terminated. In such case the terminated flag should be automatically reset before query is executed again.

OK.

monperrus:
We make great progress, thanks! one more question about FilterChildrenFunction.

What I understand now is that it provides the capability of CtScanner in the context of queries, efficiently.

Now, do we really need to have a filter there? Since the query architecture is step-based, and lazily evaluated, having the filter as the next step would be functionally equivalent and similarly efficient.

Correct? Can we we remove the filtering responsibility from FilterChildrenFunction?

pvojtechovsky:
> Can we we remove the filtering responsibility from FilterChildrenFunction?

probably yes. It is an interesting idea. I see only one (little?) problem
The CtQueryImpl#filterChildren will be implemented like this:
```java
	public <R extends CtElement> CtQueryImpl filterChildren(Filter<R> filter) {
		map(new FilterChildrenFunction()); //1st step
		select(filter); //2nd step
		return this;
	}
```
it means CtQuery#filterChildren will internally add 2 query steps, while client might expect only one query step.

Then the `CtQuery#name(String)`, will give name only the `select` step. The first `FilterChildrenFunction` step cannot be named.

I vote for doing that.

And then we have to rename `FilterChildrenFunction` to ?
A) CtScannerFunction
B) ?

monperrus:
> it means CtQuery#filterChildren will internally add 2 query steps, while client might expect only one query step.

To me it's not a problem, steps are internal, they are not part of the API.

> And then we have to rename FilterChildrenFunction to CtScannerFunction?

Yes. We're along the same line.

We're done for the core design decisions. Thanks.

Now the final steps are documentation and tests. I'll do the documentation part. Can you add the relevant tests?

pvojtechovsky:
What about return value of `CtScannerListener#enter`. Actually it returns `boolean`
* true - process element and children normally
* false - skip element and all children

May be we might change boolean to enum SCANNING_MODE {NORMAL, SKIP_CHILDREN, SKIP_ALL}

What do you think?

monperrus:
> May be we might change boolean to enum SCANNING_MODE {NORMAL, SKIP_CHILDREN, SKIP_ALL}
>
Excellent idea! It's more readable and allows for future changes and extension.

pvojtechovsky:
#1180, #1211 and #1210 are updated and are using new `ScanningMode` enum. Please review

monperrus:
merged #1210, you can rebase this one, and I'll pass over the doc.

pvojtechovsky:
rebased

monperrus:
Doc PR on this one: https://github.com/pvojtechovsky/spoon/pull/14

pvojtechovsky:
Looks like it is ready for merge. I will then rebase other PRs

monperrus:
Thanks a lot for the effort put in this very well-done contribution.

pvojtechovsky:
You are welcome! 👍

PR #1192, opened 2017-02-19
feature: automatic explicit ThisAccess for fields

### LINKED ISSUE #1190: bug: Wrong printed field reference after rename of variable to the same name
This code passes well
```java
class A {
  int var1 = 1;
  void A(int var2) {
     assertEquals(1, var1);
     assertEquals(2, var2);
  }
}
new A(2);
```
but after rename of var2 to var1, it fails, because spoon prints the field reference, wrong like this:
```java
class A {
  int var1 = 1;
  void A(int var1) {
     assertEquals(1, var1); //<---this is wrong
     assertEquals(2, var1);
  }
}
new A(2);
```
but it should be
```java
class A {
  int var1 = 1;
  void A(int var2) {
     assertEquals(1, this.var1); //<---this is correct
     assertEquals(2, var1);
  }
}
new A(2);
```
This issue has to be solved before #1005, because the tests fails on it.

pvojtechovsky:
Probably fixes #1190


surli:
I'll merged that one, could you confirm me it fixes #1190?

PR #1193, opened 2017-02-21
review 1: feature: search for CtVariableReference in defined scope

pvojtechovsky:
There are two things:
* The classes `XxxScopeFunction` have constructor with `CtScannerListener` parameter.
* The classes `XxxReferenceFunction` have constructor with <T extends CtVariable> `var` parameter. Such functions will search for references of `var` in children of mapping function input `CtElement`

The classes `LocalVariableReferenceFunction` , `CatchVariableReferenceFunction`, `ParameterReferenceFunction` has similar algorithm, so they inherit from `AbstractVariableReferenceFunction`. The `FieldReferenceFunction` is implemented individually, because java field references behaves different.

These improvements are needed by  #1005.

### DIFF
--- a/src/main/java/spoon/reflect/visitor/filter/CatchVariableReferenceFunction.java
+++ b/src/main/java/spoon/reflect/visitor/filter/CatchVariableReferenceFunction.java
@@ -18,8 +18,6 @@ package spoon.reflect.visitor.filter;
 
 import spoon.reflect.code.CtCatchVariable;
 import spoon.reflect.reference.CtCatchVariableReference;
-import spoon.reflect.visitor.chain.CtConsumableFunction;
-import spoon.reflect.visitor.chain.CtConsumer;
 
 /**
  * This Query expects a {@link CtCatchVariable} as input
@@ -34,16 +32,13 @@ import spoon.reflect.visitor.chain.CtConsumer;
  * }
  * </pre>
  */
-public class CatchVariableReferenceFunction implements CtConsumableFunction<CtCatchVariable<?>> {
+public class CatchVariableReferenceFunction extends LocalVariableReferenceFunction {
 
 	public CatchVariableReferenceFunction() {
+		super(CtCatchVariable.class, CtCatchVariableReference.class);
 	}
 
-	@Override
-	public void apply(CtCatchVariable<?> localVariable, CtConsumer<Object> outputConsumer) {
-		localVariable
-			.map(new CatchVariableScopeFunction())
-			.select(new DirectReferenceFilter<CtCatchVariableReference<?>>(localVariable.getReference()))
-			.forEach(outputConsumer);
+	public CatchVariableReferenceFunction(CtCatchVariable<?> catchVariable) {
+		super(CtCatchVariable.class, CtCatchVariableReference.class, catchVariable);
 	}
 }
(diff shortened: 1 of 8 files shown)


monperrus:
This PR looks good to me, it's a big refactoring of recent code.

Could you document the expectations and contracts for the implementors in the Javadoc of abstract method `createScopeQuery`?

pvojtechovsky:
Documentation was updated. 
I have also improved the signature of createScopeQuery method and made one class private, so there is less internal dependencies.
I am done with that. Please review

monperrus:
I have a concern with the design of the abstract method `createScopeQuery`, which is the core of this PR.

While I understand well the `CtQuery createScopeQuery(CtVariable<?> scope)`, it seems to me that the `scannerListener` parameter is a problem:
- we have no way to enforce or verify that the scanner listener methods are called in a correct manner
- it implicitly assumes that the returned query will use a scanner, but a query can be implemented in another way, since it is an abstraction, not directly tied to scanners
- method `apply` passes a very specific CtScannerListner, a context, and reuses this object after having passed it to  `createScopeQuery`, so `apply` does not protect and encapsulates its internal objects.

My proposal would thus be that the context is an internal object, under the responsibility of `apply` only, and to have a simpler `CtQuery createScopeQuery(CtVariable<?> scope)`.

What do you think?

pvojtechovsky:
I probably understand the problem you see. I would agree with You in case when the `AbstractVariableReferenceFunction` would be extended by the Spoon clients. But it is not the case!
`AbstractVariableReferenceFunction` is **package protected**. All the contracts it declares are internal and propriatary to implement limited and final set of classes :`CatchVariableReferenceFunction`, `LocalVariableReferenceFunction` and `ParameterReferenceFunction`. Because of that, I think it is correct to accept these theoretical design flaws. In this case it is OK, that implementation of AbstractVariableReferenceFunction expects an specific and matching implementation of their children.

> My proposal would thus be that the context is an internal object, under the responsibility of apply only, and to have a simpler CtQuery createScopeQuery(CtVariable<?> scope).

It cannot work, because it really needs to register that listener. And the only code which can register it is the implementation of createScopeQuery.

pvojtechovsky:
Ideas: 
I1) To move all implementations of createScopeQuery to AbstractVariableReferenceFunction and to make it private there. It will be code like
``java
if(variable instanceof CtLocalVariable) {
   return ... variable query
} else if(variable instanceof CtCatchVariable)
   return ... catch query
...
I2) optinaly to rename AbstractVariableReferenceFunction to some name which represents LocalVariable, CatchVariable, Parameter, but does not mean Field. And make it public, so it can be used directly... 

I3) optinally Then we might remove all 3 child classes, because they are nearly empty now. Or we can keep them empty...

monperrus:
That's a good idea, thanks Pavel for the constructive proposition.

I prefer strong encapsulation over questionable design (even if it's internal).

To avoid the not-so-nice sequence of instanceof, we can use a custom visitor, and then simply write
`variable.accept(new VariableReferenceVisitor())`

pvojtechovsky:
What about this implementation?
I moved the code from abstract class to the `LocalVariableReferenceFunction`, which contains package protected constructors, which are now used by `CatchVariableReferenceFunction` and `ParameterReferenceFunction`. It is not correct from conceptual point of view, because `CatchVariableReferenceFunction` is not special case of `LocalVariableReferenceFunction`, but it is the way how to make all code private, without creating an extra abstract class...

monperrus:
I like this better, thanks. To me it seems ready for merge.

> It is not correct from conceptual point of view, because CatchVariableReferenceFunction is not special case of LocalVariableReferenceFunction

Why? I would say it is correct

pvojtechovsky:
> To me it seems ready for merge.

for me too :-)

I can proceed faster with less opened PRs :-) Rebasing is not so funny like improving awesome spoon code.

Next PR for review is #1005 please.

PR #1195, opened 2017-02-22
review: feature: All type members function

pvojtechovsky:
This PR Introduces two new mapping functions:
* `SuperHierarchyFunction` - which produces all super classes and super interfaces of input `CtType`
* `AllTypeMembersFunction` - which produces all `CtTypeMember` of input `CtType` and all super classes and super interfaces.

The `AllTypeMembersFunction` is then used, as example, to implement CtTypeImpl#getAllFields` in a nicer and more correct way.

This PR is ready for review.

### DIFF
--- a/src/main/java/spoon/reflect/visitor/filter/AllTypeMembersFunction.java
+++ b/src/main/java/spoon/reflect/visitor/filter/AllTypeMembersFunction.java
@@ -62,7 +62,7 @@ public class AllTypeMembersFunction implements CtConsumableFunction<CtType<?>>, 
 
 	@Override
 	public void apply(CtType<?> input, final CtConsumer<Object> outputConsumer) {
-		final CtQuery q = input.map(new SuperHierarchyFunction(distintSet == null ? new HashSet<String>() : distintSet).includingSelf(true));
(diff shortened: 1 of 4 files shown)


monperrus:
Looks good to me. It's tested by all tests using the refactored methods of CtTypeImpl.

One thing: rename SuperHierarchyFunction to InheritanceHierarchyFunction?

pvojtechovsky:
>  It's tested 

yes, it is. It took some time until it passed all these tests. I am really glad that spoon has that level of tests!

> rename SuperHierarchyFunction to InheritanceHierarchyFunction?

I vote for a name, which contains direction of traversing through inheritance hieararchy, because I already need an opposite function with name like `SubHierarchyFunction`. I am open to any change in these two names, but `InheritanceHierarchyFunction` does not express Super/Sub so it is not good name from that point of view. The both words `Inheritance` and `Hierarchy` are good start ...
A) SuperHierarchyFunction, SubHierarchyFunction
B) SuperInheritanceHierarchyFunction, SubInheritanceHierarchyFunction
C) SuperInheritanceFunction, SubInheritanceFunction
D) ??

monperrus:
I like SuperInheritanceHierarchyFunction, SubInheritanceHierarchyFunction

pvojtechovsky:
If tests pass, then it is finished from my point of view.

monperrus:
Thanks!

PR #1208, opened 2017-03-07
Fix NPE when empty working dir

### LINKED ISSUE #1116: NullPointerException when using Spoon inside an IntelliJ Plugin (works with .jar but not with Maven)
Hi !

**Context:** 
We are using Spoon library inside an IntelliJ plugin for many months. Until now, we used version 5.1.1 downloaded from your website ("spoon-core-5.1.1-jar-with-dependencies.jar"). Earlier this week, we updated it on v 5.4, also using the .jar (with dependencies). Till then, no problem.
On the 11th January (v5.5 release), we switched to a maven dependency. Unfortunately, it looks like there is an incompatibility with IntelliJ and the JDT-core jar included by Spoon ?
I also tried Spoon v5.4.0 (from Maven) VS Spoon v5.4.0 (jar-with-dependencies), but it does the same thing: the jar is OK but when using Maven, it throws this exception:

**Exception :**
```
	at org.eclipse.jdt.internal.compiler.batch.FileSystem.<init>(FileSystem.java:141)
	at org.eclipse.jdt.internal.compiler.batch.Main.getLibraryAccess(Main.java:3086)
	at spoon.support.compiler.jdt.JDTBatchCompiler.getUnits(JDTBatchCompiler.java:107)
	at spoon.support.compiler.jdt.JDTBasedSpoonCompiler.buildUnits(JDTBasedSpoonCompiler.java:410)
	at spoon.support.compiler.jdt.JDTBasedSpoonCompiler.buildUnitsAndModel(JDTBasedSpoonCompiler.java:372)
	at spoon.support.compiler.jdt.JDTBasedSpoonCompiler.buildSources(JDTBasedSpoonCompiler.java:348)
	at spoon.support.compiler.jdt.JDTBasedSpoonCompiler.build(JDTBasedSpoonCompiler.java:119)
	at spoon.support.compiler.jdt.JDTBasedSpoonCompiler.build(JDTBasedSpoonCompiler.java:102)
	at spoon.Launcher.buildModel(Launcher.java:700)
```

Using breakpoints and debugger, I found this : 
image

When you can see the NullPointerException, apparently because "paths" is null.
However, I am not sure that the problem come from this file. Here is the diff between the two version of the FileSystem.class (decompiled by IntelliJ), on the left side, the one from IntelliJ, on the right side, the Spoon one.
image

I tried several methods to "force use" the .jar from Spoon,  without success.
EDIT: Even when deleting the lib from IntelliJ (ecj-4.5.2.jar), the NP Exception is still thrown.

Have you any idea how I could fix that ?

Thanks !

surli:
Fix a funny bug which happens when launching Spoon in an empty working directory: we have a NPE because JDT cannot find a .java file... We fix it by passing as argument to JDT the list of java files we want to Spoon. It should then fix #1116

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/JDTBasedSpoonCompiler.java
+++ b/src/main/java/spoon/support/compiler/jdt/JDTBasedSpoonCompiler.java
@@ -392,7 +392,7 @@ public class JDTBasedSpoonCompiler implements spoon.SpoonModelBuilder {
 					.classpathOptions(new ClasspathOptions().encoding(this.encoding).classpath(classpath)) //
 					.complianceOptions(new ComplianceOptions().compliance(javaCompliance)) //
 					.advancedOptions(new AdvancedOptions().preserveUnusedVars().continueExecution().enableJavadoc()) //
-					.sources(new SourceOptions().sources()) // no sources, handled by the JDTBatchCompiler
+					.sources(new SourceOptions().sources(sourcesFolder.getFiles())) // no sources, handled by the JDTBatchCompiler
 					.build();
 		} else {
 			args = jdtBuilder.build();
--- a/src/main/java/spoon/support/compiler/jdt/JDTBasedSpoonCompiler.java
+++ b/src/main/java/spoon/support/compiler/jdt/JDTBasedSpoonCompiler.java
@@ -392,7 +392,7 @@ public class JDTBasedSpoonCompiler implements spoon.SpoonModelBuilder {
 					.classpathOptions(new ClasspathOptions().encoding(this.encoding).classpath(classpath)) //
 					.complianceOptions(new ComplianceOptions().compliance(javaCompliance)) //
 					.advancedOptions(new AdvancedOptions().preserveUnusedVars().continueExecution().enableJavadoc()) //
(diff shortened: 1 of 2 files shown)


src/main/java/spoon/support/compiler/jdt/JDTBasedSpoonCompiler.java
```diff
@@ -392,7 +392,7 @@ protected boolean buildUnitsAndModel(JDTBuilder jdtBuilder, SpoonFolder sourcesF
 					.classpathOptions(new ClasspathOptions().encoding(this.encoding).classpath(classpath)) //
 					.complianceOptions(new ComplianceOptions().compliance(javaCompliance)) //
 					.advancedOptions(new AdvancedOptions().preserveUnusedVars().continueExecution().enableJavadoc()) //
-					.sources(new SourceOptions().sources()) // no sources, handled by the JDTBatchCompiler
+					.sources(new SourceOptions().sources(sourcesFolder.getAllJavaFiles())) // no sources, handled by the JDTBatchCompiler
```
tdurieux: Remove the unrelated comment

PR #1211, opened 2017-03-08
review: feature: Parent function

pvojtechovsky:
The ParentFunction can now early terminate too. 

This PR is ready for review

### DIFF
--- a/src/main/java/spoon/reflect/visitor/filter/ParentFunction.java
+++ b/src/main/java/spoon/reflect/visitor/filter/ParentFunction.java
@@ -20,6 +20,8 @@ import spoon.reflect.declaration.CtElement;
 import spoon.reflect.declaration.CtPackage;
 import spoon.reflect.visitor.chain.CtConsumableFunction;
 import spoon.reflect.visitor.chain.CtConsumer;
+import spoon.reflect.visitor.chain.CtQuery;
+import spoon.reflect.visitor.chain.CtQueryAware;
 
 /**
  * This Function expects a {@link CtElement} as input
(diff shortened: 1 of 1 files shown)


monperrus:
What's the responsibility of ParentFunction?

pvojtechovsky:
It is a simple function which iterates over all parents, starting from nearest parent and ending on model root package.

monperrus:
Hi Pavel,

Could you update testParentFunction to give an example of when and how using `ParentFunction#setQuery`?

Is there an order to review and merge your remaining PRs?

pvojtechovsky:
> update testParentFunction 

ok, but on Saturday. The example is something like 
element.map(new ParentFunction(...filter...)).first()

> Is there an order to review and merge your remaining PRs?
* #1211 - is easy, it can be done first
* #1195 - is independent on others, so it can be done too
* #1005 needs #1193, #1194. So first these two and then we can discuss the design of #1005. It is worknig, but it needs class renames and discussion of how to do refactoring ... so this is last

pvojtechovsky:
> Could you update testParentFunction to give an example of when and how using ParentFunction#setQuery?

There is already test code
```java
assertSame(varStrings.getParent(), varStrings.map(new ParentFunction().includingSelf(false)).first());
```
which internally uses `ParentFunction#setQuery` and which profits from optimization made by this PR.

I do not understand, what else should be added. 

May be you mean: "Some code which demonstrates how client should use that?" No. Client does not have to be aware of that there is an setQuery method. It is called automatically behind the scene. It should stay hidden for clients. It is a way how to let **authors of mapping functions** to make them more optimal - to early terminated, if no more results are needed. It is nothing for clients of these functions. So nothing to write into test.

And it is actually impossible to test that ParentFunction really terminated early. It would need a hook (special testing code) in implementation of CtQueryImpl

monperrus:
Thanks for the explanation and well focused PR.

Issue #1213, opened 2017-03-10
CtFieldReference.getDefaultExpression() returns initializer from a field of another class


bastik:
Hi, I'm trying to collect and evaluate certain strings in the source repository.

I tried VisitorPartialEvaluator, but it runs into an infinite loop. The reason of this is that fields get mixed up. The code setup is like this:

```java
class BaseClass {
    public final static String PREFIX = "BasePrefix";
}

class ClassB {
    public final static String PREFIX = BaseClass.PREFIX + ".b";
    public String getKey() {
        return BaseClass.PREFIX + ".x";
    }
}
```

Now if you try to read the return value of the `getKey()` method, the CtFieldReference object will return the default value of `ClassB.PREFIX`, not `BaseClass.PREFIX`.

tdurieux:
this bug is related to #756

PR #1215, opened 2017-03-10
review: fix: Dynamic lookup issues by removing dynamic lookup

surli:
The issue #1213 shows that in some cases dynamic lookup just fail to find the right class. Then we propose to remove the dynamic lookup, and to improve the setName instead. 
According to @tdurieux dynamic lookup was introduced to support the clone operations: we propose instead to always use renaming operation after using clone, and then to improve renaming to also rename children of a type. 

A mandatory change is then needed: to use the `setName` operation *after* putting a type in the package.

### DIFF
--- a/src/main/java/spoon/refactoring/Refactoring.java
+++ b/src/main/java/spoon/refactoring/Refactoring.java
@@ -28,6 +28,33 @@ import java.util.List;
  */
 public final class Refactoring {
 	/**
+	 * Changes names of the inherited reference of type qualified by oldName
+	 *
+	 * @param type
+	 * 		Type in the AST.
+	 * @param oldName
+	 * 		The old qualified name of type
+	 * @param name
+	 * 		New name of the element.
+	 */
+	public static void changeInheritedTypeName(final CtType<?> type, String oldName, String name) {
+
+		final String typeQFN = oldName;
+
+		final List<CtTypeReference<?>> references = Query.getElements(type.getFactory(), new TypeFilter<CtTypeReference<?>>(CtTypeReference.class) {
+			@Override
+			public boolean matches(CtTypeReference<?> reference) {
+				String refFQN = reference.getQualifiedName();
+				return typeQFN.equals(refFQN);
+			}
+		});
+
+		for (CtTypeReference<?> reference : references) {
+			reference.setSimpleName(name);
+		}
+	}
+
+	/**
 	 * Changes name of a type element.
 	 *
 	 * @param type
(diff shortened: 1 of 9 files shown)


monperrus:
> dynamic lookup was introduced to support the clone operations

no, dynamic lookup is a core design decision that was already there at the birth of Spoon more than ten years ago

tdurieux:
Nope this lookup based on parents has been introduced 8 months ago to solve this issue #756

pvojtechovsky:
The dynamic lookup algorithm introduced in #756, is step in wrong direction. 
The dynamic lookup - old core feature of spoon - is OK and should stay there.

The problem of #756 is that client wants to rename class A to B, and it is kind of java model refactoring, which cannot be easily done by `CtType#setName()`, but which **must be done by an appropriate refactoring algorithm**, which includes check for validity of new name, conflict with other classes, etc.
So the original solution of #756, was simply wrong.

And I think that [solution](https://github.com/INRIA/spoon/pull/1215/files#diff-6773e1d8c6dd556d70292d554c78cf7dR53) of this PR is not good idea too, because it tries to pretend that renaming of class is a primitive operation. What if somebody wants to rename field, variable, method, parameter, ...? Are we able to implement everything into CtNamedElementImpl ?

Suggestion: May be the CtNamedElementImpl#setSimpleName should only check what kind of element we are renaming and by default throw an exception which says "Modification of model is not primitive operation. Please use refactoring methods from package ...spoon.refactor...". That exception would be thrown optionally depending on settings of spoon launcher/factory, so after clients are warned and they change that setting with idea "I know what I am doing", they can continue renaming without further warning exceptions ...

src/main/java/spoon/support/reflect/declaration/CtNamedElementImpl.java
```diff
@@ -43,6 +49,22 @@ public String getSimpleName() {
 		if (factory instanceof FactoryImpl) {
 			simpleName = ((FactoryImpl) factory).dedup(simpleName);
 		}
+
+		if (this instanceof CtType && !simpleName.equals("")) {
```
surli: We should not change setSimpleName here. Instead we should change the CloneVisitorGenerator to call Refactor.rename().

pvojtechovsky: > Refactor.rename() 

I agree.

src/main/java/spoon/refactoring/Refactoring.java
```diff
@@ -28,18 +28,22 @@
  */
 public final class Refactoring {
 	/**
-	 * Changes name of a type element.
+	 * Changes names of the inherited reference of type qualified by oldName
```
monperrus: outdated doc

PR #1216, opened 2017-03-11
review1: feature: CtTypeParameter#getTypeErasure()

pvojtechovsky:
Adds new method:
* CtTypeParamer#getTypeErasure() - returns type erasure of generic parameter

This method and #1225  are needed to correctly detect whether method overrides another method following [specification](https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.4.2)

### DIFF
--- a/src/main/java/spoon/reflect/declaration/CtTypeInformation.java
+++ b/src/main/java/spoon/reflect/declaration/CtTypeInformation.java
@@ -160,4 +160,11 @@ public interface CtTypeInformation {
 	@DerivedProperty
 	Collection<CtExecutableReference<?>> getAllExecutables();
 
+	/**
+	 * @return type (not generic one), which is used by java compiler to ensure that no new classes are created for parameterized types;
+	 * consequently, generics incur no runtime overhead.
+	 * See https://docs.oracle.com/javase/tutorial/java/generics/erasure.html
+	 */
+	@DerivedProperty
+	CtTypeReference<?> getTypeErasure();
 }
--- a/src/main/java/spoon/reflect/declaration/CtTypeInformation.java
+++ b/src/main/java/spoon/reflect/declaration/CtTypeInformation.java
@@ -161,8 +161,7 @@ public interface CtTypeInformation {
 	Collection<CtExecutableReference<?>> getAllExecutables();
 
 	/**
-	 * @return type (not generic one), which is used by java compiler to ensure that no new classes are created for parameterized types;
-	 * consequently, generics incur no runtime overhead.
+	 * @return the type erasure, which is computed by the java compiler to ensure that no new classes are created for parametrized types so that generics incur no runtime overhead.
 	 * See https://docs.oracle.com/javase/tutorial/java/generics/erasure.html
 	 */
 	@DerivedProperty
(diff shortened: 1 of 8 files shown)


monperrus:
`getTypeErasure` is an excellent idea, thanks.

To me, the type erasure is computed when the type parameter is used not defined. This means that `getTypeErasure` should rather be in `CtTypeParameterReference`. What do you think?

(also could we split the discussion of `getTypeErasure` and of `getTypeAdaptedTo` in two separate PRs?)

pvojtechovsky:
[Computing](https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.4.2) whether method A overrides method B uses type erasure. So it is not linked to CtTypeParameterReference. But we can have this method on both CtTypeParameter and CtTypeParameterReference. I personally need it on CtTypeParameter

>  split the discussion of getTypeErasure and of getTypeAdaptedTo in two separate PRs

I will check if it can be split

monperrus:
> Computing <https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.4.2> whether method
> A overrides method B uses type erasure. So it is not linked to CtTypeParameterReference.
I'm not sure about this. In a signature `T foo(U x)` both the return type and the parameter types
are CtType(Parameter)Reference. The method signatures don't contani CtTypeParameter directly.

pvojtechovsky:
There are two types of type erasures. Till now I though only about this case:
```java
class<A extends InputStream> {
    A stream;
    <B extends Reader> void setReaderLike(B reader) {}
}
```
Such typed parameters naturally computes erasure using code on `CtTypeParameter` (like it was in this PR till now).

Note that 
* both `CtTypeParameter` and `CtTypeParameterReference` directly contains superClass/superType, which represents their bound. 
* only `CtTypeParameter` has declaringType.

The `bound` and `declaringType` are needed to compute type erasure. So from that point of view thos code belongs 1st to CtTypeParameter. And from client's point of view it makes sense life easier if API for `CtTypeParameterReference` exists too.

But thanks to your note, I see now that there is another place where type erasure makes sense:
```java
class X {
   void processList(List<? extends Reader> list) {};
}
```
In this case there is no CtTypeParameter or CtTypeParameterReference involved at all. In this case we need new method `CtTypeReference#getTypeErasure()`

So I suggest these steps:
1) to merge this PR with `CtTypeParameter#getTypeErasure()` only. (I will remove `getTypeAdaptedTo`)
2) to create another PR, which handles type erasures on all type references `CtTypeReference#getTypeErasure()`. This code will sometime call `CtTypeParameter#getTypeErasure()` and sometime it will use new algorithm, which is not implemented yet.

OK?

pvojtechovsky:
Now this PR contains only plain `CtTypeParameter#getTypeErasure()`.

Please review. Thank you!

pvojtechovsky:
I have to think/learn more about generics and parameters. May be you are right that CtTypeReferences has to be involved more in the Type erasure process...

pvojtechovsky:
I have simplified computing of type erasure and I added CtTypeParameterReference#getTypeErasure too. Is such implementation OK, or did you expected something more for CtTypeParameterReference#getTypeErasure ?

monperrus:
The signature `CtTypeReference<?> getTypeErasure()` is clear. 

However, I'm not sure about the location.
What about putting this method in CtTypeReference? or even maybe in CtTypeInformation?

pvojtechovsky:
I like this idea!
I implemented it as CtTypeInformation#getTypeErasure()

monperrus:
Thanks Pavel.

pvojtechovsky:
Thanks Martin.

PR #1217, opened 2017-03-12
review 1: feature: CtTypeParameter#getTypeDeclarer()

pvojtechovsky:
During work with `CtTypeParameter` I always needed to know where this parameter is declared. I suggest to introduce this new API method into `CtTypeParameter`.
```java
	/**
	 * @return the {@link CtFormalTypeDeclarer}, which declares this {@link CtTypeParameter}
	 */
	@DerivedProperty
	CtFormalTypeDeclarer getTypeDeclarer();
```

This PR fixes bug in `TypeFactory#createReference(CtTypeParameter)`, which created reference without parent, so there was lost information about type parameter declarer. CtTypeParameterReference without declarer is useless.

### DIFF
--- a/src/main/java/spoon/reflect/declaration/CtTypeParameter.java
+++ b/src/main/java/spoon/reflect/declaration/CtTypeParameter.java
@@ -34,6 +34,12 @@ public interface CtTypeParameter extends CtType<Object> {
 	@DerivedProperty
 	CtTypeParameterReference getReference();
 
+	/**
+	 * @return the {@link CtFormalTypeDeclarer}, which declares this {@link CtTypeParameter}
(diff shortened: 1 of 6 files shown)


monperrus:
This method is good. 

Two questions before merge:
- rename as `getTypeParameterDeclarer`?
- could you add one test?

Thanks.

pvojtechovsky:
Rename done:
2 tests added
* one for getTypeParameterDeclarer
* second for CtTypeParameter.getReference().getDeclaration(), which did not worked before and is fixed by this PR too.

I am done. Please review

monperrus:
thanks

Issue #1160, opened 2017-02-02
[generic type] feature: isSubtypeOf is not supported with typeParameters


surli:
When writing a code like that:
```java
class O<A extends X> {
	<B extends A> B foo() {
		return null;
	}
}
```
Testing `b.isSubtypeOf(a)` will always return false for now. 
Test has been added in #1157 but is ignored for now. We should start from there.

PR #1218, opened 2017-03-12
review1: feature: CtTypeParameter#isSubtypeOf

pvojtechovsky:
Implements #1160.

The implementation of isSubtypeOf on type parameters is quite complex task, which needs all the basic steps below, which fits to commits of this PR.

There are:
BS1) new interface GenericTypeAdapter
BS2) CtSuperTypeHierarchy - adapts generic types to Type scope
BS3) CtMethodSuperTypeHierarchy - adapts generic types to Method/Constructor scope
BS4) CtFormalTypeDeclarer#getGenericTypeAdapter(), which creates instance of CtSuperTypeHierarchy  or CtMethodSuperTypeHierarchy for the scope of this declarer
BS5) CtTypeParameterImpl#isSubtypeOf - finally the implementation of isSubtypeOf  - the target of this PR
BS6) simpler and more correct implementation of `CtTypeParameterImpl#isSubtypeOf(CtTypeReference)`, based on new algorithms
+ first and incomplete tests ... the testing of this java generics stuff is quite tricky ... too many combinations.

### DIFF
--- a/src/test/java/spoon/test/ctType/CtTypeTest.java
+++ b/src/test/java/spoon/test/ctType/CtTypeTest.java
@@ -2,17 +2,29 @@ package spoon.test.ctType;
 
 import org.junit.Test;
 import spoon.Launcher;
+import spoon.reflect.code.CtAssignment;
+import spoon.reflect.code.CtComment;
+import spoon.reflect.code.CtLocalVariable;
+import spoon.reflect.code.CtVariableAccess;
 import spoon.reflect.declaration.CtClass;
 import spoon.reflect.declaration.CtInterface;
 import spoon.reflect.declaration.CtMethod;
+import spoon.reflect.declaration.CtNamedElement;
 import spoon.reflect.declaration.CtType;
-import spoon.reflect.declaration.CtTypeMember;
 import spoon.reflect.declaration.CtTypeParameter;
 import spoon.reflect.factory.Factory;
+import spoon.reflect.reference.CtReference;
 import spoon.reflect.reference.CtTypeReference;
+import spoon.reflect.visitor.filter.NameFilter;
+import spoon.reflect.visitor.filter.TypeFilter;
 import spoon.test.ctType.testclasses.X;
 
+import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
@@ -84,27 +96,98 @@ public class CtTypeTest {
 		//using CtType implementation
 		assertTrue(xCtType.isSubtypeOf(xCtType.getReference()));
 	}
-
+	
 	@Test
 	public void testIsSubTypeOfonTypeParameters() throws Exception {
 		CtType<X> xCtType = buildClass(X.class);
+		Factory factory = xCtType.getFactory();
 
-		CtType<?> oCtType = xCtType.getFactory().Type().get("spoon.test.ctType.testclasses.O");
+		CtType<?> oCtType = factory.Type().get("spoon.test.ctType.testclasses.O");
+		CtType<?> pCtType = factory.Type().get("spoon.test.ctType.testclasses.P");
+		CtTypeReference<?> objectCtTypeRef = factory.Type().OBJECT;
 
-		List<CtTypeParameter> typeParameters = oCtType.getFormalCtTypeParameters();
-		assertTrue(typeParameters.size() == 1);
+		List<CtTypeParameter> oTypeParameters = oCtType.getFormalCtTypeParameters();
+		assertTrue(oTypeParameters.size() == 1);
+		List<CtTypeParameter> pTypeParameters = pCtType.getFormalCtTypeParameters();
+		assertTrue(pTypeParameters.size() == 2);
 
-		CtType<?> aCtType = typeParameters.get(0);
+		CtType<?> O_A_CtType = oTypeParameters.get(0);
+		CtType<?> P_D_CtType = pTypeParameters.get(0);
+		CtType<?> P_F_CtType = pTypeParameters.get(1);
 
-		List<CtMethod<?>> methods = oCtType.getMethodsByName("foo");
+		CtMethod<?> O_FooMethod = oCtType.filterChildren(new NameFilter<>("foo")).first();
+		CtMethod<?> P_FooMethod = pCtType.filterChildren(new NameFilter<>("foo")).first();
 
-		assertTrue(methods.size() == 1);
+		CtType<?> O_B_CtType = O_FooMethod.getType().getDeclaration();
+		CtType<?> P_E_CtType = P_FooMethod.getType().getDeclaration();
 
-		CtMethod<?> fooMethod = methods.get(0);
-		CtType<?> bCtType = fooMethod.getType().getDeclaration();
+		assertTrue(O_B_CtType.isSubtypeOf(xCtType.getReference()));
+		assertTrue(O_B_CtType.isSubtypeOf(O_A_CtType.getReference()));
+		
+		assertTrue(P_E_CtType.isSubtypeOf(xCtType.getReference()));
+		assertTrue(P_E_CtType.isSubtypeOf(P_D_CtType.getReference()));
+		assertTrue(P_E_CtType.isSubtypeOf(O_A_CtType.getReference()));
+		
+		assertTrue(P_D_CtType.isSubtypeOf(O_A_CtType.getReference()));
+		assertTrue(P_E_CtType.isSubtypeOf(O_B_CtType.getReference()));
 
-		//The TypeParameters are not supported yet
-//		assertTrue(bCtType.isSubtypeOf(xCtType.getReference()));
-//		assertTrue(bCtType.isSubtypeOf(aCtType.getReference()));
+		assertTrue(P_E_CtType.isSubtypeOf(objectCtTypeRef));
+		assertTrue(P_F_CtType.isSubtypeOf(objectCtTypeRef));
+	}
+	
+	@Test
+	public void testIsSubTypeOfonTypeReferences() throws Exception {
+		final Launcher launcher = new Launcher();
+		launcher.setArgs(new String[]{"-c"});
+		launcher.addInputResource("./src/test/java/spoon/test/ctType/testclasses/SubtypeModel.java");
+		launcher.buildModel();
+		Factory factory = launcher.getFactory();
+		
+		CtType<?> oCtType = factory.Class().get("spoon.test.ctType.testclasses.SubtypeModel");
+		CtMethod<?> O_FooMethod = oCtType.filterChildren(new NameFilter<>("foo")).first();
+
+		Map<String, CtTypeReference<?>> nameToTypeRef = new HashMap<>();
+		O_FooMethod.filterChildren(new TypeFilter<>(CtLocalVariable.class)).forEach((CtLocalVariable var)->{
+			nameToTypeRef.put(var.getSimpleName(), var.getType());
+		});
+		
+		int[] count = new int[1];
+
+		O_FooMethod.filterChildren(new TypeFilter<>(CtAssignment.class)).forEach((CtAssignment ass)->{
+			for (CtComment comment : ass.getComments()) {
+				checkIsNotSubtype(comment, nameToTypeRef);
+				count[0]++;
+			};
+			count[0]++;
+			checkIsSubtype(((CtVariableAccess) ass.getAssigned()).getVariable().getType(), ((CtVariableAccess) ass.getAssignment()).getVariable().getType(), nameToTypeRef);
+		});
+		
+		assertTrue(count[0]>(9*8));
+	}
+
+	private void checkIsSubtype(CtTypeReference superType, CtTypeReference subType, Map<String, CtTypeReference<?>> nameToTypeRef) {
+		String msg = getTypeName(subType)+" isSubTypeOf "+getTypeName(superType);
+		assertTrue(msg, subType.isSubtypeOf(superType));
+	}
+
+	private static final Pattern assignment = Pattern.compile("\\s*(\\w+)\\s*=\\s*(\\w+);");
+	private void checkIsNotSubtype(CtComment comment, Map<String, CtTypeReference<?>> nameToTypeRef) {
+		Matcher m = assignment.matcher(comment.getContent());
+		assertTrue(m.matches());
+		CtTypeReference<?> superType = nameToTypeRef.get(m.group(1));
+		CtTypeReference<?> subType = nameToTypeRef.get(m.group(2));
+		String msg = getTypeName(subType)+" is NOT SubTypeOf "+getTypeName(superType);
+		assertFalse(msg, subType.isSubtypeOf(superType));
+	}
+	
+	private String getTypeName(CtTypeReference<?> ref) {
+		String name;
+		CtReference r= ref.getParent(CtReference.class);
+		if(r!=null) {
+			name = r.getSimpleName();
+		} else {
+			name = ref.getParent(CtNamedElement.class).getSimpleName();
+		}
+		return ref.toString()+" "+name;
 	}
 }
--- a/src/test/java/spoon/test/generics/GenericsTest.java
+++ b/src/test/java/spoon/test/generics/GenericsTest.java
@@ -1,21 +1,6 @@
 package spoon.test.generics;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.assertSame;
-import static spoon.testing.utils.ModelUtils.build;
-import static spoon.testing.utils.ModelUtils.buildClass;
-import static spoon.testing.utils.ModelUtils.buildNoClasspath;
-import static spoon.testing.utils.ModelUtils.canBeBuilt;
-import static spoon.testing.utils.ModelUtils.createFactory;
-
-import java.util.ArrayList;
-import java.util.List;
-
 import org.junit.Test;
-
 import spoon.Launcher;
 import spoon.SpoonModelBuilder;
 import spoon.compiler.SpoonResourceHelper;
@@ -30,13 +15,13 @@ import spoon.reflect.declaration.CtClass;
 import spoon.reflect.declaration.CtConstructor;
 import spoon.reflect.declaration.CtElement;
 import spoon.reflect.declaration.CtField;
-import spoon.reflect.declaration.CtFormalTypeDeclarer;
 import spoon.reflect.declaration.CtInterface;
 import spoon.reflect.declaration.CtMethod;
 import spoon.reflect.declaration.CtNamedElement;
 import spoon.reflect.declaration.CtParameter;
 import spoon.reflect.declaration.CtType;
 import spoon.reflect.declaration.CtTypeParameter;
+import spoon.reflect.declaration.CtTypedElement;
 import spoon.reflect.factory.Factory;
 import spoon.reflect.reference.CtReference;
 import spoon.reflect.reference.CtTypeParameterReference;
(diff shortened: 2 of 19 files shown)


monperrus:
Hi Pavel,

> This PR contains #1216 and #1217, so we should first discuss them.

You're so prolific with PRs :-) that it would help if you add the review/merge priority  in the PR title itself

pvojtechovsky:
> that it would help if you add the review/merge priority in the PR title itself

I will use these prefixes:
* review 1: ... ready for review, without dependency  to others. Please check it first
* review 2: ... ready for review, with an dependency to other PRs. I am done with that too, so it is possible to discuss that solution too. But priority is lower then "review 1: "
* WiP - work in progress - contains an unfinished feature. I am not publishing everything what I do, just things where early feedback would be welcome, to influence the final solution as soon as possible.

pvojtechovsky:
I updated first comment of this PR to better fit current content. 

First I would like to ask you to have a short look at each commit separately - to get overview about all these topics. Then I would like to discuss with you how many PR's we will make out of this and what will be their content.

pvojtechovsky:
```java
//model
<T extends spoon.reflect.code.CtStatement> T getLastStatement();
//test code
CtTypeParameterReference paramT = ...//'T' from example above
assertTrue(paramT.isSubtypeOf(factory.Code().createCtTypeReference(CtElement.class)));
```
Do you agree that `T` from the model above is sub type of `CtElement`? I think it is exactly the request from issue #1160 so it is correct that new implementation returns true there.

But this method `SpoonTestHelpers#isMetamodelRelatedType` called by `CtScannerTest#testScannerCallsAllProperties` expects opposite. Do you agree to fix that method? 
@monperrus Could you please suggest new implementation of `SpoonTestHelpers#isMetamodelRelatedType` ?

pvojtechovsky:
I see now that problem is not in `SpoonTestHelpers#isMetamodelRelatedType`, which works correctly. But there is problem with missing declaration of Derived property CtBlock#getLastStatement(), which is now correctly detected by fixed test.

pvojtechovsky:
In legacy spoon code there are problems with CtExecutableReference, which 
A) has no parent
B) the invocation scope is the parent
In both cases there is not possible to resolve declaration of CtTypeParameterReference, which is used as type of executable parameters. I have committed here some workarounds [W1](https://github.com/INRIA/spoon/pull/1218/commits/ffab37263d416be0476cd26ebb23c07a6724deb4) and [W2](https://github.com/INRIA/spoon/pull/1218/commits/f4df61c23ed54895103da1b1ba15af69c1b86a65), but I am not happy from such "solutions". I would prefer something nicer. But what? It is similar problem like #1237, so I need your help here. It seems to me like there is not possible to use dynamic lookup in this case or ... ? 

WDYT?

pvojtechovsky:
Tests finally passed! Jupeee .. now "just" little step ... to pass Martin ;)

monperrus:
Thanks Pavel.

The first comment is that 22 files changed, it's a lot. I'm not sure they're baby Prs and that they are all necessary here. Maybe we'll have to break this into smaller PRs.

I first concentrate on "adaptType':

- do the changes and comments I've just pushed in the tests correspond to your idea?
- can we rename adaptType to resolveType?

pvojtechovsky:
You are right, they are not baby steps. I am aware of that, so let's discuss what how to break it into smaller useful parts.

> I first concentrate on `adaptType':
good choice. The adaptType of type parameter declared in scope of CtType is simpler.
Then I suggest to discuss new implementation of `isSubtypeOf` of CtType, which checks actual type arguments too.
Then adaptType of type parameter declared in scope of CtMethod
Then isOverriding, isSubSignature, isSameSignature of CtMethodSuperTypeHierarchy
And last it makes sense to check isSubTypeOf of type parameter (which needs all previous steps) = the target of this PR.

> do the changes and comments I've just pushed in the tests correspond to your idea?

yes, they are OK, Thanks for that!

> can we rename adaptType to resolveType?

why? [Java lang specification](https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.4.2) uses term "adapting"
> a type mentioned in N can be **adapted** to the type parameters of M

so why do you look for different name? I think that "adapt" represents that process better then "resolve", because there is source context X and target contexts Y and that process is adapting the type from one context into another context. There can be more target contexts. I think "resolving" is more related to founding of ONE solution, and it is not the case of that adapting process, where you can adapt into more then one target contexts.

But you are the master of spoon, so if you still like "resolve" more, I can rename it and live with that. No problem :-)

src/test/java/spoon/test/generics/GenericsTest.java
```diff
@@ -695,36 +725,39 @@ public void testCtSuperTypeHierarchy() throws Exception {
...
-		//adapt A to scope of enclosing class of CelebrationLunch<Integer,Long,Double>.WeddingLunch<Mole>, which is CelebrationLunch<Integer,Long,Double>
-		adapted = sthOftrWeddingLunch_Mole.getEnclosingGenericTypeAdapter().adaptType(ctClassLunch_A);
-		assertEquals("java.lang.Double", adapted.getQualifiedName());
-		
+
+		// in disgust, the A of Lunch is bound to "Mole"
+		assertEquals("spoon.test.generics.testclasses.Mole", typingContextOfDisgust.adaptType(ctClassLunch_A).getQualifiedName());
+
+		// I don't understand the goal and utility of this one
+		assertEquals("java.lang.Double", typingContextOfDisgust.getEnclosingGenericTypeAdapter().adaptType(ctClassLunch_A).getQualifiedName());
```
pvojtechovsky: This shows that java class hierarchies of inner classes are different on each inner class level. In this test model there are two INDEPENDENT class hierarchies:
First is hierachy of top level classes:
```java
CelebrationLunch<K,L,M> extends Lunch<M,K>
```
and second there is hierarchy of inner class
```java
WeddingLunch<X> extends CelebrationLunch<Tacos, Paella, X> extends Lunch<X,Tacos>
```
Because in this wild example model, both ( 1. top level class `CelebrationLunch` and 2. inner class `WeddingLunch` ) extends from `Lunch<A,B>`, there is necessary to be able to adapt type parameters A and B in **both** contexts.

So this test line shows how to move from context of inner class `WeddingLunch` to the parent context of top level class `CelebrationLunch` and to adapt type parameters there.

May be it is confusing in this example that 
1) `WeddingLunch` is inner class of `CelebrationLunch`
2) `WeddingLunch` extends `CelebrationLunch` too
The `typingContextOfDisgust.getEnclosingGenericTypeAdapter()` is move from inner class to declaring class. It is NOT move from sub type to super type.

src/test/java/spoon/test/generics/GenericsTest.java
```diff
@@ -695,36 +725,39 @@ public void testCtSuperTypeHierarchy() throws Exception {
...
+		assertEquals("X", sthOftWeddingLunch_X.adaptType(ctWeddingLunch_X).getQualifiedName());
+
+		// in WeddingLunch the "A" from Lunch of is called "X"
+		assertEquals("X", sthOftWeddingLunch_X.adaptType(ctClassLunch_A).getQualifiedName());
+
+		// ?????
 		//adapt A to scope of enclosing class of CelebrationLunch<K,L,M>.WddingLunch<X>, which is CelebrationLunch<K,L,M>
-		adapted = sthOftWeddingLunch_X.getEnclosingGenericTypeAdapter().adaptType(ctClassLunch_A);
-		assertEquals("M", adapted.getQualifiedName());
+		assertEquals("M", sthOftWeddingLunch_X.getEnclosingGenericTypeAdapter().adaptType(ctClassLunch_A).getQualifiedName());
```
pvojtechovsky: Same like above. In this simplified model
```
class Lunch<A,B> {}
class CelebrationLunch<K,L,M> extends Lunch<M,K> {}
```
it shows that type parameter `A` of `Lunch` is adapted to `M` of `CelebrationLunch`.

src/test/java/spoon/test/generics/GenericsTest.java
pvojtechovsky: It is test of contract of `getEnclosingGenericTypeAdapter` method

monperrus: To simplify, can we create instead an adapter of the super class such as  new CtSuperTypeHierarchy(Lunch).adaptType(ctClassLunch_A).getQualifiedName())? 
Can we remove this method?

src/test/java/spoon/test/generics/GenericsTest.java
```diff
@@ -633,7 +640,137 @@ public void testGenericTypeReference() throws Exception {
...
+		
+		//are these two types same?
+		CtSuperTypeHierarchy typingContextOfDisgust = new CtSuperTypeHierarchy(typeReferenceOfDisgust);
+		
+		// in disgust, X of WeddingLunch is bound to "Model"
+		assertEquals("spoon.test.generics.testclasses.Mole", typingContextOfDisgust.adaptType(ctWeddingLunch_X).getQualifiedName());
+		//adapt A to scope of CelebrationLunch<Integer,Long,Double>.WeddingLunch<Mole>
+
+		// in disgust, the A of Lunch is bound to "Mole"
+		assertEquals("spoon.test.generics.testclasses.Mole", typingContextOfDisgust.adaptType(ctClassLunch_A).getQualifiedName());
```
monperrus: I have a doubt here, the "Lunch.A" of disgust should be Double?

monperrus:
> why? Java lang specification uses term "adapting" 

OK

src/test/java/spoon/test/generics/GenericsTest.java
pvojtechovsky: client can of course create new CtSuperTypeHierarchy using correct reference or type and then s/he gets the same result.

> Can we remove this method?

There is hierarchy of GenericTypeAdapter (GTA) instances, like GTA of Method -> GTA of Type -> GTA of enclosing type (recursivelly up to top level type). It is necessary because java allows such complex structures, so I think it is correct that GenericTypeAdapter has method getEnclosingGenericTypeAdapter, which gives access to parent adapter, which is already there and whose construction may need some computation ... so it is inefficient to create GenericTypeAdapter again if the needed one is already there.

If you want to remove it just because clients will not understand it, then note that all implementations of GenericTypeAdapter may be understood as internal helper classes. We can create nice facade methods in existing spoon model elements, so the normal client's does not need to learn/see that.

src/test/java/spoon/test/generics/GenericsTest.java
pvojtechovsky: In the hierarchy of WeddingLunch<Mole> the Lunch.A is Mole
In the hierarchy of CelebrationLunch<Integer,Long,Double> the Lunch.A is Double

monperrus: OK.

src/test/java/spoon/test/generics/GenericsTest.java
monperrus: >  I think it is correct that GenericTypeAdapter has method getEnclosingGenericTypeAdapter, which gives access to parent adapter,

What does "parent" mean in this context? the AST parent? or the super class? or something else?

pvojtechovsky: I am sorry, "parent" was not the best name. I mean "declaring type". See example:
```java
class A1<K1> {
  class B1<L1> {
  <M1> M1 method(K1 k, L1 l){}
  }
}
class A2<K2> extends A1<K2> {
  class B2<L2> extends B1<L2> {
  <M2> M2 method(K2 k, L2 l){}
  }
}
```
In the model above the 
```java
//adapter for method `A2.B2#method`
CtMethodSuperTypeHierarchy methodAdapter = new CtMethodSuperTypeHierarchy(...method("A2.B2#method"));

//the enclosing adapter of `methodAdapter` is adapter of inner class `A2.B2`
CtSuperTypeHierarchy b2Adapter = methodAdapter.getEnclosingGenericTypeAdapter();
//the enclosing adapter of `b2Adapter ` is adapter of top level class `A2`
CtSuperTypeHierarchy a2Adapter = b2Adapter.getEnclosingGenericTypeAdapter();
//the enclosing adapter of `a2Adapter ` is null, because A2 is top level class
assertNull(a2Adapter.getEnclosingGenericTypeAdapter())
```
Note that `methodAdapter` internally needs adapters of all enclosing declaring types recursively, because that method can contain typed parameters of 
1) method itself - there is type parameter M2
2) declaring type B2 - there is type parameter L2
3) declaring type A2 - there is type parameter K2

pvojtechovsky:
The java language model is quite complicated if you start to combine inner classes, type parameters, generic methods and legacy methods ... Even after 17 years of daily java programming I had to learn a lot to understand how it works, to be able to implement this PR. 

It would be bad if spoon client's would need to understand it too, to be able to use that. Therefore I think that we should create easy and understandable facade methods in the spoon model, which will hide that complexity. 
The most of these facade methods are already there:
* CtTypeInformation#isSubtypeOf - already uses GenericTypeAdapter 
* CtExecutableReference#isOverriding - will use GenericTypeAdapter after we finish this PR
Others might be added soon. Something like:
* CtMethod#isOverriding

And it question whether somebody needs these methods:
//adapts type to the scope of this declarer
* CtFormalTypeDeclarer.adaptFrom(CtTypeInformation type)
* CtTypeReference.adaptFrom(CtTypeInformation)
* CtExecutableReference.adaptFrom(CtTypeInformation)
or
//adapts this type or reference to the `scope`
* CtTypeInformation.adaptTo(CtFormalTypeDeclarer scope)
* CtTypeInformation.adaptTo(CtTypeReference scope)
* CtTypeInformation.adaptTo(CtExecutableReference scope)

But it does not mean that we should make GenericTypeAdapter and it's implementations private. We can just put them into some special spoon package so they are less visible for normal clients, but can be still directly used by experienced clients and by spoon core itself.

src/test/java/spoon/test/parent/ParentContractTest.java
```diff
@@ -95,6 +95,9 @@ public void testContract() throws Throwable {
 		// contract: all setters/adders must set the parent (not necessarily the direct parent, can be upper in the parent tree, for instance when injecting blocks
 		Object o = factory.Core().create((Class<? extends CtElement>) toTest.getActualClass());
 		for (CtMethod<?> setter : SpoonTestHelpers.getAllSetters(toTest)) {
+			if("setParent".equals(setter.getSimpleName())) {
```
monperrus: move to SpoonTestHelpers.getAllSetters?

monperrus:
I agree with the core semantic and API design of  `adaptType`. Now, the plan is 1) to find good names 3) to identify if all changes in the 22 changed files of this PR are required and/or can be split into smaller separate baby PRs. 

Let's start with the naming issue, I propose the following:
CtSuperTypeHierarchy -> ClassTypingContext
CtMethodSuperTypeHierarchy -> MethodTypingContext

src/main/java/spoon/reflect/code/CtBlock.java
```diff
@@ -87,6 +88,7 @@
 	/**
 	 * Gets the last statement of this block.
 	 */
+	@DerivedProperty
```
monperrus: correct, move to separate PR?

src/main/java/spoon/reflect/declaration/CtFormalTypeDeclarer.java
```diff
@@ -43,4 +45,9 @@
 	 * Removes a type parameters from this generic element.
 	 */
 	boolean removeFormalCtTypeParameter(CtTypeParameter formalTypeParameter);
+
+	/**
+	 * @return the instance of {@link GenericTypeAdapter} which can adapt type parameters from different scopes to scope of this declarer
+	 */
+	GenericTypeAdapter getGenericTypeAdapter();
```
monperrus: really hard to understand, I wouldn't put it here. Is it required?

If we remove it, we won't have to touch all the classes implementing it, and this simplifies the PR much.

src/test/java/spoon/reflect/visitor/CtScannerTest.java
```diff
@@ -99,18 +99,22 @@ public void testScannerCallsAllProperties() throws Exception {
...
-				if (isMetamodelProperty(t, m)) {
-					//System.out.println("checking "+m.getSignature() +" in "+visitMethod.getSignature());
-					assertTrue("no "+m.getSignature() +" in "+visitMethod, visitMethod.getElements(new TypeFilter<CtInvocation>(CtInvocation.class) {
-						@Override
-						public boolean matches(CtInvocation element) {
-							return super.matches(element) && element.getExecutable().getSimpleName().equals(m.getSimpleName());
-						}
-					}).size()>0);
-				}
+				checkProperty(t,m,visitMethod);
```
monperrus: refactoring only? move in separate refactoring PR?

src/test/java/spoon/test/architecture/SpoonArchitectureEnforcerTest.java
```diff
@@ -71,6 +71,8 @@ public void process() {
 				for (CtType<?> t : factoryPackage.getTypes()) {
 					if (t.getSimpleName().startsWith("Mega")) continue; //
 					for (CtMethod<?> m : t.getMethods()) {
+						// we check only public methods
+						if (m.hasModifier(ModifierKind.PUBLIC) == false) continue;
```
monperrus: idem

pvojtechovsky:
> renaming

I like new names a lot! I will rename it.

what about name of GenericTypeAdapter? Can we use something name-compatible too? Like
TypingContext
GenericTypingContext
...

I will make smaller PRs for test bug fixing today

monperrus:
> what about name of GenericTypeAdapter?

It depends on whether we emphasize on what it represents (a typing context) or what it is used for (type adaptation).

Since an interface focuses on a usage, a responsibility, I prefer to keep GenericTypeAdapter.

pvojtechovsky:
What about package for GenericTypeAdapter and ClassTypingContext and MethodTypingContext. They are not filters, so it probably does not belongs to `spoon.reflect.visitor.filter`.

WDYT?

src/main/java/spoon/reflect/declaration/CtFormalTypeDeclarer.java
pvojtechovsky: I use it here:
```java
class CtTypeParameterImpl {
...
	public boolean isSubtypeOf(CtTypeReference<?> superTypeRef) {
		if (superTypeRef instanceof CtTypeParameterReference) {
			//the type is type parameter too. Use appropriate sub type checking algorithm
			CtTypeParameter superTypeParam = (CtTypeParameter) superTypeRef.getDeclaration();
			return isSubtypeOf(getTypeParameterDeclarer().getGenericTypeAdapter(), this, superTypeParam);
		}
		//type is normal type
		return getTypeErasure().isSubtypeOf(superTypeRef);
	}
```
Without that `getGenericTypeAdapter` method, I would need a `if` with 3 cases:
1) CtType
2) CtMethod
3) CtConstructor
where each case calls appropriate constructor of GenericTypeAdapter implementation.

So this method actually represents a way how to create instance of implementation of GenericTypeAdapter. 

What about adding a spoon factory method instead? Please suggest name and factory class if you agree. Should I use an AbstractSpoonVisitor instead of ugly `if`?

I am waiting for your feedback before I (re)move that code.

src/test/java/spoon/reflect/visitor/CtScannerTest.java
pvojtechovsky: I will skip this refactoring.

pvojtechovsky:
Rename done.

The required test fixes were moved to separate PRs #1261, #1262, #1263.
This PR still contains 3 commits from these PRs to have it compilable and testable. I will rebase it after they are merged.

Next steps please?

src/main/java/spoon/reflect/declaration/CtFormalTypeDeclarer.java
monperrus: > What about adding a spoon factory method instead? Should I use an AbstractSpoonVisitor instead of ugly if?

Excellent. What about `TypeFactory#createTypeAdapter()`?

monperrus:
we need a more appropriate package than `spoon/reflect/visitor/filter/`.

What about  simply `spoon/support/visitor/`?

pvojtechovsky:
GenericTypeAdapter, ClassTypingContext and MethodTypingContext moved to `spoon/support/visitor/`

CtFormalTypeDeclarer#getGenericTypeAdapter removed. Created `TypeFactory#createTypeAdapter(CtFormalTypeDeclarer)` instead.

What next? ;-)

src/test/java/spoon/test/ctType/CtTypeTest.java
```diff
@@ -84,27 +96,98 @@ public void testIsSubTypeOf() throws Exception {
...
+				count[0]++;
+			};
+			count[0]++;
+			checkIsSubtype(((CtVariableAccess) ass.getAssigned()).getVariable().getType(), ((CtVariableAccess) ass.getAssignment()).getVariable().getType(), nameToTypeRef);
+		});
+		
+		assertTrue(count[0]>(9*8));
+	}
 
-		CtType<?> oCtType = xCtType.getFactory().Type().get("spoon.test.ctType.testclasses.O");
```
monperrus: never delete existing test code in a feature PR.

when reviewing, if no test code is deleted, with only addition, it's easy to be sure that there is no regression.

this is particularly true for large PRs such as this one.

monperrus:
let's close this old PR. 

I'm ready to merge so as to move on.

OK to merge?

src/test/java/spoon/test/ctType/CtTypeTest.java
pvojtechovsky: Github comparer confused us. In the better comparer there is visible that red line 91 is not deleted by is located on green line 105, etc.
So I did not deleted test code ;)

pvojtechovsky:
Yes, merge is OK for me.

PR #1219, opened 2017-03-12
review 1: fix CtExecutableReference#isOverriding on anonymous class

pvojtechovsky:
I think that `CtExecutableReferenceImpl#isOverriding` handles wrong anonymous types.

I think that anonymous classs
```java
public abstract class AbstractTostada implements ITostada {
	@Override
	public ITostada make() {
//following anonymous class extends Tostada (not ITostada!), therefore it overrides Tostada#make and prepare
//---->
		return new Tostada() {
			@Override
			public void prepare() {
				super.prepare();
			}

			@Override
			public ITostada make() {
				return super.make();
			}
		};
	}

	public abstract void prepare();
}
```
The origin code 
```java
	if (getDeclaringType().isAnonymous()) {
		if (!getDeclaringType().getDeclaringType().isSubtypeOf(executable.getDeclaringType())) {
			return false;
		}
	} 
```
is not understandable for me. So I think it is buggy. May be there was some special case which has to be handled for anonymous classes, which @GerardPaligot wanted to handle, but the code is actually wrong. Any idea what was the origin purpose of that code?

### DIFF
--- a/src/main/java/spoon/support/reflect/reference/CtExecutableReferenceImpl.java
+++ b/src/main/java/spoon/support/reflect/reference/CtExecutableReferenceImpl.java
@@ -236,11 +236,7 @@ public class CtExecutableReferenceImpl<T> extends CtReferenceImpl implements CtE
 		if (!isSame) {
 			return false;
 		}
-		if (getDeclaringType().isAnonymous()) {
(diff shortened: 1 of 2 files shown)


monperrus:
My analysis and my IDE agree with you :-)

Thanks.

PR #1228, opened 2017-03-16
review: Remove deprecated elements from release 5.6.0

surli:
Here the summary of deprecated elements removed and changes to do:
  * class ReferenceTypeFilter has been removed: TypeFilter should be used instead
  * class AbstractReferenceFilter has been removed: AbstractFilter should be used
  * classes  ReferenceQueryVisitor and QueryVisitor has been removed: CtQuery#filterChildren should be used
  * interface ReferenceFilter has been removed: Filter should be used instead
  * methods Query#getReferences has been removed: Query#getElements should be used


monperrus:
Those APIs are simple to keep. I propose to keep them in order not to break any client code. 

We would simply remove the depreciation, and keep the documentation link to the preferred method.

PR #1232, opened 2017-03-17
Add 'getBinaryFiles' to CompilationUnit. Fixes #1230.

### LINKED ISSUE #1230: CompilationUnit: Get path of .class file from CompilationUnit
Is it possible to get the file path of the class file of a compilation unit? If not, would you mind if I would add such a feature?

msteinbeck:
This is a proposal to fix #1230. I decided to use `File#listFiles` to detect all available .class files instead of traversing all sub types one after another in order to return existing files only. Did I miss something (except of tests which I will add if you think this method is worth adding to Spoon).

### DIFF
--- a/src/main/java/spoon/reflect/cu/CompilationUnit.java
+++ b/src/main/java/spoon/reflect/cu/CompilationUnit.java
@@ -41,6 +41,13 @@ public interface CompilationUnit extends FactoryAccessor {
 	void setFile(File file);
 
 	/**
+	 * Gets all binary (.class) files that corresponds to this compilation unit
+	 * and have been created by calling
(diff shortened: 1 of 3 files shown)


monperrus:
Thanks for this PR.

What about returning `List<SpoonFile>` instead of `List<File>`?

Could you add a test case?

msteinbeck:
Since `getFile()` returns a `File` as well, I would stick with `File` rather than `SpoonFile` to keep consistency.

monperrus:
You're right. OK for `File`

msteinbeck:
I think I'm done here. Feel free to merge.

PR #1235, opened 2017-03-20
refactor: CtFormalTypeDeclarer extends CtTypeMember

pvojtechovsky:
What about this small change in spoon model?


monperrus:
what's the purpose?

pvojtechovsky:
The code which collects formal type arguments of declaring types is **little** bit simpler if CtFormalTypeDeclarer extends CtTypeMember, because it is so. CtFormalTypeDeclarer already extends CtFormalTypeDeclarer, but it is not declared in model.

But if you have any reason why it is not good idea then close this PR. I have no problem with that.

monperrus:
> CtFormalTypeDeclarer already extends CtFormalTypeDeclarer
what do you mean exactly? :-)

pvojtechovsky:
That all existing implementations of `CtFormalTypeDeclarer` already implements `CtTypeMember`

monperrus:
Then I'm fine to merge this one. If somebody disagrees, drop a comment :-)

msteinbeck:
look good to me

Issue #1233, opened 2017-03-18
How to resolve CtTypeParameter using Actual type arguments?


pvojtechovsky:
I would like to fix #1160 and I came to an idea how to improve handling of actual type arguments of CtTypeReference during some operations, like `CtTypeReference#getSuperclass()`, `CtTypeReference#getSuperInterfaces()`, ...?

There is a [test case](https://github.com/pvojtechovsky/spoon/commit/e331a8e74e12d54457f13c805d7729427fb866e3), which shows how it might behave in future.

The idea is 
> `CtTypeReference#getSuperclass()` called on reference `TRa`, would return the `CtTypeReference` `TRb` to super class with actual type arguments matching the actual type arguments of `TRa`. 

Simple example: 
> The type reference to `ArrayList<String>` would return super interface type reference `List<String>` and not `List<T>`, like it does it now.

Please think about this and give me feedback if such change of Spooon behavior is acceptable. I would like to base further PRs on that too. Or if it is not acceptable, then I suggest to add an new API method, which will behave like this. Please suggest how it should look like then.

monperrus:
The advantages of the current behavior is that:

1. it's the least surprising, close to the code one reads
2. the behavior of `CtType.getSuperClass()` and `CtTypeReference.getSuperClass()` are aligned

What you propose makes sense, probably in a new method `getSuperClassResolved` (or `getSuperClassBounded`) in  `CtTypeReference`.

WDYT?

pvojtechovsky:
> keep current behavior

ok, I agree.

> What you propose makes sense, 

Yes, spoon needs that support, because it is really not trivial operation to detect, which value is assigned to which type parameter.

> probably in a new method getSuperClassResolved (or getSuperClassBounded) in CtTypeReference.

I would say that we need something else. The problem of resolving value of type parameters is not only about super class. Spoon should have a way how to resolve type parameter everywhere ... in superInterfaces, methods, parameters, statements, ... nearly everywhere where is CtType may be a CtTypeParameter, which needs to be resolved.

So the API should solve this use case: Client has a `CtTypeReference` or `CtType` and wants to know what would be the "resolved" `CtTypeReference` or `CtType` for an defined `actual type arguments`.
If the input CtType(Reference) is instance of CtTypeParameter(Reference), then we need to resolve it, otherwise we can return origin CtType(Reference).

But there are some open points in this "design". For example:

**How to define the `actual type arguments`?**
[...]

PR #1237, opened 2017-03-23
Depends on #1238. feature: TypeFactory#createReference(type,includingFormalTypeParams)

pvojtechovsky:
A baby step. The TypeFactory method which creates a CtTypeReference, which has initialized actual type arguments following formal type parameters of the input CtType.

I need it to simplify architecture of CtTypeParameter resolving algorithm. This algorithm can be theoretically applied to
A) CtTypeReference - which has actual type arguments
B) CtType - which has formal type parameters

The new method of this PR let's me simplify it to A) only, because B) can be easily covered to A) by:
```java
CtTypeReference typeRef = fatory.Type().createReference(type, true);
```

### DIFF
--- a/src/main/java/spoon/reflect/factory/TypeFactory.java
+++ b/src/main/java/spoon/reflect/factory/TypeFactory.java
@@ -356,8 +356,7 @@ public class TypeFactory extends SubFactory {
 			ref.addAnnotation(ctAnnotation.clone());
 		}
 		ref.setSimpleName(type.getSimpleName());
-		//TypeParameter reference without parent is unusable. It lost information about it's declarer
-		ref.setParent(type);
+
 		return ref;
 	}
 
--- a/src/main/java/spoon/reflect/factory/TypeFactory.java
+++ b/src/main/java/spoon/reflect/factory/TypeFactory.java
@@ -356,7 +356,7 @@ public class TypeFactory extends SubFactory {
 			ref.addAnnotation(ctAnnotation.clone());
 		}
 		ref.setSimpleName(type.getSimpleName());
-
+		ref.setParent(type);
 		return ref;
 	}
 
--- a/src/main/java/spoon/reflect/factory/TypeFactory.java
+++ b/src/main/java/spoon/reflect/factory/TypeFactory.java
@@ -316,15 +316,29 @@ public class TypeFactory extends SubFactory {
 	 * Create a reference to a simple type
 	 */
 	public <T> CtTypeReference<T> createReference(CtType<T> type) {
+		return createReference(type, false);
+	}
+
+	/**
+	 * @param includingFormalTypeParameter if true then references to formal type parameters
+	 * 	are added as actual type arguments of returned {@link CtTypeReference}
+	 */
+	public <T> CtTypeReference<T> createReference(CtType<T> type, boolean includingFormalTypeParameter) {
 		CtTypeReference<T> ref = factory.Core().createTypeReference();
 
 		if (type.getDeclaringType() != null) {
-			ref.setDeclaringType(createReference(type.getDeclaringType()));
+			ref.setDeclaringType(createReference(type.getDeclaringType(), includingFormalTypeParameter));
 		} else if (type.getPackage() != null) {
 			ref.setPackage(factory.Package().createReference(type.getPackage()));
 		}
 
 		ref.setSimpleName(type.getSimpleName());
+
+		if (includingFormalTypeParameter) {
+			for (CtTypeParameter formalTypeParam : type.getFormalCtTypeParameters()) {
+				ref.addActualTypeArgument(formalTypeParam.getReference());
+			}
+		}
 		return ref;
 	}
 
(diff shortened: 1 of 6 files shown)


monperrus:
it is clearly backward-compatible.

By definition, the actual type arguments would never be bound, they would only by "E" or "T" because they come from the formal definition? Correct?

could you add a test?

pvojtechovsky:
"Houston, we have a problem"... 

The test fails... because there is a problem in a spoon model. I made issue #1238 for that.

monperrus:
OK, let's fix #1238 first then.

monperrus:
here is fix. WDYT?

pvojtechovsky:
I am not sure with these things:
```java
CtType type = ...
CtTypeReference ref = type.getReference();
assertSame(type, ref.getDeclaration());
```
1) This contract (code above) is often used for types. And all the existing code, which depends on this contract will fail if the processed class is a generic class and instead of CtTypeReference there is instance of CtTypeParameterRefrence internally.

2) Let's answer first the question: Q1 "Why anybody needs to put reference to formal type parameter as actual type argument out of the scope of it's visibility?"
A) to build a printable/compilable java model 
B) to build a special model, which is efficient for future analyze of the origin spoon model

B is correct. A makes no sense. Because **every formal type parameter is valid only in scope of it's declarer** and if it is out of that scope then it is invalid. 
```java
class Cat<T> {}
Cat<T> cat; //makes no sense because <T> is out of it's scope.
```
So that operation is not the way how to build a compilable/printable spoon model. It is the way how to create special structures, which are useful for analyzing of spoon model.
If you agree with this answer, then let's answer another question: Q2) Do we want 
A) dynamic lookup following java rules (which we already broke, by removing of parameter out of it's scope) ?
B) or we want to be able to put reference to parameter at any arbitrary place we need and to be always able to found the origin declaration of that parameter?

For me B is correct.

How to continue? I do not need to have this issue solved to be able to continue with other PRs. May be it would be helpful, but I can live without that too. I though it is clear and small BS, so I suggested to do it. But I see that we see that concept different so I suggest to postpone this discussion after somebody will really need it for some real use case. 

Your current implementation, which breaks first contract above, would be even problem for my code. So I prefer to not have it.

pvojtechovsky:
Your code is kind of compromise, which tries to be working for clients whose answer for Q1 and Q2 is both A and B too. But that compromise stops working on code like this:
```java
class Dog<T>{}
class Cat<T> {
  Dog<T> dog;
}
```
If we understand that code as printable/compilable spoon model, then it is clear that `T` of `Dog<T> dog` points to `T` of `Cat`.
But in my use case (an analytic structure needed to [adapt](https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.4.2) parameters to be able to resolve whether method overrides another method) I need to be able to create reference to 'T' which points to `Dog`

monperrus:
Very interesting discussion.

Basically, for type parameters, we have two kinds of declaration. So we need an additional method in
CtTypeParameter.

Two questions.

What should be the name of the new method? I'd suggest getTypeParameterDeclaration (following the
convention of getTypeDeclaration and getExecutableDeclaration)

what should be  the behavior of both?

class Dog<X>{}
class Cat<T> {
  Dog<T> dog;
}

for Dog<T>, I propose that getDeclaration points to T and that getTypeParameterDeclaration points to X.

OK for you?

pvojtechovsky:
I did not got it.

> So we need an additional method in CtTypeParameter.
do you mean CtTypeParameterReference ?

What will this new method return in this context?
```java
class Cat<T>{
T field;
}
```

Concerning dynamic lookup... I agree with you that in normal cases the dynamic lookup is correct. But there is question whether we need to switch off dynamic lookup in some situations and be able to directly point to declaration.

monperrus:
> I did not got it.

See recent commit

> What will this new method return in this context?
> ```java
> class Cat<T>{
> T field;
> }
> ```
T.getDeclaration would return the T of Cat
t.getParameterDeclaration... either null or the same as T.getDeclaration (currently null)

pvojtechovsky:
Now I noticed your commit. I understand it now. It is an interesting solution!
But you are solving different problem then me :-) ... there are so many ways how to use and understand spoon.
I will explain in #1243 how I understood adapting of type parameters and then it will be easier to continue discussion here. 
Thank you for your time!

monperrus:
I'm just trying to merge this PR which is about "TypeFactory#createReference(type,includingFormalTypeParams)", with an appropriate test.

To me it's OK like this. It's probably not the perfect solution but it's a working one and we can improve it in the future in the context of another PR.

WDYT?

src/main/java/spoon/reflect/factory/TypeFactory.java
```diff
@@ -342,8 +356,7 @@ public CtTypeParameterReference createReference(CtTypeParameter type) {
 			ref.addAnnotation(ctAnnotation.clone());
 		}
 		ref.setSimpleName(type.getSimpleName());
-		//TypeParameter reference without parent is unusable. It lost information about it's declarer
-		ref.setParent(type);
+
```
pvojtechovsky: This change is problem for my code. Please rollback it.

pvojtechovsky:
Yes this solution is ok, just do not create reference without parent.  It breaks basic contract of usage uf type and typeReference.

pvojtechovsky:
Thank You. It is OK for me now.

monperrus:
Good, I'll merge tomorrow and close #1238 and #1233.

pvojtechovsky:
OK, thanks!

PR #1242, opened 2017-03-28
review1: feature SuperInheritanceHierarchyFunction supports references and listener

pvojtechovsky:
refactoring of SuperInheritanceHierarchyFunction.
* can be configured if it returns CtTypeReferences (new implementation) or CtType (old implementation)
* can register CtScannerListener


monperrus:
OK for me. Could you add a test for the added capabilities? Thanks.

pvojtechovsky:
Tests were added.

Issue #1238, opened 2017-03-25
CtTypeParameterReference cannot be used as actual type argument


pvojtechovsky:
Problem: CtTypeParameterReference cannot be used as actual type argument.

Why? Because CtTypeParameterReference uses `parent` to resolve it's declaration. 
But the `parent` is changed by `CtTypeReference#addActualTypeArgument(CtTypeReference)`, so after it get's assigned it lost's it's declarer.

See test case, which can be found in PR #1237
```java
	public void testTypeParameterReferenceAsActualTypeArgument() throws Exception {
		CtType<Tacos> aTacos = buildNoClasspath(Tacos.class).Type().get(Tacos.class);
		
		CtTypeReference<?> typeRef = aTacos.getReference();
		CtTypeParameter typeParam = aTacos.getFormalCtTypeParameters().get(0);

		CtTypeParameterReference typeParamRef = typeParam.getReference();
		//typeParamRef is able to resolve it's typeParam - OK
		assertSame(typeParam, typeParamRef.getDeclaration());
		//the reference to parent is used by legacy 
		//CtTypeParameterReferenceImpl#getRecursiveDeclaration 
		//to detect declaration - it is probably wrong approach
		assertSame(typeParam, typeParamRef.getParent());
		
		//this assignment changes parent of typeParamRef to typeRef
		typeRef.addActualTypeArgument(typeParamRef);
		//stored typeParamRef is same like the added one, no clone - OK
		assertSame(typeParamRef, typeRef.getActualTypeArguments().get(0));
		//typeParamRef has got new parent 
		assertSame(typeRef, typeParamRef.getParent());

		//and therefore it lost the link to declaring CtFormalTypeDeclarer, 
		//so this assertion FAILS
		assertSame(typeParam, typeParamRef.getDeclaration());
	}
```
Solution?
I think `CtTypeParameterReference` should use `CtTypeReferenceImpl#declaringType` to reference the `CtFormalTypeDeclarer`, which declares it.
The problem is that type of `CtTypeReferenceImpl#declaringType` is `CtTypeReference` and not `CtFormalTypeDeclarer`.
So that solution would need some changes ... like:
* rename CtTypeReferenceImpl#declaringType to `declarer` and change the type to `CtFormalTypeDeclarer`
* add new method `CtType(Parameter?)Reference#getDeclarer()` with return value `CtFormalTypeDeclarer`
[...]

PR #1243, opened 2017-03-28
WiP: feature: CtSuperTypeHierarchy

pvojtechovsky:
`CtSuperTypeHierarchy` is a new helper class, which provides access to actual type arguments of any super type of type X adapted to type X

### DIFF
--- a/src/main/java/spoon/reflect/visitor/filter/CtSuperTypeHierarchy.java
+++ b/src/main/java/spoon/reflect/visitor/filter/CtSuperTypeHierarchy.java
@@ -1,0 +1,478 @@
+/**
+ * Copyright (C) 2006-2017 INRIA and contributors
+ * Spoon - http://spoon.gforge.inria.fr/
+ *
+ * This software is governed by the CeCILL-C License under French law and
+ * abiding by the rules of distribution of free software. You can use, modify
+ * and/or redistribute the software under the terms of the CeCILL-C license as
+ * circulated by CEA, CNRS and INRIA at http://www.cecill.info.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the CeCILL-C License for more details.
+ *
+ * The fact that you are presently reading this means that you have had
+ * knowledge of the CeCILL-C license and that you accept its terms.
+ */
+package spoon.reflect.visitor.filter;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import spoon.SpoonException;
+import spoon.reflect.declaration.CtClass;
+import spoon.reflect.declaration.CtElement;
+import spoon.reflect.declaration.CtFormalTypeDeclarer;
+import spoon.reflect.declaration.CtType;
+import spoon.reflect.declaration.CtTypeInformation;
+import spoon.reflect.declaration.CtTypeParameter;
+import spoon.reflect.declaration.ModifierKind;
+import spoon.reflect.reference.CtTypeParameterReference;
+import spoon.reflect.reference.CtTypeReference;
+import spoon.reflect.reference.CtWildcardReference;
+import spoon.reflect.visitor.chain.CtConsumer;
+import spoon.reflect.visitor.chain.ScanningMode;
+
+/**
+ * Helper class created from type X or reference to X.
+ * It provides access to actual type arguments
+ * of any super type of type X adapted to type X.<br>
+ * Example:<br>
+ * <pre>
+ * //reference to `ArrayList` with actual type argument `Integer`
+ * CtTypeReference arrayListRef = ... //ArrayList&lt;Integer&gt;
+ * //type java.util.List with type parameter `E`
+ * CtType list = ... //List&lt;E&gt;
+ * //adapting of type parameter `E` to scope of arrayListRef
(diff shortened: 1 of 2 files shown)


monperrus:
The core responsibility of this new class is `adaptType`?

pvojtechovsky:
yes. The core is "to be able to compute details about super inheritence hierarchy", which is needed to deliver these features:
* isSubtypeOf
* adaptType

monperrus:
OK. We'll have to discuss on how to integrate this kind of feature in the repo, but first, I want to well understand what you mean by `adaptType`. 

Could you elaborate on the example given in the Javadoc?

pvojtechovsky:
>  I want to well understand what you mean by adaptType
> Could you elaborate on the example 

The target of all these small PRs is to be able to correctly detect whether one method overrides another method. So please have a look at this [test](https://github.com/INRIA/spoon/pull/1243/commits/bbe90001d68bbd080e90372b6548f3a458947e69#diff-c20710bb8ee877d0eae2d22578a8f010R603). There is a complex class hierarchy of generic classes and generic inner classes. There are 3 implementations of method `eatMe`, which are using different generic types, but which are - after [adapting](https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.4.2) - same

pvojtechovsky:
> I want to well understand what you mean by adaptType.

here is the big picture: I want to implement a refactoring, which will remove method parameters. That needs an algorithm which founds all declarations and implementations of methods with same signature on the same type hierarchy. Therefore I need to correctly detect whether two methods have same signature (method A overrides B). And following java lang specification, one of the steps in the method signature comparation is [adapting](https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.4.2) of type parameters.
Example:
```java
class Parent<T,K> {
  void someMethod(T p1, K p2);
}
class Child1 extends Parent<Integer,Double> {
  void someMethod(Integer p1, Double p2);
}
```
Now I want to detect whether `Child1#someMethod` overrides `Parent<T,K>#someMethod(T,K)`. The idea is like this:
1) to take class Child1 and adapt all type parameters of all it's parents to scope of `Child1`. After adapting is done we get following VIRTUAL (not real spoon) model, which would be represented by sources like this:
```java
class Parent {
  void someMethod(Integer p1, Double p2);
}
class Child1 extends Parent {
  void someMethod(Integer p1, Double p2);
}
```
In this model it is easily possible to compare method signatures, because type parameters are adapted to scope of Child1. 

The class `CtSuperTypeHierarchy` holds that VIRTUAL model and is able to adapt parameters to scope of Child1. So now it should be more understandable the purpose of this PR.

Now little bit more complicated example to explain why I started with PR #1238.

```java
class Parent<T,K> {
  void someMethod(T p1, K p2);
}
class Child2<M> extends Parent<Double,M> {
  void someMethod(Double p1, M p2);
}
```
In this case the super type hierarchy model of Child2, where each parameter is adapted looks like this:
```java
class Parent {
  void someMethod(Double p1, M p2);
}
class Child2<M> extends Parent {
  void someMethod(Double p1, M p2);
}
```
Note the usage of `M` in second parameter of `Parent#someMehod`. 
In such model the `type` of second parameter of `Parent#someMehod` is `CtTypeParameterReference`, which refers formal type parameter `M` of  class `Child2`.
And that's the problem of #1238. That `CtTypeParameterReference` is actually not able to **statically** reference it's declarer from **different scope**.
From java point of view that model is nonsense! And yes, such reference is really a special case. And yes, current implementation of `CtSuperTypeHierarchy` DOES NOT NEED that. So we can close #1238 unsolved if you think it is on bad way. 
But **may be** if such reference would work then it might be helpful in similar analyzes.

monperrus:
we're getting closer to merge.
- according to the main responsibility, what about renaming class `CtSuperTypeHierarchy` to `GenericTypeAdapterImpl`?
- for discoverability, what about adding a static method `adaptType` in `Query`?

pvojtechovsky:
> according to the main responsibility, what about renaming class CtSuperTypeHierarchy to GenericTypeAdapterImpl

yes, I am open to better name. But let's wait some more days with merge and name decision until I finish the "method overrides" detection code - it is the main client of CtSuperTypeHierarchy for now and I would prefer to have free hands with further refactoring of CtSuperTypeHierarchy , when something might be done better.

Note: CtSuperTypeHierarchy is able to adapt generic types declared in scope of CtType. I have nearly finished another similar helper class, which will adapt generic types declared in scope of CtMethod. I call it actually `CtMethodSuperTypeHierarchy`. These two helper classes has similar API, so we can think of some interface, which will be implemented by both of them. But this generic type adaption algorithms are quite complicated in java, so I need more time to do it well.

> for discoverability, what about adding a static method adaptType in Query?

At the beginning I though that CtSuperTypeHierarchy is a mapping function, but now I see it is not. So it should be may be moved to different package and does not belong to Query. I guess some method(s) might be added to CtType(Parameter)(Reference).

Explanation: Mapping function has one responsibility, but CtSuperTypeHierarchy has more. There are:
* CtTypeReference adaptType(type) - adapts `type` into scope of CtSuperTypeHierarchy or CtMethodSuperTypeHierarchy
* boolean isSubtypeOf(typeRef) - detects if typeRef is super type of scope of CtSuperTypeHierarchy
* boolean isSame(typeParam1, typeParam2)  - detect whether 2 type parameters are same - this method is WiP...
* boolean isSameSignature(CtMethod, boolean canTypeErasure) - detects if method has same signature (after adapting parameters, etc.) like scope method of CtMethodSuperTypeHierarchy

monperrus:
I guess this one can be closed.

pvojtechovsky:
Yes, it's content was already merged by PR #1218

PR #1248, opened 2017-03-31
refactor: remove duplicate code

monperrus:


monperrus:
closing this one. One method is about the modifier encoding by JDT the other about the modifier encoding by the JVM. It happens to be the same but it's not bound to remain the same.

PR #1249, opened 2017-03-31
adds example of failing to set type to CtTypedElement.

anivanovic:
@surli this is the PR showing compile error on CtTypedElement commented on
[StackOverflow](http://stackoverflow.com/questions/43078951/how-to-change-type-of-cttypedelement)

### DIFF
--- a/src/test/java/spoon/test/ctTypedElement/testclass/CtTypedElementError.java
+++ b/src/test/java/spoon/test/ctTypedElement/testclass/CtTypedElementError.java
@@ -1,0 +1,13 @@
+package spoon.test.ctTypedElement.testclass;
+
+import spoon.reflect.declaration.CtTypedElement;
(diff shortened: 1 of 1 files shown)


surli:
Ok I'm not sure to understand the problem here: the compilation error is normal, the two `<?>` are considered as different types by Java. 

If you use the following code:
```java
private static <T> void changeType(CtTypedElement<T> typedElem) {
	CtTypeReference<T> type = typedElem.getType();
	typedElem.setType(type);
}
```
it will work fine. 
Even easier, you could just remove the type check on CtTypeReference here: 
```java
CtTypeReference type = typedElem.getType();
```

So, there is no spoon-related bug in the snippet you show.

surli:
I'm closing this one. Don't hesitate to continue the discussion on an issue or stackoverflow.

Issue #1246, opened 2017-03-31
Compilation bug, spoon silently doesn't compile any source code in environment where new File(".") points to the folder without any java sources (for example in tomcat environment where it points to the bin folder)


gnom7:
Hello, I have a project that uses spoon 5.3.0. but as for now due to missing org.eclipse.jdt.core/3.12.0.v20160516-2131 (on which spoon relies) in maven repo (and because of that problem some builds on my CI have been failed) I consider switching to spoon 5.6.0, but it seems that I have faced a bug -- when I tested my code in IDE everything was fine, but the same code didn't work in tomcat environment - spoon silently didn't compiled any source code.

Debugging:
When running compile() on SpoonModelBuilder it calls batchCompiler.compile(args); so those args doesn't contain input sources, but "."

```
final String[] args = new JDTBuilderImpl() //
				.classpathOptions(new ClasspathOptions().encoding(this.encoding).classpath(getSourceClasspath()).binaries(getBinaryOutputDirectory())) //
				.complianceOptions(new ComplianceOptions().compliance(javaCompliance)) //
				.annotationProcessingOptions(new AnnotationProcessingOptions().compileProcessors()) //
				.advancedOptions(new AdvancedOptions().preserveUnusedVars().continueExecution().enableJavadoc()) //
				.sources(new SourceOptions().sources()) // no sources, handled by the JDTBatchCompiler
				.build();

		getFactory().getEnvironment().debugMessage("compile args: " + Arrays.toString(args));
		System.setProperty("jdt.compiler.useSingleThread", "true");
		batchCompiler.compile(args);
```
then in org.eclipse.jdt.internal.compiler.batch.Main it creates `new File(".")` from that argument and calls org.eclipse.jdt.internal.compiler.batch.FileFinder which recursively searchs for files with `SuffixConstants.SUFFIX_STRING_JAVA` (".java") extension, so when running in tomcat `new File(".")` points to tomcat bin folder wich does not contain any .java files (and running in IDE in sandbox project it points to project base dir -- so it contains .java files), then in Main class filesCount == 0 is true (there is no .java files in tomcat bin folder) and compilation termiates even not being started and in IDE filesCount != 0 (project base dir contains .java files) and compilation proceeds. I haven't debugged further more, and I don't know how then compiled files appeared from VALID sources (not from the project base)...
when filesCount == 0, the code below in Main class switches proceed flag to false and eclipse compiler dies.
```
[...]

PR #1251, opened 2017-04-03
review: consider the sources when calling compile in JDTBasedSpoonCompiler

surli:
This PR is really similar to https://github.com/INRIA/spoon/pull/1208 and is proposed to fix https://github.com/INRIA/spoon/issues/1246


surli:
@monperrus just tested on an empty dir and it fixes the bug.

PR #1254, opened 2017-04-03
refactor: enforce the architectural rule 'never mix stateful objects and stateless helper classes containing static methods'

monperrus:
I had to move some recent methods from RtMethod (minor non-backward compatible change).


monperrus:
I make smaller PRs

PR #1255, opened 2017-04-04
Fixed declaration of non-imported method calls for noclasspath.

msteinbeck:


msteinbeck:
I found another bug not being covered by this solution.

PR #1266, opened 2017-04-20
WiP: fix bug related to CtIf and hashCodes

surli:


surli:
> To follow our discussion, tests failed right?

Indeed, follow the PR ;)

monperrus:
This is a collision. However, it's not a contract that hashcodes are always different. The hashcode implementation is a trade-off between speed and collision risk.

surli:
> This is a collision. However, it's not a contract that hashcodes are always different. The hashcode implementation is a trade-off between speed and collision risk.

Yes, but the contract equals/hashcode should be respected and here it's not the case.

PR #1273, opened 2017-04-26
fix actual type arguments resolving

pvojtechovsky:
Fixes #1218


monperrus:
awesome! could you add a test case?

pvojtechovsky:
I will probably have time tomorrow evening for that.
The problem was that `ClassTypingContext` of `ForwardingNavigableMap$StandardDescendingMap` was asked to adapt parameter `<K>` of superclass `Maps.DescendingMap<K, V>` of `StandardDescendingMap`.
There was error in algorithm that typed parameters was searched only in current type hierarchy and was not search in type hierarchy of enclosing (declaring) class. The test case should be something like
```java
class A<T> {
 class B extends List<T> {}
}
```
where 
```java
new ClassTypingContext(classB).adaptType(classB.getSuperClass().getActualTypeArguments().get(0))
```
failed before, but should return `<T>` of class `A`.

pvojtechovsky:
Test added

monperrus:
thanks a lot Pavel.

PR #1275, opened 2017-04-27
review1: fix generic type adapting

pvojtechovsky:
Here are fixed other bugs found during implementing of CtMethod#isOverriding(CtMethod).
Following problems are tested and fixed by this PR:
* The generic types are now adapted recursive into actual type arguments. E.g. `List<List<T>>` is adapted to `List<List<M>>`
* The generic types are now adapted recursive into their bounds. E.g. `List<? extendsT>` is adapted to `List<? extends M>`
* The adapting of wildcard type E.g. `List<? extendsT>` does not fail with NullPointerException
* The MethodTypingContext correctly detects isOverriding for child method whose generic types are replaced by concrete types.

### DIFF
--- a/src/main/java/spoon/support/visitor/AbstractTypingContext.java
+++ b/src/main/java/spoon/support/visitor/AbstractTypingContext.java
@@ -84,5 +84,14 @@ abstract class AbstractTypingContext implements GenericTypeAdapter {
 		return typeParamRefAdapted;
 	}
 
+	/**
+	 * adapts `typeParam` to the {@link CtTypeReference}
+	 * of scope of this {@link GenericTypeAdapter}
+	 * In can be {@link CtTypeParameterReference} again - depending actual type arguments of this {@link GenericTypeAdapter}.
+	 *
+	 * @param typeParam to be resolved {@link CtTypeParameter}
+	 * @return {@link CtTypeReference} or {@link CtTypeParameterReference} adapted to scope of this {@link GenericTypeAdapter}
+	 *  or null if `typeParam` cannot be adapted to target `scope`
+	 */
 	protected abstract CtTypeReference<?> adaptTypeParameter(CtTypeParameter typeParam);
 }
--- a/src/main/java/spoon/support/visitor/AbstractTypingContext.java
+++ b/src/main/java/spoon/support/visitor/AbstractTypingContext.java
@@ -1,0 +1,88 @@
+/**
+ * Copyright (C) 2006-2017 INRIA and contributors
+ * Spoon - http://spoon.gforge.inria.fr/
+ *
+ * This software is governed by the CeCILL-C License under French law and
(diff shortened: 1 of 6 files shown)


src/main/java/spoon/support/visitor/AbstractTypingContext.java
```diff
@@ -0,0 +1,88 @@
...
+import spoon.reflect.declaration.CtTypeInformation;
+import spoon.reflect.declaration.CtTypeParameter;
+import spoon.reflect.reference.CtTypeParameterReference;
+import spoon.reflect.reference.CtTypeReference;
+import spoon.reflect.reference.CtWildcardReference;
+
+/**
+ * Implements common adapting algorithm of {@link ClassTypingContext} and {@link MethodTypingContext}
+ */
+abstract class AbstractTypingContext implements GenericTypeAdapter {
```
monperrus: I like the package visibility

src/main/java/spoon/support/visitor/AbstractTypingContext.java
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	private CtTypeReference<?> adaptTypeParameterReferenceBoundingType(CtTypeParameterReference typeParamRef, CtTypeReference<?> boundingType) {
+		CtTypeParameterReference typeParamRefAdapted = typeParamRef.clone();
+		typeParamRefAdapted.setParent(typeParamRef.getParent());
+		typeParamRefAdapted.setBoundingType(boundingType == null ? null : adaptType(boundingType));
+		return typeParamRefAdapted;
+	}
+
+	protected abstract CtTypeReference<?> adaptTypeParameter(CtTypeParameter typeParam);
```
monperrus: what's the contract of implementors?

pvojtechovsky: There is not expected that any client will implement this method, because all possible implementations ClassTypingContext and MethodTypingContext are already done.
I have added this comment
```java
	/**
	 * adapts `typeParam` to the {@link CtTypeReference}
	 * of scope of this {@link GenericTypeAdapter}
	 * In can be {@link CtTypeParameterReference} again - depending actual type arguments of this {@link GenericTypeAdapter}.
	 *
	 * @param typeParam to be resolved {@link CtTypeParameter}
	 * @return {@link CtTypeReference} or {@link CtTypeParameterReference} adapted to scope of this {@link GenericTypeAdapter}
	 *  or null if `typeParam` cannot be adapted to target `scope`
	 */
```

src/main/java/spoon/support/visitor/AbstractTypingContext.java
pvojtechovsky: yes, it was intentional. This class is internal

monperrus:
thanks Pavel

PR #1276, opened 2017-05-01
review1: refactor Method overriding/overriden filters

pvojtechovsky:
New MethodTypingContext is used to implement these two filters.

I added also method `#includingSelf(boolean)`, which can be used to configure the behavior of these filters.


monperrus:
rebase?

pvojtechovsky:
Is it problem of my PR or problem of build process?

monperrus:
thanks

PR #1277, opened 2017-05-02
fix failing CtExecutableReference#getActualMethod

### LINKED ISSUE #1256: [generic type] CtMethod.getReference().getActualMethod() fails for generic parameter types
```java
//model
interface IFace<T extends Long> {
 void method(T param);
}
//test
CtMethod method = ...//refers `void method(T param)`
method.getReference().getActualMethod(); //returns null
```
method.getReference().getActualMethod() returns null, because parameter type reference `T` is not able to found it's declaration in `IFace`, because CtExecutableReference parent is null - there is no way how to get to declaration of generic type dynamically.

The workaround is:
```java
CtExecutableReference ref = method.getReference();
ref.setParent(method);
ref.getActualMethod(); //returns expected `Method`
```
I have seen this problem [here](https://github.com/INRIA/spoon/pull/1218/commits/ffab37263d416be0476cd26ebb23c07a6724deb4#diff-89d21fcd8875342a0be9725d0bb210f9R54)

pvojtechovsky:
should fix #1256, but I am not sure that it is correct solution. May be we should think about having reference to Formal Type declarer directly in `CtTypePareameterReference` (which would be added by compiler?) instead of searching for the declarer in all possible contexts where it might occur. How many place we need to fix next to this one?

### DIFF
--- a/src/main/java/spoon/support/reflect/reference/CtTypeParameterReferenceImpl.java
+++ b/src/main/java/spoon/support/reflect/reference/CtTypeParameterReferenceImpl.java
@@ -19,9 +19,11 @@ package spoon.support.reflect.reference;
 import spoon.SpoonException;
 import spoon.reflect.declaration.CtElement;
 import spoon.reflect.declaration.CtFormalTypeDeclarer;
+import spoon.reflect.declaration.CtMethod;
 import spoon.reflect.declaration.CtType;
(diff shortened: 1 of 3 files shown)


monperrus:
Thanks!

> How many place we need to fix next to this one?

We'll see :-)

PR #1220, opened 2017-03-12
review1: feature: CtMethod#isOverriding(CtMethod)

pvojtechovsky:
New API `CtMethod#isOverriding(CtMethod)`, which works better on generic methods, then legacy `CtExecutableReference#isOverriding(CtExecutableReference)`

This PR actually fails on wrong legacy implementation of `CtExecutableReference#isOverriding(CtExecutableReference)`

I would like to discuss with you: "How to fix that legacy implementation?"

I did not understood:
1) why isOverriding method was implemented on CtExecutableReference and not on CtMethod since beginning?
2) how Template engine works with `CtExecutableReference#isOverriding` ... I do not know how to replace that code with `CtMethod#isOverriding` ... because I do not understand that code yet...

PR #1281, opened 2017-05-03
Fix SuperInheritanceHierarchyFunction in noClassPath mode

pvojtechovsky:
This PR improves algorithm of SuperInheritanceHierarchyFunction by way it works better in no class path mode. 
This fix is needed to pass tests of #1220

### DIFF
--- a/src/main/java/spoon/reflect/visitor/filter/SuperInheritanceHierarchyFunction.java
+++ b/src/main/java/spoon/reflect/visitor/filter/SuperInheritanceHierarchyFunction.java
@@ -57,6 +57,47 @@ public class SuperInheritanceHierarchyFunction implements CtConsumableFunction<C
 	private boolean returnTypeReferences = false;
 
 	/**
+	 * Super inheritance hierarchy scanning listener.
+	 * Use it instead of {@link CtScannerListener}
+	 * if you need to know whether visited type reference is class or interface
+	 */
+	public static class Listener implements CtScannerListener {
+
+		/**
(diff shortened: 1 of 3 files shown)


src/main/java/spoon/reflect/visitor/filter/SuperInheritanceHierarchyFunction.java
```diff
@@ -57,6 +57,47 @@
 	private boolean returnTypeReferences = false;
 
 	/**
+	 * Super inheritance hierarchy scanning listener.
+	 * Use it instead of {@link CtScannerListener}
+	 * if you need to know whether visited type reference is class or interface
+	 */
+	public static class Listener implements CtScannerListener {
```
monperrus: can we make this class private or package protected?

pvojtechovsky: It is better `CtScannerListener`. so it might be interesting for some spoon clients. 
It is true that they might extend `DistinctTypeListener` too, but it provides some extra behavior, which might be NOT needed by some clients.
But technically I can make it private, so we keep Spoon API as small as possible.
So should I really make it private?

monperrus:
> But technically I can make it private, so we keep Spoon API as small as possible.
> So should I really make it private
Yes, let's keep the API surface small.

If somebody finds it and needs it, we'd be happy to make it visible.

pvojtechovsky:
It is private now

PR #1286, opened 2017-05-06
reproduce CtMethod.getReference().getActualMethod() failure

pvojtechovsky:
here is a failing test, which reproduces problem of #1256


pvojtechovsky:
The `((CtExecutableReference) parent).getDeclaration()` you have used cannot be used, because it returns wrong executable. That problem is well visible in new test `testGetGenericExecutableReference`.

Note: that older test `testGetGenericMethodFromReference` pass with Your fix, but that test is not checking that returned Method is the correct one. It simply returns the first method from these two
```
class Cloud<T> {
	<U extends InputStream> void method(T param, U param2) {}
	<U extends Reader> void method(T param, U param2) {}
```
so if you would ask for CtMethod of these two invocations
```java
		cc.method("x", (InputStream)null);
		cc.method("x", (Reader)null);
```
spoon would always return `<U extends InputStream> void method(T param, U param2) {}` ... and it is wrong in case of second invocation.

pvojtechovsky:
@surli Thank You for that interesting solution!
If I understood it well the solution is based on
C1) CtParameterTypeReference#superType now contains the information from the CtTypeParameter#superClass
C2) that extra information is used in algorithm of detection of correct declaring method.

Seems like it works, and I was not able to found a model where it does not work.
I have not clear opinion about correctness of this solution. That problem is still too complex for me or my understanding of some spoon concepts is not good enough - so I personally agree with that solution because test passed.

I have extended test little bit and there is still a little problem. May be even the contract I have added is not correct, so you can rollback that test if you think so.

surli:
> I have extended test little bit and there is still a little problem. May be even the contract I have added is not correct, so you can rollback that test if you think so.

I will check that, I want to close this PR before the release :)

src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
```diff
@@ -676,12 +686,38 @@ private String concatSubArray(String[] a, int endIndex) {
...
+				// if the type parameter has a super interface, then we'll get it too, as a superclass
+				// type parameter can only extends an interface or a class, so we don't make the distinction
+				// in Spoon. Moreover we can only have one extends in a type parameter.
+				} else if (superInterfaces != null && superInterfaces.length == 1) {
+					refSuperClass = this.getTypeReference(superInterfaces[0]);
+				}
+
+				if (refSuperClass != null) {
+					CtTypeParameter typeParameter = this.jdtTreeBuilder.getFactory().createTypeParameter();
+					typeParameter.setSuperclass(refSuperClass);
```
surli: Try to change to `createTypeParameterReference` and to use `addBound`

src/test/java/spoon/test/methodreference/MethodReferenceTest.java
```diff
@@ -225,6 +227,32 @@ public void testGetGenericMethodFromReferene() throws Exception {
 		assertEquals("method", method2.getName());
 	}
 	
+	@Test
+	public void testGetGenericExecutableReference() throws Exception {
+		CtType<?> classCloud = ModelUtils.buildClass(Cloud.class);
+		for (CtMethod<?> ctMethod : classCloud.getMethodsByName("method")) {
```
surli: Inline the test to avoid for loop inside a test

src/test/java/spoon/test/reference/ExecutableReferenceGenericTest.java
```diff
@@ -110,7 +114,11 @@ public void testMultiReferenceBetweenMethodsWithGenericInSameClass() throws Exce
 		CtMethod<?> expectedMethod5 = getCtMethodByNameFromCtClass(clazz, "method5");
 
 		assertEquals(3, refsMethod2.size());
-		assertEquals(expectedMethod1, refsMethod2.get(0).getDeclaration());
+		// Not true anymore, because T is inferred at execution time
+//		assertEquals(expectedMethod1, refsMethod2.get(0).getDeclaration());
+		CtExecutable execRefsMethods2 = refsMethod2.get(0).getDeclaration();
+
+		assertThat(execRefsMethods2.getSignature(), is("<T extends java.lang.String> void method1(T extends java.lang.String)"));
```
surli: Rewrite with assertEquals

src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
```diff
@@ -99,6 +101,9 @@
 
 public class ReferenceBuilder {
 
+	// Allow to detect circular references and to avoid endless recursivity
+	// when resolving parameterizedTypes (e.g. Enum<E extends Enum<E>>)
+	private Set<TypeBinding> exploringParameterizedBindings = new HashSet<>();
```
pvojtechovsky: I have heard that spoon elements does not work well in HashSet/Map, because their implementation of #hashCode and #equals doesn't fit to HashSet contract. I do not know details. I am using IdentityHashMap or HashSet<String> on unique identificator of the elements. Please discuss the details with @monperrus if needed.

monperrus:
@surli ready for merge before release?

src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
pvojtechovsky: I do not have counter-example. I am happy that I was wrong here ;-)

surli:
@monperrus yes it's ok for me

PR #1288, opened 2017-05-07
review: feature: MethodTypingContext#adaptMethod

pvojtechovsky:
I am implementing the algorithm, which searches for all methods, which has to be modified when changing method signature (e.g. remove method parameter). This algorithm needs to be able to search in all super types of sub types of declaring type of target methods, for methods with equal `sub signature`... yes, it is tricky sentence ... so see Example:
```java
class A {
 void method() {}
}
interface IB {
 void method();
}
class B extends A implements IB {
//note that B does not declares `method`. It inherits it from `A`.
}
class C implements IB {
 void method();
}
```
If `A#method()` has to be refactored, then that algorithm has to found `C#method()` too. But note that `C#method()`, does not overrides `A#method()`, so it is not enough to use existing overriding filter.
The algorithm first needs to detect that `IB#method()` is overridden by `B#method()` (which does not exists, but is inherited from `A`). 
We can do that by new feature implemented in this PR like this:
```java
MethodTypingContext mtc = new MethodTypingContext()
   .setClassTypingContext(new ClassTypingContext(classB));
//create a clone of `A#method` whose generic types are adapted to scope of class `B`
CtMethod classB_method = mtc.adaptMethod(classA_method);
mtc.setMethod(classB_method);
assertTrue(mtc.isOverriding(ifaceB_method));
```
or shortly like this:
```java
MethodTypingContext mtc = new MethodTypingContext()
   .setClassTypingContext(new ClassTypingContext(classB))
   .setMethod(classA_method); //it detects that method is declared in different scope and adapts it automatically
assertTrue(mtc.isOverriding(ifaceB_method));
```
The example above is simplified. There are no generic parameters in method, so it might be solved different way. But with generic parameters and longer type hierarchies, the adapting of method parameters is necessary, because `A#method` and `C#method` might have different parameter types, but they can be still related, thanks to some generic method(s) in between ...

Note: review only second commit. The first commit comes from another PR. I will rebase after first commit is merged.

### DIFF
--- a/src/test/java/spoon/test/generics/GenericsTest.java
+++ b/src/test/java/spoon/test/generics/GenericsTest.java
@@ -921,9 +921,7 @@ public class GenericsTest {
 		methodSTH.setMethod(trLunch_eatMe);
 		//contract: method typing context creates adapted method automatically, which is equal to manually adapted one
 		assertEquals(adaptedLunchEatMe, methodSTH.getAdaptationScope());
-
-		//contract: check that adapted method in role of method typing context can be used
-		//contract: check that adapted method in role of method typing context can be used
+		
 		// represents <C> void eatMe(M paramA, K paramB, C paramC)
 		CtMethod<?> trWeddingLunch_eatMe = ctClassWeddingLunch.filterChildren(new NameFilter<>("eatMe")).first();
 		assertTrue(methodSTH.isOverriding(trLunch_eatMe));
--- a/src/test/java/spoon/test/generics/GenericsTest.java
+++ b/src/test/java/spoon/test/generics/GenericsTest.java
@@ -56,6 +56,7 @@ import java.io.File;
 import java.util.ArrayList;
 import java.util.List;
 
+import static javafx.scene.input.KeyCode.M;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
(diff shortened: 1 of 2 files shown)


src/test/java/spoon/test/generics/GenericsTest.java
```diff
@@ -863,6 +875,68 @@ public void testMethodTypingContext() throws Exception {
...
 	
 	@Test
+	public void testMethodTypingContextAdaptMethod() throws Exception {
+		Factory factory = build(new File("src/test/java/spoon/test/generics/testclasses"));
+		CtClass<?> ctClassLunch = factory.Class().get(Lunch.class);
+		CtMethod<?> trLunch_eatMe = ctClassLunch.filterChildren(new NameFilter<>("eatMe")).first();
+		CtClass<?> ctClassWeddingLunch = factory.Class().get(WeddingLunch.class);
+		CtMethod<?> trWeddingLunch_eatMe = ctClassWeddingLunch.filterChildren(new NameFilter<>("eatMe")).first();
+
+		MethodTypingContext methodSTH = new MethodTypingContext().setClassTypingContext(new ClassTypingContext(ctClassWeddingLunch));
```
monperrus: what does this line mean? I thought a MethodTypingContext would be always with respect to a specific method

pvojtechovsky: yes, it is correct. But there are situations when MethodTypingContext is needed for method, which does not exist. 
In the example above (first comment of this PR), it is the case when I need MethodTypingContext for  method `void method()' in adaption scope of class `B`

monperrus:
I've updated the tests. The core contract and test of adaptMethod is fine to me.

After analysis, I see one major and one minor problem here. 

The major one is that the design and usage of  `isOverriding(CtMethod<?> thatMethod)` is hard to grasp: it seems naturally impossible that a method overrides a context (or the other way around). Do we change this here on in a subsequent PR?

monperrus:
> The major one is that the design and usage of isOverriding(CtMethod<?> thatMethod) is hard to grasp: it seems naturally impossible that a method overrides a context (or the other way around). Do we change this here on in a subsequent PR?

See  #1292

PR #1291, opened 2017-05-08
review: feature: all executables same signature function

pvojtechovsky:
Whenever anybody needs to refactor method signature, it is necessary to found all methods and lambdas in sub and super class/interface hierarchies, which overrides/are overridden by the refactored method.
Note: it is not primitive operation, because it needs to correctly detect generic methods and methods in related inheritance hierarchies.
Example:
```java
class A {
 void toBeRefactored(InputStream is) {}
}

class IFace<T> {
 void toBeRefactored(T param) {}
}

class B extends A implements IFaceC {}

class D implements IFaceC<Writer> {
 void toBeRefactored(Writer w) {}
}
class E {
 void toBeRefactored(InputStream is) {}
}
```
Note that there are 2 inheritance hierarchies with roots:
1) A, with sub type B
2) IFaceC, with sub types B and D
But they are related because they share type B. If one needs to refactor method signature of A#toBeRefactored, then s/he must refactor IFace#toBeRefactored and D#toBeRefactored too.
Note that E#toBeRefactored is not touched, because it is not member of any related inheritance hierarchy.

The mapping function `AllMethodsSameSignatureFunction` of this PR, searches for all CtMethods in all related inheritance hiearchies, which are overriding/overriden by input CtMethod.

The algorithm of this mapping function has to recursively scan sub and super inheritance hierarchies of each found sub/super type. It uses `SubInheritanceHiearchyResolver ` from #1290, which efficiently supports such recursive, iterative searching for subtypes.

### DIFF
--- a/src/main/java/spoon/reflect/visitor/filter/AllMethodsSameSignatureFunction.java
+++ b/src/main/java/spoon/reflect/visitor/filter/AllMethodsSameSignatureFunction.java
@@ -48,6 +48,7 @@ import spoon.support.visitor.SubInheritanceHierarchyResolver;
 public class AllMethodsSameSignatureFunction implements CtConsumableFunction<CtExecutable<?>>, CtQueryAware {
 
 	private boolean includingSelf = false;
+	private boolean includingLambdas = true;
 	private CtQuery query;
 
 	public AllMethodsSameSignatureFunction() {
@@ -61,8 +62,20 @@ public class AllMethodsSameSignatureFunction implements CtConsumableFunction<CtE
 		return this;
 	}
 
+	/**
+	 * @param includingLambdas if true then extra search for {@link CtLambda} executables,
+	 * with same signature will be processed too.
+	 * If false, then it returns only {@link CtMethod} instances.
+	 * By default it is true.
+	 */
+	public AllMethodsSameSignatureFunction includingLambdas(boolean includingLambdas) {
+		this.includingLambdas = includingLambdas;
+		return this;
+	}
+
 	@Override
 	public void apply(CtExecutable<?> targetExecutable, final CtConsumer<Object> outputConsumer) {
+		final LambdaFilter lambdaFilter = new LambdaFilter();
 		final List<CtExecutable<?>> targetMethods = new ArrayList<>();
 		targetMethods.add(targetExecutable);
 		if (includingSelf) {
@@ -81,12 +94,13 @@ public class AllMethodsSameSignatureFunction implements CtConsumableFunction<CtE
 			return;
 		}
 		CtType<?> declaringType = targetMethod.getDeclaringType();
+		lambdaFilter.addImplementingInterface(declaringType);
 		//search for all declarations and implementations of this method in sub and super classes and interfaces of all related hierarchies.
 		class Context {
 			boolean haveToSearchForSubtypes;
 		}
 		final Context context = new Context();
-		//at the beginning we know that we have to always search for sub types too. 
+		//at the beginning we know that we have to always search for sub types too.
 		context.haveToSearchForSubtypes = true;
 		//Sub inheritance hierarchy function, which remembers visited sub types and does not returns/visits them again
 		final SubInheritanceHierarchyResolver subHierarchyFnc = new SubInheritanceHierarchyResolver(declaringType.getFactory().getModel().getRootPackage());
@@ -110,6 +124,7 @@ public class AllMethodsSameSignatureFunction implements CtConsumableFunction<CtE
 						targetMethods.add(overriddenMethod);
 						outputConsumer.accept(overriddenMethod);
 						CtType<?> type = overriddenMethod.getDeclaringType();
+						lambdaFilter.addImplementingInterface(type);
 						subHierarchyFnc.addSuperType(type);
 						//mark that new super type was added, so we have to search for sub types again
 						context.haveToSearchForSubtypes = true;
@@ -131,6 +146,10 @@ public class AllMethodsSameSignatureFunction implements CtConsumableFunction<CtE
 					}
 				});
 			}
+		}
+		if (includingLambdas) {
+			//search for all lambdas implementing any of the found interfaces
+			declaringType.getFactory().getModel().getRootPackage().filterChildren(lambdaFilter).forEach(outputConsumer);
 		}
 	}
 
--- a/src/main/java/spoon/reflect/visitor/filter/AllMethodsSameSignatureFunction.java
+++ b/src/main/java/spoon/reflect/visitor/filter/AllMethodsSameSignatureFunction.java
@@ -27,6 +27,7 @@ import spoon.reflect.code.CtLambda;
 import spoon.reflect.declaration.CtExecutable;
 import spoon.reflect.declaration.CtMethod;
 import spoon.reflect.declaration.CtType;
+import spoon.reflect.visitor.Filter;
 import spoon.reflect.visitor.chain.CtConsumableFunction;
 import spoon.reflect.visitor.chain.CtConsumer;
 import spoon.reflect.visitor.chain.CtQuery;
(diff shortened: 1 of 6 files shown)


pvojtechovsky:
This mapping function actually looks only for CtMethod instances. Looking for lambdas needs little bit different algorithm. I am not sure whether I should extend this mapping function or implement another one. As a client, I would prefer one mapping function, which returns all executables (methods and lambdas), but from maintenance point of view two smaller classes are easier... 

WDYT?

src/main/java/spoon/reflect/visitor/filter/AllMethodsSameSignatureFunction.java
```diff
@@ -0,0 +1,189 @@
...
+import spoon.reflect.declaration.CtType;
+import spoon.reflect.visitor.chain.CtConsumableFunction;
+import spoon.reflect.visitor.chain.CtConsumer;
+import spoon.reflect.visitor.chain.CtQuery;
+import spoon.reflect.visitor.chain.CtQueryAware;
+import spoon.support.visitor.ClassTypingContext;
+import spoon.support.visitor.SubInheritanceHierarchyResolver;
+
+/**
+ * Returns all methods/lambdas with same signature in related inheritance hierarchies.
```
monperrus: in supertypes? in subtypes? or both?

monperrus: or in the whole model?

monperrus:
> As a client, I would prefer one mapping function, which returns all executables (methods and lambdas), but from maintenance point of view two smaller classes are easier...

The client viewpoint is the most important. The API should be as intuitive and regular as possible. The client code should as simple as possible.

So we need more test case for testing lambdas?

src/main/java/spoon/reflect/visitor/filter/AllMethodsSameSignatureFunction.java
pvojtechovsky: Where it searches? Nearly in whole model. I do not know the correct terms from java theory for that, but it searches. In superTypes, subTypes of all related inheritance hierarchies.

The most exact definition of the result of this mapping function is:
> Whenever anybody needs to refactor method signature, it is necessary to found all methods in sub and super class/interface hierarchies, which overrides/are overridden by the refactored method.

pvojtechovsky:
As I described above, the purpose of this mapping function is to return all related executables, whose signature has to be changed if **client wants to refactor signature of some executable**.

It is **NOT** same/simple like:
A) return all overridden/overriding method of input executable in it's **OWN type hierarchy tree**.
B) return all methods with same signature from **whole spoon model**.

The results of this mapping function are more then A) and less then B).

Example: If you have to add new parameter into method `A#getProperty(String name)` so it will become `A#getProperty(String name, String default)`, then 
A) it is not enough to refactor all methods in the same type hierarchy (because there might be some "related" type hierarchy, which has to be refactored too) 
B) and it is wrong to refactor all getProperty(String) method's of the world.

Did you understood the contract of this method from my bad English and misusing terms? I would like to have your confirmation first about that contract, otherwise next discussions will be full of misunderstandings ;-)

The problem of current implementation is that it returns only CtMethod instances. It does not return lambdas. And my question is whether I should add searching for lambdas into the same mapping function (would make sense from contract/client point of view) or keep it as simple as possible and create another mapping function for that.
And You answered me "client viewpoint is the most important", so I will try to put it into one and do my best to have the code understandable.

pvojtechovsky:
I have added support for returning `CtLambda` instances too. By default this mapping functions returns methods and lambdas. Returning Lambdas can be disabled by `includingLambdas(false)`

src/test/java/spoon/test/refactoring/MethodsRefactoringTest.java
```diff
@@ -73,4 +89,117 @@ private void checkContainsOnly(List<String> foundNames, String... expectedNames)
...
+		checkMethodHierarchy("A_method1", factory.Class().get(TypeA.class).getMethodsByName("method1").get(0));
+		checkMethodHierarchy("A_method1", factory.Class().get(TypeC.class).getMethodsByName("method1").get(0));
+		checkMethodHierarchy("A_method1", factory.Interface().get(IFaceB.class).getMethodsByName("method1").get(0));
+		checkMethodHierarchy("A_method1", factory.Class().get(TypeL.class).getMethodsByName("method1").get(0));
+		checkMethodHierarchy("A_method1", ((CtClass<?>)factory.Class().get(TypeB.class).filterChildren(new NameFilter<>("1Local")).first()).getMethodsByName("method1").get(0));
+		checkMethodHierarchy("R_method1", factory.Class().get(TypeR.class).getMethodsByName("method1").get(0));
+		checkMethodHierarchy("R_method1", factory.Class().get(TypeS.class).getMethodsByName("method1").get(0));
+	}
+	
+	private void checkMethodHierarchy(String hierarchyName, CtMethod startMethod) {
```
monperrus: This string parameter is really hard to understand, it took me scrolling to understand.

A simpler and more intuitive solution would be to create different annotations for each hierarch `AMethod1TypeHiearachy`, `RMethod1TypeHiearachy`, etc

monperrus: as far as I understand, the first parameter is a kind of expected value.

what about 
`checkMethodHierarchy(String hierarchyName, CtMethod startMethod)` -> `assertAllMethodsSameSignatureFunction(Collection<Methoc> expected,  CtMethod startMethod)`

called this way `assertAllMethodsSameSignatureFunction(collectMethodsInHierarchy(AMethod1TypeHiearachy.class),  startMethod)`

src/test/java/spoon/test/refactoring/MethodsRefactoringTest.java
```diff
@@ -73,4 +89,117 @@ private void checkContainsOnly(List<String> foundNames, String... expectedNames)
...
+		//contract: check that by default it does not includes self and returns lambdas
+		{
+			final List<CtExecutable<?>> executables = startMethod.map(new AllMethodsSameSignatureFunction()).list();
+			assertFalse(executables.contains(startMethod));
```
monperrus: new AllMethodsSameSignatureFunction never contains the origin by contract?

src/test/java/spoon/test/refactoring/MethodsRefactoringTest.java
pvojtechovsky: Would not be a explanation comment sufficient? ;-) ... so developer does not have to scroll but can read a comment and understand?
> AMethod1TypeHiearachy, RMethod1TypeHiearachy

needs more code then one annotation, does not?

There are cases when annotation cannot be used to mark CtExecutable, which belongs to hierarchy. It is the case of Lambda expression. I "annotate" them using comment, which contains hierarchy name.

monperrus:
> There are cases when annotation cannot be used to mark CtExecutable, which belongs to hierarchy.
> It is the case of Lambda expression. I "annotate" them using comment, which contains hierarchy name.
Interesting! It's complicated enough that I even missed this point :-) even I looked at the stuff
with comments.

I'm pretty sure you can annotate a lambda or a statement in moder java (such as @SuppressWarning.

pvojtechovsky:
> I'm pretty sure you can annotate a lambda or a statement in modern java (such as @SuppressWarning.

I already spent some time trying that and I failed. Here they think it is not possible too:
[http://stackoverflow.com/questions/22375891/annotating-the-functional-interface-of-a-lambda-expression]

Show me how to do it please.

monperrus:
> I already spent some time trying that and I failed. Here they think it is not possible too:
[http://stackoverflow.com/questions/22375891/annotating-the-functional-interface-of-a-lambda-expression]

We're in Spoon, so we don't need bytecode-based or runtime annotations. You'd access the annotations on the Spoon model directly.

pvojtechovsky:
I refactored tests. 
* Lambdas are now marked by a annotation on the first statement of the lambda body.
* expected executables are collected at the beginning

It is finished from my point of view.

src/main/java/spoon/reflect/visitor/filter/AllMethodsSameSignatureFunction.java
```diff
@@ -0,0 +1,208 @@
...
+		targetMethods.add(targetExecutable);
+		if (includingSelf) {
+			outputConsumer.accept(targetExecutable);
+			if (query.isTerminated()) {
+				return;
+			}
+		}
+		CtMethod<?> targetMethod;
+		if (targetExecutable instanceof CtLambda) {
+			targetMethod = ((CtLambda) targetExecutable).getOverriddenMethod();
```
monperrus: not covered (see [coverage](https://coveralls.io/builds/11635765/source?filename=src%2Fmain%2Fjava%2Fspoon%2Freflect%2Fvisitor%2Ffilter%2FAllMethodsSameSignatureFunction.java))

because startMethod is never a lambda in your test?

pvojtechovsky: thanks I will fix it

monperrus:
thanks a lot Pavel. 

There's a last point I'd like to know, see comment.

pvojtechovsky:
> There's a last point I'd like to know, see comment.

Which comment?

I had to do some fixes, because handling of lambdas and constructors as input of mapping function was not OK.

PR #1302, opened 2017-05-16
review bug: in noClasspath type without import should be considered in the same package as current class

### LINKED ISSUE #1293: getQualifiedName returns simple name when analysing a single file (sometimes)
Hello, 

I have two classes "TransporteurFragment" and "CompletableFragment". Both of them are in the same package.
TransporteurFragment extends CompletableFragment, and CompletableFragment is an abstract class extending Fragment (from Android) and containing one abstract method.

If I scan the entire project folder, and do element.getSuperclass().getQualifiedName(), where element is the CtClass corresponding to TransporteurFragment, it returns the correct QualifiedName (x.y.z.CompletableFragment). 
However, if I scan only the TransporteurFragment file, it returns "CompletableFragment". 

At first I though it was not a bug and only a consequence of scanning a single file, but then I though about the imports that indicates where the used classes are from, and the fact that it works for other files ...

I tried to reproduce this, with basic Java files but I don't know why, it works ... 
I have however detected that if CompletableFragment is in a different package than TransporteurFragment, even when scanning only TransporteurFragment.java, it returns the correct QualifiedName, which let me think that maybe there is a real issue ;)

I hope I was clear enough, if not, I can share you the files that cause me problems !

(I use v5.6.0 of Spoon, from Maven)

surli:
This PR intends to fix #1293

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
+++ b/src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
@@ -913,7 +913,8 @@ public class ReferenceBuilder {
 		} else if (declaring instanceof CtTypeReference) {
 			ref.setDeclaringType((CtTypeReference) declaring);
 		} else if (declaring == null) {
-			ref.setPackage(jdtTreeBuilder.getFactory().Package().topLevel());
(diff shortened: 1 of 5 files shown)


monperrus:
thanks!

PR #1309, opened 2017-05-18
Baby PR: SubInheritanceHierarchyResolver

monperrus:
baby PR for  #1290


monperrus:
@pvojtechovsky WDYT?

pvojtechovsky:
Martin, thanks for Your effort. It looks quite good. Such implementation probably fits to needs of #1291 too. So I like it!

I have only minor notes. I will add them into code

src/main/java/spoon/support/visitor/SubInheritanceHierarchyResolver.java
```diff
@@ -0,0 +1,230 @@
...
+						 * FOUND! we are in super inheritance hierarchy, which extends from an searched super type(s).
+						 * All `currentSubTypes` are sub types of searched super type
+						 */
+						while (currentSubTypes.size() > 0) {
+							typeRef = currentSubTypes.pop();
+							/*
+							 * Send them to outputConsumer and add then as targetSuperTypes too, to perform faster with detection of next sub types.
+							 */
+							if (!targetSuperTypes.contains(typeRef.getQualifiedName())) {
+								targetSuperTypes.add(typeRef.getQualifiedName());
```
pvojtechovsky:
```java
if (!targetSuperTypes.contains(typeRef.getQualifiedName())) {
   targetSuperTypes.add(typeRef.getQualifiedName());
```
can be replaced by
```java
if (targetSuperTypes.add(typeRef.getQualifiedName())) {
```
You have seen that in my origin code, and you changed it, so you probably think it as not understandable. So I at least suggest
```java
String qName = typeRef.getQualifiedName();
if (!targetSuperTypes.contains(qName)) {
   targetSuperTypes.add(qName);
```
... spoon is already not fastest so why to wast time again?
I have feeling (not measured) like this code is in middle of often repeated algorithm. I plan to refactor 10 000 methods of 6000 classes and I need to search many many inheritance hiearchies of all these classes for all methods ... it can end with a big number of calls ;-)

src/main/java/spoon/support/visitor/SubInheritanceHierarchyResolver.java
```diff
@@ -0,0 +1,230 @@
...
+	private Set<String> targetSuperTypes = new HashSet<>();
+	/**
+	 * if true then we have to check if type is a subtype of superClass or superInterfaces too
+	 * if false then it is enough to search in superClass hierarchy only (faster)
+	 */
+	boolean hasSuperInterface = false;
+	private boolean failOnClassNotFound = false;
+
+	public SubInheritanceHierarchyResolver(CtPackage input) {
+		inputPackage = input;
```
pvojtechovsky: What about no package at all?
The default (root) inputPackage might be taken from the factory of CtTypeInformation delivered by first call of #addSuperType.
And if somebody ever needs to search in different then root package, then we can add #setStartPackage(CtPackage) method.
That setStartPackage is better then constructor, because it can be used to recursively scan on different packages... But I personally do not need that, so I vote for removing constructor parameter

pvojtechovsky:
That's all. Thanks!

monperrus:
> You have seen that in my origin code, and you changed it, so you probably think it as not understandable

yes! maintainability always comes first :-) 

optimization is just fine, but should come with very long and detailed comments to still be readable and maintainable.

src/main/java/spoon/support/visitor/SubInheritanceHierarchyResolver.java
monperrus: > The default (root) inputPackage might be taken from the factory of CtTypeInformation delivered by first call of #addSuperType.

This would be implicit and not regular.

> then we can add #setStartPackage(CtPackage) method.

setter = state = mutability = bad. the current design is append-only which is much better.

monperrus:
OK. Let's wait for @surli to review and merge this one, probably tomorrow.

pvojtechovsky:
This PR is OK for me. Thanks!

monperrus:
Excellent! you can rebase #1290 and #1291 then.

surli:
I'm reading the code, it seems OK but I have some questions as I'm still not really familiar with all changes introduced by Pavel :)

src/main/java/spoon/support/visitor/SubInheritanceHierarchyResolver.java
```diff
@@ -0,0 +1,232 @@
...
+			/*
+			 * listen for types in super inheritance hierarchy
+			 * 1) to collect `currentSubTypes`
+			 * 2) to check if we have already found a targetSuperType
+			 * 3) if found then send `currentSubTypes` to `outputConsumer` and skip visiting of further super types
+			 */
+			.setListener(new CtScannerListener() {
+				@Override
+				public ScanningMode enter(CtElement element) {
+					final CtTypeReference<?> typeRef = (CtTypeReference<?>) element;
```
surli: Just to be sure: this cast here is always right because of the `typeFilter` and `classFilter` you're using above which guarantee you'll obtain CtTypeReference, right?

src/main/java/spoon/support/visitor/SubInheritanceHierarchyResolver.java
```diff
@@ -0,0 +1,232 @@
...
+					String qName = typeRef.getQualifiedName();
+					if (targetSuperTypes.contains(qName)) {
+						/*
+						 * FOUND! we are in super inheritance hierarchy, which extends from an searched super type(s).
+						 * All `currentSubTypes` are sub types of searched super type
+						 */
+						while (currentSubTypes.size() > 0) {
```
surli: In fact, this part remains unclear to me: as far as I understand, you assume that the elements are returned with the following order: first the leaves, and then the superclasses/interfaces, right? This is a contract of the `SuperInheritanceHierarchyFunction`? Because I'm not sure to understand that when I'm reading its documentation.

src/main/java/spoon/support/visitor/SubInheritanceHierarchyResolver.java
pvojtechovsky: This case is always right because `SuperInheritanceHierarchyFunction` by default always returns TypeReferences. My origin code declared it explicitly on line 145 where was
```java
//internally it works always with references
.returnTypeReferences(true)
```
but Martin cleared that ;-)

src/main/java/spoon/support/visitor/SubInheritanceHierarchyResolver.java
pvojtechovsky: > first the leaves, and then the superclasses/interfaces, right?

right. it is the contract of `SuperInheritanceHierarchyFunction`. It should be written in [javadoc](https://github.com/INRIA/spoon/blob/80ab53abb27cca71e30859e8cbfe9ac080fd462a/src/main/java/spoon/reflect/visitor/filter/SuperInheritanceHierarchyFunction.java#L51) of SuperInheritanceHierarchyFunction class.

surli:
Ok then thanks @pvojtechovsky for the explanations :)

PR #1314, opened 2017-05-20
review: fix: replace wrong CtTypeInformation by CtType

pvojtechovsky:
After simplifications of SubInheritanceHierarchyResolver (remove of `#returnTypeReferences(boolean)` method) there remained `CtConsumer<CtTypeInformation>`, but there should be `CtConsumer<CtType>`


monperrus:
why do you say "remove useless generic parameter", it seems to be still there?

pvojtechovsky:
Because first I tried to remove it, but it failed with compilation errors on `CtConsumer<CtType<?>>` so I had to keep generic parameter ... but with correct bound.

PR #1316, opened 2017-05-20
review: replace string literals by constants

pvojtechovsky:
replaces all "<init>" by CtExecutableReference.CONSTRUCTOR_NAME.
replaces all "lambda$" by CtExecutableReference.LAMBDA_NAME_PREFIX.


pvojtechovsky:
May be the constants are not on good place.
CtConstructor.CONSTRUCTOR_NAME
CtLambda.LAMBDA_NAME_PREFIX
might be better... We must of course keep legacy CtExecutableReference constant too. 
WDYT?

The reason for this PR is that I need lambda name prefix constant in new spoon code.

PR #1319, opened 2017-05-21
review: fix, improve and document TemplateMatcher

pvojtechovsky:
I am trying to understand how TemplateMatcher works and here are some
* javadoc added - which helps me and others to understand it better
* comments added about bugs - to plan future refactoring
* refactoring (removed some useless calls of `CtType#getReference`, which is immediately followed by `CtTypeReference#getDeclaringType`

@monperrus please have a look at it, even if it is WiP. This PR is more like list of TODO topics for future PRs. Are my changes on good way?

### DIFF
--- a/src/main/java/spoon/template/TemplateMatcher.java
+++ b/src/main/java/spoon/template/TemplateMatcher.java
@@ -61,6 +61,14 @@ import java.util.regex.Pattern;
  */
 public class TemplateMatcher implements Filter<CtElement> {
 
+	/**
+	 * Collects all AST nodes, which has to be substituted, because they represents a template parameter declared by {@link TemplateParameter}
+	 *
+	 * Searches for all invocations of {@link TemplateParameter#S()} in CtClass model of {@link Template}
+	 * @param root CtClass model of {@link Template}
+	 * @return List of all variables - only these parameters which are declared by {@link TemplateParameter}.
+	 * It doesn't includes references to parameters annotated by {@link Parameter}
+	 */
 	private List<CtInvocation<?>> getMethods(CtClass<? extends Template<?>> root) {
 		CtExecutableReference<?> methodRef = root.getFactory().Executable()
 				.createReference(root.getFactory().Type().createReference(TemplateParameter.class), root.getFactory().Type().createTypeParameterReference("T"), "S");
@@ -69,6 +77,11 @@ public class TemplateMatcher implements Filter<CtElement> {
 		return meths;
 	}
 
+	/**
+	 * @param templateType CtClass model of {@link Template}
+	 * @return list of all names of template parameters.
+	 * It includes parameters defined by {@link TemplateParameter} and parameters with annotation {@link Parameter}.
+	 */
 	private List<String> getTemplateNameParameters(CtClass<? extends Template<?>> templateType) {
 		return Parameters.getNames(templateType);
 	}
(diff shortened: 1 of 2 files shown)


pvojtechovsky:
I am finished with commenting and few safe refactorings. I would like your feedback about correctness of comments (confirms that I understood the idea of that sometime buggy code). Does it makes sense to merge that PR then?

In the mean time I will try to make some test cases - in another PRs, which will reproduce the bugs ... and then may be fix them.

monperrus:
I love commenting PRs, that's super good for long term maintenance.

However, it would be strange to merge all the "bug comments", and it would be ineffective to discuss them all here.

Would you remove the bug comments from this one so that we can review and merge the valuable explanations and refactorings first?

pvojtechovsky:
I kept only few simple fixes and java doc without TODOs.
The TODOs and BUG markers were moved to new [branch](https://github.com/pvojtechovsky/spoon/tree/todosOfTemplateMatcher)

src/main/java/spoon/template/TemplateMatcher.java
```diff
@@ -62,12 +62,9 @@
...
 
 	/**
-	 * Collects all AST nodes, which has to be substituted, because they represents a template parameter declared by {@link TemplateParameter}
+	 * Searches for all invocations of {@link TemplateParameter#S()} in "root", a CtClass model of {@link Template}
 	 *
-	 * Searches for all invocations of {@link TemplateParameter#S()} in CtClass model of {@link Template}
 	 * @param root CtClass model of {@link Template}
-	 * @return List of all variables - only these parameters which are declared by {@link TemplateParameter}.
-	 * It doesn't includes references to parameters annotated by {@link Parameter}
 	 */
```
pvojtechovsky: The idea of my comment was not "just describe what algorithm does" (like your comment), but it tried to describe the purpose of this method

monperrus: I don't understand. does it return a list of variables or parameters? or a list of invocations?

pvojtechovsky: Your new comment is absolutely correct, I just suggest to keep my comment too (in improved form), because it explains purpose of this method. 
It is used like this:
```java
public TemplateMatcher(CtElement templateRoot) {
  ...
  variables = getMethods(templateType)
  ...
}
```
so it returns template variables/parameters/substitution points (please give it correct name ;-) ) ... but as written on last javadoc line, they are not all, because variables/parameters/substitution points  annotated by {@link Parameter} are not included in returned list

monperrus:
Are you sure?

    private List<CtInvocation<?>> getMethods(CtClass<? extends Template<?>> root) {
        CtExecutableReference<?> methodRef = root.getFactory().Executable()
                .createReference(root.getFactory().Type().createReference(TemplateParameter.class),
root.getFactory().Type().createTypeParameterReference("T"), "S");
        List<CtInvocation<?>> meths = Query.getElements(root, new InvocationFilter(methodRef));

        return meths;
    }

pvojtechovsky:
Yes, I read this method carefully.
Do we speak about same thing? ;-)

monperrus:
We probably don't speak about the same thing. Let's merge it as is?

pvojtechovsky:
Yes, I agree to merge it

PR #1322, opened 2017-05-24
fix: auto-import works for compilation units with multiple classes

monperrus:


src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
```diff
@@ -1796,12 +1796,8 @@ public void reset() {
...
 	public void calculate(CompilationUnit sourceCompilationUnit, List<CtType<?>> types) {
 		this.sourceCompilationUnit = sourceCompilationUnit;
-		// reset the import scanner between each compilationunit
-		if (env.isAutoImports()) {
-			this.importsContext = new ImportScannerImpl();
-		} else {
-			this.importsContext = new MinimalImportScanner();
-		}
+		// warning: the import scanner used in this pretty-printer has been set once and for all in the constructor
+		// this is required to reuse import when multiple classes are in the same compilation unit
```
surli: > this is required to reuse import when multiple classes are in the same compilation unit

If they are in the same compulation unit, then we enter only once in this method and it's already using the same importContext. 
However it could mess things with different compilation units: let's say we are in no-autoimport and we have two CU, the first one use B but have a collision, then it should import B, if the second CU also use B, even without collision it will be imported too and that's not the expected behaviour.

src/main/java/spoon/reflect/visitor/ImportScannerImpl.java
```diff
@@ -207,9 +172,6 @@ public void scan(CtElement element) {
 
 	@Override
 	public Collection<CtReference> computeAllImports(CtType<?> simpleType) {
-		classImports.clear();
-		fieldImports.clear();
-		methodImports.clear();
```
surli: I added them by security if computeAllImports was called multiple times with the same instance.

src/main/java/spoon/reflect/visitor/ImportScannerImpl.java
```diff
@@ -224,11 +186,7 @@ public void scan(CtElement element) {
 
 	@Override
 	public Collection<CtTypeReference<?>> computeImports(CtElement element) {
-		classImports.clear();
-		fieldImports.clear();
-		methodImports.clear();
```
surli: This one is used to for `toString` so it can really be used multiple times with the same instance, it would be safer to keep it that way, don't you think?

src/main/java/spoon/reflect/visitor/ImportScannerImpl.java
```diff
@@ -404,7 +362,7 @@ protected boolean isImportedInClassImports(CtTypeReference<?> ref) {
 	 */
 	private boolean declaringTypeIsLocalOrImported(CtTypeReference declaringType) {
 		if (declaringType != null) {
-			if (isImportedInClassImports(declaringType) || classNamePresentInJavaLang(declaringType)) {
+			if (!isTypeInCollision(declaringType, false) && (isImportedInClassImports(declaringType) || classNamePresentInJavaLang(declaringType))) {
```
surli: Actually I was testing that and trying to import the declaring type just below. So I assume you had a problem somewhere importing a declaring type: the condition below is useless then...

PR #1323, opened 2017-05-25
bug in test testTemplateMatcher

pvojtechovsky:
There is test, which should check contract
> together with the appropriate @Parameter f, this means
> we match all methods with name f*

but there is no such contract in TemplateMatcher and test is wrong. See the new test code, which shows the real contract, which is implemented.

### DIFF
--- a/src/test/java/spoon/test/template/TemplateTest.java
+++ b/src/test/java/spoon/test/template/TemplateTest.java
@@ -33,7 +33,10 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertThat;
 import static org.junit.Assert.fail;
+import static org.hamcrest.CoreMatchers.is;
+import static java.util.Arrays.asList;
 
 public class TemplateTest {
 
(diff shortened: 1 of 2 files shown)


src/test/java/spoon/test/template/TemplateTest.java
```diff
@@ -297,6 +306,10 @@ public void testTemplateMatcher() throws Exception {
 			CtMethod meth = (CtMethod) templateKlass.getElements(new NameFilter("matcher4")).get(0);
 
 			// together with the appropriate @Parameter f, this means
+/*
+ * this is not correct. See the new test below.
```
monperrus: could you update the contract then so that we see the old version of the contract deleted and the version added?

monperrus:
I also love test-only PRs :-)

There is no API documentation to add or clarify in TemplateMatcher?

pvojtechovsky:
I fixed the test, by way the TemplateMatcher matches methods with "f*". OK?

monperrus:
> the TemplateMatcher matches methods with "f*". OK?

this is the current behavior. Do you see a more appropriate behavior?

pvojtechovsky:
I just wanted to fix the test. The templating engine is very powerful and one might write many many tests about similar contracts - I really do not want to do it ;-)

src/test/java/spoon/test/template/TemplateTest.java
```diff
@@ -294,11 +303,11 @@ public void testTemplateMatcher() throws Exception {
 			// contract: the name to be matched does not have to be an exact match
 			CtClass<?> templateKlass = factory.Class().get(CheckBoundMatcher.class);
 			CtClass<?> klass = factory.Class().get(CheckBound.class);
-			CtMethod meth = (CtMethod) templateKlass.getElements(new NameFilter("matcher4")).get(0);
+			CtMethod meth = (CtMethod) templateKlass.getElements(new NameFilter("matcher5")).get(0);
```
monperrus: why do you change this line?

pvojtechovsky:
because "matcher4" already returns only methods starting with "f" ;-) 
while "matcher5" returns others too, so the extra filtering contract tested by this test can be really tested

monperrus:
indeed. thanks a lot.

pvojtechovsky:
You are welcome ;-)

PR #1326, opened 2017-05-29
fix:NPE in isTypeInCollision

surli:


monperrus:
does this fix the regression?

surli:
> does this fix the regression?
yep

PR #1327, opened 2017-05-29
Fix bug with import java.lang.subpackage

surli:


surli:
@monperrus this one fixes jPerturb regression

src/main/java/spoon/reflect/visitor/printer/ElementPrinterHelper.java
```diff
@@ -250,6 +250,10 @@ public void writeActualTypeArguments(CtActualTypeContainer ctGenericElementRefer
 		}
 	}
 
+	private boolean isJavaLangClasses(String importType) {
+		return importType.matches("^(java.lang.){1}[^.]*$");
```
monperrus: do you need "{1}"? should we escape the dots java\\.lang\\.?

surli: Done.

PR #1334, opened 2017-05-29
remove redundant 'final' modifier

pvojtechovsky:


monperrus:
why are they redundant?

pvojtechovsky:
My wrong configured check style pointed to them and I agreed that they make no sense. Or is there any meaning behind these keywords in interface method parameter?
Anyway I do not persist on this change. It is cosmetic only.

PR #1336, opened 2017-05-30
chore: remove checkstyle rule

monperrus:
Too much convention is useless and no fun.
Julien Duribreux (@?) was right, we have to do something.

### DIFF
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -37,7 +37,6 @@
     <module name="NoWhitespaceBefore"/>
     <module name="OperatorWrap"/>
     <module name="ParenPad"/>
-    <module name="TypecastParenPad"/>
     <module name="WhitespaceAfter"/>
     <module name="WhitespaceAround"/>
 


msteinbeck:
I prefer having strict rules :).

monperrus:
I also prefer strict rules.

The question is whether at some point too strict becomes counter-productive.

checkstyle.xml
```diff
@@ -1,9 +1,6 @@
 <?xml version="1.0"?>
 <!DOCTYPE module PUBLIC "-//Puppy Crawl//DTD Check Configuration 1.2//EN" "http://www.puppycrawl.com/dtds/configuration_1_2.dtd">
 <module name="Checker">
-  <module name="NewlineAtEndOfFile">
-    <property name="lineSeparator" value="lf"/>
-  </module>
```
surli: I'll keep this one: keeping a line at the end of a file is better for the diff, and at least we ensure that we won't have windows-style (or mac style ;) ) end of lines.

See: http://checkstyle.sourceforge.net/config_misc.html#NewlineAtEndOfFile

monperrus:
Interesting. `NewlineAtEndOfFile` only checks the last line.

I discover that we have a mix of \\r\\n and \n in Spoon :-(  That's bad.

What's the checkstyle rule to check this and enforce we only have \\n?

surli:
> What's the checkstyle rule to check this and enforce we only have \n?

I found that:
```
<module name="RegexpMultiline">
    <property name="format" value="\r\n"/>
    <property name="message" value="Do not use Windows line endings"/>
</module>
```

Source: https://github.com/checkstyle/checkstyle/issues/893

monperrus:
awesome. would you make a PR to add it and modify the violations accordingly?

surli:
> would you make a PR to add it and modify the violations accordingly?

OK

zielint0:
I found that for Git developers git configuration during setup is important: fetch unix, commit unix.

PR #1339, opened 2017-05-30
refactor: change ImportScanner interface to add getAllImports

### LINKED ISSUE #1329: refactor: ImportScanner#computeImports should return void
In `ImportScanner`: `computeImports` and `computeAllImports` should return `void`. Add a new method to return imports.

surli:
This PR fixes #1329 by adding a new method `getAllImports` and by deprecating `computeImports` and `computeAllImports` because their return type will change in the future to void.

### DIFF
--- a/src/main/java/spoon/reflect/visitor/ImportScanner.java
+++ b/src/main/java/spoon/reflect/visitor/ImportScanner.java
@@ -34,6 +34,7 @@ public interface ImportScanner {
 	 * @return class imports computed by Spoon, it does not contain static imports
 	 * @deprecated This method signature will change to return void in order to improve encapsulation: Use getAllImports to get the imports after calling it.
 	 */
+	@Deprecated
 	Collection<CtTypeReference<?>> computeImports(CtElement element);
 
 	/**
@@ -43,6 +44,7 @@ public interface ImportScanner {
 	 * @return imports computed by Spoon, it can be CtTypeReference (for classes), but also CtFieldReference (static field) or CtExecutableReference (static methods)
 	 * @deprecated This method signature will change to return void in order to improve encapsulation: Use getAllImports to get the imports after calling it.
 	 */
+	@Deprecated
 	Collection<CtReference> computeAllImports(CtType<?> simpleType);
 
 	/**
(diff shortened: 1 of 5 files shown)


monperrus:
Do we have to return the generic type? Would a cast be enough?

(a deprecation should not come with a signature change)

--Martin

surli:
> Do we have to return the generic type? Would a cast be enough?

Yes, because it can be a `CtTypeReference`, a `CtFieldReference` or a `CtExecutableReference`. 

> (a deprecation should not come with a signature change)

Yes, how to deprecate the usage of the return type then?... 
I cannot create new methods returning void in the interface.

src/main/java/spoon/reflect/visitor/ImportScanner.java
```diff
@@ -32,18 +31,27 @@
 	 * Computes import of a {@link spoon.reflect.declaration.CtElement}
 	 *
 	 * @return class imports computed by Spoon, it does not contain static imports
+	 * @deprecated This method signature will change to return void: Use getAllImports to get the imports after calling it.
 	 */
-	Collection<CtTypeReference<?>> computeImports(CtElement element);
+	Collection<CtReference> computeImports(CtElement element);
```
surli: Undo that change

src/main/java/spoon/reflect/visitor/ImportScanner.java
```diff
@@ -32,18 +31,27 @@
 	 * Computes import of a {@link spoon.reflect.declaration.CtElement}
 	 *
 	 * @return class imports computed by Spoon, it does not contain static imports
+	 * @deprecated This method signature will change to return void: Use getAllImports to get the imports after calling it.
```
surli: Add the reason: for improving encapsulation and design

src/main/java/spoon/reflect/visitor/ImportScanner.java
```diff
@@ -32,6 +32,7 @@
 	 * Computes import of a {@link spoon.reflect.declaration.CtElement}
 	 *
 	 * @return class imports computed by Spoon, it does not contain static imports
+	 * @deprecated This method signature will change to return void in order to improve encapsulation: Use getAllImports to get the imports after calling it.
```
monperrus: the @deprecated should also be outside the javadoc

surli: Done.

PR #1348, opened 2017-05-31
review: Substitution#insertAll should bypass Template methods

surli:


surli:
@monperrus ok to merge this one?

PR #1350, opened 2017-05-31
WiP: refactor SubstitutionVisitor. Unify conversion of template parameters

pvojtechovsky:
Motivation: clean and simplify code of `SubstitutionVisitor`, before next refactoring, which will support decoupled `template model` and 'template parameters', which allows even more powerful templates.

Following changes were done on `SubstitutionVisitor` in this PR:
* the list of parameter names, was replaced by Map of parameter names mapped to parameter values - is faster and is first step on the way to decouple template model and template parameters.
* the conversion of template parameter values was unified and extracted into new `Parameters#getParameterValueAsXXX` methods - it is now cleaner what it does, corner cases are handled consistently, errors are reported. duplicate code was removed.
* few bugs were fixed

TODO: refactor array access (the places which still use `Parameters#getValue(...)`)

### DIFF
--- a/src/test/java/spoon/test/template/TemplateTest.java
+++ b/src/test/java/spoon/test/template/TemplateTest.java
@@ -24,6 +24,7 @@ import spoon.template.TemplateParameter;
 import spoon.test.template.testclasses.ArrayAccessTemplate;
 import spoon.test.template.testclasses.InvocationTemplate;
 import spoon.test.template.testclasses.SecurityCheckerTemplate;
+import spoon.test.template.testclasses.SubstituteRootTemplate;
 import spoon.test.template.testclasses.bounds.CheckBound;
 import spoon.test.template.testclasses.bounds.CheckBoundMatcher;
 import spoon.test.template.testclasses.bounds.CheckBoundTemplate;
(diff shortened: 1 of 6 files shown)


src/test/java/spoon/test/template/TemplateTest.java
```diff
@@ -503,4 +504,21 @@ public void testTemplateArrayAccess() throws Exception {
 		CtMethod<?> m2 = resultKlass.getMethod("method2");
 		assertEquals("java.lang.System.out.println(\"second\")", m2.getBody().getStatement(0).toString());
 	}
+
+	@Test
+	public void testStatementTemplateRootSubstitution() throws Exception {
+		//contract: the template engine supports substitution of root element
```
monperrus: what do you mean by "root element"?

in `new SubstituteRootTemplate(param).apply(resultKlass);` do you mean `param` or `resultKlass` or something else?

src/test/java/spoon/test/template/TemplateTest.java
```diff
@@ -503,4 +504,21 @@ public void testTemplateArrayAccess() throws Exception {
...
+		Launcher spoon = new Launcher();
+		spoon.addTemplateResource(new FileSystemFile("./src/test/java/spoon/test/template/testclasses/SubstituteRootTemplate.java"));
+
+		spoon.buildModel();
+		Factory factory = spoon.getFactory();
+
+		CtClass<?> templateClass = factory.Class().get(SubstituteRootTemplate.class);
+		CtBlock<Void> param = (CtBlock) templateClass.getMethod("sampleBlock").getBody();
```
monperrus: rename to `templateParam`? (`param` suggests method parameter)

src/test/java/spoon/test/template/testclasses/SubstituteRootTemplate.java
```diff
@@ -0,0 +1,28 @@
...
+import spoon.template.Local;
+import spoon.template.Parameter;
+import spoon.template.StatementTemplate;
+import spoon.template.TemplateParameter;
+
+public class SubstituteRootTemplate extends StatementTemplate {
+
+	@Override
+	public void statement() throws Throwable {
+		block.S();
```
pvojtechovsky: > what do you mean by "root element"?

This is root element of `testStatementTemplateRootSubstitution()`

src/test/java/spoon/test/template/TemplateTest.java
pvojtechovsky: Root element is in this case the body of the `statement` method.

pvojtechovsky:
@monperrus 
The new test `testStatementTemplateRootSubstitution()` fails on HEAD. The PR #1354 fixes it, but yesterday  I found that it is only partial fix, because client can replace any element by
A) another element
B) no element
C) more then one another element - and this is the case which is not covered yet

The solution of C) will simplify SubstitutionVisitor (forEach and CtExecutable#parameters replacing, ...), but it causes deep refactoring too ... so I am actually checking how it will look like and then we can have a look how to split it into smaller PRs...

src/test/java/spoon/test/template/TemplateTest.java
monperrus: does that mean that the  body of the statement method is never substituted?

pvojtechovsky: yes, if I remember well (90%) it was not substituted. Or it failed with exception (10%)

pvojtechovsky:

PR #1354, opened 2017-06-01
WiP: fix substitution of root statement in StatementTemplate

pvojtechovsky:


src/main/java/spoon/support/template/SubstitutionVisitor.java
```diff
@@ -482,4 +488,18 @@ public void scan(CtElement element) {
 		} catch (DoNotFurtherTemplateThisElement ignore) {
 		}
 	}
+
+	public CtElement substitute(CtElement element) {
```
surli: So the main entry point of this class is no more `scan` but `substitute` right? It could be good to document that.

src/main/java/spoon/template/StatementTemplate.java
```diff
@@ -43,8 +43,7 @@ public CtStatement apply(CtType<?> targetType) {
 		CtClass<?> c = Substitution.getTemplateCtClass(targetType, this);
 		// we substitute the first statement of method statement
 		CtStatement result = c.getMethod("statement").getBody().getStatements().get(0).clone();
-		new SubstitutionVisitor(c.getFactory(), targetType, this).scan(result);
-		return result;
+		return (CtStatement) new SubstitutionVisitor(c.getFactory(), targetType, this).substitute(result);
```
surli: It seems that the class Substitute also use the `SubstitutionVisitor.scan()` usage you let it as it is on purpose or is it a miss?

src/main/java/spoon/support/template/SubstitutionVisitor.java
pvojtechovsky: yes, it is the entry point now. I forgot to document it. Thank You for noticing that!

src/main/java/spoon/template/StatementTemplate.java
pvojtechovsky: I missed it. I focused on SubstitutionVisitor and did not checked other calls.
Now it uses `SubstitutionVisitor#substutute` too.

pvojtechovsky:

PR #1356, opened 2017-06-03
WiP: refactor SubstitutionVisitor decouple template model and parameters

pvojtechovsky:
adds new constructor into `SubstitutionVisitor`, which creates substitution visitor, whose evaluation does not need/depend classes `Template`, `TemplateParameter` or `Parameter`. 

The `SubstitutionVisitor` can be now used to substitute values in any AST model (doesn't have to extend Template) using the dynamically created parameters (doesn't have to be declared in any Template class).

I will add test case soon...


src/test/java/spoon/test/template/TemplateTest.java
```diff
@@ -360,6 +365,36 @@ public void testExtensionBlock() throws Exception {
 	}
 
 	@Test
+	public void testExtensionDecoupledSubstitutionVisitor() throws Exception {
+		final Launcher launcher = new Launcher();
```
monperrus: could you add a contract?

pvojtechovsky:

Issue #1284, opened 2017-05-05
feature: add support for sniper mode (only rewrites the changed nodes)


monperrus:
see ROADMAP

monperrus:
depends on #1168

monperrus:
useful for repairnator and co

tdurieux:
see #1358

monperrus:
closed per #1927

monperrus:
FYI:
An Algorithm for Layout Preservation in Refactoring Transformations
https://link.springer.com/chapter/10.1007/978-3-642-28830-2_3

monperrus:
More papers on high-fidelity / layout preserving code transformation:

High-fidelity C/C++ code transformation (2017)
https://www.sciencedirect.com/science/article/pii/S0167642307000718/pdf

Detaching and reconstructing the documentary structure of source code (2010)
https://www.researchgate.net/profile/Gabor_Kusper/publication/284888833_ICAI2010-volume2/links/5659ad0408aeafc2aac4c23a/ICAI2010-volume2.pdf#page=401

Layout preserving parser for refactoring in Erlang (2009)
https://pdfs.semanticscholar.org/8487/eeda4c867e8f7cd13d83391488b2abec4a39.pdf

revcpp: A reversible c++ preprocessor (2015)
https://www.lrde.epita.fr/~sigoure/revcpp.pdf

Scrap you reprinter (2017)
https://www.cs.kent.ac.uk/people/staff/dao7/publ/reprinter2017.pdf

monperrus:
Transformation tools with high-fidelity / layout preserving features:

* [ASF+SDF](http://www.meta-environment.org/), superseeded by [Rascal (generic)](https://github.com/usethesource/rascal)
* [StrategoXT (generic)](https://github.com/metaborg/strategoxt)
* [Cscout (for C)](https://github.com/dspinellis/cscout)
* [HaRe (for Haskell)](https://github.com/alanz/HaRe)
* Clair for C++

jurgenvinju:
* [Clair](https://github.com/cwi-swat/clair) is based on Rascal, provides concrete syntax matching for an external C++ parser
* Clair-related paper: concrete syntax with black box parsers: https://homepages.cwi.nl/~jurgenv/papers/Programming-2019.pdf

Relevant papers:
* Term Rewriting with Layout: https://homepages.cwi.nl/~jurgenv/papers/RULE-2000.pdf
[...]

Issue #1168, opened 2017-02-04
feature: Detection of model change


pvojtechovsky:
The  model change detection is needed when:
* caching some derived information. The cache should be invalidated when model element which is source for the information in cache changes.
* model of higher abstraction level is based on spoon model. Here it needs to detect changes to be able to keep model up to date.

# Requirements
> R1 detects change of model element

For example if CtClass X changes name, then it should be possible to detect that on X

> R2 detects change of any child model element

For example if an statement is added/removed/modified in body of an method then it should be possible to detect that change on all parents: CtBlock, CtMethod, CtClass, CtPackage, CtModel.

> R3 detects change of model element referred by defined physical or computed relation

For example if method is added/removed/modified in CtClass A, then it should be possible to detect that on all classes, which directly or indirectly inherits/extends class A.

# Solutions
Let's collect here the ideas how to solve that.

## Solution 1 - Change events
We might fire event (call an method) on each change. The clients which needs to detect the change would listen for these events (using registered listener).

### Features
The clients is informed immediately when node is changed

### Problems
it can become performance problem, when it is used heavily and each change would fire many events ...
it can need a lot of memory to keep collections of listeners

## Solution 2 - CtElement#modelVersion attribute
We might have this code
```java
class CtElement {
   long modelVersion = 0;
  //returns true if this model or any child element was changed after time represented by modelVersionSnapshot
   boolean wasChangedAfter(long modelVersionSnapshot) {
      return modelVersionSnapshot<modelVersion;
  }
  public long getModelVersion() {
     return modelVersion;
  }
}
class CtModel {
  long version = 0;
  public long incrementVersion() {
    version++
    return version
  }
/*
* increments version of model and applies that version to modified element(s) and all parents
* @param e - the element whose content was changed
*/
  public long onElementChanged(CtElement... e) {
     long version = incrementVersion();
     e.modelVersion=version;
     e.forEachParent(parent->parent.modelVersion = version);
     return version;
  }
  public long getModelVersionSnapshot() {
     return version;
  }
}
```
### Features
The clients have to ask whether element was changed. The client is not informed immediately.

We have one model version counter. 
* It allows to apply new version to model elements using complex algorithms, which visits not only parents of changed element, but also related elements. Thanks to one version number, it is easy to detect that visiting algorithm already applied new version to the element and we can easily avoid infinite loops.
* It allows clients to store only one number - as snapshot - and detect if model was changed using that snapshot.
[...]

PR #1358, opened 2017-06-04
WIP: feat: sniper mode (implements #1284)

tdurieux:
- [ ] AbstractInvocation
- [ ] AnnotationFieldAccess
- [x] Annotation
- [ ] AnnotationMethod
- [ ] ArrayAccess
- [ ] Assert
- [ ] Assignment
- [ ] BinaryOperator
- [ ] BodyHolder
- [ ] Case
- [ ] Catch
- [ ] Class
- [x] Comment
- [ ] Conditional
- [ ] ConstructorCall
- [ ] Constructor
- [ ] Do
- [ ] Element
- [ ] Enum
- [ ] Executable
- [ ] ExecutableReferenceExpression
- [ ] Expression
- [ ] FieldAccess
- [ ] FieldReference
- [ ] ForEach
- [ ] For
- [ ] FormalTypeDeclarer
- [ ] If
- [ ] IntersectionTypeReference
- [ ] Invocation
- [ ] JavaDoc
- [ ] JavaDocTag
- [ ] LabelledFlowBreak
- [ ] Lambda
- [ ] Literal
- [ ] Method
- [ ] Modifiable
- [ ] MultiTypedElement
- [ ] NamedElement
- [ ] NewArray
- [ ] NewClass
- [ ] OperatorAssignment
- [ ] Package
- [ ] Parameter
- [ ] Return
- [ ] RHSReceiver
- [ ] Statement
- [ ] StatementList
- [ ] Switch
- [ ] Synchronized
- [ ] TargetedExpression
- [ ] Throw
- [ ] Try
- [ ] TryWithResource
- [ ] TypedElement
- [ ] TypeInformation
- [ ] Type
- [ ] TypeParameterReference
- [ ] TypeReference
- [ ] UnaryOperator
- [ ] VariableAccess
- [ ] Variable
- [ ] While


pvojtechovsky:
Interesting feature! And there is a lot of work in it! Thank You for that effort!
I have one idea, how to profit from that code even more... could You replace all places where you create an Action and pushes it into stack by call of appropriate `Factory.Change()#onXxxYyy(...)` method?
Example:
Replace this code
```java
  		if (type != null) {
  			type.setParent(this);
  		}
		if (getFactory().getEnvironment().buildStackChanges()) {
			getFactory().getEnvironment().pushToStack(new UpdateAction(new ObjectContext(this, "type"), type, this.type));
		}
  		this.type = type;
  		return (C) this;
  	}
```
by
```java
  		if (type != null) {
  			type.setParent(this);
  		}
		getFactory().Change().onObjectUpdate(this, "type", type, this.type);
  		this.type = type;
  		return (C) this;
  	}
``` 
Why that?
* The code of Element implementations is shorter and cleaner - will be easier to process it automatically in future
* The implementation of onObjectUpdate (and all similar methods which would be then needed too) might be easily changed with another implementation or extended implementation - it would be then very easy and elegant way how to implement #1168 (Solution 2 - CtElement#modelVersion attribute)

WDYT?

pvojtechovsky:
Super! I am looking forward to having this PR in spoon! It unlocks another great use cases!
I just hope that you already convinced Martin, that this step is enough baby ;-)

> I used the sniper mode to make the change.

I am not sure what `sniper mode` means... Did I understood well that you can update origin source code, by applying only changes, without touching old code (without reformatting) ?

tdurieux:
This PR contains a lot of bugs, we start this PR more than one year ago. 
It is not always easy to follow the change.

> I just hope that you already convinced Martin, that this step is enough baby ;-)

3000 added line is not that big ;) 

> Did I understood well that you can update origin source code, by applying only changes, without touching old code (without reformatting) ?

Exactly, it will never work  for every case but for most the case it will work (rename variable, add methods, change statement, change invocations, ...)

pvojtechovsky:
>>  can update origin source code

> Exactly

Cool!!! 👍

src/main/java/spoon/reflect/visitor/printer/sniper/SniperElement.java
```diff
@@ -0,0 +1,35 @@
...
+package spoon.reflect.visitor.printer.sniper;
+
+import spoon.diff.Action;
+import spoon.diff.AddAction;
+import spoon.diff.DeleteAction;
+import spoon.diff.DeleteAllAction;
+import spoon.diff.UpdateAction;
+
+public interface SniperElement {
+	AddAction onAdd(AddAction action);
```
monperrus: throw an exception if action not handled

signature return void or SniperElement

monperrus:
still rebasable after #1377, see http://github.com/monperrus/spoon/tree/feat_sniper

tdurieux:
I did not push the last version of this branch.

pvojtechovsky:
I just read this code. There is really a lot of work in it and looks like it still needs some work. I personally would like this feature a lot!! It would increase usability of Spoon in real life projects, where is not acceptable to change formatting/style of whole file just because one line is added.

I suggest to extract part of this PR and merge it separately. I am speaking about all calls of new `Change()` factory like: `getFactory().Change().onXxx(...)`
The implementation of `onXxx` methods might be empty at the beginning and it would be nice if Spoon would allow client to hook at this calls easily.
I think that this part of PR is finished and there is nothing todo. And these `onXxx` calls might be used to detect changes in the model and to implement other nice features ... (improve performance, thanks to better caching, solve some features, etc....)
WDYT?

tdurieux:
I agree with you.
I have still to fix the roles, update the contact checker and add listener support in factory.Change()

Martin is also not a big fan of onXxx. I didn't really understand what he wanted to do

pvojtechovsky:
Before I read this PR I though about different implementation of this feature. May be it is bad idea and it cannot work, but actually it seems to me like it would be less effort to implement it.

The idea is like this
* do not collect each action on the model, but just detect if the element was modified since it's reading from sources. Let's mark modified element `ele.dirty=true` and let's mark element whose any child is dirty `ele.anyChildDirty=true`.

The pretty printer would use this algorithm:
* if the element `ele.dirty==false && ele.anyChildDirty==false', then do not scan children of this element, but simply copy all the characters from the source file from ele.position.start to ele.position.end
* if the element `ele.dirty==false && ele.anyChildDirty==true', then scan children until first dirty child is found. Then copy all the characters from the source file from ele.position.start to firstDirtyEle.position.start, then "somehow" ;-) render dirty element and search for next dirty child element ... copy not modified sources and render modified ...
* the removed elements would be removed from printed result, because these elements are not found by scanning and therefore it's source is not copied from old to new content.
* the added elements would be pretty printed using existing printer.
* the moved elements might move too, because their source is copied on the new position

If that algorithm would work, then it would be much less work, because we do not have to care about each possible operation on the model.
WDYT?

tdurieux:
The first problem that I found is:
If you change the modifier or the name of a class you will reprint the complete class. 

And You will still have the issue to know where to print the new element (no position).


 The goal of this approach is to minify the resulting diff

pvojtechovsky:
> Martin is also not a big fan of onXxx.
Optionally I would like one method `Factory#onChange(operationType, role, target, oldValue, newValue, ...?...)`.

> If you change the modifier or the name of a class you will reprint the complete class.
yes, it is the bigger challenge of this approach - to detect headers, footers and separators of the elements which contains other elements and in this case to update only class header.

monperrus:
> Interesting feature! And there is a lot of work in it

Indeed. The will probably a long-standing WIP PR, probably to be merged this Fall. The goal of having it here is to keep it constantly on sync with "master", while progressing on it.

> Martin is also not a big fan of onXxx. 

There is a misunderstanding, I think it's a promising direction.

> I suggest to extract part of this PR and merge it separately. I am speaking about all calls of new `Change()` factory like: `getFactory().Change().onXxx(...)`

Can we use the sniper mode itself to do this from to extract such a new PR from this PR? :-)

tdurieux:
@leventov I changed to `ChangeFactory.java` what do you think?

pvojtechovsky:
ChangeFactory.java is ok for me

pvojtechovsky:
Another idea how to implement pretty printer, which produces sources with minimum differences to origin sources... To collect and remember separators between elements in origin sources and then use them during pretty printing.

* no need to detect changes at all
* the position of each element would be extended with information about prefix and suffix separator
* the current pretty printer would use these separators

I am thinking about optional ways how to do it, because I have feeling like current approach, which applies all the model changes separatelly is much more complicated and may be for some combinations of actions impossible.

Side effect: If we would have separators stored in model, then clients might influence the formatting of generated code (e.g code generated from template would keep formatting of the template) ... I guess clients would like it too.

placeybordeaux:
This branch appears to be ~400 commits behind at this point. As a new comer to the project I couldn't figure out how to merge the changes cleanly, but something like sniper support, or better support for outputting code that conforms to some style guide is a must have to use for my purposes.

Is this still the best chance for 'sniper' support?

pvojtechovsky:
@placeybordeaux The PR #1927 tries to implement the same feature using different concept. Try it

monperrus:
Closed per #1927. Thanks Thomas for all your inputs and ideas on this topic.

PR #1359, opened 2017-06-04
review: refactor SubstitutionVisitor

pvojtechovsky:
Here is refactored SubstitutionVisitor. There are following changes:

C1) The SubstitutionVisitor can be used on any Spoon model and Map<String, Object> parameters. It does not need `Template`, `TemplateParameter`, `Parameter`. Of course they may be still used if it makes sense for client. See new constructor `SubstitutionVisitor(Factory f, Map<String, Object> templateParameters)`

C2) The SubstitutionVisitor can substitute the root node of substitution model by zero, one or more nodes.
See new method `<E extends CtElement> List<E> SubstitutionVisitor#substitute(E element)`

C3) There is possible to directly generate new Class, Interface, Enum, ... thanks to (C1). See
`CtType<T> Substitution#insertType(CtPackage targetPackage, String typeName, CtType<T> templateOfType, Map<String, Object> templateParameters)`

C4) The SubstitutionVisitor use Context, which knows current substitution parameters. It is needed to correctly substitute all parameters of blocks generated by `SubstitutionVisitor#visitCtForEach`

C5) The SubstitutionVisitor has less scanning methods now. It does all substitutions here:
* scanCtElement - substitutes comments, which are on all elements
* scanCtNamedElement - substitutes names or part's of names. Or replaces whole nodes by zero, one or more other nodes. For example legacy CtExecutable one parameter replacement by many parameters, is implemented this more generic way. Now we can replace everything this way. For example: one Annotation by many annotations; one field by many fields; one enum value many enum values, ...
* scanCtReference - substitutes all kinds of references. See code for each case
* visitCtForEach - like before
* visitCtFieldRead, visitCtFieldWrite - like before
* visitCtInvocation - like before

C6) the package protected fields of SubstitutionVisitor were made private - may be it was public intentional? Then I will rollback this change

C7) the conversion of parameter values to required types were extracted to private methods and made more reliable. Missuses throws exception, and if conversion is possible then parameter value is automatically converted to required type. See SubstitutionVisitor#getParameterValueAsXxx methods.

### DIFF
--- a/src/test/java/spoon/test/template/TemplateArrayAccessTest.java
+++ b/src/test/java/spoon/test/template/TemplateArrayAccessTest.java
@@ -24,8 +24,8 @@ public class TemplateArrayAccessTest {
 		Factory factory = spoon.getFactory();
 
 		CtClass<?> resultKlass = factory.Class().create("Result");
-		CtStatement result = new SubstituteArrayAccessTemplate(new String[]{"a","b"}).apply(resultKlass);
-		assertEquals("new java.lang.String[]{ \"a\" , \"b\" }.toString()", result.toString());
+		CtStatement result = new SubstituteArrayAccessTemplate(new String[]{"a",null,"b"}).apply(resultKlass);
+		assertEquals("new java.lang.String[]{ \"a\" , null , \"b\" }.toString()", result.toString());
 	}
 
 	@Test
@@ -38,7 +38,7 @@ public class TemplateArrayAccessTest {
 		Factory factory = spoon.getFactory();
 
 		CtClass<?> resultKlass = factory.Class().create("Result");
-		CtStatement result = new SubstituteArrayLengthTemplate(new String[]{"a","b"}).apply(resultKlass);
-		assertEquals("if (2 > 0);", result.toString());
+		CtStatement result = new SubstituteArrayLengthTemplate(new String[]{"a",null,"b"}).apply(resultKlass);
+		assertEquals("if (3 > 0);", result.toString());
 	}
 }
--- a/src/test/java/spoon/test/template/TemplateArrayAccessTest.java
+++ b/src/test/java/spoon/test/template/TemplateArrayAccessTest.java
@@ -10,6 +10,7 @@ import spoon.reflect.declaration.CtClass;
 import spoon.reflect.factory.Factory;
 import spoon.support.compiler.FileSystemFile;
 import spoon.test.template.testclasses.SubstituteArrayAccessTemplate;
+import spoon.test.template.testclasses.SubstituteArrayLengthTemplate;
 
 public class TemplateArrayAccessTest {
 
@@ -27,4 +28,17 @@ public class TemplateArrayAccessTest {
 		assertEquals("new java.lang.String[]{ \"a\" , \"b\" }.toString()", result.toString());
 	}
 
+	@Test
+	public void testArrayLengthAccess() throws Exception {
+		//contract: the template engine replaces length of collection of parameter values by number
+		Launcher spoon = new Launcher();
+		spoon.addTemplateResource(new FileSystemFile("./src/test/java/spoon/test/template/testclasses/SubstituteArrayLengthTemplate.java"));
+
+		spoon.buildModel();
+		Factory factory = spoon.getFactory();
+
+		CtClass<?> resultKlass = factory.Class().create("Result");
+		CtStatement result = new SubstituteArrayLengthTemplate(new String[]{"a","b"}).apply(resultKlass);
+		assertEquals("if (2 > 0);", result.toString());
+	}
 }
--- a/src/test/java/spoon/test/template/TemplateInvocationSubstitutionTest.java
+++ b/src/test/java/spoon/test/template/TemplateInvocationSubstitutionTest.java
@@ -24,10 +24,10 @@ public class TemplateInvocationSubstitutionTest {
 		spoon.buildModel();
 		Factory factory = spoon.getFactory();
 
-		CtBlock<?> model = factory.Class().get(InvocationSubstitutionByStatementTemplate.class).getMethod("sample").getBody();
+		CtBlock<?> templateArg = factory.Class().get(InvocationSubstitutionByStatementTemplate.class).getMethod("sample").getBody();
 		
 		CtClass<?> resultKlass = factory.Class().create("Result");
-		CtStatement result = new InvocationSubstitutionByStatementTemplate(model).apply(resultKlass);
+		CtStatement result = new InvocationSubstitutionByStatementTemplate(templateArg).apply(resultKlass);
 		assertEquals("throw new java.lang.RuntimeException(\"Failed\")", result.toString());
 	}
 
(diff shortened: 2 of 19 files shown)


monperrus:
rebase?

pvojtechovsky:
I have added some more tests to increase coverage of the new implementation of SubstitutionVisitor.
This PR is ready for review. I know that it is not baby step, but it is hard to split it, because it depends on each other ...

src/test/java/spoon/test/template/TemplateInvocationSubstitutionTest.java
```diff
@@ -0,0 +1,48 @@
...
+	@Test
+	public void testInvocationSubstitutionByStatement() throws Exception {
+		//contract: the template engine supports substitution of any method invocation
+		Launcher spoon = new Launcher();
+		spoon.addTemplateResource(new FileSystemFile("./src/test/java/spoon/test/template/testclasses/InvocationSubstitutionByStatementTemplate.java"));
+
+		spoon.buildModel();
+		Factory factory = spoon.getFactory();
+
+		CtBlock<?> model = factory.Class().get(InvocationSubstitutionByStatementTemplate.class).getMethod("sample").getBody();
```
monperrus: model confusing here what about templateArg

src/test/java/spoon/test/template/TemplateArrayAccessTest.java
```diff
@@ -23,8 +24,21 @@ public void testArrayAccess() throws Exception {
...
 		CtClass<?> resultKlass = factory.Class().create("Result");
-		CtStatement result = new SubstituteArrayAccessTemplate(new String[]{"a","b"}).apply(resultKlass);
-		assertEquals("new java.lang.String[]{ \"a\" , \"b\" }.toString()", result.toString());
+		CtStatement result = new SubstituteArrayAccessTemplate(new String[]{"a",null,"b"}).apply(resultKlass);
+		assertEquals("new java.lang.String[]{ \"a\" , null , \"b\" }.toString()", result.toString());
 	}
 
+	@Test
+	public void testArrayLengthAccess() throws Exception {
+		//contract: the template engine replaces length of collection of parameter values by number
```
monperrus: OK

src/test/java/spoon/test/template/TemplateInvocationSubstitutionTest.java
```diff
@@ -0,0 +1,48 @@
...
+		
+		CtClass<?> resultKlass = factory.Class().create("Result");
+		CtStatement result = new InvocationSubstitutionByStatementTemplate(model).apply(resultKlass);
+		assertEquals("throw new java.lang.RuntimeException(\"Failed\")", result.toString());
+	}
+
+	@Test
+	public void testInvocationSubstitutionByExpression() throws Exception {
+		//contract: the template engine supports substitution of any method invocation
```
monperrus: //contract: the template engine supports substitution of method arguments

src/test/java/spoon/test/template/TemplateTest.java
```diff
@@ -539,5 +611,75 @@ public void testTemplateArrayAccess() throws Exception {
 		//check that both @Parameter usage was replaced by appropriate parameter value
 		CtMethod<?> m2 = resultKlass.getMethod("method2");
 		assertEquals("java.lang.System.out.println(\"second\")", m2.getBody().getStatement(0).toString());
+		//check that substitution by missing value correctly produces empty expression
+		assertEquals("java.lang.System.out.println(null)", m2.getBody().getStatement(1).toString());
+	}
+
+	@Test
+	public void testStatementTemplateRootSubstitution() throws Exception {
+		//contract: the template engine supports substitution of root element
```
monperrus: I'm sure to understand. there is no "root" in this test

src/test/java/spoon/test/template/TemplateTest.java
```diff
@@ -539,5 +611,75 @@ public void testTemplateArrayAccess() throws Exception {
...
+		CtBlock<Void> templateParam = (CtBlock) templateClass.getMethod("sampleBlock").getBody();
+		
+		CtClass<?> resultKlass = factory.Class().create("Result");
+		CtStatement result = new SubstituteRootTemplate(templateParam).apply(resultKlass);
+		assertEquals("java.lang.String s = \"Spoon is cool!\"", ((CtBlock)result).getStatement(0).toString());
+	}
+
+	@Test
+	public void testInsertType() throws Exception {
+		final Launcher launcher = new Launcher();
```
monperrus: I'd rename the method `createTypeFromTemplate`

contract: the Substituion API provides a method `createTypeFromTemplate`

src/test/java/spoon/test/template/TemplateTest.java
```diff
@@ -539,5 +611,75 @@ public void testTemplateArrayAccess() throws Exception {
...
+		factory.getModel().getRootPackage().addPackage(targetPackage);
+
+		
+		Map<String, Object> parameters = new HashMap<>();
+		//replace someMethod with genMethod
+		parameters.put("someMethod", "genMethod");
+		
+		//contract: we can generate interface
+		final CtType<?> aIfaceModel = launcher.getFactory().Interface().get(AnIfaceModel.class);
+		CtType<?> genIface = Substitution.insertType(targetPackage, "GenIface", aIfaceModel, parameters);
```
monperrus: What about only passing a `String qualifiedName` to simplify the API usage?

src/test/java/spoon/test/template/TemplateInvocationSubstitutionTest.java
pvojtechovsky: it is misunderstanding. The origin contract is correct.
```java
System.out.println(_expression_().substring(1));
```
is substituted by
```java
System.out.println("abc".substring(1));
```
so not the method arguments, but any method invocation can be substituted.

src/test/java/spoon/test/template/TemplateTest.java
pvojtechovsky: SubstituteRootTemplate extends StatementTemplate, which internally calls
```java
CtStatement result = c.getMethod("statement").getBody().getStatements().get(0).clone();
```
where `result` is `block.S();` ... so it is substitution of the root element in the spoon model represented by `result`.

src/test/java/spoon/test/template/TemplateInvocationSubstitutionTest.java
pvojtechovsky: done

src/test/java/spoon/test/template/TemplateTest.java
pvojtechovsky: the Substitution.insertType was renamed to createTypeFromTemplate

src/test/java/spoon/test/template/TemplateTest.java
pvojtechovsky: old insertType renamed to createTypeFromTemplate has now only String qualifiedName instead of CtPackage and simpleName.

pvojtechovsky:
@monperrus Thanks for ideas and your time to review this bigger PR.
What next?

monperrus:
OK to merge.

PR #1361, opened 2017-06-05
fix(factory): create type reference from inner class

tdurieux:


src/main/java/spoon/reflect/factory/CodeFactory.java
```diff
@@ -573,6 +573,9 @@ public CtCatch createCtCatch(String nameCatch, Class<? extends Throwable> except
 		if (originalClass.isPrimitive()) {
 			return typeReference;
 		}
+		if (originalClass.getDeclaringClass() != null) {
+			return typeReference.setDeclaringType(createCtTypeReference(originalClass.getDeclaringClass()));
```
surli: Are you sure in that case we don't need to set the package?

tdurieux: Yes, because the package is declared the declaring type

src/test/java/spoon/test/factory/TypeFactoryTest.java
```diff
@@ -27,6 +28,10 @@ public void testCreateTypeRef() {
 
         ctTypeReference = launcher.getFactory().Code().createCtTypeReference(null);
         assertEquals(null, ctTypeReference);
+
+		ctTypeReference = launcher.getFactory().Code().createCtTypeReference(CtJavaDoc.CommentType.class);
```
surli: If you can fix the indentation there ;)

tdurieux: use space or change the indentation to tab?

tdurieux: This file is a mess...

PR #1365, opened 2017-06-06
review test(import): compare the computed imports with human imports

tdurieux:


surli:
I started to have a look on this one and I've several questions regarding the usage of `ScannerImport` and the contract we imposed. For example, first false assumptions detected by the test are the following ones:

```
spoon.support.reflect.code.CtForImpl
	spoon.reflect.ModelElementContainerDefaultCapacities.FOR_INIT_STATEMENTS_CONTAINER_DEFAULT_CAPACITY missing
	spoon.reflect.ModelElementContainerDefaultCapacities.FOR_UPDATE_STATEMENTS_CONTAINER_DEFAULT_CAPACITY missing
	spoon.reflect.ModelElementContainerDefaultCapacities unused
```

Those errors appeared because we took as a contract to always put an import preferably on the class type, instead of the field or the executable to use statically. Then, here the test detected that we don't use inside Spoon the contract we put on imports. 

So do we change our contract inside the ImportScanner or do we apply the contract on Spoon?

monperrus:
I propose to refine this test, to test for "equivalent import" import Foo = import static Foo.FIELD

surli:
>  propose to refine this test, to test for "equivalent import" import Foo = import static Foo.FIELD

OK I'll do that

surli:
Ok now I have another issue: some imports are directly related to the javadoc: for example in ProblemFixer

```
/**
 * This interface defines problem fixers. Problem fixers can be provided when a
 * problem is reported to the environment. The user can then chose what fixer to
 * use.
 *
 * @see Environment#report(Processor, org.apache.log4j.Level, CtElement, String,
 * ProblemFixer[])
 */
```
leads to the import of `spoon.compiler.Environment` which is not used inside the class.

src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
```diff
@@ -1859,6 +1859,10 @@ public void calculate(CompilationUnit sourceCompilationUnit, List<CtType<?>> typ
 		}
 
 		Set<CtReference> imports = new HashSet<>();
+		if (sourceCompilationUnit != null) {
+			imports.addAll(sourceCompilationUnit.getImports());
```
pvojtechovsky: Will it work also in cases when AST is refactored and some classes will be removed? Then it will probably produce unused imports, will not?

src/main/java/spoon/support/reflect/cu/CompilationUnitImpl.java
```diff
@@ -185,6 +190,16 @@ public int getTabCount(int index) {
 		return tabCount;
 	}
 
+	@Override
+	public Set<CtReference> getImports() {
+		return this.imports;
```
pvojtechovsky: return read only collection?

surli:
I propose that we change this test to report a warning on the logs instead of a failure to be able to merge it, and to open an issue about that problem.

I won't make improvement on how the imports are managed in Spoon in the next weeks, and it does not really help to keep this PR open IMHO.

surli:
As proposed above I changed the test to only check if no import is missing: for the unused imports, I only output a warning. For me the PR is ready.

PR #1368, opened 2017-06-06
review feat(role): field annotation with a role

tdurieux:
for each field of spoon defines for which interface it is the property

This PR is not finish I would like to have your feedback.

### DIFF
--- a/src/main/java/spoon/support/reflect/code/CtAssignmentImpl.java
+++ b/src/main/java/spoon/support/reflect/code/CtAssignmentImpl.java
@@ -21,8 +21,10 @@ import spoon.reflect.code.CtExpression;
 import spoon.reflect.code.CtRHSReceiver;
 import spoon.reflect.declaration.CtElement;
 import spoon.reflect.declaration.CtTypedElement;
+import spoon.reflect.path.CtRole;
 import spoon.reflect.reference.CtTypeReference;
 import spoon.reflect.visitor.CtVisitor;
+import spoon.support.MetamodelPropertyField;
 import spoon.support.reflect.declaration.CtElementImpl;
 
 import java.util.ArrayList;
@@ -33,12 +35,16 @@ import static spoon.reflect.ModelElementContainerDefaultCapacities.CASTS_CONTAIN
 public class CtAssignmentImpl<T, A extends T> extends CtStatementImpl implements CtAssignment<T, A> {
 	private static final long serialVersionUID = 1L;
 
+	@MetamodelPropertyField(role = CtRole.ASSIGNED)
 	CtExpression<T> assigned;
 
+	@MetamodelPropertyField(role = CtRole.ASSIGNMENT)
 	CtExpression<A> assignment;
 
+	@MetamodelPropertyField(role = CtRole.TYPE)
 	CtTypeReference<T> type;
 
+	@MetamodelPropertyField(role = CtRole.CAST)
 	List<CtTypeReference<?>> typeCasts = emptyList();
 
 	@Override
--- a/src/main/java/spoon/support/reflect/code/CtAssignmentImpl.java
+++ b/src/main/java/spoon/support/reflect/code/CtAssignmentImpl.java
@@ -24,7 +24,7 @@ import spoon.reflect.declaration.CtTypedElement;
 import spoon.reflect.path.CtRole;
 import spoon.reflect.reference.CtTypeReference;
 import spoon.reflect.visitor.CtVisitor;
-import spoon.support.MetamodelPropertyField;
+import spoon.reflect.annotations.MetamodelPropertyField;
 import spoon.support.reflect.declaration.CtElementImpl;
 
 import java.util.ArrayList;
--- a/src/test/java/spoon/test/api/MetamodelTest.java
+++ b/src/test/java/spoon/test/api/MetamodelTest.java
@@ -4,16 +4,24 @@ import org.junit.Assert;
 import org.junit.Test;
 import spoon.Launcher;
 import spoon.SpoonAPI;
-import spoon.reflect.code.CtFieldRead;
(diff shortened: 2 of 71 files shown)


src/main/java/spoon/support/reflect/code/CtAssignmentImpl.java
```diff
@@ -33,12 +34,16 @@
 public class CtAssignmentImpl<T, A extends T> extends CtStatementImpl implements CtAssignment<T, A> {
 	private static final long serialVersionUID = 1L;
 
+	@SpoonProperty(clazz = CtAssignment.class, name = SpoonProperty.PropertyName.ASSIGNED)
 	CtExpression<T> assigned;
 
+	@SpoonProperty(clazz = CtAssignment.class, name = SpoonProperty.PropertyName.ASSIGNMENT)
 	CtExpression<A> assignment;
 
+	@SpoonProperty(clazz = CtTypeReference.class, name = SpoonProperty.PropertyName.TYPE)
```
surli: CtTypedElement

pvojtechovsky:
I did not understood the purpose. What can profit from that?

The mapping of fields to correct property names, might make sense. But may be optional solution would be to rename fields, whose name do not fit to their property name. Then we might have simple rule: field name == property name (I have not checked if it is possible in all cases)

The mapping of property to interface `SpoonProperty#clazz` is ambiguous.
A) remove it completely (or let me understand what it is good for)
B) use an array of interfaces where this property occurs.

pvojtechovsky:
Did I understood well, that this PR is a step on the way to the meta model of spoon model? 

In such case I suggest something like this:

C1) we need a `property descriptor` for each spoon model property. Properties are e.g. `CtTypedElement#type` or `CtNamedElement#name`
C2) the `property descriptor` knows interface(s?) on which this property is defined. Or it has at least method `#hasProperty(CtElement target) boolean` which returns `true` if the target element has  property defined by this `property descriptor`.
C3) the `property descriptor` knows type of the property. E.g. `List` or 'Map' or 'Set' or `single value` 
C4) for each type from C3, the property descriptor is implemented as sub class/interface of `property descritor`. So we will have 'list property descriptor', 'map property descriptor', ..., 'single value property descriptor'
C5) each `property descriptor` defines spoon model accessing methods. For example `single value property descriptor` would have accessors: `#getPropertyValue(CtElement target)` and `#setPropertyValue(CtElement target, Object value)`

I1) These property descriptors might be accessible as attribute of SpoonProperty.PropertyName enum. For example `SpoonProperty.PropertyName.DECLARING_TYPE.getDescriptor()` would return `property descriptor`, which might be used to access meta information about this property and to read/modify appropriate field of the spoon model.
I2) We need a CtScannerWithContext, which would provide access to `property descriptor` of just scanned property
...

monperrus:
> Did I understood well, that this PR is a step on the way to the meta model of spoon model?

yes

> C1) we need a property descriptor for each spoon model property. Properties are e.g. CtTypedElement#type or CtNamedElement#name
> C3) the property descriptor knows type of the property. E.g. List or 'Map' or 'Set' or single value

Yes, this will be implemented as additional information in the to-be-renamed SpoonProperty.PropertyName enumeration (=I1)

> C2) the property descriptor knows interface(s?) on which this property is defined. 

Yes, this is the final goal

> Or it has at least method #hasProperty(CtElement target) boolean which returns true if the target element has property defined by this property descriptor.

Runtime support is welcome in subsequent PRs

> C5) each property descriptor defines spoon model accessing methods. For example single value property descriptor would have accessors: #getPropertyValue(CtElement target) and #setPropertyValue(CtElement target, Object value)

Yes, we aim at this too.

>  I2) We need a CtScannerWithContext, which would provide access to property descriptor of just scanned property

Welcome in subsequent PR.

It seems we're on the same wavelength.

monperrus:
per our discussion:
```
@SpoonProperty -> @MetamodelProperty
SpoonProperty.PropertyName. -> reuse and rename CtPathRole -> CtRole
name = SpoonProperty.PropertyName.LEFT_OPERAND) -> role = CtRole.LEFT_OPERAND) 
```

contract: all non-final fields must be annotated with @MetamodelProperty (checkable with architecture test)

todo: remove the `clazz` information, create `@PropertyGetter(role = X),  @PropertySetter(role = X)`, in another PR that will be merged first (for this PR: contract 1): all roles are declared at least once in a metamodel interface method) and maybe another PR for contract 2): all methods of metamodel interfaces are annotated with either `@Unsettable, @DerivedProperty @PropertyGetter or @PropertySetter`

tdurieux:
I updated this PR but I cannot create the contract because of these fields:

```java
private CtTypeReference<Void> voidType;
List<CtTypeParameterReference> formalTypeParameters = CtElementImpl.emptyList();
transient Factory factory;
protected CtElement parent;
SourcePosition position = SourcePosition.NOPOSITION;
Map<String, Object> metadata;
private CtMethod<T> valueOfMethod;
```

tdurieux:
@monperrus Done for me

src/test/java/spoon/test/api/MetamodelTest.java
```diff
@@ -46,4 +54,42 @@ public void testGetterSetterFroRole() {
 		Assert.assertEquals(expectedRoles, getterRoles);
 		Assert.assertEquals(expectedRoles, setterRoles);
 	}
+
+
+	@Test
+	/**
+	 * contract: all  all non-final fields must be annotated with {@link spoon.reflect.annotations.MetamodelPropertyField}
```
surli: Fix comment

surli:
It should be great to have a test to check if all there is a getter/setter in one interface of the class for all field with a MetamodelPopertyRole.

tdurieux:
Done

PR #1369, opened 2017-06-06
review: feature: CtElement#replace(Collection<CtElement>)

pvojtechovsky:
Support for replacing of one element by multiple.
Needed to implement more powerful `SubstitutionVisitor` ;-)

### DIFF
--- a/src/main/java/spoon/generating/replace/ReplacementVisitor.java
+++ b/src/main/java/spoon/generating/replace/ReplacementVisitor.java
@@ -21,6 +21,7 @@ import spoon.reflect.declaration.CtElement;
 import spoon.reflect.visitor.CtScanner;
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


pvojtechovsky:
Please review this one. 
What is problem of CI? Should I change something?

monperrus:
interesting feature, what about a vararg here instead of a new interface method: `void replace(CtElement... element);`

src/main/java/spoon/generating/replace/ReplacementVisitor.java
```diff
@@ -32,18 +34,24 @@
  *
  * This class is generated automatically by the processor {@link spoon.generating.ReplacementVisitorGenerator}.
  */
-class ReplacementVisitor extends CtScanner {
+class ReplacementVisitor<E extends CtElement> extends CtScanner {
```
monperrus: A new formal type parameter added to element: 'E extends spoon.reflect.declaration.CtElement'.
Why do you need this?

pvojtechovsky: Good question. I can live without that. I will remove it.

pvojtechovsky:
> vararg method: void replace(CtElement... element);

nice! I will do so.

pvojtechovsky:
It is done from my point of view

src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
```diff
@@ -177,7 +177,7 @@ public int hashCode() {
 
 	@Override
 	public void delete() {
-		replace(null);
+		replace();
```
monperrus: hard to understand. replace new CtElement[0] with a comment to explain?

monperrus:
Hi Pavel,
I've just realized that introducing varargs is breaking at the binary level (and tested with the script below). Since `replace` is commonly used, this would be quite bad for client code ... and we have to undo you last commit. I'm sorry for my bad suggestion.

```
# test to check whether the introduction of varargs breaks binary compatibility

# without vararg
cat <<EOF > Foo.java
class Foo {
public void m(String arg){}
}
EOF

javac Foo.java

rm Foo.java

ls Foo.class

# now a client class
cat <<EOF > Bar.java
class Bar {
public static void main(String[] args){
new Foo().m("vv");
System.out.println("end");
}
}
EOF

javac Bar.java

java -cp . Bar


# with vararg

cat <<EOF > Foo.java
class Foo {
public void m(String... arg){}
}
EOF
javac Foo.java

# binary compatibility?
if java -cp . Bar; 
then echo binary compatible
else echo binary incompatible
fi 

# source compatibility after recompilation?
if javac Foo.java Bar.java;
then echo source compatible
else echo source incompatible
fi 


```

pvojtechovsky:
no problem ;-) I will fix it tomorrow

PR #1377, opened 2017-06-07
review feat(model): annotes getter and setter with a CtRole

tdurieux:
This PR is related to #1368 

TODO tests

### DIFF
--- a/src/main/java/spoon/reflect/path/CtRole.java
+++ b/src/main/java/spoon/reflect/path/CtRole.java
@@ -56,6 +56,7 @@ public enum CtRole {
 	ASSIGNED,
 	MODIFIERS,
 	COMMENTS,
+	ANNOTATION_TYPE,
 	TYPES,
 	INTERFACES,
 	ANNOTATIONS,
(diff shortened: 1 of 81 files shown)


monperrus:
per our discussion
* type is for expression type (no comment, no type access)
* no role in reference interfaces
* remove contract all getters annotated 
* split MEMBER in CONSTRUCTOR/FIELD/EXECUTABLE/NESTED_TYPE

monperrus:
@surli OK for merge for me, please check. 

This will probably be adjusted later, but this is a really good start, thanks @tdurieux

src/main/java/spoon/reflect/path/CtRole.java
```diff
@@ -20,61 +20,59 @@
...
+	//IS_UPPER, // in reference only
 	IS_IMPLICIT,
 	IS_DEFAULT,
 	IS_VARARGS,
-	IS_UPPER,
 	DEFAULT_EXPRESSION,
 	THEN,
 	ELSE,
-	PACKAGE,
+	SUB_PACKAGE,
```
surli: Why not keeping `PACKAGE` here, it seems clearer?

monperrus:
because it is more explicit, the role is to be a subpackage of another one

surli:
> because it is more explicit, the role is to be a subpackage of another one

Ok I saw the usage after asking.

Issue #1183, opened 2017-02-11
feature: Spoon metamodel


pvojtechovsky:
Requirements:

* [x] annotations on metamodel properties (@Unsettable, @DerivedProperty @PropertyGetter or @PropertySetter) see #1368 
* [x] get list of metamodel classes (should be moved from test code to main code) see #1378

----------------------------------

Some code in spoon library is generated by generators. Do we have in spoon an metamodel, which can be used as good input for future generators?

Idea: if we would have a good and complete metamodel of spoon model, then this metamodel might be used to generate more sources and to implement issues like #741, #1168 faster. 

If I understood it well current spoon generators are using `CtScanner` and annotations on spoon model interfaces as source metamodel. May be we might write a code which scans current spoon sources, mainly
* class CtScanner, 
* package: spoon.reflect.code
* package: spoon.reflect.declaration
* package: spoon.reflect.reference
and to automatically build a real - easy to use :-) ... metamodel, which might be then used as input of next generation generators and may be might be even used to update spoon model itself.

... I know, it is quite big task, but may be once ...

monperrus:
To some extent, creating the Spoon of model of `spoon.reflect.declaration/code` is the spoon metamodel itself, and this is what's done for generating classes and testing the metamodel itself. 

> If I understood it well current spoon generators are using CtScanner and annotations on spoon model interfaces as source metamodel

Correct

Moving this information elsewhere make it more explicit but is one more level of indirection to understand and maintain.

pvojtechovsky:
I understand what you mean and I agree. 

I am thinking, how to minimize amount of manually written code and replace it by generated code. May be the spoon sources (even one file) might contain both manually written and generated code. The generated code would be marked (e.g by some comment) and the generator would be allowed to update only such sections. For example the sources file below contains both manually written and generated code:
```java
public class CtBreakImpl extends CtStatementImpl implements CtBreak {
	private static final long serialVersionUID = 1L;

//begin of code genereated by: ModelGeneratorTemplate#primitiveProperty
	String targetLabel;
//end of genereated code

//begin of code genereated by: ModelGeneratorTemplate#visitorPattern_accept
	@Override
	public void accept(CtVisitor visitor) {
		visitor.visitCtBreak(this);
	}
//end of genereated code

//begin of code genereated by: ModelGeneratorTemplate#primitiveGetter
	@Override
	public String getTargetLabel() {
		return targetLabel;
	}
//end of genereated code

[...]

PR #1378, opened 2017-06-07
feature: add first metamodel support

monperrus:
first step for #1183 

WDYT?

### DIFF
--- a/src/main/java/spoon/Metamodel.java
+++ b/src/main/java/spoon/Metamodel.java
@@ -1,0 +1,151 @@
+/**
+ * Copyright (C) 2006-2017 INRIA and contributors
+ * Spoon - http://spoon.gforge.inria.fr/
+ *
+ * This software is governed by the CeCILL-C License under French law and
+ * abiding by the rules of distribution of free software. You can use, modify
+ * and/or redistribute the software under the terms of the CeCILL-C license as
+ * circulated by CEA, CNRS and INRIA at http://www.cecill.info.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the CeCILL-C License for more details.
+ *
+ * The fact that you are presently reading this means that you have had
+ * knowledge of the CeCILL-C license and that you accept its terms.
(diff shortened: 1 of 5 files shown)


pvojtechovsky:
It is on good way! May be I will have Xmas sooner? ;-)

Some ideas:

May be it might return Set instead of Collection.

The returned references have assigned different Factory instance, may be we should add new type of spoon object under Factory, called  `MetaModel`, which would provide such information (and which would have access to the correct parent factory.)

Or return Set<String> with qualified names of interfaces? It is faster to search in it.

src/main/java/spoon/Metamodel.java
```diff
@@ -0,0 +1,148 @@
...
+		result.add(factory.Type().get(spoon.reflect.reference.CtIntersectionTypeReference.class));
+		result.add(factory.Type().get(spoon.reflect.reference.CtLocalVariableReference.class));
+		result.add(factory.Type().get(spoon.reflect.reference.CtPackageReference.class));
+		result.add(factory.Type().get(spoon.reflect.reference.CtParameterReference.class));
+		result.add(factory.Type().get(spoon.reflect.reference.CtReference.class));
+		result.add(factory.Type().get(spoon.reflect.reference.CtTypeParameterReference.class));
+		result.add(factory.Type().get(spoon.reflect.reference.CtTypeReference.class));
+		result.add(factory.Type().get(spoon.reflect.reference.CtUnboundVariableReference.class));
+		result.add(factory.Type().get(spoon.reflect.reference.CtVariableReference.class));
+		result.add(factory.Type().get(spoon.reflect.reference.CtWildcardReference.class));
```
surli: Why do you do this by hand instead of retrieving all classes of interesting packages by reflection?

tdurieux: Do you know the performance of reflection?

monperrus:
> May be it might return Set instead of Collection.

Agree, done.

> The returned references have assigned different Factory instance, 

Yes, but the `equals` contract works across factories.

> may be we should add new type of spoon object under Factory, called MetaModel, which would provide such information (and which would have access to the correct parent factory.)

Let's start with a stateless design first (stateless is better), we'll modify it later if needed. 

> Or return Set with qualified names of interfaces? It is faster to search in it.

I would prefer domain-specific types over strings

PR #1382, opened 2017-06-08
WiP: Fix issue related to CtTypeParameterReference and getDeclaration

surli:


surli:
I've several issue fixing this one. It seems that either it exposes even more bugs in ClassTypingContext or the way we're treating TypeParameters in Spoon is still wrong which is possible too. 

I tried to debug `testDocumentationContract`: there is a problem with a `CtTypeParameterReference` `T` from `CtTargetedExpression`. With my new implementation the declaring type is the one from `CtTargetedExpression` where before it was taking from `CtAnnotationFieldAccess` which is the parent of `CtTargetedExpression`. From my point of view, it makes more sense to take the closest one (here `CtTargetExpression`), but apparently it breaks `ClassTypingContext` because this types is not present in `typeToArguments`. Then I'm not sure if the behaviour of ClassTypingContext should change or not. Any idea @pvojtechovsky ?

Here is my breakpoint condition line 501 of `ClassTypingContext`: `typeRef.getParent(CtTypeReference.class).getQualifiedName().equals("spoon.reflect.code.CtTargetedExpression")` when executing `CommentTest#testDocumentationContract`

pvojtechovsky:
I will have a look at it

PR #1383, opened 2017-06-08
review: Fix issue related to CtTypeParameterReference and getDeclaration

pvojtechovsky:
It is the copy of #1382 with modified solution. May be it pass all tests now?

### DIFF
--- a/src/test/java/spoon/test/generics/GenericsTest.java
+++ b/src/test/java/spoon/test/generics/GenericsTest.java
@@ -1036,4 +1036,38 @@ public class GenericsTest {
 
 		assertTrue(invocationDetected);
 	}
+
+	@Test
+	public void testGetDeclarationOfTypeParameterReference() {
+		Launcher launcher = new Launcher();
+		launcher.addInputResource("./src/test/java/spoon/test/generics/testclasses/ExtendedPaella.java");
+		launcher.addInputResource("./src/test/java/spoon/test/generics/testclasses/Paella.java");
+		launcher.buildModel();
+
+		Factory factory = launcher.getFactory();
+
+		CtClass extendedPaella = factory.getModel().getElements(new NameFilter<CtClass>("ExtendedPaella")).get(0);
+		List<CtTypeParameter> typeParameterList = extendedPaella.getFormalCtTypeParameters();
+
+		assertEquals(1, typeParameterList.size());
+
+		CtMethod totoMethod = factory.getModel().getElements(new NameFilter<CtMethod>("toto")).get(0);
+		CtTypeReference returnTypeToto = totoMethod.getType();
+
+		CtType declaration = returnTypeToto.getDeclaration();
+
+		assertTrue(declaration.equals(typeParameterList.get(0)));
+
+		CtMethod machinMethod = factory.getModel().getElements(new NameFilter<CtMethod>("machin")).get(0);
+		CtTypeReference returnTypeMachin = machinMethod.getType();
+		List<CtTypeParameter> formalCtTypeParameters = machinMethod.getFormalCtTypeParameters();
+
+		assertEquals(1, formalCtTypeParameters.size());
+
+		CtType declarationMachin = returnTypeMachin.getDeclaration();
+
+		assertFalse(declarationMachin.equals(typeParameterList.get(0)));
+		assertTrue(declarationMachin.equals(formalCtTypeParameters.get(0)));
+
+	}
 }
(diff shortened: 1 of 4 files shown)


pvojtechovsky:
Even if it pass, I do not trust that algorithm. In other cases when I develop, I know what I am doing, because I understand the context. But in this case the context is too complicated.

surli:
> The new test only checks that it returns some CtTypeParameter. Can we somehow test that it returned the correct one?

I tried to add a new test but I'm not sure it's the kind of test you imagine. WDYT? Don't hesitate to remove my commit if you find it useless.

surli:
@pvojtechovsky I added some more tests, tell me if it's ok for you to merge this one.

pvojtechovsky:
I will try to have a look at Your tests today in the evening.

One idea concerning testing of correctness of type parameter references... may be we might extend code which creates Spoon model from JDT compiler AST. This would be TEST only extension, which would remember correct/expected declaration of the type parameter reference in CtElement#metadata. This extra data would allow as run a special test (e.g. on Guava library), which would check whether spoon returns correct type parameter declaration for each type parameter reference.... I really do not know if JDT compiler provides that information - I have not check that code yet ... it is just an idea how to found potential problems in code of this PR.

src/test/java/spoon/test/generics/GenericsTest.java
```diff
@@ -1036,4 +1036,38 @@ public void testDiamondComplexGenericsRxJava() {
...
+		List<CtTypeParameter> typeParameterList = extendedPaella.getFormalCtTypeParameters();
+
+		assertEquals(1, typeParameterList.size());
+
+		CtMethod totoMethod = factory.getModel().getElements(new NameFilter<CtMethod>("toto")).get(0);
+		CtTypeReference returnTypeToto = totoMethod.getType();
+
+		CtType declaration = returnTypeToto.getDeclaration();
+
+		assertTrue(declaration.equals(typeParameterList.get(0)));
```
pvojtechovsky: I preffer asserSame instead assertEquals in this case. May be there might exist a equal type parameter, whis is not correct one.

src/test/java/spoon/test/generics/GenericsTest.java
```diff
@@ -1036,4 +1036,38 @@ public void testDiamondComplexGenericsRxJava() {
...
+		CtMethod machinMethod = factory.getModel().getElements(new NameFilter<CtMethod>("machin")).get(0);
+		CtTypeReference returnTypeMachin = machinMethod.getType();
+		List<CtTypeParameter> formalCtTypeParameters = machinMethod.getFormalCtTypeParameters();
+
+		assertEquals(1, formalCtTypeParameters.size());
+
+		CtType declarationMachin = returnTypeMachin.getDeclaration();
+
+		assertFalse(declarationMachin.equals(typeParameterList.get(0)));
+		assertTrue(declarationMachin.equals(formalCtTypeParameters.get(0)));
```
pvojtechovsky: replace equals with == and != too. WDYT?

src/test/java/spoon/test/lambda/LambdaTest.java
```diff
@@ -450,4 +454,17 @@ private void assertIsWellPrinted(String expected, CtLambda<?> lambda) {
...
+	@Test
+	public void testGetDeclarationOnTypeParameterFromLambda() {
+		List<CtTypeParameterReference> listCtTPR  = launcher.getModel().getElements(new TypeFilter<>(CtTypeParameterReference.class));
+
+		for (CtTypeParameterReference typeParameterReference : listCtTPR) {
+			if (!(typeParameterReference instanceof CtWildcardReference) && typeParameterReference.getDeclaration() == null) {
+				System.err.println(typeParameterReference.getSimpleName()+" from parent "+typeParameterReference.getParent(CtClass.class).getPosition()+"  has null declaration");
+				typeParameterReference.getDeclaration();
+				fail();
+			}
```
pvojtechovsky: else { would be nice to check that returned type declaration is the correct one. The last place where it is kniwn is AST of JDT compiler.

But is just an idea. I think you can already merge this PR. The tests are fine, the code is better then before and if there is another problem hidden somewher we can fix it later.

surli: Ok let's finish it, I need it to close another one, we can always improve the tests later :) Thanks for the comments!

surli:
@monperrus could you review this one?

PR #1390, opened 2017-06-10
review: test SubstitionVisitor array access literal

pvojtechovsky:


src/test/java/spoon/test/template/TemplateArrayAccessTest.java
```diff
@@ -0,0 +1,30 @@
...
+import spoon.reflect.declaration.CtClass;
+import spoon.reflect.factory.Factory;
+import spoon.support.compiler.FileSystemFile;
+import spoon.test.template.testclasses.SubstituteArrayAccessTemplate;
+
+public class TemplateArrayAccessTest {
+
+	@Test
+	public void testArrayAccess() throws Exception {
+		//contract: the template engine supports array access substitution
```
monperrus: // contract: the template engine supports variable access substitution.

to me this is not an array access, but more a normal variable access, typed as Array.

an array access would be "array.S()[0]"  

that said the test is good,

monperrus:
I love PR with tests :-)

PR #1398, opened 2017-06-13
review: refactor(label): remove labelledStatement that was not used

### LINKED ISSUE #1381: bug: CtContinue#getLabelledStatement should be derived, computed with lookup
and setter setLabelledStatement removed

tdurieux:
The property labelledStatement was always equals to null.

Fix #1381

### DIFF
--- a/src/main/java/spoon/Metamodel.java
+++ b/src/main/java/spoon/Metamodel.java
@@ -74,6 +74,7 @@ public class Metamodel {
 		result.add(factory.Type().get(spoon.reflect.code.CtInvocation.class));
 		result.add(factory.Type().get(spoon.reflect.code.CtJavaDoc.class));
 		result.add(factory.Type().get(spoon.reflect.code.CtJavaDocTag.class));
+		result.add(factory.Type().get(spoon.reflect.code.CtLabelledFlowBreak.class));
 		result.add(factory.Type().get(spoon.reflect.code.CtLambda.class));
 		result.add(factory.Type().get(spoon.reflect.code.CtLiteral.class));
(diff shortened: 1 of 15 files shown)


monperrus:
What about keeping only the getter `getLabelledStatement()`, put it `@Derived`, and look it up dynamically, as done for references?

tdurieux:
It's nit a so trivial task and this api is buggy since the beginning and nobody complain. I will not spend time on a feature that nobody wants

monperrus:
I understand, thanks for initiating the PR.

 @surli could you have a look at this, thanks?

surli:
>  @surli could you have a look at this, thanks?

You mean to review this PR or to implement it as a derived property?

monperrus:
to implement it as a derived property.

Thanks,

--Martin

tdurieux:
you can test this
```java

		label2:
		{
			System.out.println("block");
			if (args.length != 0 && args[0].equals("test")) {
				System.out.println("block after break3");
				break label2;
			}
			if (args.length ==1 && args[0].equals("test")) {
				System.out.println("block after break1");
				break label2;
			}
			System.out.println("block after break");
		}
		sw:
		switch ("f") {
		case "f":
			label:
			do {
				System.out.println("do");
				while (true) {
					for (int i = 0; i < 4; i++) {
						System.out.println("for");
						label2:
						{
							System.out.println("test");
							break label2;
						}
						System.out.println("for 1");
					}
					break;
				}
				break;
			} while (true);
			break;
		}
	}
```

monperrus:
> Add a new interface

I'm not sure about this. To be discussed.

monperrus:
thanks @tdurieux @surli

PR #1399, opened 2017-06-13
review: fix: mark implicitly created CtBlock as implicit=true

pvojtechovsky:


surli:
Added a unit test, can you review & merge @monperrus ?

monperrus:
I would prefer explicit because one sees it in the pretty-printed code (least surprise principle)
Why this change?

pvojtechovsky:
This java code
```java
for(Object o : list) o.toString();
```
is compiled into Spoon model which already contains implicit block
and which is pretty printed correctly as
```java
for(Object o : list) o.toString();
```
but I one tries to build the same using 
CtForEach.setBody(...statement...)
then s/he actually always get
```java
for(Object o : list) {
   o.toString();
}
```
because the implicitly created CtBlock wrapper is NOT marked as implicit. And I think it is bug. So I made this PR

monperrus:
I see your point. However, this is not a bug, IMHO.

The design rule for implicitness is that if it's in the original code, it's set as implicit.

However, when the API is used directly, one should always **explicitly** call setImplicit :-)

pvojtechovsky:
So close this PR. I can live with this wrong ;-) behavior, if you think it is correct.

PR #1400, opened 2017-06-13
fix: ignore null items in ReplacementVisitor

pvojtechovsky:
The ReplacementVisitor would fail if template parameter is null or an item of to be replaced collection is null.


surli:
The code seems ok but could you create a unit test to cover it?

pvojtechovsky:
Looks like most uncovered stuff does not comes from `null` but is there for long time...
I will probably do more tests till next Monday, will see how much time I will have.

pvojtechovsky:
I am done with tests. It is ready for merge from my point of view.

Issue #1337, opened 2017-05-30
Template does not replace content of string literal


pvojtechovsky:
Spoon Templates are really excellent and powerful tool!
But the substitution of variables in strings does not work.
```java
public class RequestBuilderTemplate extends ExtensionTemplate {

	/**
	 * Runs {@link API#$methodName$(Request)}
	 * @return {@link Response}
	 */
	public void run() {
		try {
			//do something
		} catch (Exception e) {
//-----------------------------------------------> the variables in this string are not substituted
			throw new RuntimeException("Execution of AnAPI#$methodName$ failed", e);
		}
	}
	@Parameter("AnAPI")
	CtTypeReference<?> apiType;
	@Parameter
	String $methodName$;
}
```

monperrus:
closed per #1403

PR #1403, opened 2017-06-20
review: feature: Substitute string literals

pvojtechovsky:
implements #1337

### DIFF
--- a/src/test/java/spoon/test/template/TemplateTest.java
+++ b/src/test/java/spoon/test/template/TemplateTest.java
@@ -688,13 +688,9 @@ public class TemplateTest {
 		spoon.buildModel();
 		Factory factory = spoon.getFactory();
 
-		CtClass<?> templateClass = factory.Class().get(SubstituteLiteralTemplate.class);
-		
 		{
 			//contract: String value is substituted in String literal
-			Map<String, Object> params = new HashMap<>();
-			params.put("$param1$", "value1") ;
-			final CtClass<?> result = new SubstitutionVisitor(factory, params).substitute(templateClass.clone()).get(0);
+			final CtClass<?> result = (CtClass<?>) new SubstituteLiteralTemplate("value1").apply(factory.createClass());
 			assertEquals("java.lang.String stringField1 = \"value1\";", result.getField("stringField1").toString());
 			assertEquals("java.lang.String stringField2 = \"Substring value1 is substituted too - value1\";", result.getField("stringField2").toString());
 			//contract: the parameter of type string replaces only method name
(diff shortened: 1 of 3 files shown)


src/test/java/spoon/test/template/TemplateTest.java
```diff
@@ -667,7 +665,12 @@ public void createTypeFromTemplate() throws Exception {
 		assertNull(genClass.getMethod("someMethod"));
 		assertTrue(generatedIfaceMethod!=generatedClassMethod);
 		assertTrue(generatedClassMethod.isOverriding(generatedIfaceMethod));
-		
+		//contract: check replace in literals
```
monperrus: what is the parameter (in `parameters`) driving this replacement?

pvojtechovsky: It is the parameter implicitly added by this code
```java
class Substitution {
...
	public static <T> CtType<T> createTypeFromTemplate(String qualifiedTypeName, CtType<T> templateOfType, Map<String, Object> templateParameters) {
		final Factory f = templateOfType.getFactory();
		CtTypeReference<T> typeRef = f.Type().createReference(qualifiedTypeName);
		CtPackage targetPackage = f.Package().getOrCreate(typeRef.getPackage().getSimpleName());
		final Map<String, Object> extendedParams = new HashMap<String, Object>(templateParameters);
//--------Here is the implicit parameter added----------------
		extendedParams.put(templateOfType.getSimpleName(), typeRef);
		List<CtType<T>> generated = new SubstitutionVisitor(f, extendedParams).substitute(templateOfType.clone());
```

monperrus: Implicit is worse than explicit. What if we move this code in the test? ie this responsibility to the caller?

I guess that string substitution should work  outside createTypeFromTemplate, consequently, could you add a test with a template dedicated to string substitution?

thanks

pvojtechovsky:
OK, you are right. 
I have added explicit tests of several combinations of string literal substitution and parameter value type.
There is also tested how behaves invocation replacement - it is tested in same test, to easily see the difference in behavior.
It is ready for merge. WDYT?

monperrus:
That's good thanks.

What I miss in SubstituteLiteralTemplate is that the template parameter are  implicit, we don't explicitly define a template parameter related to String and annotated with @Parameter. What do you think?

pvojtechovsky:
The Template parameter is declared explicitly now. OK? ;-)

monperrus:
thanks Pavel

Issue #1407, opened 2017-06-21
issue with isOverriding behavior when extending TypeParameters


surli:
If you consider the interface `Iterable<E>` which contain a method `forEach(Consumer<? super E>)` and if you create a class with a `T extends String` which implements `Iterable<T>`, then you implement your own version of `forEach`: if you use `ClassTypingContext.isSameSignature()` on the two `forEach` methods, it will return false because they do not have the same type parameter. 

I'm really not sure if this is a bug or not: in the overriding method, the type parameter is more defined, so the signature is not exactly the same. But then, if we only use `isSameSignature` in `getAllMethods` (see #1375) we will get both `Iterable#forEach` and `ArrayList#forEach`, when using it on `ArrayList`. 

WDYT? Should we consider the signature are the same, or should we consider that `getAllMethods` should indeed return the two different methods? 

Edit: Actually I extended my test to check with `isOverriding` and it returns false which is obviously wrong. So there's a real issue here.

pvojtechovsky:
> Should we consider the signature are the same
yes, there must be reported as same in this case. It is bug, which will be fixed in #1411

PR #1408, opened 2017-06-21
Expose issue with isOverriding #1407

surli:


surli:
PR take back by @pvojtechovsky #1411

PR #1411, opened 2017-06-21
review: Fix issue with isOverriding #1407

pvojtechovsky:
This PR is based on #1408 and should fix the problem of #1407.


pvojtechovsky:
There were two problems
1) java reflection did not created CtTypeParameterReference correctly. It made CtTypeReference instead
2) CtTypeParameterReference.getBoundingType() was adapted in wrong way...

pvojtechovsky:
There is another problem ... I have added test GenericsTest#testGetExecDeclarationOfEnumSetOf

<E extends Enum<E>> EnumSet#of(E)

has different depth of boundingTypes in their CtTypeParameterReference ...

Please have a look at this problem. I have no idea how to solve it...
I think it is not correct to have boundingType for CtTypeParameterReferences which  mirrors superType of it's TypeParameter declaration.

surli:
> I think it is not correct to have boundingType for CtTypeParameterReferences which mirrors superType of it's TypeParameter declaration.

I agree, I changed the convention to solve problems like `Enum<E extends Enum<E>>` to solve them keeping only `E -> Enum -> E` as typeParameterReference. 

However to fix your test I also had to change the `CtExecutableReference#getDeclaration` to use `getTypeDeclaration`. It changes some tests, but it also brings more information...

src/main/java/spoon/support/visitor/AbstractTypingContext.java
```diff
@@ -71,9 +71,11 @@ protected AbstractTypingContext() {
 			return adaptTypeParameterReferenceBoundingType(typeParamRef, typeParamRef.getBoundingType());
 		}
 		CtTypeReference<?> typeRefAdapted = adaptTypeParameter(typeParamRef.getDeclaration());
-		if (typeRefAdapted instanceof CtTypeParameterReference) {
-			return adaptTypeParameterReferenceBoundingType((CtTypeParameterReference) typeRefAdapted, typeParamRef.getBoundingType());
-		}
+		//Do not adapt bounding type of TypeParameterReference. That bounding type is already adapted as part of typeParamRef adaptation
```
monperrus: remove commented code

surli:
it's ok for you @pvojtechovsky ?

pvojtechovsky:
I will check it in the evening

src/main/java/spoon/support/reflect/reference/CtExecutableReferenceImpl.java
```diff
@@ -145,7 +146,16 @@ public boolean isConstructor() {
 	@SuppressWarnings("unchecked")
 	public CtExecutable<T> getDeclaration() {
 		final CtTypeReference<?> typeRef = getDeclaringType();
-		return typeRef == null ? null : getCtExecutable(typeRef.getDeclaration());
+		if (typeRef == null) {
```
surli: add a test `typeRef.getDeclaration() == null` (not in source path)

src/test/java/spoon/test/generics/GenericsTest.java
```diff
@@ -1094,4 +1097,39 @@ public void testGetDeclarationOfTypeParameterReference() {
...
+	public void testGetExecDeclarationOfEnumSetOf() {
+		Launcher launcher = new Launcher();
+		launcher.addInputResource("./src/test/java/spoon/test/generics/testclasses/EnumSetOf.java");
+		launcher.buildModel();
+
+		CtClass<?> ctClass = launcher.getFactory().Class().get(EnumSetOf.class);
+
+		CtInvocation invocation = ctClass.getMethodsByName("m").get(0).getBody().getStatement(0);
+		CtExecutable<?> decl = invocation.getExecutable().getDeclaration();
+		assertNotNull(decl);
```
surli: This assert should be null (getDeclaration without the source for java.util). Add a new assert with getExecutableDeclaration

src/test/java/spoon/test/generics/GenericsTest.java
```diff
@@ -1094,4 +1097,39 @@ public void testGetDeclarationOfTypeParameterReference() {
...
+		CtType<?> iterableItf = launcher.getFactory().Type().get(Iterable.class);
+
+		List<CtMethod<?>> methodsItf = iterableItf.getMethodsByName("forEach");
+		assertEquals(1, methodsItf.size());
+
+		ClassTypingContext ctc = new ClassTypingContext(ctClass.getReference());
+		assertTrue(ctc.isOverriding(methods.get(0), methodsItf.get(0)));
+		assertTrue(ctc.isSubSignature(methods.get(0), methodsItf.get(0)));
+
+		// assertTrue(ctc.isSameSignature(methods.get(0), methodsItf.get(0)));
```
surli: Remove commented code

pvojtechovsky:
It is mainly Simon's work. It should be merged under his name.
Thank you Simon for that effort.

monperrus:
thanks!

Issue #1415, opened 2017-06-23
NPE when using a union catch type inside a lambda not known in no classpath mode


surli:
Just spotted a really specific bug: 
when calling this kind of code in noclasspath mode, it crashes with NPE: 

```java
import java.io.IOException;
import java.io.Reader;
import java.io.StringReader;
import java.util.function.Function;
import toto.NbpOperator; // the important thing is here

public class UnionCatch {
    public void toto() {
        bla((NbpOperator) t -> {
                try {
                    Reader reader = new StringReader("machin");
                    int i = -1;
                    while (reader.ready()) {
                        i = reader.read();
                    }
                    return i;

               // and here
                } catch (IOException | NullPointerException e) {
                    System.out.printf("Error");
                    return 0;
                }
            });
    }
}
```

The bug does not occur with a simple catch. Neither it does outside a lambda. Or if the lambda is known in the classpath.

surli:
Fixed in #1416

PR #1416, opened 2017-06-23
review: Fix NPE with union catch inside lambda in noclasspath

surli:
This PR fix a NPE we had when spooning a catch with multi exception inside a lambda in no classpath mode, as explained in #1415. 
However the bug is not completely solved in that PR, in that sense that the model is not exactly the same as if the try catch was outside the lambda. I decided to concentrate on getting information from multi types (which was causing the NPE). But to complete it, we should also compute the type of CtCatchVariable from the multitypes (it's normally done by JDT). I opened an issue for that and I propose to solve it in a separate PR (see #1420)

### DIFF
--- a/src/test/java/spoon/test/exceptions/ExceptionTest.java
+++ b/src/test/java/spoon/test/exceptions/ExceptionTest.java
@@ -125,7 +125,7 @@ public class ExceptionTest {
 		assertEquals(variable1.getMultiTypes(), variable2.getMultiTypes());
 
 		// for now the type of CtCatchVariable is not the same
-		// this should be fix in the future (see:
+		// this should be fix in the future (see: https://github.com/INRIA/spoon/issues/1420)
 		//assertEquals(variable2, variable1);
 	}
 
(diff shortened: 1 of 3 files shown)


src/test/java/spoon/test/exceptions/ExceptionTest.java
```diff
@@ -101,4 +107,22 @@ public void testExceptionDuplicateClass() throws Exception {
 					.build();
 	}
 
+	@Test
+	public void testUnionCatchExceptionInsideLambdaInNoClasspath() {
+		// contract: the model should be built when defining a union catch inside a lambda which is not known (noclasspath)
+		// and the catch variable should be the same than outside a lambda
```
surli: I added this contract which seems reasonable, but I have problem to achieve this second part of the contract (and so, the last equals is a wrong assertion): JDT does not provide any information about the exception type inside the lambda, even not the fully qualified name of the exception. As a result, I have only the simple name of the class, which does not allow me easily to infer any information like on any other case. 

So, do you have any idea how to improve that, or could we remove this part of the contract and consider that in this specific case the model have less information?

PR #1417, opened 2017-06-23
feat(model): Model change listener

### LINKED ISSUE #1168: feature: Detection of model change

tdurieux:
Close #1168


src/main/java/spoon/support/reflect/code/CtCommentImpl.java
```diff
@@ -30,13 +33,21 @@
 	@MetamodelPropertyField(role = CtRole.COMMENT_TYPE)
 	private CommentType type;
 
+	public CtCommentImpl() {
+	}
+
+	protected CtCommentImpl(CommentType type) {
```
monperrus: remove?

monperrus:
rename `spoon.diff` in `spoon.experimental.modelobs`
put `ChangeFactory` in `spoon.experimental.modelobs`
add `spoon.experimental.package-info.java`:

    this is unstable code, subject to potential complete redesign, not only in snapshots but also in releases. you can use this code but you'll probably be impacted upon future changes. 

move addArgument to separate PR

monperrus:
split ChangeFactory into ChangeFactory (empty on*) and `ListeningChangeFactory`

src/main/java/spoon/support/reflect/code/CtCommentImpl.java
tdurieux: I need it for CtJavaDocImpl

tdurieux:
Done for me.

@pvojtechovsky Do you see something to improve?

pvojtechovsky:
Nice code! I would not do it better. It is OK for me. Sometime I would prefer more comments ;-)

I like that it is by default disabled and that it adds no overhead if it is not needed.
I like Contexts and Actions and fast and low level FineModelChangeListener and higher level ActionBasedChangeListener.

Here are some topics for further discussion:
* Why FineModelChangeListener is not an interface?
* The `CtExecutable#setParameters(List<CtParameter<?>> parameters)` calls listener `deleteAll` and then `add` ... I can imagine some use cases where firing event about remove and add on the same element might be a problem (the element (e.g. parameter) will be marked dirty even if it was not really touched). What about another FineModelChangeListener method `onListSet`, which might internally call deleteAll and add, but implementor have a chance to detect what elements were really removed/added by listening on `onListSet`. Or we might use new UpdateMultipleAction, where value would be a kind of Collection.
* All the collections returned by any model getter should be readonly in future. So clients are forced to use appropriate modification methods.

The topics above might be discussed/solved later - after this big PR is merged.

Anyway thank You for this effort!

tdurieux:
> I would prefer more comments ;-)

I know, I will try to add some

> Why FineModelChangeListener is not an interface?

Because it is really painful the create an interface when the things evolve a lot. We will create an interface when we will have some confidence in the design.

> The CtExecutable#setParameters(List<CtParameter<?>> parameters) calls listener deleteAll

I also don't like that and our replace uses these methods to make the change. The events are thus really awful. I made a clean method in the sniper PR to detect the delete all and the related add to keep only the relevant change.

> All the collections returned by any model getter should be readonly in future. So clients are forced to use appropriate modification methods.

That our final goal but it is really painful for the client.

monperrus:
We're almost done :-)

- remove addArgument in interface/impl
- undo change in return type of `removeType`, `removeArgument`

src/main/java/spoon/reflect/factory/Factory.java
```diff
@@ -156,6 +157,8 @@
 
 	QueryFactory Query();
 
+	FineModelChangeListener Change();
```
monperrus: remove this change? The entry point is now clearly in Environment. (Thanks revapi)

tdurieux:
Done for me

monperrus:
Thanks @GerardPaligot for the intial work on this. Kudos @tdurieux for the heavyweight rebase and design improvements.

pvojtechovsky:
Thanks @everybody for that great feature!

PR #1422, opened 2017-06-23
review feature: Add a new method to add a method argument to a specific position.

surli:


monperrus:
add a test?

surli:
> add a test?

Yep it's planned.

src/main/java/spoon/reflect/code/CtAbstractInvocation.java
```diff
@@ -49,6 +49,12 @@
 	<C extends CtAbstractInvocation<T>> C addArgument(CtExpression<?> argument);
 
 	/**
+	 * Adds an argument expression to the invocation at the specified position
+	 */
```
monperrus: Shifts the element currently at that position (if any) and any subsequent elements to the right (adds one to their indices).

src/test/java/spoon/test/invocations/InvocationTest.java
```diff
@@ -58,4 +62,32 @@ public void testTargetNullForStaticMethod() throws Exception {
...
+		CtExpression secondArg = invocBar.getArguments().get(1);
+
+		assertTrue(secondArg instanceof CtLiteral);
+
+		CtLiteral arg = (CtLiteral)secondArg;
+		assertEquals(42, arg.getValue());
+
+		secondArg = launcher.getFactory().Core().createLiteral().setValue(12);
+
+		invocBar.addArgument(1, secondArg);
```
monperrus: Interesting, this should fail with an exception because it's not compatible with the method signature.

tdurieux: I do not agree with you.
Spoon is not designed to check to validity of the transformation, it is not realistic to have always consistant model.

we can create a scanner that check the consistant of the model but it is not the responsibility of the setters.

surli: > Interesting, this should fail with an exception because it's not compatible with the method signature.

If you do that, we should do it in all methods, especially if we create `addParameter(int position, ...)` in CtExecutable: then you'll have an egg-chicken problem, if you use a transformation on the executable and on the invocations, there will be a "not compatible transformation" somewhere.

So I agree with Thomas: we should not do check there, but we should document the fact that those transformation are not checked against the executable.

monperrus:
> but it is not the responsibility of the setters.
I agree

> we should document the fact that those transformation are not checked against the executable.
Yes, starting by the Javadoc of this method.

surli:
> Yes, starting by the Javadoc of this method.

I changed the javadoc, but in fact we can do the same kind of warning for any `remove`, `add` or `set` method...

monperrus:
we close this one and remove this code from #1417 (and subsequently from #1358 by rebase)

PR #1423, opened 2017-06-23
review feature: Add a new method to insert a statement to a specific position

surli:


tdurieux:
Replace all `this.statements.add(0, statement);` by `this.insertAtPosition(0, statement)`

src/main/java/spoon/reflect/code/CtStatementList.java
```diff
@@ -48,6 +48,11 @@
 	<T extends CtStatementList> T addStatement(CtStatement statement);
 
 	/**
+	 * Inserts the given statement at a specific position in the list of statements
```
monperrus: Shifts the statement currently at that position (if any) and any subsequent statements to the right (adds one to their indices).

surli: Done.

PR #1428, opened 2017-06-25
review: refactor printing of lists in Spoon pretty printer

pvojtechovsky:
I am thinking of some enhancements of DefaultJavaPrettyPrinter and this would be the first BS.

@monperrus , do not merge this PR now, but please give me first a feedback whether you will merge the similar refactorings on all 16 places where `PrinterHelper#removeLastChar` is used.

Why to like it? ;-)
* it is shorter code - 6 lines instead of 11.
* there is less places for mistakes
* the PrinterHelper is not asked to print something and then to remove it - this bad approach makes future extensions of PrinterHelper more complicated...

Why to not like it?
* developer must know how the ListPrinter works to understand how it works. But then I think it is much more readable... so I hope it is not an issue.

WDYT? 

The suggestions for better names of new classes and methods are always welcome ;-)

### DIFF
--- a/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
+++ b/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
@@ -342,7 +342,6 @@ public class DefaultJavaPrettyPrinter implements CtVisitor, PrettyPrinter {
 				}
 			}
 		}
-		printer.writeln().writeTabs();
 	}
 
 	@Override
--- a/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
+++ b/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
@@ -1773,10 +1773,7 @@ public class DefaultJavaPrettyPrinter implements CtVisitor, PrettyPrinter {
 		elementPrinterHelper = new ElementPrinterHelper(printer, this, env);
 
 		elementPrinterHelper.writeComment(pack);
-
-		for (CtAnnotation<?> a : pack.getAnnotations()) {
-			a.accept(this);
-		}
+		elementPrinterHelper.writeAnnotations(pack);
 
 		if (!pack.isUnnamedPackage()) {
 			printer.write("package " + pack.getQualifiedName() + ";");
--- a/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
+++ b/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
@@ -109,6 +109,7 @@ import spoon.reflect.visitor.PrintingContext.Writable;
 import spoon.reflect.visitor.filter.PotentialVariableDeclarationFunction;
 import spoon.reflect.visitor.printer.CommentOffset;
 import spoon.reflect.visitor.printer.ElementPrinterHelper;
+import spoon.reflect.visitor.printer.ListPrinter;
 import spoon.reflect.visitor.printer.PrinterHelper;
 
 import java.lang.annotation.Annotation;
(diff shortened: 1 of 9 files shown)


src/main/java/spoon/reflect/visitor/printer/ListPrinter.java
```diff
@@ -0,0 +1,71 @@
...
+
+		if (start != null && start.length() > 0) {
+			printerHelper.write(start);
+		}
+	}
+
+	/**
+	 * Call that before printing of list item starts
+	 */
+	public void itemStart() {
```
monperrus: call it printNextSeparator?

monperrus:
It seems reasonable to me.

monperrus:
>  do not merge this PR now

then WIP it :-)

src/main/java/spoon/reflect/visitor/printer/ListPrinter.java
```diff
@@ -0,0 +1,71 @@
...
+		if (isFirst) {
+			/*
+			 * we are starting first item. Do not print `next` separator yet
+			 */
+			isFirst = false;
```
surli: Why not printing the start token here? It would be more clear to me.

src/main/java/spoon/reflect/visitor/printer/ListPrinter.java
pvojtechovsky: Note, that this method is called before the item is generated, not after. And that separator is printed only when needed. The "printNextSeparator" sounds like separator is printed always.
Anyway I agree that my method name needs an improvement...

src/main/java/spoon/reflect/visitor/printer/ListPrinter.java
pvojtechovsky: Because in case when there is no item in the list, the start and end tokens must be printed too.
With your change, the start would be printed only for non empty lists

monperrus:
printSeparatorIfAppropriate?

src/main/java/spoon/reflect/visitor/printer/ListPrinter.java
surli: I don't know the other usage of this class, but here it's used after a check on the size of the arguments, I assumed it would be the same afterwards.

pvojtechovsky: Now I see it too, that there is `arguments.size() > 0`, so in this case your solution would work too. But it is rather a specific case. Usually, there is no such test and start/end have to be printed always.

pvojtechovsky:
> printSeparatorIfAppropriate
ok. I will do so

src/main/java/spoon/reflect/visitor/printer/ListPrinter.java
surli: Ok then

pvojtechovsky:
The new line after annotation is now not printed by `DefaultJavaPrettyPrinter#visitCtAnnotation`, but it is printed by `ElementPrintHelper#writeAnnotations`, because in the first case there are places where annotation was alone and the new line had to be deleted again. And in second case we can be sure that annotations have to be separated by EOL.

pvojtechovsky:
I have printed Guava sources from Spoon model using main line and using this PR and it produces same results. So it looks like this code is OK. It is ready for merge from my point of view.

src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
```diff
@@ -333,16 +334,14 @@ public String toString() {
...
+				for (Entry<String, CtExpression> e : annotation.getValues().entrySet()) {
+					lp.printSeparatorIfAppropriate();
+					printer.write(e.getKey() + " = ");
+					elementPrinterHelper.writeAnnotationElement(annotation.getFactory(), e.getValue());
+				}
 			}
-			printer.removeLastChar();
-			printer.write(")");
 		}
-		printer.writeln().writeTabs();
```
surli: Looking on your test, and with this line deleted it means now Spoon does not print anymore one annotation per line when there's multiple annotations?

src/test/java/spoon/test/filters/FilterTest.java
```diff
@@ -96,7 +96,7 @@ public void testFilters() throws Exception {
 		CtClass<?> foo = factory.Package().get("spoon.test.filters").getType("Foo");
 		assertEquals("Foo", foo.getSimpleName());
 		List<CtExpression<?>> expressions = foo.getElements(new RegexFilter<CtExpression<?>>(".* = .*"));
-		assertEquals(2, expressions.size());
+		assertEquals(4, expressions.size());
```
surli: Why this one changed?

src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
surli: Ok I just saw your comment and the fact that tests are presented only one annotation and not multiple ones, so don't consider my comment ;)

pvojtechovsky:
Because reg exp `.* ` by default does not matches EOL `\n` and therefore it did not matched the single line annotations which ended with useless EOL. Now the EOL is not there and therefore the filter returns more occurences

src/test/java/spoon/test/filters/FilterTest.java
pvojtechovsky: Because reg exp `.*` by default does not matches EOL. So it did not matched single line annotations which ended with EOL. Now it matches them, because there is no useless EOL at the end

surli: Ok thanks for the explanation :)

PR #1429, opened 2017-06-26
improve explanation of getDeclaration

monperrus:


msteinbeck:
I think it should be emphasized that getDeclaration should be used in noclasspath mode. The sentence: 

> It is strongly advised to use the more robust {@link CtTypeReference#getTypeDeclaration()} and {@link CtExecutableReference#getExecutableDeclaration()} that never return null" 

may be misleading here.

PR #1434, opened 2017-06-26
review: feat: sort imports alphabetically

pvojtechovsky:
The imports generated by Spoon pretty printer are sorted alphabetically now.

The main advantage is that Spoon produces same content with each printing now. Before the imports were randomly mixed (because of HashSet).

### DIFF
--- a/src/main/java/spoon/reflect/visitor/printer/ElementPrinterHelper.java
+++ b/src/main/java/spoon/reflect/visitor/printer/ElementPrinterHelper.java
@@ -53,6 +53,7 @@ import spoon.reflect.visitor.PrintingContext.Writable;
 
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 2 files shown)


monperrus:
good idea! we need a test with the corresponding contract, thanks.

tdurieux:
Personally I would separate the static import from the normal one.
Put all the static import at the beginning or at the end of the import list.

pvojtechovsky:
> Personally I would separate the static import from the normal one.

I agree that it can be done better, but I do not want to invest more time in it. The sorting was enough for now for me. If somebody wants it better, I will like it ;-)

surli:
@pvojtechovsky I created another issue for @tdurieux idea. Could you add a test for this one, then I'll merged it.

pvojtechovsky:
Test added. I am finished here.

PR #1435, opened 2017-06-26
fix printing of comments on MS Windows

pvojtechovsky:
The comments were printed wrong on MS Windows, when java source lines are separated by "\n" (it is case of sources which are checked out from GIT), while MS Windows EOL is "\n\r".

The new algorithm should split lines of comments independent on current OS.

### DIFF
--- a/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
+++ b/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
@@ -860,7 +860,7 @@ public class DefaultJavaPrettyPrinter implements CtVisitor, PrettyPrinter {
 			printer.write(docTag.getParam()).writeln().writeTabs();
 		}
 
-		String[] tagLines = docTag.getContent().split(LINE_SEPARATOR);
+		String[] tagLines = docTag.getContent().split("\\n\\r|\\n|\\r");
 		for (int i = 0; i < tagLines.length; i++) {
 			String com = tagLines[i];
 			if (i > 0 || docTag.getType().hasParam()) {
@@ -898,7 +898,7 @@ public class DefaultJavaPrettyPrinter implements CtVisitor, PrettyPrinter {
 				printer.write(content);
 				break;
 			default:
-				String[] lines = content.split(LINE_SEPARATOR);
+				String[] lines = content.split("\\n\\r|\\n|\\r");
 				for (int i = 0; i < lines.length; i++) {
 					String com = lines[i];
 					if (comment.getCommentType() == CtComment.CommentType.BLOCK) {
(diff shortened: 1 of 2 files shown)


src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
```diff
@@ -861,7 +861,7 @@ public void visitCtJavaDocTag(CtJavaDocTag docTag) {
 			printer.write(docTag.getParam()).writeln().writeTabs();
 		}
 
-		String[] tagLines = docTag.getContent().split(LINE_SEPARATOR);
+		String[] tagLines = docTag.getContent().split("\\n\\r|\\n|\\r");
```
surli: Create a constant for that one, it could be useful later.

src/test/java/spoon/test/comment/CommentTest.java
```diff
@@ -110,12 +111,37 @@ public void testCombinedPackageInfoComment() {
...
+		//that is true on Windows too.
+		String EOL = "\n";
 
 		Factory f = getSpoonFactory();
 		CtClass<?> type = (CtClass<?>) f.Type().get(JavaDocComment.class);
 
 		CtJavaDoc classJavaDoc = (CtJavaDoc) type.getComments().get(0);
+		//contract: test that java doc is printed correctly
+		String str = classJavaDoc.toString();
+		StringTokenizer st = new StringTokenizer(str, "\n\r");
```
surli: You should use a `System.getProperty("line.separator")` here for the StringTokenizer delimiter, no?

pvojtechovsky: no, because I want to split string on each occurence of `\n` and on each occurence of `\r`. See javadoc of StringTokenizer for details

surli: Ok for the behaviour of StringTokenizer but I'm not sure to understand why you want to do that here. If you're on linux, you shouldn't have `\r` at all, so using `System.getProperty("line.separator")` would be safer, and if you're on windows it will return `\n\r` so you're test remains the same. In either case, it seems safer to me. What am I missing?

pvojtechovsky: Even on Windows there are situations when EOL is \n only
And even on Linux it can happen that you open file which contains \n\r as EOL.
In case of java sources checked out from repository it happens quite often

surli: > n case of java sources checked out from repository it happens quite often

Yes but here you create a tokenizer only for the source code generated by Spoon, so it should not happen.

pvojtechovsky:
The issues found by Simon were solved. I am finished here.

PR #1437, opened 2017-06-27
review: Fix NPE with import scanner #1436

surli:


monperrus:
merged without test because we're unable to isolate the bug.

thanks @tdurieux @surli

PR #1442, opened 2017-06-28
review: fix null comment content

pvojtechovsky:
Test is needed...


pvojtechovsky:
The CtComment.getComment() returned null in some cases... it is a bug, which is fixed by this PR

monperrus:
thanks

PR #1448, opened 2017-06-28
review: fix: getFactory should never return null

### LINKED ISSUE #1446: fix: getFactory could be null if an element of the model is created without using a factory
This could now throw NPE with change of ModelListener which use intensively getFactory(). 
Simple fix: create a new default factory inside getFactory if it's null: `new FactoryImpl(new DefaultCoreFactory(),new StandardEnvironment());`

surli:
Fix #1446


monperrus:
we need a checkstyle automatic repair :-)

surli:
> we need a checkstyle automatic repair :-)

Indeed...

PR #1449, opened 2017-06-29
doc:explicit the difference between isOverriding and isSubsignature

surli:


src/main/java/spoon/support/visitor/ClassTypingContext.java
```diff
@@ -318,26 +323,32 @@ public boolean isOverriding(CtMethod<?> thisMethod, CtMethod<?> thatMethod) {
 			}
 		}
 		//TODO check method visibility following https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.4.8.1
-		return isSubSignature(thisMethod, thatMethod);
+		return isSubSignature(thisMethod, thatMethod) || isSubSignature(thatMethod, thisMethod);
```
pvojtechovsky: Not that, because in our case we can be sure that thisMethod belongs to sub class and thatMethod belongs to super class. So origin code is OK

PR #1451, opened 2017-06-29
review: fix NPE in Pretty Printer on missing variable declaration

pvojtechovsky:
If the pretty printer fails now, It throws and SpoonException with position of the element, which caused problem now.
It helps to fix the problem without need to debug printer.


surli:
Nice PR, it will be useful indeed :)

PR #1452, opened 2017-06-29
feature CtTypeMember extends CtNamedElement

pvojtechovsky:
WDYT?

### DIFF
--- a/src/main/java/spoon/reflect/declaration/CtTypeMember.java
+++ b/src/main/java/spoon/reflect/declaration/CtTypeMember.java
@@ -22,7 +22,7 @@ import spoon.support.DerivedProperty;
  * This interface represents a member of a class (field, method,
  * nested class or static/instance initializer).
  */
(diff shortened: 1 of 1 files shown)


surli:
It's specified in the javadoc : 
> nested class or static/instance initializer).

Am I wrong, or a static initializer is an anonymous block?

pvojtechovsky:
You are right, but CtAnonymousExecutable already implements CtNamedElement and returns "" for the name. So this change is breaking nothing and helps clients (me ;-) ) to query type members by name without casting...

surli:
Hmm, actually I wonder why CtExecutable implements CtNamedElement as many executables are not named (lambda, anonymous executables, ...). @monperrus ?

pvojtechovsky:
I vote to keep it as it is - it is may be not 100% correct from java model point of view, but I think it is good for clients.

monperrus:
OK for me.

PR #1455, opened 2017-06-30
review: fix splitting of comment by EOL

pvojtechovsky:
The RegExp for splitting of comments by EOL was wrong. MS Windows is using `\r\n` and not `\n\r` like it was wrong before

### DIFF
--- a/src/main/java/spoon/reflect/code/CtComment.java
+++ b/src/main/java/spoon/reflect/code/CtComment.java
@@ -52,6 +52,13 @@ public interface CtComment extends CtStatement {
 	}
 
 	/**
+	 * This line separator is used in comments returned by {@link #getContent()}.
+	 * It is OS independent.
+	 * It has no influence to pretty printed comments, which uses by default OS dependent line separator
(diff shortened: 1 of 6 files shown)


surli:
Can you create a test with a resource with windows EOL to ensure it in the future?

pvojtechovsky:
I started with writing of test, but I failed because sometime spoon uses `\n` on Windows and sometime `\r\n` ... Please define some rules how it should work in #1460 and then we can finish test of this one too.

pvojtechovsky:
1) The lines are not detected by BufferedReader#readLine, so it is correct for all OSes.

2) The comment cleaning algorithm was rewritten to handle correctly wild cases like

```java
///* this single line comment should start with /* and end with */
```
see new tests in `WildComments`

3) The CtComment#LINE_SEPARATOR explicitly states that CtComment#getContent() lines are always separated by "\n" - independent on OS. - It is legacy behavior. I have just made it visible.

It is ready for merge, if you agree with all these concepts.

PR #1457, opened 2017-06-30
fix: clarifies the contract of proxy template parameters

monperrus:
@pvojtechovsky in #1444 has uncovered foggy parts in the template engine.

this PR is a first clarification about proxy template parameters (those with a value, eg `@Parameter("$name$")`.

### DIFF
--- a/src/main/java/spoon/template/Substitution.java
+++ b/src/main/java/spoon/template/Substitution.java
@@ -669,15 +669,30 @@ public abstract class Substitution {
 
 	private static <T> void checkTemplateContracts(CtClass<T> c) {
 		for (CtField f : c.getFields()) {
-			if (f.getAnnotation(Parameter.class) != null && !f.getAnnotation(Parameter.class).value().equals("")) {
+			Parameter templateParamAnnotation = f.getAnnotation(Parameter.class);
+			if (templateParamAnnotation != null && !templateParamAnnotation.value().equals("")) {
+				String proxyName = templateParamAnnotation.value();
 				// contract: if value, then the field type must be String
 				if (!f.getType().equals(c.getFactory().Type().STRING)) {
 					throw new TemplateException("proxy template parameter must be typed as String " +  f.getType().getQualifiedName());
 				}
 
 				// contract: the name of the template parameter must correspond to the name of the field
-				if (!f.getSimpleName().equals("_" + f.getAnnotation(Parameter.class).value())) {
-					throw new TemplateException("the field name of a proxy template parameter must be called _" + f.getSimpleName());
+				// as found, by Pavel, this is not good contract because it prevents easy refactoring of templates
+				// we remove it but keep th commented code in case somebody would come up with this bad idae
+//				if (!f.getSimpleName().equals("_" + f.getAnnotation(Parameter.class).value())) {
+//					throw new TemplateException("the field name of a proxy template parameter must be called _" + f.getSimpleName());
+//				}
+
+				// contract: if a proxy parameter is declared and named "x" (@Parameter("x")), then a type member named "x" must exist.
+				boolean found=false;
+				for(CtTypeMember member: c.getTypeMembers()) {
+					if (member.getSimpleName().equals(proxyName)) {
+						found = true;
+					}
+				}
+				if (!found) {
+					throw new TemplateException("if a proxy parameter is declared and named \"x\", then a type member named \"x\" must exist.");
 				}
 
 			}
--- a/src/main/java/spoon/template/Substitution.java
+++ b/src/main/java/spoon/template/Substitution.java
@@ -692,7 +692,7 @@ public abstract class Substitution {
 					}
 				}
 				if (!found) {
-					throw new TemplateException("if a proxy parameter is declared and named \"x\", then a type member named \"x\" must exist.");
+					throw new TemplateException("if a proxy parameter is declared and named \""+proxyName+"\", then a type member named \"\"+proxyName+\"\" must exist.");
 				}
 
 			}
--- a/src/main/java/spoon/template/Substitution.java
+++ b/src/main/java/spoon/template/Substitution.java
@@ -663,7 +663,25 @@ public abstract class Substitution {
 		if (c.isShadow()) {
 			throw new SpoonException("The template " + template.getClass().getName() + " is not part of model. Add template sources to spoon template path.");
 		}
+		checkTemplateContracts(c);
 		return c;
+	}
+
+	private static <T> void checkTemplateContracts(CtClass<T> c) {
+		for (CtField f : c.getFields()) {
+			if (f.getAnnotation(Parameter.class) != null && !f.getAnnotation(Parameter.class).value().equals("")) {
+				// contract: if value, then the field type must be String
+				if (!f.getType().equals(c.getFactory().Type().STRING)) {
+					throw new TemplateException("proxy template parameter must be typed as String " +  f.getType().getQualifiedName());
+				}
+
+				// contract: the name of the template parameter must correspond to the name of the field
+				if (!f.getSimpleName().equals("_" + f.getAnnotation(Parameter.class).value())) {
+					throw new TemplateException("the field name of a proxy template parameter must be called _" + f.getSimpleName());
+				}
+
+			}
+		}
 	}
 
 	/**
--- a/src/main/java/spoon/template/Substitution.java
+++ b/src/main/java/spoon/template/Substitution.java
@@ -685,14 +685,14 @@ public abstract class Substitution {
 //				}
 
 				// contract: if a proxy parameter is declared and named "x" (@Parameter("x")), then a type member named "x" must exist.
-				boolean found=false;
-				for(CtTypeMember member: c.getTypeMembers()) {
+				boolean found = false;
+				for (CtTypeMember member: c.getTypeMembers()) {
 					if (member.getSimpleName().equals(proxyName)) {
 						found = true;
 					}
 				}
 				if (!found) {
-					throw new TemplateException("if a proxy parameter is declared and named \""+proxyName+"\", then a type member named \"\"+proxyName+\"\" must exist.");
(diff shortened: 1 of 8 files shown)


pvojtechovsky:
I have no time today to debug it. But after furst reading, I understood that you would like to introduce some rules between proxy field name and proxy template parameter name. Why that? What it is hood for? Thanks for explanation ;)

monperrus:
The rules already existed, but somehow implicitly, only with conventions.

Making them explicit avoids mis-usages or overly complex usages of template parameters.

In short, never define a template parameter with a proxy name if you can avoid it, because it makes
code much harder to understand (actually a number of test cases could be simplified this way).

This PR adds dynamic verification code to check this.

src/test/java/spoon/test/template/testclasses/InvocationTemplate.java
```diff
@@ -15,15 +15,15 @@ void invoke() {
 	
 	@Local
 	public InvocationTemplate(CtTypeReference<?> ifaceType, String methodName) {
-		this.ifaceType = ifaceType;
-		this.methodName = methodName;
+		this._IFace = ifaceType.getSimpleName();
```
pvojtechovsky: This is wrong, because ifaceType usually comes from different package then template and the package is lost at this place. You cannot replace parameters with type CtTypeReference by type String.

src/test/java/spoon/test/template/testclasses/EnumAccessTemplate.java
```diff
@@ -13,15 +12,15 @@ public void method() throws Throwable {
 	}
 	
 	@Parameter("_AnEnum_")
-	CtTypeReference<?> anEnum;
+	String __AnEnum_;
```
pvojtechovsky: Yesterday I refactored my templates a lot and I have found a way how to write them in "refactoring friendly" way. I am using this:
```java
@Parameter(_AnEnum_.class.getSimpleName())
CtTypeReference<?> anEnum;
```
This PR actually kills this usefull approach.

src/test/java/spoon/test/template/testclasses/InvocationTemplate.java
monperrus: I would disagree. Because the essence of what we do is to rename the nested Interface IFace, and for this we only need a name.

Otherwise, the semantics is unclear. Suppose we have `spoon.test.template.testclasses.InvocationTemplate.IFace` and we want to replace by `a.b.c.Foo`: should the result be `spoon.test.template.testclasses.InvocationTemplate.Foo` (current behavior) or `spoon.test.template.testclasses.InvocationTemplate.a.b.c.Foo`?

AFAIK, the package is information is useless.

src/test/java/spoon/test/template/testclasses/EnumAccessTemplate.java
monperrus: Super interesting, you're right. Let's relax the contract and remove the naming convention.

monperrus:
Last point: there is probably another contract to add:

if a proxy parameter is declared and named x `@Parameter("x")`, then a type member named "x" must exist.

This is refactoring-proof. 

WDYT?

src/test/java/spoon/test/template/testclasses/EnumAccessTemplate.java
pvojtechovsky: ok, in fact I did not used it exactly like in example above, because it is probably not compilable, but I still argue that new contract between field name and parameter vale is not good because:
* it makes the refactoring of templates harder, because it is not enough to rename parameter value, but one must rename field and all field references too.
* The class name in parameter value and field name might represent different meaning and then it is jot possible to use selfexplanatory field name, but one must use field name which is based on class name.

pvojtechovsky: Try to compile it. If it pass then relax contract. If it is not compilable then keep contract as it is. My usage, which I show you soon, is not with conflict with this PR so it is ok.

src/test/java/spoon/test/template/testclasses/InvocationTemplate.java
pvojtechovsky: The result must be `a.b.c.Foo`. Note that IFace is (or should be) marked @Local and is here to represent arbitrary interface of the java world.

pvojtechovsky:
I have understood the basic idea of this PR - to avoid usage of Parameter#value if it is not necessay, because it makes it more readable - this idea is ok for me.

But the replacing of CtTypeReferences with Strings is different topic, which is not related to this PR.

So I like this PR but please do not remove tests - see comments below

src/test/java/spoon/test/template/TemplateTest.java
```diff
@@ -743,48 +743,5 @@ public void substituteSubString() throws Exception {
...
-		{
-			//contract: The elements which cannot be converted to String should throw exception
-			SubStringTemplate template = new SubStringTemplate(factory.createSwitch());
-			try {
-				template.apply(factory.createClass());
-				fail();
-			} catch(SpoonException e) {
-				//OK
-			}
-		}
```
pvojtechovsky: Do not remove all that stuff. It tests behavior of SubstitutionVisitor depending of type of value of paramater. It is not related to your changes. It can be kept with following change - see comment below.

src/test/java/spoon/test/template/testclasses/SubStringTemplate.java
```diff
@@ -16,11 +16,11 @@ void m() {
 		set$name$("The $name$ is here too");
 	}
 	
-	@Parameter("$name$")
-	Object name;
+	@Parameter
+	String $name$;
```
pvojtechovsky: Use `Object $name$;`

src/test/java/spoon/test/template/testclasses/InvocationTemplate.java
pvojtechovsky: We both are right and spoon supports both ways.
A) if one needs to rename type the s/he use parameter value of type String
B) if one needs to substitute type by another type, then s/he use CtTypeReference

src/test/java/spoon/test/template/testclasses/InvocationTemplate.java
```diff
@@ -15,15 +15,15 @@ void invoke() {
...
 	public InvocationTemplate(CtTypeReference<?> ifaceType, String methodName) {
-		this.ifaceType = ifaceType;
-		this.methodName = methodName;
+		this._IFace = ifaceType.getSimpleName();
+		this._$method$ = methodName;
 	}
 
 	@Parameter("IFace")
-	CtTypeReference<?> ifaceType;
+	String _IFace;
```
pvojtechovsky: So please use

@Parameter
CtTypeReference<?> IFace;

because otherwise it generates not compilable type - note that nested interface is already marked @Local.

pvojtechovsky: and here I see that field name IFace doeas not start lowercase - java conventions - so may be we should realy relax the contracts of this PR? It is not a problem for me.

src/main/java/spoon/template/Substitution.java
```diff
@@ -663,9 +663,27 @@ public static void redirectTypeReferences(CtElement element, CtTypeReference<?>
 		if (c.isShadow()) {
 			throw new SpoonException("The template " + template.getClass().getName() + " is not part of model. Add template sources to spoon template path.");
 		}
+		checkTemplateContracts(c);
 		return c;
 	}
 
+	private static <T> void checkTemplateContracts(CtClass<T> c) {
+		for (CtField f : c.getFields()) {
+			if (f.getAnnotation(Parameter.class) != null && !f.getAnnotation(Parameter.class).value().equals("")) {
```
tdurieux: Please don't use f.getAnnotation(Class) several times in a row.
This API uses reflection and proxy, the performances are thus not good at all.

monperrus: done

src/test/java/spoon/test/template/TemplateTest.java
monperrus: you're right. done

monperrus:
Done last changes.
If you now both agree, @tdurieux you can proceed with merge and we can come back to #1444

src/main/java/spoon/template/Substitution.java
```diff
@@ -663,9 +663,42 @@ public static void redirectTypeReferences(CtElement element, CtTypeReference<?>
 		if (c.isShadow()) {
 			throw new SpoonException("The template " + template.getClass().getName() + " is not part of model. Add template sources to spoon template path.");
 		}
+		checkTemplateContracts(c);
 		return c;
 	}
 
+	private static <T> void checkTemplateContracts(CtClass<T> c) {
+		for (CtField f : c.getFields()) {
+			Parameter templateParamAnnotation = f.getAnnotation(Parameter.class);
```
pvojtechovsky: move it one line above?

pvojtechovsky:
I am still in doubt whether this PR is good idea at all. Will the clients apprechiate it or hate it because they have to change their templates ... just because we think that they should design it in nicer way?

What about some switch which allows to relax this constraint in case it is not compatible with legacy templates? ... I personaly do not have this problem, because I started with templates just few days ago, but I can imagine clients with many templates which will hate this PR ...

src/main/java/spoon/template/Substitution.java
pvojtechovsky: ignore this comment, I overlook something

monperrus:
As far we know, there are no client of templates .... but you :-). You're the very first!
(proof: the absence of issues and PR on this part over the last years, while it's little documented and most probably buggy :-) 

Seriously, the template engine is a very old yet still experimental feature, as illustrated by the recent PRs on this topic, incl. yours.

To this extent, it's better to clean the design and contracts now, as done in this PR.

pvojtechovsky:
OK ;

tdurieux:
@pvojtechovsky @monperrus Is it Ok for everyone?

monperrus:
OK for me.

pvojtechovsky:
ok for me

Issue #1458, opened 2017-07-01
feature: add fluent API for templates


pvojtechovsky:
```java
class _Builder_ extends AbstractBuilder implements ThisIFace, ThatIFace {
  private  _Proxy_ delegate;
  public _Builder_(_Proxy_ proxy) {
    this.proxy = proxy;
  }
  @Local
  public _Builder_ _propertyName_(_PropertyType_ _propertyName_) {
     delegate._setterName_(_propertyName_);
     return this;
  }
  @Local
  public static CtClass<?> createFluentBuilder(String builderQName, CtTypeReference<?> proxyType) {
      return new SubstitutionRequest(proxyType.getFactory(), _Builder_ .class)
          .excludeAllMembersWithAnnotation(Local.class)
          .substituteClass(_Proxy_ .class, proxyType)
          .createType(builderQName)
  }
  @Local
  public static List<CtTypeMember> addFluentSetter(CtClass<?> builder, CtMethod<?> setterMethod) {
      return new SubstitutionRequest(setterMethod.getFactory(), _Builder_ .class)
          .excludeAll().includeMethod("_propertyName_")
          .substituteClass(_Proxy_ .class, setterMethod.getDeclaringType())
          .substituteName("_setterName_", setterMethod.getSimpleName())
          .substituteName("_propertyName", Utils.lowerCaseFirst(setterMethod.getSimpleName().substring(3)))
          .applyToType(builder)
  }
}
```
and usage
```java
CtClass<?> newBuilderClass = _Builder_ .createFluentBuilder("a.b.c.FluentSetterForX", typeRefToTypeX);
for(CtMethod m : Utils.getAllSetters(typeRefToTypeX)) {
 _Builder_.addFluentSetter(newBuilderClass, m);
}
```
WDYT?

surli:
I like the idea of a more explicit API for substitution inside the template. However in this example, why the `Local` annotation are not excluded by default? And why do you use one on `_propertyName_`? It is a mistake or it is on purpose?

pvojtechovsky:
> why the Local annotation are not excluded by default? And why do you use one on _propertyName_? It is a mistake or it is on purpose?

It is on purpose. Note that type `_Builder_` defines 2 templates! (Not one like it was always till now!). First template constructed by method `createFluentBuilder` is used to generate target class itself. 
The second template constructed by method `addFluentSetter` is used to generate one method into existing (before generated) type.

[...]

PR #1463, opened 2017-07-02
WiP: feature TemplateBuilder

pvojtechovsky:
A new way how to create a template from any type or any part of type. See example in #1458.

Features:
* any type can be used as source of the template
* the part of the source type can be excluded/ included in template by `excludeXxx` and `includeXxx` methods. It allows to define one Template type and use it for more related template substitutions. For example 1) create the build type, 2) add all builder methods. Both steps based on same template type, but using different type members.
* the new type can be created from template
* the template can be applied to existing type
* substitution parameters are wrapped into `substituteXxx` methods, which allows more user friendly definition of template parameters.

This code is experimental I will play with it first and I expect it will be finished in 1 month.


pvojtechovsky:
The ideas from this PR are in #1686

PR #1476, opened 2017-07-15
review: fix replacing of field access by relaxed Parameter value contract

pvojtechovsky:
This is the alternative to the #1444 with 
* relaxed Template parameter constraint
* early detection of string literal parameter value applied to each Parameter without value


pvojtechovsky:
Travis CI is broken...

monperrus:
Thanks Pavel.
I don't understand the goal of this PR. The title of the PR suggests it's a fix. What is the wrong behavior and why?

pvojtechovsky:
> what is wrong behavior?

This PR is alternative solution to #1444. The problems are described there.

Short description: The proxy template parameter is needed not only in case of name conflict with existing field name, but is needed
* for field name in nested class (see new test case in#1444)
* for case when client wants to rename method name of method invocation - there is already older test for that.

Therefore I suggest to release the template parameter constraint you have introduced. It is the change introduced by this PR.
Optionally you can try to solve that problem represented by actually failing test in #1444 by another way - continue with #1444

monperrus:
I see, I have to deeply understand the problem of #1444 then. However, I'm traveling, it may require some time I don't have yet.

pvojtechovsky:
Yes, the deeper understanding is needed.

These are the concepts which have to be understood:

C1) Template and SubstitutionVisitor are no more deeply bound together. The only binding is in constructor of SubstitutionVisitor, where Template is converted to `substitution parameters` represented by Map<String, Object>.

C2) If the `substitution parameter` value is CtLiteral of type String then parameter CtFieldReference is replaced by that CtLiteral. So the substitution produces a string literal.

C3) If the `substitution parameter` value is String then simple name of the parameter CtFieldReference is substituted. So the substitution produces a field reference to field whose name is defined by template parameter.

These 3 concepts were origin target of the PR #1444. But then Simon came with interesting idea (commit [Change SubstitutionVisitor to treat differently case A and case B](https://github.com/INRIA/spoon/pull/1444/commits/b1a3aba6db7bf495ca923e85ef628b5f33feb789)) (if I understood him well) to make my solution more backward compatible.
The Simon's solution has these problems: 
* it does not follows clear rules C2 and C3, but tries to detect intention of the Template designer by detecting how the field reference is used in generated code. 
* it does not work for field references in inner classes. See commit named "test: field access in inner class"

Therefore I suggested to detect intention of the Template designer sooner - during extraction of `substitution parameters` from the Template (C1). See commit [convert String to Literal parameter value automatically](https://github.com/INRIA/spoon/pull/1476/commits/aa422baef4fdcddd786afdbf7d5fe2fb609d012a). This solution is compatible with the concepts C2 and C3. 

It introduces concept

C4) If the template parameter has a proxy name, then it is not substituted as String literal, but it is substituted in a simple name of the field/executable reference.

Concept C4 can be used for field parameter in Template, but it cannot be used for field parameter in inner class (See TemplateTest#testFieldAccessNameSubstitutionInInnerClass) and it cannot be used for replacing of method invocation names (See TemplateTest#testTemplateInvocationSubstitution)
because it is not compatible with Template parameter rule in Substition#checkTemplateContracts(). Therefore I disabled this rule in commit "disable one Substitution#checkTemplateContracts contract"

Final I adapted the tests InvocationTemplate and FieldAccessOfInnerClassTemplate to use C4.

I hope this PR is understandable now and we can continue here.

pvojtechovsky:
I need C1, C2, C3. I base my templates on these rules.

I am not sure concerning backward compatibility introduced by Simon and I am not sure whether C4 is the understandable solution. Do you see a better one?

src/main/java/spoon/template/Substitution.java
```diff
@@ -684,15 +684,16 @@ public static void redirectTypeReferences(CtElement element, CtTypeReference<?>
...
-				boolean found = false;
-				for (CtTypeMember member: c.getTypeMembers()) {
-					if (member.getSimpleName().equals(proxyName)) {
-						found = true;
-					}
-				}
-				if (!found) {
-					throw new TemplateException("if a proxy parameter is declared and named \"" + proxyName + "\", then a type member named \"\" + proxyName + \"\" must exist.");
-				}
+				// it is not good contract, because fields of inner classes and method invocations cannot be handled then
```
pvojtechovsky: I know what you mean, but the reason why I need to relax this constrain is not the name conflict but the ability to tell Spoon that I do not want to replace string literal but I want to replace simple name of the field or method invocation.
Try to rollback this [change](https://github.com/INRIA/spoon/pull/1476/commits/8ee34001108a310ed04ee0070417fee95e752df2) using code of this PR and then run that test. It will fail and if you try to fix it, then you understand what I mean.
The code
```java
void invoke() {
 	iface.$method$();
 }
```
is not replace to 
```java
void invoke() {
 	iface.toBeInvokedMethod();
 }
```
but is replaced with 
```java
void invoke() {
 	"toBeInvokedMethod";
 }
```
and the last one is correct from C2, C3 point of view, because the parameter value is CtLiteral. So I need to force the system to keep parameter value of type String! But how if the constraint is so strict?

surli:
I'm trying to go back on this PR, but there's still some things unclear for me. I think the current behaviour concerning String usage in parameters are mixed which make them difficult to understand and to use. I understand there's a problem on the invocation test, but I don't agree with you on relaxing the contract: template are difficult to use and we should clarify the most possible the parameters definition. The goal of the proposed contract was to clarify usage of proxy parameters and parameters, and I think it remains important to keep it. 

I look on your test, and from my point of view we should not allow to mix the following kind of behaviour on templates:

```java
public class TemplateWithFieldsAndMethods extends ExtensionTemplate {

	@Parameter
	public String PARAM;

	public TemplateParameter<String> PARAM2;

	@Local
	public TemplateWithFieldsAndMethods(String PARAM) {
		this.PARAM = PARAM;
	}

	public String methodToBeInserted() {
		return PARAM; // want to have a simple string literal
	}
}
```

```java
public class AnotherFieldAccessTemplate extends ExtensionTemplate {

    @Parameter("$name$")
    String name = "x";

    int $name$; // change field name
}
```

I agree to keep the second behaviour: it's really useful to be able to easily template the field names with a simple string. However, I propose to change the first behaviour: if you want to have a string literal, you use a CtLiteral or a TemplateParameter<String> (as you already do on the same test): then the usage of String parameter is explicitely to use on field/method renaming and not for printing literals. We are not mixing anymore behaviours. 

WDYT?

pvojtechovsky:
> I propose to change the first behaviour: if you want to have a string literal, you use a CtLiteral or a TemplateParameter (as you already do on the same test): then the usage of String parameter is explicitely to use on field/method renaming and not for printing literals. We are not mixing anymore behaviours.

I fully agree! It was the original idea of #1444 (which is then not 100% compatible). If we remove the code which tries to be backward compatible (tries to cover first case `return PARAM; // want to have a simple string literal), then it is really easier and cleaner.

I think it is good solution for future. I will try to change this PR this evening... or some longer time later when I have time for that

pvojtechovsky:
I like it as it is now. It is not compatible with some legacy templates, but it is more clear what template does.
@surli Is it the solution you mean in your last comment?

src/test/java/spoon/test/template/testclasses/constructors/TemplateWithFieldsAndMethods.java
```diff
@@ -20,7 +20,7 @@ public TemplateWithFieldsAndMethods(String PARAM,
 	}
 
 	public String methodToBeInserted() {
-		return PARAM;
+		return "PARAM";
```
surli: This one bothers me: one of the main interest of Spoon template, is that they are java typed. Here using `return PARAM;` is right concerning the typing, but throws an exception as there's no field with the new name in the test. 
With the proposed solution, (i.e. string parameters are used for renaming fields/methods), I don't think we can keep the typing safety for String (like here). 
However we should provide an immediate feedback, when applying the template: I'd like to do it when a String parameter is used in this kind of case but it seems really complicated, maybe we can do it only when an error occured, instead of throwing an NPE, WDYT?

pvojtechovsky:
> maybe we can do it only when an error occurred, instead of throwing an NPE?

Good idea! I have added error handler which throws SpoonException instead of NPE + test of this situation.

surli:
> Good idea! I have added error handler which throws SpoonException instead of NPE + test of this situation.

Nice job! It's far more clear like that for users I think. 

So, if I sum up String parameters are now used exclusively for replacing variable or method names, and we got a clear error if you use them to return a string for example. And if you want to use a parameter to return a string, you have to use a TemplateParameter or a CtLiteral, right? 

Then we are potentially breaking template clients here, but regarding all the bugs templates had we can assume that really few people used templates. 

I propose to conclude this PR that you update the javadoc of `Template` because it shows an example with String parameter which is not good anymore. And as we potentatially break client I really wonder if we put this PR in the next minor release or if we do a minor release with the most recent bug fixes and we keep this PR and some others (like the recent one with imports) for a major release in september. WDYT @pvojtechovsky @monperrus @tdurieux?

monperrus:
ok for me

pvojtechovsky:
> update documentation

I believe that current implementation is better then before, but thinking about documenation, it looks it is not good intuitive/easily understandable. There is missing a concept, which would explicitly state the meaning of template String parameter. Whether
A) it is understood as string literal
B) it is the (part of the) simple name of an model element

The inconsistency / unfriendliness is that parameter of type int is automatically handled as int literal, while string is/was ambiguous - how to explain that in documentation? 

May be the solution would be, to introduce another annotation, which would explicitly express that parameter value is not a literal, but will be used as (part of the) simple name.

Something like:
```java
@NameParameter(_name_)
String name;
```

optional solution is to mark old templates as deprecated and to introduce template builder, which does not have this problem.

surli:
> while string is/was ambiguous - how to explain that in documentation?

I agree it was ambiguous but for me, String usage is now pretty clear: you use String type for a parameter only if you want to rename some element of the code. Else you use a CtLiteral. And we should document it like that. 

>May be the solution would be, to introduce another annotation, which would explicitly express that parameter value is not a literal, but will be used as (part of the) simple name.
> Something like:
> ```java
> @NameParameter(_name_)
> String name;
> ```

It looks like a good idea indeed to use another annotation for stating that a parameter will be used only to rename code elements. I'm not sure about the usage of the argument inside the annotation: it should follow the same rule as standard parameter then. 
But then it means that we go back to the previous usage for standard parameters with String (same behaviour as CtLiteral<String>), right? 
Honestly I'm not completely sure about this solution, especially since I was pretty satisfied about the solution we have here. WDYT @monperrus ?

monperrus:
I would prefer to keep only one annotation for templates.

So I would stick and merge this solution for now

surli:
> because of reasons we 3 understand, but which is hard to explain for me in documentation

If you already started to change the documentation, push your changes, we'll be able to help you.

In my point of view, we can keep things simple to explain the stuff around String parameters: 

_String parameters are not working like other primitive type parameters, since we're using String parameters only to rename elements of the code like fields and methods. To use a parameter with a type String like other primitive types, use CtLiteral\<String\>._

pvojtechovsky:
> If you already started to change the documentation

no, I did not.

pvojtechovsky:
Please check the updated documentation and if it is OK, then this PR seems to be ready for merge.

surli:
@pvojtechovsky I fixed the failing test case and I just updated the javadoc of `Template.java` could you validate this last change then I'll merge the PR. Finally :)

src/main/java/spoon/template/Template.java
```diff
@@ -70,7 +78,7 @@
  *
  * <pre>
  * public class A {
- * 	public void insertedMethod() {
+ * 	public void methodwithpPrameterizedName() {
```
pvojtechovsky: here is a typo mistake in method name.

pvojtechovsky:
It looks OK for me. Thank you for all your input and fixing documentation ;-)

PR #1477, opened 2017-07-15
review: fix substitution of references to target type in inner type

pvojtechovsky:
Failing test and fix is part of this PR. The problem was found during development of PR #1444, which contains this fix too.


monperrus:
OK for me.

PR #1478, opened 2017-07-15
review: performance of DoNotFurtherTemplateThisElement

pvojtechovsky:
The SubstitutionVisitor throws DoNotFurtherTemplateThisElement, which internally pretty printed the to be skipped element. It is not good idea because:
* it needs time to pretty print that element - why to do it that if nobody ever needs this information?
* it happened to me that printing failed, because the model is substituted only partially at this time. It happened during development of #1444, but it was probably side effect of bug fixed by #1477, so it does not fail anymore. Therefore I am not able to create failing test for this PR.


monperrus:
Indeed. OK for me.

PR #1480, opened 2017-07-17
fix(reference): fix getTypeReference of an unknown class that has the same name that an existing class

### LINKED ISSUE #1475: ctClass#getSuperclass returns the current class
The class [it.feio.android.omninotes.models.Attachement](https://github.com/federicoiosue/Omni-Notes/blob/9ebbc0151950f8f9ec69326a9f4decc3db290f4d/omniNotes/src/main/java/it/feio/android/omninotes/models/Attachment.java) will return itself when `getSuperclass()` is called.

This happens in noclasspath mode

tdurieux:
fix #1475

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
+++ b/src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
@@ -540,6 +540,10 @@ public class ReferenceBuilder {
 			CtPackageReference packageReference = index >= 0 ? packageFactory.getOrCreate(concatSubArray(namesParameterized, index)).getReference() : packageFactory.topLevel();
 			inner.setPackage(packageReference);
 		}
+		if (!res.toString().replace(", ?", ",?").endsWith(CharOperation.toString(ref.getParameterizedTypeName()))) {
+			// verify that we did not match a class that have the same name in a different package
+			return this.jdtTreeBuilder.getFactory().Type().createReference(CharOperation.toString(ref.getParameterizedTypeName()));
+		}
 		return res;
 	}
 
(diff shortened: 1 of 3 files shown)


src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
```diff
@@ -540,6 +540,10 @@ private boolean isCorrectTypeReference(TypeReference ref) {
 			CtPackageReference packageReference = index >= 0 ? packageFactory.getOrCreate(concatSubArray(namesParameterized, index)).getReference() : packageFactory.topLevel();
 			inner.setPackage(packageReference);
 		}
+		if (!res.toString().replace(", ?", ",?").endsWith(CharOperation.toString(ref.getParameterizedTypeName()))) {
```
monperrus: a call to toString seems bad and brittle, is there another solution?

tdurieux:
In this case I compare two strings representation. It is not perfect at all but I didn't find any other solution.
This method is already dirty

Le 18 juil. 2017 9:13 AM, Martin Monperrus <notifications@github.com> a écrit :

@monperrus commented on this pull request.

________________________________

In src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java<https://github.com/INRIA/spoon/pull/1480#discussion_r127897434>:

> @@ -540,6 +540,10 @@ private boolean isCorrectTypeReference(TypeReference ref) {
                        CtPackageReference packageReference = index >= 0 ? packageFactory.getOrCreate(concatSubArray(namesParameterized, index)).getReference() : packageFactory.topLevel();
                        inner.setPackage(packageReference);
                }
+               if (!res.toString().replace(", ?", ",?").endsWith(CharOperation.toString(ref.getParameterizedTypeName()))) {


a call to toString seems bad and brittle, is there another solution?

monperrus:
OK for the noclasspath workaround, it's not the first one :-)

Issue #1484, opened 2017-07-22
feature request: SingleMemberAnnotation support


Wreulicke:
In pretty print mode with spoon, 
annotation having only "value" member converts normal annotation.

JDT Core has 3 tree model related to annotation.

* MarkerAnnotation like ( @ RequestMapping)
* SingleMemberAnnotation ( @ RequestMapping("path"))
* NormalAnnotation (@ RequestMapping(value="path"))

May I obtain code used single member annotation ?

I presents before/actual/expected.

* before (spoon process)
```java
class Controller{
    @RequestMapping("get")
    public String get() {
        return "get";
    }
}
```

* after
```java
class Controller{
    @RequestMapping(value = "get")
    public String get() {
        return "get";
    }
}
```

* expected
```java
class Controller{
    @RequestMapping("get")
    public String get() {
        return "get";
    }
}
```

# 

I used below code.
If there is my mistake, could you show me.

```java
public class ProcessorTest {
	@Test
	public void test() {
		SpoonAPI api = new Launcher();
		api.getEnvironment().setAutoImports(true);
		api.getEnvironment().setNoClasspath(true);
		api.getEnvironment().setComplianceLevel(8);
		api.addInputResource("src/test/resources/Controller.java");
		api.setSourceOutputDirectory("spooned");
		api.run();
	}
}
```

monperrus:
closed per #1485

PR #1485, opened 2017-07-22
support single member annotation

Wreulicke:
ref #1484

* only "value" member Annotation converts SingleMemberAnnotation in pretty print.
* fix tests

### DIFF
--- a/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
+++ b/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
@@ -378,7 +378,10 @@ public class DefaultJavaPrettyPrinter implements CtVisitor, PrettyPrinter {
 			try (ListPrinter lp = printer.createListPrinter("(", ", ", ")")) {
 				for (Entry<String, CtExpression> e : annotation.getValues().entrySet()) {
 					lp.printSeparatorIfAppropriate();
-					printer.write(e.getKey() + " = ");
+					if ((annotation.getValues().size() == 1 && "value".equals(e.getKey())) == false) {
+						//it is not a default value attribute. We must print a attribute name too.
+						printer.write(e.getKey() + " = ");
+					}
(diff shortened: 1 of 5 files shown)


src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
```diff
@@ -374,9 +374,15 @@ public String toString() {
...
+		if (size == 1 && values.containsKey("value")) {
+			printer.write('(');
+			elementPrinterHelper.writeAnnotationElement(annotation.getFactory(), annotation.getValues().get("value"));
+			printer.write(')');
+		} else if (size > 0) {
 			try (ListPrinter lp = printer.createListPrinter("(", ", ", ")")) {
-				for (Entry<String, CtExpression> e : annotation.getValues().entrySet()) {
+				for (Entry<String, CtExpression> e : values.entrySet()) {
 					lp.printSeparatorIfAppropriate();
 					printer.write(e.getKey() + " = ");
```
pvojtechovsky: Thank You for this PR!
What about simply change only this line like:
```java
if ((annotation.getValues().size() == 1 && "value".equals(e.key())) == false) {
   //it is not a default value attribute. We must print a attribute name too.
   printer.write(e.getKey() + " = ");
}
```
WDYT?

Wreulicke: good.
I will reflect your opinion to this PR.

monperrus:
Hi @Wreulicke, thanks a lot for your PR, it looks really good to me.

are you sure this is 100% compatible? does the Java language specification state something about this equivalence?

pvojtechovsky:
> does the Java language specification state something about this equivalence?

There is following chapter [9.7.3. Single-Element Annotations](http://docs.oracle.com/javase/specs/jls/se7/html/jls-9.html#jls-9.7.3)

monperrus:
thanks a lot!

PR #1486, opened 2017-07-24
perf: Collect list of source files only once

pvojtechovsky:
This is the implementation of FileCompilerConfig constructor which was originally called:
```java
public FileCompilerConfig(SpoonFolder folder) {
	this(folder.getAllJavaFiles());
}
```
That call can be avoided because we already have a list of all files in variable `sourceFiles`.

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/FileCompilerConfig.java
+++ b/src/main/java/spoon/support/compiler/jdt/FileCompilerConfig.java
@@ -50,6 +50,10 @@ public class FileCompilerConfig implements SpoonModelBuilder.InputType {
 		this.files = files;
 	}
 
+	/*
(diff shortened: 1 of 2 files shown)


surli:
> This is the implementation of FileCompilerConfig constructor which was originally called

Then this constructor of FileCompilerConfig is unused, right? Shouldn't we removed it?

pvojtechovsky:
Yes, it is unused.
> Shouldn't we removed it?

I agree with remove. But may be correct is to mark it as Deprecated and remove it later? I do not know if anybody uses it anywhere ... it is public API.

surli:
> mark it as Deprecated and remove it later

yes let's do that

pvojtechovsky:
Done

PR #1488, opened 2017-07-24
perf: print type only once during compiling from Factory

pvojtechovsky:
Assure that CtType is printed only once during compilation from the type ... and may be it is little bit more clear how that compiling from type works


pvojtechovsky:
This is probably not good idea, because it prints all sources before compilation and keeps them till the end - therefore it can consume a lot of memory.

pvojtechovsky:
Note: each type is pretty printed 3 times during compilation. So some kind of caching might make it faster.

Issue #1460, opened 2017-07-02
bug: Comments are probably not parsed well on Mac


pvojtechovsky:
`JDTCommentBuilder#cleanComment` method splits lines by Linux `\n`
```java
String[] lines = comment.split("\n");
```
and then cleans up remaining MS Windows `\r`. I guess that this algorithm will not work on Mac, which uses `\r` as EOL.

The handling of EOL in spoon seems to be not consistent. May be the best would be to 
1) preprocess all parsed java source file and to normalize EOL from `\r\n` and `\r` to Linux `\n`
2) then all internal algorithms in spoon can be easier and may expect that EOL is always `\n`
3) to configure what EOL is used when pretty printing

WDYT?

I am not not going to develop anything here. I just noted that it should be handled somehow during #1455, which is hard for me to test, because I am not able to say what behavior is correct concerning EOL... and Spoon. 

Are there already some rules how EOLs should be handled by Spoon?

pvojtechovsky:
I have checked current implementation and `CtComment#getContent()` [already separates lines by constant](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/compiler/jdt/JDTCommentBuilder.java#L552) "\n" - independent on OS.

So the remaining problem is to fix [bug](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/compiler/jdt/JDTCommentBuilder.java#L536) in comments handling on Mac OS, which are using "\r"

surli:
> So the remaining problem is to fix bug in comments handling on Mac OS, which are using "\r"

Actually "\r" was used on older version of MacOS (9 and below): I'm using Mac and `System.getProperty("line.separator")` returns "\n".

I tried to create a new test in #1679 and it seems to return the documentation correctly.

PR #1489, opened 2017-07-24
review: refactor: remove unused and wrong PrettyPrinter#getPackageDeclaration

pvojtechovsky:
This method is unused and unusable. It's existence avoids refactoring of PrettyPrinter concerning #1460 too.

### DIFF
--- a/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
+++ b/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
@@ -1806,6 +1806,7 @@ public class DefaultJavaPrettyPrinter implements CtVisitor, PrettyPrinter {
 	}
 
 	@Override
+	@Deprecated
 	public String getPackageDeclaration() {
 		return printPackageInfo(context.currentTopLevel.getPackage());
 	}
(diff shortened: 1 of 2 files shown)


surli:
> This method is unused and unusable

Why "unusable"? And it's unused inside Spoon but still it remains public API, maybe consider it as deprecated first?

pvojtechovsky:
I though that it is unusable, because `context.currentTopLevel` is null, so that call fails. But I see now that under some conditions there is possible to use it...

> make it deprecated

I did not wanted to wait with remove of this method until next Spoon release, because it's existence cause problem in implementation of #1490 

I will check whether I can live with that method in the evening ...

surli:
> because it's existence cause problem in implementation of #1490

I assume that the problem concern your changes with `printPackageInfo` but I'm not sure to understand why? You remove the creation of PrinterHelpers, but you call the reset method so it should do the job doesn't it?

pvojtechovsky:
You are right Simon. I checked the code again and it is no problem to keep the `PrettyPrinter#getPackageDeclaration` there. Other PRs works well with it.

Anyway I still do not understand purpose of this method so I suggest one of this:
A) we might update JavaDoc comment to make it more understandable and to describe how to use it - when it makes sense. Does anybody knows?
B) we should Deprecate it and remove it with next release, because it is just confusing...

Or just close this PR - if nobody wants to care about that ;-)

surli:
> we should Deprecate it and remove it with next release, because it is just confusing...

IMHO it's the best solution: I don't know either what it's the "good" usage for this method, and anyway users can always use `printPackageInfo`, so it's an helper not really helpful, we don't want to maintain it: let's get rid of it on the next release.

pvojtechovsky:
ok, so it is deprecated now

PR #1492, opened 2017-07-27
chore: Upgrade to org.eclipse.jdt:org.eclipse.jdt.core:3.12.3

pvojtechovsky:
This JDT is newer and has published sources in maven repository, so it is easier to debug...

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -205,7 +205,7 @@
     <dependency>
       <groupId>org.eclipse.tycho</groupId>
       <artifactId>org.eclipse.jdt.core</artifactId>
-      <version>3.12.0.v20160516-2131</version>
+      <version>3.13.0.v20170516-1929</version>
     </dependency>
     <dependency>
       <groupId>junit</groupId>


pvojtechovsky:
looks like it is compiled with JDK 1.8 :-( so not for Spoon ... yet ...

surli:
We can ignore the status of the non-working CI check for this one. 
However I wonder if we could upgrade to the **org.eclipse.jdt:org.eclipse.jdt.core:3.12.3** released in march or to **org.eclipse.tycho:org.eclipse.jdt.core:jar:3.13.0.v20170516-1929** released in june?
See https://mvnrepository.com/artifact/org.eclipse.jdt/org.eclipse.jdt.core vs https://mvnrepository.com/artifact/org.eclipse.tycho/org.eclipse.jdt.core

monperrus:
I'm fine with both of them.

pvojtechovsky:
I personally agree with everything what has accessible sources, so it is easier to debug problems or to understand how it works. So newer is better

Issue #1269, opened 2017-04-21
GWT JSNI code Comment position


marciodel:
To create a native Javascript method in GWT a comment must be used:

void native alert(String msg)/*-{
   // Javascript code here
  $wnd.alert(msg);
}-*/ ;

Spoon generates the resulting code like this

/* -{
   // Javascript code here
  $wnd.alert(msg);
}- */ 
void native alert(String msg);

Is there a way to change the position?

Thanks

monperrus:
> Is there a way to change the position?

No

But it would be an interesting feature to add, for instance by adding a value to the enumeration `CommentOffset`. Pull request welcome!

PR #1494, opened 2017-07-29
review: feat: TokenWriter - support for custom pretty printed formatting

pvojtechovsky:
This PR implements new interface
```java
/**
 * The token based and context aware printer.
 */
public interface PrinterTokenWriter {
	/**
	 * Writes one token of printed java source code to the registered {@link PrinterHelper}
	 * @param type The type of the token
	 * @param token the value of the token. The value type depends on TokenType. See {@link TokenType#getTokenClass()}
	 */
	void writeToken(TokenType type, Object token);
	/**
	 * @return {@link PrinterHelper} which this {@link PrinterTokenWriter} delegates printing to
	 */
	PrinterHelper getPrintHelper();
	/**
	 * @param printerHelper {@link PrinterHelper} which this {@link PrinterTokenWriter} delegates printing to
	 */
	void setPrintHelper(PrinterHelper printerHelper);

	/**
	 * Notifies about actually printed AST node
	 * @param element to be printed element
	 */
	void onElementEnter(CtElement element);
	/**
	 * Notifies that printing of AST node was finished
	 * @param element just printed element
	 */
	void onElementExit(CtElement element);

	/**
	 * resets to the initial state
	 */
	void reset();
}
/**
 * Defines type of token produced by PrettyPrinter
 */
public enum TokenType {
	/**
	 * One of the separator characters (){}[];,.:@=&lt;&gt;?
	 */
	SEPARATOR(String.class),
	/**
	 * The operator of an expression. For example "==", "*", ...
	 */
	OPERATOR(String.class),
	/**
	 * java keyword. For example "class", "public", ...
	 */
	KEYWORD(String.class),
	/**
	 * identifier. For example "TheNameOfMyClass_Or_VariableEtc"
	 */
	IDENTIFIER(String.class),
	/**
	 * all the spaces, new lines and tabs
	 */
	WHITESPACE(String.class),
	/**
	 * the java literals. For example
	 * "I am string"
	 * 'c'
	 * -4785.45
	 * 4456.65e10
	 * 0x00101011B
	 */
	LITERAL(String.class),
	/**
	 * The java comment. The token is instance of {@link CtComment}
	 */
	COMMENT(CtComment.class),
	/**
	 * The java doc tag. The token is instance of {@link CtJavaDocTag}
	 */
	JAVA_DOC_TAG(CtJavaDocTag.class),
	/**
	 * Special case of printing of CtCodeSnippetStatement.
	 * The tokens are not resolved here!
	 * Note:  the token value is {@link CtComment}
	 */
	CODE_SNIPPET(String.class),
	/**
	 * Special case of printing method signature.
	 * It does not happen during printing of full sources,
	 * but happen when printing an individual {@link CtExecutableReference}
	 */
	METHOD_SIGNATURE(String.class),
	/**
	 * represents new line
	 * Note:  there is no token value expected here
	 */
	NEW_LINE(Void.class),
	/**
	 * represents the defined number of tabs
	 * Note:  there is no token value expected here
	 */
	TABS(Void.class);
}
```
and there is method `DefaultJavaPrettyPrinter#setPrinterTokenWriter(PrinterTokenWriter)` which  can be used to register that token based writer.

The implementer of `PrinterTokenWriter#writeToken(...)' has to call registered PrinterHelper to print that token. The implementer can ignore/add some tokens, depending on context. It makes sense when client wants to influence formatting of pretty printed source code. Thanks to #onElementEnter and #onElementExit client is aware of the printed context and can influence formatting in context specific way.

This PR keeps complexity of DefaultJavaPrettyPrinter on the same level. The detection of tokens is done automatically by new `PrinterWriterTokenStreamAdapter`.

It even makes possible to simplify DefaultJavaPrettyPrinter in future, because it does not have to call PrinterHelper#writeTabs(), which might be called automatically by `DefaultPrinterTokenWriter`.

I believe that in future we can write a `PrinterTokenWriter` implementation, which uses same formatting (autodetected) like origin source code or template which was used to generate new source code ... but is just a far dream yet ;-)


surli:
> The implementer can ignore/add some tokens, depending on context. It makes sense when client wants to influence formatting of pretty printed source code. Thanks to #onElementEnter and #onElementExit client is aware of the printed context and can influence formatting in context specific way.

So if I understand well, the main purpose of this PR is to introduce a strategy mechanism to be able to pretty print using different styles, right? So do you plan to implement the current strategy (which follows Oracle coding convention) as an example of what's possible? It could help to understand the usage of the new classes.

pvojtechovsky:
> So if I understand well, the main purpose of this PR is to introduce a strategy mechanism to be able to pretty print using different styles, right? 

Yes, It is one of the possible use cases. Next to mentioned one it might be used like this:
* to keep formatting of source code - automatically detect what kind of separators are used in origin code or in template, which was used to generate that code.
* to improve keeping of comments on the places where they were in origin code. With token based printer we can position the comments relative to these tokens too (so not only positioning before/after class - like now, but in future also before class modifiers, before class name, before '{', before '}', and after class. Etc.

Note: But all these features needs a lot of effort, so I do not know if we ever will have it. Anyway this PR is first step to this. And I am glad that it was possible to have this step without increasing of complexity of DefaultJavaPrettyPrinter!

> So do you plan to implement the current strategy (which follows Oracle coding convention) as an example of what's possible? 

no I do not plan to do it.

src/main/java/spoon/reflect/visitor/PrinterWriterTokenStreamAdapter.java
```diff
@@ -0,0 +1,312 @@
...
+		if (canAppendToBuffer == false) {
+			//print content of buffer first, because there is character of different token type
+			flushBuffer();
+		}
+		bufferMode = m;
+		tokenBuffer.append(c);
+		return this;
+	}
+	private static final String[] javaKeywordsJoined = new String[] {
+			"abstract continue for new switch",
```
tdurieux: Is there a reason to store the java keywords in an array of strings?

pvojtechovsky: You are right, it is not nice. I converted it to array of strings now.

monperrus:
Hi Pavel,

Thanks for this proposal. It is very rich and contains many ideas. It seems to me that they are not all necessary together (for instance one does not absolutely need to refactor BinaryOperatorKind). As such, I would propose:

1. to identify the core atomic new feature
2. to split this PR into well-tested baby PRs if necessary

According to the branch name and my understanding, the core idea is to add #onElementEnter and #onElementExit in the pretty-printing mechanism. Is that correct?

pvojtechovsky:
> According to the branch name and my understanding, the core idea is to add #onElementEnter and #onElementExit in the pretty-printing mechanism. Is that correct?

yes + Stream of print tokens will help us implement other cool features on the printed source. See my previous comment for details.

> smaller PRs

yes, this PR is more like experiment - I wanted to know if this way is doable and I also need the comunity feedback - whether we go similar way or not.

monperrus:
>  I also need the comunity feedback

To sum up, there are interesting ideas here, but I'd say we need to isolate one core new feature that we want to have

pvojtechovsky:
> we need to isolate one core new feature

The core new features would be:
1) The DefaultJavaPrettyPrinter produces stream of well defined tokens, which can be easily printed to get the printed source code. 

2) The DefaultJavaPrettyPrinter fires event before and after each AST node is converted to stream of tokens

These two features together allows:
* to easily implement customer pretty printer with client specific formatting (new lines, tabs, spaces, ...)
* to improve placing of comments, which might be now placed between any token. For example
```java
/* c1 */ public /*c2*/ class /*c3*/ SomeName /*c4*/ {...}
```
this is not yet possible, but this PR would be a necessary step
* to implement pretty printer which keeps formatting of origin source code - such pretty printer would detect what separator is used between appropriate tokens in origin source code.
* to simplify DefaultJavaPrettyPrinter, which would be responsible only for firing of tokens and would be not responsible for adding of EOL, TAB and SPACE between these tokens. That responsibility would be moved to the code formater.

monperrus:
>   * to easily implement customer pretty printer with client specific formatting (new lines, tabs,
>     spaces, ...)
>   * to improve placing of comments, which might be now placed between any token. For example
>
This is what I meant, a end-user level feature. They would be tested as such.

I like more the second one, it would enable to fix #1269 (use Spoon for GWT code).

--Martin

pvojtechovsky:
@surli  @monperrus What are next steps to get it merged?

I see following topics for discussion
Q1) Do we want to have that token based printing optional (as it is in this PR) or should it be used always?
I think we might use it always. It is simpler to maintain one way then two ways.

Q2) Check the names of all new interfaces and methods - are they understandable? Do you have better names?

Q3) Do we want to have the code of PrinterHelper backward compatible? If not then the new code of this PR might by simpler.

monperrus:
Hi Pavel,
I'd say once we have split this PR into smaller PRs: 1) the refactoring on the one hand, 2) the new core atomic new feature on the other hand, and 3) backward-compatible as much as possible (currently it's binary breaking, see revapi report). 

As a first simplification, I would not change the model interfaces to add the pretty-printing information there (BinaryOperatorKind and UnaryOperatorKind), because the modeling responsibility is different from the pretty-printing one.

pvojtechovsky:
What steps do you suggest next to get it merged?

Each commit above might be merged individually as small PR (with tests of course), but does it makes sense, if they make more sense together?

monperrus:
I would say that the next step would be to have two separate discussions about PrinterTokenWriter and PrinterListener.

Let's start about PrinterListener (we can move the discussion in a separate PR too).

How is PrinterListener different from the responsibility of `scan` and `exit` of CtScanner? What do you think of #1538?

pvojtechovsky:
@monperrus, we are finished with small things, now the real fun: what to do with PrinterTokenWriter to get it merged?

monperrus:
indeed, we start the real fun :-). AFAIU, we try to introduce some configurability in DJPP. I now see more where you aim at.

Here we have two competing solutions:

* we configure DJPP by setting a PrinterTokenWriter
* we configure DJPP by setting a subclass of PrinterHelper

Currently, we have a mix of both (we hard code the field printer to PrinterTokenDetecor and setPrinterTokenWriter actually delegates to the printer).

To simplify towards a good design, one solution is to type the field `printer` with an appropriate interface (most probably close to PrinterTokenWriter) so that the the `printer` is the configuration point without an additional level of indirection.

src/main/java/spoon/reflect/visitor/CommentHelper.java
```diff
@@ -34,7 +34,7 @@
 /**
  * Computes source code representation of the Comment literal
  */
-class CommentHelper {
+public class CommentHelper {
```
monperrus: could

src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
```diff
@@ -182,9 +184,9 @@
 	private Environment env;
 
 	/**
-	 * Printer helper.
+	 * Token detector, which delegates tokens to {@link PrinterTokenWriter}
 	 */
-	private PrinterHelper printer;
+	private PrinterTokenDetecor printer;
```
monperrus: typo -> Detector

src/main/java/spoon/reflect/visitor/CommentHelper.java
monperrus: could be removed

pvojtechovsky: it is needed in tests, which are in different package.
May be I can remove part of the new tests, because the DJPP is already well tested by many other tests. Then CommentHelper can stay package protected

pvojtechovsky:
> Here we have two competing solutions:

I see it needs an explanation, so you understand the intention.

The `DJPP#printer` property is not intended for configuration. It always points to `PrinterTokenDetecor` (there is no code, which would let client to change it!). 

The purpose of internal package protected `PrinterTokenDetecor` is to detect tokens in the stream of calls made by DJPP on `printer` value. 
Note: current DJPP implementation does not provide correct stream of tokens. It provides stream of strings, which sometime contains one token and sometime more then one tokens together. So we ALWAYS have to first call `PrinterTokenDetecor`, which extracts **correct individual tokens**. Then after we have the correct tokens we can send them to a **configurable** implementation of `PrinterTokenWriter`, which is actually 
A) standard internal `PrinterHelperTokenWriter`, which delegates all tokens directly to PrinterHelper, which produces printed source code.
B) client specific extension of `DefaultPrinterTokenWriter`, which can manipulate token stream (and influence for example formatting of printed source code) and then usually delegates to `PrinterHelperTokenWriter` too.

> To simplify towards a good design, one solution is to type the field printer with an appropriate interface (most probably close to PrinterTokenWriter)

I agree with You, but that design would need much more changes in DJPP. These changes would be needed to assure that all DJPP calls of `printer` produces only **correct individual tokens**. Then extra level represented by `PrinterTokenDetecor` would be not needed. I would like to have it like this in future too, but 
* I am not sure whether the DJPP complexity will not increase
* There will be probably more space to make a mistake in implementation of DJPP - but it can be automatically tested, so such mistakes can be fixed soon.
* It would be really not small baby step, but >100 of changes of DJPP code

So I suggest to keep the design little bit ugly and to use internal `PrinterTokenDetecor` as extra level which assures  **correct individual tokens** for next configurable `PrinterTokenWriter`

src/main/java/spoon/reflect/visitor/PrinterListener.java
```diff
@@ -0,0 +1,35 @@
...
+ * knowledge of the CeCILL-C license and that you accept its terms.
+ */
+package spoon.reflect.visitor;
+
+import spoon.reflect.declaration.CtElement;
+
+/**
+ * The listener of {@link DefaultJavaPrettyPrinter}, which is notified before/after printing of each element
+ */
+public interface PrinterListener {
```
monperrus: why has it reappeared :-)?

src/main/java/spoon/reflect/visitor/PrinterTokenDetector.java
```diff
@@ -0,0 +1,299 @@
...
+import spoon.reflect.declaration.CtElement;
+import spoon.reflect.visitor.printer.TokenType;
+
+/**
+ * The special {@link PrinterHelper} implementation,
+ * which detects printed tokens and sends them to registered {@link PrinterTokenWriter}
+ *
+ * Nothing is sent to parent PrinterHelper, which is extended just to assure minimal code changes, but which is not really used
+ */
+class PrinterTokenDetector extends PrinterHelper {
```
monperrus: why having a subclass? why not modifying directly PrinterHelper and add there the PrinterTokenWriter? (since PrinterTokenDetector is already hard-wired in DJPP)

src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
```diff
@@ -1773,4 +1795,35 @@ public void calculate(CompilationUnit sourceCompilationUnit, List<CtType<?>> typ
...
+	public PrinterTokenWriter getPrinterTokenWriter() {
+		return printer.getTokenStream();
+	}
+	/**
+	 * Set {@link PrinterTokenWriter}, which has to be used to print tokens
+	 * The {@link PrinterTokenWriter#setNext(PrinterTokenWriter)} will be called
+	 * with internal implementation of {@link PrinterTokenWriter},
+	 * which delegates all tokens to {@link PrinterHelper}
+	 */
+	public DefaultJavaPrettyPrinter setPrinterTokenWriter(PrinterTokenWriter tokenWriter) {
```
monperrus: if the configuration happens in the helper, then do we need to have a setter there? this is confusing. The setter would only be in the delegate object (the helper probably)

pvojtechovsky: I want to make helper invisible/internal, so clients cannot call it. 
I expect that we will get rid of helper once in future, after DJPP code is refactored enough and produces nice print tokens directly.

src/main/java/spoon/reflect/visitor/PrinterListener.java
pvojtechovsky: I will make it package protected. Thanks

pvojtechovsky:
>> to use internal |PrinterTokenDetecor| as extra level

> why not keeping an extra level, if this encapsulates the configuration
point with the new interface.

Is it question? Or suggestion?

Actually the design which needs minimum changes in existing code of DJPP is:

DJPP calls internally PrinterTokenDetecor, which calls PrinterTokenWriter

As PrinterTokenWriter there can be 
A) standard PrinterHelperTokenWriter
B) client specific implementation which can manipulate tokens and then delegate to next PrinterTokenWriter implementation (usually PrinterHelperTokenWriter)

In future it can be like this
DJPP calls PrinterTokenWriter directly. The implementations of PrinterTokenWriter are same like in previous case.

src/main/java/spoon/reflect/visitor/PrinterTokenDetector.java
pvojtechovsky: I will check what I can do here to avoid subclassing. 

Just note: I am sure that we cannot add behavior of PrinterTokenWriter into PrinterHelper, because PrinterHelper with origin behavior is needed at the end of the token chain too.

src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
```diff
@@ -1773,4 +1795,35 @@ public void calculate(CompilationUnit sourceCompilationUnit, List<CtType<?>> typ
 	public Map<Integer, Integer> getLineNumberMapping() {
 		return printer.getLineNumberMapping();
 	}
+
+	/**
+	 * @return current implementation of {@link PrinterTokenWriter}
+	 */
+	public PrinterTokenWriter getPrinterTokenWriter() {
```
monperrus: do we need a getter? for encapsulation of configuration, the setter only is fine

src/main/java/spoon/reflect/visitor/CommentHelper.java
monperrus: > May be I can remove part of the new tests, because the DJPP is already well tested by many other tests. Then CommentHelper can stay package protected

OK. This would be consistent with the rest.

src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
```diff
@@ -182,9 +185,14 @@
 	private Environment env;
 
 	/**
-	 * Printer helper.
+	 * Token detector, which delegates tokens to {@link PrinterTokenWriter}
+	 */
+	private PrinterTokenDetector printer;
```
monperrus: what about adding Helper in the class name of PrinterTokenDetector to make the intention clear?

src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
```diff
@@ -182,9 +185,14 @@
...
+
+	/**
+	 * The listener which is called before/after each printed node
 	 */
-	private PrinterHelper printer;
+	private PrinterListener printerListener;
```
monperrus: why do you need two helpers, a PrinterTokenDetector and a PrinterListener?

monperrus:
>> why not keeping an extra level, if this encapsulates the configuration
    point with the new interface.

>Is it question? Or suggestion?

an approbation: I'm fine with having an extra level.

> I want to make helper invisible/internal, so clients cannot call it.

I understand. 

See also my other comments.

According to what I understand, the core of this PR would be to add one private setter in DJPP (`setPrinterTokenWriter`) taking one interface (`PrinterTokenWriter`) as parameter?

src/main/java/spoon/reflect/visitor/PrinterTokenWriter.java
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * The token based and context aware printer.
+ */
+public interface PrinterTokenWriter {
+	/**
+	 * Writes one token of printed java source code to the registered {@link PrinterHelper}
+	 * @param type The type of the token
+	 * @param token the value of the token. The value type depends on TokenType. See {@link TokenType#getTokenClass()}
+	 */
+	void writeToken(TokenType type, Object token);
```
monperrus: Typing by Object is rarely good.  Any way to have a more specific type?

src/main/java/spoon/reflect/visitor/PrinterTokenWriter.java
```diff
@@ -0,0 +1,46 @@
...
+
+	/**
+	 * @return next PrinterTokenWriter in the chain
+	 */
+	PrinterTokenWriter getNext();
+
+	/**
+	 * sets next PrinterTokenWriter in the chain
+	 */
+	void setNext(PrinterTokenWriter next);
```
monperrus: I don't understand the pattern here. Why do we need a chain? Does it have to be part of the API? Can this be encapsulated in DJPP or the helper?

pvojtechovsky:
@monperrus  What is good package for new `PrinterTokenWriter` and `TokenType`?
```java
public interface PrinterTokenWriter {
	/**
	 * Writes one token of printed java source code to the registered {@link PrinterHelper}
	 * @param type The type of the token
	 * @param token the value of the token. The value type depends on TokenType. See {@link TokenType#getTokenClass()}
	 */
	void writeToken(TokenType type, Object token);

	/**
	 * @return next PrinterTokenWriter in the chain
	 */
	PrinterTokenWriter getNext();

	/**
	 * sets next PrinterTokenWriter in the chain
	 */
	void setNext(PrinterTokenWriter next);

	/**
	 * resets to the initial state
	 */
	void reset();
}
```

A) spoon.reflect.visitor
B) spoon.reflect.visitor.printer

src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
pvojtechovsky: PrinterListener is there by mistake. I will remove it

src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
pvojtechovsky: Good idea. I will rename it.

src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
pvojtechovsky: ok, I will remove it

src/main/java/spoon/reflect/visitor/PrinterTokenWriter.java
pvojtechovsky: token value has actually these types:
* String
* CtComment
* CtJavaDocTag

I do not know yet how to avoid Object. May be later ... Do you have any idea?

src/main/java/spoon/reflect/visitor/PrinterTokenWriter.java
pvojtechovsky: I fight with that problem too...
I think I need it to keep support of `Environment#isPreserveLineNumbers()`.
But may be my approach is really wrong, because feature "preserving of line numbers" is not conceptually compatible with TokenPrinterWriter concept, which should let client to format printer sources as client needs.

So may be we should somehow support two modes:
1) default - which supports isPreserveLineNumbers
2) client specific TokenPrinterWriter - which will ignore that setting.
WDYT?

pvojtechovsky:
> `PrinterTokenWriter#setNext(PrinterTokenWriter)`
> Why do we need a chain? Does it have to be part of the API? Can this be encapsulated in DJPP or the helper?

There are several things behind:

C1) DJPP will produce print tokens and send them to standard or client specific PrinterTokenWriter implementation. That's clear.

C2) PrinterHelper handles formatted writing of sources. It has nice methods like `writeln`, `writeTabs`, `incTab`, `decTab`, ... each terminal implementation of PrinterTokenWriter will like to use PrinterHelper for that.

P1) PrinterHelper is not public API (yet?) so the client specific implementations of PrinterTokenWriter should not touch it.

The `PrinterTokenWriter#setNext(PrinterTokenWriter)` was a solution for (P1). The client specific implementation of PrinterTokenWriter will have `setNext(PrinterTokenWriter)`, which is called by DJPP at registration time and gets internal implementation of `DefaultPrinterTokenWriter`, which already has initialized `PrinterHelper`.
... so it is explanation for `PrinterTokenWriter#setNext(PrinterTokenWriter)` ...
... but it is ugly.

Idea: 
PrinterHelper has more features inside:
F1) formatting of printer text: 'write', `writeln`, `writeTabs`, `incTab`, `decTab`, `getTabCount`, `setTabCount`, `getLineSeparator`, `setLineSeparator`
F2) `adjustStartPosition`, `adjustEndPosition`, `removeLine`
F3) `undefineLine`, `mapLine`, `putLineNumberMapping`, `getLineNumberMapping`

Feature (F1) is needed by PrinterTokenWriter implementations and would be nice to have it public
Features F2, F3 are internal.
So I suggest to split PrinterHelper into two parts. 
1) `public class TextPrinter` with feature (F1)
2) internal `class PrinterHelper extends TextPrinter` with features F2, F3.

Then clients implementations of `PrinterTokenWriter` might use own instance of TextPrinter and there is no need for tricky chain of PrinterTokenWriters.

Internal default implementation of PrinterTokenWriter will use internal PrinterHelper and will support all legacy DJPP features.

WDYT?

monperrus:
Why do we need subclassing? What about having F1 methods public, and F2 and F3 methods package visible?

src/main/java/spoon/reflect/visitor/PrinterTokenWriter.java
monperrus: one alternative would be that instead of having the additional enum, to have a visitor like design with three methods
writeString(String), writeComment(CtComment), visitJavaDocTag(CtJavaDocTag)

pvojtechovsky:
The TextPrinter will look nicer - shorter and cleaner code if it will contain only F1 - easier for clients to understand and use that new public class.
Refactored PrinterHelper  with F2 and F3, will contain only nearly obsolete(?) features. Clients does not have to see that code at all.

pvojtechovsky:
Looks like (F3: undefineLine, mapLine, putLineNumberMapping, getLineNumberMapping). makes sense for custom PrinterTokenWriter too. So only (F2: adjustStartPosition, adjustEndPosition, removeLine) makes no sense for custom PrinterTokenWriter?

pvojtechovsky:
@monperrus please review this new implementation, where all till now discussed problems should be solved.

The PrinterHelper is now public API. The feature F1 has public methods. F2, F3 are package protected.
It was simpler to use one implementation, then PrinterHelper and TextPrinter, so I skipped this idea following your suggestion.

`PrinterTokenWriter` now has one method for each token type, so there is no token of type `Object` now.

The `SimplePrinterTokenWriter` implements standard behavior - all tokens are sent to PrinterHelper
The `PrinterTokenWriterDelegate` supports chaining of `PrinterTokenWriter`s. 
The `PrinterTokenDetectorHelper` extends `PrinterTokenWriterDelegate` + it detects tokens from the strings produces by DJPP.

What next?

monperrus:
Hi Pavel,

Thanks a lot. I really like this new PrinterTokenWriter, the design intent and its usage is clear. As a result, SimplePrinterTokenWriter is very clear too.

* Q1. In DJPP, what about typing field printer with PrinterTokenWriter (always good to type a field with an interface), this would be super clear that we introduce a new configuration point based on a strategy design pattern and a single new interface 
* Q2. In PrinterTokenWriter, what about removing getPrinterHelper. This would 1) remove a dependency to an implementation class in an interface and further clarify the goal of implementors.
* Q3. I have many questions about PrinterTokenWriterDelegate and PrinterTokenDetectorHelper but it seems they can be moved to a subsequent PR & discussion, esp if one can make Q1 and this PR becomes a smaller baby PR.

pvojtechovsky:
> Q1 In DJPP, what about typing field printer with PrinterTokenWriter

yes, I want that in future too - after DJPP code is more friendly to PrinterTokenWriter. But actually DJPP very often calls method `write(String)` and `write(char)`, which isn't (and isn't going to be) part of PrinterTokenWriter. 
Handling of calls of `write(String)` and `write(char)` is the main purpose of `PrinterTokenDetectorHelper`, which transforms these calls to call of appropriate `PrinterTokenWriter` method.

> many questions about PrinterTokenDetectorHelper 

> Q2. In PrinterTokenWriter, what about removing getPrinterHelper.

I am open to do it, the design would be nicer, but then we need some alternative solution, which will let DJPP implement legacy contracts like:
* `PrettyPrinter#getResult()` - which needs to call internally `PrinterHelper.toString()`. So there must be some way how DJPP can access current PrinterHelper
* and all other contracts related PrinterHelper features F2 and F3 - (adjust lines and line mapping)

Do you have any idea how to remove `PrinterTokenWriter#getPrinterHelper()`, but to have it still accessible in DJPP, which needs it on many places?

> PrinterTokenWriterDelegate 

we can make it package protected at the beginning - until design is more stable. Then I expect that most of the implementors will use PrinterTokenWriterDelegate as the best place to base their custom writer on.

pvojtechovsky:
> commit: refactor: flush token buffer ASAP

The `PrinterTokenDetectorHelper` now sends tokens to `PrinterTokenWriter` immediately (in scope of call of `PrinterTokenDetectorHelper#write` method). It is needed by implementations of `PrinterTokenWriter`, which needs to know correct printing context (actually printed CtElement).

monperrus:
>> Q1 In DJPP, what about typing field printer with PrinterTokenWriter

> yes, I want that in future too [...] But actually DJPP very often calls method `write(String)` and `write(char)`, which isn't (and isn't going to be) part of PrinterTokenWriter. 
> Do you have any idea how to remove `PrinterTokenWriter#getPrinterHelper()`

Interestineg. What about having two printer fields in DJPP: `PrinterHelper legacyPrinter` and `PrinterTokenWriter tokenWriterPrinter`? Then, we do not need to make the complete massive refactoring of DJPP, client code can already configure DJPP with the nice features of PrinterTokenWriter, and we can remove `PrinterTokenWriter#getPrinterHelper()`. 

If we can do a bay PR with PrinterTokenWriter and postpone the probably long discussion and merge of both PrinterTokenWriterDelegate and PrinterTokenDetectorHelper, that would be probably quite effective.

pvojtechovsky:
> two printer fields in DJPP: PrinterHelper legacyPrinter and PrinterTokenWriter tokenWriterPrinter?

I do not understand that concept. Which one will be called by DJPP? 
* How will custom PrinterTokenWriter get access to PrinterHelper?
* How will DJPP get access to PrinterHelper of customer PrinterTokenWriter to be able to implement DJPP#getResult()?

> baby PR with PrinterTokenWriter and postpone the probably long discussion ...

OK, I can make a PR which changes nothing in Spoon API (just internal implementation of DJPP), so we can merge something and others can continue with changes on DJPP too.

monperrus:
>> two printer fields in DJPP: PrinterHelper legacyPrinter and PrinterTokenWriter tokenWriterPrinter?

> I do not understand that concept.

My bad, I was confused. So we have to keep getPrinterHelper in TokenWriterPrinter.

Now, is it possible to remove PrinterTokenWriterDelegate and PrinterTokenDetectorHelper from this PR and type field `printer` by PrinterTokenWriter?

pvojtechovsky:
> is it possible to remove PrinterTokenWriterDelegate and PrinterTokenDetectorHelper from this PR and type field printer by PrinterTokenWriter?

no, PrinterTokenWriter hasn't method write, so DJPP will not compile. There are many calls of write, so it would need heavy refactoring DJPP. The purpose of PrinterTokenDetectorHelper is to implement this write method, split the content of sent strings into tokens and to sent these tokens to next PrinterTokenWriter.

pvojtechovsky:
I can make PrinterTokenWriterDelegate package protected (temporarily until the conept is well discussed). PrinterTokenDetectorHelper  is already package protected. So it is internal API and we can then nearly merge this PR. WDYT?

monperrus:
> I can make PrinterTokenWriterDelegate package protected

Hum, that would be an option, but I would prefer to converge to a good and simple design now (it's always hard to refactor at the design level later on)

> PrinterTokenWriter hasn't method write, so DJPP will not compile. There are many calls of write, so it would need heavy refactoring DJPP.

This would be a large refactoring I agree, but a rather simple one AFAIU, consisting of careful search&replace of `printer.` by `printer.getPrinterHelper()`. Correct?

Then the new abstraction `PrinterTokenWriter` and its default implementation would be enough for introducing this new configurability in DJPP.

pvojtechovsky:
> Then the new abstraction PrinterTokenWriter and its default implementation would be enough for introducing this new configurability in DJPP.

yes

> This would be a large refactoring I agree, but a rather simple one

Yes, super! So I will replace each `printer.write(...)` call by `printer.<some method from PrinterTokenWriter>()` call.
I think that it is good solution. It just will need some time and I do not like wasting of my time ;-), so please @monperrus agree that solution before I start with that. And it would be also good to merge that solution in few days, because solving of possible conflicts with other PRs would be painful.
Then PrinterTokenDetectorHelper can be deleted and design will be simple.

monperrus:
Yes! It's very good that we reach a consensus. We would basically have one changed files (DJPP) + two added ones (PrinterTokenWriter and Simple/DefaultTokenWriter), + the corresponding tests. This makes a clear baby PR with one clear goal and an atomic new feature.

pvojtechovsky:
> We would basically have one changed files (DJPP) + two added ones (PrinterTokenWriter and Simple/DefaultTokenWriter)

YES, that's my understanding too!

... uhhh ... some little problems ... it cannot be so "baby" PR, because:
* DJPP calls `snapshotLength()` and `hasNewContent()` methods to detect whether nested `DJPP.scan` add some relevant stuff. The implementation of this feature is done by `SnapshotPrinterTokenWriter`. So we have again helper internal class, which is ALWAYS called by DJPP and which then delegates to some `PrinterTokenWriter`.
* DJPP writes qualified names directly, so they must be tokenized first - implemented by `ElementPrinterHelper.writeQualifiedName`
* ElementPrinterHelper has to be refactored same like DJPP

But I have deleted `PrinterTokenWriterDelegate` and 'PrinterTokenDetectorHelper'.

@monperrus, please review it and merge it soon - I would like to avoid solving of merge conflicts in DJPP ;-)

monperrus:
Super progress, thanks a lot Pavel. 

However, the field `printer` is not yet typed by the introduced interface `PrinterTokenWriter`, which is our main goal now. 

AFAIU, to achieve this, the methods from `SnapshotPrinterTokenWriter` that have to be used in DJPP must be either put in the interface or put in `PrinterHelper` directly and called by delegation using `printer.getPrinterHelper().xxx`. We don't have to choose one or the other, some methods of  `SnapshotPrinterTokenWriter` may be moved to `PrinterTokenWriter` while others to `PrinterHelper`.

pvojtechovsky:
> However, the field printer is not yet typed by the introduced interface PrinterTokenWriter, which is our main goal now.

With my current level of understanding of DJPP, I am sure this idea is wrong.

There are these methods in SnapshotPrinterTokenWriter:
* `writeSpace()` - it is just shortcut for printer.writeWhitespace(" ").
If there will be no SnapshotPrinterTokenWriter, then I will replace it by `writeWhitespace(" ")`. Please ignore this method for now.
* `snapshotLength()` and `hasNewContent()`
You suggest to move them PrinterTokenWriter or PrinterHelper. 
  A) PrinterTokenWriter is interface, which has to be implemented by client and the contract of `snapshotLength()` and `hasNewContent()` is pure internal thing of DJPP, so we should not bother client with this
  B) PrinterHelper. Yesterday evening I already had exactly this solution. But it is bad solution, because it delegates the internal contract of DJPP "Did DJPP produced a java tokens?" to the end of the chain (PrinterHelper), which is called by potentially CLIENT SPECIFIC and ARBITRARY implementation of PrinterTokenWriter. And if client's code does something unexpected (from DJPP point of view, but expected from client point of view - e.g. just count tokens and do not sent to PrinterHelper), then DJPP will produce different tokens then it would produce with standard implementation of PrinterTokenWriter.
So I think that both solutions are design mistake.

Why do you think we should do it?

monperrus:
> A) PrinterTokenWriter is interface, which has to be implemented by client and the contract of `snapshotLength()` and `hasNewContent()` is pure internal thing of DJPP, so we should not bother client with this

I agree.

> B) PrinterHelper. Yesterday evening I already had exactly this solution. But it is bad solution, because it delegates the internal contract of DJPP "Did DJPP produced a java tokens?" to the end of the chain (PrinterHelper), which is called by potentially CLIENT SPECIFIC and ARBITRARY implementation of PrinterTokenWriter. And if client's code does something unexpected (from DJPP point of view, but expected from client point of view - e.g. just count tokens and do not sent to PrinterHelper), then DJPP will produce different tokens then it would produce with standard implementation of PrinterTokenWriter.
So I think that both solutions are design mistake.

I see your point but:

* implementing this interface is a super-advanced configuration point of Spoon, which means that clients would do so at their own risk, and that we don't provide any guarantee whatsoever
* one solution maybe that DJPP keeps the responsibility of storing the PrinterHelper in its own dedicated field, and passes it to PrinterTokenWriter via a call to setPrinterTokenWriter at the end of setPrinterTokenWriter.

pvojtechovsky:
> one solution maybe that DJPP keeps the responsibility of storing the PrinterHelper in its own dedicated field

it doesn't help if PrinterTokenWriter does not calls it.

What is wrong with current solution? The internal contract of DJPP is handled by internal DJPP class. I see nothing bad on this solution. What is the problem?

monperrus:
> What is wrong with current solution?

Good question. From the client code view point, that's fine.

What annoys me is that the current design is quite hard to understand and consequently hard to maintain.
In particular, we don't have clean delegation, because we don't type DJPP's field `printer` with PrinterTokenWriter.

It seems that the problem is that we aim for interface-based extension whereas the responsibilities are not clear-cut enough.  Maybe this case is more appropriate for inheritance-based extension as follows.

Imagine that we merge SnapshotPrinterTokenWriter and PrinterHelper in the same class. This would not impact the current code of DJPP. Let's call this class TokenPrinter.  Then:

* DJPP would offer a set setTokenPrinter
* Client code would extend DJPP by passing a subclass of TokenPrinter.

Would there be a problem with this solution. Would we agree on this?

pvojtechovsky:
What is answer to this question: "What is the contract of snapshotLength and hasNewContent methods from new DJPP point of view?"
A) to detect whether there was printed at least one character into the output (into PrinterHelper)
B) to detect whether there was sent at least one token into PrinterTokenWriter

Before it was A. Because we had no PrinterTokenWriter. But now the correct answer is B, because 
1) DJPP should not care about what client will do with tokens. DJPP should produce correct tokens independent on client.
2) before it was even buggy, because the characters, which represents whitespaces and comments must be ignored - hasNewContent should return FALSE in such case.

If you agree that B is correct, then let's found some maintainable and understandable solution, which is based on "now token detection" and not on "new character detection". 

So first please answer: A or B?

pvojtechovsky:
Would be this implementation more understandable and maintenable?
```java
/**
 * The special {@link PrinterTokenWriter} implementation,
 * detects whether some identifier, keyword, literal, operator or separator.
 * Comments, white spaces, tabs and new lines are ignored, because they can appear in the middle of printed elements.
 * was written since last call of {@link #snapshotLength()}
 */
class SnapshotPrinterTokenWriter extends PrinterTokenWriterDelegate {

	private int countOfTokens = 0;
	private ArrayDeque<Integer> lengths = new ArrayDeque<>();

	SnapshotPrinterTokenWriter(PrinterTokenWriter next) {
		super(next);
	}

	@Override
	protected void onToken(TokenType tokenType, Object token) {
		switch (tokenType) {
			case IDENTIFIER:
			case KEYWORD:
			case LITERAL:
			case OPERATOR:
			case SEPARATOR:
			case CODE_SNIPPET:
				countOfTokens++;
				break;
			case COMMENT:
			case INCREMENT_TABS:
			case DECREMENT_TABS:
			case TABS:
			case NEW_LINE:
				break;
			default:
				throw new SpoonException("Unexpected token type: " + tokenType.name());
		}
	}

	/** stores the length of the printer */
	public void snapshotLength() {
		//if the buffer contains some data then flush them first to send that token and use correct `countOfTokens`
		lengths.addLast(countOfTokens);
	}

	/** returns true if something has been written since the last call to snapshotLength() */
	public boolean hasNewContent() {
		return lengths.pollLast() < countOfTokens;
	}
}
```
It would be used same like now - as first mandatory internal `PrinterTokenWriter` implementation in the chain of `PrinterTokenWriter`s. It might be even private in DJPP, because ElementPrinterHelper and ListPrinter does not have to know that it is a special implementation of PrinterTokenWriter.

WDYT?

monperrus:
> A) to detect whether there was printed at least one character into the output (into PrinterHelper)
> B) to detect whether there was sent at least one token into PrinterTokenWriter
> So first please answer: A or B?

As you say, the current status is A, and you propose B, which is a good idea :-). We now look for a clean way to implement B.

Let me think about your last proposal.

pvojtechovsky:
And this will be next internal mandatory PrinterTokenWriter, which will let us **simplify** and fix DJPP code.
```java
/**
 * The special {@link PrinterTokenWriter} implementation,
 * ignores all writeTabs tokens produced by {@link DefaultJavaPrettyPrinter}.
 * This implementation automatically correctly writes tabs at places where they belongs to.
 * - calling writeTabs is no more responsibility of developer of {@link DefaultJavaPrettyPrinter}
 */
class AutoWriteTabsPrinterTokenWriter extends PrinterTokenWriterDelegate<AutoWriteTabsPrinterTokenWriter> {

	boolean shouldWriteTabs = true;

	AutoWriteTabsPrinterTokenWriter(PrinterTokenWriter next) {
		super(next);
	}

	@Override
	public AutoWriteTabsPrinterTokenWriter writeTabs() {
		//ignore that call. DJPP does not have to call it now at all - we can remove all these calls from DJPP sources -> easier maintenance
		return this;
	}

	@Override
	protected void onToken(spoon.reflect.visitor.PrinterTokenWriterDelegate.TokenType tokenType, Object token) {
		if (tokenType == TokenType.NEW_LINE) {
			//mark that it should write tabs after new line
			shouldWriteTabs = true;
		} else if (shouldWriteTabs && isTabCounterChangingToken(tokenType) == false) {
			/*
			 * the last token was new line. Now it is going to write some text. Write tabs to have text aligned by expected number of tabs.
			 * Note that tabs are not written if there is more empty new lines
			 */
			next.writeTabs();
			shouldWriteTabs = false;
		}
	}

	private boolean isTabCounterChangingToken(TokenType tokenType) {
		return tokenType == TokenType.INCREMENT_TABS || tokenType == TokenType.DECREMENT_TABS;
	}

	@Override
	public void reset() {
		shouldWriteTabs = true;
		super.reset();
	}
}
```
And I plan few more internal PrinterTokenWriters, which will let us simplify DJPP code even more.

pvojtechovsky:
And here is suggested (public?) `PrinterTokenWriterDelegate`, which is needed by all non terminal PrinterTokenWriters.
```java
/**
 * {@link PrinterTokenWriter} implementation, which simply sends all tokens to next {@link PrinterTokenWriter}
 * Inherit your own implementation from this class to manipulates print tokens - e.g. to change formatting of printed output
 */
public class PrinterTokenWriterDelegate<T extends PrinterTokenWriterDelegate<?>> implements PrinterTokenWriter {

	protected PrinterTokenWriter next;

	PrinterTokenWriterDelegate() {
	}

	PrinterTokenWriterDelegate(PrinterTokenWriter next) {
		setNext(next);
	}

	@Override
	public PrinterHelper getPrinterHelper() {
		return next.getPrinterHelper();
	}

	public PrinterTokenWriter getNext() {
		return next;
	}

	public void setNext(PrinterTokenWriter next) {
		if (next == null) {
			throw new SpoonException("Next PrinterTokenWriter must not be null");
		}
		this.next = next;
	}

	@Override
	public void reset() {
		next.reset();
	}

	/**
	 * Defines type of token produced by PrettyPrinter
	 */
	public enum TokenType {
		/**
		 * One of the separator characters (){}[];,.:@=&lt;&gt;?
		 */
		SEPARATOR,
		/**
		 * The operator of an expression. For example "==", "*", ...
		 */
		OPERATOR,
		/**
		 * java keyword. For example "class", "public", ...
		 */
		KEYWORD,
		/**
		 * identifier. For example "TheNameOfMyClass_Or_VariableEtc"
		 */
		IDENTIFIER,
		/**
		 * all the spaces, new lines and tabs
		 */
		WHITESPACE,
		/**
		 * the java literals. For example
		 * "I am string"
		 * 'c'
		 * -4785.45
		 * 4456.65e10
		 * 0x00101011B
		 */
		LITERAL,
		/**
		 * The java comment. The token is instance of {@link CtComment}
		 */
		COMMENT,
		/**
		 * Special case of printing of CtCodeSnippetStatement.
		 * The tokens are not resolved here!
		 * Note:  the token value is {@link CtComment}
		 */
		CODE_SNIPPET,
		/**
		 * represents new line
		 * Note:  there is no token value expected here
		 */
		NEW_LINE,
		/**
		 * prints the indentation - expected number of tabs or spaces
		 * Note:  there is no token value expected here
		 */
		TABS,
		/**
		 * increments the indentation
		 * Note:  there is no token value expected here
		 */
		INCREMENT_TABS,
		/**
		 * decrements the indentation
		 * Note:  there is no token value expected here
		 */
		DECREMENT_TABS;
	}

	/**
	 * Called at the beginning of each token handler
	 */
	protected void onToken(TokenType tokenType, Object token) {
	}

	@Override
	public T writeWhitespace(String token) {
		onToken(TokenType.WHITESPACE, token);
		next.writeWhitespace(token);
		return (T) this;
	}

	@Override
	public T writeSeparator(String token) {
		onToken(TokenType.SEPARATOR, token);
		next.writeSeparator(token);
		return (T) this;
	}

	@Override
	public T writeOperator(String token) {
		onToken(TokenType.OPERATOR, token);
		next.writeOperator(token);
		return (T) this;
	}

	@Override
	public T writeLiteral(String token) {
		onToken(TokenType.LITERAL, token);
		next.writeLiteral(token);
		return (T) this;
	}

	@Override
	public T writeKeyword(String token) {
		onToken(TokenType.KEYWORD, token);
		next.writeKeyword(token);
		return (T) this;
	}

	@Override
	public T writeIdentifier(String token) {
		onToken(TokenType.IDENTIFIER, token);
		next.writeIdentifier(token);
		return (T) this;
	}

	@Override
	public T writeCodeSnippet(String token) {
		onToken(TokenType.CODE_SNIPPET, token);
		next.writeCodeSnippet(token);
		return (T) this;
	}

	@Override
	public T writeComment(CtComment comment) {
		onToken(TokenType.COMMENT, comment);
		next.writeComment(comment);
		return (T) this;
	}

	@Override
	public T writeln() {
		onToken(TokenType.NEW_LINE, null);
		next.writeln();
		return (T) this;
	}

	@Override
	public T writeTabs() {
		onToken(TokenType.TABS, null);
		next.writeTabs();
		return (T) this;
	}

	@Override
	public T incTab() {
		onToken(TokenType.INCREMENT_TABS, null);
		next.incTab();
		return (T) this;
	}

	@Override
	public T decTab() {
		onToken(TokenType.DECREMENT_TABS, null);
		next.decTab();
		return (T) this;
	}
}
```

monperrus:
Let's concentrate on what we agree: the design of interface `PrinterTokenWriter`.

What if PrinterHelper implements it as a first baby step? Would it make sense too you?

monperrus:
I'm not sure about the Delegate concept / pattern here. It seems that you have in mind is more a decorator. Correct? If yes, this makes the idea of PrinterHelper implementing PrinterTokenWriter even more appealing.

pvojtechovsky:
> What if PrinterHelper implements it as a first baby step? Would it make sense too you?

how it implements `B) to detect whether there was sent at least one token into PrinterTokenWriter`?

It makes no sense or I did not understand why we should do that. The contract of SnapshotPrinterTokenWriter must be implementat at the beginning of chain of PrinterTokenWriters, while PrinterHelper must be always at the end of the chain. We cannot put it into one class.

> I'm not sure about the Delegate concept / pattern here. It seems that you have in mind is more a decorator. Correct? 

I am sorry, but I am not sure with these terms. So may be I use bad words, but the design of current PR seems to me still correct ;-) - I have no problem to change class names, of course.

monperrus:
OK. Let me make a try on the code directly and push to this PR.

pvojtechovsky:
It is Yours ;-)

pvojtechovsky:
And try it without changing of tests ;-)

monperrus:
Removed SnapshotPrinterTokenWriter. See https://github.com/pvojtechovsky/spoon/pull/16 

It is was much easier than expected. 

I know that `snapshotLength()` and `hasNewContent()` are still character based and some tests fail, but before going further I'd like to know your opinion on this.

WDYT?

pvojtechovsky:
The changes in pvojtechovsky#16 returns the code to the exact state I already had in my workspace few days ago. (I just had no `writeSpace()` method - but it is not relevant for this discussion)

> WDYT?

It is not final solution. Just intermediate step, where DJPP and PrinterHelper contracts are not well split.

DJPP contract is: "Produce tokens from AST"
SimplePrinterTokenWriter contract is: Print tokens to file using PrinterHelper

How to proceed?
A) we can comment the failing test and merge that to Spoon master. And then accept SnapshotPrinterTokenWriter or implement something on DJPP side what will have same behavior.
B) we can merge current PR without your changes and then think how to make SnapshotPrinterTokenWriter more understandable and maintenable.
C) you can continue with development until tests pass and then we merge into master.

A) is invalid implementation in master - probably temporarily.
B) is less work for me and immediately fully working feature
C) is risk that we have to merge DJPP with something else - acceptable, if it will not take weeks.
I think further refactoring of A or B is possible in same way.

I accept all ways. Just choose.

pvojtechovsky:
By the way I do not think that `PrinterTokenWriter#writeSpace()` method is good idea. It does not belong to PrinterTokenWriter, which should stay clear and simple. `writeSpace` is a helper method needed only by DJPP.

monperrus:
> By the way I do not think that |PrinterTokenWriter#writeSpace()| method is good idea. It does not
> belong to PrinterTokenWriter, which should stay clear and simple. |writeSpace| is a helper method
> needed only by DJPP.
I do agree with you.

But 1) it's trivial for clients to understand how to implement it, and 2) removing it from
|PrinterTokenWriter| requires a simple, yet massive code-rewriting of DJPP (replacing all calls to
writeSpace by getPrinterHelper().writeSpace() + chained call breaking). Although it would be fun to
write a processor to do so, this is not mandatory for now.

I've added some text about it.

monperrus:
I would go for A) in order to have a baby PR and proceed.

About snapshotLength and hasNewContent, I think they reflect a design problem: with those methods,
DJPP depends on the printing state to reason about what to do. We have a dedicated concept for this,
the PrintingContext. It now seems to me that we should remove snapshotLength and hasNewContent
(whether char-based or token-based) and add some logic in PrintingContext. This is not necessary in
this PR, this can be done in an upcoming PR.

WDYT?

pvojtechovsky:
> I would go for A) in order to have a baby PR and proceed.

ok, could you please fix/comment the test so we can merge soon? I might do it this evening, but may be we can be further if you do it now.

> we should remove snapshotLength and hasNewContent (whether char-based or token-based) and add some logic in PrintingContext. 

It would be cool solution. I also tried to get rid of these methods, but I failed. I did not checked the way which would use PrintingContext. Make a PR if you have exact idea or let me know some details so I can do it.

pvojtechovsky:
If you would have time for fixing the checkstyle and test problem today, then I have no problem if you make new - your own PR with all my changes in it, so you can easily see whether CI pass or not. Then we merge it as your commit - I do not care.

pvojtechovsky:
Or may be you are allowed to push directly to my branch?

monperrus:
I'm fixing the test.  

Your new PrinterTest is super interesting and the checkRepeatingOfTokens is quite cool!

However, I don't understand why it still fails. Do you have an idea?

monperrus:
Also I'd like to rename PrinterTokenWriter to simply TokenWriter. OK for you?

pvojtechovsky:
> Also I'd like to rename PrinterTokenWriter to simply TokenWriter. OK for you?

yes

pvojtechovsky:
> However, I don't understand why it still fails. Do you have an idea?

No. I cannot debug it now.

I just know that part of my test does not sends tokens to PrinterHelper, so current/wrong DJPP "is there new char detection" algorithm will be confused by that

pvojtechovsky:
> However, I don't understand why it still fails. Do you have an idea?

It fails becasue DJPP#setPrinterTokenWriter does not updates `printer` in elementPrinterHelper, so elementPrinterHelper still prints to default TokenWriter.

pvojtechovsky:
The test was buggy - so it passed ;-)
Now it is correct and it fails on the invalid use of PrinterHelper#hasNewContent(), which does not work well now.

src/test/java/spoon/test/prettyprinter/PrinterTest.java
```diff
@@ -411,6 +412,9 @@ private void checkRepeatingOfTokens(String tokenType) {
 			//contract: each printed character is handled by listener. PrinterHelper is not called directly
 			//and because PrinterTokenListener above does not use PrinterHelper, the result must be empty
 			assertEquals(0, withEmptyListenerResult.length());
+			
+			//contract: result built manually from tokens is same like the one made by DefaultTokenWriter
+			assertEquals(standardPrintedResult, allTokens.toString());
```
monperrus: this contract is not required for now IMHO
I'd remove it and proceed with merge

pvojtechovsky: ok, just comment it please

pvojtechovsky:
I checked your changes and I am ok with them. So you can merge after tests pass

monperrus:
Thanks a lot @pvojtechovsky for this long-standing effort and maturation.

src/test/java/spoon/test/prettyprinter/PrinterTest.java
```diff
@@ -165,4 +173,269 @@ public void testPrintingOfOrphanFieldReference() throws Exception {
...
+						char c = identifier.charAt(i);
+						if(i==0) {
+							if (!"?".equals(identifier)) { // generic wildcard
+								assertTrue(Character.isJavaIdentifierStart(c));
+							}
+						} else {
+							assertTrue(Character.isJavaIdentifierPart(c));
+						}
+					}
+					if (!"class".equals(identifier)) { // happens when writing Foo.class
```
pvojtechovsky: This is bug, which should be fixed in DJPP

src/test/java/spoon/test/prettyprinter/PrinterTest.java
```diff
@@ -165,4 +173,269 @@ public void testPrintingOfOrphanFieldReference() throws Exception {
...
+				}
+				
+				@Override
+				public TokenWriter writeIdentifier(String identifier) {
+					checkRepeatingOfTokens("writeIdentifier");
+					checkTokenWhitespace(identifier, false);
+					for (int i = 0; i < identifier.length(); i++) {
+						char c = identifier.charAt(i);
+						if(i==0) {
+							if (!"?".equals(identifier)) { // generic wildcard
```
pvojtechovsky: This is bug, which should be fixed in DJPP

pvojtechovsky:
Hi Martin, thanks for your time to review, understand and discuss. Thanks for merging even it is not 100% finished. Future changes will be smaller.

PR #1498, opened 2017-08-08
review: refactor Environment to add encoding

### LINKED ISSUE #1497: CompilationUnitImpl: getOriginalSourceCode() always uses the platform's default charset.
I think `getOriginalSourceCode()` should try to guess the encoding before creating a String. I did some research and found the following libraries: [juniversalchardet](https://github.com/albfernandez/juniversalchardet) and [ICU4J](https://wiki.eclipse.org/ICU4J). Maybe combining them might be useful. That is,  `getOriginalSourceCode()` tries to guess the encoding using juniversalchardet at first and if it wasn't able to find an appropriate encoding, uses ICU4J afterwards. What do you think?

surli:
Add encoding in Environment and use it in CompilationUnit#getOriginalSourceCode() 
Fix #1497

### DIFF
--- a/src/main/java/spoon/Launcher.java
+++ b/src/main/java/spoon/Launcher.java
@@ -53,6 +53,7 @@ import spoon.support.gui.SpoonModelTree;
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
+import java.nio.charset.Charset;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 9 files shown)


msteinbeck:
How about using [Charset](https://docs.oracle.com/javase/7/docs/api/java/nio/charset/Charset.html) instead of a String?

surli:
> How about using Charset instead of a String?
Done.

monperrus:
add test?

PR #1500, opened 2017-08-09
review fix: NameFilter of T also return elements of other types

surli:
There is a global problem with generic which prevent from type checking: you cannot use at runtime information about generic types and we often lose the generic type inside operation in Spoon. 
So I propose to solve a bug with generic type in NameFilter by adding a new constructor with a parameter wich is the class of the type we finally want. It's the only solution I can see to prevent an eventual ClassCastException when iterating on the result list. 
I also deprecate the old NameFilter constructor as it cannot guarantee the type of the results.

### DIFF
--- a/src/main/java/spoon/reflect/visitor/filter/NameFilter.java
+++ b/src/main/java/spoon/reflect/visitor/filter/NameFilter.java
@@ -19,6 +19,7 @@ package spoon.reflect.visitor.filter;
 import spoon.reflect.declaration.CtNamedElement;
 import spoon.reflect.visitor.Filter;
 
+
 /**
  * Filters elements by name (for instance to find a method). Example:
  *
@@ -29,6 +30,7 @@ import spoon.reflect.visitor.Filter;
  */
 public class NameFilter<T extends CtNamedElement> implements Filter<T> {
 	private final String name;
+	private Class<T> acceptedClass;
 
 	public NameFilter(String name) {
 		if (name == null) {
@@ -37,9 +39,18 @@ public class NameFilter<T extends CtNamedElement> implements Filter<T> {
 		this.name = name;
 	}
 
+	public NameFilter(String name, Class<T> acceptedClass) {
+		this(name);
+		this.acceptedClass = acceptedClass;
+	}
+
+	public void setAcceptedClass(Class<T> zeClass) {
+		this.acceptedClass = zeClass;
+	}
+
 	public boolean matches(T element) {
 		try {
-			return name.equals(element.getSimpleName());
+			return (acceptedClass == null || acceptedClass.isAssignableFrom(element.getClass())) && name.equals(element.getSimpleName());
 		} catch (UnsupportedOperationException e) {
 			return false;
 		}
--- a/src/main/java/spoon/reflect/visitor/filter/NameFilter.java
+++ b/src/main/java/spoon/reflect/visitor/filter/NameFilter.java
@@ -32,6 +32,12 @@ public class NameFilter<T extends CtNamedElement> implements Filter<T> {
 	private final String name;
 	private Class<T> acceptedClass;
 
+	/**
+	 *
+	 * @param name Name of the expected element
+	 * @deprecated Prefer to use the constructor with a class to check the type of the results
+	 */
+	@Deprecated
 	public NameFilter(String name) {
 		if (name == null) {
 			throw new IllegalArgumentException();
--- a/src/main/java/spoon/reflect/visitor/filter/NameFilter.java
+++ b/src/main/java/spoon/reflect/visitor/filter/NameFilter.java
@@ -28,7 +28,7 @@ import spoon.reflect.visitor.Filter;
  * 		new NameFilter&lt;CtMethod&lt;?&gt;&gt;(&quot;normalFor&quot;)).get(0);
  * </pre>
  *
- * @deprecated Use {@link NamedElementFilter} instead
+ * @deprecated Use {@link NamedElementFilter} instead: the actual NameFilter could return wrongly typed results. NamedElementFilter explicit the use of a type.
  */
 @Deprecated
 public class NameFilter<T extends CtNamedElement> implements Filter<T> {
--- a/src/main/java/spoon/reflect/visitor/filter/NameFilter.java
+++ b/src/main/java/spoon/reflect/visitor/filter/NameFilter.java
@@ -27,17 +27,17 @@ import spoon.reflect.visitor.Filter;
  * CtMethod&lt;?&gt; normalFor = type.getElements(
  * 		new NameFilter&lt;CtMethod&lt;?&gt;&gt;(&quot;normalFor&quot;)).get(0);
  * </pre>
+ *
+ * @deprecated Use {@link NamedElementFilter} instead
  */
+@Deprecated
 public class NameFilter<T extends CtNamedElement> implements Filter<T> {
 	private final String name;
-	private Class<T> acceptedClass;
 
 	/**
 	 *
 	 * @param name Name of the expected element
-	 * @deprecated Prefer to use the constructor with a class to check the type of the results
 	 */
-	@Deprecated
 	public NameFilter(String name) {
 		if (name == null) {
 			throw new IllegalArgumentException();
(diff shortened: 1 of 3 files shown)


src/test/java/spoon/test/filters/FilterTest.java
```diff
@@ -1124,4 +1125,17 @@ public void accept(CtType<?> ctType) {
...
+	@Test
+	public void testNameFilterWithGenericType() {
+		// contract: NameFilter of T should only return T elements
+
+		Launcher spoon = new Launcher();
+		spoon.addInputResource("./src/test/java/spoon/test/imports/testclasses/internal4/Constants.java");
+		spoon.buildModel();
+
+		CtType type = spoon.getFactory().Type().get(Constants.class);
+		List<CtMethod> ctMethods = type.getElements(new NameFilter<CtMethod>("CONSTANT"));
```
monperrus: keep only this test, add explanation, and ask @pvojtechovsky because the bug is somewhere in the query engine

pvojtechovsky: I cannot check it now, but I would say that it newer worked different. The NameFilter checks the name only. If one wants to filter by type s/he has to use TypeFilter. So it is not a bug of query engine, but old behavior of namefilter.
Anaway as client I would like the new constructor, which allows to filter by name and by type too - it is often the case...

surli: > So it is not a bug of query engine, but old behavior of namefilter.

I agree but then usage of generic type parameter for NameFilter results is dangerous, as sometimes the result type are not expected. I sometimes get a `ClassCastException` when iterating over a result list of a query with NameFilter, because the type is not checked. 

So the new constructor is a first step, but maybe we should simply get rid of the type parameter for the NameFilter and explicitely return a collection of CtNamedElement: it will be clear that it does not check types. WDYT?

monperrus:
There are two things here.

Fixing the bug.

Adding a new constructor (new feature).

They could be done in two separate PRs.

For fixing the bug, it may be possible that NameFilter extends AbstractFilter so as to get the
generic type inference right.

--Martin

monperrus:
>  For fixing the bug, it may be possible that NameFilter extends AbstractFilter so as to get the generic type inference right.

I've tried this, this fails too, no way to get the actual value of generic type T by reflection.

monperrus:
but there is a workaround, by creating an artificial matches method, see PR #1506

```
		List<CtMethod> ctMethods2 = type.getElements(new NameFilter<CtMethod>("CONSTANT") {
			@Override
			public boolean matches(CtMethod element) {
				return super.matches(element);
			}
		});
```
this works.

However, the confusing and buggy behavior is still there. If we're not able to fix it, it means that NameFilter is ill-designed, and that we have to remove its generic type parameter. This would be source-breaking (maybe not binary breaking) and would be appropriate for upcoming major release coming.



If we don't find a way to

pvojtechovsky:
What about moving of generic type of Filter from class to matches method?
It would allow writing 
new TypeFilter(SomeType.class)

...actually there is necessary to fill in generic type argument of class too.

surli:
> If we're not able to fix it, it means that NameFilter is ill-designed, and that we have to remove its generic type parameter. This would be source-breaking (maybe not binary breaking) and would be appropriate for an upcoming major release coming.

More generally I propose the following solution:
  1. Remove from the existing `NameFilter` the generic type parameter
  2. Create a `NameTypeFilter` which combines NameFilter and TypeFilter and which is often used

monperrus:
>  new TypeFilter(SomeType.class)

This can already be done, right?

src/main/java/spoon/reflect/visitor/filter/NameFilter.java
```diff
@@ -29,17 +30,33 @@
  */
 public class NameFilter<T extends CtNamedElement> implements Filter<T> {
```
monperrus: - deprecate NameFilter
- add NamedElementFilter with this code (new feature)

src/main/java/spoon/reflect/visitor/filter/NameFilter.java
```diff
@@ -29,17 +30,33 @@
...
+	 */
+	@Deprecated
 	public NameFilter(String name) {
 		if (name == null) {
 			throw new IllegalArgumentException();
 		}
 		this.name = name;
 	}
 
+	public NameFilter(String name, Class<T> acceptedClass) {
```
monperrus: thow new IllegalArg if acceptedClass is null

src/main/java/spoon/reflect/visitor/filter/NamedElementFilter.java
```diff
@@ -0,0 +1,67 @@
...
+package spoon.reflect.visitor.filter;
+
+import spoon.reflect.declaration.CtNamedElement;
+import spoon.reflect.visitor.Filter;
+
+/**
+ * Filters elements by name and by type (for instance to find a method). Example:
+ *
+ * <pre>
+ * CtMethod&lt;?&gt; normalFor = type.getElements(
```
surli: The documentation is wrong: be removing the class type parameter, we also force the type to CtNamedElement, and so we force the user to cast the returned type, even if it's checked. Is that the behaviour we want? Other solution is to remove the simple constructor, to force to specify the expected type (which was specified almost all the time to have the right return type) and then we can keep the class type parameter and the old behaviour.

pvojtechovsky:
>>  new TypeFilter(SomeType.class)

> this can already be done. Right?

no, actually you have to write

```java
new TypeFilter<SomeType>(SomeType.class)
```

src/main/java/spoon/reflect/visitor/filter/NamedElementFilter.java
monperrus: I prefer to keep the two constructors and to update the doc accordingly

monperrus:
I understand. I meant this:

    List<CtMethod> l = p.getElements(new TypeFilter(CtMethod.class));

src/main/java/spoon/reflect/visitor/filter/NamedElementFilter.java
surli: > I prefer to keep the two constructors and to update the doc accordingly

And to keep the NamedElementFilter without a type parameter? Meaning forcing all users to cast explicitely the results?

monperrus:
Do you mean that if one uses the constructor with type, you still have to use a cast?

surli:
>Do you mean that if one uses the constructor with type, you still have to use a cast?

Yes that's exactly my point. Unless we push back the type parameter on the class. And then the constructor with only one parameter can be confusing.

monperrus:
I understand. I'm OK with removing the constructor with only name. In most cases, the caller knows
what she expects.

--Martin

surli:
@monperrus for me it's done.

src/main/java/spoon/reflect/visitor/filter/NamedElementFilter.java
```diff
@@ -0,0 +1,59 @@
...
+public class NamedElementFilter<T extends CtNamedElement> implements Filter<T> {
+	private final String name;
+	private Class<T> acceptedClass;
+
+	/**
+	 *
+	 * @param name Name of the expected element
+	 * @param acceptedClass Expected class of the results
+	 */
+	public NamedElementFilter(String name, Class<T> acceptedClass) {
```
anonymous042: T super CtNamedElement?

monperrus: reverse order

src/main/java/spoon/reflect/visitor/filter/NamedElementFilter.java
```diff
@@ -0,0 +1,59 @@
...
+
+import spoon.reflect.declaration.CtNamedElement;
+import spoon.reflect.visitor.Filter;
+
+/**
+ * Filters elements by name and by type (for instance to find a method). Example:
+ *
+ * <pre>
+ * CtMethod&lt;?&gt; normalFor = type.getElements(
+ * 		new NamedElementFilter&lt;CtMethod&lt;?&gt;&gt;(&quot;normalFor&quot;, CtMethod.class)).get(0);
```
monperrus: reverse here too

PR #1504, opened 2017-08-10
review fix: issue related with Processor and generic types

### LINKED ISSUE #1503: Processor is not using templated type when process method is in an abstract class
We define an abstract spoon Processor declaring a process method on a templated type:
- https://github.com/HSarah/Tandoori/blob/master/src/main/java/analyzer/TypeProcessor.java

And create a concrete class extending the previous one:
- https://github.com/HSarah/Tandoori/blob/master/src/main/java/analyzer/ClassProcessor.java
- https://github.com/HSarah/Tandoori/blob/master/src/main/java/analyzer/InterfaceProcessor.java

The problem is that both the concrete processors are receiving every classes that matches the  upper bound of our template.

surli:
This PR propose a first solution to fix #1503 by overriding the method in the subclass. 

It changes the behaviour of `ProcessingVisitor#canBeProcessed` to consider the intersection of types instead of the union. => I don't know if considering the union of types was a wanted behavior or a mistake: in the first case, we have to had a new test to explicit the contract and the condition to use it.


src/main/java/spoon/support/visitor/ProcessingVisitor.java
```diff
@@ -45,10 +45,11 @@ private boolean canBeProcessed(CtElement e) {
 		if (!factory.getEnvironment().isProcessingStopped()
 				&& processor.getProcessedElementTypes() != null) {
 			for (Object o : processor.getProcessedElementTypes()) {
```
monperrus: add a multi-processor test with two incompatible elements

src/main/java/spoon/support/visitor/ProcessingVisitor.java
```diff
@@ -45,10 +45,11 @@ private boolean canBeProcessed(CtElement e) {
 		if (!factory.getEnvironment().isProcessingStopped()
 				&& processor.getProcessedElementTypes() != null) {
```
monperrus: bug is in getProcessedElementTypes()

PR #1507, opened 2017-08-24
review: Bug fix when using a URLClassLoader with file: URLs.

seintur:
setInputClassLoader was failing when the URLs retrieved from the specified URLClassLoader were file URLs. The URLs were transmitted to verifySourceClasspath that is expected path elements, not URLs.

This PR fixes this bug by retrieving path elements from the file URLs.

### DIFF
--- a/src/main/java/spoon/support/StandardEnvironment.java
+++ b/src/main/java/spoon/support/StandardEnvironment.java
@@ -353,13 +353,22 @@ public class StandardEnvironment implements Serializable, Environment {
 		if (aClassLoader instanceof URLClassLoader) {
 			final URL[] urls = ((URLClassLoader) aClassLoader).getURLs();
 			if (urls != null && urls.length > 0) {
-				List<String> classpath = new ArrayList<>();
+				// Check that the URLs are only file URLs
+				boolean onlyFileURLs = true;
 				for (URL url : urls) {
-					classpath.add(url.toString());
+					if( ! url.getProtocol().equals("file")) {
+						onlyFileURLs = false;
+					}
 				}
-				setSourceClasspath(classpath.toArray(new String[0]));
+				if( onlyFileURLs ) {
+					List<String> classpath = new ArrayList<>();
+					for (URL url : urls) {
+						classpath.add(url.getPath());
+					}
+					setSourceClasspath(classpath.toArray(new String[0]));
+				}
+				return;
 			}
-			return;
 		}
 		this.classloader = aClassLoader;
 	}
--- a/src/main/java/spoon/support/StandardEnvironment.java
+++ b/src/main/java/spoon/support/StandardEnvironment.java
@@ -366,6 +366,8 @@ public class StandardEnvironment implements Serializable, Environment {
 						classpath.add(url.getPath());
 					}
 					setSourceClasspath(classpath.toArray(new String[0]));
+				} else {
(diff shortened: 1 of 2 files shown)


src/main/java/spoon/support/StandardEnvironment.java
```diff
@@ -353,13 +353,22 @@ public void setInputClassLoader(ClassLoader aClassLoader) {
...
+				// Check that the URLs are only file URLs
+				boolean onlyFileURLs = true;
 				for (URL url : urls) {
-					classpath.add(url.toString());
+					if( ! url.getProtocol().equals("file")) {
+						onlyFileURLs = false;
+					}
 				}
-				setSourceClasspath(classpath.toArray(new String[0]));
+				if( onlyFileURLs ) {
```
surli: I'm not sure about this check: basically you consider that if your `URLClassLoader` contains other things than local files, you ignore everything silently. 
So:
   1. Why don't you do the loop above here, and at least add the file anyway if it's local, even if the classloader contains other things
   2. To be exhaustive, if URLs do not concern local files, maybe we should write them locally and add their local path, WDYT?

seintur: As far as I understand the behavior that was previously implemented in setInputClassLoader, if the given class loader was an URLClassLoader, the URLs were extracted and transmitted as path elements to setSourceClasspath that is verifying with verifySourceClasspath that they are all correct path elements. If the given class loader is not an URLClassLoader, setInputClassLoader is not trying to extract path elements and simply sets the class loader in this.classloader.

The check you refer to in point 1 intents to implement a behavior that is as close as possible to the previous one. If the URLClassLoader contains anything that is not an URL that refers to a file, we revert to the second alternative (simply setting the class loader in this.classloader).

I do not mind if we choose to download URLs that are not file URL and extract their path from the locally downloaded version (your point number 2). This is a choice that needs to be agreed upon.

surli: > If the URLClassLoader contains anything that is not an URL that refers to a file, we revert to the second alternative (simply setting the class loader in this.classloader).

OK I see your point, but you kept a `return` line 370, meaning, if it's a URLClassloader with other things than paths, the classloader is not set at all... 
Maybe it should be good to add another test to check the behaviour with a URLClassloader with other things than simple paths: it would be our first contract and then we'll be able to evolve it in other PR in the future.

src/main/java/spoon/support/StandardEnvironment.java
```diff
@@ -353,11 +353,20 @@ public void setInputClassLoader(ClassLoader aClassLoader) {
...
+						onlyFileURLs = false;
+					}
+				}
+				if (onlyFileURLs) {
+					List<String> classpath = new ArrayList<>();
+					for (URL url : urls) {
+						classpath.add(url.getPath());
+					}
+					setSourceClasspath(classpath.toArray(new String[0]));
 				}
```
monperrus: else { throw new SpoonException()?; }

monperrus:
@seintur OK for you now?

seintur:
Yes, good for me, thanks.

monperrus:
@seintur your commit of this PR is in an IEEE TSE paper :) http://www0.cs.ucl.ac.uk/staff/F.Sarro/resource/papers/TSE2021_API_repair.pdf

PR #1515, opened 2017-09-04
review: fix a stack overflow when calling getDeclaration on a CtExecutableReference related to a CtTypeParameter in noclasspath

msteinbeck:
As you can see, getDeclaration causes a stack overflow. I have no idea why this is happening.


surli:
@msteinbeck Thanks for reporting this one! I may have fixed it but could you improve your test for example by adding a real assert on the CtExecutableReference#getDeclaration?

msteinbeck:
> could you improve your test for example by adding a real assert on the CtExecutableReference#getDeclaration?

Do you have a good idea what kind of assertion I should add? Maybe asserting the name of the reference and the name of the declaration?

surli:
> Maybe asserting the name of the reference and the name of the declaration?

Yes, for example! I'll see if I come with something better: I'll add it myself then.

surli:
@msteinbeck ok actually I'm doing it so don't bother it ;)

msteinbeck:
> ok actually I'm doing it so don't bother it ;)

Thank you very much!

PR #1518, opened 2017-09-06
review: Remove deprecated from Spoon 5.9.0

surli:
* NameFilter is moved to spoon.legacy: NamedElementFilter is preferred
* get/set encoding of SpoonModelBuilder no longer exists: the method of Environment should be used instead
* `getPackageDeclaration()` is removed from PrettyPrinter interface and DefaultJavaPrettyPrinter
* `reset()` is removed from PrettyPrinter interface but kept as private in DefaultJavaPrettyPrinter


monperrus:
Following the [guidelines of XWiki for maximum backward compatibility](https://www.xwiki.org/xwiki/bin/view/Blog/HowToBreakAPI), instead of removing NameFilter, I propose to move it in a package `spoon.legacy`, so that impacted client code only has to modify the import statements.

surli:
>  I propose to move it in a package spoon.legacy

Done, despite what revapi is saying...

monperrus:
thanks @surli

PR #1521, opened 2017-09-06
Fixes file resource leak

tiagodrcarvalho:
Wrap the output streams for the generated files inside a try-with-resources. Was preventing generated files from being deleted/moved while Spoon was still executing.


pvojtechovsky:
Thank You @tiagodrcarvalho for fix of this little, but critical problem! The file resource leaks are really not good ;-)

tiagodrcarvalho:
Hi @pvojtechovsky. I'm glad you accepted my fix! 

We use spoon in a peculiar way (in a more "low-level" approach in which we work directly with the AST) and a feature we added in our work required a second execution of spoon (in the same Java execution) over the generated code, which lead to this resource leak (after some debugging).

PR #1523, opened 2017-09-07
review: fix: prevent renaming CtRootPackage

### LINKED ISSUE #1522: Renaming the package name, is creating a compilation issue
Hello,

Please find the below code

```java
package app;

Class Test{
        public static void main(String[] args){
                  System.out.println("test");
        }
}
```

I am parsing the code using spoon, and I am trying to change the name of the package app

```java
(element instanceof CtPackage){
         ((CtPackage) element).setSimpleName("newtest");
}
````

The output of the code is occurring as:

```java
package .newtest;

Class Test{
        public static void main(String[] args){
                  System.out.println("test");
        }
}
```

There is PACKAGE_SEPERATOR '.' added before the package. I am not sure if this is the correct way of renaming the package for a class.

Regards
Sourabh

surli:
Fix #1522

### DIFF
--- a/src/test/java/spoon/test/pkg/testclasses/ElementProcessor.java
+++ b/src/test/java/spoon/test/pkg/testclasses/ElementProcessor.java
@@ -1,0 +1,17 @@
+package spoon.test.pkg.testclasses;
+
+import spoon.processing.AbstractProcessor;
+import spoon.reflect.declaration.CtElement;
+import spoon.reflect.declaration.CtPackage;
+
+/**
+ * Created by urli on 11/09/2017.
+ */
+public class ElementProcessor extends AbstractProcessor<CtElement> {
+    public void process(CtElement element) {
+        if(element instanceof CtPackage){
+            ((CtPackage) element).setSimpleName("newtest");
+        }
+    }
+}
+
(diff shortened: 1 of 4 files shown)


surli:
This bug occured because `CtRootPackage` is renamed: IMHO this package should never be renamed, and trying to rename it should throw an exception, WDYT @monperrus @pvojtechovsky ?

pvojtechovsky:
> This bug occured because CtRootPackage is renamed: IMHO this package should never be renamed, and trying to rename it should throw an exception

I agree, that call of setSimpleName on root package should fail with exception. May be root package might silently ignore call of setSimpleName(null) and setSimpleName("")

monperrus:
> May be root package might silently ignore call of setSimpleName(null) and setSimpleName("")
I like this idea, from the viewpoint of client code, this would be better (no additional check to write)

src/test/java/spoon/test/pkg/testclasses/ElementProcessor.java
```diff
@@ -0,0 +1,17 @@
...
+import spoon.processing.AbstractProcessor;
+import spoon.reflect.declaration.CtElement;
+import spoon.reflect.declaration.CtPackage;
+
+/**
+ * Created by urli on 11/09/2017.
+ */
+public class ElementProcessor extends AbstractProcessor<CtElement> {
+    public void process(CtElement element) {
+        if(element instanceof CtPackage && element != getFactory().Package().getRootPackage()){
```
monperrus: there is a confusion, the goal is that client code indeed never has to write this. 

so the implementation RootPackage.setSimpleName would simply be ... empty.

do you see what I mean?

surli: > do you see what I mean?

Yes, I just kept going on my initial idea. I'll change that.

PR #1525, opened 2017-09-07
review: refactor: PrinterHelper to CommentHelper, LiteralHelper and OperatorHelper

pvojtechovsky:
First basic refactoring step needed to implement all the changes needed by #1494.

### DIFF
--- a/src/main/java/spoon/reflect/visitor/CommentHelper.java
+++ b/src/main/java/spoon/reflect/visitor/CommentHelper.java
@@ -38,7 +38,7 @@ import static spoon.reflect.visitor.DefaultJavaPrettyPrinter.COMMENT_STAR;
 class CommentHelper {
 
 	/**
-	 * RegExp which matches all possible line separators.
+	 * RegExp which matches all possible line separators
 	 */
 	private static final Pattern LINE_SEPARATORS_RE = Pattern.compile("\\n\\r|\\n|\\r");
 
--- a/src/main/java/spoon/reflect/visitor/CommentHelper.java
+++ b/src/main/java/spoon/reflect/visitor/CommentHelper.java
@@ -1,0 +1,134 @@
+/**
+ * Copyright (C) 2006-2017 INRIA and contributors
+ * Spoon - http://spoon.gforge.inria.fr/
+ *
+ * This software is governed by the CeCILL-C License under French law and
+ * abiding by the rules of distribution of free software. You can use, modify
+ * and/or redistribute the software under the terms of the CeCILL-C license as
+ * circulated by CEA, CNRS and INRIA at http://www.cecill.info.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the CeCILL-C License for more details.
+ *
+ * The fact that you are presently reading this means that you have had
+ * knowledge of the CeCILL-C license and that you accept its terms.
+ */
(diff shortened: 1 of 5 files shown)


src/main/java/spoon/reflect/code/UnaryOperatorKind.java
```diff
@@ -23,34 +23,47 @@
...
 	/**
 	 * Decrementation pre assignment.
 	 */
-	PREDEC, // -- _
+	PREDEC(true), // -- _
 	/**
 	 * Incrementation post assignment.
 	 */
-	POSTINC, // _ ++
+	POSTINC(false), // _ ++
```
monperrus: I'd say we already have the postfix information in the naming convention. If it starts with POST it's postfixed, else if's prefixed. Correct?

monperrus:
can the helper classes and methods be package visible? WDYT?

src/main/java/spoon/reflect/code/UnaryOperatorKind.java
pvojtechovsky: Will this be OK from your point of view?
```java
UnaryOperatorKind() {
  this.prefixOperator = name().startsWith("POST")==false;
}
```

pvojtechovsky:
> can the helper classes and methods be package visible?

ok, no problem. Just when somebody will LATER write own printer, these helper classes will be very useful. But till then they can be package protected, so we can minimize API. I will do so.

monperrus:
This code could even be put in one of the helper methods, in order not to touch and impact this core
model interface.

monperrus:
I agree, it's important to be careful on the public API surface, so as to maximize our long-term API
stability.

If we create a new public class or method, then we have to document and test it extensively.

pvojtechovsky:
> This code could even be put in one of the helper methods, in order not to touch and impact this core
model interface.

I guess you speak about `UnaryOperatorKind#isPrefixOperator()`, @monperrus why do you not like it? It is compatible is not?

pvojtechovsky:
I made helpers package protected. It needed move of `CommentHelper` to `spoon.reflect.visitor`. 
Note: during later changes needed to reach #1494 there is probably necessary to make them public again. But we will see later.

I rolled back all the changes on operator enums. I created OperatorHelper methods
```java
public static boolean isPrefixOperator(UnaryOperatorKind o) {
	return isSufixOperator(o) == false;
}
public static boolean isSufixOperator(UnaryOperatorKind o) {
	return o.name().startsWith("POST");
}
```

pvojtechovsky:
> [WARNING] Failure to transfer commons-io:commons-io:2.5-RC4-SNAPSHOT/maven-metadata.xml from http://spoon.gforge.inria.fr/repositories/snapshots was cached in the local repository, resolution will not be reattempted until the update interval of gforge.inria.fr-snapshot has elapsed or updates are forced. Original error: Could not transfer metadata commons-io:commons-io:2.5-RC4-SNAPSHOT/maven-metadata.xml from/to gforge.inria.fr-snapshot 

CI has same problems...

surli:
> CI has same problems...

Try to trigger it again: apparently there was some issue with Spoon website last days.

pvojtechovsky:
CI finally built it. Please merge it or give me feedback, so I can continue with next steps ;-)

monperrus:
Thanks Pavel!

monperrus:
> I guess you speak about UnaryOperatorKind#isPrefixOperator(), @monperrus why do you not like it?

For the record: 1) because it impacts the core Spoon metamodel and this should be done super-carefully 2) it mixes the modeling responsibility and the pretty-printing responsibility

PR #1528, opened 2017-09-11
Change to use JDK8 for compilation

### LINKED ISSUE #1519: Spoon: towards compilation using JDK 8?
Until now Spoon was compiled using JDK7 even if the tests in Spoon should use JDK8 for compilation / execution. 
I don't really see the point in changing over a compilation in JDK8 for Spoon except that, as @pvojtechovsky noticed, new versions of eclipse JDT are released (3.13.0 or 3.12.3 depending if looking on tycho or not) which are compiled using JDK8. 

As Spoon is based on JDT usage, it should be great to be able to use the last version of the library, and some statistics tends to show that as of now JDK 8 is mainly used everywhere: https://plumbr.eu/blog/java/java-version-and-vendor-data-analyzed-2017-edition

Moreover, I think it's good to ask the question now as the next release will be a major one, so if we want to change now seems a good moment. 
So what do you think about changing Spoon compilation in favor of JDK8 to be able to support new version of JDT?

surli:
Fix #1519

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -165,7 +165,7 @@
   </scm>
 
   <properties>
-    <java.src.version>1.7</java.src.version>
+    <java.src.version>1.8</java.src.version>
     <java.test.version>1.8</java.test.version>
     <runtime.log>target/velocity.log</runtime.log>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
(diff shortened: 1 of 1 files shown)


monperrus:
should we also change the animal sniffer?

surli:
> should we also change the animal sniffer?

I just did it, but is it still used?

monperrus:
thanks. I thing it is automatically used in the verify phase, correct?

surli:
> I thing it is automatically used in the verify phase, correct?

Just checked in the log, apparently it's still there but it's really not verbose at all: https://travis-ci.org/INRIA/spoon/builds/274223429#L6405

PR #1529, opened 2017-09-11
review: refactor: use LiteralHelper in DefaultJavaPrettyPrinter

pvojtechovsky:
Next baby step to the #1494

### DIFF
--- a/src/main/java/spoon/reflect/visitor/CommentHelper.java
+++ b/src/main/java/spoon/reflect/visitor/CommentHelper.java
@@ -19,7 +19,6 @@ package spoon.reflect.visitor;
 import spoon.reflect.code.CtComment;
 import spoon.reflect.code.CtJavaDoc;
 import spoon.reflect.code.CtJavaDocTag;
-import spoon.reflect.visitor.printer.PrinterHelper;
 
 import static spoon.reflect.visitor.DefaultJavaPrettyPrinter.JAVADOC_START;
 import static spoon.reflect.visitor.DefaultJavaPrettyPrinter.INLINE_COMMENT_START;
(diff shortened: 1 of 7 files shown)


pvojtechovsky:
It is ready for merge. 

Note that LiteralHelper was made public, because it is used by PrinterHelper and DefaultJavaPrettyPrinter, which are in different packages.

monperrus:
Interesting. This shows that class DJPP and the helpers should be in the same package. Since package `printer` is new and only contains 4 recent, non-API classes, I propose to move them all in package `visitor` and to keep them package visible.

pvojtechovsky:
It is misunderstanding.  Package `printer` is not new. It is old and contais PrinterHelper, which already uses all these helper classes too!

Note in future I will suggest to mark all these usages in PrinterHelper deprecated, so later helpers will be really used only from DJPP. But it is really not the case now.

monperrus:
> It is misunderstanding. Package |printer| is not new.
It's a misunderstanding :-) I know that printer is not new. I propose to remove it and move the
classes in the same package as DJPP so as to keep/restore package-visibility of the helper classes.

--Martin

pvojtechovsky:
Did I understood well, that you suggest to move class PrinterHelper into same package like DJPP?
I like it, but I want the confirmation because it will break compatibility to old versions.

monperrus:
Yes, this is what I mean.

This breaks compatibility, but PrinterHelper was never meant (or documented) to be part of the
public API.

pvojtechovsky:
The helpers were moved to the package of DJPP so they can stay package protected.
It is ready for merge or what else?

monperrus:
OK for me, thanks. Will merge tomorrow if everybody agrees.

PR #1533, opened 2017-09-15
feat(maven): create a launcher from a maven pom

### LINKED ISSUE #1396: Using Maven pom.xml to build model
Is it possible to create a Spoon model from a single pom.xml? Since the pom contains all dependencies, it should be possible to create a model without noclasspath mode. This feature is not limited to Maven, but could also be added for Ant, Gradle, and even Eclipse project files.

tdurieux:
fix #1396


src/main/java/spoon/MavenLauncher.java
```diff
@@ -0,0 +1,233 @@
...
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * Create a Spoon launcher from a maven pom file
+ */
+public class MavenLauncher extends Launcher {
```
surli: It should be useful to launch the MavenLauncher from command line with a specific argument. Do you plan to do it in another PR?

src/main/java/spoon/MavenLauncher.java
```diff
@@ -0,0 +1,233 @@
...
+			File file = dependencies.get(i);
+			classpath[i] = file.getAbsolutePath();
+		}
+		this.getModelBuilder().setSourceClasspath(classpath);
+	}
+
+
+	private InheritanceModel readPOM(String path, InheritanceModel parent) throws IOException, XmlPullParserException {
+		File parentPOM = Paths.get(path, "pom.xml").toFile();
+		if (!parentPOM.exists()) {
```
surli: What about distant parentPom like in https://github.com/pac4j/pac4j/blob/master/pom.xml for example? I had some issues with that in my own code, can you check it works well here?

tdurieux: No idea if it works, I will not spend more time on this PR. I don't need this feature.

src/main/java/spoon/MavenLauncher.java
tdurieux: We have a plugin for this

src/main/java/spoon/MavenLauncher.java
```diff
@@ -0,0 +1,236 @@
...
+import java.util.List;
+import java.util.Set;
+
+/**
+ * Create a Spoon launcher from a maven pom file
+ */
+public class MavenLauncher extends Launcher {
+	private String m2RepositoryPath;
+
+	public MavenLauncher(String pomPatch, boolean includeTest) {
```
monperrus: enum instead of boolean for sake of reability new MavenLauncher("./", TEST_INCLUDED) or new MavenLauncher("./", TEST_EXCLUDED)

tdurieux:
It is ready to merge.

src/main/java/spoon/MavenLauncher.java
```diff
@@ -0,0 +1,348 @@
...
+/**
+ * Create a Spoon launcher from a maven pom file
+ */
+public class MavenLauncher extends Launcher {
+	private String m2RepositoryPath;
+
+	/**
+	 * The type of source to consider in the model
+	 */
+	public enum SOURCE_TYPE {
```
monperrus: I would rename source to app, and add the explicit "source" (since we can also build a shadow model of binary classes)
APP_SOURCE, TEST_SOURCE,  ALL_SOURCE

src/test/java/spoon/MavenLauncherTest.java
```diff
@@ -0,0 +1,35 @@
...
+		assertEquals(8, launcher.getEnvironment().getComplianceLevel());
+
+		assertEquals(10, launcher.getEnvironment().getSourceClasspath().length);
+		// 54 because of the sub folder
+		assertEquals(54, launcher.getModelBuilder().getInputSources().size());
+
+		// with the tests
+		launcher = new MavenLauncher("./", MavenLauncher.SOURCE_TYPE.ALL);
+		assertEquals(10, launcher.getEnvironment().getSourceClasspath().length);
+		// 54 because of the sub folder
```
monperrus: 54?

src/main/java/spoon/MavenLauncher.java
```diff
@@ -0,0 +1,348 @@
...
+				if (parent == null) {
+					return null;
+				}
+				return parent.getProperty(key);
+			}
+			return value;
+		}
+
+		/**
+		 * Get the source version of the project (default 7)
```
surli: Why java 7 by default? Java 8 seems to be more used: https://plumbr.eu/blog/java/java-version-and-vendor-data-analyzed-2017-edition

src/main/java/spoon/MavenLauncher.java
```diff
@@ -0,0 +1,348 @@
...
+	public enum SOURCE_TYPE {
+		// only the main code of the application
+		APP_SOURCE,
+		// only the tests of the application
+		TEST_SOURCE,
+		// all the sources
+		ALL_SOURCE
+	}
+
+	public MavenLauncher(String pomPatch, SOURCE_TYPE sourceType) {
```
surli: typo: pomPath

surli: Actually it seems to be the projectRoot, and not the pomPath, as it should be a directory, right?

src/test/java/spoon/MavenLauncherTest.java
```diff
@@ -0,0 +1,35 @@
...
+		// with the tests
+		launcher = new MavenLauncher("./", MavenLauncher.SOURCE_TYPE.ALL_SOURCE);
+		assertEquals(10, launcher.getEnvironment().getSourceClasspath().length);
+		// 235 because of the sub folders of src/main/java and src/test/java
+		assertEquals(235, launcher.getModelBuilder().getInputSources().size());
+	}
+
+	@Test(expected = SpoonException.class)
+	public void MavenLauncherOnFileTest() {
+		new MavenLauncher("./pom.xml", MavenLauncher.SOURCE_TYPE.APP_SOURCE);
```
surli: Why this behaviour? Why only accepting the root directory and not the pom.xml itself?

tdurieux: because it's boring to handle a lot of different cases.

src/main/java/spoon/MavenLauncher.java
tdurieux: Because the default compliance level in spoon is 7

src/main/java/spoon/MavenLauncher.java
tdurieux: Yes

src/test/java/spoon/MavenLauncherTest.java
surli: It seems pretty easy (and useful IMHO) to handle this one...

src/main/java/spoon/MavenLauncher.java
surli: Ok, maybe we should change that I'll open an issue.

surli: And then do we have a constant value for this default compliance level?

tdurieux: Of course no...

monperrus:
Ok for me. Will merge it this afternoon.

surli:
> Ok for me. Will merge it this afternoon.

Could you wait a bit? I want to do the following changes:
  * Use the constant provided in #1543 
  * Allow to provide a pom.xml path instead of the root path (I'll do it this afternoon)

monperrus:
Thomas tells me that both are already done in the PR.

src/main/java/spoon/MavenLauncher.java
```diff
@@ -0,0 +1,394 @@
...
+		super();
+		this.m2RepositoryPath = m2RepositoryPath;
+		this.sourceType = sourceType;
+
+		File mavenProjectFile = new File(mavenProject);
+		if (!mavenProjectFile.exists()) {
+			throw new SpoonException(mavenProject + " does not exist.");
+		}
+		if (mavenProjectFile.isFile()) {
+			mavenProject = mavenProject.substring(0, mavenProject.length() - mavenProjectFile.getName().length() - 1);
```
surli: It seems safer to take `mavenProjectFile.getParent().getPath()` here, no?

surli:
> Thomas tells me that both are already done in the PR.

Indeed, ok then.

Issue #1513, opened 2017-09-01
Get list of accessible variables at point in code


sourabhsparkala:
I am trying to ffind all the possible list of CtLocalVariable or CtField which are accessible depending on the position at a point in code. A small example is give below.

Is it possible to query for something like this in spoon?

```java
package app;


public class App 
{
	
	static int test = 0;
	public static void main(String[] args)
	{
		int local = 0;  /* The iftest local variable from ifblock should not be accessible within this outer parentblock. similarly the variables "local" and "test" should be accessible within this block.*/
		System.out.println("Hello world");
		
		if(test<3){   /* The iftest local variable should be accessible within this ifblock, not outside. similarly the variables "local" and "test" should be accessible within this block as well.*/
			
			int iftest = 10;
			test += 1;
		}
		 /* The iftest local variable from ifblock should not be accessible now since we are in the parent block*/
		System.out.println("2nd line");
	}
}
```

Regards
Sourabh

surli:
Hi @sourabhsparkala 

>  I am trying to ffind all the possible list of CtLocalVariable or CtField which are accessible depending on the position at a point in code

Actually I think we had the contrary tool in Spoon: given a CtLocalVariable or a CtField we can give you the list of CtElements in which the variable or the field are visible. You may have a look on `LocalVariableScopeFunction` and `FieldScopeFunction`. Those functions are designed to be used inside a query, like in this snippet:

```java
ctVar.map(new LocalVariableScopeFunction()).list()
```

Maybe @pvojtechovsky can give you some insights on those functions.

pvojtechovsky:
Hi Sourabh,

yes, you can query for somethig like that. Have a look at documentation of LocalVariableReferenceFunction, LocalVariableScopeFunction and PotentialVariableDeclarationFunction. There is also a generic query function VariableReferenceFunction, which returns references to all kinds of variables. All of them can be used using example mentioned by Simon in the comment above.

I had similar need like you when I did algorithm for renaming of local variable. You can have a look at code of CtRenameLocalVariableRefactoring. It is not easy reading - rules for visibility of local variables in Java are tricky, but Spoon knows them all! So may be this is exactly what you need - You did not described why do you need such function.

Regards, 
Pavel

sourabhsparkala:
@pvojtechovsky @surli 

Thank you for the reply. I want to find the influence of these variables in a code when statically analyzing the code. Hence I need this feature.

package app;


public class App 
{
	
	static int test = 0;
[...]

PR #1536, opened 2017-09-17
review: fix PotentialVariableDeclarationFunction ignoring the static scope

pvojtechovsky:
As reported in #1513 the `PotentialVariableDeclarationFunction` returns non static field `test` as potential variable declaration of local variable reference declared in static method.
```java
int test = 0;  //this one should not be returned, because it is not static
public static void main(String[] args)
{
	int test = 1; 
	assertEqualst(1, test); //reference to local test variable
}
```

### DIFF
--- a/src/test/java/spoon/MavenLauncherTest.java
+++ b/src/test/java/spoon/MavenLauncherTest.java
@@ -19,7 +19,7 @@ public class MavenLauncherTest {
 		// with the tests
 		launcher = new MavenLauncher("./", MavenLauncher.SOURCE_TYPE.ALL_SOURCE);
 		// 235 because of the sub folders of src/main/java and src/test/java
-		assertEquals(235, launcher.getModelBuilder().getInputSources().size());
+		assertEquals(236, launcher.getModelBuilder().getInputSources().size());
 
 		// specify the pom.xml
 		launcher = new MavenLauncher("./pom.xml", MavenLauncher.SOURCE_TYPE.APP_SOURCE);
--- a/src/test/java/spoon/MavenLauncherTest.java
+++ b/src/test/java/spoon/MavenLauncherTest.java
@@ -3,6 +3,7 @@ package spoon;
 import org.junit.Test;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
 
 public class MavenLauncherTest {
 	@Test
@@ -18,8 +19,8 @@ public class MavenLauncherTest {
 
 		// with the tests
 		launcher = new MavenLauncher("./", MavenLauncher.SOURCE_TYPE.ALL_SOURCE);
-		// 235 because of the sub folders of src/main/java and src/test/java
-		assertEquals(236, launcher.getModelBuilder().getInputSources().size());
+		// 236 because of the sub folders of src/main/java and src/test/java
+		assertTrue(launcher.getModelBuilder().getInputSources().size() >= 236);
 
 		// specify the pom.xml
 		launcher = new MavenLauncher("./pom.xml", MavenLauncher.SOURCE_TYPE.APP_SOURCE);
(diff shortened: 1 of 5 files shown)


pvojtechovsky:
This PR is now based on #1541 because it was needed to debug problem of `MainTest`

The implementation of this PR fixes two more problems
1) this implementation contains first lambda expression in Spoon java sources ... and that lambda expression code was not accepted by `MainTest#testMain`. See my change [here](https://github.com/INRIA/spoon/pull/1536/commits/794ae4107c095ff6781b956671027c6f8c609b33#diff-fa9f42f56ead24693fdd70080c5adf50R182)

2) The MavenLauncherTest#spoonMavenLauncherTest is quite fragile - we are probably going to change these number quite often. Should we change it?

src/test/java/spoon/MavenLauncherTest.java
```diff
@@ -19,7 +19,7 @@ public void spoonMavenLauncherTest() {
 		// with the tests
 		launcher = new MavenLauncher("./", MavenLauncher.SOURCE_TYPE.ALL_SOURCE);
 		// 235 because of the sub folders of src/main/java and src/test/java
-		assertEquals(235, launcher.getModelBuilder().getInputSources().size());
+		assertEquals(236, launcher.getModelBuilder().getInputSources().size());
```
tdurieux: Maybe we can remove this assertion because it will be horrible to maintain.

src/test/java/spoon/test/filters/FilterTest.java
```diff
@@ -883,16 +884,16 @@ public void testClassCastExceptionOnForEach() throws Exception {
 		{
 			Context context = new Context();
 			//contract: if the query produces elements which cannot be cast to forEach consumer, then they are ignored
-			launcher.getFactory().Package().getRootPackage().filterChildren(f->{return true;}).forEach((CtType t)->{
+			launcher.getFactory().Package().getRootPackage().filterChildren(null).forEach((CtType t)->{
```
monperrus: why do you need to change this contract?

src/test/java/spoon/test/filters/FilterTest.java
```diff
@@ -883,16 +884,16 @@ public void testClassCastExceptionOnForEach() throws Exception {
...
 			});
 			assertTrue(context.count>0);
 		}
 		{
 			Context context = new Context();
-			//contract: if the for each implementation throws CCE then it is reported
+			//contract: if the for each implementation made by lambda throws CCE then it is reported
 			try {
-				launcher.getFactory().Package().getRootPackage().filterChildren(f->{return true;}).forEach((CtType t)->{
+				launcher.getFactory().Package().getRootPackage().filterChildren(null).forEach((CtType t)->{
```
monperrus: same question

src/test/java/spoon/test/main/MainTest.java
```diff
@@ -177,6 +178,9 @@ public void scan(CtElement element) {
 					if (reference.getParameters().get(i) instanceof CtArrayTypeReference && ((CtArrayTypeReference) reference.getParameters().get(i)).getComponentType() instanceof CtTypeParameterReference) {
 						continue;
 					}
+					if (executableDeclaration instanceof CtLambda) {
```
monperrus: could you add a comment line to explain why we can not make the assertion for Lambdas?

monperrus:
looks good, commented with minor questions.

monperrus:
Or we can have something like 
launcher.getModelBuilder().getInputSources().size() >= 230

src/test/java/spoon/test/filters/FilterTest.java
pvojtechovsky: it is not change. It is just shorter way how write the same: "no filter = return all children"

src/test/java/spoon/test/main/MainTest.java
pvojtechovsky: I do not know whether we cannot make assertions about lambdas. Lambdas are too different (comparing to method) and they are not supported yet by this test. So may be there is necessary to write lambda specific assertion. But I did not understood the purpose of this test, so I would rather let You to fix it.

pvojtechovsky:
I have removed commits which belongs to #1541, so this PR is simpler now. The minor problems were fixed. So it can be merged

PR #1538, opened 2017-09-18
add simple support for enter/exit in DJPP as in CtScanner

monperrus:
design exploration for #1494

### DIFF
--- a/src/main/java/spoon/reflect/visitor/CtAbstractVisitor.java
+++ b/src/main/java/spoon/reflect/visitor/CtAbstractVisitor.java
@@ -97,6 +97,7 @@ import java.lang.annotation.Annotation;
  *  See {@link CtScanner} for a much more powerful implementation of CtVisitor.
  */
 public abstract class CtAbstractVisitor implements CtVisitor {
+
 	@Override
 	public <A extends Annotation> void visitCtAnnotation(
 			CtAnnotation<A> annotation) {
--- a/src/main/java/spoon/reflect/visitor/CtAbstractVisitor.java
+++ b/src/main/java/spoon/reflect/visitor/CtAbstractVisitor.java
@@ -97,7 +97,6 @@ import java.lang.annotation.Annotation;
  *  See {@link CtScanner} for a much more powerful implementation of CtVisitor.
  */
 public abstract class CtAbstractVisitor implements CtVisitor {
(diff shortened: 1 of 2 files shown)


pvojtechovsky:
Why? To use CtScanner#enter and #exit methods for listening? Or do you mean something else?

monperrus:
> To use CtScanner#enter and #exit methods for listening?
Yes, to extend DJPP and override #enter and #exit methods for listening.

pvojtechovsky:
I did not debugged it, but looking at code, it looks like CtScanner#enter/exit are NEVER called by DJPP, because it always directly calls DJPP#scan method.

monperrus:
Good point! See  c256281

My point here is not to duplicate existing responsibilities... and thanks to this discussion, we can even make DJPP come under CtScanner.

pvojtechovsky:
> add call to enter and exit
Then, I suggest to keep `DJPP extends CtVisitor` and just add new 
`protected enter/exit(CtElement){}` methods into DJPP, same like they are in CtScanner.

I have feeling like the extending from CtScanner is confusing, because all the code of CtScanner is never called. 

But new enter/exit methods of DJPP, might be good substitution for my `PrinterListener`. I can accept that solution for now.

monperrus:
> I have feeling like the extending from CtScanner is confusing, because all the code of CtScanner is never called.

I tend to agree... 

958c353 is an alternative. WDYT?

pvojtechovsky:
> depends on CtAbstractVisitor instead

is better then CtScanner.

Do we want to force change of DJPP if we add new method into CtVisitor?
If DJPP extends CtAbstractVisitor, then new CtVisitor method is added to CtAbstractVisitor and may stay empty in DJPP, so DJPP will do nothing for that new method. 
Actually all the methods of DJPP implements some behavior for each method of CtVisitor. So it would make sense to force it by `DJPP implements CtVisitor`.

But I can live with both solutions. It is up to You ;-)

monperrus:
good point again. [KISSed](https://en.wikipedia.org/wiki/KISS_principle) it accordingly: e59a7fd

pvojtechovsky:
Looks good for me. Even the `scan(CtReference ref)` seems to be correct way. If tests pass, I agree with merge.

PR #1539, opened 2017-09-18
fix: java compliance to 8 - allow lambdas in spoon code

pvojtechovsky:
I would like to use lambdas in spoon code now, but this test fails on it.


monperrus:
looks natural now :-) do we have a consensus there?

msteinbeck:
> I would like to use lambdas in spoon code now

I really would like to use them, too :D

surli:
Ok for me

PR #1540, opened 2017-09-18
remove ant related code and dependency

monperrus:
is this old untested code still relevant?


surli:
I agree on removing it.

PR #1541, opened 2017-09-18
review: perf: Avoid slow CCE. Check acceptable type when possible

pvojtechovsky:
I wanted to debug `MainTest#testMain` and I have found that it is terribly slow (needs nearly 19 min) in debug mode. 18 min is spent in creation of ClassCastException stacktraces.

This PR detects acceptable type from message of first CCE and then checks whether class of input is acceptable - so it avoids throwing of CCE - performs faster in debug (52s)

The performance improvement is only little in non debug mode (28s origin, 25s after improvement).

Actually this PR fails on java 7/8 source compliance (will be fixed by #1539)

### DIFF
--- a/src/main/java/spoon/reflect/visitor/chain/CtQueryImpl.java
+++ b/src/main/java/spoon/reflect/visitor/chain/CtQueryImpl.java
@@ -17,15 +17,19 @@
 package spoon.reflect.visitor.chain;
 
 import java.lang.reflect.Array;
+import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import spoon.Launcher;
 import spoon.SpoonException;
 import spoon.reflect.declaration.CtElement;
 import spoon.reflect.visitor.Filter;
 import spoon.reflect.visitor.filter.CtScannerFunction;
+import spoon.support.util.RtHelper;
 
 /**
  * The facade of {@link CtQuery} which represents a query bound to the {@link CtElement},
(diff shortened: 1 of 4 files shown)


tdurieux:
THANK YOU!

src/main/java/spoon/reflect/visitor/chain/CtQueryImpl.java
```diff
@@ -325,7 +329,52 @@ private boolean isFailOnCCE() {
...
+						}
+					} else {
+						canExtractTypeFromCCE = false;
+					}
+				}
+			}
+		}
 	}
+	private static final Pattern cceMessagePattern = Pattern.compile("(\\S+) cannot be cast to (\\S+)");
+	private static boolean canExtractTypeFromCCE = true;
```
tdurieux: Is it normal that canExtractTypeFromCCE cannot be set to true?

pvojtechovsky: the canExtractTypeFromCCE is here as fallback for the case, when somebody is using different java implementation whose CCE error message looks different. Only in such case the value is set to false and the detection of acceptable type from message is OFF.

> Is it normal that canExtractTypeFromCCE cannot be set to true?

So I see actually no case when such setter would be needed.

pvojtechovsky:
The ignoring of CCE caused by CtQuery filters was buggy, because for example CtConsumer implemented
A) by lambda
B) by local class
has different length of stacktrace when CCE is thrown on input parameter.

This PR solves this problem by these changes:
1) CtQuery tries to detect acceptable input type from each callback (CtFunction, CtConsumableFunction, CtConsumer) by java reflection - but it is not possible for Lambdas, which does not provide correct parameter type.
2) so in case of Lambdas the accepted input type is not known and only in their case the accepted input type is detected from ClassCastException after it fails first time.
3) when accepted input type is known then it is always checked before the callback is called, so CCE is not thrown and it performs much faster in debug mode in Eclipse.
4) if CCE is thrown on any other place then it is correctly rethrown

This PR fixes two other bugs, which are now visible
1) RtHelper#getMethod must ignore synthetic methods, otherwise incorrect acceptable type is detected. For example `NamedElementFilter#matches` has
A) synthetic NamedElementFilter#matches(CtElement)
B) public NamedElementFilter#matches(CtNamedElement)

2) APITest#testSetterInNodes thrown unexpected ClassCastException, which was silently ignored by old code. @surli please review my fix in `APITest` - you are the author of this test.

This PR is finished and ready for merge from my point of view.

surli:
It seems ok to me. @tdurieux ?

tdurieux:
To be honest, I don't understand completely the impact of this change on CtQuery.
I will trust the tests.

surli:
Thanks @pvojtechovsky

Issue #1537, opened 2017-09-18
Allow passing properties to a Processor trough Maven plugin


tmortagne:
Maybe I missing something but I can't find any way to pass properties to a Processor when using the Maven plugin.

tmortagne:
Hmm actually if I understand well SpoonConfigurationFactory I'm support to provide a spoon.xml file at the root of my module resources. Probably just missing some example/documentation, sorry for the noise.

monperrus:
> Probably just missing some example/documentation, sorry for the noise. 
Sorry for the missing documentation :-) We also love documentation PRs :-D

tmortagne:
> We also love documentation PRs :-D

Any pointer on how to find out what is the format in that file and where it's supposed to be located ?

surli:
> Any pointer on how to find out what is the format in that file and where it's supposed to be located ?

Actually reading to the test of spoon-maven-plugin, the `spoon.xml` file is only used to pass the template and processor's names configuration, look at this spoon.xml example:

```xml
<spoonModel>
    <processors>
        <processor>com.dooapp.processor.wrapper.GenerateMissingWrapperClassProcessor</processor>
    </processors>
    <templates>
        <template>com.dooapp.template.WrapperConstructorTemplate</template>
    </templates>
</spoonModel>
```

So given the name of the issue I don't think it's what you want to do, right? 
Could you give more details then on the kind of properties you would like to give to the processor?

tmortagne:
> So given the name of the issue I don't think it's what you want to do, right?

Yep.

>  Could you give more details then on the kind of properties you would like to give to the processor?

I'm writing a processor that takes a list of forbidden invocations so I need to pass at least classes and method names (and probably parameters to be fully clean even if I don't have the need right now). Should maybe put that is a pull request on Spoon side when I'm happy with it as it's probably generic enough.

surli:
> I'm writing a processor that takes a list of forbidden invocations so I need to pass at least classes and method names

Ok that's something which is not supported yet throught the maven plugin.
So, first: an easy workaround is to create a Maven module based on spoon-core, which will use your processor and to create a command line launcher to specify your parameters. 

Second: what you really want (in the future) is to be able to specify name of the Spoon Model (classes, methods, ...) and to use them inside the Processor to match some elements, right?

Something like:

```xml
<spoonModel>
  <processors>
     <processor>
        <name>my.package.CustomProcessor</name>
        <configuration>
          <include>
[...]

Issue #1534, opened 2017-09-15
Maven plugin should fail when getEnv().report(Level.error, msg) is called


tmortagne:
At XWiki we plan to use Spoon Maven plugin to fail the build when some module use a forbidden method. I wrote a custom Processor (which was super easy, good job :)) but I can't find any way to tell the plugin to fail the build when there is errors in the report.

tdurieux:
Hi,

The easiest way to stop the build is to throw a runtime exception.

tmortagne:
Sure but it would be nicer to see all the errors of the report instead of crashing the execution. Just need a switch to fail or not the build when there is errors in the report.

tmortagne:
> I'm pretty sure that the only way to change the build status of maven is to throw an exception.

Yes but that exception should be through by the plugin itself when finding errors in the report IMO. It allows seeing all errors and not just the first one and it also allows enable/disable failing the build for the same processor depending on the need. There is several processor provided by Spoon and which don't trough any exception right now for example.

vmassol:
Spoon advocates itself as a tool to enforce architecture rules. As such it should provide a "check" goal that would run the processors and have options to fail the build if a processor reports an ERROR. That seems a very basic need for such a tool. All Maven plugins I know that do checks provide this ability.

@tdurieux Hi. For the context, I met Martin at a STAMP meeting and he advocated the use of Spoon for the XWiki project. I said I'd look into it and Thomas worked on trying it out for XWiki. It's working quite well except on this point. I think it would benefit the tool nicely if you guys could add this feature. Thanks!

tdurieux:
> @tdurieux Hi. For the context, I met Martin at a STAMP meeting and he advocated the use of Spoon for the XWiki project. I said I'd look into it and Thomas worked on trying it out for XWiki. It's working quite well except on this point. I think it would benefit the tool nicely if you guys could add this feature. Thanks!

I agree, I just gave a workaround since we don't have this feature (yet).

@tmortagne Do you use `getEnvironment().report(...)`?

monperrus:
Interesting! That would be indeed a useful feature. 

What would be a good API to your opinion?

    Something.failBuild() // to be called in the processor
    Something.failBuild(String msg)

tdurieux:
> Something.failBuild() // to be called in the processor
Something.failBuild(String msg)

I'm not a big fan because the processors are not only designed for the Maven Plugin. 
And these API outside the maven plugin are hard to understand.

[...]

PR #1545, opened 2017-09-19
feat: propose a new way to filter elements in AbstractProcessor

surli:
This PR intends to propose a new way to filter elements in AbstractProcessor in order to be used in spoon-maven-plugin as discussed in #1537 

The idea of the PR is to be able to give some rules for filtering elements like: I want 
   * all classes containing "Ct"
   * and the interface of the following name

Or I want all types coming from package containing "testclasses"

### DIFF
--- a/src/main/java/spoon/processing/AbstractProcessor.java
+++ b/src/main/java/spoon/processing/AbstractProcessor.java
@@ -20,13 +20,16 @@ import org.apache.log4j.Level;
 import spoon.Launcher;
 import spoon.compiler.Environment;
 import spoon.reflect.declaration.CtElement;
+import spoon.reflect.declaration.CtNamedElement;
 import spoon.reflect.factory.Factory;
 import spoon.testing.utils.ProcessorUtils;
 
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.lang.reflect.Method;
+import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Map;
 import java.util.Set;
 
 /**
@@ -38,6 +41,8 @@ public abstract class AbstractProcessor<E extends CtElement> implements Processo
 	Factory factory;
 
 	Set<Class<? extends CtElement>> processedElementTypes = new HashSet<>();
+
+	Map<Class<? extends CtNamedElement>, Set<String>> acceptedElements = new HashMap<>();
 
 	/**
 	 * Empty constructor only for all processors (invoked by Spoon).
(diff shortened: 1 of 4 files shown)


tdurieux:
This feature should not be implemented in AbstractProcessor but in a new processor

monperrus:
> This feature should not be implemented in AbstractProcessor but in a new processor

Probably yes. Should it this new processor be in spoon-core or in spoon-maven-plugin?

pvojtechovsky:
This PR seems to me quite **use-case specific** (only for few clients). I am not using maven and not using Processors, so I am not the one who should decide, but is it really the case that most of the clients will need to process elements **by type and occurrence of substring in element name**?

What about moving of that element filtering code into some new implementation of existing `spoon.reflect.visitor.Filter` ? And then let client configure processor filter.

surli:
> Should it this new processor be in spoon-core or in spoon-maven-plugin?

IMHO it should be in spoon-core: you don't use the spoon-maven-plugin dependency when writing a processor for using it with the maven plugin... 

> This PR seems to me quite use-case specific (only for few clients). I am not using maven and not using Processors, so I am not the one who should decide, but is it really the case that most of the clients will need to process elements by type and occurrence of substring in element name?

Actually I see the problem in the opposite way: for now there's no way to completely reuse a custom Processor in different project, and to configure it by passing properties in a pom.xml. So for me this PR is a first step towards reusing processors in different contexts by passing them parameters in the pom.xml. 

I'm totally open about the design for doing that and I agree with you on the fact that here, I mostly want to be able to **filter** elements before passing them to the processor. 

So regarding all your comments, we may have two options: 
  * create a new processor let's say `FilterProcessor` which can be used with properties in the pom
   * create a filter and allow in the pom to specify which filter to use and its configuration

WDYT it's the best solution? @tmortagne do you have an opinion on that?

tmortagne:
> @tmortagne do you have an opinion on that?

Well all I was asking for was a way to have stuff end up in AbstractProcessor#loadProperties() when I use the Maven plugin, then I'm fine dealing with those properties myself :) It seems to me there is already a system to configure processors, what is missing is just a way to provide those properties from Maven plugin configuration.

What you propose is a lot more specific and while I agree it will be less work for me and probably better for performances IMO the important subject (pass custom properties to a processor) is a Maven plugin issue only.

surli:
> Well all I was asking for was a way to have stuff end up in AbstractProcessor#loadProperties() when I use the Maven plugin

Ok my bad, I was a bit quick when looking on the problem, and I wasn't aware about this method which seems not used anywhere in Spoon. 
Then I'll look on it and indeed it should be more a spoon-maven-plugin PR. I'll let this PR open for discussion for now.

tmortagne:
> I wasn't aware about this method which seems not used anywhere in Spoon

I did not saw it either at first but @monperrus told me about ProcessorProperties so I digged a bit to see how this was used and finally found that there was what looked like some properties injection system in AbstractProcessor.

monperrus:
To make it clear, should we change the title of this issue to "document how to inject processor configuration from the pom.xml and the Maven Plugin"?

tmortagne:
> To make it clear, should we change the title of this issue to "document how to inject processor configuration from the pom.xml and the Maven Plugin"?

It's not possible right now so it's not just documentation but it's on Maven plugin side.

monperrus:
discussion continues on #1534

tmortagne:
@monperrus should it be https://github.com/INRIA/spoon/issues/1537 ?

monperrus:
you're right

PR #1547, opened 2017-09-20
fix(test): wrong number of dependency MavenLauncherTest

surli:
The ant dependency was removed so the test should be changed


surli:
@monperrus Travis is broken because of a previous merge, could you merge this one quickly?

PR #1549, opened 2017-09-20
review: refactor: Launcher#buildModel return a CtModel

### LINKED ISSUE #1542: Launcher.buildModel() should return the CtModel
What do you think?

surli:
Fix #1542


monperrus:
add a test?

PR #1551, opened 2017-09-21
refactor: Move ElementPrinterHelper

pvojtechovsky:
* ElementPrinterHelper moved next to DJPP - it belongs there some like others.
* ListPrinter made package protected too


pvojtechovsky:
@monperrus WDYT?

monperrus:
indeed, thanks. 

the fact that it is not tested directly clearly shows that it was never part of the public API.

PR #1553, opened 2017-09-21
refactor: DefaultJavaPrettyPrinter better uses OperatorHelper

pvojtechovsky:
* DefaultJavaPrettyPrinter uses OperatorHelper instead of now deprecated PrinterHelper methods
* Some code of DefaultJavaPrettyPrinter is more friendly to future PrinterTokenWriter #1494


monperrus:
RevApi detects that the move of ListHelper is not finsihed, one needs to make `public ListPrinter createListPrinter` package-protected. Let's do it here?

monperrus:
if a method is not the API we can remove it without depreciation, so we can remove the ones in PrinterHelper.

monperrus:
here we clearly see the configuration that will be made possible with the indirection level of OperatorHelper.

I'm OK with it.

PR #1559, opened 2017-09-24
feat: add a static method to getting started with one single line of code

monperrus:


pvojtechovsky:
Nice! 

It is true that Launcher is not intuitive and I am often fighting with initial configuration of Spoon too.

src/main/java/spoon/Launcher.java
```diff
@@ -798,4 +800,12 @@ public CtModel getModel() {
 		return factory.getModel();
 	}
 
+	/** returns the AST of an inline class */
+	public static CtClass<?> parseClass(String code) {
```
surli: I'm not sure about your method return: I think this method will be quickly used with code containing multiple types (enum + class for example or multiple classes). 
Then I see two solutions:
1. Either the method return a list of CtTypes (and then we may have to rename it)
2. Either it checks that the model contains only one CtClass (and then we explictely do not accept interfaces or enum, and most importantly we give a comprehensive error for the user)

tdurieux: Or even code can contain only an interface...
I would return a CtType.

tdurieux:
> It is true that Launcher is not intuitive and I am often fighting with initial configuration of Spoon too.

Indeed, we need to improve the documentation of the Launcher.

src/main/java/spoon/Launcher.java
monperrus: The idea of this PR is to have the easiest possible introduction to Spoon. Having CtClass means that users would have directly getConstructors etc. which they would naturally expect (which is not in CtType). CtType is a slight less natural, more abstract concept. Once newcomers understand this method, they would obviously move to the complete configurable Launcher and the rest of the metamodel (incl CtType)

surli: I'm ok for keeping CtClass return but we should then check the type before casting, to help the user in case he did not pass a simple class.

pvojtechovsky: > new Launcher

may be yes! But only as a replacement/wrapper of old non intuitive Launcher. I do not think we should have many ways/launchers how to start processing. The one "well designed" launcher would be nice ... but I actually have no constructive suggestion how it should look like. I just have problems with the old Launcher.

surli: Maybe something as a LauncherBuilder: 

```java
CtModel model = LauncherBuilder().useAutoimport()
         .useNoclasspath()
         .addInputSource("./myDirectory/")
         .addInputSource("./anotherOne")
         .setOutputDir("./target")
         .build();
```

pvojtechovsky: yes exactly! But someone should analyze all Launcher use cases and organize them into intuitive API. I am not the one who knows use cases of Launcher well enough. Let's make an issue where we can build such a concept.

monperrus:
There is a confusion there.
The goal of this PR is to provide a one-liner intuitive introduction for newcomers.
Not to improve or redesign Launcher (which is a good idea, but for another PR)

surli:
> The goal of this PR is to provide a one-liner intuitive introduction for newcomers.

Yep, that's why we opened a new issue to discuss the design of a new launcher.

tdurieux:
> There is a confusion there.
The goal of this PR is to provide a one-liner intuitive introduction for newcomers.
Not to improve or redesign Launcher (which is a good idea, but for another PR)

For me the design of this PR is wrong. It will introduce more confusion in the API than solve issues.

PR #1561, opened 2017-09-26
fix: cleaning of method signature computation

### LINKED ISSUE #1483: Inconsistent executable signature on generic types
Given 

 ```Java
List<String> strings = new ArrayList<String>
strings.add("foo")
```
 and the resulting ctInvocation on the executable `Java.util.ArrayList#add(..)`,
spoon will return different signatures depending on the call chain of the signature retrieval:
1. `ctInvocation.executable.signature` -> `java.util.ArrayList#add(E)`
2. `ctInvocation.executable.executableDeclaration.signature` -> `boolean add(java.lang.Object)`
3. `ctInvocation.executable.executableDeclaration.reference.signature` -> `java.util.ArrayList#add(java.lang.Object)`

where 1 is an exec reference, 2 an exec declaration and 3 an exec reference again.

1 and 3 are inconsistent and should probably result in the object signature, not the generic one (like case 3).

### PR DESCRIPTION
**Pull request by monperrus**
We have problems regarding method signatures:

* 1) the current signatures of method references are completely different which is particularly confusing. One contains the declaring type which is about method links in Javadoc, not about signatures. 
* 2) There is no clear definition of method signatures in the Java specification (while usually the return type is not  considered as part of the signature, there are cases with generics where it actually matters).

This PR fixes 1) and makes a sensible choice for 2)

Fix #1483

### DIFF
--- a/src/main/java/spoon/reflect/visitor/ImportScannerImpl.java
+++ b/src/main/java/spoon/reflect/visitor/ImportScannerImpl.java
@@ -448,7 +448,9 @@ public class ImportScannerImpl extends CtScanner implements ImportScanner {
 	protected boolean isImportedInMethodImports(CtExecutableReference<?> ref) {
 		if (!(ref.isImplicit()) && methodImports.containsKey(ref.getSimpleName())) {
 			CtExecutableReference<?> exist = methodImports.get(ref.getSimpleName());
-			if ((exist.getDeclaringType().getQualifiedName() + "." + exist.getSignature()).equals(ref.getDeclaringType().getQualifiedName() + "." + ref.getSignature())) {
+			if ((exist.getDeclaringType() != null ? exist.getDeclaringType().getQualifiedName() : "" + "." + exist.getSignature()).equals(
+					ref.getDeclaringType() != null ? ref.getDeclaringType().getQualifiedName() : "" + "." + ref.getSignature())
+					) {
 				return true;
 			}
 		}
--- a/src/main/java/spoon/reflect/visitor/ImportScannerImpl.java
+++ b/src/main/java/spoon/reflect/visitor/ImportScannerImpl.java
@@ -448,7 +448,7 @@ public class ImportScannerImpl extends CtScanner implements ImportScanner {
 	protected boolean isImportedInMethodImports(CtExecutableReference<?> ref) {
 		if (!(ref.isImplicit()) && methodImports.containsKey(ref.getSimpleName())) {
 			CtExecutableReference<?> exist = methodImports.get(ref.getSimpleName());
-			if ((exist.getDeclaringType().getQualifiedName()+"."+exist.getSignature()).equals(ref.getDeclaringType().getQualifiedName()+"."+ref.getSignature())) {
+			if ((exist.getDeclaringType().getQualifiedName() + "." + exist.getSignature()).equals(ref.getDeclaringType().getQualifiedName() + "." + ref.getSignature())) {
 				return true;
 			}
 		}
--- a/src/main/java/spoon/reflect/visitor/ImportScannerImpl.java
+++ b/src/main/java/spoon/reflect/visitor/ImportScannerImpl.java
@@ -448,7 +448,7 @@ public class ImportScannerImpl extends CtScanner implements ImportScanner {
 	protected boolean isImportedInMethodImports(CtExecutableReference<?> ref) {
 		if (!(ref.isImplicit()) && methodImports.containsKey(ref.getSimpleName())) {
 			CtExecutableReference<?> exist = methodImports.get(ref.getSimpleName());
-			if ((exist.getType().getQualifiedName()+"."+exist.getSignature()).equals(ref.getType().getQualifiedName()+"."+ref.getSignature())) {
+			if ((exist.getDeclaringType().getQualifiedName()+"."+exist.getSignature()).equals(ref.getDeclaringType().getQualifiedName()+"."+ref.getSignature())) {
 				return true;
 			}
 		}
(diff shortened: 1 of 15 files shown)

### CODE REVIEW

**Comment by monperrus** [comment]
need #1562 first in order to strengthen the test case

**Comment by pvojtechovsky** [comment]
>  makes a sensible choice for 2)

how can client choose that? I see that return value is actually not printed only if it is undefined. 
May be there will be some clients, who will like signatures without return type even if return type is known - it is more fitting to: "usually the return type is not considered as part of the signature"

What about new method `getSignature(boolean includingReturnType)`? Or an global switch or hook which let's client register own signature printer (it is probably over-engineered?)

**Comment by monperrus** [comment]
> What about new method |getSignature(boolean includingReturnType)|?
I'm not in favor of this kind of dual-behavior method because there are hard to test, maintain and
understand.
> Or an global switch or hook which let's client register own signature printer (it is probably
> over-engineered?)
Interesting. That's an option for a future PR.

For now, I propose to keep this new behavior, and to reuse the existing signature. This new behavior
subsumes the older and also enables to know whether a type can be instantiated or not. In
particular, it makes it possible to test and enforce an interesting contract.

typeRef.isGenerics() == false <=> typeRef.getSignature()== typeRef.getDeclaration().getSignature()

WDYT?

**Comment by pvojtechovsky** [comment]
I looked at gumtree-spoon-ast-diff project today and the labels for nodes of methods are made using method.getSignature(). This PR will may be break this algorithm, because in this case it was correct to ignore type of returned value. Will there be an easy way to generate method signature without return type, which may be different and is really not relevant in many cases?

**Comment by monperrus** [comment]
> I looked at gumtree-spoon-ast-diff project today and the labels for nodes of methods are made
> using method.getSignature(). This PR will may be break this algorithm, because in this case it was
> correct to ignore type of returned value.
I'm not sure. It may do the opposite: make it work better :-)

I've just started Travis with a signature without the return type to see what tests are broken ...
or not.

**Comment by monperrus** [comment]
It seems to be doable to remove the return type from the signature. 
Just basic assertEquals to change. Only the failure of ImportTest.testWithInnerEnumDoesNotImportStaticInnerMethods is more difficult.

**Comment by monperrus** [comment]
ImportTest.testWithInnerEnumDoesNotImportStaticInnerMethods is also in fact easy to fix.

So it seems we can go for a getSignature without return type.

WDYT?

**Comment by pvojtechovsky** [comment]
> So it seems we can go for a getSignature without return type.

I can imagine that some clients will want to see return type and some wont ...
And I personally do not use getSignature - only for debugging, so I have no real use case.

Anyway I agree that this PR makes sense and Spoon should print signatures more consistent then till now. So I agree that your PR is a good way!

**Comment by monperrus** [comment]
when #1562 is merged, I'll improve the test, and then  if everybody's OK, we'll proceed then.

**Comment by monperrus** [comment]
ready for review

File: src/main/java/spoon/reflect/visitor/ImportScannerImpl.java [hunk 1]
```diff
@@ -448,7 +448,9 @@ protected boolean addMethodImport(CtExecutableReference ref) {
 	protected boolean isImportedInMethodImports(CtExecutableReference<?> ref) {
 		if (!(ref.isImplicit()) && methodImports.containsKey(ref.getSimpleName())) {
 			CtExecutableReference<?> exist = methodImports.get(ref.getSimpleName());
-			if (exist.getSignature().equals(ref.getSignature())) {
+			if ((exist.getDeclaringType() != null ? exist.getDeclaringType().getQualifiedName() : "" + "." + exist.getSignature()).equals(
```
pvojtechovsky: could you add brackets around `(x?y:z)`, so the precedence of operators is clear even for me ;-)
Or better to create private (or public?) method `getQualifiedSignature()'

src/main/java/spoon/support/template/SubstitutionVisitor.java
```diff
@@ -630,10 +631,12 @@ private String getParameterValueAsString(Object parameterValue) {
...
 	/*
-	 * cut the package name. We always convert types to simple names here
+	 * return the typical Javadoc style link Foo#method(). The class name is not fully qualified.
 	 */
-	private static String getShortSignature(String fullSignature) {
-		return fullSignature.substring(fullSignature.lastIndexOf('.') + 1);
+	private static String getShortSignatureForJavadoc(CtExecutableReference<?> ref) {
+		SignaturePrinter sp = new SignaturePrinter();
+		sp.writeNameAndParameters(ref);
+		return ref.getDeclaringType().getSimpleName() + CtExecutable.EXECUTABLE_SEPARATOR + sp.getSignature();
```
pvojtechovsky: What about new public class `MethodSignatureHelper`, which would implement
* getSignature(...)
* getJavaDocSignature(...)
* getQualifiedSignature(...)
* .,. ?

src/main/java/spoon/support/visitor/SignaturePrinter.java
```diff
@@ -38,6 +36,8 @@
 
 	private final StringBuffer signature = new StringBuffer();
```
pvojtechovsky: what about faster `StringBuilder` ?

src/test/java/spoon/test/comment/CommentTest.java
```diff
@@ -686,7 +686,7 @@ public void testSnippedWithComments(){
 		assertTrue(clazz1==builder.getSnippetCompilationUnit().getDeclaredTypes().get(0));
 
 		CtMethod<?> methodString = (CtMethod<?>) clazz1.getMethods().toArray()[0];
-		assertEquals("java.io.File foo(java.lang.String)", methodString.getSignature());
+		assertEquals("foo", methodString.getSimpleName());
```
pvojtechovsky: Why this change?  methodString.getSignature() fails here?

pvojtechovsky:
This implementation is OK for me. The comments above are optional.

monperrus:
That could be interesting, yes.

monperrus:
thanks, changed.

pvojtechovsky:
Thank You Martin for this fix!

PR #1562, opened 2017-09-27
feat: add support for analyzing whether a type reference is generic or not (method isGeneric)

monperrus:


pvojtechovsky:
Hi Martin, there is already `CtTypeInformation#isGenerics()` with similar contract. May be we should implement your new method as `CtTypeInformation#isGenerics(boolean recursive)` ... or at least use more expressive name, because actually there are
* boolean isGenerics()
* boolean isGeneric()

with different behavior

monperrus:
You're right thanks.

For now, I propose to keep this new behavior, and to reuse the existing signature. 
This new behavior subsumes the older and also enables to know whether a type can be instantiated or not. In
particular, it makes it possible to test and enforce an interesting contract.

typeRef.isGenerics() == false <=> typeRef.getSignature()== typeRef.getDeclaration().getSignature()

WDYT?

pvojtechovsky:
> For now, I propose to keep this new behavior, and to reuse the existing signature.

do you mean to change behavior of `isGenerics()`? 

I checked how this method is used in spoon and it is only needed in two tests. 
It is OK for me to change this behavior.

What about adding another method `isParametrized()`, with this contract

Type | isParametrized() | isGenerics() 
------|-------------------|----
Integer | false | false
List\<Integer\> | true | false
List\<T\> | true | true

At the beginning I had problem to understand your change because I understood the behavior as `isParametrized`

monperrus:
>  It is OK for me to change this behavior.

Cool, then we may merge.

> What about adding another method isParametrized(), with this contract

Good idea! https://github.com/INRIA/spoon/issues/1565

Issue #1567, opened 2017-09-29
refactor: get rid of snapshotLength and hasNewContent


monperrus:
Spin off of #1494 

SnapshotLength and hasNewContent  reflect a design problem: with those methods, DJPP depends on the printing state to reason about what to do. We have a dedicated concept for this, the PrintingContext. It now seems to me that we should remove snapshotLength and hasNewContent (whether char-based or token-based) and add some logic in PrintingContext instead.

pvojtechovsky:
I checked the code which uses hasNewContent method and I have not found an easy and maintenable way how to get rid of it. The only nice solution I see is to observe the TokenWriter and to detect whether some tokens were fired (internal SnapshotTokenWriter). That would be an easy and clean solution from my point of view.

@monperrus please provide some ideas how to do it different so we can finish TokenWriter and fix remaining bugs.

monperrus:
> and fix remaining bugs.
I would not call this is bug. I would call it a feature that is not yet implemented. However, I'm
not sure to perfectly understand your use case? What do you need exactly that cannot be done yet?

> @monperrus <https://github.com/monperrus> please provide some ideas how to do it different
For instance, for removing the calls to snapshot in printCtFieldAccess, to add methods in the
context and replace

printer.hasNewContent() by context.typeNameStack() == stackBefore.

pvojtechovsky:
> What do you need exactly that cannot be done yet?

If I run the PrinterTest#testPrinterTokenListener on all classes of spoon model, then it fails on duplicate separator token. Why that? Because DJPP produces wrong tokens if PrinterHelper is not filled by characters - OK this is missing feature, but this missing feature causes that we never came through all spoon sources and therefore there is chance that we produce wrong tokens - different bug then one mentioned above (and this I call bug) - even on spoon sources itself. The best would be try the same test on sources of other complicated libraries too.

> printer.hasNewContent() by context.typeNameStack() == stackBefore.

I do not understand this idea. If `typeNameStack` is a stack then it will be always `==stackBefore` because the `scan()` method was already finished. But may be do not try to explain it. Simply implement it ;-)

monperrus:
https://github.com/INRIA/spoon/pull/1571 is a solution to the token stream problem. testPrinterTokenListener now goes over src/main/java

> But may be do not try to explain it. Simply implement it ;-)

Would love to do so, but requires precious time first

PR #1571, opened 2017-10-02
fix: DJPP sends a correct stream of tokens

monperrus:


pvojtechovsky:
I like this PR as a check that there is no other bug in DJPP token writing found-able by printing of spoon sources.

But I do not think we should really merge this PR, because 
1) it introduces "fake separator", which is otherwise confusing and should not be printed by DJPP in future.
2) DJPP really does not send correct stream of tokens. It just found a way how to pass the test, which is commented at the end, because the main contract still does not work.

So I suggest to close this PR and I will also keep silence and wait until you have time to fix #1567 (or you accept that my simple solution of internal SnapshotTokenWriter is not so bad).

monperrus:
Agree.

PR #1574, opened 2017-10-05
fix: DefaultJavaPrettyPrinter prints tokens correctly

pvojtechovsky:
This is solution for #1567 - token-based snapshotLength and hasNewContent internal implementations of DJPP, which are independent on implementation of client's TokenWriter.


pvojtechovsky:
May be ... somebody ... not influenced by Martin ... will merge it "by mistake" ;-)

pvojtechovsky:
There is better solution #1576

PR #1578, opened 2017-10-06
fix: the query engine is broken on some JVM because it relies on stack trace elements

monperrus:
The query engine is broken on some JVM (eg Oracle 1.8.40), because it relies on stack trace elements, and the stack trace elements change on some JVM.

The problem is that `StackTraceElement stackEle = stackEles[0]` is not necessarily the right one (in Oracle 1.8.40 the stack frame from an anonymous lambda is not there).

This PR removes this brittleness and has to change a little bit the specification when queries or functions throw a ClassCastException

### DIFF
--- a/src/main/java/spoon/reflect/visitor/chain/CtQueryImpl.java
+++ b/src/main/java/spoon/reflect/visitor/chain/CtQueryImpl.java
@@ -399,49 +399,15 @@ public class CtQueryImpl implements CtQuery {
 		 * @param parameters
 		 */
 		protected void onClassCastException(ClassCastException e, String exceptionMessage, Object input) {
-			if (isFailOnCCE() || expectedClass != null) {
+			if (isFailOnCCE()) {
 				//expected class is known and it was checked, so the CCE must be thrown by something else. Report it
-				throw new SpoonException(exceptionMessage == null ? getStepDescription(this, e.getMessage(), input) : exceptionMessage, e);
+				// we want to point to this line of code
+				e.setStackTrace(new Exception().getStackTrace());
+				throw e;
 			}
-			StackTraceElement[] stackEles = e.getStackTrace();
-			StackTraceElement stackEle = stackEles[0];
-			if (stackEle.getMethodName().equals(cceStacktraceMethodName) && stackEle.getClassName().equals(cceStacktraceClass)) {
-				//the CCE exception was thrown in the expected method - OK, it can be ignored
-				detectExpectedClassFromCCE(e, input);
-				log(this, e.getMessage(), input);
-				return;
-			}
-			//Do not ignore this exception in client's code. It is not expected. It cannot be ignored.
-			throw new SpoonException(exceptionMessage == null ? getStepDescription(this, e.getMessage(), input) : exceptionMessage, e);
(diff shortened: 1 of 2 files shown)


pvojtechovsky:
Could you provide a test case where origin query engine failed?

Please do not merge this PR too fast. I would like to analyze and discuss the impact of this change deeper.

pvojtechovsky:
OK, I have reproduced it with standard Spoon test cases on Oracle 1.8.40.

pvojtechovsky:
> This PR removes this brittleness and has to change a little bit the specification when queries or functions throw a ClassCastException

I agree that looking into Stacktrace and parsing of Exception message is really not the best practice. 

Note: that this is necessary only for handling of lambda expressions. So callbacks implemented by old way (without lambda), should behave same like before - should still report CCE in client's buggy code. There is no reason to get rid of this correct behavior. So please fix it and rollback related tests.

It is just question how to handle Lambdas
A) to fix brittle code to be less brittle and to have fast query engine, which reports CCE in client;s buggy code correctly - see #1579 
B) to remove brittle code (as done in Your PR) and to have slow query engine in case of nice lambdas and to hide CCE in client's buggy code.
C) to found some cleaner way how to detect, what is the type of parameter lambda expression.

monperrus:
> A) to fix brittle code to be less brittle and to have fast query 
> engine, which reports CCE in client;s buggy code correctly - see #1579 
> <https://github.com/INRIA/spoon/pull/1579>
> B) to remove brittle code (as done in Your PR) and to have slow query 
> engine in case of nice lambdas and to hide CCE in client's buggy code.
Yes, that's the choices. I would prefer B) because it's simpler to 
understand and maintain (as you know, I'm a strong advocate of KISS), 
and 2) we can be quite confident that in works for all JVMs, even the 
exotic ones, and across all versions, while this is unknown for A).

pvojtechovsky:
>  we can be quite confident that it works for all JVMs

this can be fixed with little effort for A too. It is possible to detect whether JVM behaves as expected and to have this optimization ON and in opposite case we can fallback to B) => slow throwing of CCEs and to eating of exceptions in clients code - to make it more tricky for clients - but working correctly on old and exotic JVMs.

And you already mentioned somewhere: "It is no problem to write more code on Spoon side, if it will be better for clients." And it is this case: A) is more code, but 
1) it is 20 times faster. As mentioned in #1541 the MainTest#testMain in debug mode before optimization needed 19 minutes on my PC. After optimization, which avoided CCEs it needs only 52s. 
2) it correctly throws exceptions produced by client's code.

While B)
1) is slower
2) it silently eats ClassCastException thrown in client's code and clients will spent hours on detecting, why their algorithm does not work ... and after creation of spoon issue and discussion with us (we spent some ours on that too) we may be found that there is CCE in client's code, caused by client's mistake. Eating of client's exceptions is fatal bug from my point of view.

monperrus:
> And you already mentioned somewhere: "It is no problem to write more code on Spoon side, if it will be better for clients.

I would still sign this statement :-)

Let's discuss your alternative fix then.

monperrus:
#1579 is the final solution

PR #1579, opened 2017-10-07
alternative fix: the query engine is broken on some JVM because it relies on stack trace elements

pvojtechovsky:
This is an alternative solution to #1578

### DIFF
--- a/src/main/java/spoon/reflect/visitor/chain/CtQueryImpl.java
+++ b/src/main/java/spoon/reflect/visitor/chain/CtQueryImpl.java
@@ -307,7 +307,7 @@ public class CtQueryImpl implements CtQuery {
 			try {
 				result = _accept(input);
 			} catch (ClassCastException e) {
-				onClassCastException(e, getErrorMessage(), input);
+				onClassCastException(e, input);
 				return;
 			}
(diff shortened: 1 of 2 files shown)


pvojtechovsky:
Now there is solution, which detects whether detection based on CCE works on current JVM. 
If it works then future CCEs are avoided and client's CCEs are correctly rethrown.
If JVM is old or exotic, then it falls back to slow throwing of CCEs and to wrong ignoring of all exceptions.

src/main/java/spoon/reflect/visitor/chain/CtQueryImpl.java
```diff
@@ -552,4 +533,61 @@ protected void handleResult(Object result, Object input) {
 			nextStep.accept(result);
 		}
 	}
+
+	private static Integer indexOfCallerInStack = null;
```
monperrus: remove mutability?

private final static Integer indexOfCallerInStack = getIndexOfCallerInStackOfLambda();

pvojtechovsky: ok, done

monperrus:
thanks!

PR #1588, opened 2017-10-11
review: fix: ImportBuilder to check visibility when importing starred imports

### LINKED ISSUE #1587: Bug: error with star imports and visibilities of classes
Hi all,

Spoon doesn't check the visibility of classes when build imports, that results in error if Spoon wants to replace the star `*` by an non-public class.

Cheers.

surli:
Fix #1587

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/JDTImportBuilder.java
+++ b/src/main/java/spoon/support/compiler/jdt/JDTImportBuilder.java
@@ -78,7 +78,9 @@ class JDTImportBuilder {
 
 					if (ctPackage != null) {
 						for (CtType type : ctPackage.getTypes()) {
-							this.imports.add(type.getReference());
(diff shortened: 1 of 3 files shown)


src/main/java/spoon/support/compiler/jdt/JDTImportBuilder.java
```diff
@@ -78,7 +78,9 @@ public void build() {
 
 					if (ctPackage != null) {
 						for (CtType type : ctPackage.getTypes()) {
-							this.imports.add(type.getReference());
+							if (type.getVisibility() == ModifierKind.PUBLIC) {
```
monperrus: should also work with package-visible if both classes are in the same package?

surli: Yes, but if both classes are in the same packages, it's useless to import them...

PR #1589, opened 2017-10-11
review fix: Bug related with cloning newClass invocation with anonymous class

surli:


surli:
This bug is a bit tricky: when cloning an element, the newly cloned element does not have a parent. 
Then when printing it in autoimport mode, the code could crash because it's using the context of the class to know if it can print fully qualified name or not, depending on what's imported or not. 

Here it crashes because the class is anonymous, and after cloning it does not have a fully qualified name, only a simple name (i.e. 1 instead of spoon.test.ctClass.testclasses.AnonymousClass.1) then getting the actualClass it throws a SpoonNotFoundException.

I don't know what is exactly the right way to fix the bug: to put a package name when cloning a CtClass? To throw a clear exception when trying to print a cloned element without parent? ...

pvojtechovsky:
As a client I would expect that such cloned class can be printed - so I can see what is inside the cloned model. So throwing of the exception during printing is probably not wanted.
Then there is question what should be the class name
A) unknown.Unknown.1
B) 1
C) spoon.test.ctClass.testclasses.AnonymousClass.1
D) ?

It seems to me that C) is the most expected concept: "If I clone something, I get the copy of the origin, so it looks exactly like origin."

So I vote for: "Cloning process should set parent of cloned element to the same element like parent of origin node."

surli:
> So I vote for: "Cloning process should set parent of cloned element to the same element like parent of origin node."

I completely agree with you but I think there was client that are/were using cloning feature with the intent to not attach it to the previous model.  Any hint on that @monperrus ?
So could we break some clients by changing this behaviour? 

The idea of throwing an exception was, from my point of view, a middle path: we throw an exception, then the client can see he must attach it to the model for his purpose. 

WDYT?

monperrus:
> there was client that are/were using cloning feature with the intent to not attach it to the previous model

Indeed. The original should never be impacted (neither at the model level nor at the pretty-printing level) when one of its elements is cloned.

So we may copy the parent if and only if it does not impact the parent. Or an alternative would be to clone the parent when necessary, such as in this case.

pvojtechovsky:
> The original AST should never be impacted 

I agree. But I do not understand how is the parent impacted if clone of children is made and then setParent is called on their root. setParent does not automatically add child into parent. Or am I wrong?

monperrus:
> setParent does not automatically add child into parent. Or am I wrong?
yes, I have just verified, see https://github.com/INRIA/spoon/pull/1591

so it is safe to add the previous parent as the parent to the cloned 
element.

surli:
> yes, I have just verified, see #1591

Ok, so do we agree to set the parent of the previous element automatically when cloning?

pvojtechovsky:
I agree

tdurieux:
> Ok, so do we agree to set the parent of the previous element automatically when cloning?

I'm pretty sure that we had this behavior before but we remove it, I don't remember why.

surli:
> I'm pretty sure that we had this behavior before but we remove it, I don't remember why.

If there was a good reason, I hope somebody puts a test to show it ;)

surli:
So, there is a test to check that a clone should not have an initialized parent: https://github.com/INRIA/spoon/blob/79c399223da0f0735f450220c30e6e24d6d86a72/src/test/java/spoon/test/factory/FactoryTest.java#L47

The test is 4 years old, so I don't exactly know why there was such a contract...

pvojtechovsky:
I vote to change the test ;-) ... and in case of some problems we will know what was the reason ;-)

surli:
@pvojtechovsky could you have on CtBodyHolderTest#checkCtBody: https://github.com/INRIA/spoon/blob/e204d072e77cbef5a46ff2bc8c9acfe7bee67371/src/test/java/spoon/test/ctBodyHolder/CtBodyHolderTest.java#L112

You create this tests, and apparently you check after applying a template if an element has no parent. With the new contract this test is broken too.
I can fix it by removing those checks, but I want to be sure we do not break anything in template contracts.

pvojtechovsky:
You can remove the checks that parent is null after template or clone. I guess 2 of 3 checks can be removed. One can stay because element is no made by clone, so parent is really null.
I do not remember why I did these checks.

surli:
Don't pay attention about revapi report: we have problems with CI today.

monperrus:
> The test is 4 years old, so I don't exactly know why there was such a contract...

This behavior is important for having a natural way of testing whether an element comes from an existing AST or has been freshly created or cloned. We should probably fix this bug in another way (eg where the exception is thrown) and keep this behavior.

pvojtechovsky:
> This behavior is important for having a natural way of testing whether an element comes from an existing AST or has been freshly created or cloned.

I see it as conflict between client´s needs and testing needs. From my point of view client should win.

For testing we can implement new method `boolean CtElement#isOrphan()`, which would check whether parent holds this element.

pvojtechovsky:
> For testing we can implement new method boolean CtElement#isOrphan(), which would check whether parent holds this element.

Or may be even better and useful method: `CtRole CtElement#getMyRoleInParent()`, which would return the role of this element in parent element and return null in case of cloned element, because such element is not referred by any parent role.

This method can be implemented using new scanner mentioned in #1614 
This method together with support in RoleHandler would allow much simpler implementation of Element#delete() (ReplacementVisitor), etc....

tdurieux:
`CtRole CtElement#getMyRoleInParent()` that is a really cool feature

surli:
For now I simply put a parent for CtNewClass elements, in order to fix the bug. For the long term view I created #1658

src/test/java/spoon/test/ctBodyHolder/CtBodyHolderTest.java
```diff
@@ -124,12 +119,7 @@ private void checkCtBody(CtBodyHolder p_bodyHolder, String p_constant, int off)
 
 		//try to set CtBlock and get the same CtBlock
 		CtStatement newStat2 = newStat.clone();
-		try {
-			newStat2.getParent();
-			fail();
```
monperrus: undo?

monperrus:
closing after agreement on #1667.

thanks @surli for doing the hard analysis and testing work

PR #1582, opened 2017-10-08
review: feat: Spoon CtRole based attributes access

pvojtechovsky:
Early implementation of spoon model CtRole based access to element attributes. See discussion in #1581

PR #1600, opened 2017-10-15
fix(role): CtPackage.types is CtRole.CONTAINED_TYPE

pvojtechovsky:
Here is the change of role of  CtPackage.types as suggested by @monperrus in #1582

### DIFF
--- a/src/main/java/spoon/reflect/declaration/CtPackage.java
+++ b/src/main/java/spoon/reflect/declaration/CtPackage.java
@@ -20,11 +20,12 @@ import spoon.reflect.reference.CtPackageReference;
 import spoon.support.DerivedProperty;
 import spoon.reflect.annotations.PropertyGetter;
 import spoon.reflect.annotations.PropertySetter;
+import spoon.reflect.path.CtRole;
 
 import java.util.Set;
 
 import static spoon.reflect.path.CtRole.SUB_PACKAGE;
-import static spoon.reflect.path.CtRole.TYPE;
+import static spoon.reflect.path.CtRole.CONTAINED_TYPE;
 
 /**
  * This element defines a package declaration. The packages are represented by a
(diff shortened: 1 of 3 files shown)


tdurieux:
please also change this: 

https://github.com/pvojtechovsky/spoon/blob/5ee803799a84c693fb7da377ece094394a168628/src/main/java/spoon/support/reflect/declaration/CtPackageImpl.java#L196

https://github.com/pvojtechovsky/spoon/blob/5ee803799a84c693fb7da377ece094394a168628/src/main/java/spoon/support/reflect/declaration/CtPackageImpl.java#L218

https://github.com/pvojtechovsky/spoon/blob/5ee803799a84c693fb7da377ece094394a168628/src/main/java/spoon/support/reflect/declaration/CtPackageImpl.java#L218

pvojtechovsky:
@monperrus Will we refactor each role change individually in own PR or can I make one bigger PR for that?

Individual PRs are problem, because all of them are changing CtRole class, which then causes conflicts.  So I would prefer to make one PR, which will contain all the CtRole improvements we see now.

WDYT?

pvojtechovsky:
By the way, all other roles have singular form, while this new role is plural. Should we rename 
CtRole.CONTAINED_TYPES
to
CtRole.CONTAINED_TYPE

WDYT?

monperrus:
What about a strong contract: all roles which return a collection/list/set are plural?

pvojtechovsky:
> What about a strong contract: all roles which return a collection/list/set are plural?

I vote for usage "singular always". The meaning of role itself should not contain information about how many items there is.

monperrus:
Actually, you convince me. I now think that always singular is a good idea to make it clear that it is a role and not property/field.

So let's make them all singular.

pvojtechovsky:
I am finished here. It is ready for merge.

PR #1603, opened 2017-10-15
WIP: Refactoring: Parameterize constants with role names

seintur:
Associate constants with role names to factorize the code of the forName factory method.

### DIFF
--- a/src/main/java/spoon/reflect/path/CtRole.java
+++ b/src/main/java/spoon/reflect/path/CtRole.java
@@ -16,40 +16,44 @@
  */
 package spoon.reflect.path;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
 /**
  * Created by nicolas on 27/08/2015.
  */
 public enum CtRole {
-	NAME,
-	TYPE,
-	DECLARING_TYPE,
-	BODY,
+	NAME("name", "simplename"),
+	TYPE("returntype", "componenttype"),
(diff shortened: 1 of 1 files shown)


pvojtechovsky:
By the way, why we need synonyms for role names? Backward comaptibility? More friendly for clients? If there is not goodanswer for that, then lets remove these synonyms.

src/main/java/spoon/reflect/path/CtRole.java
```diff
@@ -59,25 +63,31 @@
...
 	ARGUMENT,
-	SUPER_TYPE,
+	SUPER_TYPE("superclass"),
 	NESTED_TYPE,
 	CONSTRUCTOR,
 	EXECUTABLE,
-	FIELD,
-	CAST,
-	VALUE,
+	FIELD("typemembers"),
```
pvojtechovsky: executables and nested types are type members too ... it probably needs further refactoring ... but after tjis PR.

monperrus:
> If there is not goodanswer for that, then lets remove these synonyms.
I agree. We should remove them. CtPath should simply use Rol.tolowercase().

This is not backward compatible but AFAIK nobody uses CtPath so far (it is not comprehensive  and
too experimental).

surli:
> We should remove them. CtPath should simply use Rol.tolowercase().

Does this mean we should close this PR? I fixed the conflict in case.

monperrus:
closed per #1702 

thanks @seintur for bringing this on the table.

PR #1610, opened 2017-10-17
Review: feat: be able to change the destination of each created file

tdurieux:
What do you think?

The idea of this is to be able to override DefaultOutputDestination and defines your output strategy in `getOutputPath(File defaultOutputDirectory, Path packagePath, String filename)`

### DIFF
--- a/src/main/java/spoon/compiler/Environment.java
+++ b/src/main/java/spoon/compiler/Environment.java
@@ -26,7 +26,7 @@ import spoon.processing.Processor;
 import spoon.processing.ProcessorProperties;
 import spoon.reflect.declaration.CtElement;
 import spoon.reflect.declaration.CtMethod;
-import spoon.support.OutputDestination;
+import spoon.support.OutputDestinationHandler;
 
 import java.io.File;
 import java.nio.charset.Charset;
@@ -348,12 +348,12 @@ public interface Environment {
 	/**
 	 * Set the output destination that handles where source files are written
 	 */
-	void setOutputDesination(OutputDestination outputDestination);
+	void setOutputDestinationHandler(OutputDestinationHandler outputDestinationHandler);
 
 	/**
 	 * Returns the output destination that handles where source files are written
 	 */
-	OutputDestination getOutputDestination();
+	OutputDestinationHandler getOutputDestinationHandler();
 
 	/**
 	 * get the model change listener that is used to follow the change of the AST.
--- a/src/main/java/spoon/compiler/Environment.java
+++ b/src/main/java/spoon/compiler/Environment.java
@@ -18,6 +18,7 @@ package spoon.compiler;
 
 import org.apache.log4j.Level;
 import spoon.OutputType;
+import spoon.experimental.modelobs.FineModelChangeListener;
 import spoon.processing.FileGenerator;
 import spoon.processing.ProblemFixer;
 import spoon.processing.ProcessingManager;
@@ -25,7 +26,7 @@ import spoon.processing.Processor;
 import spoon.processing.ProcessorProperties;
 import spoon.reflect.declaration.CtElement;
 import spoon.reflect.declaration.CtMethod;
-import spoon.experimental.modelobs.FineModelChangeListener;
+import spoon.support.OutputDestination;
 
 import java.io.File;
 import java.nio.charset.Charset;
@@ -343,6 +344,16 @@ public interface Environment {
 	 * Returns the directory where source files are written
 	 */
 	File getSourceOutputDirectory();
+
+	/**
+	 * Set the output destination that handles where source files are written
+	 */
+	void setOutputDesination(OutputDestination outputDestination);
+
+	/**
+	 * Returns the output destination that handles where source files are written
+	 */
+	OutputDestination getOutputDestination();
 
 	/**
 	 * get the model change listener that is used to follow the change of the AST.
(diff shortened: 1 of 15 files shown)


src/main/java/spoon/support/OutputDestination.java
```diff
@@ -0,0 +1,34 @@
...
+
+import spoon.reflect.declaration.CtPackage;
+import spoon.reflect.declaration.CtType;
+
+import java.io.File;
+import java.nio.file.Path;
+
+public interface OutputDestination {
+
+	Path getOutputPath(File defaultOutputDirectory, Path packagePath, String filename);
```
monperrus: Path, Path?

src/main/java/spoon/support/OutputDestination.java
```diff
@@ -0,0 +1,34 @@
...
+
+	Path getOutputPath(CtPackage type, File defaultOutputDirectory,
```
monperrus: additional arg at the end?

pvojtechovsky:
I like that idea.
Note: Interface OutputDestination with 3 methods cannot be implemented using lambda. May be you can implement two of them as `default` methods directly in interface?

tdurieux:
I think three methods is too much.
I propose to only have
`getOutputPath(CtElement element, File defaultOutputDirectory, Path packagePath, String filename)`.

The users that will use this interface should be able to do an instanceof .

WDYT?

pvojtechovsky:
What about method like this:
```
Path getOutputPath(CtPackage package, CtType type, File defaultOutputDirectory, Path packagePath, String filename)
```
the `type` might be sometime null

monperrus:
ping?

tdurieux:
I have an issue because the output destination is stored in two places in the:

* ModelBuilder
* FileGenerator

I think it is an issue, but I don't know which one is the best to keep

surli:
> I think it is an issue, but I don't know which one is the best to keep

I propose to put the output destination in Environment and to only store it there. Then other classes could get it through Environment. WDYT?

pvojtechovsky:
>to put the output destination in Environment

sounds good for me

monperrus:
rebase now that we have #1770 ?

monperrus:
travsi fails? @tdurieux ?

tdurieux:
no there is still a lot of work to do.
The current implementation is not nice, way to much state

tdurieux:
I finished to clean up this PR.

But I'm not sure if the interface and the implementation of `OutputDestination` is good enough.
I made `DefaultOutputDestination` in the way that it is easy to overwrite.
Probably, the method that will be overwritten by the client is `Path getDirectoryPath(CtModule module, CtPackage pack, CtType type)` which returns the root path of the destination.

src/main/java/spoon/compiler/Environment.java
```diff
@@ -344,6 +345,16 @@ void report(Processor<?> processor, Level level,
 	 */
 	File getSourceOutputDirectory();
 
+	/**
+	 * Set the output destination that handles where source files are written
+	 */
+	void setOutputDesination(OutputDestination outputDestination);
```
surli: The name is a bit confusing with `sourceOutputDirectory`, and for me it looks like a strategy for the final location of the output. So WDYT about renaming it to something like `setOutputStrategy` or `setOuputHandler`?

src/main/java/spoon/support/JavaOutputProcessor.java
```diff
@@ -217,38 +217,32 @@ private void createModuleFile(CtModule module) {
...
-		if (pack.isUnnamedPackage()) {
-			packageDir = this.getModuleFile(pack.getDeclaringModule());
-		} else {
-			// Create current package dir
-			packageDir = new File(this.getModuleFile(pack.getDeclaringModule()).getAbsolutePath() + File.separatorChar + pack.getQualifiedName().replace('.', File.separatorChar));
-		}
-		if (!packageDir.exists()) {
-			if (!packageDir.mkdirs()) {
+	private Path getElementPath(CtPackage type) {
+		return createFolders(getEnvironment().getOutputDestination()
```
surli: Why don't you delegate the creation of the folder to the output destination strategy? We can imagine different implementation e.g. a strategy for not creating new directory, or for not overwritten existing files.

src/main/java/spoon/compiler/Environment.java
tdurieux: I agree

src/main/java/spoon/support/JavaOutputProcessor.java
surli: > it is not exactly the same responsibility, what you ask is more something like, isToPrint, not whereToPrint

In fact I see the `OutputDestinationStrategy` as the main handler in Spoon to manage the pathes of the output. In that sense, its responsibility is indeed, not only to say where to print, but also to say if it's ok to print things at this specific location. 
The idea behind it is that you only need to provide your own implementation of `OutputDestinationStrategy` whenever you need a specific behaviour related to path, beeing where, if or how.

tdurieux: I don't know how to communicate with the output processor. The use case is not clear to me.

surli: Maybe just rename `getOutputPath` to `createOutputPath`, and inside this method you create the needed folders?..

surli:
I changed some names so for me it's ready to merge @tdurieux @monperrus

monperrus:
That's good. Could you add a test case for this new feature?

surli:
test added @monperrus

monperrus:
thanks a lot!

PR #1586, opened 2017-10-10
feat(role): add CtRole to CtReference

tdurieux:

PR #1613, opened 2017-10-17
doc: clarifies contracts of @DerivedProperty

monperrus:
per our discussion on #1586,  PropertyGetter can be used for @DerivedProperty as well

### DIFF
--- a/src/main/java/spoon/support/DerivedProperty.java
+++ b/src/main/java/spoon/support/DerivedProperty.java
@@ -23,10 +23,16 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
-/** Tells that a metamodel property is derived, ie computed from the value of another property.
+/**
+ * Tells that a metamodel property is derived, ie computed from the value of another property.
+ *
  * For instance {@link CtType#getFields()}  is derived from {@link CtType#getTypeMembers()}
  *
- * This annotation is used for specifying CtScanner: derived properties are never scanned.
+ * It can be put on getter ond setters.
+ *
+ * Contracts:
+ * - A setter with @DerivedProperty never triggers a model intercession event.
+ * - A getter with @DerivedProperty is never called in CtScanner and derived classes (clone, replace)
  */
 @Retention(RetentionPolicy.RUNTIME)
 @Target({ ElementType.METHOD })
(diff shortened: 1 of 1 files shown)


pvojtechovsky:
I do not understand purpose of this change. But I do not need to know everything. I just cannot review this PR. I personaly would keep this annotation in the model. It had a meening

src/main/java/spoon/support/reflect/code/CtInvocationImpl.java
```diff
@@ -175,6 +176,8 @@ public void replace(CtStatement element) {
 		if (type != null) {
 			type.setParent(this);
 		}
+		// even if the setter delegates to getExecutable, it makes sense to create an event
+		getFactory().getEnvironment().getModelChangeListener().onObjectUpdate(this, TYPE, type, getExecutable().getType());
```
pvojtechovsky: So model change listener will be called twice for single modification? I do not like it. 
@monperrus Could you explain why it is good?

monperrus: you're right, it's not good.

monperrus:
The most convincing reason to keep @DerivedProperty is in ASTChecker with this contract.

Contract: a setter with @DerivedProperty never triggers a model intercession event.

So let's make this PR only a clarification one.

src/main/java/spoon/support/DerivedProperty.java
```diff
@@ -23,10 +23,16 @@
...
 import java.lang.annotation.Target;
 
-/** Tells that a metamodel property is derived, ie computed from the value of another property.
+/**
+ * Tells that a metamodel property is derived, ie computed from the value of another property.
+ *
  * For instance {@link CtType#getFields()}  is derived from {@link CtType#getTypeMembers()}
  *
- * This annotation is used for specifying CtScanner: derived properties are never scanned.
+ * It can be put on getter ond setters.
```
pvojtechovsky: ond -> and

src/main/java/spoon/support/DerivedProperty.java
```diff
@@ -23,10 +23,16 @@
...
+ *
+ * Contracts:
+ * - A setter with @DerivedProperty never triggers a model intercession event.
```
pvojtechovsky: > intercession event

Do you mean something like this?
```java
class CtConstructorCallImpl
...
	@DerivedProperty
	public <T extends CtActualTypeContainer> T addActualTypeArgument(CtTypeReference<?> actualTypeArgument) {
		if (getExecutable() != null) {
			getExecutable().addActualTypeArgument(actualTypeArgument);
		}
		return (T) this;
	}
```
It triggers the event, but not directly on the `this` node, but on some delegate? Did I understood it well? Improve description?

Is there a test for setters which does not start with `set` prefix, but with add/remove?

monperrus:
clarified contract?

PR #1620, opened 2017-10-19
review: fix: JDTImportBuilder should not import all public method from Interfaces

### LINKED ISSUE #1628: bug: import of interface class in JDTImportBuilder
I don't understand the `klass.isInterface()` in this line:

https://github.com/INRIA/spoon/blob/4ed07f7395ad0b46a0ec360388fe5a0be6f98ac0/src/main/java/spoon/support/compiler/jdt/JDTImportBuilder.java#L115

For me we should never import method from interfaces, it means nothing.
Do I miss something?

surli:
fix #1628


monperrus:
test case?

monperrus:
thanks!

PR #1623, opened 2017-10-19
refactor(role): CtCodeSnippet has CtRole.SOURCE_CODE

pvojtechovsky:
This PR introduces new CtRole.SOURCE_CODE, which is used for value of CtCodeSnippet. Before there was CtRole.EXPRESSION, which has different meaning.


monperrus:
it makes indeed much sense. I would rename it to "SNIPPET" to be even more explicit.

pvojtechovsky:
Nice name! Done.

Issue #1614, opened 2017-10-17
CtScanner which provides CtRole of scanned element


pvojtechovsky:
It would be nice if CtScanner (or similar class) would report the role of the scanned element.
The code would simply look like this:
```java
class CtScanner {
	public void scan(Object o, CtRole role) {
		...
	}
...
	public <T> void visitCtClass(final CtClass<T> ctClass) {
		enter(ctClass);
		scan(ctClass.getAnnotations(), CtRole.ANNOTATION);
		scan(ctClass.getSuperclass(), CtRole.SUPER_TYPE);
		scan(ctClass.getSuperInterfaces(), CtRole.INTERFACE);
		scan(ctClass.getFormalCtTypeParameters(), CtRole.TYPE_PARAMETER);
		scan(ctClass.getTypeMembers(), CtRole.TYPE_MEMBER);
		scan(ctClass.getComments(), CtRole.COMMENT);
		exit(ctClass);
	}
...
}
```

I am opening this issue NOW, because it might influence the design of CtRole enum values.

Note: `scan(ctClass.getTypeMembers(), CtRole.TYPE_MEMBER);`

Here is again visible the conflict in current CtRole concept. In this case the roles should be defined for root attribute (TYPE_MEMBER) and not for all leafs (FIELD, CONSTRUCTOR, NESTED_TYPE, METHOD, ANNONYMOUS_EXECUTABLE).

So what about to have roles for all cases?
* TYPE_MEMBER - is general for FIELD, NESTED_TYPE, EXECUTABLE
* EXECUTABLE - is general for CONSTRUCTOR, METHOD, ANNONYMOUS_EXECUTABLE
* FIELD
* CONSTRUCTOR
* NESTED_TYPE
* METHOD
*  ANNONYMOUS_EXECUTABLE

tdurieux:
I just thought something related, for example, we could also use the roles to refine the element that your want to scan.

Example:

```
CtStatement with CtRole.STATEMENT
CtExpression with CtRole.ASSIGNEMENT
...
```

pvojtechovsky:
This issue is solved by #1632. May be just create new issue for the last comment of @tdurieux, before closing this one. I do not understand this request, so I do know if it is already solved or not.

surli:
@tdurieux I close this one, reopen a new issue if necessary for your last comment: I confess I don't understand here either.

PR #1625, opened 2017-10-19
review1: refactor(role): add CtRole.TYPE_MEMBER

pvojtechovsky:
Adds CtRole.TYPE_MEMBER, which will be needed by #1614 and which makes sense for CtPath clients too.

@tdurieux does the sniper mode need to distinguish between CONSTRUCTOR, FIELD, ... Would not it be simpler to fire change listener even using CtRole.TYPE_MEMBER ???

If it would be simpler, then implementation of CtTypeImpl should fire CtRole.TYPE_MEMBER instead of CONSTRUCTOR, FIELD, METHOD, ... roles.


monperrus:
OK to add the role now, the intercession event can be modified later on. One solution is to attach more than one role to an intercession event.

pvojtechovsky:
Another solution would be to express inheritance between roles.
```java
assertSame(CtRole.TYPE_MEMBER, CtRole.FIELD.getSuperRole())
assertNull(CtRole.TYPE_MEMBER.getSuperRole())
```
Then we would fire event with CtRole.FIELD - as now, and client might use getSuperRole() to access more generic role if needed.

May be the opposite would be interesting too `CtRole CtRole.getSubRole(Class)`
CtField field = ...
assertSame(CtRole.FIELD, CtRole.TYPE_MEMBER.getSubRole(field.getClass())

WDYT?

monperrus:
That's also an interesting solution, yes. The one with multiple roles by event seems slightly easier (no need to commit to a predefined hierarchy).

PR #1631, opened 2017-10-20
review2: feature: CtRoleScanner provides CtRole of scanned element

pvojtechovsky:
This implements #1614.

It would be nicer to not have new CtRoleScanner, but to replace CtScanner directly. But that would kill compatibility and would need refactoring of all CtScanner children...

WDYT?

### DIFF
--- a/src/test/java/spoon/processing/CtGenerationTest.java
+++ b/src/test/java/spoon/processing/CtGenerationTest.java
@@ -5,10 +5,12 @@ import org.junit.Test;
 import spoon.Launcher;
 import spoon.generating.CloneVisitorGenerator;
 import spoon.generating.CtBiScannerGenerator;
+import spoon.generating.CtRoleScannerGenerator;
 import spoon.generating.ReplacementVisitorGenerator;
 import spoon.reflect.declaration.CtClass;
 import spoon.reflect.declaration.CtType;
 import spoon.reflect.visitor.CtBiScannerDefault;
+import spoon.reflect.visitor.CtRoleScanner;
 import spoon.reflect.visitor.Filter;
 
 import java.io.File;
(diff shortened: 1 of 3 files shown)


src/test/java/spoon/processing/CtGenerationTest.java
```diff
@@ -53,6 +55,33 @@ public void testGenerateReplacementVisitor() throws Exception {
 	}
 
 	@Test
+	public void testGenerateCtRoleScanner() throws Exception {
+		// contract: generates the biscanner that is used for equality checking
```
monperrus: comment to be updated

monperrus:
Super interesting. 

There is no need to replace CtScanner, we can live with both CtScanner and CtRoleScanner in parallel.

Comments / questions:

* rename to CtRoleBasedScanner?
* what about extracting an interface to clarify the new public methods that makes CtRoleScanner unique?
   * AFAIU, the main method is currently `scan(CtRole, CtElement / List&lt;CtElement&gt;)`;
* Another  core new method in this interface could be "scan(CtElement receiver, CtRole role)" 
* How to test this new scanner?

pvojtechovsky:
> There is no need to replace CtScanner, we can live with both CtScanner and CtRoleScanner in parallel.

The problem is that all infrastructure around CtScanner has to be duplicated. I mean:
* EarlyTerminatingRoleScanner
* CtBiScanner
* ??

monperrus:
It seems there are different ideas here:

* I1 providing a new way for clients to scan the model (eg by overriding `scan(CtElement receiver, CtRole role)`). CtRoleScanner would be generated from CtScanner as CtBiScanner, etc.
* I2 replacing CtScanner as the primary specification of scanning, replacing it by the new annotations, and consequently CtScanner becomes generated.

What are we discussing in this PR?

pvojtechovsky:
This PR is about something like I1.
But I do not want new way of scanning. I just need more information (role of actually scanned element) during scanning process. Old clients can ignore this new information and new clients can profit from that. From that point of view I more like #1632, which does the same like this PR, but with lower future maintenance and less classes to learn for client's.

Concerning I2 ...  may be we need that as first step to minimize maintenance of refactored CtScanner. Because it is simpler to generate CtScanner comparing to manually maintain `role name - method name` pairs without mistakes. Then we need new annotations in spoon model, which defines order of fields in scanning process (let's discuss it in different PR, if you like this idea)

pvojtechovsky:
This alternative is worst then #1632, closing it

PR #1637, opened 2017-10-22
review1: fix(role): CtTypeAccess getTypeAccess uses CtRole.TYPE_ACCESS

pvojtechovsky:
There was conflict with CtTypeAccess#getType(), which returns different value, but had same role CtRole.TYPE


monperrus:
I'd propose to name it ACCESSED_TYPE

pvojtechovsky:
It is ready for merge.

PR #1639, opened 2017-10-22
fix: ClassTypingContext#isSameSignature for generic methods

pvojtechovsky:
The method with this declaration
```java
<U> void visitCtConditional(final CtConditional<U> conditional)
```
defined in class and inteface returned false for ClassTypingContext#isSameSignature(methodFromIface, methodFromClass).

### DIFF
--- a/src/main/java/spoon/support/visitor/ClassTypingContext.java
+++ b/src/main/java/spoon/support/visitor/ClassTypingContext.java
@@ -35,7 +35,6 @@ import spoon.reflect.declaration.CtType;
 import spoon.reflect.declaration.CtTypeInformation;
 import spoon.reflect.declaration.CtTypeParameter;
 import spoon.reflect.declaration.ModifierKind;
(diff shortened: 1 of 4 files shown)


monperrus:
Indeed, there is a bug, thanks.

However, this bug fix is super hard to review because there are too many changes in the test. 

The reason is that the test data (test classes and methods in this case) is changed (and not only added).

It's better to keep the same test classes/methods, and only add new ones to demonstrate the bug. Doing that, there is mostly new content in tests (as opposed to modified content), and at most a few well documented changed assertions (if the incorrect behavior was already specified and enforced).

pvojtechovsky:
You are right. Now I did only minimum changes in existing test - because of change of useless hard to implement API method.
I also added new small test to reproduce problem.

pvojtechovsky:
I am done here

monperrus:
thanks Pavel

PR #1640, opened 2017-10-23
review1: fix: StackOverflowError in MethodTypingContext on <T, U extends T>

pvojtechovsky:
The method like this caused StackOverflowError
```java
<U, K extends U void visitCtConditional(final CtConditional<U> conditional, K k)
```


pvojtechovsky:
It is ready for merge

Issue #1593, opened 2017-10-12
Inconsistency of CtCatchVariable#type and #multiTypes


pvojtechovsky:
The fields CtCatchVariableImpl#type and CtCatchVariableImpl#multiTypes both represents same role CtRole.TYPE - I am OK with that. because both works with same field - type.

But it is actually implemented inconsistently
```java
//set type to x
CtCatchVariableImpl.setType(x)
//set type to y
CtCatchVariableImpl.setMultiType([y])
//ok, the type is y
assertEquals([y], CtCatchVariableImpl.getMultiType())
//what should next line do? This is how it behaves now
assertEquals(x, CtCatchVariableImpl.getType())
//but it should be
assertEquals(y, CtCatchVariableImpl.getType())
//set type to a,b,c
CtCatchVariableImpl.setMultiType([a,b,c])
//ok, the type is a,b,c
assertEquals([a,b,c], CtCatchVariableImpl.getMultiType())
//what should next line do? This is how it behaves now
assertEquals(x, CtCatchVariableImpl.getType())
//but it should probably throw an exception "There are multiple types in catch. Use getMultiType()"
```

WDYT?

I need to solve this problem before #1582 can be implemented.

monperrus:
> //but it should be
> assertEquals(y,CtCatchVariableImpl.getType())
yes, this should be the result

> //what should next line do? This is how it behaves now
> assertEquals(x,CtCatchVariableImpl.getType())
x is is bug.

however, instead of an exception, which makes client code brittle,

I would prefer to return the first element of the list, or even better 
the first common super type of a,b,c

monperrus:
closed per #1668

PR #1641, opened 2017-10-23
to be closed: fix: CtCatchVariable merge type and multiType

pvojtechovsky:
fix of #1593


pvojtechovsky:
It needs a change in CtScanner and classes generated from CtScanner. I would prefer to merge it together with #1642, because there are conflicts.

PR #1644, opened 2017-10-24
doc: Clarify CtType#getAllMethods documentation

surli:


pvojtechovsky:
What about some info about overridden methods?

> It returns only top overridden method. Each method signature is returned only once ...

monperrus:
you're right. modified.

PR #1655, opened 2017-10-26
review chore: improve Travis configuration and speed-up builds

### LINKED ISSUE #1653: chore: improve Travis building time
We currently run  3 different jobs on Travis, always replicating some step like checkstyle: I propose to only do checkstyle, verify, etc once, and to only call mvn test on the 3 different configurations, to speed-up the CI.

WDYT?

surli:
Fix #1653


tdurieux:
I think you can simplify 

- mvn dependency:resolve
- pip install --user CommonMark requests pygithub
- sudo apt-get install xmlstarlet

surli:
What do you mean by "simplify"?

tdurieux:
is useless: `mvn dependency:resolve`

is only required for the verify job
`pip install --user CommonMark requests pygithub`

is util for something but I don't remember for what (website?):
`sudo apt-get install xmlstarlet`

surli:
Thanks, I'll check that and try to improve the script.

surli:
ping @monperrus ok for merging this one? It would speed up the builds

monperrus:
It's a really good start. I would propose to split in two and refine:

Split in two:
- the Travis job design
- the test parallelization (here we may introduce flaky tests because many tests read and write to the same location, we have to make sure first that two tests never pretty-print / compile in the same target folder)

About the Travis job design, instead of a build Matrix, I would propose explicit names in Travis, and hence in the UI of travis (currently the job names are not super clear). The names would be:

* JDK8
* JDK9
* CHECKSTYLE
* SITE
* etc

tdurieux:
> About the Travis job design, instead of a build Matrix, I would propose explicit names in Travis, and hence in the UI of travis (currently the job names are not super clear). The names would be:

I think it is not possible: https://github.com/travis-ci/travis-ci/issues/5898

monperrus:
If you look at the Astor jobs: it's pretty clear: https://travis-ci.org/SpoonLabs/astor

So we can probably workaround if we have only one environment variable.
env:
   SCRIPT=jdk9.sh
   SCRIPT=jdk8.sh
   SCRIPT=checkstyle.sh

and the scripts would call Travis "jdkswitcher" themselves.

--Martin

tdurieux:
See also: https://docs.travis-ci.com/user/speeding-up-the-build/

surli:
@monperrus for me it's ready to merge

monperrus:
super maintainable! thanks!

PR #1659, opened 2017-10-27
review: feat: add missing create method in Factory

### LINKED ISSUE #1272: *Factory.create() should be added to Factory.create$1
We should put all create() method of different factories to Factory with a name like `createMethod` if `create` comes from MethodFactory. 

The SpoonArchitectureEnforcerRule test should be updated (see line 78).

surli:
Fix #1272


monperrus:
thanks! will be super useful for client code.

PR #1660, opened 2017-10-27
feature: TypeFactory LIST, SET, MAP, ... constant type references

pvojtechovsky:
I often needs CtTypeReferences of List, Set, Map, ... may be they might be available as constansts?

WDYT?


pvojtechovsky:
@surli CI must be ill... this small PR cannot produce such REVAPI problems...

PR #1661, opened 2017-10-27
review1: fix: CtTypeParameterReference#getBoundingType() Object -> null

pvojtechovsky:
If CtTypeParameterReference#getBoundingType() returns null or Object, then it has same meaning - there is no bound = java.lang.Object is a bound.

The problem is that some type references returns Object and some returns null. I do not have test case yet, but it happened to me that 2 CtTypeReferences with exactly same printed representation were not equal because one of them had bounding type = null and second had Object.

For future I would prefer to fix it opposite way: to always return Object reference in such case (never null), but this solution needs much more changes in Spoon code, because it expects on many places null value.

So this PR is fast fix to be able to finish other PRs related to CtScanner, etc.

WDYT?

### DIFF
--- a/src/main/java/spoon/support/reflect/reference/CtTypeParameterReferenceImpl.java
+++ b/src/main/java/spoon/support/reflect/reference/CtTypeParameterReferenceImpl.java
@@ -155,6 +155,16 @@ public class CtTypeParameterReferenceImpl extends CtTypeReferenceImpl<Object> im
 
 	@Override
 	public CtTypeReference<?> getBoundingType() {
+		if (superType != null) {
+			/*
+			 * Spoon expects that superType is null on many places.
+			 * But sometime  there is Object in it, which has same meaning like null in this case
+			 * But EqualsVisitor returns false that it is not equal
+			 */
+			if (Object.class.getName().equals(superType.getQualifiedName())) {
+				return null;
+			}
+		}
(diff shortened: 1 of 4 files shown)


pvojtechovsky:
One test fails...
Does this "? super java.lang.Object" has same meaning like "?"

WDYT?

monperrus:
>  Does this "? super java.lang.Object" has same meaning like "?"

Interesting! I'd say yes.

monperrus:
I agree with you, we should at least be consistent. 
I'm OK with this PR.

monperrus:
but we need a well documented test case.

pvojtechovsky:
I am done here. It is true that he test might be simplified. Only the last three lines reproduces the problems. The others are checking that parameter types from methods from different implementations are equal. It was so before, but I think it is not tested.

Can test stay like this?

monperrus:
> For future I would prefer to fix it opposite way: to always return Object reference in such case (never null), but this solution needs much more changes in Spoon code

This would be much better indeed. Have you tried it already? How many tests fail if one does so?

pvojtechovsky:
I tried that. There is a lot of code which calls getBoundingType() and expects null and fails if there is Object instead. Let's do first this simple fix and then we can refactor it deeper.

I need this PR to pass all others

PR #1667, opened 2017-10-29
review: alternative to #1589

monperrus:
completely different way of fixing #1589

WDYT?


pvojtechovsky:
It seems to print classes well, even if parent == null after cloning. I am OK with that PR

@surli do you see any problem?

surli:
> @surli do you see any problem?

Nop, the fix is fine for me 👍

monperrus:
then we can merge

PR #1668, opened 2017-10-29
review1: fix: CtCatchVariable merge type and multiType

pvojtechovsky:
fix of #1593


pvojtechovsky:
It is ready for merge from my point of view.

There are already tests of CtCatchVariable#getType(), where it expects it returns common supertype, because it behaved like that before too - JDT compiler provided that common super type. The test is:
TryCatchTest#testExceptionJava7

monperrus:
super! 

in #1593 you wrote some nice assertions to highlight the bug, would it make sense to put them in a test here?

pvojtechovsky:
Test was added. Let's merge it, so we can continue with next PRs ;-)

Issue #1658, opened 2017-10-27
feat: Create  CtElement#getMyRoleInParent()


surli:
> CtRole CtElement#getMyRoleInParent(), which would return the role of this element in parent element and return null in case of cloned element, because such element is not referred by any parent role.

PR #1669, opened 2017-10-29
WiP: feature: CtElement#getMyRoleInParent()

pvojtechovsky:
implements #1658


monperrus:
Looks super promising. I'd propose to call the new method `getRoleForParent`

pvojtechovsky:
What about `getRoleInParent` ?

monperrus:
getRoleInParent is OK

pvojtechovsky:
@monperrus ? 
[ERROR] Failures: 
[ERROR]   CtScannerTest.testScan:251 expected:<3968> but was:<3972>

monperrus:
Strange. This commit was verified in PR and in master (https://github.com/INRIA/spoon/commits/master). 
Anyway, this assertion is meant to catch differences in scanning behavior. How does this PR change the scanning behavior?

pvojtechovsky:
The scanner behaves well, but this PR contains fix of CtVariableAccessImpl#getVariable(), which causes that spoon model is bigger (no information is lost, because of bug) and therefore scanner founds more nodes.

I will try to make a test case for that bug

monperrus:
The implementation is so elegant now!

PR #1671, opened 2017-10-30
review: feat: Add Environment#setFactory

surli:
Add a setFactory method to environment and set it when calling the main factory constructor with a given environment. 
This allows to use the `getFactory()` of Environment without NPE.

### DIFF
--- a/src/main/java/spoon/compiler/Environment.java
+++ b/src/main/java/spoon/compiler/Environment.java
@@ -295,6 +295,11 @@ public interface Environment {
 	Factory getFactory();
 
 	/**
+	 * Sets the factory of the environment
+	 */
+	void setFactory(Factory factory);
+
+	/**
 	 * Gets the level of loggers asked by the user.
 	 */
 	Level getLevel();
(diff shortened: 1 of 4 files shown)


tdurieux:
Why the environment has a factory?

surli:
> Why the environment has a factory?

Because of the method `getFactory` I assume.
Anyway, I need it to access factory from `DefaultJavaPrettyPrinter` environment.

tdurieux:
My point is the `getFactory` in Environment makes no sense.

Currently, there is no usage of `env.getFactory` In `DefaultJavaPrettyPrinter`.

surli:
> My point is the getFactory in Environment makes no sense.

In that kind of situation: https://github.com/INRIA/spoon/pull/1672/files#diff-5f33bf5d1f38230de815e4c60a67a215 

I consider it cleaner to call `this.env.getFactory()` than `wildcardReference.getFactory()`.

tdurieux:
I do not agree, the Environment is only used for configuration and logging.
They should not be sate in the Environment.

The double reference factory -> Environment; Environment -> Factory is really awful.

I looked where Environment.getFactory is used and it is only in the StandardEnvironment... 
and getFactory can only return null. 
I propose to remove getFactory from the Environment.

monperrus:
I would agree with @tdurieux . The responsibility of Environment is to handle the configuration not the state.

A better solution would be to remove `getFactory` from Environment.

surli:
> The cycle reference Factory -> Environment; Environment -> Factory is really awful.

Agreed! 
Just wanted to use it as it was already there, but I'm ok to remove it :)

Issue #1202, opened 2017-03-03
[generic type] JDT-related bug concerning type inference with generics and wildcard


surli:
See PR: #1200

monperrus:
Three solutions:

- report the bug to JDT and wait for a fix
- report the bug to JDT and fix it ourselves
- deactivate some generics checks from Spoon with special configuration of JDT's `Compiler`.

msteinbeck:
@surli Did you already reported the bug to JDT developers?

surli:
@msteinbeck not yet. I'll do it next week probably.

monperrus:
See https://bugs.eclipse.org/bugs/show_bug.cgi?id=469404

surli:
Apparently a new release of JDT is available: https://mvnrepository.com/artifact/org.eclipse.jdt/org.eclipse.jdt.core we should see if it can help solving this one.

monperrus:
If no action can be taken, I propose to close this one in order to keep our bug dashboard clean.

surli:
> I propose to close this one in order to keep our bug dashboard clean

OK to me, I just put the label to be able to find it back easily.

PR #1672, opened 2017-10-30
review refactor: getBoundingType now returns Object by default

### LINKED ISSUE #1666: CtTypeParameterReference#getBoundingType() should return Object appropriately
CtTypeParameterReference#getBoundingType() should always return Object reference when bounding type is null or Object

spin-off of #1661

### LINKED ISSUE #1670: bug: Regression with the use of generics
Apparently we got a regression with generics maybe because of #1661 
We got the following error when using Spoon on assertj: 

> Failed to execute goal fr.inria.gforge.spoon:spoon-maven-plugin:2.4:generate (default) on project assertj-core: Exception during the spoonify of the target project. Type mismatch: cannot convert from AbstractListAssert<capture#76-of ?,List<? extends Object>,Object,ObjectAssert<Object>> to AbstractListAssert<?,List<?>,Object,ObjectAssert<Object>> at /builds/workspace/AssertJ/src/main/java/org/assertj/core/api/AbstractIterableAssert.java:491 -> [Help 1]

Full log available there: https://ci.inria.fr/sos/job/AssertJ/428/console

surli:
Instead of returning null, default behaviour of getBoundingType is to return Object. 
This PR is an alternative to #1661

Fix #1670
Fix #1666

### DIFF
--- a/src/main/java/spoon/reflect/reference/CtTypeParameterReference.java
+++ b/src/main/java/spoon/reflect/reference/CtTypeParameterReference.java
@@ -104,5 +104,5 @@ public interface CtTypeParameterReference extends CtTypeReference<Object> {
 	 * Returns true if the given typeReference is the default bounding type (i.e. java.lang.Object)
 	 */
 	@DerivedProperty
-	boolean isDefaultBoundingType(CtTypeReference typeReference);
+	boolean isDefaultBoundingType();
 }
--- a/src/main/java/spoon/reflect/reference/CtTypeParameterReference.java
+++ b/src/main/java/spoon/reflect/reference/CtTypeParameterReference.java
@@ -101,7 +101,7 @@ public interface CtTypeParameterReference extends CtTypeReference<Object> {
 	<T extends CtActualTypeContainer> T setActualTypeArguments(List<? extends CtTypeReference<?>> actualTypeArguments);
 
 	/**
-	 * Returns true if the given typeReference is the default bounding type (i.e. java.lang.Object)
+	 * Returns true if this has the default bounding type that is java.lang.Object (which basically means that there is no bound)
 	 */
 	@DerivedProperty
 	boolean isDefaultBoundingType();
--- a/src/main/java/spoon/reflect/reference/CtTypeParameterReference.java
+++ b/src/main/java/spoon/reflect/reference/CtTypeParameterReference.java
@@ -99,4 +99,10 @@ public interface CtTypeParameterReference extends CtTypeReference<Object> {
 	@Override
 	@UnsettableProperty
 	<T extends CtActualTypeContainer> T setActualTypeArguments(List<? extends CtTypeReference<?>> actualTypeArguments);
+
+	/**
+	 * Returns true if the given typeReference is the default bounding type (i.e. java.lang.Object)
+	 */
+	@DerivedProperty
+	boolean isDefaultBoundingType(CtTypeReference typeReference);
 }
--- a/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
+++ b/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
@@ -618,7 +618,7 @@ public class DefaultJavaPrettyPrinter implements CtVisitor, PrettyPrinter {
 		} else {
 			printer.writeIdentifier(ref.getSimpleName());
 		}
-		if (ref.getBoundingType() != null && !ref.getBoundingType().equals(this.env.getFactory().Type().OBJECT)) { // we ignore printing "extends Object"
+		if (ref.getBoundingType() != null && !ref.getBoundingType().equals(ref.getFactory().Type().OBJECT)) { // we ignore printing "extends Object"
 			if (ref.isUpper()) {
 				printer.writeSpace().writeKeyword("extends").writeSpace();
 			} else {
@@ -1570,7 +1570,7 @@ public class DefaultJavaPrettyPrinter implements CtVisitor, PrettyPrinter {
 		}
 		elementPrinterHelper.writeAnnotations(wildcardReference);
 		printer.writeSeparator("?");
-		if (wildcardReference.getBoundingType() != null && !wildcardReference.getBoundingType().equals(this.env.getFactory().Type().OBJECT)) { // we ignore printing "extends Object"
+		if (wildcardReference.getBoundingType() != null && !wildcardReference.getBoundingType().equals(wildcardReference.getFactory().Type().OBJECT)) { // we ignore printing "extends Object"
 			if (wildcardReference.isUpper()) {
 				printer.writeSpace().writeKeyword("extends").writeSpace();
 			} else {
--- a/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
+++ b/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
@@ -618,7 +618,7 @@ public class DefaultJavaPrettyPrinter implements CtVisitor, PrettyPrinter {
 		} else {
 			printer.writeIdentifier(ref.getSimpleName());
 		}
-		if (ref.getBoundingType() != null && !ref.getBoundingType().equals(ref.getFactory().Type().OBJECT)) { // we ignore printing "extends Object"
+		if (!ref.isDefaultBoundingType(ref.getBoundingType()) || !ref.getBoundingType().isImplicit()) {
 			if (ref.isUpper()) {
 				printer.writeSpace().writeKeyword("extends").writeSpace();
 			} else {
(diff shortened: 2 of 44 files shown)


src/main/java/spoon/support/reflect/reference/CtTypeParameterReferenceImpl.java
```diff
@@ -50,6 +50,7 @@
 
 	public CtTypeParameterReferenceImpl() {
 		super();
+		this.setBoundingType(getFactory().Type().objectType()); // by default set bounding type to object
```
pvojtechovsky: we should create new CtTypeReference, because AST should not share nodes

src/main/java/spoon/support/reflect/reference/CtTypeParameterReferenceImpl.java
```diff
@@ -173,6 +164,13 @@ public boolean removeBound(CtTypeReference<?> bound) {
 		if (superType != null) {
 			superType.setParent(this);
 		}
+
+		// ugly but else make testSetterInNodes failed
+		if (superType == null) { // if null, set bounding type to object
+			superType = getFactory().Type().objectType();
+			superType.setParent(this);
```
pvojtechovsky: use new CtTypeReference too

pvojtechovsky:
This PR is compatible with algorithms, which needed fast fix #1661. This solution is on good way. I like it!

monperrus:
I like it too! `return null` is always such a bad idea!

src/main/java/spoon/support/reflect/reference/CtTypeParameterReferenceImpl.java
surli: Actually the methods already create a clone of OBJECT in `TypeFactory`, see: https://github.com/INRIA/spoon/blob/e2e359e2e59c23a94cd042a6851e8058cd62dc5c/src/main/java/spoon/reflect/factory/TypeFactory.java#L250

So I assume the nodes are not shared.

pvojtechovsky: you are right. I did not knew that. So it is OK

surli:
**Any help is welcome for that PR!**

The changes are done for passing the current Spoon test suite, but I don't manage to fix #1670 or to reproduce it in a small unit test.
The bug occured when using the script https://github.com/INRIA/spoon/blob/master/doc/jenkins/build.sh on assertj-core project. I now got the following error:

>  Failed to execute goal fr.inria.gforge.spoon:spoon-maven-plugin:2.4:generate (default) on project assertj-core: Exception during the spoonify of the target project.: Type mismatch: cannot convert from AbstractListAssert<capture#76-of ?,List<? extends Object>,Object,ObjectAssert<Object>> to AbstractListAssert<?,List<?>,Object,ObjectAssert<Object>> at /Users/urli/Github/assertj-core/src/main/java/org/assertj/core/api/AbstractIterableAssert.java:491

I manage to fix the bug manually by changing `AbstractListAssert<?,List<?>,Object,ObjectAssert<Object>>` to `AbstractListAssert<?,List<? extends java.lang.Object>,Object,ObjectAssert<Object>>` manually in the code, but I don't completely understand why explicitely extending the wildcard to object in the source code fix the bug. 
I think it's because this method use another method which have the following return signature:
`AbstractListAssert<?,List<? extends E>,Object,ObjectAssert<E>>` and JDT does not manage to make the binding properly, but I still did not manage to reproduce the bug yet.

Still, my manual bug fix seems to show that sometimes we need to explicitely output an `extends Object`. Then I may need to know when there was an explicit extends object in source, and when it was implicit. Any idea to fix that?

src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
```diff
@@ -618,7 +618,7 @@ public void visitCtTypeParameter(CtTypeParameter typeParameter) {
 		} else {
 			printer.writeIdentifier(ref.getSimpleName());
 		}
-		if (ref.getBoundingType() != null) {
+		if (ref.getBoundingType() != null && !ref.getBoundingType().equals(ref.getFactory().Type().OBJECT)) { // we ignore printing "extends Object"
```
monperrus: one solution may be to simply remove this and always write "extends Object"

pvojtechovsky: I vote for shorter printed source code. So never print "... extends Object"

surli: > So never print "... extends Object"

And then I have a bug with assertj-core.

surli:
So... in fact as here as in #1202 the bug is related to JDT own interpretation of wildcard generics in Java. 
If you take this line https://github.com/joel-costigliola/assertj-core/blob/9d45e93be7f41da5c0709fc852bebd4ef79f6162/src/main/java/org/assertj/core/api/AbstractIterableAssert.java#L1435

in Intellij Idea it shows a warning on  `? extends Object` saying its redundant, and in Eclipse it shows an error when you remove the extends. 
Then as I already said, here the only way to fix the bug is to have a way to know if this extends belongs to the original source code or not.

monperrus:
OracleJDK and OpenJDK don't handle generics the same way, and this is yet another instance of the
problem.

> Then as I already said, here the only way to fix the bug is to have a way to know if this extends
> belongs to the original source code or not.
We have isImplicit that we can put in the bounding type.

--Martin

surli:
> We have isImplicit that we can put in the bounding type.
I'll take a look on that.

surli:
The last changes now fix the error in #1670 we use the `isImplicit` to tell DJPP if it must be output or not.

src/main/java/spoon/reflect/reference/CtTypeParameterReference.java
```diff
@@ -99,4 +99,10 @@
 	@Override
 	@UnsettableProperty
 	<T extends CtActualTypeContainer> T setActualTypeArguments(List<? extends CtTypeReference<?>> actualTypeArguments);
+
+	/**
+	 * Returns true if the given typeReference is the default bounding type (i.e. java.lang.Object)
+	 */
+	@DerivedProperty
+	boolean isDefaultBoundingType(CtTypeReference typeReference);
```
monperrus: parameter needed? could be simply `isDefaultBound()` with no parameter?

src/test/java/spoon/test/reference/TypeReferenceTest.java
```diff
@@ -510,15 +510,17 @@ public void testClearBoundsForTypeParameterReference() throws Exception {
...
 		reference.addBound(factory.Type().createReference(String.class));
 
 		assertNotNull(reference.getBoundingType());
 
 		reference.setBounds(new ArrayList<>());
 
-		assertNull(reference.getBoundingType());
+		assertEquals(factory.Type().OBJECT, reference.getBoundingType());
+		assertTrue(reference.isDefaultBoundingType(reference.getBoundingType()));
 	}
```
monperrus: add a test corresponding to the assert4j problem?

src/main/java/spoon/reflect/reference/CtTypeParameterReference.java
surli: The parameter is needed for case like https://github.com/INRIA/spoon/pull/1672/files#diff-e14ef777963d781b479e6fdd3e8d419dR79 where we want to check before setting the boundingType.

src/test/java/spoon/test/reference/TypeReferenceTest.java
surli: > add a test corresponding to the assert4j problem?

I cannot reproduce exactly the assert4j problem, but we already have in our test a case with an explicit "? super Object", I changed the assert accordingly: https://github.com/INRIA/spoon/pull/1672/files#diff-c20710bb8ee877d0eae2d22578a8f010R506

monperrus:
I'm not convinced by the design and the encapsulation of this new method. Can the the test be rewritten?

surli:
> I'm not convinced by the design and the encapsulation of this new method. Can the the test be rewritten?

You're talking about `CtTypeParameterReference#isDefaultBoundingType`, right?
I'm completely open to redesign it, but I'm not sure to understand what you want.

monperrus:
I'd propose to remove the parameter of the method. the "is" applied to "this".

src/main/java/spoon/reflect/reference/CtTypeParameterReference.java
pvojtechovsky: The code on the link will work if object is sent to adapt method too. Only check for null is needed.
The implementation of `is default bounding type` can be made static. May be we do not need such method at all.

surli:
> I'd propose to remove the parameter of the method. the "is" applies to "this" not to the parameter.

I'm ok with that but I also want to have a place where the default value of boundingType is defined and that I can use without creating a new CtTypeParameterReference object. This will allow me to do some checks with a single point of definition for the default value of bounding type. 

Where do you think I can create such a method?

monperrus:
In a helper class with only static methods? such as JDTTreeBuilderHelper?

surli:
It's ready for merging

monperrus:
Thanks a lot.

PR #1677, opened 2017-10-31
refactor(modifier): createCtModifierHandler that handles the modifiers

tdurieux:
Goal: remove duplicate code


monperrus:
looks good to me, this will probably be very useful for #1649

tdurieux:
I would like to do the same with bodyholder, there are a lot of duplicate code to add/remove/ statements

monperrus:
@surli if it's OK for you, don't hesitate to merge, so that we can proceed with #1649?

PR #1679, opened 2017-10-31
test: CtComment with \r separator

surli:


monperrus:
thanks

pvojtechovsky:
I afraid that this PR destroyed existing test testJavaDocCommentOnUnix which expects that JavaDocComment file contains unix EOL. I suggest to make new  JavaDocCommentMac.java file with MAC EOL.

surli:
> I afraid that this PR destroyed existing test testJavaDocCommentOnUnix which expects that JavaDocComment file contains unix EOL

Actually I kept your own file and copy it in resources after changing the new line character. So I don't think so :)

pvojtechovsky:
The ok, thanks for explanation ;-)

PR #1680, opened 2017-10-31
chore: Fork test to run in parallel

surli:


pom.xml
```diff
@@ -274,6 +274,8 @@
           https://maven.apache.org/surefire/maven-surefire-plugin/examples/class-loading.html -->
           <useSystemClassLoader>true</useSystemClassLoader>
           <useManifestOnlyJar>false</useManifestOnlyJar>
+          <forkCount>4</forkCount>
```
monperrus: do we have to hard code a value here?

surli: > do we have to hard code a value here?

We have to specify a value but according to documentation, it can be related to the number of processor on the machine: 
> if you terminate the value with a 'C', that value will be multiplied with the number of available CPU cores in your system. For example forkCount=2.5C on a Quad-Core system will result in forking up to ten concurrent JVM processes that execute tests.

tdurieux:
Apparently it takes more time with this parameter

surli:
> Apparently it takes more time with this parameter

On Travis yes, it was faster on my machine. Maybe we can improve it by playing with the number of forks?

tdurieux:
I think there is only one thread on travis. I'm not sure we can improve it.

surli:
> I think there is only one thread on travis. I'm not sure we can improve it.

I propose yo use `forkCount: 1C` then: it will use the architecture of the machine. I'll change it to test.

surli:
Ok even with 1C it takes the same speed, and apparently it takes more time on my machine... I close this one for now then.

PR #1681, opened 2017-10-31
feat: MavenLauncher Warning when a provided dependency is not loaded 

### LINKED ISSUE #1621: feat: SpoonMavenLauncher display warnings when presence of provided dependencies in pom.xml
When a pom.xml contains provided dependencies, they still might be mandatory to build the model: it would be useful to the user to see that they were missing dependencies (because provided) on the form of a console warning.

surli:
Fix #1621

### DIFF
--- a/src/main/java/spoon/MavenLauncher.java
+++ b/src/main/java/spoon/MavenLauncher.java
@@ -282,7 +282,7 @@ public class MavenLauncher extends Launcher {
 				}
 				// ignore not transitive dependencies
 				if (isLib && ("test".equals(dependency.getScope()) || "provided".equals(dependency.getScope()))) {
-					LOGGER.log(Level.WARN, "Dependency ignored (scope: provided or test): "+dependency.toString());
+					LOGGER.log(Level.WARN, "Dependency ignored (scope: provided or test): " + dependency.toString());
 					continue;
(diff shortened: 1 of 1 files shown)


src/main/java/spoon/MavenLauncher.java
```diff
@@ -281,10 +282,10 @@ private String extractVariable(String value) {
 				}
 				// ignore not transitive dependencies
 				if (isLib && ("test".equals(dependency.getScope()) || "provided".equals(dependency.getScope()))) {
+					LOGGER.log(Level.WARN, "Dependency ignored (scope: provided or test): " + dependency.toString());
 					continue;
 				}
 				String fileName = dependency.getArtifactId() + "-" + version;
-				// TODO: Check the scope of the dependency (local dependency is not handled)
```
tdurieux: this TODO is still not handled

tdurieux:
This  TODO is still not handled "Check the scope of the dependency (local dependency is not handled)"

surli:
> this TODO is still not handled "Check the scope of the dependency

Isn't it the job of this condition? Else what do you mean by handle them?

tdurieux:
Now I was lazy I did not handle System dependencies

PR #1683, opened 2017-10-31
chore: Revapi now deletes its own comments before commenting

### LINKED ISSUE #1426: chore: spoon-bot should  delete old Revapi reports
spoon-bot should post a Revapi comment only if the computed API change are different from the last post.

Possible implementation, in `revapy.py`, use the Github API to go over the list of comments to the PR and compare them against the newly computed one.

surli:
Fix #1426

### DIFF
--- a/chore/revapy.py
+++ b/chore/revapy.py
@@ -65,6 +65,9 @@ if (action in accepted_actions):
         repo = gh.get_repo(repo_name,True)
 
         pr = repo.get_pull(pr_id)
+        for comments in pr.get_comments():
+            if comments.user == gh.get_user():
+                comments.delete()
         pr.create_issue_comment(file_content)
     except GithubException as e:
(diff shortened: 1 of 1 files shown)


monperrus:
What about something slightly more powerful?

```python
#spoon-bot posts only if the computed API changes are different from the last post.
last_post = None
for comment in pr.get_comments():
  if comment.user == gh.get_user():
                last_post = comment
if comment.text != file_content:
  comment.delete()
  pr.create_issue_comment(file_content)
```

surli:
I prefer the idea which consist in always deleting older comments from revapi, as it will also avoid duplicate message with a different order in changes and it ensures that we always have the very last report message from revapi.

monperrus:
ok, let's try this and see how it feels. Thanks.

PR #1686, opened 2017-11-01
review: feature: introduce the concept of "Pattern"

pvojtechovsky:
This PR introduces a new concept `Pattern`, which is an improvement of TemplateMatcher, it allows to match EVERY attribute of Spoon model - not only a subset like current TemplateMatcher.

The concept of Pattern is like this:
* `Pattern` consists of List of `RootNode`s. Main implementation of `RootNode` is `ConstantNode`, which handles constant values  of attributes and `ElementNode`, which knows the node type and a Map<CtRole, RootNode> of nodes for handling of specific attributes. So `ElementNode`s and `ConstantNode`s together makes up the tree of the pattern - it mirrors to be generated/matched Spoon AST.
* The variables in Pattern are represented by `ParameterNode` which links `ParameterInfo`, which defines which value from Map of parameters (`ParameterValueProvider`) has to substituted.
* Structure of Pattern is usually created by copying/mirroring of an example template Spoon AST, but after it is initialized the Pattern processing is based on implementations of RootNode interface only. It doesn't need origin Spoon AST - so the matching and generating algorithms are much simpler and more flexible then before. The pattern can be created dynamically by code. No need to declare it in source as Template class. But there is of course a way how to create Pattern from legacy Template class too because in some cases it is possible. More complex patters can be first created from Template (like before) and then adapted by extra PatternBuilder calls to adjust things which were not possible with legacy Templates.
* The Patterns are created by PatternBuilder, which accept any spoon AST. Even the legacy Template can be used to initialize Pattern. It is subject of future work to make a PatternBuilder intuitive and usable for clients. 

Both algorithms generating of code from template and matching code by template are primary behavior of new patterns and are 100% compatible. Everything what can be generated can be matched and opposite.


monperrus:
Cool! Could you give an example snippet of how to use this?

pvojtechovsky:
Here are some examples
A) 
https://github.com/INRIA/spoon/pull/1686/files#diff-dfa8a4b46f7408d58548e5797a9239e1R30
B) 
https://github.com/INRIA/spoon/pull/1686/files#diff-b5d51a0eb70403715f1134a17f6fce51R31
C)
https://github.com/INRIA/spoon/pull/1686/files#diff-25f39a55233b37bb5f826e3ad922186cR23
https://github.com/INRIA/spoon/pull/1686/files#diff-70de42aee1116615f166a1b492efa1b0R19

src/test/java/spoon/test/prettyprinter/templates/OldPattern_methods.java
```diff
@@ -0,0 +1,82 @@
...
+		try (ListPrinter lp = printer.createListPrinter(start(), next(), end())) {
+			for (Item value : entity.$getItems$()) {
+				lp.printSeparatorIfAppropriate();
+				statements();
+			}
+		}
+	}
+	
+	static Pattern create(Factory factory) {
+		return PatternBuilder
```
monperrus: could you explain as comments what it's meant to do?

pvojtechovsky:
Here is updated and commented example, which does this: it looks for all occurrences of code like
```java
try (ListPrinter lp = elementPrinterHelper.createListPrinter(false, "(", false, false, ",", true, false, ")")) {
	for (Entry<String, CtExpression> e : annotation.getValues().entrySet()) {
		lp.printSeparatorIfAppropriate();
		if ((annotation.getValues().size() == 1 && "value".equals(e.getKey())) == false) {
			//it is not a default value attribute. We must print a attribute name too.
			printer.writeIdentifier(e.getKey()).writeSpace().writeOperator("=").writeSpace();
		}
		elementPrinterHelper.writeAnnotationElement(annotation.getFactory(), e.getValue());
	}
}
```
and replaces it by this code:
```java
elementPrinterHelper.printList(annotation.getValues().entrySet(), false, "(", false, false, ",", true, false, ")"), e -> {
	if ((annotation.getValues().size() == 1 && "value".equals(e.getKey())) == false) {
		//it is not a default value attribute. We must print a attribute name too.
		printer.writeIdentifier(e.getKey()).writeSpace().writeOperator("=").writeSpace();
	}
	elementPrinterHelper.writeAnnotationElement(annotation.getFactory(), e.getValue());
});
```
The example consists of these files:
* [OldPattern_ParamsInNestedType.java](https://github.com/pvojtechovsky/spoon/blob/feaPattern/src/test/java/spoon/test/prettyprinter/templates/OldPattern_ParamsInNestedType.java) - defines Pattern of old code
* [NewPattern.java](https://github.com/pvojtechovsky/spoon/blob/feaPattern/src/test/java/spoon/test/prettyprinter/templates/NewPattern.java) - defines Patter of new code
* [NewPattern#replaceOldByNew(CtElement)](https://github.com/pvojtechovsky/spoon/blob/feaPattern/src/test/java/spoon/test/prettyprinter/templates/NewPattern.java#L47) - the method which does the replacement
```java
	/**
	 * Looks for all matches of Pattern defined by `OldPattern_ParamsInNestedType.class`
	 * and replaces each match with code generated by Pattern defined by `NewPattern.class`
	 * @param rootElement the root element of AST whose children has to be transformed
	 */
	public static void replaceOldByNew(CtElement rootElement) {
		Factory f = rootElement.getFactory();
		Pattern<OldPattern_ParamsInNestedType.Parameters> newPattern = NewPattern.createPattern(f);
		Pattern<OldPattern_ParamsInNestedType.Parameters> oldPattern = OldPattern_ParamsInNestedType.createPattern(f);
		oldPattern.forEachMatch(rootElement, (match, params) -> {
			match.replace(newPattern.substitute(params));
		});
	}
```
It does not work yet, this PR is under construction, but I have feeling like it is on good way ;-) ... but it will be NOT a baby step PR.

pvojtechovsky:
Please help me with detection of the reason why these tests fails on CI. 
They pass in my environment without any error and I haven't found any details about that `ArrayIndexOutOfBounds` in the CI logs :-(

[ERROR]   APITest.testSetterInNodes:404 » ArrayIndexOutOfBounds 0
[ERROR]   TemplateArrayAccessTest.testArrayAccess:27 » ArrayIndexOutOfBounds 0
[ERROR]   TemplateArrayAccessTest.testArrayLengthAccess:41 » ArrayIndexOutOfBounds 0
[ERROR]   TemplateClassAccessTest.testClassAccessTest:26 » ArrayIndexOutOfBounds 0
[ERROR]   TemplateEnumAccessTest.testEnumAccessTest:30 » ArrayIndexOutOfBounds 0
[ERROR]   TemplateTest.createTypeFromTemplate:818 » ArrayIndexOutOfBounds 0
[ERROR]   TemplateTest.substituteStringLiteral:870 » ArrayIndexOutOfBounds 0
[ERROR]   TemplateTest.testStatementTemplateRootSubstitution:798 » ArrayIndexOutOfBounds
[ERROR]   TemplateTest.testSubstituteInnerClass:773 » ArrayIndexOutOfBounds 0
[ERROR]   TemplateTest.testTemplateArrayAccess:748 » ArrayIndexOutOfBounds 0
[ERROR]   TemplateTest.testTemplateInvocationSubstitution:695 » ArrayIndexOutOfBounds 0
[ERROR]   TemplateTest.testTemplateMatcherMatchTwoSnippets:663 » ArrayIndexOutOfBounds 0
[ERROR]   TemplateTest.testTemplateMatcherWithWholePackage:620 » ArrayIndexOutOfBounds 0

@surli Any idea?

Note: I have found only logs similar to this:
```
[ERROR] substituteStringLiteral(spoon.test.template.TemplateTest)  Time elapsed: 0.101 s  <<< ERROR!
java.lang.ArrayIndexOutOfBoundsException: 0
	at spoon.test.template.TemplateTest.substituteStringLiteral(TemplateTest.java:870)
```
but it is only part of the stacktrace, which has not enough details to detect cause of problem...

pvojtechovsky:
pattern
simplevalueresolver
subpatternvalueresolver
parametervalueprovider
conversioncontext

pvojtechovsky:
@monperrus , @surli I am nearly finished with that PR and would like to start merging process. But this is little bit bigger PR...

So first questions: 
* Do you know what this PR is about ?
  It is a rewrite of existing template matching and generating algorithm.

* Do you want that (after some cleaning) in Spoon ?



The steps to merge might be:
1) to takeover the core matching/generating algorithms, which are new - they will not influence existing Spoon code.
2) to develop/improve pattern builders, so they are understandable for clients and to merge them too
3) to merge builders which brings compatibility with legacy templates
4) to replace legacy templates by new code

So I would split it to at least 4 smaller PRs (like 4 steps above)

WDYT? Any suggestions?

pvojtechovsky:
I have just updated all my comments in this PR above. Please read them again to see current state.
There are also examples of how to create a pattern in a new way. Mainly that PatternBuilder is a subject for future improvements - to make it as much user friendly as possible. Note that PatternBuilder is just a wrapper, which creates the internal Pattern structure. Pattern structure is already finished and fine tuned and works well.

pvojtechovsky:
And here is a very simple example:
Here is the template model
```java
public class LoggerModel {
	private String _classname_;
	private String _methodName_;
	private CtBlock<?> _block_;

	public void block() throws Throwable {
		try {
			Logger.enter(_classname_, _methodName_);
			_block_.S();
		} finally {
			Logger.exit(_methodName_);
		}
	}
}
```
and here is the code which creates Pattern from type member `block` of class `LoggerModel`
```java
		spoon.pattern.Pattern pattern = PatternBuilder.create(launcher.getFactory(), LoggerModel.class, model -> model.setTypeMember("block"))
				.configureAutomaticParameters()
				.build();
```
and use that pattern to generate method
```java		
//prepare parameters
		final CtClass<?> aTargetType = launcher.getFactory().Class().get(Logger.class);
		final CtMethod<?> toBeLoggedMethod = aTargetType.getMethodsByName("enter").get(0);
		Map<String, Object> params = new HashMap<>();
		params.put("_classname_", factory.Code().createLiteral(aTargetType.getSimpleName()));
		params.put("_methodName_", factory.Code().createLiteral(toBeLoggedMethod.getSimpleName()));
		params.put("_block_", toBeLoggedMethod.getBody());
//generate a method using a before made pattern and a parameters
		final List<CtMethod> aMethods = pattern.applyToType(aTargetType, CtMethod.class, params);

```

monperrus:
interesting! Beyond the snippet, we can discuss the high-level functional test which also has assertions. We start by which one?

pvojtechovsky:
>  we can discuss the high-level functional test

* The modified TemplateTest#testTemplateMatcher shows how matcher provides matched parameter values
* CodeReplaceTest#testMatchSample1 is a test around the examples from descriptions above
* TemplateMatcherTest contains many other tests

pvojtechovsky:
This PR replaces old generating/matching engine by the new one. So all the existing template tests are checking the new Pattern algorithm too.

monperrus:
I start looking at this beast, starting with the tests.

Now looking at testTemplateMatchOfMultipleElements.

Two first questions:

- are you OK that I push directly to your branch with my test comments and refactoring?
- what is the meaning of the selectors in `PatternBuilder#create`

pvojtechovsky:
> are you OK that I push directly to your branch with my test comments and refactoring?

yes. I will stop rebasing on this branch, so your work is not lost

> what is the meaning of the selectors in PatternBuilder#create

All PatternBuilder#create gets 
1) a CtType as input (in form of Class, CtType or CtTypeReference)
2) selector which defines which parts of AST of the type has to be taken as the Pattern/Template model.

> what is the difference/added value between a Pattern and a Template?

From some point of view they are synonyms. Template is legacy implementaiton. Pattern is new implementation.
The main problem of Template is that definition of template is coupled with AST of the template model. The things which cannot be expressed using template model are simply impossible. And the matching/generating code is compicated because it always directly work with origin template AST, which is designed for different purposes...

On the other side Pattern is based on `RootNode`s, which has nothing to do with Spoon AST - they are independent from spoon AST and they are designed to be ideal for generating/matching of Spoon AST nodes. Therefore new generating/matching code is very simple and is not polluted by all the dirty tricks how to define template parameter, etc.

So Pattern building means usually these steps:
1) take an AST of an CtType
2) select a part of (or whole) AST of CtType, which has to be used as a `template AST model` for matching/generating
3) create `tree of RootNodes`, which mirrors `template AST model`
4) define substitution parameters on `tree of RootNodes`
5) define live statements on `tree of RootNodes`
Wrap `tree of RootNodes` by a `Pattern` object and return it. Now you have Pattern which can be used for generating and matching.

> what is the difference/added value between a Pattern and a TemplateMatcher?

Pattern can do a template matching too. It is able to deliver matched parameters and it is able to match list of elements (not only single element)

> could you extract an interface for Pattern?

Why? Do you try to understand how it works or you think it needs an interface? If you try to understand, then I will rather extract a smaller PR with core engine of Pattern (without Pattern object itself and without builder - which are only a nice facade around that core). Then we can discuss that smaller PR as first step and then as a next step we can discuss how Pattern and PatternBuilder works.

WDYT?

monperrus:
>     could you extract an interface for Pattern?
>
> Why?
>
To extract the essence of the design, both for the code reviewer to understand (me), and as
documentation for the future users.

> Template is legacy implementaiton. Pattern is new implementation.
Template if an interface. Would it make sens that Pattern is a subclass/subinterface of Template?

monperrus:
pushed first changes, added questions directly in testTemplateMatchOfMultipleElements

pvojtechovsky:
> Template if an interface. Would it make sens that Pattern is a subclass/subinterface of Template?

From conceptual point of view yes. But technically the legacy Template is used little bit different then Pattern.
`Template#apply(CtType<?> targetType)` does not fit to `Pattern#applyToType(CtType<?> targetType, Class<T> valueType,  Map<String, Object> params)`

... and see the pushed changes with answers to your questions

monperrus:
>> Template if an interface. Would it make sens that Pattern is a subclass/subinterface of Template?

>From conceptual point of view yes. 

OK. `Pattern` is a good name, but we don't see the conceptual relation with `Template` and `TemplateMatcher` in the name. Once I'll fully understand what it does, we may change its name.

Thanks a lot for your answers (pushed back some minor changes).

Next question: the meaning of `forEachMatch` is clear, but I don't understand why, for a given match, one can have several elements, as returned by `getMatchingElements`.

In what case does `getMatchingElements` return multiple elements for a single match?

pvojtechovsky:
> In what case does getMatchingElements return multiple elements for a single match?

for example when you want to match all sequences of these two statements:
```java
LOG.debug("It failed", e);
throw new RuntimeException("It failed", e);
```
The code above is an antipattern, which one might want to search for. It is not wrapped by block. It may occur next to other statements.

Such Pattern returns 2 elements as a match.

src/main/java/spoon/pattern/PatternBuilder.java
```diff
@@ -120,6 +120,10 @@ public static PatternBuilder create(CtTypeReference<?> templateTypeRef, List<CtE
 		return new PatternBuilder(templateTypeRef, patternModel);
 	}
 
+	public static PatternBuilder create(CtElement patternModel) {
```
pvojtechovsky: I expect that most of the patterns will need a reference to a declaring type. So I suggest that all PatternBuilder methods have that reference. It forces clients to think whether they have such a reference and provide it (and it MUST be provided, to assure that matching/generating works well) or to provide null if they are sure that in their special/rare case the pattern is independent on the type.

Note: that there is no way how to add that reference later. It must be added at the beginning of Pattern building process.

But I can live with that new method if you really like it ;-)

monperrus:
done with testTemplateMatchOfMultipleElements, pushed some minor changes

I continue with testExtensionDecoupledSubstitutionVisitor

Could you explain this snippet:

``` 
		final List<CtMethod> aMethods = PatternBuilder.create(
                launcher.getFactory(), LoggerModel.class, 
                        model -> model.setTypeMember("block")
                )
				.configureAutomaticParameters()
				.build().applyToType(aTargetType, CtMethod.class, params);
```

pvojtechovsky:
> Could you explain this snippet:

I added this comment to code too
```java
//create a patter from the LoggerModel#block
spoon.pattern.Pattern pattern = PatternBuilder.create(launcher.getFactory(), LoggerModel.class,
		//type member named "block" of LoggerModel is used as pattern model of this pattern
		model -> model.setTypeMember("block"))
		//all the variable references which are declared out of type member "block" are automatically considered
		//as pattern parameters
		.configureAutomaticParameters()
		.build();
```

surli:
@pvojtechovsky could you also review the documentation of Template pushed on the website (https://github.com/INRIA/spoon/blob/master/doc/template_definition.md) with your new changes? 

As the PR is pretty big and it concerns a new feature, would you consider to put it in the `experimental` package? It will be more clear for developers that it might still be a bit buggy and the API can still move. WDYT?

pvojtechovsky:
> review the documentation of Template 

I agree that it is a step which must be done. But I have feeling like it would be wasting of time now, because 
1) This PR is 100% backward compatible - at least from old test case point of view. So the existing Template documentation is still valid
2) The PatternBuilder which would be mainly described in documentation is subject for change before it gets merged - I would like to discuss with you the method names and the ways how to create patterns in more easy way. So I expect that such documentation would be rewritten soon anyway.

>  PR is pretty big 

I agree. So we should found a special way how to take over it. There are some ideas how to split this into smaller parts in my comments above.

> to put it in the experimental package

yes, I am ok with that. But it does not help to takeover that stuff. And I think it is not good idea to move it into experimental and merge it without deeper discussion and understanding how it works.

surli:
> This PR is 100% backward compatible - at least from old test case point of view. So the existing Template documentation is still valid

Yes but as far as I understood, using a Pattern would make templating easier because you have a clear separation of concerns between the "template part" and the "matching part", am I right?

If so, it could be an interesting exercise to take over the example given in Template documentation and do them with the Pattern mechanism to show the before/after. Or maybe to show an example of what we __cannot__ do with Template and we can do with Pattern. WDYT?

pvojtechovsky:
> because you have a clear separation of concerns between the "template part" and the "matching part", am I right?

There is clear separation of concerns between 
C1) way how template is defined (e.g. using 1) legacy Template, 2) new PatternBuilder, 3) ... any other future way how to build a template)
C2) processing of the template model (tree of `RootNode`s). Where processing can mean A) matching template on existing code, B) generating new code from template.

> Maybe a good way to discuss the API is to use some documentation snippet

ok, good idea. I will write some documentation for that purpose

pvojtechovsky:
I am finished with update of Template documentation. I hope it is understandable ;-)

Thanks @surli for that idea. It looks like a good step to explain this PR.

doc/template_definition.md
```diff
@@ -4,20 +4,63 @@ tags: [template]
...
+interface, type, ... any Spoon model subtree),
+where parts of that code may be **template parameters**.
 
-A Spoon template is a regular Java class that taken as input by the Spoon templating engine to perform a transformation.
+**Spoon template** can be used in two ways:
+
+A) **to generate new code**. The generated code is a copy of code
+of **Spoon template**, where each **template parameter** is substituted
+by it's value. We call this operation **Generating**.
+```java
```
surli: I really think it would be easier for a reader to see directly an example of really small couple template/code and the obtained result, before displaying any Spoon code to achieve this result.

doc/template_definition.md
```diff
@@ -4,20 +4,63 @@ tags: [template]
...
 contexts and give different results, depending on its parameters.
 
 ![Overview of Spoon's Templating System]({{ "/images/template-overview.svg" | prepend: site.baseurl }})
 
-Definition of templates
------------------------
+B) **to search for a code**. The found code is same like code of **Spoon template**,
+where code on position of **template parameter** may be arbitrary and is copied
+as value of **template parameter**. We call this operation **Matching**.
+```java
```
surli: My previous comment is particulary true here: I did not really understood what you meant by: 

> where code on position of **template parameter** may be arbitrary and is copied
as value of **template parameter**.

it might be easier with an example code/template -> obtained matching

doc/template_definition.md
```diff
@@ -4,20 +4,63 @@ tags: [template]
...
+Factory spoonFactory = ...
+//build a Spoon template
+Pattern spoonTemplate = ... build a spoon template. For example for an method ...
+//search for all occurences of the method like spoonTemplate in whole model
+spoonTemplate.forEachMatch(spoonFactory.getRootPackage(), (Match match) -> {
+	//this Consumer is called once for each method which matches with spoonTemplate
+	Map<String, Object> parameters = match.getParametersAsMap();
+	CtMethod<?> matchingMethod = match.getMatchingElement(CtMethod.class);
+	String aNameOfMatchedMethod = parameters.get("methodName");
```
surli: Is it right that `aNameOfMatchedMethod == matchingMethod.getSimpleName()` here?

doc/template_definition.md
```diff
@@ -4,20 +4,63 @@ tags: [template]
...
+	...
+});
+```
+There are several ways how to build a **spoon template**
+A) Using a regular java class, which implements a `Template` interface
```
surli: Typo: it's not displayed as a list here in markdown

doc/template_definition.md
```diff
@@ -81,8 +124,79 @@ method.getBody().insertBegin(injectedCode);
 
 ```
 
-Kinds of templating
--------------------
+Definition of templates using `PatternBuilder`
```
surli: Patterns are more powerful than Template, so why don't you present them first?

doc/template_definition.md
```diff
@@ -81,8 +124,79 @@ method.getBody().insertBegin(injectedCode);
...
+}
+```
+
+The code, which creates a Spoon template using `PatternBuilder` looks like this:
+```java
+Pattern t = PatternBuilder.create(factory,
+	//defines template class.
+	CheckBoundTemplate.class, 
+	//defines which part of template class will be used as template model
+	model -> model.setBodyOfMethod("statement"))
```
surli: Could it be a specific statement of the method?

doc/template_definition.md
```diff
@@ -81,8 +124,79 @@ method.getBody().insertBegin(injectedCode);
...
+	//tells builder that all variables defined out of scope of template model (body of the method)
+	//are considered as template parameters
+	.configureAutomaticParameters()
+	//builds an instance of Pattern
+	.build();
+```
+
+
+This template specifies a
+statements (all statmenets of body of method `statement`) that is a precondition to check that a list
```
surli: typo on statements

doc/template_definition.md
```diff
@@ -81,8 +124,79 @@ method.getBody().insertBegin(injectedCode);
...
+template-generated code is consistent with the template scope.
+
+```java
+Pattern t = PatternBuilder.create(...building a template. See code above...);
+// creating a holder of parameters
+Map<String, Object> parameters = new HashMap<>();
+parameters.put("_col_", createVariableAccess(method.getParameters().get(0))); 
+
+// getting the final AST
+CtStatement injectedCode = t.substituteSingle(factory, CtStatement.class, parameters);
```
surli: Why do you need to specify the `CtElement` type in `substituteSingle`: you know that `t` refers to a `CtStatement` so you only can substitute here to another `CtStatement`, no?

In other word: what would happen if I type: 

`CtMethod injectedCode = t.substituteSingle(factory, CtMethod.class, parameters);`

doc/template_definition.md
```diff
@@ -262,9 +376,106 @@ String someMethod() {
 }
 ```
 
+## PatternBuilder parameters
```
surli: Same comment as above: why not pushing that before Template parameters?

doc/template_definition.md
```diff
@@ -262,9 +376,106 @@ String someMethod() {
...
+```java
+Pattern t = PatternBuilder.create(...select template model...)
+	.configureParameters(pb -> 
+		pb.parameter("firstParamName")
+			//...select which AST nodes are parameters...
+			//e.g. using parameter selector
+			.byType(Request.class)
+			//...modify behavior of parameters...
+			//e.g. using parameter modifier
+			.setMinOccurence(0);
```
surli: I did not understood this example: is there a template example attached?

doc/template_definition.md
```diff
@@ -290,3 +501,23 @@ is transformed into:
     java.lang.System.out.println(1);
 }
 ```
+## Inlining with PatternBuilder
+The template code in spoon templates made by `PatternBuilder` is never inlined automatically.
+But you can mark code to be inlined this way:
+```java
+Pattern t = PatternBuilder.create(...select template model...)
+	//...configure parameters...
+	configureLiveStatements(ls ->
```
surli: so `configureLiveStatements` is used only for `inlining`? Shouldn't it be renamed then?

doc/template_definition.md
pvojtechovsky: The documentation I made now is primarily targeted to explain this PR to You Simon and Martin.

I agree with you, but I suggest to postpone refectoring of this documentation until architecture of this PR is understood by you and after this PR is near to merge.

doc/template_definition.md
pvojtechovsky: I speak about template matching, that matched code may have any AST subtree on the position of template parameter - because template parameter is a kind of "variable"

doc/template_definition.md
pvojtechovsky: It depends on the definition of the template (which is not really exactly defined here - to keep example simple), but yes, it was the idea it is equal to simple name of the method

doc/template_definition.md
pvojtechovsky: I suggest to refactor documentation later. First your understanding, then refactoring of code and then refactoring of documentation ... and note, I am mainly developer. I am not so good in writing good documentation - in communication.

doc/template_definition.md
pvojtechovsky: Actually not, but if you see this use case interesting, then it is no problem to add new method into that builder.

doc/template_definition.md
pvojtechovsky: It will fail because CtStatement cannot be converted to CtMethod. But if you would use
```java
CtBlock injectedCode = t.substituteSingle(factory, CtBlock.class, parameters);
```
then you would get a CtBlock which contains your generated statements. Usually the definition of required type is useless, but I already seen cases when the definition is absolutely necessary - it is case of legacy templates too. So it is part of that API now.

doc/template_definition.md
pvojtechovsky: yes, it is for inlining. This PR really needs a lot of renaming and I am open to any better names.

doc/template_definition.md
pvojtechovsky: It is not really an example which would do something real ... it is just an example which shows where to use `template parameter selectors` and where to use  `template parameter modifiers`

doc/template_definition.md
surli: I just wanted to know what kind of filter could be done here, but I get an idea looking on `TemplateModelBuilder` :)

doc/template_definition.md
surli: ok

doc/template_definition.md
surli: ok so here, you will create a map of parameters, where `firstParamName` is the value of the first optional variable of type `Request`? or should `firstParamName` be the name of the variable?

doc/template_definition.md
surli: why not `configureInlineStatements` then?

doc/template_definition.md
surli: Ok

doc/template_definition.md
surli: Sorry but I still don't understand what you exactly mean here: what do you exactly intend by "on the position of the template parameter"? 

Let's keep an example: my template is the following: 

```java
public int mymethod(int _myvar_) {
   return _anotherVar_;
}
```

and I got this code:

```java
public int foo(int x) {
  return x*2;
}

public int bar(int y) { 
  return foo(y);
}
```

Then the template match against the 2 methods, and the param `anotherVar` will point for both CtExpression `x*2` and `foo(y)` right? or as `y` is also pointed by `myvar` it won't be that simple?

src/main/java/spoon/pattern/PatternBuilder.java
```diff
@@ -0,0 +1,935 @@
...
+				});
+		});
+		return this;
+	}
+
+	/**
+	 * Builder which allows to define which part of AST of template {@link CtType}
+	 * has to be used as a model of the {@link Pattern}
+	 */
+	public static class TemplateModelBuilder {
```
surli: Why do you keep this class internal? `PatternBuilder` is already pretty big, it seems easier to maintain/get access to the doc if the class is in its own CompilationUnit.

doc/template_definition.md
pvojtechovsky: It matches twice. 
First match points to method "foo" and have parameters: 
```
_myvar_ = int x
_anotherVar_ = x*2
```
Second match points to method "bar" and have parameters: 
```
_myvar_ = int y
_anotherVar_ = foo(y)
```

doc/template_definition.md
pvojtechovsky: "firstParamName" is the name of the template parameter and all the CtTypeReferences which points to `Request.class` will be substituted by that parameter.

src/main/java/spoon/pattern/PatternBuilder.java
pvojtechovsky: good idea, I will move it to own file

doc/template_definition.md
pvojtechovsky: ok, I will rename all "live" to "inline"

doc/template_definition.md
surli: ok that's what I expected then.

doc/template_definition.md
surli: hmmm, so here the `setMinOccurrence` is only used to skip some values if needed?

For example, if the following method is matched:

```java

Request foo(Request bar) {
   Request result = ... 
   return result;
}
```

if I used setMinOccurence(0) it takes all `Request`: the CtReturn, `bar` and `result`; if I used `setMinOccurence(1)` it takes only `bar` and `result` etc?

pvojtechovsky: `setMinOccurence` combined with `byType` is bad example. I will search for a better one ... because this combination makes no sense together.
May be search for setMinOccurence in test cases of this PR. There are some running examples.

doc/template_definition.md
pvojtechovsky: Thanks, should be fixed with next commit.

doc/template_definition.md
pvojtechovsky: Thanks, should be fixed with next commit.

doc/template_definition.md
pvojtechovsky:
```java
//a template model 
void method(String _x_) {
	zeroOneOrMoreStatements();
	System.out.println(_x_);
}

//a pattern definition
Pattern t = PatternBuilder.create(...select template model...)
	.configureParameters(pb -> 
		pb.parameter("firstParamName")
			//...select which AST nodes are parameters...
			//e.g. using parameter selector
			.bySimpleName("zeroOneOrMoreStatements")
			//...modify behavior of parameters...
			//e.g. using parameter modifier
			.setMinOccurence(0);
			
		//... you can define as many parameters as you need...
		
		pb.parameter("lastParamName").byVariable("_x_");
	)
	.build();
```
next commit will contain this example. Here the the invocation of `zeroOneOrMoreStatements();` represents parameter "firstParamName", whose value will match with 0, 1 or more statements.

monperrus:
working on the API, doc and tests (done TemplateTest and testMatchSample1).

I start to understand and appreciate the full power of Patterns.
(+pushed my changes of today)

question: `CodeReplaceTest#testMatchSample1` is nice! but is there any replacement happening in this test?

will continue tomorrow.

pvojtechovsky:
> question: CodeReplaceTest#testMatchSample1 is nice! but is there any replacement happening in this test?

there is no replacement. See CodeReplaceTest#testTemplateReplace. It does replacement ... but there are missing assertions ... I just checked it manually. It must be checked automatically somehow

monperrus:
Hi Pavel,

> It is little bit more clear what it does
This is the goal. I assume that if some Spoon code is hard to understand for me, it will be
impossible for end-users.

That said, Iknew that you would react :-). Here, the problem is to have it in the constructor. If we
have it in a method, it would be much more explicit, something like

Pattern pattern = PatternBuilder.create(ctClass).selectASTNodes(new TemplateModelBuilder(ctClass).setTypeMember("matcher1"))

WDYT?

pvojtechovsky:
Hi Martin,

> I assume that if some Spoon code is hard to understand for me, it will be impossible for end-users.

I agree. I am aware the meta programming is harder then programming and templates are another extra meta level ... so it is even harder to understand. So I highly welcome any ideas how to make it more understandable ;-)

**So I can live with that change** if, it moves this PR near to the merge. There will be 100 of such topics, so if I would fight too hard, it would be never merged.

Anyway here are some arguments, why it was directly in create:

The pattern building consists on these steps. Note that these steps MUST be processed in this exact order.
1) take a CtType and select which part of THAT CtType will be a model. Result of this step is CtTypeReference + List of CtElements.
2) define a template parameters on the List of CtElements
3) mark inline statements

This idea:
> Pattern pattern = PatternBuilder.create(ctClass).selectASTNodes(new TemplateModelBuilder(ctClass).setTypeMember("matcher1"))

brings another place where client can make a mistake during building of pattern, because it allows to try to run step 2) before step 1).

The nicer would be the Pattern builder API which
1) enforces correct order (guides client to use correct order)
2) to offers autocompletion in each place, so client only chooses from available choices 
3) good method names + documentation, which makes it easily understandable.

monperrus:
AFAIU, the essence of a pattern is a list of elements. The type is the context.

What if we have `PatternBuilder.create(CtElement...)` and that we compute the context type on the fly, as the common super type of all elements? In this case, we also enforce the contract, correct?

This would also solve another issue I have: all `create` methods are duplicated with a CtType and a CtTypeReference, which is not elegant.

monperrus:
Hi Pavel, I'm having hard time to understand `configureLocalParameters` and `InlineStatementsBuilder`, could you add a test for them (see `testInlineStatementsBuilder`)? Thanks. --Martin

pvojtechovsky:
> I'm having hard time to understand configureLocalParameters and InlineStatementsBuilder, 

`configureLocalParameters` is needed mainly for internal purposes by inline forEach statement do declare `localParam`
```
for(localParam : globalParam) {...}
```
Note that such `localParam` is not visible accessible in global list of template parameters, because it changes value with each iteration over items of `globalParam`... so it's declaration is purpose of `configureLocalParameters`. We can hide this method so it is accessible only internally.

There is test `TemplateMatcherTest#testMatchForeach`, which uses that `configureLocalParameters` internally

pvojtechovsky:
> // TODO: specify what InlineStatementsBuilder does

See end of https://github.com/pvojtechovsky/spoon/blob/feaPattern/doc/template_definition.md page, is not that description enough?

... InlineStatementsBuilder allows to mark which statements has to be inlined. Actually ant CtIf and CtForEach can be inlined.

pvojtechovsky:
> AFAIU, the essence of a pattern is a list of elements. The type is the context.

yes, it is correct.

> What if we have PatternBuilder.create(CtElement...) and that we compute the context type on the fly, as the common super type of all elements? In this case, we also enforce the contract, correct?

You are probably right. I will think about that.
Note that list of elements might be theoretically enclosed by several levels of nested types ... but this is not supported by current Pattern implementation anyway. But would it make sense?

monperrus:
Thanks for your answers.

For configureLocalParameters can we make it private or package visible then? (there are mnay such untested public methods to look at)

For InlineStatementsBuilder, by specify, I mean a dedicated test with assertions :-)

Your solution `A) PatternBuilder.create(CtElement...) -> returns PatternBuilder` looks good, you can proceed (I have no pending changes on my side).

pvojtechovsky:
This PatternBuilder method
```java
	/** adds the given nodes to the pattern */
	public PatternBuilder selectNodes(TemplateModelBuilder selector) {
		this.patternModel.addAll(selector.getTemplateModels());
		return this;
	}
```
is wrong, because `PatternBuilder#patternModel` and `PatternBuilder#patternNodes` are mirrors, which are mapped to each other using `PatternBuilder#patternElementToSubstRequests`.
Therefore the only place where elements can be initialized is constructor. That is the simplest approach, with less place for client's mistake. So I deleted this method. OK?

pvojtechovsky:
Martin, please run the tests before git push from /spoon-core/src/test/java/spoon/test/template (in eclipse it is 2 clicks and and it needs about 10s) and you immediately see whether your changes broke something. You broke 6 tests. I am trying to found what change it caused now

monperrus:
I ran the tests :-) but not all, sorry!

pvojtechovsky:
> I ran the tests 

The debugging of problem in patterns is quite hard. I am sorry, that I was disappointed, but I afraid that I have to spent next hours with debugging. Fortunately the git history helped me and the fixes were easy ;-)

I am finished with refactoring we spoke above. It is yours again. 

When I have some time, I will add the tests for public methods to fix the failing architectonic test you made.

pvojtechovsky:
> For InlineStatementsBuilder, by specify, I mean a dedicated test with assertions :-)

@monperrus  there are already 8 tests dedicated to InlineStatementsBuilder. What is wrong with them?

monperrus:
Continued my pass:

- centralized doc in `pattern.md` (I'm going through it slowly)
- centralized tests in `PatternTest`.

Two questions:

Q1) so far, I'm only reading code for matching code, not for transforming. What parts of `Pattern` are used for transformation? Where are the corresponding tests?

About inline, in `testMatchForeach`

```java
.configureParameters(pb -> {
        pb.parameter("values").byVariable("values").setContainerKind(ContainerKind.LIST);
})
.configureInlineStatements(lsb -> lsb.byVariableName("values"))
```

This is hard to read and understand, in part because the configuration is spread over the ParameterBuilder object and the PatternBuilder object. 

Q2) Would it be possible to rework the API to only at the level of ParameterBuilder, eg 


```java
.configureParameters(pb -> {
        pb.parameter("values")
             .byVariable("values")
             .setContainerKind(ContainerKind.LIST)
             .matchInlinedStatements()
})
```

doc/pattern.md
```diff
@@ -0,0 +1,439 @@
...
+## PatternBuilder parameters
+The `PatternBuilder` takes all the Template parameters mentioned in the chapters above
+and understands them as pattern parameters, when `PatternBuilder#configureTemplateParameters()`
+is called.
+
+```java
+Pattern t = PatternBuilder.create(...select pattern model...)
+	.configureTemplateParameters()
+	.build();
+```
```
pvojtechovsky: The description from `## PatternBuilder parameters` till here is related to integration of legacy templates with new Pattern. It probably doesn't belong into this document ... or is at lest confusing in current content.

src/main/java/spoon/pattern/PatternBuilder.java
```diff
@@ -396,7 +398,12 @@ PatternBuilder configureLocalParameters(Consumer<ParametersBuilder> parametersBu
 		return this;
 	}
 	/**
-	 * adds all standard Template parameters based on {@link TemplateParameter} and {@link Parameter} annotation
+	 *
+	 * 	All variables defined out of scope of the model (eg body of the method)
+	 * 	are considered as template parameters
```
pvojtechovsky: It is not true for this method. The method `configureTemplateParameters` creates parameters in same way like legacy Templates did it. The comment you have entered here would be valid for `createPatternParameters` method.
So I suggest to rollback this javadoc to origin version.

monperrus:
> WDYT? Should I refactor it this way too?

Actually, I prefer the first version, because the code is more explicit, one clearly see that one creates new parameters.

pvojtechovsky:
> Q2) Would it be possible to rework the API to only configure it at the level of ParameterBuilder, eg
```java
.configureParameters(pb -> {
        pb.parameter("values")
             .byVariable("values")
             .setContainerKind(ContainerKind.LIST)
             .matchInlinedStatements()
})
```
The new method `matchInlinedStatements` was added into ParametersBuilder. I have replaced all the tests where it was possible by this new method.
Note that we cannot remove `configureInlineStatements`, because there are still some cases when it is needed. See the remaining tests, which are still using it.

monperrus:
Thanks Pavel. I'll continue my review from testMatchForeach.

monperrus:
I don't understand the tested feature and intent of #testMatchIfElse, could you add some comments?

I continue with testGenerateMultiValues

monperrus:
Question 3: In testMatchGreedyMultiValueUnlimited, why all statements are matched (see TODO in code).

pvojtechovsky:
> I don't understand the tested feature and intent of #testMatchIfElse, could you add some comments?

It is a pattern model which contains 3 optional parts
Example: This pattern model from the #testMatchIfElse test
```java
		if (option) {
			//matches String argument
			System.out.println("string");
		} else if (option2) {
			//matches int argument
			System.out.println(1);
		} else {
			//matches double argument
			System.out.println(4.5);
		}
```
contains 3 sub models. The idea is that submodels can be any model, which has to be optionally matched as part of bigger model. In this test case the submodels are quite similar and differs only in type of println method parameter: 1) StringLiteral, 2) IntegerLiteral 3) DoubleLiteral
... and whole pattern model matches, when at least one submodel matches.

OK, this simple case might be handled by 3 different models where each would be matched individually (so no inline if would be needed), but I think it makes sense to combine mandatory submodels with optional submodels, which allows to match and transform more complex real life models.
The combination of optional and mandatory model is tested in `spoon.test.template.testclasses.replaceOldPattern`

src/main/java/spoon/pattern/PatternBuilder.java
```diff
@@ -0,0 +1,699 @@
...
+				//for the compatibility reasons with Parameters.getNamesToValues(), use the proxy name as parameter name
+				String parameterName = stringMarker;
+
+				CtTypeReference<?> paramType = paramField.getType();
+
+				if (paramType.isSubtypeOf(f.Type().ITERABLE) || paramType instanceof CtArrayTypeReference<?>) {
+					//parameter is a multivalue
+					// changed from bySimpleName to byReferenceName
+					// everything is OK but for testTemplateInheritance, where I (Martin)
+					// suspects that there is a hidden bug or implicit contract
```
pvojtechovsky: I do not understand, what is the problem

pvojtechovsky:
> In testMatchGreedyMultiValueUnlimited, why all statements are matched?

The pattern model is:
```java
statements();
```
and the pattern definition is:
```java
pb.parameter("statements").byReferenceName("statements").setContainerKind(ContainerKind.LIST);
```
That means the CtStatement `statements()` is a pattern parameter called 'statements'. And it's type is List, so it is a sorted multivalue.
Note: there is no constraint about what kind of statement it should match, so it matches everything ... and because of `ContainerKind.LIST` it matches zero, one or more then one arbitrary statement.

src/test/java/spoon/test/template/PatternTest.java
```diff
@@ -0,0 +1,1281 @@
...
+		// in this example the main if statement starting with "if (option) {"
+		// is inlined
+		// and any of the branch can be matched
+		CtType<?> ctClass = ModelUtils.buildClass(MatchIfElse.class);
+
+		CtType<?> type = ctClass.getFactory().Type().get(MatchIfElse.class);
+		Pattern pattern = PatternBuilder.create(new PatternBuilderHelper(type).setBodyOfMethod("matcher1").getPatternElements())
+				.configureParameters(pb -> {
+					pb.parameter("option").byVariable("option");
+					pb.parameter("option2").byVariable("option2");
```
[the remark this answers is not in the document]
pvojtechovsky:
```java
boolean a = ...
boolean b = ...

if (a) { ... 
} else if (b) { ...
} else { ...
}
```
One parameter is enough for one if. But if you have chain of IFs like above and like in the test case, then you need one parameter for each if expression.
Or do you speak about something else?

src/test/java/spoon/test/template/PatternTest.java
```diff
@@ -0,0 +1,1281 @@
...
+					"dd++"), listToListOfStrings(match.getMatchingElements()));
+
+			// correctly matching the outer parameter
+			assertEquals("dd", match.getParameters().getValue("varName"));
+		}
+	}
+
+	@Test
+	public void testMatchIfElse() throws Exception {
+		//contract: if statements can be inlined
```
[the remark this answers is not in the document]
pvojtechovsky: yes, this contract is correct.

src/test/java/spoon/test/template/testclasses/match/MatchMultiple.java
```diff
@@ -0,0 +1,55 @@
...
+
+	/** return a pattern built from {}@link {@link #matcher1()} */
+	public static Pattern createPattern(Quantifier matchingStrategy, Integer minCount, Integer maxCount) throws Exception {
+		CtType<?> type = ModelUtils.buildClass(MatchMultiple.class);
+		return PatternBuilder.create(new PatternBuilderHelper(type).setBodyOfMethod("matcher1").getPatternElements())
+			.configureParameters(pb -> {
+
+				// matching anything that is called "statements".
+				// the setContainerKind(ContainerKind.LIST) means match zero, one or more then one arbitrary statement
+				pb.parameter("statements").byReferenceName("statements").setContainerKind(ContainerKind.LIST);
```
monperrus: I see, what I don't understand is that the reference is matched, which means the executable ref in this case.
while the pattern is actually the CtInvocation, ie the parent of the executable reference.

When does the link to parent is made? Is it automated?

src/test/java/spoon/test/template/PatternTest.java
```diff
@@ -0,0 +1,1281 @@
...
+
+		// in this example the main if statement starting with "if (option) {"
+		// is inlined
+		// and any of the branch can be matched
+		CtType<?> ctClass = ModelUtils.buildClass(MatchIfElse.class);
+
+		CtType<?> type = ctClass.getFactory().Type().get(MatchIfElse.class);
+		Pattern pattern = PatternBuilder.create(new PatternBuilderHelper(type).setBodyOfMethod("matcher1").getPatternElements())
+				.configureParameters(pb -> {
+					pb.parameter("option").byVariable("option");
```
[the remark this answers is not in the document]
pvojtechovsky: I agree, it is tricky. The inlineing of statements is advanced feature built on top of normal parameters, therefore it is more complex.
All these steps must be done internally
1) client has to define, which expressions will be parameters (code: 'byVariable("option")')
2) client has to define name of each parameter (code: 'parameter("option")')
3) and after all parameters are defined client can mark that chain of IFs as inline statement (code: `configureInlineStatements(lsb -> lsb.byVariableName("option"))`

I am sure that there is possible to define some nicer builder method, which will do all these steps together. I would prefer to do it in next PR, to keep this one ""simple"" ;-)

monperrus:
* testMatchIfElse: made progress, see comments
* testGenerateMultiValues: made great progress, almost done, see question about whether substituteList does matching also
* testMatchGreedyMultiValueUnlimited: made great progress, see question about reference and invocation.

src/test/java/spoon/test/template/PatternTest.java
```diff
@@ -0,0 +1,1296 @@
...
+			assertEquals("3.14", match.getParameters().getValue("value").toString());
+		}
+	}
+	@Test
+	public void testGenerateMultiValues() throws Exception {
+		// contract: the pattern parameter (in this case 'statements')
+		//can have type List and can be replaced by list of elements
+		//(in this case by list of statements)
+
+		// here, in particular, we test method "substituteList"
```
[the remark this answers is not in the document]
pvojtechovsky: use substituteList when you expect 0, 1 or more generated root elements
use substituteSingle when you expect exactly one generated root element

Note: internally both cases are handled same, here is the code from `spoon.pattern.Generator'
```java
	/**
	 * Generates one target depending on kind of this {@link RootNode}, expected `expectedType` and input `parameters`
	 * @param node to be generated node
	 * @param parameters {@link ParameterValueProvider}
	 * @param expectedType defines {@link Class} of returned value
	 *
	 * @return a generate value or null
	 */
	default <T> T generateTarget(RootNode node, ParameterValueProvider parameters, Class<T> expectedType) {
		ResultHolder.Single<T> result = new ResultHolder.Single<>(expectedType);
		generateTargets(node, result, parameters);
		return result.getResult();
	}

	/**
	 * Generates zero, one or more targets depending on kind of this {@link RootNode}, expected `expectedType` and input `parameters`
	 * @param node to be generated node
	 * @param parameters {@link ParameterValueProvider}
	 * @param expectedType defines {@link Class} of returned value
	 *
	 * @return a {@link List} of generated targets
	 */
	default <T> List<T> generateTargets(RootNode node, ParameterValueProvider parameters, Class<T> expectedType) {
		ResultHolder.Multiple<T> result = new ResultHolder.Multiple<>(expectedType);
		generateTargets(node, result, parameters);
		return result.getResult();
	}
```

src/test/java/spoon/test/template/testclasses/match/MatchMultiple.java
pvojtechovsky: Yes, it is automated by this code:
/spoon-core/src/main/java/spoon/pattern/ParametersBuilder.java (line 629)
```java
	public ParameterElementPair getSubstitutedNodeOfElement(ParameterInfo parameter, CtElement element) {
		ParameterElementPair parameterElementPair = new ParameterElementPair(parameter, element);
		parameterElementPair = transformVariableAccessToVariableReference(parameterElementPair);
		parameterElementPair = transformArrayAccess(parameterElementPair);
		parameterElementPair = transformTemplateParameterInvocationOfS(parameterElementPair);
		parameterElementPair = transformExecutableRefToInvocation(parameterElementPair);
		parameterElementPair = transformCtReturnIfNeeded(parameterElementPair);
		//if spoon creates an implicit parent (e.g. CtBlock) around the pattern parameter, then replace that implicit parent
		parameterElementPair = getLastImplicitParent(parameterElementPair);
		return parameterElementPair;
	}
```

pvojtechovsky:
>> ParameterValueProvider

> Martin wrote: is it specific to Pattern? could this be moved to spoon.support.util?

It is actually used only by Pattern, but you are right that it is independent feature. 
Yes, I can move it and I can make a special PR which contains only this feature.

@monperrus Let me know when you are finished with your changes and I will then do a move of package and rebase to master.

pvojtechovsky:
I have just noticed that `testTemplateInheritance` fails after your refactoring of `PatternBuilder line 473`
```java
//parameter is a multivalue
// changed from bySimpleName to byReferenceName
// everything is OK but for testTemplateInheritance, where I (Martin)
// suspects that there is a hidden bug or implicit contract
pb.parameter(parameterName).setContainerKind(ContainerKind.LIST).byReferenceName(stringMarker);
```
What was wrong? There is template like this:
```java
public class SubTemplate extends SuperTemplate {
	public void methodWithTemplatedParameters(Object params) {
           ...
        }
}
```
and there is expected that parameter 'params' is substituted. But the `params` here is `CtParameter`, which is `CtNamedElement` and not `CtReference`, so it is not enough to call `ParameterBuilder#byReferenceName`. It needs a call of `ParameterBuilder#byNamedElement` too.

Note: legacy Template engine replaced both NamedElements and References, so new Pattern has to do it too... and I think it is correct like this.

src/test/java/spoon/test/template/PatternTest.java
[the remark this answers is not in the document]
monperrus: thanks. that helps a lot. further reworked the test according to your answer. 

... I'm done with testMatchIfElse!

src/test/java/spoon/test/template/PatternTest.java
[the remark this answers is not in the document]
monperrus: that's clear, see latest test changes

src/test/java/spoon/test/template/PatternTest.java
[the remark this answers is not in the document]
monperrus: Here? I have a concern with the API: we have 4 public methods `substitute*`. It is hard for client to quickly grasp the differences. Is it possible to have a single `substituteMethod?`

src/test/java/spoon/test/template/PatternTest.java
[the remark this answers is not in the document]
pvojtechovsky: I am OK with these changes in test. Thanks

Note: I am just not sure whether `if` with more branches is tested somewhere else. So may be origin test might be kept too next to test modified by You.

monperrus:
* testMatchIfElse: DONE :-)
* testGenerateMultiValues: 
   * can we simplifiy the API with only one method substitute?
* testMatchGreedyMultiValueUnlimited: .
   * question: byReferenceName points to an executable reference, while the element that is matched is the ctInvocation, which is the parent of the executable reference. where is the automated pointer to the parent made? in which cases?

> so it is not enough to call `ParameterBuilder#byReferenceName`. It needs a call of `ParameterBuilder#byNamedElement` too.

You're right. Could you add it?

> I will then do a move of package and rebase to master.

Super. I'm done for today.

src/test/java/spoon/test/template/PatternTest.java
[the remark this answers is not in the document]
pvojtechovsky: We can easily get rid of the methods with `Map` parameters. Then `substitueList` is more generic so it can survive.

But I would prefer to keep all 4. Client can use the one which fits best to its needs. I have experience with APIs of other SW and I always approaciate when I as client can choose the best fitting method.

pvojtechovsky:
> question: byReferenceName points to an executable reference, while the element that is matched is the ctInvocation, which is the parent of the executable reference. where is the automated pointer to the parent made? in which cases?

Yes, it is automated by this code:
/spoon-core/src/main/java/spoon/pattern/ParametersBuilder.java (line 629)
```java
	public ParameterElementPair getSubstitutedNodeOfElement(ParameterInfo parameter, CtElement element) {
		ParameterElementPair parameterElementPair = new ParameterElementPair(parameter, element);
		parameterElementPair = transformVariableAccessToVariableReference(parameterElementPair);
		parameterElementPair = transformArrayAccess(parameterElementPair);
		parameterElementPair = transformTemplateParameterInvocationOfS(parameterElementPair);
		parameterElementPair = transformExecutableRefToInvocation(parameterElementPair);
		parameterElementPair = transformCtReturnIfNeeded(parameterElementPair);
		//if spoon creates an implicit parent (e.g. CtBlock) around the pattern parameter, then replace that implicit parent
		parameterElementPair = getLastImplicitParent(parameterElementPair);
		return parameterElementPair;
	}
```

pvojtechovsky:
> can we simplifiy the API with only one method substitute?

We can easily get rid of the methods with Map parameters. 
Then substituteList is more generic so it can survive.

But I would prefer to keep all 4. Client can use the one which fits best to its needs. I have experience with APIs of other SW and I always appreciate when I as client can choose the best fitting method. Note: c class `Pattern` is kind of facade around low level engine, so it is OK that it offers several variants of method which delivers similar functionality.

>> so it is not enough to call ParameterBuilder#byReferenceName. It needs a call of ParameterBuilder#byNamedElement too.

> You're right. Could you add it?

I already added it before you wrote that comment.

pvojtechovsky:
I moved ParameterValueProvider to spoon.support.util and did rebase. 
Could you please avoid merging of our branches in future and to use rebase when possible? I expect that we have to split this PR into smaller parts in future and after some merging it might be complicated.

See my answers to your questions in above comments too. I am done here. It is yours again ;-)

pvojtechovsky:
> ParameterValueProvider ... did so method renaming there, please check

I like all these new names. Thanks ;-)

> I have some doubts here. It is super implicit for client code, hence hard to understand. It seems better to me that client code explicitly says it wants: eg it looks for an invocation as pattern parameter. WDYT?

Arguments for that implicit code:
* The legacy Template was super implicit here too. 
* most of the clients doesn't understand the Spoon model deep enough to distinguish between CtInvocation, CtExecutableReference, ... so the implicit fallback to the most probable case makes sense ... **to the elements which are directly visible in source code**
* the Pattern builder is simpler for clients

Arguments agains that implicit code:
* it kills flexibility and avoids possibility to define some rare Patterns

Note: I am aware of that there are many TODOs in that Pattern building area - I mean how to define what exactly has to be matched?
I see big challenge in matching of method invocations. There are many cases:
* I want to match everything what has same name like template method
* I want to match everything what has same number of parameters
* I want to match everything what has same name, arbitrary parameters but same return type
* ... many many combinations :-(

But let's solve that later. The legacy Templates are NOT handling these cases too and they are already quite usable ;-)

monperrus:
Continued the review. Questions:

* testMatchPossesiveMultiValueMaxCount4: why does it match something while testMatchPossesiveMultiValueUnlimited there is no match?
* testMatchPossesiveMultiValueMinCount2: what additional behavior is tested compared to testMatchPossesiveMultiValueMinCount?
* testMatchInSet: why don't we have the pattern parameters for the last matched methods sample4?
* testTemplateReplace: what is the contract?

pvojtechovsky:
> testMatchPossesiveMultiValueMaxCount4: why does it match something while in testMatchPossesiveMultiValueUnlimited there is no match?

because the template is
```java
	public void matcher1() {
		statements(); //this statement possesively eats everything until maxCount
		System.out.println("something");
	}
```
So testMatchPossesiveMultiValueMaxCount4 matches 4 elements by `statements()` and 5th element by `System.out.println("something");`
And testMatchPossesiveMultiValueUnlimited  matches all elements by `statements()` and there remains no element for mandatory single match of `System.out.println("something");`, which has nothing to match then.


> testMatchPossesiveMultiValueMinCount2: what additional behavior is tested compared to testMatchPossesiveMultiValueMinCount?

It is similar but it uses different template. 
A) mandatory single element at the end
```
System.out.println("something");
```
B) 0,1 or more elements at the end
```java
for (String v : something) {
System.out.println(v);
}
```
and therefore it tests behavior of different part of matching code.

> testMatchInSet: why don't we have the pattern parameters for the last matched methods sample4?

because it is empty. The template is
```java
public void matcher1() throws IOException, FileNotFoundException {
	}
```
so it matches with 
```java
private void sample4() throws IOException, FileNotFoundException
```
because there are exactly the expected exceptions. But there are no other exceptions, so match.getParameters().getValue("otherThrowables") is empty/null ... it is one of the TODOs, to have empty collection instead of null value, but it is not so easy actually...


> testTemplateReplace: what is the contract?

it is not good unit test. It was a try whether Pattern engine can be really used for a transformation. See PR #1956, which was made by the code in testTemplateReplace. 

So we can delete that "test"

monperrus:
Hi Pavel,

Thanks for your super clear answers, I've update the tests based on them.

DONE:

* testMatchInSet: OK: added documentation
* testTemplateReplace: OK removed
* testGenerateClassWithSelfReferences: OK done
* testGenerateMethodWithSelfReferences: OK done
* testMatchPossesiveMultiValueMinCount: OK. Note that I removed if/then/else with assertions, it's a nightmare to understand and debug

Questions: 

* testMatchPossesiveMultiValueMaxCount4:  why method matcher1 is not matched? (see question in code)
* testMatchPossesiveMultiValueMinCount2: I refactored it for better understanding and now it fails: why?

pvojtechovsky:
> testMatchPossesiveMultiValueMaxCount4: why method matcher1 is not matched?

because this method contains a `for` statement, 
```java
public void matcher1(List<String> something) {
 		statements1.S();
 		statements2.S();
 		for (String v : something) {
 			System.out.println(v);
 		}
 	}
```
which is not part of template. Why is not a part? It was inlined.

> testMatchPossesiveMultiValueMinCount2: I refactored it for better understanding and now it fails: why?

I do not know yet. I will check it next days

pvojtechovsky:
I am sorry I looked at another test ;-). So again:

the template is
```java
public void matcher1() {
		statements(); //here is possessive quantifier with min 0 and max count 4
		System.out.println("something"); //this is mandatory
	}
```
and the question is
> testMatchPossesiveMultiValueMaxCount4: why method matcher1 is not matched?

Because the possessive quantifier eats both statements. Because it never returns back there remains nothing for the second template statement, which cannot match then.

pvojtechovsky:
> testMatchPossesiveMultiValueMinCount2: I refactored it for better understanding and now it fails: why?

because after refactoring there were missing two method calls. So pattern was different then expected. I fixed it by `fix test` commit

Anyway thank You Martin 👍 for Your time spent on making these tests more readable.

monperrus:
Thanks Pavel for your answer.

I'm done with all tests. Also merged with latest master.

The next step is to clearly identify the public API. We cannot only rely on Java modifiers. So we can use the the `internal` convention with a package `spoon.pattern.internal`. All classes (or sub-packages) in that package, even public one, are considered as private.

* OK for you? Could you do this refactoring?
* What would be the classes in the public API? Pattern, PatternBuilder, ParametersBuilder, something else?

monperrus:
> I guess these must be public API: Pattern, PatternBuilder, ParametersBuilder, InlineStatementsBuilder, PatternBuilderHelper, ConflictResolutionMode, Match, Quantifier

Good, + ParameterValueProvider from spoon.utils.

> What package names should we have after refactoring?

Your list is perfect.

> Is it OK that public API classes will contain **public** methods with parameters with types from internal packages?

(added the second **public** to your question). No it's not OK, this would be a bad design smell. Public API methods should not point to internal types. If we can make those methods private or package protected, that's the easy solution.

monperrus:
just pushed doc change (+one visibility change).

pvojtechovsky:
I am done with moving of classes. Martin, it is your side now ;-)

> ParameterValueProvider from spoon.utils.

I do not understand what would you like to do with this class.

The class ParameterInfo (or at least some methods) might be interesting as API too. It is actually accessible from the API `Pattern#getParameterInfos` ... it lists the parameters of the Pattern. WDYT?

monperrus:
Summary of my latest changes:

* renamed classes to obtain PatternParameterConfigurator and InlinedStatementConfigurator
* added an annotation Experimental, first used for this feature
* reviewed all public methods of all API classes (renamed some of them, removed or reduce visibility of untested ones)
* heavily worked on documentation

pvojtechovsky:
> Summary of my latest changes:

I will check it today in the evening.

> in Pattern we have two responsibilities: matching code and generating code.
> Do you think we can move the methods related to generating code (substitute*, createType applyToType) to another class?

Why do you think generating of code doesn't belong to pattern and matching belongs there?
Both features have same weight from my point of view. So 
A) they both belong there
B) they both belong somewhere else
... but I do not persist on that ;-) ... if your feeling is different.

In case of B) I suggest 
```java
class Patter {
PatternMatcher match() {...}
PatternGenerator generate() {...}
}
```
WDYT?

doc/pattern.md
```diff
@@ -0,0 +1,183 @@
...
+// creates pattern from the body of method "matcher1"
+Pattern t = PatternBuilder.create(
+    new PatternBuilderHelper(fooClass).setBodyOfMethod("matcher1").getPatternElements())
+    .configurePatternParameters()
+    .build();
+```
+
+
+This pattern matches all statements of the body of method `statement`, ie. a precondition to check that a list  is smaller than a certain size. 
+This pattern has  one single pattern parameter called `_col_`, which is automatically considered as a pattern parameter because it is declared outside of the AST node. This automatic configuration happens when `configurePatternParameters` is called.
```
pvojtechovsky: The keywords `statement` and `_col_` are wrong with updated example

monperrus:
pushed some changes (removed ParameterValueProviderFactory)

monperrus:
> Why do you think generating of code doesn't belong to pattern and matching belongs there?

I'm only saying that they are two different responsibilities. But sill, the concept of "Pattern" is more related to matching to my understanding and the meaning of methods forEachMatch and getMatches is super clear to me (and hopefully for the clients).  

Based on your proposal what do you think of:

```
class Pattern {
forEachMatch
getMatches
Generator generator() {...}
}
```

Main question, for the generation, we also use pattern parameters? how do we specify what to put in the parameters?

doc/pattern.md
monperrus: could you fix them directly? that would likely be more effective.

pvojtechovsky:
I have found these problems:
1) rename of all similar methods to `configurePatternParameters` was not good idea. There is mainly naming conflict in:
* configurePatternParameters(Map)
* configurePatternParameters(Consumer<PatternParameterConfigurator>)

which is visible when `configurePatternParameters(null)` has to be called, which is OFTEN needed to create pattern parameters from 
A) field annotated by @Parameter 
B) field with type `TemplateParameter`

2) rename of all similar methods to 'byVariable' cannot be done, because `byVariable` and (new) `byFieldRefOfVariable` does different thinks. 
This is new comment added to `byFieldRefOfVariable` method
```java
class Params {
   int paramA;
   int paramB;
}
void matcher(Params p) {
   return p.paramA + p.paramB;
}
//called with `byFieldRefOfVariable("p")` will create pattern parameters: `paramA` and `paramB`
```
I hope it is clear now what is difference between byVariable and `byFieldRefOfVariable`.

3) TemplateBuilder has to call PatternBuilderHelper#removeSuperClass to remove `class T extends AbstractTemplate`  to create correct pattern from legacy template ... even if it is not tested by legacy template tests, this code is correct.

pvojtechovsky:
I am OK with move of generator behavior to `Pattern#generator()`

> Main question, for the generation, we also use pattern parameters? 

yes

> how do we specify what to put in the parameters?
Put that anything what can be converted to value expected by the place where parameter is evaluated. The conversion is done by implementation of `ValueConvertor`.

Just an example:
parameter value is 
* SomeType.class
* CtType which points to SomeType
* CtTypeReference which points to SomeType
* String "somePackage.SomeType"

all these values are understood as type reference and are converted depending on target place
* to type reference 
* to type
* to javadoc type name

monperrus:
Thanks for the explanation on byFieldRefOfVariable, questions:

* why do we need a `String...` parameter? could we instead chain the calls: `.byFieldRefOfVariable("params").byFieldRefOfVariable "item")`
* can we call it `byFieldAccessOnVariable`?

monperrus:
Does it mean that the we only support generating code with varying type references? Assume that the pattern parameter is a CtExpression, we cannot provide a new CtExpression for generation (as we would do with a template)?

doc/pattern.md
pvojtechovsky: I started to do it before I wrote this comment, but then I had feeling like there is too many changes and I do not know your vision, so my fix might ho in wrong direction:
A) to remove comment about _col_
B) to re-add the example you removed
... I would prefer B), but you already removed it so, you probably prefer something else :-)

pvojtechovsky:
> Does it mean that the we only support generating code with varying type references? 

no

> Assume that the pattern parameter is a CtExpression, we cannot provide a new CtExpression for generation (as we would do with a template)?

the parameter value is Object. So it accepts everything including CtExpression. The concept is simple:
* Generator knows what is expected value type at substitution place
* Generator knows what is current value of parameter
* Generator calls ValueConvertor to adapt current parameter value to expected type ... if parameter already has expected type, then ValueConvertor simply returns it

> why do we need a String... parameter? could we instead chain the calls: .byFieldRefOfVariable("params").byFieldRefOfVariable "item")

yes

> can we call it byFieldAccessOnVariable?

sure. I always welcome better name. 👍 

> ambiguous `configurePatternParameters` methods

may be we might move code of configurePatternParameters(Map) from PatternBuilder to PatternParameterConfigurator and rename it to:
* byAnnotationParameter - will handle `@Parameter`
* byTemplateParameter - will handler `TypeParameter _col_`

or may be we do not want to distinguish between annotation and template parameter and we want to have there only one method
* byPatternParameter

... or any better name ;-)

monperrus:
> > can we call it byFieldAccessOnVariable?

> sure. I always welcome better name. 👍 

Done


> may be we might move code of configurePatternParameters(Map) from PatternBuilder to PatternParameterConfigurator and rename it to:
> * byAnnotationParameter - will handle `@Parameter`
> * byTemplateParameter - will handler `TypeParameter _col_`

Perfect!

* Do you proceed with extracting the `Generator` class?
* Also we need to make a separate PR for the changes to `src/main/java/spoon/Metamodel.java` and  `src/main/java/spoon/reflect/*`

I've also started to look at the uncovered / untested new methods thanks to the Coverall report.

pvojtechovsky:
> byAnnotationParameter - will handle @Parameter
> byTemplateParameter - will handler TypeParameter _col_

it is not compatible with legacy templates, which allows this
```java
@Parameter
TemplateParameter<Void> block;
```
and then in legacy templates spoon calls now:
```java
configureParameters(pb->pb.byAnnotationParameter().byTemplateParameter())
```
and the parameter is processed twice, because it is both @Parameter and TemplateParameter :-(
... and there are related other ugly hidden "features" of legacy templates:
* The CtBlock value in TemplateParameter field is understood as list of statements
* The CtBlock value in @Parameter field is understood as CtBlock

pvojtechovsky:
I merged `byAnnotationParameter` and `byTemplateParameter` into one method `byTemplateParameter`. It solves problem mentioned in previous comment. 
WDYT?

Do you have better name then `byTemplateParameter` ?

pvojtechovsky:
I created `Generator` class and moved there generator related methods from `Pattern`.

I finish today.

monperrus:
working on this

monperrus:
I love the new generator design, with a clear responsibility, I moved things a little bit around such that the interface Generator is in the public API, and the rest is in the internal package.

Questions:
* the map of objects passed to the generator, the objects are mostly expected to be CtElement? something else?
* in the enums: ConflictResolutionMode, Quantifier, what are the default values?
* getModelValueResolver is public but not in the public API. I can live with that. Yet, it returns a `ModelNode` which is empty. Can we remove class `ModelNode`?

pvojtechovsky:
> the map of objects passed to the generator, the objects are mostly expected to be CtElement? something else?

yes, the values can be CtElement, primitive literals like String, Integer, ... and or their Map, List, Set

I will answer other questions when I have more time.

pvojtechovsky:
> in the enums: ConflictResolutionMode, Quantifier, what are the default values?

ConflictResolutionMode.FAIL is the default in most cases. But there are some standard Pattern builder algorithms (mainly these which deals with legacy Templates), which are using other modes.

Quantifier.GREEDY is default for parameters.

> getModelValueResolver is public but not in the public API. I can live with that. Yet, it returns a ModelNode which is empty. Can we remove class ModelNode?

I will remove it from public API and I will remove ModelNode ... working on it now

pvojtechovsky:
ModelNode deleted.

Pattern#getModelValueResolver is removed too, but it was little bit tricky to get access to protected members ... change it if you know better approach. I do not persist on that.

I finished here today.

monperrus:
> Pattern#getModelValueResolver is removed too

Awesome. The public API of Pattern is beautiful now.

Final step of this marathon PR: to test all new methods that are not covered yet.

pvojtechovsky:
Can we remove `SubstitutionVisitor`? It was a core of legacy Templates and it is not used now.

monperrus:
Yes, we can remove `SubstitutionVisitor`.

pvojtechovsky:
All API methods seems to be tested now. There some untested internal methods only.

monperrus:
Awesome, thanks.

SwitchNode seems to be an outlier in terms of number of untested lines (43) and coverage (60%), with large untested methods, see <https://coveralls.io/builds/17219444/source?filename=src%2Fmain%2Fjava%2Fspoon%2Fpattern%2Finternal%2Fnode%2FSwitchNode.java#L58>

Is there a way to better test it?

pvojtechovsky:
> SwitchNode seems to be an outlier in terms of number of untested lines
> Is there a way to better test it?

yes, to rollback one test you removed/simplified :-)
ok, I will keep the simplified test and rollback the origin test into new test.

monperrus:
perfect, thanks :+1:

sorry for that :-) I simply aimed at clear tests and contracts.

pvojtechovsky:
I was wrong. Adding of the removed part of `inlined if` test did not helped to cover more lines. :-)
So I made new test, which 
* generates code using `if`, where one can select what is generated using a parameter
* this pattern is made in more complicated way (configurePatternParameters after configureInlineStatements) to cover even more lines.

What next?

monperrus:
it seems that the next step is to merge #2016?

pvojtechovsky:
I just merged master. Now I will check whether new metamodel is still fitting to Pattern needs and then will refactor Pattern to use new Metamodel.

pvojtechovsky:
I moved MetamodelTest.testRuntimeMetamodel and related classes to #2050

monperrus:
Pavel, thanks a lot for this great contribution!

Having new and big features in Spoon is really awesome.

Next one on our roadmap: the sniper mode!

monperrus:
There is a regression on the template side, in https://github.com/SpoonLabs/spoon-examples, see https://ci.inria.fr/sos/job/spoon-examples/507/console

Could you have a look at it?

pvojtechovsky:
I will have a look in the evening ... if I make internet at home running after tonight storm ...

PR #1690, opened 2017-11-06
review: Fix: explicit final modifier should be kept in CtCatchVariable

### LINKED ISSUE #1689: bug: regression related to modifiers
Apparently some modifiers are still not correctly handled. 
We got an error when using Spoon with JUnit, see: 
> /builds/workspace/JUnit/target/generated-sources/spoon/org/junit/rules/Timeout.java:[97,91] error: local variable e is accessed from within inner class; needs to be declared final

in https://ci.inria.fr/sos/job/JUnit/858/console

surli:
In JDK5, final modifier should be explicit in CtCatchVariable to be used elsewhere in the code. 
Fix #1689


surli:
This PR should be fixed only after #1649 is merged.

PR #1697, opened 2017-11-08
chore: Implement class name shortening by a bunch of Freemarker macros.

metlos:
I played with Freemarker macros and was able to implement class name shortening - i.e. leaving out the package names and type parameters.

I was not able to test on spoon itself (I didn't see the rendered output as you have it on github) but I tested the macros with a number of changes for which it produced desirable output.


monperrus:
awesome! thank you very much!

PR #1702, opened 2017-11-09
review refactor: compute CtRole#forName without synonym

surli:
This refactor is an alternative of #1603 it just get rid of synonyms and only consider given role names. As @monperrus said it's not backward compatible but it's not a huge problem especially if it's integrated in Spoon 6.0.0


monperrus:
OK for me. 
Would you add a test?

tdurieux:
It can be a problem for me, I used it but I don't remember where ...

surli:
> It can be a problem for me, I used it but I don't remember where ...

So, when you'll have a crash you'll know where you used it 😄

monperrus:
I think this is a good PR that simplifies maintenance

src/main/java/spoon/reflect/path/CtRole.java
```diff
@@ -197,11 +109,18 @@ public static CtRole fromName(String name) {
...
 	public String getCamelCaseName() {
 		String s = name().toLowerCase();
-		int i = s.indexOf("_");
-		if (i != -1) {
-			s = s.substring(0, i) + Character.toUpperCase(s.charAt(i + 1)) + s.substring(i + 2);
+		String[] tokens = s.split("_");
+		if (tokens.length == 1) {
+			return s;
+		} else {
+			StringBuffer buffer = new StringBuffer(tokens[0]);
```
msteinbeck: I think you can use a StringBuilder here: https://stackoverflow.com/questions/355089/difference-between-stringbuilder-and-stringbuffer

PR #1706, opened 2017-11-09
review refactor: CtModifiable add/removeModifier

### LINKED ISSUE #1616: addModifier documentation out of sync
http://spoon.gforge.inria.fr/mvnsites/spoon-core/apidocs/spoon/reflect/declaration/CtModifiable.html

> <T extends CtModifiable> T addModifier(ModifierKind modifier)
> Returns: true if this element changed as a result of the call

However, the code does not return a Boolean:

``` .java
	@Override
	public <C extends CtModifiable> C addModifier(ModifierKind modifier) {
		if (modifiers == CtElementImpl.<ModifierKind>emptySet()) {
			this.modifiers = EnumSet.noneOf(ModifierKind.class);
		}
		getFactory().getEnvironment().getModelChangeListener().onSetAdd(this, MODIFIER, this.modifiers, modifier);
		modifiers.add(modifier);
		return (C) this;
	}
```

This documentation issue is applicable to all implementations of `addModifier`.

Recommend the Javadocs be updated to indicate whether there's a way to determine that a modifier was added versus already existed.

surli:
Fix the documentation for `CtModifiable#addModifier` (Fix #1616)
Change the signature of `CtModifiable#removeModifier` to be consistent

### DIFF
--- a/src/main/java/spoon/reflect/declaration/CtModifiable.java
+++ b/src/main/java/spoon/reflect/declaration/CtModifiable.java
@@ -59,7 +59,6 @@ public interface CtModifiable extends CtElement {
 	 * add a modifier
 	 *
 	 * @param modifier
-	 * @return <tt>true</tt> if this element changed as a result of the call
 	 */
 	@PropertySetter(role = MODIFIER)
 	<T extends CtModifiable> T addModifier(ModifierKind modifier);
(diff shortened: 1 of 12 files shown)


surli:
> could you add a test on the returned type / object of add/removeModifier?

Yes, in fact I was thinking about creating an enforcer rule for the set/get/add/remove methods of the different interfaces. But in another PR.

surli:
> could you add a test on the returned type / object of add/removeModifier?

Done.

DaveJarvis:
There appears to be a large amount of code duplication here that could be served by a common superclass.

PR #1707, opened 2017-11-09
review feature(import): add CtImport in the model

surli:
The goal of this PR is to manage imports properly with CtImport.


src/main/java/spoon/reflect/reference/CtImport.java
```diff
@@ -0,0 +1,31 @@
...
+ * knowledge of the CeCILL-C license and that you accept its terms.
+ */
+package spoon.reflect.reference;
+
+import spoon.reflect.declaration.CtElement;
+
+public interface CtImport extends CtElement {
+	<T extends CtImport> T setKindImport(ImportKind importKind);
+
+	ImportKind getKindImport();
```
monperrus: should be getImportKind

surli:
@monperrus @pvojtechovsky @tdurieux Even if every test is not passing, I'm curious about your insight on this PR. In particular, concerning the changes in the architecture.
 
CtImport extends CtReference and encapsulate a CtReference: it was easier for managing scanner, but I'm mixed about this architecture. I was first considering a CtImport as only extending CtElement. What seem the better choice for you? 

For now I propose to introduce CtImport in keeping the old behaviour of Spoon: imports are only contained in a CompilationUnit and are computed by ImportScanner. 
After this PR, I'll start to work on linking CtImport to CtType.

pvojtechovsky:
> CtImport extends CtReference ... I was first considering a CtImport as only extending CtElement.
> What seem the better choice for you?

Interesting question! 
I vote for CtImport extends CtElement. Why?

What is the difference between CtReference and CtElement?
D1) CtReference has getSimpleName/setSimpleName
The CtComment#getSimpleName() should reflect the `importKind`, but it doesn't

D2) CtReference has getDeclaration
The CtComment#getDeclaration() should cannot return declaration of all Types of the package if it is of kind `STAR_x`

D3) CtReference has unsetable Comment
The CtImport can have comments.

So all 3 differences seems to be NOT fitting to CtImport extends CtReference.

WDYT?

Does any code which uses CtImport profits from `getSimpleName` or `getDeclaration`?

src/main/java/spoon/support/reflect/reference/CtImportImpl.java
```diff
@@ -0,0 +1,101 @@
...
+import spoon.reflect.path.CtRole;
+import spoon.reflect.reference.CtImport;
+import spoon.reflect.reference.CtReference;
+import spoon.reflect.reference.ImportKind;
+import spoon.reflect.visitor.CtVisitor;
+import spoon.support.reflect.declaration.CtElementImpl;
+
+public class CtImportImpl extends CtElementImpl implements CtImport {
+	@MetamodelPropertyField(role = CtRole.IMPORT_KIND)
+	private ImportKind importKind;
```
pvojtechovsky: what about
```java
boolean includingChildren
```

src/main/java/spoon/support/reflect/reference/CtImportImpl.java
```diff
@@ -0,0 +1,101 @@
...
+	@Override
+	public <T extends CtImport> T setImportKind(ImportKind importKind) {
+		getFactory().getEnvironment().getModelChangeListener().onObjectUpdate(this, CtRole.IMPORT_KIND, importKind, this.importKind);
+		this.importKind = importKind;
+		return (T) this;
+	}
+
+	@Override
+	public ImportKind getImportKind() {
+		return this.importKind;
```
pvojtechovsky: what about
```
if (this.localReference instanceof CtTypeReference) {
  return includingChildren ? STAR_TYPE : TYPE;
}
...
```
That would assure that code in ElementPrinterHelper#writeImports will always work
```java
			switch (ctImport.getImportKind()) {
				case TYPE:
					CtTypeReference typeRef = (CtTypeReference) ctImport.getReference();
					importTypeStr = typeRef.getQualifiedName();
					if (!isJavaLangClasses(importTypeStr)) {
						setImports.add(importTypeStr);
					}
					break;

				case STAR_PACKAGE:
					CtPackageReference packageRef = (CtPackageReference) ctImport.getReference();
					importTypeStr = packageRef.getQualifiedName() + ".*";
					if (!isJavaLangClasses(importTypeStr)) {
						setImports.add(importTypeStr);
					}
					break;

				case METHOD:
					CtExecutableReference execRef = (CtExecutableReference) ctImport.getReference();
					if (execRef.getDeclaringType() != null) {
						setStaticImports.add(this.removeInnerTypeSeparator(execRef.getDeclaringType().getQualifiedName()) + "." + execRef.getSimpleName());
					}
					break;

				case FIELD:
					CtFieldReference fieldRef = (CtFieldReference) ctImport.getReference();
					setStaticImports.add(this.removeInnerTypeSeparator(fieldRef.getDeclaringType().getQualifiedName()) + "." + fieldRef.getSimpleName());
					break;

				case STAR_TYPE:
					CtTypeReference typeStarRef = (CtTypeReference) ctImport.getReference();
					importTypeStr = typeStarRef.getQualifiedName() + ".*";
					if (!isJavaLangClasses(importTypeStr)) {
						setStaticImports.add(importTypeStr);
					}
					break;
			}
```

pvojtechovsky:
Anyway I like this PR! It moves in good direction!

src/main/java/spoon/support/reflect/reference/CtImportImpl.java
surli: I'm not sure to understand your point here.

surli:
> What is the difference between CtReference and CtElement?
D1) CtReference has getSimpleName/setSimpleName
The CtComment#getSimpleName() should reflect the importKind, but it doesn't

I assume you mean `CtImport#getSimpleName()` here :)
Actually we can override getSimpleName to use ImportKind, so no big deal on this

> D2) CtReference has getDeclaration
The CtComment#getDeclaration() should cannot return declaration of all Types of the package if it is of kind STAR_x

IMHO the `CtImport#getDeclaration` should always return the contained reference of the CtImport. So even if you have a kind  `STAR_PACKAGE`, it will return the declaration of the package itself. I think it keeps the semantic of the `getDeclaration` on an import. 

> D3) CtReference has unsetable Comment
The CtImport can have comments.

This is a real problem. I did not check that when I implement it. 

> Does any code which uses CtImport profits from getSimpleName or getDeclaration?

It's hard to say as `CtImport` uses the scanner for `CtReference`, as in the current implementation it is a `CtReference`.
I'd say that `getSimpleName` is really used, at least for comparing `CtImports` in sets (see #1717) but I'm really not sure for `getDeclaration`.

So maybe I can inherit `CtImport` from `CtNamedElement`: it makes sense to me that a `CtImport` has a name.

src/main/java/spoon/support/reflect/reference/CtImportImpl.java
pvojtechovsky: I mean that current CtImport API is not safe for client's mistake. Client can create CtImport with kind = package, but providing CtTypeReference ... it is insconsistency.
That inconsistency might be solved if ctImport.getImportKind() would be DERIVED property and we would have instead CtImport#includingChildren which is not derived

pvojtechovsky:
> So maybe I can inherit CtImport from CtNamedElement

Interesting idea ...  
@monperrus WDYT about all these concepts above?

surli:
> So maybe I can inherit CtImport from CtNamedElement

I just made the changes: it shows that the `getDeclaration` was really never used. 
I also change implementation of `getSimpleName`. 
The whole bunch of test is passing so it looks like a good solution for me. 

Concerning CtComment I did not test for now to put comments on an import, I propose to see that later.

src/main/java/spoon/reflect/reference/ImportKind.java
```diff
@@ -0,0 +1,25 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE. See the CeCILL-C License for more details.
+ *
+ * The fact that you are presently reading this means that you have had
+ * knowledge of the CeCILL-C license and that you accept its terms.
+ */
+package spoon.reflect.reference;
+
+public enum ImportKind {
+	TYPE,
+	STAR_PACKAGE,
```
monperrus: STAR_PACKAGE->ALL_TYPES
STAR_TYPE -> ALL_STATIC_MEMBERS

src/test/java/spoon/reflect/visitor/CtScannerTest.java
```diff
@@ -180,14 +181,14 @@ public boolean matches(CtInvocation element) {
 				} else {
 					c.nbChecks++;
 					//System.out.println(invocation.toString());
-				}
 
-				// contract: the scan method is called with the same role as the one set on field / property
-				CtRole expectedRole = metaModel.getRoleOfMethod((CtMethod<?>)invocation.getExecutable().getDeclaration());
```
monperrus: why is this PR?

src/test/java/spoon/test/imports/ImportTest.java
```diff
@@ -1252,4 +1257,23 @@ public void testImportWithGenerics() throws IOException {
...
+		// contract: two imports of same kind with same reference should be equals
+		final Launcher launcher = new Launcher();
+
+		CtType typeA = launcher.getFactory().Type().get(A.class);
+
+		CtImport importsA1 = launcher.getFactory().createImport(ImportKind.TYPE, typeA.getReference());
+		CtImport importsA2 = launcher.getFactory().createImport(ImportKind.TYPE, typeA.getReference());
+
+		assertEquals(importsA1, importsA2);
+		assertEquals(importsA1.hashCode(), importsA2.hashCode());
```
monperrus: will disappear with rebase

src/main/java/spoon/reflect/factory/Factory.java
```diff
@@ -915,4 +918,10 @@
 	 * @see CoreFactory#create(Class<? extends CtElement>)
 	 */
 	CtElement createElement(Class<? extends CtElement> klass);
+
+	/**
+	 *
+	 * @see TypeFactory#createImport(ImportKind, CtReference)
+	 */
+	CtImport createImport(ImportKind importKind, CtReference reference);
```
monperrus: ImportKind -> CtImportKind

src/test/java/spoon/test/imports/ImportTest.java
surli: Actually no: this one is a new test related to CtImport. It's different from the test on CtReference introduced in the other PR.

surli:
@pvojtechovsky @monperrus for me it's ready.

src/main/java/spoon/support/reflect/reference/CtImportImpl.java
```diff
@@ -0,0 +1,89 @@
...
+ * circulated by CEA, CNRS and INRIA at http://www.cecill.info.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the CeCILL-C License for more details.
+ *
+ * The fact that you are presently reading this means that you have had
+ * knowledge of the CeCILL-C license and that you accept its terms.
+ */
+package spoon.support.reflect.reference;
```
pvojtechovsky: move it to different package, when it is no more `extends CtReference`
Including the CtImport interface.

pvojtechovsky:
`CtComment extends CtNamedElement` is OK for me. Thanks!

But I still have feeling like having `CtImportImpl.importKind` stored as attribute is wrong decision.
Each call of `CtImportImpl.setReference` must be followed by useless and error prone `CtImportImpl.setImportKind`, which MUST FIT to class of actually used reference, otherwise Spoon code will fails ClassCastException later during printing.

It would be simpler for clients to NOT HAVE `CtImportImpl.setImportKind` at all, but to have only
* CtImportImpl.getReference/setReference(CtReference)
* CtImportImpl.isImportAll/setImportAll(boolean)
* CtImportImpl.getImportKind() //would be derived property computed from class of `localReference` and value of `importAll`.

The Factory `createImport(CtImportKind importKind, CtReference reference)` would be replaced by
* createImport(CtReference reference) and createImportAll(CtReference reference) 
* createImport(CtReference reference, boolean importAll)

or you can keep `createImport(CtImportKind importKind, CtReference reference)` but do a check whether input is consistent and compute value of importAll from the importKind.

surli:
> or you can keep createImport(CtImportKind importKind, CtReference reference)  but do a check whether input is consistent and compute value of importAll from the importKind.

I really prefer to keep only this explicit constructor. But I agree with you that I should add a check between the ImportKind and the CtReference. I will add that.

pvojtechovsky:
OK, I hope you have some good reason to implement the model in a way, which allows inconsistencies and doubled storage of the same information (the type of import is stored in class of reference and in importKind enum too).
I can live with that - it is not big problem for me - just not nice. May be you are right ;-)

But please move CtImport(Impl) into correct package before merging.

surli:
> But please move CtImport(Impl) into correct package before merging.

Where do you want I move CtImport? For me it makes sense to put it with references.

pvojtechovsky:
> Where do you want I move CtImport?

Into `spoon.reflect.declaration`, same like all other implementations of CtNamedElement.
Note that all implementations of CtReference are in `spoon.reflect.reference`.

surli:
@pvojtechovsky so I finally change my mind after writing the ugly check for ImportKind. I really don't like to have a boolean to set in the import for imports of all static type members, but I don't see any other solution. And its usage should remain rare... 

@monperrus could you review this one with the last changes?

pvojtechovsky:
> I really don't like to have a boolean to set in the import for imports of all static type members, but I don't see any other solution. And its usage should remain rare...

I agree with You, but it is the best solution we have for now.

A wild idea: We might model `import static some.package.with,Type.*` using `CtWildcardStaticTypeMemberReference extends CtTypeReference`. Then we do not need any extra boolean property on CtImport. WDYT?

surli:
> We might model import static some.package.with,Type.* using CtWildcardStaticTypeMemberReference extends CtTypeReference. Then we do not need any extra boolean property on CtImport. WDYT?

I like the idea it keeps semantic of import and it does not change much in Spoon model.

src/main/java/spoon/reflect/path/CtRole.java
```diff
@@ -87,7 +87,9 @@
 	JAVADOC_TAG_VALUE,
 	POSITION,
 	SNIPPET,
-	ACCESSED_TYPE;
+	ACCESSED_TYPE,
+	IMPORT_REFERENCE,
+	IMPORT_ALL_TYPE_MEMBERS;
```
monperrus: this is really too low level compared to the concept of role. Is it possible to remove it?

src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
```diff
@@ -199,10 +200,16 @@
 	private CompilationUnit sourceCompilationUnit;
 
 	/**
+	 * Imports computed
+	 */
```
monperrus: longer name and documentation to explain the intention and usage of this new field?

should it be reset for each new type?

monperrus:
I'm OK with the PR but my two last comments. @pvojtechovsky if you're OK as well, would you merge?

pvojtechovsky:
> We might model import static some.package.with,Type.* using CtWildcardStaticTypeMemberReference extends CtTypeReference

May be Simon will go this way. So I am waiting for his final decision. That would solve one of your comments too.

pvojtechovsky:
It looks OK for me. Thank you! 
@surli Are You finished? Can I merge it? Or should we wait until Spoon 6.0?

surli:
> Can I merge it? Or should we wait until Spoon 6.0?

Yes, wait Spoon 6.0 :)

src/main/java/spoon/support/reflect/cu/CompilationUnitImpl.java
```diff
@@ -42,7 +42,7 @@
 
 	CtPackage ctPackage;
 
-	Collection<CtReference> imports = new HashSet<>();
+	Collection<CtImport> imports = new HashSet<>();
```
pvojtechovsky: What about using List instead of Set... here and in JDTImportBuilder#imports too, then we have a chance to keep the origin order of imports - minimize changes in generated sources.

pvojtechovsky:
I have found a conceptual mistake in this PR:
CtImportImpl#getReference() is not derived property, while CtNamedElement.getReference() is derived.
It causes that Equals checker never checks equality of CtImportImpl#localReference.
This problem must be solved before merging.

It shows that CtImport is PROBABLY a special kind of entity. It is not a reference and it is not a named element, because it cannot be referred.

I really do not know actually what is the good solution, but may be `CtImport` entity is not need at all.
 Because if have `CtWildcardStaticTypeMemberReference` now, then origin 
`Collection<CtReference> CompilationUnit#getImports()` might be probably enough...

May we just need new marker interface `CtImportableReference` and then
* CtFieldReference extends CtImportableReference
* CtExecutableReference extends CtImportableReference
* CtPackageReference extends CtImportableReference
* CtWildcardStaticTypeMemberReference extends CtImportableReference
* CtTypeReference extends CtImportableReference ... but here is the problem in subhierarchy of CtTypeReference, which is not importable ...

@surli, @monperrus, WDYT?

pvojtechovsky:
Simpler would be CtImport extends CtElement. Then it is clear that it is special entity and we have a place where to write import specific code.

surli:
> CtImportImpl#getReference() is not derived property, while CtNamedElement.getReference() is derived.

I agree with you that's a problem. 
I changed `CtImport` to inherit from CtElement.

pvojtechovsky:
@monperrus, is there any test which checks that all non derived properties are checked by EqualsChecker? May be the EqualsChecker might be implemented in future using RoleHandler - so we might get rid of this manually written code too.

monperrus:
> is there any test which checks that all non derived properties are checked by EqualsChecker?
I don't remember any.

monperrus:
oops, a conflict here.

src/main/java/spoon/Metamodel.java
```diff
@@ -146,6 +149,9 @@ private Metamodel() { }
 		result.add(factory.Type().get(spoon.reflect.reference.CtUnboundVariableReference.class));
 		result.add(factory.Type().get(spoon.reflect.reference.CtVariableReference.class));
 		result.add(factory.Type().get(spoon.reflect.reference.CtWildcardReference.class));
+		result.add(factory.Type().get(CtImport.class));
+		result.add(factory.Type().get(CtImportKind.class));
+		result.add(factory.Type().get(CtWildcardStaticTypeMemberReference.class));
```
monperrus: remove interface but keep implementation and neat polymorphism

src/test/java/spoon/test/prettyprinter/DefaultPrettyPrinterTest.java
```diff
@@ -177,11 +177,11 @@ public void autoImportUsesFullyQualifiedNameWhenImportedNameAlreadyPresent() thr
 
 		String expected =
 			"public void setFieldUsingExternallyDefinedEnumWithSameNameAsLocal() {" +nl+
-			"    localField = E1.ordinal();" +nl+
+			"    localField = spoon.test.prettyprinter.testclasses.sub.TypeIdentifierCollision.ENUM.E1.ordinal();" +nl+
```
monperrus: add an assert to the conflicting one which is written short

src/main/java/spoon/Metamodel.java
surli: Actually this is more complex than I was thinking as the whole metamodel (and associated tests) suppose that we have an interface for each implementation: if I want to put only implementation, I can't put any method to create it in factory etc.

monperrus:
> if I want to put only implementation, I can't put any method to create it in factory
why?

surli:
>  if I want to put only implementation, I can't put any method to create it in factory

Because `CtInheritanceScannerTest` - for example - uses methods defined in CoreFactory in its test.

monperrus:
what about having

CtYpeReference createwildcardStaticTypeMemberReference()?

which internally creates the appropriate implementation?

src/main/java/spoon/support/reflect/cu/CompilationUnitImpl.java
surli: > here and in JDTImportBuilder#imports too, then we have a chance to keep the origin order of imports - minimize changes in generated sources.

I agree, but I propose to do it in another PR: this one start to be difficult to understand. So I propose that you open an issue for that, and I'll make new tests and changes to assert that. WDYT?

surli:
@monperrus I propose to merge the PR as it is: your last comment concerning DefaultJavaPrettyPrinterTest is actually already checked few lines later, but not exactly with a call to `E1.ENUM` (it still use the name of the class).

Once we have this PR merged I want to check first if #1365 is improved, and then to completely refactor `ImportScanner` classes: those are a mess, and I can simplify them by using CtImport.

monperrus:
Week-end merge!

Thanks a lot @surli

Rahamathulla:
Version : Spoon 7.0.0

I am facing the similar or same issue in Spoon 7.0.0.
I m having sample code like this

package com.somelocalpack;
import javax.ejb.*;
public class BankMdb implements MessageDrivenBean, MessageListener {
......
}

When I parse this class

Set<CtTypeReference<?>> superInterfaces = element.getSuperInterfaces();

The superInterfaces ( MessageDrivenBean, MessageListener ) are identified as
com.somelocalpack.MessageDrivenBean & com.somelocalpack.MessageListener instead of 
javax.ejb.MessageDrivenBean and javax.ejb.MessageListener

monperrus:
Thanks for the bug report. This is normal, because the jar file containing `javax.ejb.*` is not in the classpath. Next time, could you please create a new issue? Thanks!

PR #1709, opened 2017-11-10
URGENT review: Fix typo in RoleHandlerGenerator

surli:


surli:
The combination of two PR breaks Spoon, this PR fixes it.

PR #1711, opened 2017-11-10
review fix: issue when using recursively defined annotation in shadow class

### LINKED ISSUE #1698: StackOverflowError using Spoon with Spring dependencies
When using Spoon with Spring dependencies, e.g.
```
	<dependency>
		<groupId>org.springframework</groupId>
		<artifactId>spring-webmvc</artifactId>
		<version>4.3.12.RELEASE</version>
	</dependency>
```

I got a `StackOverflowError` at line 504 of class `spoon.reflect.factory.TypeFactory` inside the method `public <T> CtType<T> get(Class<?> cl)` with parameter `interface org.springframework.cache.annotation.Cacheable`

My code is dependent of Spring (e.g. some methods have a `@Cacheable` annotation) but without the Spring dependency inside the Spoon classpath, the problem do not occur.

I'm using Spoon library:
```
  	<dependency>
  		<version>5.9.0</version>
  	</dependency>
```

surli:
Fix #1698

### DIFF
--- a/src/main/java/spoon/support/visitor/java/JavaReflectionTreeBuilder.java
+++ b/src/main/java/spoon/support/visitor/java/JavaReflectionTreeBuilder.java
@@ -252,6 +252,11 @@ public class JavaReflectionTreeBuilder extends JavaReflectionVisitorImpl {
 
 	@Override
 	public void visitMethod(RtMethod method) {
+		this.visitMethod(method, null);
+	}
+
+	@Override
+	public void visitMethod(RtMethod method, Annotation parent) {
 		final CtMethod<Object> ctMethod = factory.Core().createMethod();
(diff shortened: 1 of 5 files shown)


surli:
The fix here consists in checking that when visiting an annotation for shadow class creation, the parent call wasn't already the visit of an annotation of the same type. 
It means that in the given `AliasFor` annotation (see the test), if you inspect the fields, they won't have any annotation on them even if in the file definition it uses `AliasFor` annotation.

So, it's not perfect but it was the easiest way to propose a fix in that kind of case.

monperrus:
OK for me, thanks @surli . Will merge it on Monday if everybody agrees.

monperrus:
thanks @surli!

PR #1718, opened 2017-11-13
review: Fix test assertion with JDK8 on Travis

surli:
Travis move from Java 1.8.0_144 to Java 1.8.0_151 for their JDK8. 
Our assertion no longer pass on CtTypeInformationTest, this PR fixes it.


monperrus:
now Gakoci fails. 

we should have 61 <= expectedNumberInJDK8 <= 62

PR #1724, opened 2017-11-14
review fix: fix bug with repeatable annotations

### LINKED ISSUE #1722: Unwanted factorized annotation values
I am currently writing a program aiming at refactoring tests from junit 4 to 5.
To do so, I want to change Categories to Tags.

Here is the junit4 version:
`@Category({SlowTests.class, FastTests.class})`

Here is the expected result in junit 5 wanted to be produced (one tag per category):
```
@Tag("SlowTests")
@Tag("FastTests")
```

Here is my code, a for-loop on a list:
```
for (String tag : tagList) {
            method.getFactory().Annotation().annotate(method, Tag.class,"value",tag);
        }
```

Here is the result:
`@Tag({ "TestAnnotation.SlowTests" , "TestAnnotation.FastTests" })`

Any idea why the tags are "factorized"?
Thanks !

surli:
fix #1722

### DIFF
--- a/src/main/java/spoon/reflect/factory/AnnotationFactory.java
+++ b/src/main/java/spoon/reflect/factory/AnnotationFactory.java
@@ -16,6 +16,7 @@
  */
 package spoon.reflect.factory;
 
+import spoon.SpoonException;
 import spoon.reflect.declaration.CtAnnotation;
 import spoon.reflect.declaration.CtAnnotationType;
 import spoon.reflect.declaration.CtElement;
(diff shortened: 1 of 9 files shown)


surli:
This PR intends to fix #1722 but it also fixes cases when trying to annotate twice on an annotation type that does not take an Array: the current behavior was to create an array nevertheless, now it throws a SpoonException. 

I also had to make a choice when calling repeatedly annotate with a String value on this kind of annotation:
```java
@Repeatable(Tags.class)
public @interface Tag {
    String[] value();
}
```

We could have indeed those different results: 
```java
@Tag("value1")
@Tag("value2")
@Tag("value3")
void foo1();

@Tag({"value1", "value2"})
@Tag("value3")
void foo2();

@Tag({"value1", "value2", "value3"})
void foo3();
```

Previous behaviour did not take into account the ```@Repeatable``` meta-annotation of annotation, so the results corresponding to `foo3` were used (then issue #1722), but now with the fix proposed I consider that we should check in priority if the annotation is repeatable or not: if it is, then it creates a new annotation in priority. 
My guess is that if the user really wants to put an array of objects in the value, then it can always annotate with the array of objects, instead of calling multiple times annotate.

pvojtechovsky:
I have checked this PR and it looks good for me.  I would merge it but the meaning of word "factorized" in the title is not clear for me. So I am just not sure with merge comment ...

monperrus:
fixed title. ready for merge!

PR #1730, opened 2017-11-16
review feat: Add Java9 CtModule to Spoon model

surli:


surli:
@monperrus @pvojtechovsky I'm starting to work on Java 9 modules to add the concept in Spoon model. Based on the JLS a module has the following definition (see: https://docs.oracle.com/javase/specs/jls/se9/html/jls-7.html#jls-7.7):

```
ModuleDirective:
requires {RequiresModifier} ModuleName ;
exports PackageName [to ModuleName {, ModuleName}] ;
opens PackageName [to ModuleName {, ModuleName}] ;
uses TypeName ;
provides TypeName with TypeName {, TypeName} ;
RequiresModifier:
(one of)
transitive static
```

I don't know how to express properly the different directives, without creating new interfaces in the model to represent them. Could you give me a quick opinion about the first proposed hierarchy of classes: then I'll be able to work on it.

monperrus:
A quick proposal:

|ModuleDirective: -> CtModule ModuleName -> get/setSimpleName PackageName -> CtPackageReference
TypeName -> CtTypeReference transitive static -> add transitive to ModifierKind|

pvojtechovsky:
> |ModuleDirective: -> CtModule ModuleName -> get/setSimpleName PackageName -> CtPackageReference
TypeName -> CtTypeReference transitive static -> add transitive to ModifierKind|

@monperrus, I don't understand this language ;-)

> Could you give me a quick opinion about the first proposed hierarchy of classes

Note: I have zero exprience with java9 modules. I just read some tutorials and specification linked by Simon. I have some questions ...

Q1) May be we should introduce a concept of a `CtModuleMember` (similar to CtTypeMember), to have one **ORDERED** list of all `CtModuleMember`s. Then each CtModule member should exist as own entity extending CtModuleMember.

Q2) Is `CtModuleExport extends CtPackageReference` correct? Or should we use
```java
CtModuleExport extends CtModuleMember {
  CtPackageReference package;
  Set<CtModuleReference> targetExport;
}
```
Q3) Are comments allowed in module-info.java filed? If [yes](https://issues.apache.org/jira/browse/MCOMPILER-289), then each CtModule member should NOT extend a `CtReference` (reference is not allowed to have comments).

src/main/java/spoon/reflect/CtModelImpl.java
```diff
@@ -39,6 +43,23 @@
 
 	private static final long serialVersionUID = 1L;
 
+	@Override
+	public <R extends CtElement> CtQuery filterChildren(Filter<R> filter) {
+		return getUnnamedModule().getFactory().Query().createQuery(this.getAllModules().toArray()).filterChildren(filter);
```
pvojtechovsky: `QueryFactory#createQuery` has actually parameter type Object, not Object[], so this code will not work as expected. We have to add new QueryFactory#createQuery(Object[]) method.

surli: My bad, that's because `CtQueryImpl` constructor as an `Object...` arg. I'll fix that.

src/main/java/spoon/reflect/CtModel.java
```diff
@@ -40,6 +42,16 @@
 	/** returns all packages of the model */
 	Collection<CtPackage> getAllPackages();
 
+	/**
+	 * returns the unnamed module
```
monperrus: returns the unnamed module. Warning: if there are other modules, they are not contained in that unnamed root module.

src/main/java/spoon/reflect/CtModel.java
```diff
@@ -29,9 +31,9 @@
 /** represents a Java program, modeled by a set of compile-time (Ct) objects
  * where each objects is a program element (for instance, a CtClass represents a class).
  */
-public interface CtModel extends Serializable {
+public interface CtModel extends Serializable, CtQueryable {
 
-	/** returns the root package */
+	/** returns the root package of the unnamed module */
```
monperrus: If there are several modules, it throws an exception

src/main/java/spoon/Metamodel.java
```diff
@@ -146,6 +146,11 @@ private Metamodel() { }
 		result.add(factory.Type().get(spoon.reflect.reference.CtUnboundVariableReference.class));
 		result.add(factory.Type().get(spoon.reflect.reference.CtVariableReference.class));
 		result.add(factory.Type().get(spoon.reflect.reference.CtWildcardReference.class));
+		result.add(factory.Type().get(spoon.reflect.declaration.CtModule.class));
+		result.add(factory.Type().get(spoon.reflect.declaration.CtModuleRequirement.class));
```
monperrus: CtModuleProvidedService should be called  CtProvidedService

add CtUsedService

pvojtechovsky:
> Q1) May be we should introduce a concept of a CtModuleMember (similar to CtTypeMember), to have one ORDERED list of all CtModuleMembers. Then each CtModule member should exist as own entity extending CtModuleMember.

no reaction here? Did you understood this problem? Will be spoon able to distinguish between these two cases?
```java
module com.example.foo {
    requires com.example.foo.http;
    exports com.example.foo.bar;
}
```
and 
```java
module com.example.foo {
    exports com.example.foo.bar;
    requires com.example.foo.http;
}

... one of the bigger problem which avoids using spoon productive is that it changes formatting of the pretty printed source files. So it cannot be easily applied to existing sources to let it change few things only.

I think it makes sense to keep order of module directives to minimize changes between origin and pretty printed code.

**Comment by monperrus** [comment]
> May be we should introduce a concept of a CtModuleMember (similar to CtTypeMember), to have one ORDERED list of all CtModuleMembers.

That's a good idea.

File: src/main/java/spoon/experimental/modelobs/FineModelChangeListener.java [hunk 5]
```diff
@@ -51,6 +52,8 @@
 
 	void onSetAdd(CtElement currentElement, CtRole role, Set field, ModifierKind newValue);
 
+	void onSetAdd(CtElement currentElement, CtRole role, Set field, CtModuleRequirement.RequiresModifier newValue);
```
pvojtechovsky: Are we really going to have copy of this method (and others) for each non CtElement type?

surli:
@monperrus @pvojtechovsky I consider this PR as finished: it's a first support of modules in Spoon. It will be followed by other PR to fully supports module later, when the tools will be more matured. 

I first planned to add more tests, especially to consider compiling project with multiple modules. However I have several issue doing that: intelliJ currently does not support putting multiple files named "module-info.java" in resources (it crashes the compilation...), and the current release of JDT is buggy concerning package references. The bug has been fixed (see: https://bugs.eclipse.org/bugs/show_bug.cgi?id=525603) but we have to wait next release.

Tell me if you see other interesting things to test, that do not imply to add module-info.java files in resources, nor to compile a project with module.

src/main/java/spoon/reflect/path/CtRole.java
```diff
@@ -99,7 +99,8 @@
 	OPENED_PACKAGE,
 	SERVICE_TYPE,
 	IMPLEMENTATION_TYPE,
-	PROVIDED_SERVICE;
+	PROVIDED_SERVICE,
+	MODULE_MEMBER;
```
pvojtechovsky: MODULE_MEMBER is super role of other roles. Move it before the first sub rule, and then use `CtRole(CtRole)` constructor on the sub roles to link them.

src/main/java/spoon/support/reflect/declaration/CtModuleImpl.java
```diff
@@ -0,0 +1,446 @@
...
+		if (this.moduleMembers.isEmpty()) {
+			return CtElementImpl.emptyList();
+		} else {
+			List<CtUsedService> usedServices = new ArrayList<>();
+			for (CtModuleMember moduleMember : this.moduleMembers) {
+				if (moduleMember instanceof CtUsedService) {
+					usedServices.add((CtUsedService) moduleMember);
+				}
+			}
+			return usedServices;
```
pvojtechovsky: I suggest that all such geters return unmodifiable collection - to make it clear to clients that it is not read/write collection.

src/main/java/spoon/reflect/declaration/CtModule.java
```diff
@@ -0,0 +1,184 @@
...
+public interface CtModule extends CtNamedElement {
+
+	/**
+	 * The name for the top level module.
+	 */
+	String TOP_LEVEL_MODULE_NAME = "unnamed module";
+
+	/**
+	 * Returns the unnamed module
+	 * Warning: if there are other modules, they are not contained in that unnamed root module.
```
pvojtechovsky: I do not understand this comment ... the method returns `boolean` ...

src/main/java/spoon/reflect/declaration/CtModule.java
```diff
@@ -0,0 +1,184 @@
...
+	<T extends CtModule> T addModuleMemberAt(int position, CtModuleMember moduleMember);
+
+	@PropertyGetter(role = MODULE_MEMBER)
+	List<CtModuleMember> getModuleMembers();
+
+	@PropertySetter(role = MODULE_MEMBER)
+	<T extends CtModule> T removeModuleMember(CtModuleMember moduleMember);
+
+	@PropertySetter(role = MODIFIER)
+	<T extends CtModule> T setIsOpenModule(boolean openModule);
```
pvojtechovsky: what about move it up, next to isOpenModule?

src/main/java/spoon/reflect/declaration/CtModuleMember.java
```diff
@@ -0,0 +1,9 @@
+package spoon.reflect.declaration;
+
+/**
+ * Represents a component from a module
+ */
+public interface CtModuleMember extends CtElement {
```
pvojtechovsky: Note: the java specification calls it `ModuleDirective`. So why we invent new name `ModuleMember`? I personally have no problem with both names, so if You think ModuleMember is good one. I will like it too ;-)

src/main/java/spoon/reflect/factory/Factory.java
```diff
@@ -787,6 +795,11 @@
 	 */
 	CtQuery createQuery(Object input);
 
+	/**
+	 * @see QueryFactory#createQuery(Object...)
+	 */
+	CtQuery createQuery(Object[] input);
```
pvojtechovsky: would that `createQuery(Object... input)` work?

src/main/java/spoon/reflect/factory/ModuleFactory.java
```diff
@@ -0,0 +1,148 @@
...
+			visitor.visitCtModule(this);
+			getFactory().Module().getAllModules().forEach(module -> {
+				if (!module.getSimpleName().equals(CtModule.TOP_LEVEL_MODULE_NAME)) {
+					module.accept(visitor);
+				}
+			});
+		}
+	}
+
+	final Map<String, CtModule> allModules = new HashMap<>();
```
pvojtechovsky: Modules shouldn't be hold by ModuleFactory, because one Spoon factory might be used to create more `CtModels`.
Actually I suggest to move `allModules` into `CtUnnamedModule` and to register CtModule there when `CtModule.setParent(CtElement unnamedModule)` is called

src/main/java/spoon/support/DefaultCoreFactory.java
```diff
@@ -938,5 +968,45 @@ public CtTypeReference createWildcardStaticTypeMemberReference() {
 		result.setFactory(getMainFactory());
 		return result;
 	}
+	public CtModule createModule() {
+		CtModule module = new CtModuleImpl();
+		module.setFactory(getMainFactory());
+		return module;
```
pvojtechovsky: may be call `setParent(unnamedModule)` directly here? So it can be called directly by client?

src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java
```diff
@@ -657,4 +672,147 @@ public boolean onAccess(char[][] tokens, int index) {
...
+		}
+
+		if (moduleDeclaration.services != null && moduleDeclaration.services.length > 0) {
+			List<CtProvidedService> moduleProvidedServices = new ArrayList<>();
+			for (ProvidesStatement providesStatement : moduleDeclaration.services) {
+				moduleProvidedServices.add(this.createModuleProvidedService(providesStatement));
+			}
+
+			module.setProvidedServices(moduleProvidedServices);
+		}
```
pvojtechovsky: The order of ModuleDirectives in source code is not kept. May be we might use ModuleStatement#declarationSourceStart to sort them after they are added into CtModule.

src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java
```diff
@@ -657,4 +672,147 @@ public boolean onAccess(char[][] tokens, int index) {
...
+
+		return module;
+	}
+
+	CtUsedService createUsedService(UsesStatement usesStatement) {
+		CtTypeReference typeReference = this.jdtTreeBuilder.references.getTypeReference(usesStatement.serviceInterface);
+		return this.jdtTreeBuilder.getFactory().Module().createUsedService(typeReference);
```
pvojtechovsky: missing source position?

pvojtechovsky:
> intelliJ currently does not support putting multiple files named "module-info.java" in resources

An idea ... may be you can create "module-info-template.java" files and copy them into a special compilation directory and rename them just before the test starts. All the module-info based tests would use the same special compilation directory and would clean it before they start. So there is always only one module-info.java.

>  things to test

* de/serialization of Spoon model, which contains some modules

Anyway it is great PR! Thank You for that Simon!

src/main/java/spoon/reflect/declaration/CtModuleMember.java
monperrus: agree

monperrus:
for a such an important new feature, I propose to add some documentation in `/doc/`

src/main/java/spoon/reflect/factory/Factory.java
surli: Actually it should not be there, as I created those methods in #1739  I'll remove it.

src/main/java/spoon/reflect/factory/ModuleFactory.java
surli: Indeed, it's something we discuss with Martin and I forget to get rid of it, thanks :)

src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java
surli: I will create a unit test to check that and fix it

src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java
surli: Yes you're right !

surli:
> An idea ... may be you can create "module-info-template.java" files and copy them into a special compilation directory and rename them just before the test starts. All the module-info based tests would use the same special compilation directory and would clean it before they start. So there is always only one module-info.java.

Yes, I think about it later and I will do something like that to add some more tests, for now. I'll see if there is any bug reported to IntelliJ about this bug. 

> de/serialization of Spoon model, which contains some modules

I'll look into that.

src/main/java/spoon/reflect/declaration/CtModule.java
surli: Yeah it was a confusion between CtModel and CtModule...

src/main/java/spoon/support/DefaultCoreFactory.java
surli: I don't agree with you on this point.

surli: I don't agree with you on this point: in my understanding, the default core factory is used to create model element, without given them any state except for the factory.

However, in order to put the module in the map, you need to have its name: you cannot have twice a module with the same name. 
So I vote to explicitely put in `ModuleFactory#getOrCreate` the mechanism to create a module and to put it in the map of modules properly. And to keep here, the creation of a simple CtModule detached from the model.

surli:
Ok @pvojtechovsky I update my code according to your reviews, but I'm now stuck... 

I provided more tests to Spoon, and I do not manage to fix the last one.
First of all, I think this PR is officially breaking behaviour: I completely removed the previous implementation of `CtRootpackage` because it was conceptually broken.
CtUnnamedModule is now the root element, so this one do not have any parent, and you can get the root package, by calling `getRootPackage` on it: then it does not make sense that the `CtRootPackage` do not provide any parent. I kept the class name for now, but in long term view this class should disappear. 

Now I have another issue, which is the binding between a package and its supermodule: in JDTTreeBuilder, we create packages when discovering them in TypeDeclaration. IMHO we should get the proper module here and create the package in specifying its parent module here (this is why I changed PackageFactory). However I did not find any way in JDT to find the parent module of a type in TypeDeclaration or CompilationUnitDeclaration: all modules referenced is the unnamed module currently.
So maybe it's related to the JDT bug that prevents me to compile a project with modules, I'm not sure.

src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java
surli: This is managed directly in CtModuleImpl: I'm using a sorted list which uses the position to put elements in the right order. This is checked by a test.

pvojtechovsky:
> I propose, at least, to remove the implementation of CtRootPackage even if we keep the class, for backward compatibility reason.

I am ok wit that idea. Optionaly we might just modify implementation of root package and
1) to have rootPackage.getParent() == module
2) to have one instance of root package for each module.
It would keep old concept of root package together with new modules. But may be discuss it with @monperrus. I do not know concept of java modules well enough.

pvojtechovsky:
> I propose, at least, to remove the implementation of CtRootPackage even if we keep the class, for backward compatibility reason.

I am ok wit that idea. Optionaly we might just modify implementation of root package and
1) to have rootPackage.getParent() == module
2) to have one instance of root package for each module.
It would keep old concept of root package together with new modules. But may be discuss it with @monperrus. I do not know concept of java modules well enough.

src/main/java/spoon/reflect/factory/ModuleFactory.java
```diff
@@ -0,0 +1,163 @@
...
+
+			this.setRootPackage(new CtModelImpl.CtRootPackage());
+			this.addModule(this);
+		}
+
+		public boolean addModule(CtModule module) {
+			if (allModules.containsKey(module.getSimpleName())) {
+				return false;
+			}
+			allModules.put(module.getSimpleName(), module);
```
pvojtechovsky: I see one more problem. Later call of CtModule.setSimpleName(...) will cause inconsistency in modules, because the module is registered under old name. 
May be we should store modules as List<CtModule> and `getModule` should always search that list by name.
Then you can add the module into that list sooner directly in `factory.Core().createModule()` call - at time when it has no name. WDYT?

src/main/java/spoon/reflect/factory/ModuleFactory.java
```diff
@@ -0,0 +1,163 @@
...
+
+	public CtModule getOrCreate(String moduleName) {
+		if (moduleName == null || moduleName.isEmpty()) {
+			return getUnnamedModule();
+		}
+
+		CtModule ctModule = getUnnamedModule().getModule(moduleName);
+		if (ctModule == null) {
+			ctModule = factory.Core().createModule().setSimpleName(moduleName).setParent(getUnnamedModule());
+			getUnnamedModule().addModule(ctModule);
```
pvojtechovsky: What about move of this code `getUnnamedModule().addModule(ctModule);` into `factory.Core().createModule()`. WDYT?

src/main/java/spoon/reflect/CtModel.java
```diff
@@ -41,6 +42,17 @@
 	/** returns all packages of the model */
 	Collection<CtPackage> getAllPackages();
 
+	/**
+	 * Returns the unnamed module.
+	 * Warning: if there are other modules, they are not contained in that unnamed root module, use getAllModules instead.
```
pvojtechovsky: Is it still true? I guess `getUnnamedModule().getAllModules()` now works same like Model.getAllModules(). OR?

src/main/java/spoon/reflect/CtModelImpl.java
```diff
@@ -135,12 +139,22 @@ public CtPackage getRootPackage() {
...
 	}
 
+	@Override
+	public CtModule getUnnamedModule() {
+		return this.unnamedModule;
+	}
+
+	@Override
+	public Collection<CtModule> getAllModules() {
+		return this.unnamedModule.getFactory().Module().getAllModules();
```
pvojtechovsky: Shorter `return this.unnamedModule.getAllModules();` should return the same result now.

src/main/java/spoon/support/reflect/declaration/CtModuleImpl.java
```diff
@@ -0,0 +1,467 @@
...
+
+	@Override
+	public <T extends CtModule> T setModuleDirectives(List<CtModuleDirective> moduleDirectives) {
+		getFactory().getEnvironment().getModelChangeListener().onListDeleteAll(this, CtRole.MODULE_DIRECTIVE, this.moduleDirectives, new ArrayList<>(this.moduleDirectives));
+		if (moduleDirectives == null || moduleDirectives.isEmpty()) {
+			this.moduleDirectives = CtElementImpl.emptyList();
+			return (T) this;
+		}
+		if (this.moduleDirectives == CtElementImpl.<CtModuleDirective>emptyList()) {
+			this.moduleDirectives = new SortedList<>(new CtLineElementComparator());
```
pvojtechovsky: I suggest to sort directives only once at time when CtModule is created by JDT builder. I think, later automatic sorting of client specific module directives by position is not expected by clients. WDYT?

src/main/java/spoon/support/reflect/declaration/CtModuleImpl.java
```diff
@@ -0,0 +1,467 @@
...
+	public CtModule clone() {
+		return (CtModule) super.clone();
+	}
+
+	@Override
+	public <T extends CtElement> T setParent(T parent) {
+		if (!this.getSimpleName().equals(CtModule.TOP_LEVEL_MODULE_NAME) && parent != getFactory().Module().getUnnamedModule()) {
+			throw new SpoonException("The parent of a module should necessarily be the unnamed module.");
+		}
+		return (T) super.setParent(parent);
```
pvojtechovsky: May be we can mark setParent here as Derived property and ignore call of setParent here?
I guess there is no case when somebody should be able to set different parent then `unnamedModule`...

surli:
> I am ok wit that idea. Optionaly we might just modify implementation of root package and
    1. to have rootPackage.getParent() == module
    2. to have one instance of root package for each module.
    It would keep old concept of root package together with new modules.

I'm ok with those idea. I'll try to implement it in that way.

src/main/java/spoon/reflect/factory/ModuleFactory.java
surli: > I see one more problem. Later call of CtModule.setSimpleName(...) will cause inconsistency in modules, because the module is registered under old name.

Agree that's a problem. I'll try to produce a test to show the bug, and fix it the way you propose.

src/main/java/spoon/reflect/CtModel.java
surli: No I think you're right.

src/main/java/spoon/reflect/CtModelImpl.java
surli: Indeed.

src/main/java/spoon/support/reflect/declaration/CtModuleImpl.java
surli: I don't agree. I think we should document that the module directives are ordered by position. Now I think users don't put position attribute in the element when they're modifying the model. So they will use the method `addModuleDirectiveAtPosition` when they want to specify a position.

src/main/java/spoon/support/reflect/declaration/CtModuleImpl.java
surli: I agree, it will avoid unexpected behaviours there.

surli:
So I'm kind of stuck with two problems related to the JDT usage right now: 
   1. I don't manage to compile source code containing modules with JDT
   2. I don't manage to get the module of a given type with JDT.

The master problem around those is that I don't find any proper documentation on how JDT handles modules. Actually I got some piece of information in JDT bug tracker, but I don't know if it's reliable and my experiments don't show many improvement: I suspect that the next release of JDT should add improvement on modules, but I'm not sure right now.

I was even trying to compile myself JDT from the sources to see if it helps, but even then the documentation is either hidden either inexistent. 
So, my guess would be to put the two failing tests with a `@Ignore`, and to wait the next JDT release to see if it improves things. 

WDYT @monperrus @pvojtechovsky?

pvojtechovsky:
> So, my guess would be to put the two failing tests with a @Ignore, and to wait the next JDT release to see if it improves things.

I agree that we should merge this work even if it is not 100% finished.

surli:
So I added documentation and ignored the failing tests. For me it's ready for merging @pvojtechovsky @monperrus 

Once it's merged I'll create issues for the remaining unsupported features associated with modules.

monperrus:
excellent! if everybody agrees I'll merge it on Monday.

pvojtechovsky:
I agree, Nice code and documentation! Thank You Simon!

surli:
@monperrus it's ready for me too

monperrus:
Thanks a lot for the hard and deep work!

Issue #1629, opened 2017-10-20
Trying to find the imports of a class model


krlu:
I have the following scala code: 

```
    val launcher = new Launcher
    launcher.addInputResource("SomeClass.java")
    launcher.getEnvironment.setAutoImports(true)
    launcher.getEnvironment.setNoClasspath(true)
    launcher.buildModel()
    val model: CtModel = launcher.getModel
    println(model.getAllTypes)
```
However, when I print `model.getAllTypes` I don't see the import statements at the top being printed. Is there a way for me to get them? I noticed there is an `ImportScanner` object, but it looks like I'd have to go into each element to find the relevant imports.

surli:
Hi @krlu 

For now the import are not directly part of Spoon model, so cannot get computed imports unless you use yourself the ImportScanner.
However, if you only want to retrieve the original imports, you can get them through CompilationUnit:

```java
String path = "path/of/compilationUnit.java";
CompilationUnit cu = launcher.getFactory().CompilationUnit().create(path);
cu.getImports(); // return a set of CtReference
```

You can retrieve your interesting path from a CtType by using `getPosition().getFile()`.

krlu:
Thanks so much!

kenail2002:
> Hi @krlu
> 
> For now the import are not directly part of Spoon model, so cannot get computed imports unless you use yourself the ImportScanner. However, if you only want to retrieve the original imports, you can get them through CompilationUnit:
> 
> ```java
> String path = "path/of/compilationUnit.java";
> CompilationUnit cu = launcher.getFactory().CompilationUnit().create(path);
> cu.getImports(); // return a set of CtReference
> ```
> 
> You can retrieve your interesting path from a CtType by using `getPosition().getFile()`.

perfect

PR #1732, opened 2017-11-17
doc: Bump version number after release 6.0.0

surli:


README.md
```diff
@@ -90,7 +90,7 @@ Snapshot version:
 	<dependency>
 		<groupId>fr.inria.gforge.spoon</groupId>
 		<artifactId>spoon-core</artifactId>
-		<version>6.0.0-SNAPSHOT</version>
+		<version>7.0.0-SNAPSHOT</version>
```
monperrus: should be 6.1.0

doc/_config.yml
```diff
@@ -18,9 +18,9 @@ include: [_includes]
 # variables
 
 sidebar_tagline: Spoon
-sidebar_version: version 5.9.0
-spoon_release: "5.9.0"
-spoon_snapshot: "6.0.0-SNAPSHOT"
+sidebar_version: version 6.0.0
+spoon_release: "6.0.0"
+spoon_snapshot: "7.0.0-SNAPSHOT"
```
monperrus: idem

doc/doc_homepage.md
```diff
@@ -35,6 +35,7 @@ You have different options:
 Star Spoon on Github: [https://github.com/INRIA/spoon](https://github.com/INRIA/spoon). :-)
 
 <!-- .* Marker comment. -->
+- November 17, 2017: Spoon 6.0.0 is released! Check the [(changelog)](https://github.com/INRIA/spoon/releases/tag/spoon-core-6.0.0) as this major release may introduce breaking changes :warning:
```
monperrus: I would say "as this major release may introduce breaking changes :warning:" here because we actually break very little

I'd say "there are few non backward-compatible changes, see changes"

README.md
surli: I hesitated but as CtImport is almost ready, it will be immediately a new major release of Spoon... So IMHO we won't have another minor release before Spoon 7.0.0.

monperrus:
Imports are breaking but not part of the main public API.I'd propose first aim at 6.1 and we see
something really big breaking we could change to 7.0.

surli:
> Imports are breaking but not part of the main public API.

They are part of the main public API through `CompilationUnit#getImports()` which is used by clients.

monperrus:
Yes, I know, but this is not the kind of methods that are typically used by clients.
In default mode, you never manipulate the imports. In auto-import mode, you don't manipulate them as
well, Spoon does it for you.

surli:
> In auto-import mode, you don't manipulate them as well, Spoon does it for you.

Yes, but you can get them as in #1629. Then if we change the return type, we break their code. I'm ok to bump the version to 6.1.0 if we do a release before merging CtImport, else I really don't see the point to not follow the semantic versioning. 
Moreover I really hope that we'll be able to put CtModule for the next release :)

surli:
@monperrus when you merge this PR take care to delete the Spoon 7.0.0 snapshot from gforge.inria.fr.

PR #1738, opened 2017-11-22
review test: add test for getAnnotation().value() on shadow classes

monperrus:
@surli would you have a look at this one?


surli:
@monperrus check on travis logs: the test does not fail with a NPE but with an equality failure between TODO and SOURCE: which should be because it's the content of Retention annotation.

monperrus:
interesting, I had another error on my machine, which has now disappeared

PR #1740, opened 2017-11-22
review refactor: Add a new way to build a query in QueryFactory

surli:


src/main/java/spoon/reflect/factory/QueryFactory.java
```diff
@@ -48,4 +48,12 @@ public CtQuery createQuery() {
 	public CtQuery createQuery(Object input) {
 		return new CtQueryImpl(input);
 	}
+
+	/**
+	 * Creates a bound query with an optional number of inputs
```
monperrus: what is the meaning of inputs?

not fan of pointing to an implementation in an interface.

monperrus:
interesting. it seems we need a test case.

surli:
@monperrus @pvojtechovsky if you're ok with this one, it's ready for merging.

monperrus:
ok for me. @pvojtechovsky do you merge?

PR #1741, opened 2017-11-22
review refactor: add CtModel#getRootElement and deprecates CtModel#getRootPackage

surli:
The purpose of this PR is to help evolving the model without losing the clients.
I propose to : 
  1. add the method `getRootElement`, which would return a CtPackage or a CtModule depending on the java compliance version
  2. Set `CtModel#getRootPackage` as deprecated, and advise clients to use `getRootModel` or `PackageFactory#getRootPackage`. 

For now, the rootPackage is still stored on `CtModel` but I don't know if it's a good design on the long term view.

### DIFF
--- a/src/main/java/spoon/reflect/CtModel.java
+++ b/src/main/java/spoon/reflect/CtModel.java
@@ -20,6 +20,7 @@ import spoon.processing.Processor;
 import spoon.reflect.declaration.CtElement;
 import spoon.reflect.declaration.CtPackage;
 import spoon.reflect.declaration.CtType;
+import spoon.reflect.factory.PackageFactory;
 import spoon.reflect.visitor.Filter;
 
 import java.io.Serializable;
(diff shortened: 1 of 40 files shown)


src/main/java/spoon/reflect/CtModel.java
```diff
@@ -31,7 +32,20 @@
...
+	 */
+	CtElement getRootElement();
+
+	/**
+	 * Returns the root package
+	 * @deprecated If you want to traverse the model,
+	 * prefer using {@link #getRootElement()}. If you really need the rootPackage,
+	 * prefer using {@link PackageFactory#getRootPackage()}
+	 */
+	@Deprecated
```
monperrus: remove @Deprecated

monperrus:
Per our discussion, the design of CtModel in #1730 is better

PR #1746, opened 2017-11-24
review chore: Travis: Skip checkstyle when calling mvn install for installing dependencies

surli:
See: https://docs.travis-ci.com/user/languages/java/#Maven-Dependency-Management

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ env:
   - SCRIPT=travis-coverage.sh
   - SCRIPT=travis-verify.sh
 
+install: mvn install -DskipTests=true -Dmaven.javadoc.skip=true -Dcheckstyle.skip=true -B -V
 script: ./chore/travis/$SCRIPT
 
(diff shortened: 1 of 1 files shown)


monperrus:
it seems to me that this switch should be in travis-jdk9.sh and travis-jdk8.sh instead

surli:
> it seems to me that this switch should be in travis-jdk9.sh and travis-jdk8.sh instead

No, it needs to be everywhere: by default Travis calls `mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V` before executing the script to install the dependencies. 
In our case, it makes all fails when there's a checkstyle error, instead of only checkstyle script. 

After the change, every job will execute the installation of dependencies with skipping checkstyle, and only then calls its own script like checkstyle checking for the checkstyle script or testing for the jdk8/9 scripts.

monperrus:
OK, then that's OK for me.

PR #1748, opened 2017-11-24
fix(test): CheckBoundMatcher

pvojtechovsky:
I suggest to change this test, because the new template matcher (#1686), which collects matching parameter values, fails on this. Note: left and right expression are not same in the matching examples, so there should be different variables (_x_ and _y_) and not only one, like it is now.


monperrus:
I'm OK with this.

Could you add the test in 1686 showing that you match correctly `x.S() == x.S()` if the two expressions are the same?

Also, it would be great that you avoid in #1686 to modify existing test code for refactoring, so that we directly see the changed behavior (red lines in `src/test`) as opposed to new specification (green lines in `src/test`).

pvojtechovsky:
> Could you add the test in 1686 showing that you match correctly x.S() == x.S() if the two expressions are the same?

Good idea, I will do.

PR #1751, opened 2017-11-25
fix: partial evaluator

pvojtechovsky:


src/test/java/spoon/test/eval/EvalTest.java
```diff
@@ -45,6 +43,17 @@ public void testArrayLength() throws Exception {
...
 
+	@Test
+	public void testDoNotSimplify() throws Exception {
+		CtClass<?> type = build("spoon.test.eval", "ToEvaluate");
+		assertEquals("ToEvaluate", type.getSimpleName());
+
+		CtBlock<?> b = type.getMethodsByName("testDoNotSimplify").get(0).getBody();
+		assertEquals(1, b.getStatements().size());
+		b = b.partiallyEvaluate();
+		assertEquals("java.lang.System.out.println((((\"enter: \" + className) + \" - \") + methodName))", b.getStatements().get(0).toString());
```
monperrus: what was the result before the fix?

pvojtechovsky: it was this
```
java.lang.System.out.println(methodName)
```

PR #1755, opened 2017-11-28
test: substitution of `return expression.S()`

pvojtechovsky:
Improves a test of substitution of `return expression.S()`, which actually substitutes
A) `return expression.S()` if the parameter value is CtBlock
B) `expression.S()` if the parameter value is an expression

The case B) was not tested before, so I would like to add this test before #1686 so it visible that this contract is still working and did not changed.


monperrus:
do we still have an assertion for case A?

pvojtechovsky:
Yes, here https://github.com/INRIA/spoon/pull/1755/files#diff-623cb8c19b37a2dc0f4b2ac7063eacf1L21

monperrus:
thanks!

PR #1756, opened 2017-11-28
fix: test meta model - MMField value(Item)Type is not implicit

pvojtechovsky:
Value type of MMField `CtLiteral#value` was Object with implicit = true, which caused problem in new template substitution and then invalid pretty printing of such type references... DJPP doesn't prints such TypeReferences, even if they are on place where they cannot be implicit.

### DIFF
--- a/src/test/java/spoon/test/reflect/meta/MetaModelTest.java
+++ b/src/test/java/spoon/test/reflect/meta/MetaModelTest.java
@@ -36,7 +36,6 @@ public class MetaModelTest {
 	 */
 	//enable this test after everything is covered.
 	@Test
-	@Ignore
 	public void spoonMetaModelTest() {
 		SpoonMetaModel mm = new SpoonMetaModel(new File("./src/main/java"));
 		List<String> problems = new ArrayList<>();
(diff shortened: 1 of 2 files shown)


src/test/java/spoon/test/reflect/meta/MetaModelTest.java
```diff
@@ -57,14 +56,17 @@ public void spoonMetaModelTest() {
...
+				assertFalse("Item value type of Field " + mmField.toString() + " is implicit", mmField.getItemValueType().isImplicit());
 				
 				mmField.forEachUnhandledMethod(ctMethod -> problems.add("Unhandled method signature: " + ctMethod.getDeclaringType().getSimpleName() + "#" + ctMethod.getSignature()));
 			});
 		});
 		
 		unhandledRoles.forEach(it -> problems.add("Unused CtRole." + it.name()));
 		
-		assertTrue(String.join("\n", problems), problems.isEmpty());
+//		assertTrue(String.join("\n", problems), problems.isEmpty());
```
monperrus: explain why this assertion is commented out and when we'll be able to get it back?

src/test/java/spoon/test/reflect/meta/MetaModelTest.java
```diff
@@ -36,7 +36,6 @@
 	 */
 	//enable this test after everything is covered.
 	@Test
-	@Ignore
```
monperrus: remove line comment " 	//enable this test after everything is covered. "?

monperrus:
thanks.

PR #1759, opened 2017-11-29
fix: CtQueryImpl handling of preallocated exceptions

pvojtechovsky:
The new Templating engine (#1686) uses CtQuery a lot and Java runtime decided to optimize exception handling in runtime. It was the reason why many tests started to fail with strange exception without the stacktrace.

This PR handles such "preallocated exceptions" well.

See java [release notes](http://www.oracle.com/technetwork/java/javase/relnotes-139183.html#vm) for details:
> The compiler in the server VM now provides correct stack backtraces for all "cold" built-in exceptions.
For performance purposes, when such an exception is thrown a few times, the method may be recompiled. After recompilation, the compiler may choose a faster tactic using preallocated exceptions that do not provide a stack trace. To disable completely the use of preallocated exceptions, use this new flag: -XX:-OmitStackTraceInFastThrow.


monperrus:
super interesting!

Issue #1757, opened 2017-11-29
isSubtypeOf is not working with class hierarchies


mohammadnaseri:
Hi, 
I am using spoon to analyze an Android project. I am creating the launcher like the following:
```
Launcher launcher = new Launcher();
launcher.addInputResource("path to my Android project");
launcher.getEnvironment().setSourceClasspath(new String[]{"path to Android jar file");
launcher.getEnvironment().setNoClasspath(true);
CtModel model = launcher.buildModel();
```
In my Android project, I have a class like following:
```
public class MyEditText extends AppCompatEditText {
}
```
```AppCompandEditText``` is an Android class which extends ```EditText```
Now, I want to check if ```MyEditText``` is a subType of ```EditText```. 
I am doing like following:
```
CtTypeReference androidEditRef = launcher.getFactory().Type().createReference("android.widget.EditText");

CtTypeReference nodeRef = App.launcher.getFactory().Type().createReference("MyEditTextPackagePath");

nodeRef.isSubtypeOf(androidEditRef)
```
Now this one returns false, I feel like ```isSubType``` is not checking the class hierarchy. 
Please let me know if I am doing something wrong here.

pvojtechovsky:
Hi,

I guess that `android.widget.EditText` is not on spoon class path. 
Try `androidEditRef.getTypeDeclaration()`, what it returns? Null or an exception or instance of CtType?
Try `nodeRef.getTypeDeclaration()`, what it returns? Null or an exception or instance of CtType?
Try `nodeRef.getTypeDeclaration().getSuperclass()' and then `nodeRef.getTypeDeclaration().getSuperclass().getSuperclass()' ...

mohammadnaseri:
```android.widget.EditText``` is in android class path. and I added the jar path of android to the launcher class path. 
```androidEditRef.getTypeDeclaration()``` returns instance of CtType.
```nodeRef.getTypeDeclaration()``` returns instance of CtType too.
```nodeRef.getTypeDeclaration().getSuperclass()``` returns ```AppCompatEditText```
But 
```nodeRef.getTypeDeclaration().getSuperclass().getSuperclass()``` returns null.

surli:
Hi @mohammadnaseri 

please have a look on my first step in the proposed PR #1760 : I tried to reproduce your bug, however in my unit test it seems to pass without problem. Maybe I forget something compared to your own conditions ?
[...]

PR #1760, opened 2017-11-30
review test: check inheritance behaviour when using noclasspath with a sourceclasspath

surli:


surli:
Ok this PR shows that I don't manage to reproduce the error in #1757 (the first test does not fail).

However I really wonder why it fails when specifying the full path of a class to put in source classpath?
see: https://github.com/INRIA/spoon/pull/1760/files#diff-1ab3a44b929ed1b3f45ae2718c5bc427R190
 @tdurieux @pvojtechovsky @monperrus any idea on that? Is it expected?

src/test/java/spoon/test/api/NoClasspathTest.java
```diff
@@ -150,4 +153,57 @@ public void testIssue1747() {
...
+
+		Launcher spoon = new Launcher();
+		spoon.getEnvironment().setShouldCompile(true);
+		spoon.addInputResource(sourceInputDirPath);
+		spoon.setBinaryOutputDirectory(targetBinPath);
+		spoon.run();
+
+		spoon = new Launcher();
+		spoon.getEnvironment().setNoClasspath(true);
+		spoon.getEnvironment().setSourceClasspath(new String[] { targetBinPath+"/fr/acme/MyClass.class" });
```
surli: If I put the directory here, it works. Why?

PR #1761, opened 2017-11-30
review doc: Add more precision for CtTypeReference#getActualClass

surli:


src/main/java/spoon/reflect/reference/CtTypeReference.java
```diff
@@ -56,8 +56,8 @@
 	 * Use {@link #getTypeDeclaration()} instead,
 	 * in order to only stay in the Spoon world and manipulate CtType instead of java.lang.Class
 	 *
-	 * @return the Java class or null if the class is not found (not in
-	 * classpath)
+	 * @return the Java class or throws an exception if the class is not found.
```
monperrus: a SpoonClassNotFoundException

PR #1762, opened 2017-11-30
review fix: a NPE in CtTypeImpl#isSameParameter in noclasspath

surli:


src/main/java/spoon/support/reflect/declaration/CtTypeImpl.java
```diff
@@ -765,18 +765,18 @@ private boolean isSameParameter(CtMethod<?> method, CtTypeReference<?> ctParamet
 				return false;
 			}
 		} else if (expectedType instanceof CtTypeParameterReference) {
-			if (!ctParameterType.isSubtypeOf(factory.Type().createReference(expectedType.getActualClass()))) {
+			if (!ctParameterType.isSubtypeOf(factory.Type().createReference(expectedType.getQualifiedName()))) {
```
msteinbeck: A more stable approach could be the check if `getDeclaration` returns null and only if it is not null using `getQualifiedName()` to create the reference. Otherwise, skip this check.

monperrus: getActualClass should be avoided as much as possible to have portable code between full-classpath and noclasspath mode.
So this change seems good to me.

monperrus:
OK for me.

monperrus:
@pvojtechovsky if you're OK, do you merge?

src/main/java/spoon/support/reflect/declaration/CtTypeImpl.java
```diff
@@ -759,24 +759,30 @@ private boolean isSameParameter(CtMethod<?> method, CtTypeReference<?> ctParamet
 				return false;
 			}
 		}
+		if (ctParameterType.equals(factory.Type().OBJECT) && (expectedType instanceof CtTypeParameterReference)) {
+			CtTypeParameterReference typeParameterReference = (CtTypeParameterReference) expectedType;
+			if (!typeParameterReference.isUpper() || (typeParameterReference.isUpper() && typeParameterReference.getBoundingType().equals(ctParameterType))) {
```
pvojtechovsky: in this `if` statement this is true `ctParameterType.equals(factory.Type().OBJECT)`, so call of `typeParameterReference.getBoundingType().equals(ctParameterType))` might be changed to
`typeParameterReference.getBoundingType().equals(factory.Type().OBJECT))` ... but may be it points to mistake on the line 762 ... and the condition `ctParameterType.equals(factory.Type().OBJECT)` should be changed?

src/test/resources/noclasspath/elasticsearch1753/TaskManager.java
```diff
@@ -0,0 +1,463 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.Consumer;
+
+import static org.elasticsearch.common.unit.TimeValue.timeValueMillis;
+
+/**
+ * Task Manager service for keeping track of currently running tasks on the nodes
+ */
+public class TaskManager extends AbstractComponent implements ClusterStateListener {
```
pvojtechovsky: Could you simplify this class, so it contains only type members, which are needed to reproduce the problem? Do it please only if it is low effort - it would make tests more readable. Otherwise it is OK like it is.

src/main/java/spoon/support/reflect/declaration/CtTypeImpl.java
```diff
@@ -759,24 +759,30 @@ private boolean isSameParameter(CtMethod<?> method, CtTypeReference<?> ctParamet
...
+			}
+		}
 		if (expectedType instanceof CtTypeParameterReference && ctParameterType instanceof CtTypeParameterReference) {
 			// Check if Object or extended.
 			if (!ctParameterType.equals(expectedType)) {
 				return false;
 			}
 		} else if (expectedType instanceof CtTypeParameterReference) {
-			if (!ctParameterType.isSubtypeOf(factory.Type().createReference(expectedType.getActualClass()))) {
+			if (!ctParameterType.isSubtypeOf(factory.Type().createTypeParameterReference(expectedType.getQualifiedName()))) {
```
pvojtechovsky: I do not understand the purpose of both - old and new code. Could anybody explain it?

PR #1766, opened 2017-11-30
refactor: setSourceClasspath throws an exception in case a .class is given

surli:


msteinbeck:
Why throwing an exception if a '.class' file is present? Why don't you just ignore it?

monperrus:
Because users would think it is taken into account and would not understand the behavior afterwards.

PR #1767, opened 2017-11-30
alternative to #1762: fix: a NPE in CtTypeImpl#isSameParameter in noclasspath

pvojtechovsky:


pvojtechovsky:
@surli seems like tests passed too. This solution seems to me more clean then #1767. @monperrus please merge it if Simon agrees and you are OK with that too.

surli:
It looks good to me :)

monperrus:
Thanks a lot @surli and @pvojtechovsky

PR #1770, opened 2017-12-04
review: refactor: manage sourceOutputDirectory in Environment

surli:
This PR intends to fix issues bring by the multiple usage of SourceOutputDirectory


src/main/java/spoon/support/compiler/jdt/JDTBasedSpoonCompiler.java
```diff
@@ -279,12 +278,12 @@ public void addTemplateSources(List<SpoonResource> resources) {
 
 	@Override
 	public void setSourceOutputDirectory(File outputDirectory) {
```
tdurieux: put deprecated?

tdurieux:
:heart:

src/main/java/spoon/support/compiler/jdt/JDTBasedSpoonCompiler.java
surli: I wondered: it's still usable to put the right value in the environment...

monperrus:
OK for me. Merge?

src/main/java/spoon/support/compiler/jdt/JDTBasedSpoonCompiler.java
tdurieux: What do you mean?

surli:
It's ready for me.

src/main/java/spoon/support/compiler/jdt/JDTBasedSpoonCompiler.java
surli: The method is used in some test and maybe some clients are still using it to put the value, so it can still work except now it puts the value in the Environment.

tdurieux: This PR will already break the API.
I would break it more to reduce the number of API, to simplify the usage of the API.
At least I would say in the doc that this API is just a different way to access the Environment

surli: >I would break it more to reduce the number of API, to simplify the usage of the API.
At least I would say in the doc that this API is just a different way to access the Environment

I agree with you. I'll deprecate it then.

surli:
@tdurieux @monperrus I made few changes to deprecate methods `setOutputDirectory` in `SpoonModelBuilder` and `FileGenerator` in favor of `Environment#setOutputSourceDirectory`. 
However I think the API of `JavaOutputProcessor` is not really clear: it's a processor, except we never set its factory in Spoon, so here the `Environment` comes from the prettyprinter I pass in constructor, but it may came from the factory.

So maybe we should override the `setFactory()` to do nothing in `JavaOutputProcessor` and to only use the environment from the pretty printer to avoid inconsistency if two different environments are used in Factory and Prettyprinter.

src/main/java/spoon/support/JavaOutputProcessor.java
```diff
@@ -96,7 +122,7 @@ public void init() {
 			}
 		}
 		try {
-			directory = directory.getCanonicalFile();
+			this.getEnvironment().setSourceOutputDirectory(directory.getCanonicalFile());
```
surli: I'm not sure about this one: this should be done directly in the setter of Environment, no? WDYT?

src/main/java/spoon/support/compiler/jdt/JDTBasedSpoonCompiler.java
surli: I put the deprecated annotation in the interface.

src/main/java/spoon/support/JavaOutputProcessor.java
```diff
@@ -44,19 +46,28 @@
...
 public class JavaOutputProcessor extends AbstractProcessor<CtNamedElement> implements FileGenerator<CtNamedElement> {
 	PrettyPrinter printer;
-
-	File directory;
+	private Environment environment;
 
 	List<File> printedFiles = new ArrayList<>();
 
+	public JavaOutputProcessor(PrettyPrinter printer) {
+		this.printer = printer;
```
tdurieux: that does not seem nice :/

surli: You're talking about the line below?

tdurieux: No the constructor

surli: Could you elaborate ?

monperrus:
OK for me. @tdurieux do you see a problem?

tdurieux:
I have a general issue with the printing process, there is too much state.

src/main/java/spoon/support/JavaOutputProcessor.java
tdurieux: why did you need to copy the env?
it is accessible via getFactory().getEnv().

tdurieux:
I don't like the creation of getEnvironment method in several classes, I am pretty sure it is not required

monperrus:
> I don't like the creation of getEnvironment method in several classes

I agree, if we can avoid it, that's better

src/main/java/spoon/support/JavaOutputProcessor.java
```diff
@@ -44,19 +46,28 @@
  */
 public class JavaOutputProcessor extends AbstractProcessor<CtNamedElement> implements FileGenerator<CtNamedElement> {
 	PrettyPrinter printer;
-
-	File directory;
+	private Environment environment;
```
monperrus: we don't need some state here, we alread have `getFactory().getEnvironment()`

src/main/java/spoon/support/JavaOutputProcessor.java
```diff
@@ -44,19 +46,28 @@
...
 
 	List<File> printedFiles = new ArrayList<>();
 
+	public JavaOutputProcessor(PrettyPrinter printer) {
+		this.printer = printer;
+		this.environment = printer.getEnvironment();
```
monperrus: we can replace `printer.getEnvironment()` by `getFactory().getEnvironment()`. And then we don't need the new method in interface PrettyPrinter

tdurieux:
thanks!

surli:
@tdurieux @monperrus so it's ready for me.

monperrus:
thanks a lot!

monperrus:
seems we have a regression due to this one: https://ci.inria.fr/sos/job/juliac/423/consoleFull

seintur:
I was using https://github.com/INRIA/spoon/blob/a3a96502f07984e861e2da4efbf604b6ec432374/src/main/java/spoon/support/JavaOutputProcessor.java#L64 that is now deprecated. I updated my project to remove the use of this now deprecated method. The problem is fixed.

monperrus:
thanks Lionel. --Martin

PR #1771, opened 2017-12-04
WiP fix(module): Spoon should be able to use module in full classpath

### LINKED ISSUE #1769: bug: Project with modules cannot be spooned in full classpath mode yet
We have to use some arguments with JDT in order to make it work with full classpath mode.

surli:
This PR intends to fix #1769


surli:
This PR will be a WiP for a while, because we have a bug with JDT. I opened a new bug in their tracker, we'll see what happens next: https://bugs.eclipse.org/bugs/show_bug.cgi?id=528109

Moreover, according to this specification: http://openjdk.java.net/jeps/261 we need to use a new argument when compiling some code with modules, called "--module-source-path". JDT is implemented this way: then I had to add this argument in the environment.

surli:
So, according to the release date of JDT (see: https://wiki.eclipse.org/Simultaneous_Release) I should be able to continue this one at the end of the month...

surli:
This PR will be break in several parts as I'm trying to manage module in Spoon piece by piece.

PR #1772, opened 2017-12-04
WiP feat: implement #1267 to improve API for imports

### LINKED ISSUE #1267: Feature request: Allow to flag in the model an import as static import
I'd be happy if the API would provide something like

	getFactory().Class().addStaticImport("fqdn.to.static.method");

# Use case
I'd like to generate unit tests by using Hamcrester matcher. I don't want to use either the AST to accomplish this but also don't want to use the full FQDN. A real usage would be

	addStatement(new CodeSnippert("assertThat(" + variableName + ", is(1))"));

where I would have previously statically imported *assertThat* by using

	getFactory().Class().addStaticImport("org.hamcrest.MatcherAssert.assertThat");

surli:
This PR intends to fix #1267

### DIFF
--- a/src/main/java/spoon/reflect/visitor/ImportScanner.java
+++ b/src/main/java/spoon/reflect/visitor/ImportScanner.java
@@ -49,8 +49,6 @@ public interface ImportScanner {
 	 */
 	List<CtImport> getAllImports();
 
-	void setOriginalImports(List<CtImport> imports);
-
 	void setImports(List<CtImport> importList);
 
 	void addImport(CtImport ctImport);
--- a/src/main/java/spoon/reflect/visitor/ImportScanner.java
+++ b/src/main/java/spoon/reflect/visitor/ImportScanner.java
@@ -19,6 +19,7 @@ package spoon.reflect.visitor;
 import spoon.reflect.cu.CompilationUnit;
 import spoon.reflect.declaration.CtElement;
 import spoon.reflect.declaration.CtImport;
+import spoon.reflect.factory.Factory;
 import spoon.reflect.reference.CtReference;
 
 import java.util.Collection;
@@ -38,6 +39,8 @@ public interface ImportScanner {
 	 * Computes import of a {@link CompilationUnit}
 	 */
 	void computeImports(CompilationUnit cu);
+
+	void setFactory(Factory factory);
 
 	/**
 	 * Use computeImports or computeAllImports before getting the different imports.
--- a/src/main/java/spoon/reflect/visitor/ImportScanner.java
+++ b/src/main/java/spoon/reflect/visitor/ImportScanner.java
@@ -16,11 +16,13 @@
  */
 package spoon.reflect.visitor;
 
+import spoon.reflect.cu.CompilationUnit;
 import spoon.reflect.declaration.CtElement;
 import spoon.reflect.declaration.CtImport;
 import spoon.reflect.reference.CtReference;
 
 import java.util.Collection;
+import java.util.List;
 
 /**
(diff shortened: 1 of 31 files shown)


surli:
So this PR was meant to be a simple one, with the main goal to add helpers in CompilationUnit to add and remove imports.
The client contract would be that given a class with a call to `Assert.assertNull()` (and so an import to `org.junit.Assert`), he should be able to: 
1. remove the import to `org.junit.Assert`
2. add a new static import to `org.junit.Assert.assertNull`

However, even with this simple use case I have a problem in the way itself we manage imports right now. We currently have to ways to process imports: 
 - when parsing the file, with `JDTImportBuilder`
 - when printing the file with `ImportScanner`

The problem is that, even if I add some methods to add/remove imports in JDTImportBuilder, they will be computed back in `ImportScanner` just before printing... 

I have the feeling that in order to do that, the ImportScanner should not be available only for the printer, but more globally for the whole model. Then manipulating the imports should be done through it, in order to ensure the consistency of the operations.

pvojtechovsky:
> the ImportScanner should not be available only for the printer, but more globally for the whole model. Then manipulating the imports should be done through it, in order to ensure the consistency of the operations.

I like this idea.

src/main/java/spoon/reflect/visitor/ImportScanner.java
```diff
@@ -16,31 +16,71 @@
...
+	void removeImport(CtImport ctImport);
 
 	/**
-	 * Checks if the type is already imported.
+	 * Returns true iff the given reference belongs to the list of imports.
+	 */
+	boolean isEffectivelyImported(CtReference reference);
+
+	/**
+	 * Returns true iff the given reference has been imported {@link #isEffectivelyImported(CtReference)}
```
seintur: iff -> if

src/main/java/spoon/reflect/visitor/ImportScanner.java
```diff
@@ -16,31 +16,71 @@
...
+	 * OR if the given reference is considered as implicitely imported (e.g. a reference from java.lang or from
+	 * the current target type)
 	 */
 	boolean isImported(CtReference ref);
+
+	/**
+	 * Checks if the fully qualified name should be printed
+	 * This can return false even if {@link #isImported(CtReference)} return true:
```
seintur: "return true" -> "returns true"

src/test/java/spoon/test/imports/ImportTest.java
```diff
@@ -406,7 +416,7 @@ public void testImportStaticAndFieldAccess() throws Exception {
 
 	@Test
 	public void testImportStaticAndFieldAccessWithImport() throws Exception {
-		// contract: Qualified field access and an import static with import should import the type first, and not use static import
+		// contract: The same static import should be used, only one import should be called an the type should not be printed
```
seintur: "an the type" -> "and the type"

src/test/java/spoon/test/imports/ImportTest.java
```diff
@@ -1302,4 +1320,27 @@ public void testGetImportKindReturnRightValue() {
 		ctImport = spoon.getFactory().createImport(aType.getPackage().getReference());
 		assertEquals(CtImportKind.ALL_TYPES, ctImport.getImportKind());
 	}
+
+	@Test
+	public void testRemoveAndAddImports() {
+		// contract: the API authorize to add and remove imports
```
seintur: authorize -> authorizes

surli:
I close this one for now. I might take it back later but not under this form.

PR #1774, opened 2017-12-05
review fix(annotation): bug when using an annotation targeting declaration and types

### LINKED ISSUE #1773: Bug: Annotation @javax.validation.constraints.NotNull not correctly parsed
Maven dependencies:
```
<dependency>
    <groupId>javax.validation</groupId>
    <artifactId>validation-api</artifactId>
   <version>2.0.0.Final</version>
</dependency>
 <dependency>
    <version>6.0.0</version>
</dependency>
```

It seems like the field parsing is not working with this specific annotation (others seems to be working fine!)
Here is the field: 
@NotNull
public String mandatoryField;

Here is the toString() of the spoon model constructed:
```
public java.lang.@javax.validation.constraints.NotNull
String mandatoryField;
```

It should be this according t me:
```
@javax.validation.constraints.NotNull
public java.lang.String mandatoryField;
```

I don't understand how this is possible. This is working fine for example with annotation @javax.xml.bind.annotation.XmlTransient

surli:
Fix #1773 

In this case the JLS says that: "It is possible for @foo to be both a declaration annotation and a type annotation simultaneously." So we should attach the annotation to the field/method AND to the type in Spoon model. 
Then in this case we don't want to print it twice: we print it only for the field/method.


surli:
Actually this bug seems not to be one, and my unit test is wrong. See my answer on #1773

surli:
This PR is now ready to be reviewed @pvojtechovsky @monperrus 

Please note that the PR is potentially breaking.

When using a code like:
```java
@MyAnnotation
String myField;

@Target({ElementType.TYPE_USE, ElementType.FIELD})
public @interface MyAnnotation {}
```

the model will now contain `MyAnnotation` annotation on the field AND on the type reference String. 
This has been done according to the following sentence of the JLS:
> @Foo int f;
> @Foo is a declaration annotation on f if Foo is meta-annotated by @Target(ElementType.FIELD), and a type annotation on int if Foo is meta-annotated by @Target(ElementType.TYPE_USE). It is possible for @Foo to be both a declaration annotation and a type annotation simultaneously. 

So this is not true only for fields, but for all declaration. 
Moreover, in case of the usage of a an annotation without `@Target` the JLS says that 
>  If an annotation of type java.lang.annotation.Target is not present on the declaration of an annotation type T, then T is applicable in all declaration contexts except type parameter declarations, and in no type contexts. 

So I considered those exceptions for adding the annotation or not. 
It means that in no classpath, if we have no information about the annotation type, `MyAnnotation` would be only attached to the field on the example below: this is the previous Spoon behaviour.

src/main/java/spoon/reflect/visitor/ElementPrinterHelper.java
```diff
@@ -77,6 +78,10 @@ public ElementPrinterHelper(TokenWriter printerTokenWriter, DefaultJavaPrettyPri
 	 */
 	public void writeAnnotations(CtElement element) {
 		for (CtAnnotation<?> annotation : element.getAnnotations()) {
+			if (element.isParentInitialized() && element instanceof CtTypeReference && (element.getParent() instanceof CtTypedElement) && element.getParent().getAnnotations().contains(annotation)) {
```
monperrus: explain this complex condition with a fat comment?

src/main/java/spoon/reflect/declaration/CtAnnotation.java
```diff
@@ -165,4 +167,42 @@
 	@Override
 	@UnsettableProperty
 	<C extends CtExpression<A>> C setTypeCasts(List<CtTypeReference<?>> types);
+
+	static CtAnnotatedElementType getAnnotatedElementTypeForCtElement(CtElement element) {
```
monperrus: use an anonymous scanner to avoid multiple instanceof?

src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderQuery.java
```diff
@@ -169,22 +169,37 @@ static boolean hasAnnotationWithType(Annotation a, CtAnnotatedElementType elemen
...
+				for (Object field : fields) {
+					if (field instanceof FieldBinding && elementType.name().equals(CharOperation.charToString(((FieldBinding) field).name))) {
+						return true;
+					}
 				}
 			}
 		}
-		return false;
+
+		// true here means that the target annotation is not mandatory and we don't have found it
```
seintur: don't have -> have not

src/main/java/spoon/support/compiler/jdt/ParentExiter.java
```diff
@@ -133,29 +133,43 @@ public void setChild(ASTNode child) {
 	@Override
 	public void scanCtElement(CtElement e) {
 		if (child instanceof CtAnnotation && this.jdtTreeBuilder.getContextBuilder().annotationValueName.isEmpty()) {
-			e.addAnnotation((CtAnnotation<?>) child);
+			// we check if the current element can have the annotation attached
+			CtAnnotatedElementType annotatedElementType = CtAnnotation.getAnnotatedElementTypeForCtElement(e);
+			annotatedElementType = (e instanceof CtTypeParameter || e instanceof CtTypeParameterReference) ? CtAnnotatedElementType.TYPE_USE : annotatedElementType;
+
+			// in case of noclasspath, we can be 100% sure, so we guess it must be attached...
```
seintur: "can be sure" or "cannot be sure"?

src/main/java/spoon/reflect/declaration/CtAnnotation.java
surli: I don't see how to use easily an anonymous scanner here: I must return a variable and I cannot store it as a local variable if I use the anonymous scanner. And I don't have any method to call to setup the value.

monperrus: I agree, it's not straightforward.

> I must return a variable and I cannot store it as a local variable if I use the anonymous scanner
 
Unless you create an inline class

```
class Result { CtVariable v }
final Result result = new Result();
// then the anonymous class

```

monperrus:
merge?

surli:
> merge?

ok for me

PR #1776, opened 2017-12-06
review feat: Add a method to get the unit type of the compilation unit

surli:


src/test/java/spoon/test/compilationunit/TestEncoding.java
```diff
@@ -1,33 +0,0 @@
...
-import java.io.File;
-import java.nio.charset.Charset;
-import java.nio.file.Files;
-
-import static org.junit.Assert.assertEquals;
-
-/**
- * Created by urli on 18/08/2017.
- */
-public class TestEncoding {
```
surli: This one is not deleted but renamed (see above)

PR #1777, opened 2017-12-06
feature: Factory#createQuery(Iterable)

pvojtechovsky:
Possibility to create query on Collection of elements. No need to create helper array, just to convert collection to format acceptable by API.

### DIFF
--- a/src/test/java/spoon/test/filters/FilterTest.java
+++ b/src/test/java/spoon/test/filters/FilterTest.java
@@ -865,16 +865,16 @@ public class FilterTest {
 		
 		//contract: empty query returns no element
 		assertEquals(0, launcher.getFactory().createQuery().list().size());
-		assertEquals(0, launcher.getFactory().createQuery(null).list().size());
+		assertEquals(0, launcher.getFactory().createQuery((Object) null).list().size());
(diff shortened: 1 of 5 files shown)


src/test/java/spoon/test/filters/FilterTest.java
```diff
@@ -865,16 +872,16 @@ public void testEmptyQuery() throws Exception {
 		
 		//contract: empty query returns no element
 		assertEquals(0, launcher.getFactory().createQuery().list().size());
-		assertEquals(0, launcher.getFactory().createQuery(null).list().size());
+		assertEquals(0, launcher.getFactory().createQuery((Object) null).list().size());
```
monperrus: Adding a cast to null is not elegant IMHO. One alternative solution would be to replace  `createQuery(null)` by  `createQuery()`

pvojtechovsky: `createQuery()` is already used on the line 867, the purpose of this test is to check handling of null values. I guess that cast is OK for the test. Or optionally I can use new API and call `createQuery(Arrays.asList(null))`. WDYT?

monperrus:
you're right. Thanks.

PR #1781, opened 2017-12-11
review refactor: CompilationUnit should return unmodifiable collection 

surli:


monperrus:
I'm OK.

but we need a test for the new method.

surli:
@monperrus Done. for me it's ready.

PR #1782, opened 2017-12-11
review fix: CompilationUnits are created on demand

surli:


surli:
@pvojtechovsky could you have a look on this one? 
I had a problem in #1772 because I compute imports from compilation unit and when generated RoleHandlers, the generated type is not part of a CompilationUnit, so I was not able to create its imports. 

I'm really not sure if it's a proper fix, or if we must do something more generic for Spoon: I have the feeling that the compilation unit were never really used except for printing...

pvojtechovsky:
I would prefer that CompilationUnit with no exact source position would be created automatically if needed anywhere. 

I do not think that client's who just want to print each generated Type into own file, should write any code to create CompilationUnit.

src/main/java/spoon/template/Substitution.java
```diff
@@ -114,9 +116,20 @@ private Substitution() {
...
+		CompilationUnit cu = f.createCompilationUnit();
+		cu.setDeclaredPackage(targetPackage);
+		cu.setDeclaredTypes(generated);
+
 		for (CtType<?> ctType : generated) {
 			targetPackage.addType(ctType);
+			if (ctType.getPosition() != SourcePosition.NOPOSITION) {
+				SourcePosition olderPosition = ctType.getPosition();
+				SourcePosition sourcePosition = f.createSourcePosition(cu, olderPosition.getSourceStart(), olderPosition.getSourceEnd(), null);
+				ctType.setPosition(sourcePosition);
```
pvojtechovsky: I think that source positions of generated type are not related to source position of template file. The generated file should have "undefined" source start/end positions. Or what is the use case behind the copying of source positions. I afraid that such invalid source positions might only confuse the pretty printer, if it is in mode "keep lines".

And if any code needs a CompilationUnit, then that code should create it. I think it is not the responsibility of type generation code.

surli: > I think that source positions of generated type are not related to source position of template file. 

I agree, I kept this because it was the previous behaviour: the position was just a clone of the sourceposition of the template. 

> The generated file should have "undefined" source start/end positions. Or what is the use case behind the copying of source positions.

I don't know if we have some rules in Spoon for such case: @monperrus @tdurieux any idea?

surli:
> I do not think that client's who just want to print each generated Type into own file, should write any code to create CompilationUnit.

I agree with you and I think it raises a more important question for Spoon: what about newly created type out of the box from a factory? We currently do not create a CU: they are only used when pretty printing the type. However it might no longer be true if we manage imports in compilation unit. 

A way to do it might be to add a new method `getCompilationUnit` on CtType which would get the sourcePosition#compilationUnit if it exists, or which would create the compilation unit if it does not? WDYT?

src/main/java/spoon/template/Substitution.java
tdurieux: For the moment all the new elements have a SourcePosition.NOPOSITION position

surli: > For the moment all the new elements have a SourcePosition.NOPOSITION position

Yes but if we want to put a SourcePosition with a compilation unit, but without specifying the positions, is there a recommended practice?

tdurieux: My first thought is to move out the compilation unit from the position.
But I don't really like the idea, I don't see a good practice.

surli: Using "-1" for positions?...

tdurieux: It was like that before and it is not a good specification

surli: So what about a `PartialSourcePosition`?

pvojtechovsky:
> A way to do it might be to add a new method getCompilationUnit on CtType which would get the sourcePosition#compilationUnit if it exists, or which would create the compilation unit if it does not?

I agree

tdurieux:
I just thought to introduce a new interface for all the elements that are printed:

* CtType
* CtPackage
* CtModule

which contains the compilationUnit and the imports.

WDYT?

tdurieux:
I just thought that all CtType are not all top-level thus that does not really make sense

monperrus:
OK for me. @pvojtechovsky if you're OK, do you merge?

src/main/java/spoon/reflect/factory/CompilationUnitFactory.java
```diff
@@ -54,6 +60,89 @@ public CompilationUnit create() {
...
+			baseDir = this.getBaseDirectory();
+		} else {
+			baseDir = this.getModuleDirectory(module);
+		}
+
+		return new File(baseDir, ctPackage.getQualifiedName().replace(CtPackage.PACKAGE_SEPARATOR_CHAR, File.separatorChar));
+	}
+
+	public CompilationUnit getOrCreate(CtPackage ctPackage) {
+		if (ctPackage.getPosition().getCompilationUnit() != null) {
```
pvojtechovsky: ctPackage.getPosition() may be null, we should check for null here too

src/main/java/spoon/reflect/factory/CompilationUnitFactory.java
```diff
@@ -54,6 +60,89 @@ public CompilationUnit create() {
...
+				throw new SpoonException("Cannot get path for file: " + file.getAbsolutePath(), e);
+			}
+		}
+	}
+
+	public CompilationUnit getOrCreate(CtType type) {
+		if (type == null) {
+			return null;
+		}
+		if (type.getPosition().getCompilationUnit() != null) {
```
pvojtechovsky: type.getPosition() may be null, we should check for null here too

src/main/java/spoon/reflect/factory/CompilationUnitFactory.java
```diff
@@ -54,6 +60,89 @@ public CompilationUnit create() {
...
+			} catch (IOException e) {
+				throw new SpoonException("Cannot get path for file: " + file.getAbsolutePath(), e);
+			}
+		} else {
+			return getOrCreate(type.getTopLevelType());
+		}
+	}
+
+	public CompilationUnit getOrCreate(CtModule module) {
+		if (module.getPosition().getCompilationUnit() != null) {
```
pvojtechovsky: module.getPosition() may be null, we should check for null here too

src/test/java/spoon/test/compilationunit/TestCompilationUnit.java
```diff
@@ -127,4 +133,24 @@ public void testAddDeclaredTypeInCU() throws IOException {
...
+        // contract: when a type is created, a CU can be created and added as partial position
+        final Launcher launcher = new Launcher();
+        assertTrue(launcher.getFactory().CompilationUnit().getMap().isEmpty());
+
+        CtClass myNewClass = launcher.getFactory().createClass("my.new.MyClass");
+        assertEquals(SourcePosition.NOPOSITION, myNewClass.getPosition());
+
+        CompilationUnit cu = launcher.getFactory().CompilationUnit().getOrCreate(myNewClass);
+
+        assertNotNull(cu);
```
pvojtechovsky: //next call of getOrCreate returns existing cu
assertSame(cu, launcher.getFactory().CompilationUnit().getOrCreate(myNewClass))

src/test/java/spoon/test/compilationunit/TestCompilationUnit.java
```diff
@@ -127,4 +133,24 @@ public void testAddDeclaredTypeInCU() throws IOException {
...
+        SourcePosition sourcePosition = myNewClass.getPosition();
+        assertTrue(sourcePosition instanceof PartialSourcePositionImpl);
+        assertEquals(cu, sourcePosition.getCompilationUnit());
```
pvojtechovsky: replace assertEquals -> assertSame

pvojtechovsky:
I have found only little problems, and some improvements in tests. Otherwise this code is OK for me. You can merge it tomorrow after Simon decides whether he will fix it or keep it as it is. Or I will merge it tomorrow evening.

src/main/java/spoon/reflect/factory/CompilationUnitFactory.java
surli: OK I'll change that.

src/test/java/spoon/test/compilationunit/TestCompilationUnit.java
surli: I'll add it too.

src/test/java/spoon/test/compilationunit/TestCompilationUnit.java
surli: I'm not quite sure about changing this one: in the current implementation it's true, but I don't think it should be a contract. We can imagine implementation in which it can return a clone of CompilationUnit for example.

pvojtechovsky: > return a clone

sourcePosition is an unmodifiable element. There are no setters on it. It should always return the same values. I think it is the contract.

surli: > sourcePosition is an unmodifiable element.

Yes but you can change the SourcePosition object itself... But ok I'm not sure on this one: I'll change it if you think it should be a contract.

surli:
@pvojtechovsky for me it's ready then.

pvojtechovsky:
Simon, thank You!

PR #1784, opened 2017-12-12
review refactor(api): OutputType available in Environment

surli:
This PR aims at simplifying the usage of API for defining output type: until now the only way to specify the output type was using the argument "output-type" or using its own call of `JDTSpoonCompiler#generateProcessedSourceFiles`. Now it can be done simply by setting `Environment#setOutputType`


monperrus:
thanks!

PR #1794, opened 2018-01-01
feature: RoleHandlerHelper#getRoleHandlers(Class), forEachRoleHandler, getParentRoleHandler

pvojtechovsky:
Some helper methods which allows generically access Spoon model

### DIFF
--- a/src/main/java/spoon/reflect/meta/impl/RoleHandlerHelper.java
+++ b/src/main/java/spoon/reflect/meta/impl/RoleHandlerHelper.java
@@ -53,7 +53,7 @@ public class RoleHandlerHelper {
 	 * @param targetClass the class of the to be manipulated node
 	 * @param role defines the to be manipulated attribute
 	 * @return {@link RoleHandler} implementation which knows how to manipulate the attribute of {@link CtRole} on `targetClass`
-	 * or throws exception if such role does not exists on the `targetClass`
+	 * or throws exception if such role doesn't exist on the `targetClass`
 	 */
 	public static RoleHandler getRoleHandler(Class<? extends CtElement> targetClass, CtRole role) {
 		RoleHandler rh = getOptionalRoleHandler(targetClass, role);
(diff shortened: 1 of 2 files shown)


src/main/java/spoon/reflect/meta/impl/RoleHandlerHelper.java
```diff
@@ -71,4 +78,47 @@ public static RoleHandler getOptionalRoleHandler(Class<? extends CtElement> targ
...
+	 */
+	public static void forEachRoleHandler(Consumer<RoleHandler> consumer) {
+		for (List<RoleHandler> list : roleHandlers) {
+			for (RoleHandler roleHandler : list) {
+				consumer.accept(roleHandler);
+			}
+		}
+	}
+
+	public static RoleHandler getParentRoleHandler(CtElement element) {
```
monperrus: rename to getRoleHandler?

add API documentation?

thanks!

pvojtechovsky: > rename to getRoleHandler?

are you sure? I think that there should be something about `parent` in the method name, because it returns role of this element in parent.

> api doc.
I forgot that ... It compiles without documentation well :-) I will add it

monperrus: > because it returns role of this element in parent.

a role is always with respect to the parent, right?

pvojtechovsky: in case of this method yes.

src/main/java/spoon/reflect/meta/impl/RoleHandlerHelper.java
```diff
@@ -71,4 +78,51 @@ public static RoleHandler getOptionalRoleHandler(Class<? extends CtElement> targ
 		}
 		return null;
 	}
+
+	/**
+	 * @param targetClass a Close whose handlers we are looking for
```
surli: I don't understand you're documentation here: what's a "Close"?

pvojtechovsky: I should sleep more ... at least sometime ... "Close" is wrong. Correct is "Class" ;-) I will fix it.

monperrus:
This PR is OK but for the method name `getParentRoleHandler`.
WDYT of `getRoleHandlerWrtParent` (wrt = with respect to) or `getRoleWrtParentHandler`?

monperrus:
thanks!

PR #1797, opened 2018-01-01
fix: DefaultJavaPrettyPrinter prints an ERROR comment instead of failure

pvojtechovsky:
In case  of some model inconsistency (missing field declaration) the DJPP failed before. It is much more debugging friendly if there is no failure, but if it prints the comment with error description instead.
WDYT?

### DIFF
--- a/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
+++ b/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
@@ -854,14 +854,16 @@ public class DefaultJavaPrettyPrinter implements CtVisitor, PrettyPrinter {
 						 * Search for potential variable declaration until we found a class which declares or inherits this field
 						 */
 						final CtField<?> field = f.getVariable().getFieldDeclaration();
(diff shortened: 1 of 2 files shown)


monperrus:
I agree that DJPP is not the most appropriate location for this check.

As error message I'd suggest something even more explicit `"/* ERROR: Missing field "fieldname", please check your model. The code may not compile."`.

We may move the exception raising in CtBody#addStatement directly.

pvojtechovsky:
> We may move the exception raising in CtBody#addStatement directly.

I don't think so, or I didn't understood you.

There are many ways how to break consistency of the model. And it is not good idea to introduce any kind of check, which assures that model is ALWAYS consistent, because:
* it will have performance impact
* in the middle of model building/refactoring the model can be temporarily inconsistent - until all the changes are done

May be we might introduce new method #checkConsistenty on some spoon model elements, so the client might call it when needed and to get the report about problems. But it would be different task.

And in all cases I already sometime needed to be able to print inconsistent model too - for debugging purposes - to understand what is the problem.

For me this PR is finished and read for merge.

PR #1798, opened 2018-01-01
feature: Spoon Filter implements java.util.function.Predicate

pvojtechovsky:


monperrus:
Java 8 forever!

monperrus:
This breaks more code than expected: all client-side implementations of Filter, which are common, are broken.
See https://ci.inria.fr/sos/ (NpeFix and Cocospoon)

I suggest to undo this change and rather have `AbstractFilter` implementing `Predicate`.

WDYT?

pvojtechovsky:
ok, I have no problem with undo. Please do it if you have some tool for that. Otherwise we can make new PR for that...

Could you show me short and fast example of the problem? Do they implement method test? or Not using java 8??

PR #1799, opened 2018-01-01
feature: SpoonModelTree can be created with custom root element

pvojtechovsky:
new constructor  SpoonModelTree(CtElement rootElement) allows to visualize spoon model of subtree.


pvojtechovsky:
Revapi seems to be confused ... this PR really did not changed anything on Filter

PR #1801, opened 2018-01-01
fix: detection of TemplateParameter in Parameters#isParameterSource

pvojtechovsky:
Each field whose type extends TemplateParameter is a template parameter now.

For example this is template parameter.
```java
CtExpression<String> _expression_;
```


monperrus:
very nice! could you mention this in the documentation?

PR #1805, opened 2018-01-01
review: feat: CtScanner#scan of Collection and Map

pvojtechovsky:
The main purpose of this PR is, to be able to handle scanning of whole collection or whole map. It is needed for #1686 for the matching algorithms which matches a sequence of elements. For example in list of 5 commands match first `if` followed by `for`

### DIFF
--- a/src/main/java/spoon/reflect/visitor/CtScanner.java
+++ b/src/main/java/spoon/reflect/visitor/CtScanner.java
@@ -149,6 +149,16 @@ public abstract class CtScanner implements CtVisitor {
 			}
 		}
 	}
+	/**
+	 * Generically scans a Map of meta-model elements.
+	 */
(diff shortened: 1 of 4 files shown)


pvojtechovsky:
This PR contains more thinks - but they are related and it is not easily possible to deliver them individually.
1) CtScanner - has a scan method now which is called for whole scanned Collection or Map
2) EarlyTerminatingScanner - implements this method and terminates early
3) EarlyTerminatingScanner - is more encapsulated now - ScanningMode contract is checked directly here - so clients can use it more easily
4) CtScannerFunction$Scanner - uses that EarlyTerminatingScanner ScanningMode contract

WDYT?

monperrus:
I'm OK with that.

Could you add a test case to specify to contract of this new public method?

monperrus:
thanks!

Issue #1791, opened 2017-12-31
Getting enum values from shadowed enum type


Egor18:
Hi. I'm trying to get all elements of some enum.
For the enum, which declaration is present in the source code this works fine:

```
...
CtTypeReference typeRef = expr.getType();
if (typeRef.isEnum())
{
  CtType type = typeRef.getTypeDeclaration();
  if (type instanceof CtEnum)
  {
    System.out.println(((CtEnum) type).getEnumValues());
  }
}
```

But once the enum is not present in the source (i.e. it is located in some .jar), this approach does not work.
For shadowed type reference `.isEnum()` always returns false, and the actual type of such enum is `CtClass` (why not `CtEnum`?).
So, it is impossible to cast it to `CtEnum` and call `.getEnumValues()`.

Obviously, to get enum values from enum which is located in .jar, I can check if it's parent is `java.lang.Enum` and then get all the fields, excluding `.ENUM$VALUES`, `name` and `ordinal`.
But it seems to be quite overcomplicated, because I need to use two different approaches for shadowed and non-shadowed enums.

So can you please explain, why is this happening? Is it possible to get shadowed enum values easier?
Furthermore, I'm interested, are there any similar problems with other data types?

P.S. This partially concerns [incremental build](https://github.com/INRIA/spoon/pull/1688).

Thank you.

monperrus:
> But once the enum is not present in the source (i.e. it is located in 
> some .jar), this approach does not work.
> For shadowed type reference |.isEnum()| always returns false, and the 
> actual type of such enum is |CtClass| (why not |CtEnum|?).
this sounds like a bug. could you propose a failing test case as pull 
request? thanks!

PR #1806, opened 2018-01-02
fix: Shadow classes should be interpreted according to their kind (enum, interface, etc)

Egor18:
Failing test for https://github.com/INRIA/spoon/issues/1791.
As you can see, test fails, for the enum from jar.

Class from EnumJar.jar:
```

public class EnumJar {

	enum Colors { Red, Black, White };
}

```


pvojtechovsky:
Thank you @Egor18  for excellent reporting of problem including test case!
Thank you @Simon for taking care!

Egor18:
Thank you.

src/main/java/spoon/support/visitor/java/JavaReflectionVisitorImpl.java
```diff
@@ -61,7 +61,15 @@ public void visitPackage(Package aPackage) {
 			visitField(field);
 		}
 		for (Class<?> aClass : clazz.getDeclaredClasses()) {
-			visitClass(aClass);
+			if (aClass.isAnnotation()) {
+				visitAnnotationClass((Class<Annotation>) aClass);
```
tdurieux: I think it will produce class cast exception.
The correct type is `Class<? extends Annotation>`

surli: I get this code from there: https://github.com/INRIA/spoon/blob/cef5e079f335f48133395cb1f369ce1e0cdf83cc/src/main/java/spoon/support/visitor/java/JavaReflectionTreeBuilder.java#L88

and apparently it's used for 2 years... But PR with a failing test case is welcome :)

PR #1808, opened 2018-01-03
review fix: Filter extends Predicate is now working with projects in JDK7

surli:
This PR intends to fix the bug introduced in #1798 which make projects using JDK7 and custom filters fail. The error comes from the usage of a default method: overriding it in AbstractFilter fixes the bug.

### DIFF
--- a/src/main/java/spoon/reflect/visitor/filter/AbstractFilter.java
+++ b/src/main/java/spoon/reflect/visitor/filter/AbstractFilter.java
@@ -60,4 +60,9 @@ public abstract class AbstractFilter<T extends CtElement> implements Filter<T> {
 	public boolean matches(T element) {
 		return type.isAssignableFrom(element.getClass());
 	}
+
+	@Override
+	public boolean test(T element) {
+		return matches(element);
+	}
 }


pvojtechovsky:
I wonder how these projects works? Are they compiling java 8 library (spoon) using java 7 compatible compiler?

Anyway if it helps, I am OK to merge it. This PR is OK for me.

monperrus:
really good, thanks!

surli:
Ok actually this was not a proper fix: code that implements the interface Filter might be broken too.
So, either I completely remove the `extends Predicate` on AbstractFilter, or we consider the change as breaking for JDK7. 

> Are they compiling java 8 library (spoon) using java 7 compatible compiler?

They are using a JDK 8 but the pom.xml target a Java 7 sourcecode, so I assume the default method are not managed in the produced binary leading to a compilation error.

pvojtechovsky:
lets remove Predicate from Filter. It is not so great idea - if it breaks compatibility. We can add it again later after some years ... after everybody migrates

surli:
> We can add it again later after some years ... after everybody migrates

Actually I thought that AbstractFilter was more used, so we would be able to implement it in there, but no: AbstractFilter is almost a  specific kind of Filter, not really an **abstract** filter. 

I think a good first step, might be to refactor the hierarchy of filters to create a real abstract filter class that - in the future - we can enrich if needed.

pvojtechovsky:
> to refactor the hierarchy of filters to create a real abstract filter class

I think on many places in spoon there is already used Filter implementation using Lambda
```java
anElement.filterChildren((CtAnnotation m) -> val.equals(m.getValue(...)))...
```
There is no way how to replace it by AbstractFilter, without making code ugly again.

surli:
> There is no way how to replace it by AbstractFilter, without making code ugly again.

Hmmm indeed.

Issue #1814, opened 2018-01-08
Using maven dependency 3.5.0 might cause issues with Maven <= 3.3.9 


surli:
I got the following comment from @tmortagne on https://github.com/SpoonLabs/spoon-maven-plugin/pull/51#issuecomment-355937450 : 

> Now I just noticed that soon-core 6.1.0 itself has a dependency on Maven 3.5.0 so it might create some issues (on my side I can't build the processor because of that, because a dependency require a dependency more recent that the one I allow).

PR #1815, opened 2018-01-08
review chore: Check if Spoon compile with Maven 3.3.9

surli:


surli:
This PR aims at fixing #1814. However, I did not reproduce any error by just compiling Spoon using Maven 3.3.9 before the downgrade (see second commit status). I assume the error occured when building a custom processor with Spoon as dependency: could you confirm that @tmortagne? 

On the other hand, downgrading the Maven dependency introduced in Spoon for MavenLaucher to 3.3.9 seems to work without errors. Maven 3.5.0 is not largely deployed so I'm in favor of merging this PR as it is. 

WDYT @pvojtechovsky @monperrus ?

monperrus:
OK for me.

tmortagne:
> I assume the error occured when building a custom processor with Spoon as dependency: could you confirm that @tmortagne?

It's a bit special in my case because I have a <dependencyManagement> with Maven 3.3.9 and at the same time an enforcer plugin rules that check the depencencies version compatibility. And since Spoon "requires" 3.5.0 so the enforcer rule fail.

The other issue I have is that revapi plugin seems to be disturbed by the dependencies coming from Maven 3.5.0 when building with 3.3.9.

In any case it's safer to not use a too recent version of Maven.

surli:
So if everyone agree, please merge it.

Issue #1835, opened 2018-02-06
bug: Exception raised by `ClassTypingContext#isOverriding` ("Cannot adapt type parameters of non type scope")


monperrus:
In some cases,  `ClassTypingContext#isOverriding` crasheswith an exception:

```
spoon.SpoonException: Cannot adapt type parameters of non type scope

	at spoon.support.visitor.ClassTypingContext.resolveTypeParameters(ClassTypingContext.java:473)
	at spoon.support.visitor.ClassTypingContext.access$000(ClassTypingContext.java:62)
	at spoon.support.visitor.ClassTypingContext$1.accept(ClassTypingContext.java:234)
	at spoon.support.visitor.ClassTypingContext$1.accept(ClassTypingContext.java:203)
	at spoon.reflect.visitor.chain.CtQueryImpl$OutputFunctionWrapper._accept(CtQueryImpl.java:471)
	at spoon.reflect.visitor.chain.CtQueryImpl$AbstractStep.accept(CtQueryImpl.java:319)
	at spoon.reflect.visitor.filter.SuperInheritanceHierarchyFunction.sendResult(SuperInheritanceHierarchyFunction.java:347)
	at spoon.reflect.visitor.filter.SuperInheritanceHierarchyFunction.visitSuperClasses(SuperInheritanceHierarchyFunction.java:281)
	at spoon.reflect.visitor.filter.SuperInheritanceHierarchyFunction.apply(SuperInheritanceHierarchyFunction.java:248)
	at spoon.reflect.visitor.filter.SuperInheritanceHierarchyFunction.apply(SuperInheritanceHierarchyFunction.java:51)
	at spoon.reflect.visitor.chain.CtQueryImpl$LazyFunctionWrapper._accept(CtQueryImpl.java:505)
	at spoon.reflect.visitor.chain.CtQueryImpl$AbstractStep.accept(CtQueryImpl.java:319)
	at spoon.reflect.visitor.chain.CtQueryImpl.forEach(CtQueryImpl.java:105)
	at spoon.support.visitor.ClassTypingContext.resolveActualTypeArgumentsOf(ClassTypingContext.java:203)
	at spoon.support.visitor.ClassTypingContext.isSubtypeOf(ClassTypingContext.java:133)
	at spoon.support.visitor.ClassTypingContext.isOverriding(ClassTypingContext.java:274)
```

One can reproduce the problem with 

```java
		Launcher spoon = new Launcher();
[...]

PR #1837, opened 2018-02-06
fix: ClassTypingContext resolveTypeParameter of outer parameters

pvojtechovsky:
fix of #1835

The problem was caused by type parameters, which were declared out of the scope of ClassTypingContext. Such type parameters must be kept as they are. It is not model inconsistency or reason to fail.


monperrus:
thanks a lot!

PR #1838, opened 2018-02-06
fix some small bugs

Egor18:
Hi! I was working on static analysis tool prototype, and it probably found few nasty misprints in spoon.
So, I fixed them.


pvojtechovsky:
Hi! These fixes looks correct for me. Many thanks!

Could you add simple test cases, which failed before your fix? But if you have no failing test case, then no problem. I can merge it as it is ;-)

Test cases would be helpful for future to avoid similar problems. Note: Spoon code is quite complex and without good test cases it cannot evolve well.

Egor18:
Sorry, I can't provide actual failing tests, since I'm not so familliar with this particular code.
But still, these are definately misprints.

PR #1845, opened 2018-02-09
fix EqualsChecker for assignments

Egor18:
Hi! I have found a bug, in `EqualsChecker` class, and this PR is a possible fix.
The problem is that `.equals` method called on `CtAssignment` and `CtOperatorAssignment` (which extends `CtAssignment`) returns true, once rhs and lhs statements are the same. But this is not correct.
For example, statements `x = 42` and `x += 42` are not the same thing at all.
To fix that, I implemented `visitCtAssignment` method.
Also, I provided some tests (second assert fails, without this fix).

### DIFF
--- a/src/main/java/spoon/support/visitor/equals/EqualsChecker.java
+++ b/src/main/java/spoon/support/visitor/equals/EqualsChecker.java
@@ -16,6 +16,7 @@
  */
 package spoon.support.visitor.equals;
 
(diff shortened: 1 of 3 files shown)


monperrus:
Thanks a lot.

I wonder whether we don't have similar bugs due to covariance / contravariance in the equality visitor.

PR #1847, opened 2018-02-13
fix getFile() for serialization

Egor18:
Hi! I found out that filename from `SourcePosition` is lost during serialization (it's null). Although line number, column, etc. are present in deserialized model.
This happens because `getFile()` uses `compilationUnit`, which is null in deserialized model.
I fixed that by introducing field '`file`' in `SourcePosition`.
Also, I provided some tests. Particularly, without this fix the following assertions fail:
```
assertTrue(type.getPosition().getFile().equals(typeFromFile.getPosition().getFile()));
....
assertTrue(elem1.getPosition().getFile().equals(elem2.getPosition().getFile()));
```

Really want this fix, because without it we can't get any filename from SourcePosition from deserialized model.

### DIFF
--- a/src/main/java/spoon/support/reflect/cu/position/SourcePositionImpl.java
+++ b/src/main/java/spoon/support/reflect/cu/position/SourcePositionImpl.java
@@ -107,9 +107,15 @@ public class SourcePositionImpl implements SourcePosition, Serializable {
 	 */
 	int[] lineSeparatorPositions;
 
+	/** The file for this position. */
+	private File file;
+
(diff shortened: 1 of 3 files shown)


monperrus:
OK for me.

pvojtechovsky:
OK for me too

surli:
Hi @Egor18 thanks for the fix! Could you just add a contract in comment of the test, then I'll merge the PR?

Egor18:
Thanks!

PR #1853, opened 2018-02-16
refactor: push down the equality checking code in EqualsVisitor

monperrus:
the design here was quite broken. this PR fixes it.

    CtAbstractVisitor <- CtBiScannerDefault (generated from CtBiScannerTemplate) <- EqualVisitor


src/main/java/spoon/reflect/visitor/CtAbstractBiScanner.java
```diff
@@ -19,81 +19,15 @@
...
-			}
-			return isNotEqual;
-		} else if (other == null) {
-			return fail();
-		}
-		if (element == other) {
-			return isNotEqual;
-		}
+	/** This method is called to compare `element` and `other` when traversing two trees in parallel.*/
+	public abstract boolean biScan(CtElement element, CtElement other);
```
pvojtechovsky: CtAbstractBiScanner sounds now like general purpose bi-scanner. Do I understood it well? If yes, then please change comments here (it is no more about "compare") and change return value from `boolean` to `void`
WDYT?

src/main/java/spoon/support/visitor/equals/EqualsVisitor.java
```diff
@@ -41,5 +44,60 @@ protected void enter(CtElement e) {
...
+			}
+			return isNotEqual;
+		} else if (other == null) {
+			return fail();
+		}
+		if (element == other) {
+			return isNotEqual;
+		}
+
+		stack.push(other);
```
pvojtechovsky: Could we move `stack.push(other);` and `stack,pop()` somewhere into `CtBiScannerDefault`? I looks like implementation of `CtBiScannerDefault` in this PR is useless without calls of `stack.push(other);` and `stack,pop()`. Because `CtBiScannerDefault` calls `stack.peek()`, but nobody pushes data there.
WDYT?

pvojtechovsky:
It would be probably more readable if all biScan methods would return void instead of boolean which is on most places ignored. The only place where return value is actually used is 
```java
public class EqualsVisitor extends CtBiScannerDefault {
	public static boolean equals(CtElement element, CtElement other) {
		return !new EqualsVisitor().biScan(element, other);
	}
```
which might be replaced by
```java
	public static boolean equals(CtElement element, CtElement other) {
		return new EqualsVisitor().isEqual(element, other);
	}
...
boolean isEqual(CtElement element, CtElement other) {
  biScan(element, other);
  return !isNotEqual;
}
```

src/main/java/spoon/reflect/visitor/CtAbstractBiScanner.java
monperrus: > CtAbstractBiScanner sounds now like general purpose bi-scanner. Do I understood it well?

yes, correct.

I agree with you, done.

src/main/java/spoon/support/visitor/equals/EqualsVisitor.java
monperrus: Good point. Done.

src/main/java/spoon/reflect/visitor/CtBiScannerDefault.java
```diff
@@ -28,7 +36,35 @@
...
+			stack.pop();
+		}
+	}
+
+	public void biScan(CtRole role, CtElement element, CtElement other) {
+		biScan(element, other);
+	}
+
+	protected boolean biScan(CtRole role, Collection<? extends CtElement> elements, Collection<? extends CtElement> others) {
+		for (Iterator<? extends CtElement> firstIt = elements.iterator(), secondIt = others.iterator(); (firstIt.hasNext()) && (secondIt.hasNext());) {
```
pvojtechovsky: I guess this code does not work for `Set` since beginning, because you cannot assure that elements in 2 Sets are iterated in consistent pairs. I wonder why it works ... WDYT?

monperrus:
for the beauty of it, added a test to do a parallel biscan visit of the same tree.

monperrus:
This is a question we discussed when we introduced it.

A strong assumption -- even a a requirement --  is we only use implementations of `Set` which have a deterministic iterator order.

pvojtechovsky:
> we only use implementations of Set which have a deterministic iterator order
Interesting! Good to know! Thanks!

PR #1864, opened 2018-02-17
feat: one could add a type member that already exists (equals but not same) and modify it afterwards

### LINKED ISSUE #1862: ConstructorFactory.create(someClass, someConstructorOfThisClass) doesn't set the parent of the created constrcutor properly
The exact problem is that `ConstructorFactory.create(someClass, someConstructorOfThisClass)` doesn't work properly: parent of the created constructor is null rather than `someClass`.

The fix that I applied and it helped is:
```
diff --git a/src/main/java/spoon/support/reflect/declaration/CtTypeImpl.java b/src/main/java/spoon/support/reflect/declaration/CtTypeImpl.java
index 61e727ea..bf654d15 100644
--- a/src/main/java/spoon/support/reflect/declaration/CtTypeImpl.java
+++ b/src/main/java/spoon/support/reflect/declaration/CtTypeImpl.java
@@ -123,7 +123,9 @@ public abstract class CtTypeImpl<T> extends CtNamedElementImpl implements CtType
 		if (this.typeMembers == CtElementImpl.<CtTypeMember>emptyList()) {
 			this.typeMembers = new SortedList<>(new CtLineElementComparator());
 		}
-		if (!this.typeMembers.contains(member)) {
+		// Don't use typeMembers.contains(), because member is potentially a clone of another
+		// member and "not yet" different, so it could be found in the members collection.
+		if (!this.typeMembers.stream().anyMatch(m -> m == member)) {
 			member.setParent(this);
 			CtRole role;
 			if (member instanceof CtMethod) {
```

But, I think it reveals a much larger problem, touching many types of elements across the library. Related to #646.

monperrus:
fix #1862

### DIFF
--- a/src/main/java/spoon/support/reflect/declaration/CtClassImpl.java
+++ b/src/main/java/spoon/support/reflect/declaration/CtClassImpl.java
@@ -104,6 +104,9 @@ public class CtClassImpl<T extends Object> extends CtTypeImpl<T> implements CtCl
 		Set<CtConstructor<T>> constructors = new SignatureBasedSortedSet<>();
 		for (CtTypeMember typeMember : typeMembers) {
 			if (typeMember instanceof CtConstructor) {
+				if (constructors.contains(typeMember)) {
+					throw new SpoonException("ooops, impossible to have two constructors with the same signature");
+				}
 				constructors.add((CtConstructor<T>) typeMember);
 			}
 		}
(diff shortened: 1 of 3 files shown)


pvojtechovsky:
`throw new SpoonException("ooops, impossible to have two constructors with the same signature");`
I would say that is not good idea to fail when Spoon model is not 100% compilable/consistent.
1) we cannot assure that Spoon model is 100% consistent
2) during Spoon model transformation/refactoring the model might be temporarily inconsistent - until refactoring is finished. And your change might cause that refactoring fails in the middle or it would need more complicated algorithm, which avoids inconsistent states.

I would prefer instead a new spoon model method `checkConsistency`, which would report all inconsistencies. Such check would be then part of this consistency checking.

WDYT?

monperrus:
Yes, I agree. Here is the alternative solution.

Issue #1865, opened 2018-02-18
Return public DefaultJavaPrettyPrinter.writeHeader() method


leventov:
As a side effect of #784 (https://github.com/INRIA/spoon/pull/784/files#diff-5f33bf5d1f38230de815e4c60a67a215L2328), `DefaultJavaPrettyPrinter` doesn't have public `writeHeader()` method anymore. It makes public `computeImports()` and `scan()` much less useful, because you cannot print header (imports) anyway (`calculate()` does all those things together). However the ability to compute imports and write them (via `writeHeader()`) is useful for reorganizing or changing the imports list.

monperrus:
closed per  #1866

PR #1866, opened 2018-02-18
refactor: restore DJPP#writeHeaders that clients can override

monperrus:
simple fix for #1865 

@leventov WDYT?


leventov:
LGTM

PR #1870, opened 2018-02-18
fix: fully comply with the contract of CtScannerTest#testScannerCallsAllProperties

### LINKED ISSUE #1712: CtScanner doesn't scan all comments.
The missing code in CtScanner can be found by change in CtScannerTest#testScannerCallsAllProperties.
Comment this code:
```java
				if(getter.getName().equals("getComments")) {
					//ignore missing getComments ... until discussion about its contracts is finished
					return;
				}
```
and test will fail and report all missing `scan(CtRole.COMMENT, ...)` calls.

monperrus:
all comments are actually scanned. only in references, there is no scan of comments, but this is normal, since references cannot be documented.
 
fix #1712


src/main/java/spoon/reflect/visitor/CtBiScannerDefault.java
```diff
@@ -37,15 +29,15 @@
  */
 // autogenerated by CtBiScannerGenerator
 public class CtBiScannerDefault extends spoon.reflect.visitor.CtAbstractBiScanner {
-	protected Deque<CtElement> stack = new ArrayDeque<>();
+	protected java.util.Deque<spoon.reflect.declaration.CtElement> stack = new java.util.ArrayDeque<>();
```
pvojtechovsky: Is it correct that current generator produces fully qualified names here?

monperrus: yes, the BiScanner generator does not use autoimport

pvojtechovsky:
Thank You for solving this issue!

Issue #1872, opened 2018-02-21
bug: CtElement.getRoleInParent() and CtElement.getValueByRole(CtRole) are not consistent for catch with multple types


nharrand:
When parsing 
```Java
	try {
		Object val = m.get(o);
		if (val != null) {
			return val.getClass().getCanonicalName();
		}
	} catch (IllegalArgumentException | IllegalAccessException e) {
		Launcher.LOGGER.error(e.getMessage(), e);
	}
```
the node corresponding to the type of `java.lang.IllegalArgumentException` has a role `CtRole.TYPE` in its parent `CtCatchVariable`, but if `getValueByRole(CtRole.TYPE)` is called on this parent, it returns the type `java.lang.Exception`. (Wheras it should return, if I understand correctly, a List containing both types `java.lang.IllegalArgumentException` and `java.lang.IllegalAccessException`)

pvojtechovsky:
Hi Nicolas! Thanks for reporting that problem and defining test case. 

@monperrus, @surli it is a tricky case ... We probably have to add new CtRole#MULTI_TYPE

But note that there is a relation between new CtRole#MULTI_TYPE and existing CtRole#TYPE. The `type` is derived from `multiType` in this case... and because we do not provide access to derived properties using CtRole yet, we should - may be - remove CtRole#TYPE out of there??

WDYT?

monperrus:
> we should - may be - remove CtRole#TYPE out of there

OK for me.

monperrus:
closed per #1877

PR #1874, opened 2018-02-21
feat: Adds unique serializable path for each CtElement from Root Element.

nharrand:
The goal of this PR is to add support for getting an unique `CtPath` for each `CtElement` of a model. In particular this could be used for persistence. Any given `CtElement` can be access through an unique deterministic serializable path. All the following transformations are supported: `CtElement` -> `CtPath` -> `String` -> `CtPath` -> `CtElement`.

Main content:
 * Method `CtPath CtElement.getPath()`: Returns the path from root element to `this`.
 * Class `CtElementPathBuilder`: Build a `CtPath` from a given `CtElement`
 * Class `CtUniqueRolePathElement `: Extends CtRolePathElement with a new behavior and a new syntax.
 * Method `MainTest.testElementToPathToElementEquivalency()`: Test `CtElement` -> `CtPath` -> `String` -> `CtPath` -> `CtElement`

In order to integrate this to `CtPathStringBuilder` without conflict, the char '@' is used for this new path element.

The syntax for paths is the following
 * For unique CtElement -> `@role`
 * For elements in a List -> `@role[index=ID]`
 * For elements in a  Set of CtElement for which name exists and is unique (Either CtNamedElement or CtReference) -> `@role[name=element.getSimpleName()]`
 * For elements in a Map<String, CtElement> -> `@role[key=String]`

`@subPackage[name=spoon]@subPackage[name=compiler]@containedType[name=Environment]@typeMember[index=0]` -> `spoon.compiler.Environment` 's first type member ie: `int getComplianceLevel();`

Currently the test fails (related to #1872).

What do you think?


src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
```diff
@@ -536,4 +540,12 @@ public CtElement clone() {
 		rh.setValue(this, value);
 		return (E) this;
 	}
+
+	public CtPath getPath() {
+		try {
```
monperrus: why not removing the try/catch and let the exception go?

src/test/java/spoon/test/main/MainTest.java
```diff
@@ -439,6 +435,35 @@ public void scan(CtRole role, CtElement element) {
...
+					CtPath path = element.getPath();
+					String pathStr = path.toString();
+					try {
+						CtPath pathRead = new CtPathStringBuilder().fromString(pathStr);
+						Collection<CtElement> returnedElements = pathRead.evaluateOn(list);
+						//contract: CtUniqueRolePathElement.evaluateOn() returns a unique elements if provided only a list of one inputs
+						assertEquals(returnedElements.size(), 1);
+						CtElement actualElement = (CtElement) returnedElements.toArray()[0];
+						//contract: Element -> Path -> String -> Path -> Element leads to the original element
+						assertEquals(element, actualElement);
```
monperrus: assertSame?

monperrus:
looks super cool, I love the strong generic test.

are you working on fixing #1872?

src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
nharrand: The exception is supposed to be risen from CtElementPathBuilder().fromElement(CtElement el, CtElement from) when from is not a parent of el. So in this case, I think it can't happen and I didn't want to annoy the user with the handling of an exception which is not supposed to happen.
But still, we can let it go.

src/test/java/spoon/test/main/MainTest.java
nharrand: Yes, that sounds better. I'll fix it ASAP.

nharrand:
@monperrus I tried to add:
```java
	static class CtCatchVariable_TYPE_RoleHandler extends ListHandler<CtCatchVariable, CtTypeReference<?>> {
		private CtExpression_CAST_RoleHandler() {
			super(CtRole.CAST, CtExpression.class, CtTypeReference.class);
		}

		@Override
		public <T, U> U getValue(T element) {
			return ((U) ((Object) (castTarget(element).getMultiTypes())));
		}

		@Override
		public <T, U> void setValue(T element, U value) {
			castTarget(element).setMultiTypes(castValue(value));
		}
}
```
to `ModelRoleHandlers`, which fixed the issue but broke  another test. I realized that his code is actually generated, if I'm not mistaken. So, this is where I got stuck.

nharrand:
@monperrus I'm not sure why does the Travis CI build fails. It mentions Javadoc errors in files that I haven't changed. Any ideas?

Do you want to keep the current syntax for `CtUniqueRolePathElement` when transformed to String? Do you want me to merge it to the pre existing syntax?

monperrus:
I propose to replace the previous implementation, to keep the old syntax with slash, and thus to update `CtPathTest`. It means we would drop the support for wildcard temporarily, but I'm OK with that. (This was an experimental partial feature, with no user to my knowledge).

WDYT?

monperrus:
Fixed the Javadoc error at CtPathStringBuilder.java line 71 (please pull).

(Maven prefixes "warnings" of the Javadoc linter by `[Error]` which is super confusing, but later on the line it's clear that it is only a warning)

monperrus:
> Do you want me to merge it to the pre existing syntax?

Yes, see https://github.com/INRIA/spoon/pull/1874#issuecomment-368088909

nharrand:
We can keep the wildcards, they are a part of `CtNamedPathElement`, and can already be used in combination of the new syntax.
The only changes are that block cannot be omitted anymore.
Example:
```Java
new CtPathStringBuilder().fromString(".spoon.test.path.Foo.foo#body[index=0]");
```
become
```Java
new CtPathStringBuilder().fromString(".spoon.test.path.Foo.foo#body#statement[index=0]");
```

src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
monperrus: "let it go" then :-)

monperrus:
the change is OK for me.

it seems we're close to merge!

src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
nharrand: I have a counter proposal: Why not catch `CtPathException` (which should not happen in that context) and throw a `SpoonException` to show that it's an spoon implementation error. Further more it would avoid that the user has to handle it.

src/main/java/spoon/reflect/path/impl/CtRolePathElement.java
```diff
@@ -106,14 +55,57 @@ public CtRole getRole() {
...
+				} else if (root.getValueByRole(getRole()) instanceof Map) {
+					if (getArguments().containsKey("key")) {
+						String name = getArguments().get("key");
+						matchs.add((CtElement) ((Map) root.getValueByRole(getRole())).get(name));
+					}
+				} else {
+					CtElement el = root.getValueByRole(getRole());
+					matchs.add(el);
+				}
+			} catch (SpoonException e) { } //When no element are found for a given role, return empty list.
```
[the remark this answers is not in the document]
nharrand: I guess one alternative would be adding a method `hasChildFor(CtRole role)` to `CtElement`.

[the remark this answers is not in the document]
pvojtechovsky: It is already there. See `RoleHandlerHelper#getOptionalRoleHandler`
Have a look also at `RoleHandler#getContainerKind`, which might be used instead of `root.getValueByRole(getRole()) instanceof Set` and similar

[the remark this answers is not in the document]
nharrand: @pvojtechovsky Thanks for the tip.
Problem solved.

monperrus:
OK for me. Will merge it. That's a really nice contribution to Spoon, thanks!

monperrus:
what about changing the signature of `evaluateOn` to `evaluateOn(CtElement...)` (varargs)?. the API would be easier for clients.

monperrus:
sorry @nharrand , I fixed the test by calling `super.scan(element)`, and it seems it now finds a bug :-)

nharrand:
Yes, I see that! I'm looking into that.

nharrand:
The error is a bit disturbing:
```
MainTest.testElementToPathToElementEquivalency:455 expected same:<java.lang.String> was not:<java.lang.String>
```
I bet that the assumption that name of named element was enough to differentiate did not hold for this case.

nharrand:
Ok this should work now.

src/main/java/spoon/reflect/path/impl/CtRolePathElement.java
```diff
@@ -106,14 +54,65 @@ public CtRole getRole() {
...
+				switch (roleHandler.getContainerKind()) {
+					case SINGLE:
+						matchs.add(roleHandler.getValue(root));
+						break;
+
+					case LIST:
+						if (getArguments().containsKey("index")) {
+							int index = Integer.parseInt(getArguments().get("index"));
+							matchs.add((CtElement) roleHandler.asList(root).get(index));
+						}
```
pvojtechovsky: else it should add all items of the into `matchs`. WDYT?

src/main/java/spoon/reflect/path/impl/CtRolePathElement.java
```diff
@@ -106,14 +54,65 @@ public CtRole getRole() {
...
+								//System.err.println("[ERROR] Element not found for name: " + name);
+								//No element found for name.
+							}
+						}
+						break;
+
+					case MAP:
+						if (getArguments().containsKey("key")) {
+							String name = getArguments().get("key");
+							matchs.add((CtElement) roleHandler.asMap(root).get(name));
```
pvojtechovsky: I guess it is not correct to add null into matchs, when map doesn't contain a key

src/main/java/spoon/reflect/path/impl/CtRolePathElement.java
pvojtechovsky: and what about case when `index >= size()`?

nharrand: Yes, you are right. I only had in mind my Unique path case in mind here, but yes. I'll fix this

src/main/java/spoon/reflect/path/impl/CtRolePathElement.java
nharrand: Indeed. I'll fix this too.

src/main/java/spoon/reflect/path/impl/CtRolePathElement.java
nharrand: Right, that also means that I should add negative test cases...

src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
```diff
@@ -536,4 +541,12 @@ public CtElement clone() {
...
+			return new CtElementPathBuilder().fromElement(this, getParent(CtModelImpl.CtRootPackage.class));
+		} catch (CtPathException e) {
+			throw new SpoonException();
```
pvojtechovsky: `throw new SpoonException(e);` ... the orgin exception should be used as cause of the SpoonException

pvojtechovsky: Yes, no exception should happen, but if it happens because of some mistake in spoon code, then your current implementation will avoid efficient detection of the cause, because the cause of the problem is lost. So just type in `e` into brackets and I am OK with that ;-)

nharrand: > just type in e into brackets

What do you mean? Shall I add `e.printStackTrace();` or change `throw new SpoonException();` into `throw new SpoonException(e);` ?

pvojtechovsky: yes please

nharrand: Done in 8cfc1adeff6331dc05d73a316ab6d78d2cc54e7d

src/main/java/spoon/reflect/path/impl/CtRolePathElement.java
nharrand: > else it should add all items of the into `matchs`.
We should extends that behavior to sets and maps, don't you think?

src/main/java/spoon/reflect/path/impl/CtRolePathElement.java
```diff
@@ -106,14 +55,78 @@ public CtRole getRole() {
 
 	@Override
 	public String toString() {
-		return STRING + role.toString() + getParamString();
+		return STRING + getRole().toString() + getParamString();
+	}
+
+	public CtElement getFromSet(Set set, String name) throws CtPathException {
```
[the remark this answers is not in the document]
nharrand: No, we don't. I'll change it.

src/main/java/spoon/reflect/path/impl/CtRolePathElement.java
```diff
@@ -106,14 +55,78 @@ public CtRole getRole() {
...
+				}
+			} else if (o instanceof CtReference) {
+				if (((CtReference) o).getSimpleName().equals(name)) {
+					return (CtElement) o;
+				}
+			} else {
+				throw new CtPathException();
+			}
+		}
+		throw new CtPathException();
```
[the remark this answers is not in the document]
nharrand: Fixed in 7ee992176b01e138e1d5d4f07d9f53247c04eba6

pvojtechovsky:
It looks good to me. I am looking forward to use this new Spoon feature! Thank You @nharrand .

I guess we can merge. @monperrus WDYT?

monperrus:
OK to merge! This opens super exciting new usages.

Kudos and thanks to @nharrand

nharrand:
You are welcome. Thanks for the reviews.

I'm a little disappointed that it counts as only one commit though... ;-)

monperrus:
that's the tough law of squashing!

Issue #1869, opened 2018-02-18
CtExecutableReference in cloned method should refer to the clone, not to the method being cloned.


leventov:
Diff of the test which fails:
```
diff --git a/src/test/java/spoon/test/method/MethodTest.java b/src/test/java/spoon/test/method/MethodTest.java
index 0b585aa0..d689d4c0 100644
--- a/src/test/java/spoon/test/method/MethodTest.java
+++ b/src/test/java/spoon/test/method/MethodTest.java
@@ -17,6 +17,7 @@

 package spoon.test.method;

+import org.junit.Assert;
 import org.junit.Test;
 import spoon.Launcher;
 import spoon.reflect.declaration.CtClass;
@@ -24,6 +25,8 @@ import spoon.reflect.declaration.CtMethod;
 import spoon.reflect.declaration.CtType;
 import spoon.reflect.declaration.ModifierKind;
 import spoon.reflect.factory.Factory;
+import spoon.reflect.reference.CtExecutableReference;
+import spoon.reflect.visitor.CtScanner;
 import spoon.reflect.visitor.filter.NamedElementFilter;
 import spoon.test.delete.testclasses.Adobada;
 import spoon.test.method.testclasses.Tacos;
@@ -110,4 +113,22 @@ public class MethodTest {
 		assertTrue(compareFound);
 	}

+	@Test
+	public void testCloneMethod() throws Exception {
+		Launcher l = new Launcher();
+		l.getEnvironment().setNoClasspath(true);
+		l.addInputResource("src/test/resources/noclasspath/A2.java");
+		l.buildModel();
+		CtMethod<?> methodB = l.getFactory().Class().get("A2").getMethodsByName("b").get(0);
+		CtMethod<?> methodClone = methodB.clone();
+		methodClone.accept(new CtScanner() {
+			@Override
+			public <T> void visitCtExecutableReference(CtExecutableReference<T> reference) {
+				if (reference.getDeclaration() == methodB) {
+					Assert.fail("Cloned method shouldn't have references to the former method");
+				}
+				super.visitCtExecutableReference(reference);
+			}
+		});
+	}
 }
```

monperrus:
I have two concerns with this.

First, there is no clearly better semantics. One can object that the cloned should call the same methods as before. Both semantics can be confusing depending on one's assumption.

Second, it is impossible, since reference resolving is name-based and the method name after cloning is the same, so it would resolve to the same declaration. That's normal.

[...]

PR #1875, opened 2018-02-21
refactor: remove CtParameterReferenceImpl#declaringExecutable

monperrus:
By analyzing #1869, I realize that `CtParameterReference` and `CtParameterReferenceImpl` are badly designed. The problem is that `CtParameterReference` is stateful with a superfluous pointer to an executable reference:

- this goes against the design of philosophy of Spoon references that are mean to be resolved dynamically (instead of hard-wired links)
- when a parameter reference is moved in a transformation (eg from one variable to another), the result of `getDeclaration()`, done with dynamic lookup, can become inconsistent with the result of `getDeclaringExecutable()`
- when a parameter reference is cloned, the result of `getDeclaration()`, done with dynamic lookup is correct (see https://github.com/INRIA/spoon/issues/1869#issuecomment-366798381), yet inconsistent with the result of `getDeclaringExecutable()`, which is not changed during cloning. This is one facet of the problem discussed in #1869.

The solution presented here it to remove the `get/setDeclaringExecutable` from interface `CtParameterReference`, and to remove the corresponding state in the implementation (field `declaringExecutable`).

This fixes the three problems described above. There are a few test changes, which are minor, they were simply directly unit-testing `get/setDeclaringExecutable`.

Once this one is merged, there will be other PRs for #1869.


pvojtechovsky:
I agree that field `CtParameterReferenceImpl#executable` should be removed.

But would it be possible to KEEP `CtParameterReference[Impl]#getDeclaringExecutable()` ?? May be we can mark it @Derived and compute it's value from the Spoon model?

monperrus:
you're right. it's better for backward compatibility. done.

pvojtechovsky:
Looks good to me. @monperrus are you done with this PR? Do you want to merge it soon (tomorrow morning) or should we wait for feedback of others here?

monperrus:
OK to merge.

PR #1878, opened 2018-02-22
review: refactor: refactor cloning code to increase extendability

monperrus:
We want to be able, in the future, to slightly adapt the cloning behavior without touching the generator of cloning code.

In particular, it will be useful for #1869

### DIFF
--- a/src/main/java/spoon/support/visitor/equals/CloneHelper.java
+++ b/src/main/java/spoon/support/visitor/equals/CloneHelper.java
@@ -137,23 +137,22 @@ public class CloneHelper {
 		targetMap.put(key, clone(value));
 	}
 
-	private CtElement initialTopLevel;
-	private CtElement cloneTopLevel;
 
-	public void tailor(spoon.reflect.declaration.CtElement element, spoon.reflect.declaration.CtElement other) {
-		this.initialTopLevel = element;
-		this.cloneTopLevel = other;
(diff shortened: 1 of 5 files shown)


monperrus:
CloneTailoringVistor may rather be a BiScanner

src/main/java/spoon/support/visitor/equals/CloneHelper.java
```diff
@@ -134,4 +136,24 @@
 	protected <T extends CtElement> void addClone(Map<String, T> targetMap, String key, T value) {
 		targetMap.put(key, clone(value));
 	}
+
+	private CtElement initialTopLevel;
+	private CtElement cloneTopLevel;
```
pvojtechovsky: What is purpose of these two members? They are actually useless - not accessible.

src/main/java/spoon/support/visitor/equals/CloneHelper.java
```diff
@@ -134,4 +136,24 @@
...
+
+	public void tailor(spoon.reflect.declaration.CtElement element, spoon.reflect.declaration.CtElement other) {
```
pvojtechovsky: javadoc? ;-)

monperrus:
done :-)

monperrus:
you're right. not needed anymore since we use an inner class in "tailor". changed.

PR #1881, opened 2018-02-23
chore: fix javadoc linter config in pom.xml

monperrus:


monperrus:
not that with this change, the warning are still displayed, but not with the confusing "[Error]" prefix. see https://travis-ci.org/INRIA/spoon/jobs/345365932

PR #1884, opened 2018-02-25
feature: add methods Refactoring#copyType and Refactoring#copyMethod

monperrus:
plus changeMethodName as useful bonus, see testCopyMethod

Advanced cloning usages requires more than only copying all metamodel fields. For instance, it also requires to update references and to set a parent (for future resolving with `getDeclaration`).

This is what is done here.


pvojtechovsky:
Thank you Martin!

PR #1890, opened 2018-03-02
fix: RoleHandler setter for SUPER_TYPE and INTERFACE

### LINKED ISSUE #1887: RoleHandler setter isn't implemented well in some cases
For example RoleHandler can't be used to call `CtClass#setSuperClass`, because CtTypeInformation has no setter for that. It is a bug in RoleHandlerGenerator.

I am going to fix that.

pvojtechovsky:
fix #1887 

By the way, is it correct that some unsettable properties sets a value? This PR contains a comment assertion, which reports that "problem".


monperrus:
Thanks. Impressive coverage increase.

> By the way, is it correct that some unsettable properties sets a value?

probably not. or it's a misplaced `@Unsettable`.

Issue #1889, opened 2018-03-02
Parsing class with name containing $ and with a nested classe fails


nharrand:
```Java
Launcher launcher = new Launcher();
launcher.addInputResource("./$Poor$Name.java");
launcher.buildModel();
CtModel model = launcher.getModel();
System.out.println(model.getRootPackage().getType("$Poor$Name"));
```
with file `$Poor$Name.java` containing:
```Java
public class $Poor$Name {
    public Choice lookingForTroubles() {
        return new Choice();
    }
    class Choice {
        public Choice() {}
    }
}
```
produces the following exceptions:
```
Exception in thread "main" spoon.support.SpoonClassNotFoundException: cannot load class: $Poor$Name
	at spoon.support.reflect.reference.CtTypeReferenceImpl.findClass(CtTypeReferenceImpl.java:164)
	at spoon.support.reflect.reference.CtTypeReferenceImpl.getActualClass(CtTypeReferenceImpl.java:146)
	at spoon.support.reflect.reference.CtTypeReferenceImpl.getModifiers(CtTypeReferenceImpl.java:491)
	at spoon.support.reflect.reference.CtTypeReferenceImpl.canAccess(CtTypeReferenceImpl.java:621)
	at spoon.support.reflect.reference.CtTypeReferenceImpl.getAccessType(CtTypeReferenceImpl.java:691)
	at spoon.reflect.visitor.DefaultJavaPrettyPrinter.visitCtTypeReference(DefaultJavaPrettyPrinter.java:1824)
	at spoon.reflect.visitor.DefaultJavaPrettyPrinter.visitCtTypeReference(DefaultJavaPrettyPrinter.java:1794)
	at spoon.support.reflect.reference.CtTypeReferenceImpl.accept(CtTypeReferenceImpl.java:83)
	at spoon.reflect.visitor.DefaultJavaPrettyPrinter.scan(DefaultJavaPrettyPrinter.java:345)
	at spoon.reflect.visitor.DefaultJavaPrettyPrinter.visitCtMethod(DefaultJavaPrettyPrinter.java:1369)
	at spoon.support.reflect.declaration.CtMethodImpl.accept(CtMethodImpl.java:73)
	at spoon.reflect.visitor.DefaultJavaPrettyPrinter.scan(DefaultJavaPrettyPrinter.java:345)
	at spoon.reflect.visitor.ElementPrinterHelper.writeElementList(ElementPrinterHelper.java:196)
[...]

PR #1891, opened 2018-03-02
refactor(TypeFactory#get(String qualifiedName)): Look for type in factory.Package() first, and not at the end. Workaround for #1889

nharrand:
As mentioned in the name of the PR, this would refactor `TypeFactory.get(String qualifiedName)`. Instead of looking in existing packages for the wanted type only if it is not resolve otherwise, it starts by that.

It prevent to resolve recursively nested type from their root type as a first option. If you look for `A$B$C` it won't resolve `A` then `A$B` and then `A$B$C`. if `A$B` is already reachable, it will stop there.

This allow the resolution of class whose name contains `$`. The current behavior can't resolve a class named `A$B` if it's not a class `B` nested in `A`.

### DIFF
--- a/src/main/java/spoon/reflect/factory/TypeFactory.java
+++ b/src/main/java/spoon/reflect/factory/TypeFactory.java
@@ -419,6 +419,21 @@ public class TypeFactory extends SubFactory {
 	 */
 	@SuppressWarnings("unchecked")
 	public <T> CtType<T> get(final String qualifiedName) {
+		int packageIndex = qualifiedName.lastIndexOf(CtPackage.PACKAGE_SEPARATOR);
+		CtPackage pack;
+		if (packageIndex > 0) {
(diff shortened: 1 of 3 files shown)


monperrus:
Heavily used method yet all tests pass. 

OK for me.

src/main/java/spoon/reflect/factory/TypeFactory.java
```diff
@@ -419,6 +419,21 @@ public CtTypeParameterReference createReference(CtTypeParameter type) {
 	 */
 	@SuppressWarnings("unchecked")
 	public <T> CtType<T> get(final String qualifiedName) {
+		int packageIndex = qualifiedName.lastIndexOf(CtPackage.PACKAGE_SEPARATOR);
+		CtPackage pack;
+		if (packageIndex > 0) {
```
pvojtechovsky: `if (packageIndex >= 0)` would be better

PR #1893, opened 2018-03-03
review: fix: Unsettable properties sets no value or sets different primary value

pvojtechovsky:
Here is the test, which reports inconsistencies in Spoon model role get/set.

Generally there are these issues:
P1) Unsettable property CtXxxReference#comment<java.util.List<spoon.reflect.code.CtComment>> sets the value
P2) UnsettableProperty CtEnumValue|CtLocalVariable|CtField#assignment<spoon.reflect.code.CtExpression<T>> sets the value into different role defaultExpression
P3) UnsettableProperty CtLambda#thrown<java.util.Set<spoon.reflect.reference.CtTypeReference<? extends java.lang.Throwable>>> sets the value
P4) Settable field CtCatchVariable#type<spoon.reflect.reference.CtTypeReference<T>> is accessible by multiType too

I am not sure whether we should solve these problems. May be they are features for somebody?

Solutions might be 
>P1

to implement empty CtReference#setComment().

> P2

?? 

> P3
 
to implement empty CtLambda#setThrownTypes

>P4

to mark role CtCatchVariable#TYPE as derived

WDYT?

### DIFF
--- a/src/main/java/spoon/reflect/code/CtCatchVariable.java
+++ b/src/main/java/spoon/reflect/code/CtCatchVariable.java
@@ -19,6 +19,7 @@ package spoon.reflect.code;
 import spoon.reflect.declaration.CtMultiTypedElement;
 import spoon.reflect.declaration.CtVariable;
 import spoon.reflect.reference.CtCatchVariableReference;
+import spoon.reflect.reference.CtTypeReference;
 import spoon.support.DerivedProperty;
 import spoon.support.UnsettableProperty;
 
@@ -44,4 +45,8 @@ public interface CtCatchVariable<T> extends CtVariable<T>, CtMultiTypedElement, 
 	@Override
 	@UnsettableProperty
 	<C extends CtVariable<T>> C setDefaultExpression(CtExpression<T> assignedExpression);
+
+	@Override
+	@DerivedProperty
+	CtTypeReference<T> getType();
 }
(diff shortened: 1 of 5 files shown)


monperrus:
OK for me. For P2, I would simply remove the `@DerivedProperty` (the API documentation is clear about the goal).

pvojtechovsky:
The `@DerivedProperty` on getter is no problem. The problem is meaning of `@UnsettableProperty`. I understand the meaning: "you can call this method, but it will do nothing" ... but this contract is broken by code like
```java
class CtFieldImpl { ...
	@UnsettableProperty
	public <C extends CtRHSReceiver<T>> C setAssignment(CtExpression<T> assignment) {
		setDefaultExpression(assignment);
		return (C) this;
	}
```
May be we can live with that, I just wanted to be sure that we speak about the same thing.

monperrus:
My bad, I indeed meant UnsettableProperty:

For P2, I would simply remove the `@UnsettableProperty` (the API documentation is clear about the goal).

pvojtechovsky:
I am mainly not sure with CtLambdaImpl#setThrownTypes, where I deleted code added by You ;-)
A) it is correct, it has to be deleted
B) we have to remove UnsettableProperty from CtLambda#setThrownTypes

So if you agree with A), then this is finished

monperrus:
OK for me. Will merge it.

PR #1900, opened 2018-03-09
review chore:Use Maven INRIA instead of Gforge to deploy snapshots

surli:
Fix #1900 

⚠️ This PR might breaks for some time the snapshot deployment: I should have fixed it in the CI but I did not tested it, as I need this changes in the pom.xml file. Nervertheless I already deployed manually a first snapshot in order to avoid complete service discontinuity.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -169,11 +169,6 @@
       <url>http://spoon.gforge.inria.fr/repositories/releases</url>
     </repository>
     <repository>
-      <id>gforge.inria.fr-snapshot</id>
-      <name>Maven Repository for Spoon Snapshots</name>
-      <url>http://spoon.gforge.inria.fr/repositories/snapshots</url>
-    </repository>
-    <repository>
       <id>maven.inria.fr-snapshot</id>
       <name>Maven Repository for Spoon Snapshots</name>
       <url>http://maven.inria.fr/artifactory/repo/spoon-public-snapshot</url>
--- a/pom.xml
+++ b/pom.xml
@@ -173,6 +173,11 @@
       <name>Maven Repository for Spoon Snapshots</name>
       <url>http://spoon.gforge.inria.fr/repositories/snapshots</url>
     </repository>
+    <repository>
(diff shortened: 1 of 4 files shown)


monperrus:
There are also urls to change in the scripts in /doc

surli:
> There are also urls to change in the scripts in /doc

I think I replaced all mentions now: checked with a grep in the folder. If you know other locations of the mention, tell me.

pom.xml
```diff
@@ -169,9 +169,9 @@
       <url>http://spoon.gforge.inria.fr/repositories/releases</url>
     </repository>
     <repository>
-      <id>gforge.inria.fr-snapshot</id>
+      <id>maven.inria.fr-snapshot</id>
```
monperrus: I guess we can remove the whole `repository` tag since we release on Central, and lonly keep the `snapshotRepository`?

pom.xml
```diff
@@ -181,9 +181,9 @@
       <url>https://oss.sonatype.org/service/local/staging/deploy/maven2/</url>
     </repository>
     <snapshotRepository>
-      <id>gforge.inria.fr-snapshot</id>
-      <name>inria-snapshots</name>
-      <url>scp://scm.gforge.inria.fr/home/groups/spoon/htdocs/repositories/snapshots</url>
+      <id>maven.inria.fr-snapshot</id>
+      <name>http://maven.inria.fr-snapshots</name>
```
monperrus: What about avoiding `http://`, it may be confusing, looking like a URL and not a name.

pom.xml
surli: I suppose we can indeed. I just kept the same pattern as before.

pom.xml
surli: Ok I'll change it.

monperrus:
I guess we have to update the Jenkins job? (at least to give the crendentials?)

surli:
> I guess we have to update the Jenkins job? (at least to give the crendentials?)

Yes, I'm on it.

PR #1903, opened 2018-03-10
fix: fix a bad exception in DJPP

monperrus:
fix bad exception in DJPP after heavy transformation, no way to reproduce in a test case.

### DIFF
--- a/src/main/java/spoon/reflect/visitor/filter/SiblingsFunction.java
+++ b/src/main/java/spoon/reflect/visitor/filter/SiblingsFunction.java
@@ -73,7 +73,7 @@ public class SiblingsFunction implements CtConsumableFunction<CtElement> {
 			boolean visitNext = mode == Mode.ALL || mode == Mode.NEXT;
 			@Override
 			public void scan(CtElement element) {
-				if (element != null && element.getParent() == parent) {
+				if (element != null && element.isParentInitialized() && element.getParent() == parent) {
 					//visit only elements whose parent is same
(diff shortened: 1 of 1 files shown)


pvojtechovsky:
This PR looks like a workaround of some deeper problem. So I would say it is not good idea to merge it.

It looks like that heavy transformation produced an inconsistent spoon model, where an element (parent) has some children, but these children doesn't point back to that parent. Their parent is uninitialized.

May be 
1) run a spoon model consistency check
2) found which element is invalid and which code made it
3) fix the spoon code which allows producing such inconsistencies

WDYT?

monperrus:
The transformation code was working very well with previous versions of Spoon, so it can be considered as a regression. As for all regressions, I prefer to fix it in Spoon's core and not in the client code.

It's a semi-workaround, the alternative fix would be to identify the intercession method that did not set the parent, but I cannot find it.

This PR fixes the regression.

pvojtechovsky:
this PR is fixing secondary problem.

>  to identify the intercession method that did not set the parent,

that would be fix of the primary problem

monperrus:
agree, hope we'll find the primary problem at some point!

(actually, I'm surprised because we have a strong test on whether setters set the parent)

pvojtechovsky:
> unless getStatements() returns a super-list similar to those you have in the role part.

yes, that would make sense.
In future I would like to refactor whole Spoon model by way all the multivalue properties are read/write and automatically binds new items with parent and sends change events - like "in the role part"

monperrus:
agree, see https://github.com/INRIA/spoon/issues/1916

PR #1904, opened 2018-03-10
fix: all unsettable properties are derived too

pvojtechovsky:
Examples are CtTypeReference#MODIFIER and COMMENT. Before they were derived==false unsettable==true. Now they are derived==true unsettable==true

WDYT?


monperrus:
I agree, we have "unsettable implies derived" (but not the opposite, because we have derived setters).

do we check "unsettable implies derived" in a test ?

pvojtechovsky:
> do we check "unsettable implies derived" in a test ?

no. But how to do it?

It is interesting question, how to test that metamodel stuff at all...? It is somehow tested by other tests which are based on metamodel ... but it is secondary test. I have no idea how to write a primary test without increasing maintenance too much - by duplicate definition of metamodel data in test.

monperrus:
> It is interesting question, how to test that metamodel stuff at all...?

Simply like this?
```
for all metamodel classes:
   if a setter is @Unsettable then the corresponding getter is @Derived
```

pvojtechovsky:
Tests was added. It is finished form my point of view

monperrus:
thanks!

PR #1905, opened 2018-03-10
WIP? add support for incremental build

Egor18:
Moved from https://github.com/INRIA/spoon/pull/1816

Hi, this is my new implementation of the incremental build.

A few words about performance.
On my machine spoon model for RxJava project builds in about 15 minutes, but I can get cached model in less than 1 minute. So, on some projects cache itself could be quite useful.
I added `BufferedOutputStream` and `BufferedInputStream` to `SerializationModelStreamer` because they drastically increase model serialization/deserialization speed (dozens of times or even more).
However, it's possible that reference resolution part for _incremental_ build could be rewritten in a more effective way (as I mentioned [here](https://github.com/INRIA/spoon/pull/1816#issuecomment-371052913)).

So what do you think about this incremental launcher?


src/main/java/spoon/Launcher.java
```diff
@@ -75,7 +75,7 @@
 
 	public static final String OUTPUTDIR = "spooned";
 
-	private final Factory factory;
+	protected Factory factory;
```
msteinbeck: Is it really necessary to make the factory protected?

Egor18: I can replace that with `super(oldFactory)`, but I would like to avoid undesirable call of `processArguments()` in the `Launcher` constructor. Alternatively, I can add another protected constructor to the `Launcher`, which does not call `processArguments()`.

surli:
I'm not really sure why you created a new Launcher for that feature: why not creating a class managing the serialization that you can use in the current Launcher if a specific option is given? And maybe adding a new property in environment to manage the cache directory. 

My point is your new `IncrementalLauncher` is breaking the current behaviour of Launcher: what if I create an `IncrementalLauncher` and add resources in the environment after it?

monperrus:
Before adding new code, could you, in a first PR, remove the code related to the old and broken 'build-outdated' feature? Thanks!

Egor18:
@surli Ok, so what API do you suggest?
Something like that?
```
launcher.addInputResources(....);
launcher.getEnvironment.setClasspath(....);
launcher.getEnvironment().setCacheEnabled(true);
launcher.getEnvironment().setCacheDir(....);
launcher.setIncremental(true);
launcher.build();
```
I think, I can refactor it this way.

surli:
Actually you could even put the `setIncremental` inside the environment: we might need this value in other classes for specific treatments. 

And isn't the cache directory mandatory in case of incremental build? 

So I suggest this kind of API then:

```java
launcher.addInputResources(....);
launcher.getEnvironment.setClasspath(....);
launcher.getEnvironment().setIncremental(true);
launcher.getEnvironment().setCacheDir(....); // optional: a default value should be available
launcher.build();
```

WDYT?

Egor18:
Ok. I'll try to implement this API.

surli:
Any news on this one @Egor18 ?

Egor18:
@surli I'll do that as soon, as I can.

mehdi-kaytoue:
Any news? I am very interested in this feature, it looks quite related to https://github.com/INRIA/spoon/issues/1983

Would be happy to help

Cheers,
Mehdi

monperrus:
just restarted Travis

monperrus:

PR #1907, opened 2018-03-11
review: feature: SpoonMetaModel can be build without spoon sources - using shadow classes

pvojtechovsky:


pvojtechovsky:
This PR prepares metamodel code for move from test to spoon runtime. It is needed in runtime by new `Pattern`, and many other new nice features...

PR #1913, opened 2018-03-15
review refactor: Remove old mechanism "buildOnlyOutdatedFiles" as it was broken.

surli:
This mechanism aims at be replacing by the PR #1905 It's marked as breaking as it changes the interfaces.

### DIFF
--- a/src/main/java/spoon/Launcher.java
+++ b/src/main/java/spoon/Launcher.java
@@ -346,17 +346,6 @@ public class Launcher implements SpoonAPI {
 			sw1.setDefault("false");
 			jsap.registerParameter(sw1);
 
-			// Enable building only outdated files
-			sw1 = new Switch("buildOnlyOutdatedFiles");
-			sw1.setLongFlag("buildOnlyOutdatedFiles");
-			sw1.setHelp(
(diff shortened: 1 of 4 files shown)


monperrus:
ok for me, thanks.

monperrus:
For backward compatibility, we should keep the method signatures `buildUnits(JDTBuilder, SpoonFolder, .Array, String, boolean)` and `#buildUnitsAndModel(JDTBuilder, SpoonFolder, .Array, String, boolean)` even if the last parameter is useless. And we document this in the javadoc.

Otherwise, I'm pretty sure we'll break somebody.

surli:
> Otherwise, I'm pretty sure we'll break somebody. 

I can push them back with a deprecated

monperrus:
we don't even need to deprecate, as long as we document properly.
(so many people don't see the deprecation warnings).

surli:
> (so many people don't see the deprecation warnings).

That's not a good reason why we should not use it: it seems irrelevant to maintain methods that have dead arguments in the future! So even for us it's useful to have the deprecated information. It does not necessarily means that we have to delete the methods in the next release....

monperrus:
let's deprecate then!

monperrus:
thanks @surli !

PR #1914, opened 2018-03-15
WiP fix: Manage generics from Interface in shadow mode

### LINKED ISSUE #1906: superInterfaces on shadow type miss actual type arguments
I am refactoring SpoonMetaModel by way it can be created without spoon sources - just using spoon runtime reflection. And I have found this problem. For example CtConditionalImpl is defined like this
```java
class CtConditionalImpl<T> implements CtConditional<T>
````

but when I get shadow CtClass of CtConditionalImpl, then getSuperInterfaces() returns CtTypeReference without actuall type arguments. So it looks like this
```java
class CtConditionalImpl<T> implements CtConditional
````
I need that actual type argument on interface references, in ClassTypingContext.

This information was not available until `Class#getAnnotatedInterfaces()` was introduced in Java8. Could you please use that new method to deliver actual type arguments of extended interfaces too?

I created a branch with a failing test https://github.com/pvojtechovsky/spoon/tree/fixSuperIfaceArgs

### PR DESCRIPTION
**Pull request by surli**
Fix #1906

### DIFF
--- a/src/main/java/spoon/support/visitor/java/JavaReflectionVisitorImpl.java
+++ b/src/main/java/spoon/support/visitor/java/JavaReflectionVisitorImpl.java
@@ -356,6 +356,10 @@ class JavaReflectionVisitorImpl implements JavaReflectionVisitor {
 		if (type.getEnclosingClass() != null) {
 			visitClassReference(type.getEnclosingClass());
 		}
+
+		for (TypeVariable<Class<T>> generic : type.getTypeParameters()) {
+			visitTypeParameter(generic);
+		}
 	}
 
 	private <T> List<RtMethod> getDeclaredMethods(Class<T> clazz) {
--- a/src/main/java/spoon/reflect/factory/TypeFactory.java
+++ b/src/main/java/spoon/reflect/factory/TypeFactory.java
@@ -45,6 +45,7 @@ import spoon.support.visitor.MethodTypingContext;
 import spoon.support.visitor.java.JavaReflectionTreeBuilder;
 
 import java.lang.annotation.Annotation;
+import java.lang.reflect.TypeVariable;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -314,19 +315,31 @@ public class TypeFactory extends SubFactory {
 		return array;
 	}
 
+	public <T> CtTypeReference<T> createReference(Class<T> type) {
+		return createReference(type, false);
+	}
+
 	/**
 	 * Creates a reference to a simple type
 	 */
-	public <T> CtTypeReference<T> createReference(Class<T> type) {
+	public <T> CtTypeReference<T> createReference(Class<T> type, boolean includingFormalTypeParameter) {
 		if (type == null) {
 			return null;
 		}
 		if (type.isArray()) {
 			CtArrayTypeReference<T> array = factory.Core().createArrayTypeReference();
-			array.setComponentType(createReference(type.getComponentType()));
+			array.setComponentType(createReference(type.getComponentType(), includingFormalTypeParameter));
 			return array;
 		}
-		return createReference(type.getName());
+		CtTypeReference typeReference = createReference(type.getName());
+
+		if (includingFormalTypeParameter) {
+			for (TypeVariable<Class<T>> generic : type.getTypeParameters()) {
+				typeReference.addActualTypeArgument(createReference(generic.getName()));
+			}
+		}
+
+		return typeReference;
 	}
 
 	/**
--- a/src/main/java/spoon/support/reflect/reference/CtTypeReferenceImpl.java
+++ b/src/main/java/spoon/support/reflect/reference/CtTypeReferenceImpl.java
@@ -516,7 +516,7 @@ public class CtTypeReferenceImpl<T> extends CtReferenceImpl implements CtTypeRef
 			if ((sis != null) && (sis.length > 0)) {
 				Set<CtTypeReference<?>> set = new QualifiedNameBasedSortedSet<CtTypeReference<?>>();
 				for (Class<?> si : sis) {
-					set.add(getFactory().Type().createReference(si));
+					set.add(getFactory().Type().createReference(si, true));
 				}
 				return set;
 			}
--- a/src/main/java/spoon/support/visitor/java/internal/TypeReferenceRuntimeBuilderContext.java
+++ b/src/main/java/spoon/support/visitor/java/internal/TypeReferenceRuntimeBuilderContext.java
@@ -18,6 +18,7 @@ package spoon.support.visitor.java.internal;
 
 import spoon.reflect.declaration.CtAnnotation;
 import spoon.reflect.declaration.CtPackage;
+import spoon.reflect.declaration.CtTypeParameter;
 import spoon.reflect.reference.CtTypeParameterReference;
 import spoon.reflect.reference.CtTypeReference;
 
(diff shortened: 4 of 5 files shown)

### CODE REVIEW

**Comment by monperrus** [comment]
LGTM!

**Comment by pvojtechovsky** [comment]
I have tried/tested this code with #1907 and there is still some problem. I have improved the test no so the problem is reproducible here.
Anyway many thanks @surli for this PR it is on good way and it lets me continue with other PRs.

**Comment by surli** [comment]
> I have tried/tested this code with #1907 and there is still some problem

Great, thank you, I was not quite confident on my changes ;)

**Comment by pvojtechovsky** [comment]
This wrong code is generated by this PR
```java
public abstract interface CtField<T extends java.lang.Object> 
   extends spoon.reflect.code.CtRHSReceiver<A> , 
    spoon.reflect.declaration.CtShadowable , 
   spoon.reflect.declaration.CtTypeMember , 
    spoon.reflect.declaration.CtVariable<T> { ... }
```
The problem is 
`CtRHSReceiver<A>`
while it should be
`CtRHSReceiver<T>`

I have extended the test to check that situation

pvojtechovsky:
There is some problem with qualified names of interfaces. I have added failing test. For example CtExpression looks like this:
```java
public abstract interface CtExpression<T extends java.lang.Object> extends spoon.reflect.code.CtCodeElement , spoon.reflect.declaration.spoon.reflect.declaration.CtTypedElement<T> , spoon.template.spoon.template.TemplateParameter<T> {
```
where
`spoon.reflect.declaration.spoon.reflect.declaration.CtTypedElement`
should be
`spoon.reflect.declaration.CtTypedElement`

pvojtechovsky:
I have found two more problems:
* CtTypeReference#getSuperInterfaces() returns correctly initialized CtTypeReferences with actual type arguments, but these CtTypeReferences has no parent CtType, so actual type arguments cannot be resolved well. I fixed that problem with last commit
* CtType#getSuperClass() returns CtTypeReferences without actual type arguments. I have added the failing test. Could you please fix it?

pvojtechovsky:
Next problem is with bounding type of actual type arguments of references of parameter types and return types of methods. See failing test.

I am not sure whether it belongs to this PR. May be we merge this one and fix the problem reproduced by last test case in next PR ... but may be it is quite related ... it is up to you, to decide ;-)

By the way, I am really not sure that current behavior of non shadow type (from sources) checked by first part of the test is wanted (see comments in test). For me the shadow class behaves more correct ... but the change would probably destroy many things... hard to estimate...

surli:
For me it should be fixed in this PR: we are trying here to align the behaviour between elements coming from sources and shadow elements, and obviously concerning the generic types, shadow classes are currently badly supported. 

Now I don't agree with your comment on the last test:
> //is it really correct to have bounding type T?
//There should be NO bounding type - may be a special AST node?

You were talking about the type `A` of `setAssignment` with this signature:
```java
public <C extends CtRHSReceiver<A>> C setAssignment(CtExpression<A> assignment)
```

and `A` is defined above on the class signature:
```java
public class CtAssignmentImpl<T, A extends T> extends CtStatementImpl implements CtAssignment<T, A>
```

so for me it shoud really use the bounding type T which should itself be bounding only by object.
Moreover, we normally use the same bounding definition as coming from JDT for source-coming elements, so the definition should be right.

src/main/java/spoon/support/visitor/java/JavaReflectionVisitorImpl.java
```diff
@@ -43,9 +43,9 @@ public void visitPackage(Package aPackage) {
 			clazz.getPackage();
 		}
 		if (clazz.getSuperclass() != null) {
```
tdurieux: is it ok that the check is not on `getGenericSuperclass`?

surli: Yeah I certainly should change that one for safety.

pvojtechovsky:
I have added a test which compares shadow types and normal types. So it should report all remaining problems. 
This test reports some real problems, but note that it is possible that this test is buggy, so after you fix real problems, may be the test has to be fixed too ;-)

src/main/java/spoon/support/reflect/reference/CtTypeParameterReferenceImpl.java
```diff
@@ -204,12 +204,21 @@ public CtTypeParameter getDeclaration() {
 
 		CtElement e = this;
 		CtElement parent = getParent();
+
 		if (parent instanceof CtTypeReference) {
-			if (parent.isParentInitialized() == false) {
-				return null;
+			if (!parent.isParentInitialized()) {
+				// we might enter in that case because of a call
+				// of getSuperInterfaces() for example
```
pvojtechovsky: This case should not happen after my change in `CtTypeReference#getSuperInterfaces`. Let's rollback this change. WDYT?

surli: I just checked and I have the following broken test if I remove it: GenericsTest#testTypeParameterReferenceAsActualTypeArgument.

surli:
@pvojtechovsky could you have a look on your test? There's still errors in the diff but somehow I did not manage to debug them (it takes a very long time) :-/ Apparently the trick you used to bypass Override annotation did not work, so I changed it a bit but I'm not sure it's working. 

I think most of the work is done here, but the test should be improved: it's hard to read and fix the diff without debug.

pvojtechovsky:
You are right. There was a bug in the test. I fixed it.

Now there remains probably the last problem (thank you for fixing all others  👍  ) in the interface `CtConditional` there is method `clone`
Shadow: 
```java
public volatile default java.lang.Object clone();
```
Normal:
```java
spoon.reflect.code.CtConditional<T> clone();
```

pvojtechovsky:
I just fixed these problems (all detected by existing `testShadowModelEqualsNormalModel`):
* wrong generic types were used when method generic type hides type generic type e.g. here:
```java
class CtLocalVariableImpl<T> {
<T extends CtModifiable> T setExtendedModifiers(Set<CtExtendedModifier> extendedModifiers) {}
}
```
* The synthetic methods of interface were not filtered out. E.g. CtConditional#clone

I have fixed test
* EqualsChecker doesn't fail when element or other is null 
* it ignores annotation `SuppressWarnings`, which has source retention too.

But I do not know that code enough to fix remaining problem. May be it is easy to fix, but not for me.... The **simple name** of CtTypeReference is "spoon.reflect.reference.CtTypeReference<?>"
It is easy to debug this problem:
* I have added this debugging code into `CtReferenceImpl`
	public <T extends CtReference> T setSimpleName(String simplename) {
		if ("spoon.reflect.reference.CtTypeReference<?>".equals(simplename)) {
			this.getClass();
		}
* put breakpoint to this.getClass() and debug the `testShadowModelEqualsNormalModel`

Now it is yours @surli

pvojtechovsky:
Do we have some clients which persist on API of JavaReflectionTreeBuilder? Or in different words: Can I refactor that even more, so it is more readable and maintanable?
I have mainly problem with 3 ways of handling of `Class reference`:
1) for references of super class
2) for references of super interfaces
3) for other references to a class
These 3 ways causes copied code and causes hard readability of the code...

surli:
Thanks @pvojtechovsky to take back this one. Sorry I'm a lot busy elsewhere these days :-/

monperrus:
we have a conflict here. rebase?

pvojtechovsky:
Merge done, but we should review it after #1998, which is contained in this PR too ... so another merge will be needed

monperrus:
awesome collaborative work, thanks to all!

tdurieux:
This would be a nice reflection library!

Issue #1916, opened 2018-03-15
feature: collection-based property on steroids


monperrus:
In Spoon, collection-based properties (ie getters than return set or list of ctelements) are mutable.

This poses two problems:

* the parent is not set
* the change events are not sent

By using specific implementations of lists, we could get this fixed while remaining backward compatible.

@pvojtechovsky has already done in the heavy-lifting in the role-related classes.

pvojtechovsky:
This is  somehow duplicate to #1633. Both issues has same solution - to use dynamic lists and sets

monperrus:
duplicate of #1633

Issue #1633, opened 2017-10-21
feature: better collection-based properties of Spoon model (non-derived getters)


pvojtechovsky:
Example of problem: Use case: Client needs to add argument at the beginning of CtInvocation#arguments list.
Actually the only solution is
```java
List<CtExpression<?>> args = new ArrayList<>(invocation.getArguments());
args.add(0, newArgument);
invocation.setArguments(args);
```

Solutions?
> S1) to have method CtInvocation#addArgument(int, CtExpression)
> S2) to have method CtInvocation#addFirstArgument(int, CtExpression)

it is the way how Spoon handles it actually. Each element implements (or does not implement (... yet?)) own set of collection manipulation methods. May be we might generate all possible add/addList/remove/contains methods of Set and List interface with appropriate attribute name suffix automatically.
It would be compatible with current model, but 
* it would mean a lot of methods in spoon API - more complicated for clients
* it avoids usage of generic helper methods, which are able to do operations on List and Set

> S3) to have method `List CtInvocation#getArguments()`, which would return modifiable implementation of `List`.

We cannot of course return internal list and let client to modify it. 
The idea is to return a List/Set wrapper which would handle all the List/Set method calls. It would
* generate appropriate model change events 
* handle consistency of model (change parent, ...)

It would be not backward compatible, but the code of spoon and code of client's would be simpler and spoon API would be more friendly to clients who needs model modifications.

> S4) to use generic `List modifiableList = invocation.getValueByRoleAsList(CtRole.ARGUMENT)`

This is already partially implemented in  #1582.
* I like that API, because it will be useful for some special kind of model operations.
* But it is not so nice solution like S3 would be, because client has to access attributes by CtRole.ARGUMENT, which is more complicated then simple call of `List getArguments()`
* Current implementation of #1582 is based on current spoon model (`get/setArguments(...)`, ...). If S3 would be already implemented in spoon model, then implementation of #1582 would be much simpler too.

monperrus:
closed per #1582

PR #1917, opened 2018-03-15
review: refactor: mutable collections of Spoon model handles parent and fire change events

pvojtechovsky:
A prototype of solution for #1916 and #1633
... at the beginning only for CtBlock#statements
... later the same solution should be applied to each List, Set and Map of Spoon model

WDYT?

### DIFF
--- a/src/main/java/spoon/support/reflect/code/CtBlockImpl.java
+++ b/src/main/java/spoon/support/reflect/code/CtBlockImpl.java
@@ -22,16 +22,15 @@ import spoon.reflect.code.CtInvocation;
 import spoon.reflect.code.CtStatement;
 import spoon.reflect.code.CtStatementList;
 import spoon.reflect.declaration.CtConstructor;
+import spoon.reflect.declaration.CtElement;
 import spoon.reflect.declaration.ParentNotInitializedException;
 import spoon.reflect.path.CtRole;
 import spoon.reflect.visitor.CtVisitor;
 import spoon.reflect.visitor.Filter;
 import spoon.reflect.visitor.Query;
-import spoon.support.reflect.declaration.CtElementImpl;
-import spoon.support.util.EmptyIterator;
+import spoon.support.util.ModelList;
 
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 
@@ -42,7 +41,27 @@ public class CtBlockImpl<R> extends CtStatementImpl implements CtBlock<R> {
 	private static final long serialVersionUID = 1L;
 
 	@MetamodelPropertyField(role = CtRole.STATEMENT)
-	private List<CtStatement> statements = emptyList();
+	private final ModelList<CtStatement> statements = new ModelList<CtStatement>() {
+		private static final long serialVersionUID = 1L;
+		@Override
+		protected CtElement getOwner() {
+			return CtBlockImpl.this;
+		}
+		@Override
+		protected CtRole getRole() {
+			return STATEMENT;
+		}
+		@Override
+		protected int getDefaultCapacity() {
+			return BLOCK_STATEMENTS_CONTAINER_DEFAULT_CAPACITY;
+		}
+		@Override
+		protected void onSizeChanged(int newSize) {
+			if (isImplicit() && (newSize > 1 || newSize == 0)) {
+				setImplicit(false);
+			}
+		}
+	};
 
 	public void accept(CtVisitor visitor) {
 		visitor.visitCtBlock(this);
@@ -50,7 +69,6 @@ public class CtBlockImpl<R> extends CtStatementImpl implements CtBlock<R> {
 
 	@Override
 	public List<CtStatement> getStatements() {
-		ensureModifiableStatementsList();
 		return this.statements;
 	}
 
@@ -86,14 +104,9 @@ public class CtBlockImpl<R> extends CtStatementImpl implements CtBlock<R> {
 			getStatements().get(0).insertAfter(statements);
 			return (T) this;
 		}
-		ensureModifiableStatementsList();
-		for (CtStatement statement : statements.getStatements()) {
-			statement.setParent(this);
-			this.addStatement(0, statement);
-		}
-		if (isImplicit() && this.statements.size() > 1) {
-			setImplicit(false);
-		}
+		List<CtStatement> copy = new ArrayList<>(statements.getStatements());
+		statements.setStatements(null);
+		this.statements.addAll(0, copy);
 		return (T) this;
 	}
 
@@ -103,28 +116,23 @@ public class CtBlockImpl<R> extends CtStatementImpl implements CtBlock<R> {
 			getStatements().get(0).insertAfter(statement);
 			return (T) this;
 		}
-		ensureModifiableStatementsList();
-		statement.setParent(this);
-		this.addStatement(0, statement);
-
-		if (isImplicit() && this.statements.size() > 1) {
-			setImplicit(false);
-		}
+		this.statements.add(0, statement);
 		return (T) this;
 	}
 
 	@Override
 	public <T extends CtStatementList> T insertEnd(CtStatement statement) {
-		ensureModifiableStatementsList();
 		addStatement(statement);
 		return (T) this;
 	}
 
 	@Override
 	public <T extends CtStatementList> T insertEnd(CtStatementList statements) {
-		for (CtStatement s : statements.getStatements()) {
-			insertEnd(s);
-		}
+		List<CtStatement> tobeInserted = new ArrayList<>(statements.getStatements());
+		//remove statements from the `statementsToBeInserted` before they are added to spoon model
+		//note: one element MUST NOT be part of two models.
+		statements.setStatements(null);
+		this.statements.addAll(this.statements.size(), tobeInserted);
 		return (T) this;
 	}
 
@@ -162,80 +170,31 @@ public class CtBlockImpl<R> extends CtStatementImpl implements CtBlock<R> {
 
 	@Override
 	public <T extends CtStatementList> T setStatements(List<CtStatement> statements) {
-		if (statements == null || statements.isEmpty()) {
-			this.statements = CtElementImpl.emptyList();
-			return (T) this;
-		}
-		getFactory().getEnvironment().getModelChangeListener().onListDeleteAll(this, STATEMENT, this.statements, new ArrayList<>(this.statements));
-		this.statements.clear();
-		for (CtStatement s : statements) {
-			addStatement(s);
-		}
+		this.statements.set(statements);
 		return (T) this;
 	}
 
 	@Override
 	public <T extends CtStatementList> T addStatement(CtStatement statement) {
-		return this.addStatement(this.statements.size(), statement);
+		this.statements.add(statement);
+		return (T) this;
 	}
 
 	@Override
 	public <T extends CtStatementList> T addStatement(int index, CtStatement statement) {
-		if (statement == null) {
-			return (T) this;
-		}
-		ensureModifiableStatementsList();
-		statement.setParent(this);
-		getFactory().getEnvironment().getModelChangeListener().onListAdd(this, STATEMENT, this.statements, index, statement);
 		this.statements.add(index, statement);
-		if (isImplicit() && this.statements.size() > 1) {
-			setImplicit(false);
-		}
 		return (T) this;
 	}
 
-	private void ensureModifiableStatementsList() {
-		if (this.statements == CtElementImpl.<CtStatement>emptyList()) {
-			this.statements = new ArrayList<>(BLOCK_STATEMENTS_CONTAINER_DEFAULT_CAPACITY);
-		}
-	}
 
 	@Override
 	public void removeStatement(CtStatement statement) {
-		if (this.statements != CtElementImpl.<CtStatement>emptyList()) {
-			boolean hasBeenRemoved = false;
-			// we cannot use a remove(statement) as it uses the equals
-			// and a block can have twice exactly the same statement.
-			for (int i = 0; i < this.statements.size(); i++) {
-				if (this.statements.get(i) == statement) {
-					getFactory().getEnvironment().getModelChangeListener().onListDelete(this, STATEMENT, statements, i, statement);
-					this.statements.remove(i);
-					hasBeenRemoved = true;
-					break;
-				}
-			}
-
-			// in case we use it with a statement manually built
-			if (!hasBeenRemoved) {
-				getFactory().getEnvironment().getModelChangeListener().onListDelete(this, STATEMENT, statements, statements.indexOf(statement), statement);
-				this.statements.remove(statement);
-			}
-
-			if (isImplicit() && statements.size() == 0) {
-				setImplicit(false);
-			}
-		}
+		this.statements.remove(statement);
 	}
 
 	@Override
 	public Iterator<CtStatement> iterator() {
-		if (getStatements().isEmpty()) {
-			return EmptyIterator.instance();
-		}
-		// we have to both create a defensive object and an unmodifiable list
-		// with only Collections.unmodifiableList you can modify the defensive object
-		// with only new ArrayList it breaks the encapsulation
-		return Collections.unmodifiableList(new ArrayList<>(getStatements())).iterator();
+		return this.statements.iterator();
 	}
 
 	@Override
--- a/src/main/java/spoon/support/util/ModelList.java
+++ b/src/main/java/spoon/support/util/ModelList.java
@@ -1,0 +1,238 @@
+/**
+ * Copyright (C) 2006-2017 INRIA and contributors
+ * Spoon - http://spoon.gforge.inria.fr/
+ *
+ * This software is governed by the CeCILL-C License under French law and
+ * abiding by the rules of distribution of free software. You can use, modify
+ * and/or redistribute the software under the terms of the CeCILL-C license as
+ * circulated by CEA, CNRS and INRIA at http://www.cecill.info.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the CeCILL-C License for more details.
+ *
+ * The fact that you are presently reading this means that you have had
+ * knowledge of the CeCILL-C license and that you accept its terms.
+ */
+package spoon.support.util;
+
+
+import java.io.Serializable;
+import java.util.AbstractList;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.ConcurrentModificationException;
+import java.util.List;
+
+import spoon.experimental.modelobs.FineModelChangeListener;
+import spoon.reflect.declaration.CtElement;
+import spoon.reflect.path.CtRole;
+import spoon.support.reflect.declaration.CtElementImpl;
+
+/**
+ * The implementation of the {@link List}, which is used by Spoon model objects.
+ * It assures:
+ * 1) each inserted {@link CtElement} gets assigned correct parent
+ * 2) each change is reported in {@link FineModelChangeListener}
+ */
+public abstract class ModelList<T extends CtElement> extends AbstractList<T> implements Serializable {
+	private static final long serialVersionUID = 1L;
+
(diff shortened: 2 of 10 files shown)


src/main/java/spoon/support/util/ModelList.java
```diff
@@ -0,0 +1,192 @@
...
+	public int indexOf(Object o) {
+		return list.indexOf(o);
+	}
+
+	@Override
+	public int lastIndexOf(Object o) {
+		return list.lastIndexOf(o);
+	}
+
+	private void ensureModifiableStatementsList() {
```
surli: This method should be changed to be generic

src/main/java/spoon/support/reflect/code/CtBlockImpl.java
```diff
@@ -42,15 +39,29 @@
...
+	};
 
 	public void accept(CtVisitor visitor) {
 		visitor.visitCtBlock(this);
 	}
 
 	@Override
 	public List<CtStatement> getStatements() {
-		ensureModifiableStatementsList();
 		return this.statements;
```
surli: Why does this method return a mutable list? It seems on purpose regarding the previous call of `ensureModifiableStatementsList` but it feels wrong...

src/main/java/spoon/support/reflect/code/CtBlockImpl.java
```diff
@@ -163,79 +158,35 @@ private boolean shouldInsertAfterSuper() {
 	@Override
 	public <T extends CtStatementList> T setStatements(List<CtStatement> statements) {
 		if (statements == null || statements.isEmpty()) {
-			this.statements = CtElementImpl.emptyList();
+			this.statements.clear();
```
surli: We already call `clear` below in the method: this could now be simlified by just adding a if not null around the final `addAll`

surli:
There's some minor issues but it looks much more elegant than before ;)

src/main/java/spoon/support/reflect/code/CtBlockImpl.java
pvojtechovsky: It returns a mutable ModelList, which can be mutated using List API and which assures that spoon model is kept consistent (parent-child is kept) and that modification events are sent.

This change is the CORE idea of this PR ;-)

src/main/java/spoon/support/reflect/code/CtBlockImpl.java
pvojtechovsky: you are right

src/main/java/spoon/support/util/ModelList.java
pvojtechovsky: why? It is private and it uses generic argument <T> from ListModel<T> class

surli: Actually no you the name says it's for statements and you have this: `list == CtElementImpl.<CtStatement>emptyList()`

It should use a `T` and have a name like `ensureModifiableList`

pvojtechovsky: Now I see it ;-), Thanks. I will fix it probably today in the evening

src/main/java/spoon/support/reflect/code/CtBlockImpl.java
surli: > This change is the CORE idea of this PR ;-)

Hmmm. It looks a bit dangerous for me: old users won't change their API and won't beneficiate the changes, but it might disturb new users because they won't necessarily expect the API to return a pointer on the model. 

WDYT about keeping a `getStatements` that returns an immutable list, and adding a `getModifiableStatements` like the one you proposed?

pvojtechovsky: Note: that this method already returned modifiable list before. I just assured that this modifiable list behaves correctly.

surli: > that this method already returned modifiable list before

My point is that I believe we should have the same consistent behaviour everywhere in Spoon. And I think that almost everywhere - but maybe not here - this kind of method is currently returning a new ArrayList, detached from the model  (see for example CtClass#getConstructors or CtType#getDeclaredExecutables): so they are mutable, but not linked to the model. If we make those methods returning a mutable element which is now linked to the model, it really can bring unexpected behaviours for the clients. 

That's why I suggest to start here to fix a definitive behaviour that we'll use everywhere. And I'll definitively prefer to change those methods to return an immutable object and to create another methods which return explicitely a mutable linked to the model: we might break some clients, but the fix is easy and explicit.

pvojtechovsky:
> My point is that I believe we should have the same consistent behaviour everywhere in Spoon.

I understand the problem and agree that spoon should behave consistent. So there are two questions:
1) what to do until all spoon properties are switched to mutable ModelList/ModelSet/ModelMap - can we expect that we are able to change whole spoon model until next spoon release?
2) let's say we changed all properties to mutable and safe ModelList/ModelSet/ModelMap: Do we want 
A) to have each property duplicated like `getStatements` and `getMutableStatements` 
B) or we can simply use only `getStatements` which will return mutable list.

I vote for B), because it ends with lower maintenance, nicer API and good compatibility with old clients. 

Note: there are actually these implementations in spoon model:
A) getter returns immutable collection - OK it is compatible if we return mutable collection now
B) getter returns muttable collection, whose mutation adds values into model, but which does not sets parent and does not call change events - OK it is compatible too, because we just fixed the bug and spoon finally works as it was already expected by clients
C) getter returns detached muttable copy of collection - whose mutations has no influence to model. Here we are incompatible. I am not aware of this code yet, but if there is really some then we need some solution ... the easier is to break compatibility here and to keep nice and simple model API for future.

surli:
Maybe a good way to prepare this is to setup a test architecture to check what we want:

1. Get all methods returning collections and check if the returning object is an immutable or not, then we might have an idea of what we will break in case C
2. Assert that the returning object should be a ModifiableCollection for the future implementation: then we release only when this test will pass

And to take into account the problem C an easy way is to release it as major release and document why. @monperrus won't be happy but it looks like the easiest way to avoid maintaining methods with mutable and immutable.

monperrus:
> Get all methods returning collections and check if the returning object is an immutable or not, then we might have an idea of what we will break in case C

Probably none of them if it's not the empty list.

monperrus:
> And I think that almost everywhere - but maybe not here - this kind of method is currently
> returning a new ArrayList, detached from the model (see for example CtClass#getConstructors or
> CtType#getDeclaredExecutables)

IMHO it's the opposite. It's mostly bound to the model but for a few specific cases such as the ones
you mention.

Anyway, we all agree that this should be 1) consistent and 2) specified and checked in a test.

Let's write the test to measure the quantity of immutable / mutable-detached / mutable-attached
collection getters.

surli:
> Probably none of them if it's not the empty list.

I don't agree. See:

https://github.com/INRIA/spoon/blob/0b45cf90bd380b8cdfc8dd2c2d29cf3f74a445d7/src/main/java/spoon/support/reflect/declaration/CtTypeImpl.java#L589

monperrus:
>
https://github.com/INRIA/spoon/blob/0b45cf90bd380b8cdfc8dd2c2d29cf3f74a445d7/src/main/java/spoon/support/reflect/declaration/CtTypeImpl.java#L589

This is a derived getter. This is one point of confusion, here we are talking about non-derived
getters. (updated my previous comment accordingly)

surli:
> This is one point of confusion, here we are talking about non-derived getters. (updated my previous comment accordingly)

Actually that's interesting: if all getters become mutable, how will they do the distinction between derived ones and the others, attached to the model? Without reading the doc I mean.

```java
CtType myType = // a way to get my type

 // change nothing
myType.getAllFields().add(myNewField);
myType.getAllMethods().add(myNewMethod);

 // change the model and fire events
myType.getAllTypeMembers().add(myNewField);
```

pvojtechovsky:
> Actually that's interesting: if all getters become mutable, how users will do the distinction between derived ones and the others, attached to the model?

interesting question...

I vote to make immutable these methods (derived) whose mutation has no influence to the model. So it fails soon and client is immediately notified that it is not the correct way how to modify the model

monperrus:
> if all getters become mutable
almost all non-derived getters are already mutable :-) (my personal bet!)

the derived ones should be immutable, and we won't break anything serious if we do so.

pvojtechovsky:
The tests are actually failing on this old test code:
```java
    CtBlock<?> body = ...

    for (CtStatement s : body) {
    	body.removeStatement(s);
    }
```
the body contains two statements, but it processes only one of them, because concurrent modification is actually not check correctly in `ModelList`. I can fix it, but the question is how it should behave???

A) it should throw concurrent modification exception - standard java behavior
B) it should create a copy of Array before iterator is created. And the iterator should be unmodifiable
C) it should create a copy of Array before iterator is created. But the iterator should be still modifiable

if A), then we break compatibility of Spoon API
if B), then we limit possibility to modify content of statements using iterator - I would prefer to have new collections fully compatible with java collections
if C), then it tries some compromise between expected java behavior and backward compatibility ... and I am not sure what problems will come later...

I actually vote for A). WDYT?

pvojtechovsky:
I have created a test (see #1922) which checks collection types. Here is the detailed results
[collection_check.txt](https://github.com/INRIA/spoon/files/1822109/collection_check.txt)
... rename it to CSV and let Excel sort it and count it

A summary:
* 18 tested derived collections are all mutable-detached
* 3 collections are mutable attached
  * CtBlock#statement
  * CtPackage#containedType
  * CtPackage#subPackage
* 1 collection is mutable detached
  * CtJavaDoc#commentTag
* 210 collections are read only

Test actually fails on 
* CtIntersectionTypeReference#interface
* CtTypeParameterReference#interface
* CtWildcardReference#interface

pvojtechovsky:
I have fixed remaining non-derived mutable attached/detached collections of spoon model. There were
* CtBlock#statements - was mutable attached
* CTPackage#packs - was mutable attached
* CTPackage#types - was mutable attached
* CtJavaDoc#tags - was mutable detached

So now there should be not possible to create a model with children, which has unitialized parents.

src/main/java/spoon/support/util/ModelList.java
```diff
@@ -0,0 +1,238 @@
...
+	public int indexOf(Object o) {
+		return list.indexOf(o);
+	}
+
+	@Override
+	public int lastIndexOf(Object o) {
+		return list.lastIndexOf(o);
+	}
+
+	private static class InternalList<T> extends ArrayList<T> {
```
surli: Could you document that this internalList purpose is to manage the ConcurrentModificationException through modCount, maybe linking the URL: https://docs.oracle.com/javase/7/docs/api/java/util/AbstractList.html#modCount

It will ease the future maintaining ;) 

And more generally I'm not sure why you update this value only when calling `clear` method: shouldn't you manage also the add/remove cases?

pvojtechovsky: > Could you document

ok, I will do so.

> And more generally I'm not sure why you update modCount only when calling `clear` method

I am directly changing it in `clear` method because this one changes content of internal list, but doesn't calls any method of that internal `list`. In all other cases the internal method of `list` is called so the modCount of that internal list is managed automatically. Then I just call `updateModCount` to read list.modCount and write it into this.modCount.

pvojtechovsky:
This PR now fails on many places because of this `ModelList` code:
```java
	static void linkToParent(CtElement owner, CtElement element) {
		if (owner.getFactory().getEnvironment().checksAreSkipped() == false && element.isParentInitialized() && element.getParent() != owner) {
			//the `e` already has an different parent. Check if it is still linked to that parent
			if (element.getRoleInParent() != null) {
				throw new SpoonException("The element " + element + " is already used by another part of SpoonModel. Remove it from previous model or clone it before.");
			}
		}
		element.setParent(owner);
	}
```
I think we should check whether added element is not already used on some other part of spoon model. Because otherwise you can add one element on two places and you will be not warned that you made invalid model....
But there are some cases when this technique is used by spoon internally ... for example creates helper CtStatementsList to insert some statements ... and then this `linkToParent` reports a problem...
How to solve it?
A) to remove check from `linkToParent`
B) to improve check form `linkToParent` by way it detects whether element is connected to real model - the one which starts with `CtUnnamedModule` and to report error only in that case. When there is just a model fragment, then it is OK.
C) to refactor spoon code to not use this technique or to remove the elements from CtStatementsList before they are inserted into another part of model.
D) ?

WDYT?

src/main/java/spoon/support/util/ModelList.java
surli: Thanks for the explanation ;)

surli:
> How to solve it?
> C) to refactor spoon code to not use this technique or to remove the elements from CtStatementsList before they are inserted into another part of model.

I think we should keep the contract explicit as it avoid unexpected side effects in Spoon model. So first let's try refactoring the current Spoon to assert that, and maybe then - if it makes the code harder to understand/maintain - add a flag to skip the contract in some very specific cases?

pvojtechovsky:
The test `FactoryTest#testIncrementalModel` merges two models together. There are several inconsistencies:
* the produced target model contains elements whose getFactory() returns different values
* the source model becomes inconsistent, because that model still points to a children elements, but getParent of these elements points to different model.
* the model change notifications will not work well in such mixed model
* ...?

@monperrus you seems to be the author. Could you please fix this test somehow?

pvojtechovsky:
Please have a look at `CtPackageAssertTest#testEqualityBetweenTwoCtPackage` test too. It seems to be broken multiple times
* actually assertSame(factory.getModel().getRootPackage();, factory.Package().getOrCreate("")) pass, but it was not so at the time when author made that test
* the `CtPackage#addType` adds type directly into this package. It ignores origin package of the type. So qualified name "spoon.testing.testclasses.Foo" becomes "Foo"

So there is question what is the intended  behavior of `CtPackage#addType`
A) it should move type from it's current package to new package (to change qualified name)
B) it should add the type into package where it actually is

WDYT?

pvojtechovsky:
ping @monperrus, please have a look at failing tests here. I have no idea how to fix them

monperrus:
ok, I will have a look at it.

monperrus:
Good news! when one removes:

```
//			if (element.getRoleInParent() != null) {
//				throw new SpoonException("The element " + element + " is already used by another part of SpoonModel. Remove it from previous model or clone it before.");
//			}
```

1) We restore backward compatibility, see https://github.com/INRIA/spoon/pull/1917/files#r178425580
2) testIncrementalModel passes again.

Double win!

To me that's the way to go.

pvojtechovsky:
> backward compatibility

it is already configurable (see `if(getEnvironment().checksAreSkipped()==false ...`), so if one really needs to use that dirty and unsafe technique to build spoon model, then s/he still can do it using one switch.

But I vote to keep that check by default activated, because it reported REAL problems in spoon and spoon test code.

And even if we comment this code out, then these two tests:
* FactoryTest#testIncrementalModel
* CtPackageAssertTest#testEqualityBetweenTwoCtPackage
are buggy anyway, so they have to fixed ... even if the problem is not so visible after commenting that check ;-)

WDYT?

monperrus:
Why do you think that those tests are buggy?

pvojtechovsky:
See my explanations of the problems in this PR above https://github.com/INRIA/spoon/pull/1917#issuecomment-375071418

pvojtechovsky:
@monperrus, I copied the comments from above so it is easily visible what is the problem:

The test `FactoryTest#testIncrementalModel` merges two models together. Problems:
* the produced target model contains elements whose getFactory() returns different values
* the source model becomes inconsistent, because that model still points to a children elements, but getParent of these elements points to different model.
* the model change notifications will not work well in such mixed model
* ...?

Test `CtPackageAssertTest#testEqualityBetweenTwoCtPackage` has these problems:
* actually assertSame(factory.getModel().getRootPackage();, factory.Package().getOrCreate("")) pass, but it was not so at the time when author made that test
* the `CtPackage#addType` adds type directly into this package. It ignores origin package of the type. So qualified name "spoon.testing.testclasses.Foo" becomes "Foo"

So there is question what is the intended  behavior of `CtPackage#addType`
A) it should move type from it's current package to new package (to change qualified name)
B) it should add the type into package where it actually is

monperrus:
> So there is question what is the intended behavior of CtPackage#addType

I would go for `A) it should move type from it's current package to new package (to change qualified name)`

when can do this first in a separate baby PR.

pvojtechovsky:
> AstCheckerTest line 152: here, AFAIU, we basically discard the check for plenty of nodes, correct? If yes, we should do it directly before the assertions (and not in toBeProcessed) to clearly understand what we change.

It skips assertion for these nodes:
CtBlock#statement;MUTABLE_ATTACHED_CORRECT
CtJavaDoc#commentTag;MUTABLE_ATTACHED_CORRECT
CtPackage#containedType;MUTABLE_ATTACHED_CORRECT
CtPackage#subPackage;MUTABLE_ATTACHED_CORRECT

> ImportTest: is it related to this PR?

no, it is not related. I made #2006 out of that

> RemoveTest

added comment: iterate on copy of list of statements, otherwise it fails with concurrent modification exception

> SignatureTest: what happens there?

`createCodeSnippetStatement` now returns a new element, which is ready to be added into some AST tree. It means that new element is actually not part of any other AST (otherwise consistency constraint will complain). That is the change. Before the returned element was part of the wrapper AST and after adding into some AST there was an inconsistency - new element was a child of two AST nodes.

I do not know why SignatureTest#testLiteralSignature expects that there exist some parent of the new element, ... I do not thing we should support that. So I removed that, because it failed with this PR.

> InsertBlockProcessor: getStatements() ... what does this change mean?
> block.getStatements().add(element.getBody());

the line above causes model inconsistency. The body of element is child of two nodes. So AST is no more tree ... This PR (by default) fails on such operations to warn about that inconsistency early. So test was fixed.

> CtPackageAssertTest: why do lines 19,20 and 21 and 33 are removed?

because they would fail following the contract you agreed in #2000.
Note: `factory.Class().create("spoon.testing.testclasses.Foo")` creates a type, which is already assigned to the package `spoon.testing.testclasses`, whose parents are already root package.
So it makes no sense to re-add that type into root package.

monperrus:
Thanks a lot for the explanation.

AstCheckerTest: why do we need to skip those nodes now?

RemoveTest: why this worked before? and this does not work anymore now?

src/main/java/spoon/support/util/ModelList.java
```diff
@@ -0,0 +1,245 @@
...
+		getModelChangeListener().onListAdd(owner, getRole(), list, index, element);
+		list.set(index, element);
+		updateModCount();
+		return oldElement;
+	}
+
+	static void linkToParent(CtElement owner, CtElement element) {
+		if (owner.getFactory().getEnvironment().checksAreSkipped() == false && element.isParentInitialized() && element.getParent() != owner) {
+			//the `e` already has an different parent. Check if it is still linked to that parent
+			if (element.getRoleInParent() != null) {
```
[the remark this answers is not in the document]
monperrus: This exception was essential for finding errors. But should we keep it now? It will break a lot of client code.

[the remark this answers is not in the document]
pvojtechovsky: it is already configurable here `owner.getFactory().getEnvironment().checksAreSkipped()` so one can switch it off

> break a lot of client code.

It breaks only client code which is already broken, because that code creates invalid AST, which can cause unexpected behavior of spoon and tricky issues.

monperrus:
I still don't feel comfortable with this PR. It seems to me that we are trying to do two different things here:

1. A) a refactoring, as stated in the title of the PR "mutable collections of Spoon model handles parent and fire change events"
1. B) a behavior change related to not being in two parents at the same time. 

Correct?

pvojtechovsky:
yes, correct.

pvojtechovsky:
I moved the consistency check and related fixes to #2009, so this one is simpler now

pvojtechovsky:
AST checker test checks that each Spoon model property modifier (set/add/remove,....) method calls model change notification ... the new implementation with ModelList and ModelSet is now different case, so it must be ignored by this test

remove test iterates over collection and then removes from this collection. Before it passed because it worked on the copy of the collection. Now it would fail because of concurrent modification exception.

monperrus:
Now, I remember, AST checker works by pure static analysis.

We have other tests which check the behavior dynamically (which is better and less fragile), such as
ParentTest.

Do we also have a dynamic test for checking the presence of change events?

monperrus:
Thanks a lot for this step towards #1922

PR #1922, opened 2018-03-17
WIP: test: spoon model collections contract

pvojtechovsky:
As discussed in #1917 here is the test, which checks type of each (most) Spoon model collections.

PR #1923, opened 2018-03-18
fix: derived properties return unmodifiable collections

pvojtechovsky:
As discussed in #1917 and detected by #1922, this PR fixes all main derived properties of Spoon model by way they return unmodifiable collections.

Before they were modifiable, but detached. So the modifications had no influence to the model. Now they are unmodifiable, so client is warn early that modifications makes no sense here.


surli:
I mark this one as breaking as it can be a breaking behaviour in some clients. We'll be able to notice it easily in the changelog then. Thanks @pvojtechovsky !

PR #1924, opened 2018-03-18
Use error-prone compiler; fix bugs and resource leaks

leventov:
See http://errorprone.info

Also I found a problem that `spoon.test.control.Ifs` class is not used.


src/test/java/spoon/test/compilationunit/GetBinaryFilesTest.java
```diff
@@ -115,7 +115,7 @@ public void testNestedTypes() throws IOException {
 
 	@Test
 	public void testAnonymousClasses() throws IOException {
-		final String input = "./src/test/java/spoon/test/secondaryclasses/AnonymousClass.java";
+		final String input = "./src/test/java/spoon/test/secondaryclasses/testclasses/AnonymousClass.java";
```
pvojtechovsky: The origin line was correct. Please check it again

pvojtechovsky: ok, I see. You moved classes into different package. Why?

leventov: To exclude them from analysis succinctly here: https://github.com/INRIA/spoon/pull/1924/files#diff-600376dffeb79835ede4a0b285078036R296

pvojtechovsky:
Looks good to me. I am just not sure with that move of testclasses. Is not there another way how to exclude them from analyzis? E.g. an annotation?
@monperrus @surli

leventov:
Yes there are other ways to suppress errors in testclasses. But moving them all to testclasses/ package actually seems to be the right thing to me. This PR doesn't try to move all of them, though.

surli:
I'm personally ok with moving all test resources in testclasses packages: I think it makes test much more readable.

monperrus:
can we have a test to check those problems automatically in CI?

surli:
> can we have a test to check those problems automatically in CI?

if I'm right it's already the case as the pom.xml is changed to add error-prone, right?

monperrus:
I see. I'm not familiar with errorprone. Can it be activated outside javac directly, such as checkstyle?

leventov:
Probably it could be implemented as a separate "maven-compiler-plugin" run, but why?

monperrus:
because the compiler is the most central component in the build, I don't trust any arbitrary dependency for this.

since it can be done separately, this seems like a better option.

leventov:
error-prone compiler should be viewed as a maintained brunch of OpenJDK's javac, which adds checks, but doesn't change bytecode generation logic. It's usually closer to OpenJDK's javac `head` development state, that the javac that is distributed, for example, with OpenJDK 8 or 9. Sometimes there are bugs in OpenJDK's javac. Considering this, error-prone is actually a slightly more robust option.

monperrus:
updated the PR

- fix CtPathStringBuilder exception message + add test
- temporarily remove the compiler change to merge those valuable changes without having to settle down the question of changing the compiler, we can postpone this decision (see https://stackoverflow.com/questions/49416014/how-to-use-error-prone-in-maven-without-changing-the-compiler)
 
One question:

Why did error-prone make you change this `Object value = annotation.annotationType().getMethod(ctMethod.getSimpleName()).invoke(annotation);`

leventov:
> Why did error-prone make you change this Object value = annotation.annotationType().getMethod(ctMethod.getSimpleName()).invoke(annotation);

See http://errorprone.info/bugpattern/GetClassOnAnnotation. In this case it is irrelevant probably, but to cease the error, I had to make the change.

monperrus:
Thanks a lot for this cleaning PR. 

For the record, if we find a solution to run error-prone without changing the compiler, here is the magic config line:

                 <arg>-XepExcludedPaths:(.*/testclasses.*/.*|.*/Foo.*)</arg>

leventov:
In Druid it's done via a separate profile: https://github.com/druid-io/druid/blob/master/pom.xml#L1136-L1204

PR #1926, opened 2018-03-24
review: test: add new test for shadow enums

Egor18:
Some time ago I added PR (https://github.com/INRIA/spoon/pull/1806) with failing tests, considering shadow classes kind (enum, interface, etc). And @surli fixed the issue.
However, I found out that shadow enums **from interfaces** are still considered as classes, instead of enums.
Can't really say why is this happening, so I provided failing test case.
Thanks.


surli:
Hi @Egor18 thanks for reporting this with a testcase. I did not checked yet, but I think it's closely related to #1914 on which I'm working on. I'll try to take back your tests in #1914 to fix it.

Egor18:
Ok, thanks.

surli:
@Egor18 your test has been fixed with the work done in #1914 
Should be ready for reviewing.

PR #1931, opened 2018-03-27
review: fix class with comments has invalid DeclarationSourcePosition#modifier

pvojtechovsky:
I am experimenting with another approach to implement Sniper mode and have found a bug:
Class with sources like this
```java
public class /*comment*/ Test {}
```
has wrong DeclarationSourcePosition#modifier end.

The problem is in code 
/spoon-core/src/main/java/spoon/support/compiler/jdt/PositionBuilder.java (line 137)
because it doesn't count with comments. So we need some comment aware parser/tokenizer here.

I guess, I can fix it. But if anybody else would like to implement that comment parser, I would gladly do something else ;-)

### DIFF
--- a/src/test/java/spoon/test/position/PositionTest.java
+++ b/src/test/java/spoon/test/position/PositionTest.java
@@ -22,6 +22,7 @@ import spoon.test.position.testclasses.FooAbstractMethod;
 import spoon.test.position.testclasses.FooAnnotation;
 import spoon.test.position.testclasses.FooClazz;
 import spoon.test.position.testclasses.FooClazz2;
+import spoon.test.position.testclasses.FooClazzWithComments;
 import spoon.test.position.testclasses.FooField;
 import spoon.test.position.testclasses.FooGeneric;
 import spoon.test.position.testclasses.FooInterface;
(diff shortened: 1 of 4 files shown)


pvojtechovsky:
this PR contains another fix related to source position too. The type reference: `List<T>` had a source position which included only `List`, The `<T>` was missing.
But `List<T>[]` already included everything. It looks like JDT bug.
This PR fixes source position of `List<T>` so it is now `List<T>`. 
The comments are handled well in both fixes - therefore I joined them together...
List<T> field`

tdurieux:
Thank you for this fix!

src/test/java/spoon/test/position/PositionTest.java
```diff
@@ -77,6 +77,65 @@ public void testPositionClass() throws Exception {
 		assertEquals("public", contentAtPosition(classContent, position.getModifierSourceStart(), position.getModifierSourceEnd()));
 	}
 	
+	
+	@Test
+	public void testPositionClassWithComments() throws Exception {
+		final Factory build = build(new File("src/test/java/spoon/test/position/testclasses/"));
```
monperrus: contract?

src/test/java/spoon/test/position/PositionTest.java
```diff
@@ -77,6 +77,65 @@ public void testPositionClass() throws Exception {
...
+
+		assertEquals("FooClazzWithComments", contentAtPosition(classContent, position.getNameStart(), position.getNameEnd()));
+		assertEquals("/*c1*/\n" + 
+				"//lc1\n" + 
+				"public", contentAtPosition(classContent, position.getModifierSourceStart(), position.getModifierSourceEnd()));
+	}
+
+	@Test
+	public void testPositionParameterTypeReference() throws Exception {
+		//contract: the parameterized type refernce has a source position which includes parameter types, etc.
```
monperrus: typo in refernce

monperrus:
seems like a serious series of changes about positions :-)

pvojtechovsky:
> seems like a serious series of changes about positions :-)

the bugs becomes visible in the light of my experimenting with the sniper mode.

This is ready for merge too.

PR #1932, opened 2018-03-29
fix IndexOutOfBoundsException in getType()

Egor18:
Hi! I found out that `IndexOutOfBoundsException` is occurred in cases when `getType()` is called on parameter of a catch with multiple exception types, some of which is unknown (noclasspath mode):
```
java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
	at java.util.ArrayList.rangeCheck(ArrayList.java:657)
	at java.util.ArrayList.get(ArrayList.java:433)
	at spoon.support.reflect.code.CtCatchVariableImpl.getType(CtCatchVariableImpl.java:109)
        ....
```

I think it would be better to return null from getType() in such cases instead, so I provided this PR.

### DIFF
--- a/src/test/java/spoon/test/trycatch/TryCatchTest.java
+++ b/src/test/java/spoon/test/trycatch/TryCatchTest.java
@@ -218,7 +218,7 @@ public class TryCatchTest {
 		CtTry tryStmt = (CtTry) clazz.getElements(new TypeFilter<>(CtTry.class)).get(0);
 		List<CtCatch> catchers = tryStmt.getCatchers();
 		assertEquals(1, catchers.size());
-		
+
 		CtCatchVariable<?> catchVariable = catchers.get(0).getParameter();
 
 		assertEquals(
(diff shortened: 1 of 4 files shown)


src/test/java/spoon/test/trycatch/TryCatchTest.java
```diff
@@ -289,4 +289,18 @@ public void testCatchWithExplicitFinalVariable() throws IOException {
...
+		// contract: unknown exception type in multicatch should not cause IndexOutOfBoundsException
+		String inputResource = "./src/test/resources/spoon/test/noclasspath/exceptions/Foo.java";
+		Launcher launcher = new Launcher();
+		launcher.addInputResource(inputResource);
+		launcher.getEnvironment().setNoClasspath(true);
+		CtModel model = launcher.buildModel();
+
+		List<CtCatch> catches = model.getElements(new TypeFilter<CtCatch>(CtCatch.class));
+		catches.get(0).getParameter().getType(); // catch with single UnknownException
+		catches.get(1).getParameter().getType(); // multicatch with UnknownException
```
pvojtechovsky: could you please add assert, which checks the returned value too?
assertNull(...)

pvojtechovsky:
It would be nice to have it documented in javaDoc of `CtCatchVariable#getType()` too.

pvojtechovsky:
Thank You!

PR #1935, opened 2018-03-30
review fix: contract CtAnnotation and array of values

surli:
This PR intends to fix the contract defined after discussion in #1928

### DIFF
--- a/src/main/java/spoon/support/reflect/declaration/CtAnnotationImpl.java
+++ b/src/main/java/spoon/support/reflect/declaration/CtAnnotationImpl.java
@@ -359,7 +359,7 @@ public class CtAnnotationImpl<A extends Annotation> extends CtExpressionImpl<A> 
 	}
 
 	@Override
-	public <T extends CtExpression> T getTypedValue(String key) {
+	public <T extends CtExpression> T getWrappedValue(String key) {
 		CtExpression ctExpression = this.getValue(key);
 
 		if (ctExpression instanceof CtLiteral) {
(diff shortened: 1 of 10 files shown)


src/main/java/spoon/support/reflect/declaration/CtAnnotationImpl.java
```diff
@@ -357,6 +358,31 @@ private Object getDefaultValue(String fieldName) {
...
+			if (type != null) {
+				CtMethod method = type.getMethod(key);
+				if (method != null) {
+					CtTypeReference returnType = method.getType();
+					if (returnType instanceof CtArrayTypeReference) {
+						CtNewArray newArray = getFactory().Core().createNewArray();
+						CtArrayTypeReference typeReference2 = this.getFactory().createArrayTypeReference();
+						typeReference2.setComponentType(ctExpression.getType().clone());
+						newArray.setType(typeReference);
+						newArray.addElement(ctExpression);
```
pvojtechovsky: `newArray.addElement(ctExpression.clone());`
clone is missing

src/main/java/spoon/support/reflect/declaration/CtAnnotationImpl.java
```diff
@@ -357,6 +358,31 @@ private Object getDefaultValue(String fieldName) {
 		return (T) this.elementValues.get(key);
 	}
 
+	@Override
+	public <T extends CtExpression> T getTypedValue(String key) {
+		CtExpression ctExpression = this.getValue(key);
+
+		if (ctExpression instanceof CtLiteral) {
+			CtTypeReference typeReference = this.getAnnotationType();
+			CtType type = typeReference.getDeclaration();
```
monperrus: use getTypeDeclaration?

monperrus:
I have a doubt about the name `getTypedValue` because all values are already typed.

Does this method always return a CtNewArray?

surli:
> Does this method always return a CtNewArray?

only if it respects some conditions (it's the only CtLiteral value and the annotation should return an array) else it returns the same value as `getValue`.

monperrus:
what about getWrappedValue?

monperrus:
Thanks Simon!

PR #1936, opened 2018-03-30
review Set default no classpath as true and trigger a warning when it's not set

### LINKED ISSUE #1934: Use noClasspath by default in Spoon
Most bugs we are getting there can be simply resolved by using a `setNoClasspath(true)` in the environment: why not setting this by default? 

1. It would ease the adoption of Spoon
2. For old clients we can trigger a warning to set the default value has changed
3. For old clients, as they should normally already set the classpath it should not be a problem: the classpath is still used 
4. We can always change that and release it as a major release

WDYT?

surli:
Fix #1934

### DIFF
--- a/src/main/java/spoon/support/StandardEnvironment.java
+++ b/src/main/java/spoon/support/StandardEnvironment.java
@@ -469,7 +469,7 @@ public class StandardEnvironment implements Serializable, Environment {
 	public boolean getNoClasspath() {
 		if (this.noclasspath == null) {
 			logger.warn("Spoon is currently use with the default noClasspath option set as true. Read the documentation for more information: http://spoon.gforge.inria.fr/launcher.html#about-the-classpath");
-			return true;
+			this.noclasspath = true;
 		}
 		return noclasspath;
 	}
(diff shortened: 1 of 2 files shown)


src/main/java/spoon/support/StandardEnvironment.java
```diff
@@ -458,15 +460,17 @@ public void setPreserveLineNumbers(boolean preserveLineNumbers) {
...
-
 	@Override
 	public void setNoClasspath(boolean option) {
 		noclasspath = option;
 	}
 
 	@Override
 	public boolean getNoClasspath() {
+		if (this.noclasspath == null) {
+			logger.warn("Spoon is currently use with the default noClasspath option set as true. Read the documentation for more information: http://spoon.gforge.inria.fr/launcher.html#about-the-classpath");
```
pvojtechovsky: Do we want to print this warning
A) only once
B) for each call ... to fill the log by these warnings?

I vote for A) ... then add `this.noclasspath = true`

monperrus: I propose to move this warning from `getNoClasspath` to a more higher level method such as `Launcher#run`

surli: I'm ok with @pvojtechovsky but I don't understand the purpose of what you propose @monperrus .

monperrus:
thanks @surli

Issue #1933, opened 2018-03-30
ModelBuildingException when spoon classes with static method calls


XuSihan:
I am trying to analyze the source code of apache cassandra.
```java
 Launcher launcher = new Launcher();
 launcher.addInputResource("path\\to\\cassandra");
 launcher.buildModel(); <-- this crashes
```
Here's the exception:

2018-03-30 11:27:49,893 ERROR spoon.support.compiler.jdt.JDTTreeBuilder - Could not find declaration for variable Pair at (/Users/Sihan/eclipse-workspace/Test/resources/cassandra/src/java/org/apache/cassandra/schema/Schema.java:152)
2018-03-30 11:27:49,902 ERROR spoon.support.compiler.jdt.JDTTreeBuilder - Could not find declaration for variable Pair at (/Users/Sihan/eclipse-workspace/Test/resources/cassandra/src/java/org/apache/cassandra/schema/Schema.java:177)
2018-03-30 11:27:49,913 ERROR spoon.support.compiler.jdt.JDTTreeBuilder - Could not find declaration for variable Pair at (/Users/Sihan/eclipse-workspace/Test/resources/cassandra/src/java/org/apache/cassandra/schema/Schema.java:191)
2018-03-30 11:27:49,926 ERROR spoon.support.compiler.jdt.JDTTreeBuilder - Could not find declaration for variable Pair at (/Users/Sihan/eclipse-workspace/Test/resources/cassandra/src/java/org/apache/cassandra/schema/Schema.java:207)
2018-03-30 11:27:49,936 ERROR spoon.support.compiler.jdt.JDTTreeBuilder - Could not find declaration for variable Pair at (/Users/Sihan/eclipse-workspace/Test/resources/cassandra/src/java/org/apache/cassandra/schema/Schema.java:290)
2018-03-30 11:27:52,211 ERROR spoon.support.compiler.jdt.JDTTreeBuilder - Could not find declaration for variable e at (/Users/Sihan/eclipse-workspace/Test/resources/cassandra/src/java/org/apache/cassandra/metrics/ClientMetrics.java:47)
2018-03-30 11:27:52,561 ERROR spoon.support.compiler.jdt.JDTTreeBuilder - Could not find declaration for variable Metrics at (/Users/Sihan/eclipse-workspace/Test/resources/cassandra/src/java/org/apache/cassandra/metrics/MessagingMetrics.java:56)
[...]

PR #1937, opened 2018-03-30
review Add a new warning in verifySourceClasspath

surli:
This PR aims at avoiding issues like #1933


surli:
> by: Only .jar files or directories with *.java files are accepted. The directories with *.jar files or *.class files are ignored.

I changed a bit the text: 
Only .jar files or directories with *.class files are accepted. The *.jar or *.java files contained in given directories are ignored.

> Or may be we should automatically replace classpath which point to a directory by list of contained jar files? ... looks like clients expects .that .. and the java doc would be easier too...

I though about that but it could be quite expensive: in #1933 the user tried to add its complete .m2 directory as sourceClasspath...

PR #1938, opened 2018-03-30
review: Add mechanism to sort compilation units

### LINKED ISSUE #1929: discussion: buildModel and CompilationUnits order
While working on #1928 I get stuck on a bug which was appearing when building the model from the input path `./src/test/java/spoon/test/` and disappearing when building the model from path: `./src/test/java/spoon/test/annotation/testclasses/shadow`. 
And my entire test was built only with types coming from `./src/test/java/spoon/test/annotation/testclasses/shadow`.
The bug was in fact coming because one type was resolving **before** another one I needed in the first case, and **after** it in the second case. 

So in other word, I got a bug because of the order of resolution of the compilation units. The interesting point is that I was quite confident on my implementation, until my test crashes because I changed the path and so changed the order of resolution. 

What action should we take about this? 

1. None 
2. To fix the order in which compilation units are sorted in JDTBasedSpoonCompiler and to rely only on this order
3. To define a random order with a seed, and to check in CI tests with different seeds, to ensure everything's passing even if the classes are not resolved in the same order

WDYT?

surli:
This PR should fix #1929

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/CompilationUnitComparator.java
+++ b/src/main/java/spoon/support/compiler/jdt/CompilationUnitComparator.java
@@ -1,0 +1,29 @@
+package spoon.support.compiler.jdt;
+
+import org.eclipse.jdt.internal.compiler.ast.CompilationUnitDeclaration;
+import spoon.support.StandardEnvironment;
+
+import java.util.Comparator;
+import java.util.Random;
+
+public class CompilationUnitComparator implements Comparator<CompilationUnitDeclaration> {
+    private int seed;
+
+    public CompilationUnitComparator(int seed) {
+        this.seed = seed;
+    }
+
+
+    @Override
+    public int compare(CompilationUnitDeclaration o1, CompilationUnitDeclaration o2) {
+        if (this.seed == StandardEnvironment.DEFAULT_SEED_CU_COMPARATOR) {
+            String s1 = new String(o1.getFileName());
+            String s2 = new String(o2.getFileName());
+            return s1.compareTo(s2);
+        } else {
+            Random random = new Random(seed);
+            int r = random.nextInt(3); // can be 0, 1 or 2
+            return r-1; // can be -1, 0 or 1
+        }
+    }
+}
--- a/src/main/java/spoon/support/compiler/jdt/CompilationUnitComparator.java
+++ b/src/main/java/spoon/support/compiler/jdt/CompilationUnitComparator.java
@@ -18,7 +18,6 @@ package spoon.support.compiler.jdt;
 
 import org.eclipse.jdt.internal.compiler.ast.CompilationUnitDeclaration;
 import spoon.Launcher;
-import spoon.support.StandardEnvironment;
 
 import java.util.Comparator;
 import java.util.Random;
--- a/src/main/java/spoon/support/compiler/jdt/CompilationUnitComparator.java
+++ b/src/main/java/spoon/support/compiler/jdt/CompilationUnitComparator.java
@@ -1,3 +1,19 @@
+/**
+ * Copyright (C) 2006-2017 INRIA and contributors
+ * Spoon - http://spoon.gforge.inria.fr/
+ *
+ * This software is governed by the CeCILL-C License under French law and
+ * abiding by the rules of distribution of free software. You can use, modify
+ * and/or redistribute the software under the terms of the CeCILL-C license as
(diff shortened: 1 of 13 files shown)


src/main/java/spoon/support/compiler/jdt/CompilationUnitComparator.java
```diff
@@ -0,0 +1,49 @@
...
+
+import java.util.Comparator;
+import java.util.Random;
+
+public class CompilationUnitComparator implements Comparator<CompilationUnitDeclaration> {
+
+	@Override
+	public int compare(CompilationUnitDeclaration o1, CompilationUnitDeclaration o2) {
+		int seed = 0;
+		if (System.getenv("SPOON_SEED_CU_COMPARATOR") != null) {
```
pvojtechovsky: > copied from javadoc of System#getenv
> If a security manager exists, its checkPermission method is called with a RuntimePermission("getenv."+name) permission. This may result in a SecurityException being thrown. 

It would be cleaner to try catch security exception here.

src/main/java/spoon/support/compiler/jdt/JDTBasedSpoonCompiler.java
```diff
@@ -430,8 +430,11 @@ protected boolean buildUnitsAndModel(JDTBuilder jdtBuilder, SpoonFolder sourcesF
 
 	protected void buildModel(CompilationUnitDeclaration[] units) {
 		JDTTreeBuilder builder = new JDTTreeBuilder(factory);
+		List<CompilationUnitDeclaration> unitList = Arrays.asList(units);
+		unitList.sort(new CompilationUnitComparator());
```
pvojtechovsky: Is it intentional that this sort sorts the origin `units` array? It actually modifies the array sent from outside. I have checked who actually calls it and it looks like it is actually no problem, but it would be safer to sort a copy of that array.
WDYT?

surli: You're right: I wasn't aware that `Arrays.asList` was keeping up the link with the original array. Good thing to know 👍

pvojtechovsky:
Java code is OK for me.
@monperrus please check the CI stuff ... I do not know that.

monperrus:
Interesting.

Two comments:

- AFIAU, we do fixed order by default, I thought you wanted to do random order by default.
- I propose to put the variation point outside the comparator, ie having a `RandomizeCompilationUnitOrderComparator` and `FixedOrderBasedOnFileNameCompilationUnitComparator`

surli:
> we do fixed order by default, I thought you wanted to do random order by default.

No as I said in #1929 that we should have always the same order but we should test that we are not dependent of the order: a way to do that is to randomize the order when testing. 

> I propose to put the variation point outside the comparator

OK with that change, I'll do it

surli:
> I propose to put the variation point outside the comparator

Before I do the change, just to be clear, you want me to integrate my test with `getenv` inside `JDTBasedSpoonCompiler` then? It looks a bit ugly...

monperrus:
Yes. We can add a method `JDTBasedSpoonCompiler#orderCompilationUnits` to make it less ugly.

It seems to me that adding a new Travis job for such a specific need and rare bug is somehow overkill, because it is run on all PRs. You suggested a Jenkins job which I tend to prefer. WDYT?

surli:
> We can add a method

OK I'll do it that way

> You suggested a Jenkins job which I tend to prefer. WDYT?

Jenkins does not provide an immediate feedback, however I agree it's overkill to run this test on all PRs commits: a middle way might be to put it only for commits on master.

monperrus:
then we can have a Gakoci job for master.

monperrus:
what about adding a test case?

src/test/java/spoon/test/architecture/SpoonArchitectureEnforcerTest.java
```diff
@@ -328,6 +328,7 @@ public void testSpecPackage() throws Exception {
 		officialPackages.add("spoon.reflect.visitor");
 		officialPackages.add("spoon.reflect");
 		officialPackages.add("spoon.support.comparator");
+		officialPackages.add("spoon.support.comparator.cu");
```
monperrus: seems overkill to add a new package for those two small classes? spoon.support.comparator would be enough?

surli:
@monperrus it's ready for me.

monperrus:
thanks!

PR #1939, opened 2018-03-31
review: fix: comments after class name and before type members

pvojtechovsky:
This PR fixes this situation.
```java
class A
  // this comment belongs to class
   extends B /* this comment belongs to class too*/ { /* but this comment is comment of first member*/
  int firstMember;
}
```
Before all the comments from the example above were associated to the `firstMember`


monperrus:
thanks!

PR #1940, opened 2018-03-31
help: fix: always create new type reference for primitive type

pvojtechovsky:
The type references to primitive types like `void` sometime copied the source position from another type reference. See test for details.

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
+++ b/src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
@@ -93,7 +93,6 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.TreeMap;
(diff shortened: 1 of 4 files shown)


pvojtechovsky:
This PR actually fails on `AnnotationTest#testSpoonSpoonResult`. I would say that this test is wrong. Hard to say for me what was origin intent.

@surli @monperrus please have a look at it and fix the test or explain me what was the idea, so I can fix it.

monperrus:
it was a fix, see #970.

however, like you, when looking at the test and the spooned code, it's unclear.

I would suggest to go ahead, change the assertion!(and we'll see if dropwizard fails again, and we'll try to better fix it)

pvojtechovsky:
I guess the assertion was intended for the annotated field. So I changed it this way.

I am finished here. It is ready for merge from my point of view

monperrus:
Thanks Pavel.

PR #1945, opened 2018-04-03
review fix: Object[]::new references resolution

### LINKED ISSUE #1883: bug: Java8 "Foo[]::new" cannot be resolved
Assuming a Java8 `new ArrayList().stream().toArray(Foo[]::new)`

The `Foo[]::new` is a CtExRefExpression which cannot be resolved with `getExecutableDeclaration` even if Foo is in the classpath.

See failing test in https://github.com/INRIA/spoon/compare/master...monperrus:bug-execrefexpr

surli:
fix #1883


surli:
The proposed solution is far from being perfect but it's a first step. I propose that before improving it further we define first a better test contract.

src/test/java/spoon/test/reference/ExecutableReferenceTest.java
```diff
@@ -204,4 +207,23 @@ public void testHashcodeWorksWithReference() {
 
 		assertNotEquals(hashCode1, hashCode2);
 	}
+
+	@Test
+	public void testPbWithStream() {
+		// contract: array constructor references are well represented
```
monperrus: add assert on type of element?
add assert on toString of element?

src/main/java/spoon/support/reflect/declaration/InvisibleArrayConstructor.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+package spoon.support.reflect.declaration;
+
+/**
+ * This class is used to represent the constructor of an array when calling with an expression like
+ * <pre>
+ *     String[]::new
+ * </pre>
+ */
+public class InvisibleArrayConstructor<T> extends CtConstructorImpl<T> {
```
monperrus: By convention, should end by Impl: `InvisibleArrayConstructorImpl`, and should be in `spoon.support.reflect.declaration` (all implementation classes are in spoon.support)

monperrus:
Thanks @surli.

LGTM, see comments.

src/main/java/spoon/support/reflect/declaration/InvisibleArrayConstructor.java
surli: It is in spoon.support. I will rename it.

surli:
For me it's ready again.

monperrus:
Thanks Simon!

Issue #1833, opened 2018-01-30
In noclasspath mode, all calls to `CtTypeReference#getActualClass` should catch SpoonClassNotFoundException


surli:
The current behaviour in Spoon is that if `getActualClass` is called a `SpoonClassNotFoundException` is thrown if the class cannot be found. It is still true in noclasspath mode, and it makes sense when a client calls the method directly. 

However IMHO it does not make sense in the internal API of Spoon: in noclasspath mode, we should never propagate a SpoonClassNotFoundException in the internal API. I suggest that we change the API to manage those as standard exception and no more RuntimeException. The benefit will be that we know exactly where we need to catch those or to propagate them. WDYT?

monperrus:
In general, client code should never call getActualClass() because it breaks Spoon's encapsulation
(source code model or shadow class).

This is even more true in noclasspath, in noclasspath there is absolutely no reason to call
getActualClass.

So I would say it's better to throw this exception, to violently warn clients that they are doing
something really dangerous.

tdurieux:
The problem is that our API uses getActualClass in noclasspath

msteinbeck:
> The problem is that our API uses getActualClass in noclasspath

I observed several errors resulting from this fact. noclasspath mode and getActualClass should be strictly isolated to make Spoon more fail-safe.

surli:
> The problem is that our API uses getActualClass in noclasspath

I don't agree it's a problem: you can use the noclasspath mode with a classpath, so you can use getActualClass. The problem is that we do not throw the exception, so in other methods we cannot rely on different behaviour when we are in noclasspath.

monperrus:
We can continue the discussion with a PR, and see the impact on tests.

monperrus:
this is done at the client method level, in particular getTypeDeclaration, see #2028

PR #1946, opened 2018-04-04
WiP refactor: catching SpoonClassNotFoundException

surli:
This PR is a first step to solve #1833 
SpoonClassNotFoundException should be properly managed in `noclasspath` mode to avoid crashing the process. However this PR shows that they're used everywhere.


surli:
In order to see where we should catch `SpoonClassNotFoundException` I decided to change its inheritance to see it in method signature. And I created another exception called `SpoonClassNotFoundRuntimeException` that inherits from SpoonException. 

So: 
  * `SpoonClassNotFoundException` is managed internally in Spoon
  * `SpoonClassNotFoundRuntimeException` is always thrown and is crashing the system

 1. I want to know if you're ok with my changes, if I make the systems crashes on the right place and manage the internal exception on the right place
 2. I really wonder if we should rollback the `SpoonClassNotFoundException` to an inheritance towards RuntimeException: `getActualClass` is really used everywhere and can be potentially destructive if we don't use well. I think it would ease the maintenance if we keep throwing an exception from that specific method.

monperrus:
Hi Simon,

Thanks for the proposal. 

Here, you are talking about the old debate between checked and unchecked exception. In Spoon, it's been a long time that we have decided to only use unchecked exception (as Spring and many other Java projects).

That said, introducing a `SpoonInternalException` seems a useful good idea. We would differentiate between `SpoonException` (for end users) and `SpoonInternalException` (for us to catch).

monperrus:
that seems a good solution!

PR #1954, opened 2018-04-09
review: Fix failing builds because of the usage of random in CompilationUnit comparison

surli:
This PR intends to fix the failing build created after the merge of #1938 
Apparently we got lucky with the random when it was first executed: the usage of some random number could lead to the trigger of the following exception `IllegalArgumentException:Comparison method violates its general contract!`

This happens during the sort algorithm because as we use a random answer for sorting, the result is not transitive.
To avoid this problem I decided to completely remove the concept of seed and to just sort or shuffle the CU list, based on a switch value in the environment. 

Before merging the PR I need to uncoment the lines of the shell script to make it only available on master commits: here I uncommented them only to check if travis is fixed.


surli:
There's actually one remaining failing test on the job with the shuffling of CUs. The funny part is that actually I got the same error on my computer with JDK 9 for a while but I thought my configuration was broken somewhere.

Actually it seems to be related with the order of resolution of CUs: so we already have a bug because of that.

monperrus:
So I guess it's a WIP?

surli:
Apparently the error does not appear using JDK8 so I fixed it by changing TravisCI script to use JDK8.

surli:
ping @monperrus it's ready now

PR #1956, opened 2018-04-11
refactor: printing of lists in DefaultJavaPrettyPrinter

pvojtechovsky:
This code is a side effect of my experiments with printing of changes #1927. I though that I will need to know, which list is actually printed by DJPP, so I implemented that `ElementPrinterHelper#printList`.
Now I see it is not needed, but the code of this PR doesn't look bad from my point of view.

So if you like it, take it. Otherwise close this PR - I do not need that.

This PR also moves printing of method modifier "default" into `ElementPrinterHelper#writeModifiers`


monperrus:
ok for me, refactoring PRs are always welcome.

PR #1959, opened 2018-04-12
Review: feat(position): add the position of the modifier in the ast

tdurieux:


src/main/java/spoon/support/reflect/CtExtendedModifier.java
```diff
@@ -53,6 +55,14 @@ public void setKind(ModifierKind kind) {
 		this.kind = kind;
 	}
 
+	public SourcePosition getPosition() {
+		return position;
```
pvojtechovsky: `if (position == null) return NoSourcePosition;`
WDYT? CtElement already never returns null.

src/main/java/spoon/support/compiler/jdt/PositionBuilder.java
```diff
@@ -234,13 +235,51 @@ SourcePosition buildPositionCtElement(CtElement e, ASTNode node) {
...
+
+	private void setModifiersPosition(CtModifiable e, int start, int end) {
+		CoreFactory cf = this.jdtTreeBuilder.getFactory().Core();
+		CompilationUnit cu = this.jdtTreeBuilder.getFactory().CompilationUnit().getOrCreate(new String(this.jdtTreeBuilder.getContextBuilder().compilationunitdeclaration.getFileName()));
+		CompilationResult cr = this.jdtTreeBuilder.getContextBuilder().compilationunitdeclaration.compilationResult;
+		char[] contents = cr.compilationUnit.getContents();
+
+		Set<CtExtendedModifier> modifiers = e.getExtendedModifiers();
+		if (start < 0 || end + 1 > contents.length) {
+			System.out.println(e);
```
pvojtechovsky: Is it really wanted?

tdurieux: sorry, debug point that I forget to remove

src/main/java/spoon/support/compiler/jdt/PositionBuilder.java
```diff
@@ -133,19 +139,16 @@ SourcePosition buildPositionCtElement(CtElement e, ASTNode node) {
...
-			if (modifiersSourceStart == 0) {
+			if (modifiersSourceStart <= 0) {
 				modifiersSourceStart = declarationSourceStart;
 			}
 			//look for start of first keyword before the type keyword e.g. "class". `sourceStart` points at first char of type name
 			int modifiersSourceEnd = findPrevNonWhitespace(contents, modifiersSourceStart - 1,
 										findPrevWhitespace(contents, modifiersSourceStart - 1,
 											findPrevNonWhitespace(contents, modifiersSourceStart - 1, sourceStart - 1)));
+			if (e instanceof CtModifiable) {
+				setModifiersPosition((CtModifiable) e, modifiersSourceStart, bodyStart);
```
pvojtechovsky: I have not read whole code yet, so may be it is correct, but usage of `bodyStart` here seems to be wrong.

src/main/java/spoon/support/compiler/jdt/PositionBuilder.java
```diff
@@ -234,13 +235,51 @@ SourcePosition buildPositionCtElement(CtElement e, ASTNode node) {
...
+		CoreFactory cf = this.jdtTreeBuilder.getFactory().Core();
+		CompilationUnit cu = this.jdtTreeBuilder.getFactory().CompilationUnit().getOrCreate(new String(this.jdtTreeBuilder.getContextBuilder().compilationunitdeclaration.getFileName()));
+		CompilationResult cr = this.jdtTreeBuilder.getContextBuilder().compilationunitdeclaration.compilationResult;
+		char[] contents = cr.compilationUnit.getContents();
+
+		Set<CtExtendedModifier> modifiers = e.getExtendedModifiers();
+		if (start < 0 || end + 1 > contents.length) {
+			System.out.println(e);
+		}
+		String modifierContent = String.valueOf(Arrays.copyOfRange(contents, start, end + 1));
```
pvojtechovsky: I guess there is faster `new String(contents, start, end+1)` or something like that. Or why do we copy that array?

src/main/java/spoon/support/compiler/jdt/PositionBuilder.java
pvojtechovsky: ok, the searched string is little bit longer then necessary, but I cannot imagine code where it would cause problem, so it is OK for me.

tdurieux:
I thought about this and also in an annotation contains public or final or static but handle them is not that easy and it is probably pretty rare.
Because you need an annotation followed by a comment that contains the modifier name.

PR #1963, opened 2018-04-13
fix(position): PartialSourcePosition only support getCompilationUnit

tdurieux:


pvojtechovsky:
I fixed the code which asks for position and doesn't checks that there is partial source position only.

But I am not happy with that solution, so feel free to rollback it ... it would be nicer if there would be a better way how to detect whether position
A) is valid  - SourcePositionImpl
B) not valid - either `NoSourcePosition` or `PartialSourcePositionImpl`

I can imagine several solutions:
S1) PartialSourcePositionImpl extends NoSourcePosition
S2) `boolean SourcePosition#isValid()` ... or any better name
S3) `SourcePosition#getSourceStart` returns -1 for invalid SourcePosition  and >=0 for valid source position. Other methods of SourcePosition might behave same
S4) ?

@tdurieux @monperrus WDYT?

tdurieux:
I think I prefer solution S1 but I would prefer to throw an exception instead of -1 in NoSourcePosition.

monperrus:
ready for merge?

pvojtechovsky:
1) I like the idea of throwing of exception, because it avoids bugs caused by using invalid value `-1`.

what about change of NoSourcePosition and to throw exception there too?

2) I would like to have an easy way how to detect that SourcePostion is correct.

> S1) PartialSourcePositionImpl extends NoSourcePosition

validity detected by: `position instanceof NoSourcePosition == false`
* it is longer then S2
* client must know internal implementation class to detect that
> S2) boolean SourcePosition#isValid() ... or any better name

validity detected by: `position.isValid()`

pvojtechovsky:
This PR is a step in a good way. So it is ready for merge. We can do other thinkgs in another PRs.

PR #1969, opened 2018-04-18
fix(position): fix the position of catch variable

### LINKED ISSUE #1968: CtCatchVariable does not include source position of modifiers
`CtCatchVariable` implements the `CtModifiable` interface and therefore can include modifiers.
For all other classes that implement `CtModifiable` I found that their SourcePosition is of type `DeclarationSourcePosition` containing the getters `getModifierSourceStart()` and `getModifierSourceEnd()`.
This is not the case for the CtCatchVariable.

tdurieux:
fix #1968


tdurieux:
Done for me

Issue #1077, opened 2016-12-26
feature: iterable Scanner


monperrus:
spin-off of #1070 

**Idea**: an iterable scanner.

we could then have 

    interface CtElement extends Iterator<CtElement>

so as to easily go over the children elements of an element:

    for (CtElement el : klass) { // do something }

The `CtIterableScanner` class will also be useful for some advanced queries.

It can probably be generated out of CtScanner directly, as we do for Clone, Replace and Equal implementation classes.

retuxx:
You need to define the iteration order, is it pre-, in-, or postfix?

pvojtechovsky:
It is still true that I do not need iterable Scanner for queries, so do not understand this comment as: "Yes it makes sense to do it, because Pavel needs it too". 
But if you think that iterable scanner would be useful, then here is one more idea:
Do not implement just `Iterator` (which has method `next`, but implement these model iterating methods:
* nextChild
* nextSibling
* nextParent
* ?
So the client can drive the way of iterating through the spoon model.

pvojtechovsky:
One note: There are two strategies how to access the data: [push and pull](http://stackoverflow.com/questions/5792966/data-pull-vs-push-oop-approach). 

> Pushing : The advantage of pushing is that **you know your data** and you know what you are pushing. No component knows (or should not know) the data better than the component who owns the data, which will theoretically **imply a better design and a more robust system**.

> Pulling : The only advantage I see in pulling approach is the component who pulls exactly knows when it should pull. It can initiate the conversation when it needs the data and no component knows (or should not know) when the data is needed than the component who needs it.

It depends on situation which one is better - as usually. 

Current CtScanner is implemented using **push** strategy.
The Iterable Scanner will use **pull** strategy.

I do not know your plans with Iterable scanner, but I vote for keeping and future supporting of origin CtScanner, because in some (in case of spoon ... most?) scenarios CtScanner will be much better then iterable scanner.

Second note: the pull and push are incompatible like fire and water. If spoon will support both strategies, then it will probably end with duplicate implementation of all data accessing algorithms. So it increases maintenance cost of the spoon library. May be I am wrong, but I wanted to warn you.

Question: Could you please describe your plans with iterable scanner? Which use cases of spoon clients are covered by this new feature?

monperrus:
> I vote for keeping and future supporting of origin CtScanner

We will keep it. CtIterableScanner will be a separate class. Note that in Spoon, we always aim at maximum backward compatibility.

[...]

PR #1986, opened 2018-05-08
feature: add CtElement#descendantIterator and CtElement#asIterable

AndreFCruz:
This is a method for obtaining an Iterator over the CtElement's descendants, as suggested in #1077 .
It aims to make the CtIterator class easier to find.


monperrus:
Thanks a lot for this pull-request.

Could you add a test to verify that one can use this feature in a foreach loop?

AndreFCruz:
I added a static method ```CtElement.asIterable``` that I think facilitates the use of the Iterator.

In this way, a user can iterate with the regular for loop:
```java
for (CtElement elem : CtElement.asIterable(ctElem)) {
        ...
}
```
and using the iterator's forEachRemaining
```java
ctElem.descendantIterator().forEachRemaining((CtElement elem) -> ... );
```

And added a test for both methods.

monperrus:
Thanks a lot!

In Spoon, we don't use static interface methods.

Instead of a static method, is it possible to  directly have `Iterable<CtElement> asIterable()` as normal interface method + implementation in CtElementImpl?

Side-note: would you put your test directly in `CtIteratorTest.java`?

Cheers --Martin

AndreFCruz:
Sure, moved the method to the class implementation as non-static, and moved all iteration tests to CtIteratorTest.java

monperrus:
Thanks a lot. I'm OK to merge if nobody disagrees with those two new methods in CtElement.

pvojtechovsky:
I like this PR too 👍

monperrus:
Thanks a lot André for your contribution.

PR #1992, opened 2018-05-17
review: fix: setModifiers(Collections.emptySet()) resets modifiers now

pvojtechovsky:


monperrus:
are we able to write a test case to specify this?

PR #1993, opened 2018-05-17
refactor: EqualsChecker#setNotEqual can be used to debug non-equality

pvojtechovsky:
If I need to know what attribute is not equal, then this PR is very helpful, because one can put breakpoint into new method setNotEqual()


monperrus:
Thanks Pavel

tdurieux:
It would even be nicer to provide the two CtElements that are different.

PR #1997, opened 2018-05-19
review fix: actualTypeArguments of array type reference

pvojtechovsky:
I guess I have found a bug in actual type arguments of array type reference.
The problem is that array type reference like `List<?>[]` has set wildcard `<?>` actual type arguments on 2 places
1) on List type: `List<?>` - here it is OK
2) in List array type: `List[]` - here it is wrong. Or WDYT is it bug or feature?

I added two failing test, which both actually fails on last test line.


monperrus:
thanks!

PR #1998, opened 2018-05-19
review: feat: EqualsVisitor can report which attribute is not equal

pvojtechovsky:
please check if performance of Spoon changes with that change. Equals visitor is called very often, so it must be really fast ... and I am not sure about performance impact of this PR


tdurieux:
That is a good feature for debug

surli:
> We don't have a systematic performance evaluation infrastructure.

Actually we might see the difference in the test duration after the merge of the PR here: https://ci.inria.fr/sos/job/Spoon%20Install%20Commit%20Hook/

PR #2008, opened 2018-05-25
refactor: get*Field now use shadow types

monperrus:


pvojtechovsky:
private method `getDeclaredFieldReferences` is unused now. Let's remove it too.

PR #2009, opened 2018-05-26
review: feature: fail when element is added twice

pvojtechovsky:
There is basic spoon concept: each element of AST has 0 or 1 parent and is in children collection of exactly one parent.

When some element which is already part of AST is added into another part of AST, then this rule concept is not true and Spoon can behave unpredictable.

This PR assures that element add will fail in such situation.

### DIFF
--- a/src/main/java/spoon/support/compiler/SnippetCompilationHelper.java
+++ b/src/main/java/spoon/support/compiler/SnippetCompilationHelper.java
@@ -92,6 +92,9 @@ public class SnippetCompilationHelper {
 		// Clean up
 		c.getPackage().removeType(c);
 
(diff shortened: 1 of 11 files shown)


monperrus:
> There is basic spoon concept: each element of AST has 0 or 1 parent and is in children collection of exactly one parent.

I agree with this contract.

This PR is not the only solution for this. 

This other solution to maintain this contract is to automatically do call `delete` before any call to `setParent`. This would be much more backward compatible.

WDYT?

ashutosh1598:
I think it is better to throw an exception rather than silently deleting the node. Two days ago I had to spend some time to debug because I forgot to clone a CtBlock.

monperrus:
thanks Pavel!

PR #2015, opened 2018-05-30
review: fix: position of (String arg[])

pvojtechovsky:
Position of parameter `arg` is incorrect in this code
```java
void method(String arg[])
```


pvojtechovsky:
@tdurieux this is an interesting case ... `String arg[]` the CtParameter.getType.getPosition must return `String arg[]` ... including parameter name... What SourcePosition type we will use for that special case???

tdurieux:
:/ Java...

I'm not sure, I would say DeclarationSourcePosition

and the failing test-case is strange:
`PositionTest.testArrayArgParameter:767 expected:<String arg[[]]> but was:<String arg[]>` I don't understand the diff

PR #2020, opened 2018-06-01
fix(reference): fix npe in getTypeReference when the reference is null

tdurieux:
fix npe in #2014

No idea how to reproduce it in a failing test-case.


pvojtechovsky:
Thank you Thomas

tdurieux:
thanks @bilak for the issue:

Issue #2003, opened 2018-05-23
SpoonException when computing imports


mboussaa:
Hi all,

I hit this exception when I run the following code. Any help? 

```
ImportScanner importContext = new ImportScannerImpl();
importContext.computeImports(classA);
```

`spoon.SpoonException: The lambda can be based on interface, which has only one method. But java.util.Comparator has at least two: compare(T,T) and equals(java.lang.Object)`

The concerned class is: [org.springframework.web.reactive.result.method.RequestMappingInfoHandlerMapping](https://github.com/spring-projects/spring-framework/blob/master/spring-webflux/src/main/java/org/springframework/web/reactive/result/method/annotation/RequestMappingHandlerMapping.java)

Thanks in advance.

Best,
Mohamed

pvojtechovsky:
Hi Mohamed,

it is really interesting case. It looks like `Comparator#equals` is a special case of method which can be ignored in Spoon Lambda expression resolving algorithm, because that method is always implemented by java.lang.Object.

Could you create a pull request with a failing test? I can then fix it quite fast.

filkra:
Hi Pavel,

I created a pull request over here https://github.com/INRIA/spoon/pull/2023. Let me know what you think.

Best Regards,
Filip

PR #2023, opened 2018-06-01
Fix Interface method handling

filkra:
## Description

From [FunctionalInterface](https://docs.oracle.com/javase/8/docs/api/java/lang/FunctionalInterface.html)'s JavaDoc:
> If an interface declares an abstract method overriding one of the public methods of java.lang.Object, that also does not count toward the interface's abstract method count since any implementation of the interface will have an implementation from java.lang.Object or elsewhere.

[CtLambdaImpl](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/reflect/code/CtLambdaImpl.java#L126-L134) doesn't take this into account resulting in Exceptions whenever an Interface declaring one or more of `java.lang.Object`'s methods is used in combination with a Lambda Expression.

## Related Issues

- Fixes https://github.com/INRIA/spoon/issues/2003

### DIFF
--- a/src/main/java/spoon/support/reflect/code/CtLambdaImpl.java
+++ b/src/main/java/spoon/support/reflect/code/CtLambdaImpl.java
@@ -23,11 +23,11 @@ import spoon.reflect.code.CtBodyHolder;
 import spoon.reflect.code.CtExpression;
 import spoon.reflect.code.CtLambda;
 import spoon.reflect.code.CtStatement;
-import spoon.reflect.declaration.CtExecutable;
 import spoon.reflect.declaration.CtMethod;
 import spoon.reflect.declaration.CtNamedElement;
 import spoon.reflect.declaration.CtParameter;
 import spoon.reflect.declaration.CtType;
+import spoon.reflect.declaration.CtExecutable;
 import spoon.reflect.declaration.ModifierKind;
 import spoon.reflect.path.CtRole;
 import spoon.reflect.reference.CtExecutableReference;
@@ -124,6 +124,9 @@ public class CtLambdaImpl<T> extends CtExpressionImpl<T> implements CtLambda<T> 
 			lambdaExecutableMethod = lambdaTypeMethods.iterator().next();
 		} else {
 			for (CtMethod<?> method : lambdaTypeMethods) {
+				if (getFactory().Method().OBJECT_METHODS.stream().anyMatch(method::isOverriding)) {
+					continue;
+				}
 				if (method.isDefaultMethod() || method.hasModifier(ModifierKind.PRIVATE) || method.hasModifier(ModifierKind.STATIC)) {
 					continue;
 				}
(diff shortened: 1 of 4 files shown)


pvojtechovsky:
Filip, welcome! And thank you for this high quality PR 👍

src/main/java/spoon/support/reflect/code/CtLambdaImpl.java
```diff
@@ -124,6 +119,9 @@ public String getSimpleName() {
 			lambdaExecutableMethod = lambdaTypeMethods.iterator().next();
 		} else {
 			for (CtMethod<?> method : lambdaTypeMethods) {
+				if (getFactory().Method().OBJECT_METHODS.contains(method)) {
```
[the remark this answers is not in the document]
filkra: Thanks for the incredibly fast feedback!

I just tried using `CtMethod.isOverriding`.

```java
for (CtMethod<?> method : lambdaTypeMethods) {
    if (getFactory().Method().OBJECT_METHODS.stream().anyMatch(method::isOverriding)) {
        continue;
    }
...
```

Unfortunately the method does not return `true` since the Interface is no subtype of `java.lang.Object` ([Reference](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/visitor/ClassTypingContext.java#L274-L277)).

It seems the only solution is comparing method signatures, as you said.

pvojtechovsky:
> CtMethod.isOverriding.
> Unfortunately the method does not return true since the Interface is no subtype of java.lang.Object

I think it is a bug, which should be fixed first. Then we can use isOverriding here too.
@monperrus WDYT?

pvojtechovsky:
The comparation of signatures doesn't work with java generics
```java
class B<T> extends A<T> {
void m(T arg){}
}
class A<U> {
void m(U arg){}
}
```
here B#m overrides A#m, but they have different signatures.

So do not invest more effort on signatures, because it is probably wrong way.

filkra:
[CtLambdaImpl](https://github.com/filkra/spoon/blob/2ed24f68ffb75ae54677fb6ede157d793746291c/src/main/java/spoon/support/reflect/code/CtLambdaImpl.java#L127) now checks if the interface's methods are overriding `java.lang.Object` methods. Once https://github.com/INRIA/spoon/pull/2025 is merged this should work.

pvojtechovsky:
#2025 was merged, please do some little change in your PR so the next CI build is triggered using latest sources.

filkra:
Thanks Pavel!

I rebased on top of master and the test succeeded on my local machine.

src/test/java/spoon/test/lambda/LambdaTest.java
```diff
@@ -403,6 +404,18 @@ public void testLambdaFilter() throws Exception {
 		assertHasStrings(methodNames, "m2", "m4", "m7", "m8");
 	}
 
+	@Test
+	public void testInterfaceWithObjectMethods() throws Exception {
+		CtInterface<?> checkPersons = factory.Interface().get(Foo.CheckPersons.class);
```
pvojtechovsky: please add contract to the first line of the test body. Something like
//contract: Lambda expression works on interfaces with methods inherited from java.lang.Object

filkra: Done :)

pvojtechovsky:
Thank You Filip 👍

PR #2027, opened 2018-06-03
Review: getActualClass never returns null

monperrus:


src/main/java/spoon/support/reflect/reference/CtArrayTypeReferenceImpl.java
```diff
@@ -85,7 +86,7 @@ public String getQualifiedName() {
 	public Class<T> getActualClass() {
 		Class<?> c = getComponentType().getActualClass();
 		if (c == null) {
-			return null;
+			throw new SpoonClassNotFoundException("you should never call getActualClass", null);
```
pvojtechovsky: Could the exception contain the qualified name of the class which is not found? .. it might improve debugging

monperrus:
add `@Deprecated` on getActualClass (we will remove it from the public API, but may keep it in the implementation)

PR #2030, opened 2018-06-04
chore: Change README to encourage usage/contribution

surli:
The README of Spoon always fear me when seeing the whole bibtex citation on the first lines: I think it does not encourage its usage and the "getting started" is currently not immediately viewable.

This PR intends to propose another first impression for the repo.


monperrus:
Simon, I really like your improvement... but the Bibtex change.

One of the unique feature of Spoon is that it combines the best of academic research and the best of high-quality open-source.

This is part of Spoon's identity since its birth with @renaudpawlak and @petitpre, and I want to keep that in our DNA.

(slightly refactored the README accordingly, readme for merge).

surli:
I'm still not sure why moving the bibtex after the getting started change the DNA of Spoon, but at least here we have a better getting started IMHO. 

Anyone to merge it? @tdurieux @pvojtechovsky?

PR #2033, opened 2018-06-04
fix: npe in getEnclosingType when getTypeDeclaration is null

tdurieux:
@pvojtechovsky can you check the semantic of what I did.
I'm not sure if it is correct


pvojtechovsky:
It is better then it was. But it is hard to say what is correct behavior if some types are not available - it cannot return 100% correct results then.

So I am ok with that fix. Thank you Thomas. Are you going to add some test? Or is it done?

tdurieux:
I planned to try to add test. 
I will do another pr if I succeed to do it.

PR #2036, opened 2018-06-06
feature: one can get annotation values directly

monperrus:
Two advantages:
* this eases the API: direct `annot.getValueAsObject("foo") ` instead of  `((CAST)annot.getActualAnnotation()).foo()`
* this enables getting the annotation value even if the annotation binary class file is not in the classpath

### DIFF
--- a/src/main/java/spoon/reflect/declaration/CtAnnotation.java
+++ b/src/main/java/spoon/reflect/declaration/CtAnnotation.java
@@ -68,7 +68,10 @@ public interface CtAnnotation<A extends Annotation> extends CtExpression<A>, CtS
 	CtTypeReference<A> getAnnotationType();
 
 	/**
-	 * Gets a value for a given key without any conversion.
+	 * Gets a value, as a CtExpression, for a given key without any conversion.
+	 *
+	 * If you need the actual value (eg an integer and not a literal, see {@link #getValueAsObject(String)} and similar methods.
+	 *
 	 * Note that this value type does not necessarily corresponds to the annotation
 	 * type member. For example, in case the annotation type expect an array of Object,
 	 * and a single value is given, Spoon will return only the object without the CtNewArray.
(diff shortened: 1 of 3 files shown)


src/main/java/spoon/reflect/declaration/CtAnnotation.java
```diff
@@ -81,6 +84,18 @@
 	@PropertyGetter(role = VALUE)
 	<T extends CtExpression> T getValue(String key);
 
+	/** Returns the actual value of an annotation property */
+	@DerivedProperty
+	Object getValueAsObject(String key);
+
+	/** Returns the actual value of an annotation property, as an integer (utility method) */
+	@DerivedProperty
+	int getValueAsInt(String key);
```
surli: What if the key does not exist? We would get a ClassCastException?

monperrus: It would return null. In Java cast(null) returns null.

surli: It will work with the string but not int as an int can never return null.
Actually I just tested it and it throws a NPE. Could you add a check?

monperrus: done

pvojtechovsky:
Thank You Martin.

src/main/java/spoon/support/reflect/declaration/CtAnnotationImpl.java
```diff
@@ -312,7 +312,22 @@ private CtExpression getDefaultExpression(String fieldName) {
 		return (T) getValueAsExpression(key);
 	}
 
-	private Object getValueAsObject(String key) {
+	@Override
+	public int getValueAsInt(String key) {
+		Object val = getValueAsObject(key);
+		if (val == null) {
+			throw new IllegalStateException(key + " not in the annotation");
```
surli: Why not returning null here like in the other methods?

PR #2040, opened 2018-06-06
fix(shadow): fix the creation of a partial shadow class

### LINKED ISSUE #2038: Shadow and partial classpath
I think there is a major bug in the Shadow.

You have:
- A in the source
- B referenced in A and is in the classpath
- C referenced in B but is not in the classpath

The creation of the shadow of B will fail because when we will try to create the shadow of C a SpoonClassNotFound will be thrown and the creation of the shadow of B will stop.

tdurieux:
It is a little bit dirty, do you have better idea how to do it?

Fix #2038

### DIFF
--- a/src/main/java/spoon/support/visitor/java/JavaReflectionVisitorImpl.java
+++ b/src/main/java/spoon/support/visitor/java/JavaReflectionVisitorImpl.java
@@ -49,28 +49,28 @@ class JavaReflectionVisitorImpl implements JavaReflectionVisitor {
 			for (TypeVariable<Class<T>> generic : clazz.getTypeParameters()) {
 				visitTypeParameter(generic);
 			}
-		} catch (Throwable ignore) {
+		} catch (NoClassDefFoundError ignore) {
 			// partial classpath
 		}
 		try {
 			if (clazz.getGenericSuperclass() != null && clazz.getGenericSuperclass() != Object.class) {
 				visitTypeReference(CtRole.SUPER_TYPE, clazz.getGenericSuperclass());
 			}
-		} catch (Throwable ignore) {
+		} catch (NoClassDefFoundError ignore) {
 			// partial classpath
 		}
 		try {
 			for (Type anInterface : clazz.getGenericInterfaces()) {
 				visitTypeReference(CtRole.INTERFACE, anInterface);
 			}
-		} catch (Throwable ignore) {
+		} catch (NoClassDefFoundError ignore) {
 			// partial classpath
 		}
 		try {
 			for (Annotation annotation : clazz.getDeclaredAnnotations()) {
 				visitAnnotation(annotation);
 			}
-		} catch (Throwable ignore) {
+		} catch (NoClassDefFoundError ignore) {
 			// partial classpath
 		}
 		try {
@@ -80,7 +80,7 @@ class JavaReflectionVisitorImpl implements JavaReflectionVisitor {
 				}
 				visitConstructor(constructor);
 			}
-		} catch (Throwable ignore) {
+		} catch (NoClassDefFoundError ignore) {
 			// partial classpath
 		}
 		try {
@@ -90,7 +90,7 @@ class JavaReflectionVisitorImpl implements JavaReflectionVisitor {
 				}
 				visitMethod(method);
 			}
-		} catch (Throwable ignore) {
+		} catch (NoClassDefFoundError ignore) {
 			// partial classpath
 		}
 		try {
@@ -100,14 +100,14 @@ class JavaReflectionVisitorImpl implements JavaReflectionVisitor {
 				}
 				visitField(field);
 			}
-		} catch (Throwable ignore) {
+		} catch (NoClassDefFoundError ignore) {
 			// partial classpath
 		}
 		try {
 			for (Class<?> aClass : clazz.getDeclaredClasses()) {
 				visitType(aClass);
 			}
-		} catch (Throwable ignore) {
+		} catch (NoClassDefFoundError ignore) {
 			// partial classpath
 		}
 	}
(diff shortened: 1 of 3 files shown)


src/main/java/spoon/support/visitor/java/JavaReflectionVisitorImpl.java
```diff
@@ -46,38 +45,70 @@ public void visitPackage(Package aPackage) {
...
 			clazz.getPackage();
 		}
-		for (TypeVariable<Class<T>> generic : clazz.getTypeParameters()) {
-			visitTypeParameter(generic);
+		try {
+			for (TypeVariable<Class<T>> generic : clazz.getTypeParameters()) {
+				visitTypeParameter(generic);
+			}
+		} catch (Throwable ignore) {
+			// partial classpath
```
pvojtechovsky: can we catch more exact exception? Such error handling may cause big problems during future debugging.

src/main/java/spoon/support/visitor/java/JavaReflectionVisitorImpl.java
```diff
@@ -46,38 +45,70 @@ public void visitPackage(Package aPackage) {
...
+				visitTypeReference(CtRole.SUPER_TYPE, clazz.getGenericSuperclass());
+			}
+		} catch (Throwable ignore) {
+			// partial classpath
 		}
-		for (Type anInterface : clazz.getGenericInterfaces()) {
-			visitTypeReference(CtRole.INTERFACE, anInterface);
+		try {
+			for (Type anInterface : clazz.getGenericInterfaces()) {
+				visitTypeReference(CtRole.INTERFACE, anInterface);
```
pvojtechovsky: I suggest to try/catch this line only. So some interfaces are correctly created and some are ignored. Actually it process all interfaces until first which is not on classpath.

Do the same in all cycles of this method, please.

src/main/java/spoon/support/visitor/java/JavaReflectionVisitorImpl.java
```diff
@@ -131,46 +187,70 @@ public void visitPackage(Package aPackage) {
...
 		}
-		for (Constructor<?> constructor : clazz.getDeclaredConstructors()) {
-			if (Modifier.isPrivate(constructor.getModifiers())) {
-				Class<?>[] paramTypes = constructor.getParameterTypes();
-				if (paramTypes.length == 2 && paramTypes[0] == String.class && paramTypes[1] == int.class) {
-					//ignore implicit enum constructor
+		try {
+			for (Annotation annotation : clazz.getDeclaredAnnotations()) {
+				visitAnnotation(annotation);
+			}
```
pvojtechovsky: This for cycle seems to be copy past here by mistake. It is already above. Please fix it

src/main/java/spoon/support/visitor/java/JavaReflectionVisitorImpl.java
tdurieux: I cannot the Error is thrown by clazz.getGenericInterfaces()

src/main/java/spoon/support/visitor/java/JavaReflectionVisitorImpl.java
tdurieux: I don't understand it is the only loop on clazz.getDeclaredAnnotations() for visitEnum(Class<T> clazz).
Did I miss something?

src/main/java/spoon/support/visitor/java/JavaReflectionVisitorImpl.java
tdurieux: and I did not succeed to find a way to get a partial list :/

src/main/java/spoon/support/visitor/java/JavaReflectionVisitorImpl.java
pvojtechovsky: See line 142-144 in this file

src/main/java/spoon/support/visitor/java/JavaReflectionVisitorImpl.java
pvojtechovsky: So then try/catch only clazz.getGenericSuperclass(). So the exceptions thrown in visitTypeReference are not catched

src/main/java/spoon/support/visitor/java/JavaReflectionVisitorImpl.java
tdurieux: It is not the same method

pvojtechovsky: I see. First one processes Interface and second one processes Enum. You are right. It is correct. Thanks for explanation.

src/main/java/spoon/support/visitor/java/JavaReflectionVisitorImpl.java
tdurieux: I don't have the motivation to do it, right now. 
There is probably 50 invocations to protect in this class.

pvojtechovsky: it can be written in nice way using lambda:
`for (Type anInterface : ignoreMissingClass(()->clazz.getGenericInterfaces()))`
where ignoreMissingClass will return empty collection in case of exception

tdurieux: clazz.getGenericInterfaces() returns an array not a collection.
I don't see a way to do it nicely :/

tdurieux: but the problem is to create EMPTY_ARR, because EMPTY_ARR has to be different for each T

monperrus:
LGTM.

PR #2042, opened 2018-06-07
feature: add CtAnnotation#getAllValues to reason about all annotation values

monperrus:
add CtAnnotation#getAllValues to reason about all annotation values regardless of whether they are defined in the annotation or as default value.


pvojtechovsky:
Looks good to me

PR #2043, opened 2018-06-08
review feat: add the ability to follow the process of the creation of the Spoon model

### LINKED ISSUE #1778: feature: Add Progress Listener for Model Building
Using Spoon to create a model for large size projects takes quite some time. Thus, I would like to attach some sort of progress listener to the `Launcher` class allowing me to track the current progress while creating a model. The progress listener should be able to distinguish between source compilation (JDT) and model creation (Spoon). I did some research and found that the JDT compiler allows to attach a [CompilationProgress](https://help.eclipse.org/mars/index.jsp?topic=%2Forg.eclipse.jdt.doc.isv%2Freference%2Fapi%2Forg%2Feclipse%2Fjdt%2Fcore%2Fcompiler%2FCompilationProgress.html). For the model creation, I would suggest to fire a progress event for each compilation unit.

Wrapping JDT's `CompilationProgress` class with a custom progress listener should be straightforward.

What do you think?

tdurieux:
fix #1778 

I notice with this feature that printing the model is the slowest task.

JTD takes: 4794 ms
Spoon model: 4098 ms
Printing: 5507 ms

It surprised me

### DIFF
--- a/src/main/java/spoon/support/compiler/ProgressLogger.java
+++ b/src/main/java/spoon/support/compiler/ProgressLogger.java
@@ -1,0 +1,32 @@
+package spoon.support.compiler;
+
+import spoon.support.compiler.jdt.JDTTreeBuilder;
+
+import java.util.GregorianCalendar;
+
+public class ProgressLogger implements SpoonProgress {
+	private long stepTimer;
+	private long timer;
+
+	@Override
+	public void start(STEP step) {
+		System.out.println("Start " + step);
+		timer = getCurrentTimeInMillis();
+		stepTimer = timer;
+	}
+
+	@Override
+	public void step(STEP step, String element, int taskId, int nbTask) {
+		JDTTreeBuilder.getLogger().trace("Step " + step + " " + taskId + "/" + nbTask + " " + element + " in " + (getCurrentTimeInMillis() - timer) + " ms");
+		timer = getCurrentTimeInMillis();
+	}
+
+	@Override
+	public void end(STEP step) {
+		JDTTreeBuilder.getLogger().trace("End " + step + " in " + (getCurrentTimeInMillis() - stepTimer) + " ms");
+	}
+
+	private long getCurrentTimeInMillis() {
+		return new GregorianCalendar().getTimeInMillis();
+	}
+}
--- a/src/main/java/spoon/support/compiler/ProgressLogger.java
+++ b/src/main/java/spoon/support/compiler/ProgressLogger.java
@@ -1,35 +1,40 @@
 package spoon.support.compiler;
 
-import spoon.support.compiler.jdt.JDTTreeBuilder;
+import spoon.support.StandardEnvironment;
 
 import java.util.GregorianCalendar;
 
 public class ProgressLogger implements SpoonProgress {
 	private long stepTimer;
 	private long timer;
+	private StandardEnvironment environment;
+
+	public ProgressLogger(StandardEnvironment environment) {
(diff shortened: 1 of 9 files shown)


src/main/java/spoon/support/compiler/SpoonProgress.java
```diff
@@ -0,0 +1,38 @@
+package spoon.support.compiler;
+
+/**
+ * is the interface to follow the progress of the creation of the Spoon model.
+ */
+public interface SpoonProgress {
+
+	enum PROCESS {
```
surli: Why not calling it `Process`?

src/main/java/spoon/support/compiler/SpoonProgress.java
```diff
@@ -0,0 +1,38 @@
...
+	 * is called when a new process is started
+	 * @param process the started process
+	 */
+	void start(PROCESS process);
+
+	/**
+	 * is called when a step in the precess is started
+	 * @param process the current process
+	 * @param task the task that has been processed
+	 * @param taskId the task if
```
surli: I think the end of the comment is missing ;)

surli:
Is it ready?

src/main/java/spoon/support/compiler/SpoonProgress.java
```diff
@@ -0,0 +1,38 @@
...
+	void start(Process process);
+
+	/**
+	 * is called when a step in the precess is started
+	 * @param process the current process
+	 * @param task the task that has been processed
+	 * @param taskId the task id
+	 * @param nbTask the number of task in the process
+	 */
+	void step(Process process, String task, int taskId, int nbTask);
```
surli: Am I wrong to say that everywhere in your usage, the task is in fact the filename of the file being processed?

src/main/java/spoon/support/compiler/SpoonProgress.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+public interface SpoonProgress {
+
+	enum Process {
+		COMPILE,
+		COMMENT,
+		MODEL,
+		IMPORT,
+		PROCESS,
+		PRINT
```
surli: I don't see in the diff the usage of PRINT? Shouldn't it used in DJPP?

src/main/java/spoon/support/compiler/SpoonProgress.java
tdurieux: for the moment it contains only the absolute path of the file.

surli: So don't you think that the parameter should be renamed then?

surli: I mean: it's not the task, it's at best a description. The task itself is given by the process parameter.

tdurieux: it is the task of the process.

src/main/java/spoon/support/compiler/SpoonProgress.java
tdurieux: Done

src/main/java/spoon/support/compiler/ProgressLogger.java
```diff
@@ -0,0 +1,43 @@
+package spoon.support.compiler;
```
surli: The license header is missing. Could you add it please, else it will break the CI on master: merge with master and `mvn license:format`

PR #2045, opened 2018-06-08
review refactor: Fix the blocker smells reported by SonarQube with try-with-resources

surli:
This PR intends to improve our score on SonarQube
See: https://sonarqube.ow2.org/project/issues?id=fr.inria.gforge.spoon%3Aspoon-core&resolved=false&severities=BLOCKER&types=BUG

### DIFF
--- a/src/main/java/spoon/support/compiler/ZipFolder.java
+++ b/src/main/java/spoon/support/compiler/ZipFolder.java
@@ -72,11 +72,7 @@ public class ZipFolder implements SpoonFolder {
 		// Indexing content
 		if (files == null) {
 			files = new ArrayList<>();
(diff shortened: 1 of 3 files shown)


monperrus:
FYI, 
@ashutosh1598 works on an automatic fixing tool for SonarQube warnings based on Spoon and the sniper mode: https://github.com/kth-tcs/sonarqube-repair

Ashutosh: this PR gives you a ground truth on how to fix this bug typ.

ashutosh1598:
The diff of this PR is much better than my automatic repair for this rule. I am going to change my repair so that nested try blocks are not unnecessarily created for each resource.

PR #2050, opened 2018-06-10
test: derived and unsettable of spoon.metamodel.Metamodel

### LINKED ISSUE #2047: How to test correctness of spoon.metamodel.Metamodel?
I compared results returned by new `spoon.metamodel.Metamodel` with hardcoded Metamodel which was in PR #1686 and it found some problems.

Many problems were fixed by readding contract "unsettable implies derived" from #1904, which was removed during Metamodel refactorings.

Here are remaining problems
```
Field CtInterface#modifier<java.util.Set<spoon.reflect.declaration.ModifierKind>>.derived hardcoded value = false but computed value is true
Field CtInterface#interface<java.util.Set<spoon.reflect.reference.CtTypeReference<? extends java.lang.Object>>>.derived hardcoded value = false but computed value is true
Field CtTypeParameter#modifier<java.util.Set<spoon.reflect.declaration.ModifierKind>>.unsettable hardcoded value = true but computed value is false
Field CtTypeParameter#superType<spoon.reflect.reference.CtTypeReference<? extends java.lang.Object>>.derived hardcoded value = false but computed value is true
Field CtConstructor#name<java.lang.String>.derived hardcoded value = true but computed value is false
Field CtConstructor#name<java.lang.String>.unsettable hardcoded value = true but computed value is false
Field CtAnonymousExecutable#name<java.lang.String>.derived hardcoded value = true but computed value is false
Field CtAnonymousExecutable#name<java.lang.String>.unsettable hardcoded value = true but computed value is false
Field CtWildcardReference#name<java.lang.String>.derived hardcoded value = true but computed value is false
Field CtWildcardReference#name<java.lang.String>.unsettable hardcoded value = true but computed value is false
Field CtEnum#label<java.lang.String>.derived hardcoded value = true but computed value is false
Field CtEnum#label<java.lang.String>.unsettable hardcoded value = true but computed value is false
Field CtEnum#modifier<java.util.Set<spoon.reflect.declaration.ModifierKind>>.derived hardcoded value = false but computed value is true
Field CtEnum#interface<java.util.Set<spoon.reflect.reference.CtTypeReference<? extends java.lang.Object>>>.derived hardcoded value = false but computed value is true
Field CtClass#label<java.lang.String>.derived hardcoded value = true but computed value is false
Field CtClass#label<java.lang.String>.unsettable hardcoded value = true but computed value is false
Field CtClass#modifier<java.util.Set<spoon.reflect.declaration.ModifierKind>>.derived hardcoded value = false but computed value is true
Field CtClass#superType<spoon.reflect.reference.CtTypeReference<? extends java.lang.Object>>.derived hardcoded value = false but computed value is true
Field CtClass#interface<java.util.Set<spoon.reflect.reference.CtTypeReference<? extends java.lang.Object>>>.derived hardcoded value = false but computed value is true
Field CtAnnotationType#modifier<java.util.Set<spoon.reflect.declaration.ModifierKind>>.derived hardcoded value = false but computed value is true
```

pvojtechovsky:
The values of derived and unsettable attributes of MetamodelProperties of MetamodelConcepts of spoon.metamodel.Metamodel are compared to the hardcoded expected values

It is a way how to fix #2047

### DIFF
--- a/src/test/java/spoon/generating/MetamodelGenerator.java
+++ b/src/test/java/spoon/generating/MetamodelGenerator.java
@@ -1,0 +1,104 @@
+package spoon.generating;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.apache.commons.lang3.text.StrSubstitutor;
+
+import spoon.SpoonException;
+import spoon.metamodel.ConceptKind;
+import spoon.metamodel.Metamodel;
+import spoon.reflect.declaration.CtElement;
+import spoon.reflect.factory.Factory;
+import spoon.reflect.path.CtRole;
+import spoon.reflect.visitor.CtScanner;
+
+public class MetamodelGenerator {
+
+	public static void main(String[] args) {
+		Metamodel mm = Metamodel.getInstance();
+		Factory factory = mm.getConcepts().iterator().next().getMetamodelInterface().getFactory();
+		factory.getEnvironment().useTabulations(true);
+		StringBuilder sb = new StringBuilder();
+		for (spoon.metamodel.MetamodelConcept type : mm.getConcepts()) {
+			if (type.getKind()==ConceptKind.LEAF) {
+				sb.append(printType(factory, type));
+			}
+		}
+		System.out.println(sb.toString());
+	}
+	
+
+	private static String printType(Factory factory, spoon.metamodel.MetamodelConcept type) {
+		Map<String, String> valuesMap = new HashMap<>();
+		valuesMap.put("typeName", type.getName());
+		valuesMap.put("ifaceName", type.getMetamodelInterface().getQualifiedName());
+		valuesMap.put("implName", type.getImplementationClass().getQualifiedName());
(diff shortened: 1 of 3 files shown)


monperrus:
> A) to move hardcoded metamodel from #1686 to test package and to change it manually in case of real change of spoon metamodel. This let's us write a test which checks whethere these two metamodels are equal.
> B) To persist the metamodel somehow and commit the persisted representation into git. Then tests can check if metamodel is equal like persisted version

Both are fine if if have a test.

The problem in this PR is:
``` 
static class Type
static class Field
```
which are duplicate of MetamodelConcept and MetamodelProperty (hard to understand, hard to maintain). 

Is it possible to direclty use MetamodelConcept and MetamodelProperty instead?

pvojtechovsky:
> Is it possible to direclty use MetamodelConcept and MetamodelProperty instead?

I think no. The purpose of this PR is to make a snapshot of the Metamodel state. So we need a duplicate structure, which represents that snapshot. May be we can make that test Metamodel package protected so nobody can use that duplicated structure.

monperrus:
Now, I understand, LGTM.

* we need to revolve conflicts
* do we check that **all** properties are hard-coded?

PR #2051, opened 2018-06-11
review feat: API for incremental build

surli:
This PR aims at finishing the incremental build mechanism proposed by @Egor18 in #1905  with an API well integrated with Spoon Environment and standard launcher.


surli:
@Egor18 WDYT about this API? It's not polished yet: I'm thinking about removing the new fields from Launcher that seem unecessary and to put them as local variable, do you agree?

Egor18:
This API looks good, thanks. Yes, unecessary fields can be removed.
Also, probably we need some perfomance tests.
And finally, I'm not sure if comparison of the modification time of the files is really sufficient for incremental build. What should we do if someone just copies another file?

surli:
ok I'll try first to improve it without changing the way it works.

We'll be able to improve the way we detect changes later: for now modification time can already be used in several cases.

src/main/java/spoon/support/IncrementalBuildInformation.java
```diff
@@ -0,0 +1,227 @@
...
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * This class store information for incremental build.
+ */
+public class IncrementalBuildInformation {
```
monperrus: put `@Internal` or in package called something.internal (eg spoon.launcher.internal)

monperrus:
Incremental launcher is a different responsibility. What about having an IncrementalLauncher for this? (which would delegate to Launcher)

surli:
So actually the incremental launcher was really the API provided by @Egor18 in #1816 Maybe his original PR is enough then?

monperrus:
Why was #1905 still marked as WIP?

surli:
> Why was #1905 still marked as WIP?

I suppose because I asked for changes regarding the API...

monperrus:
:-)

did you add new tests here?

surli:
No

monperrus:
Thanks a lot @surli to have revived this topic!

monperrus:
That will be a great release!

PR #2054, opened 2018-06-11
review: feat: Add support to Java 10 'var' keyword

surli:
This PR add the support to Java 10 `var` keyword for local type inference.


src/main/java/spoon/reflect/path/CtRole.java
```diff
@@ -101,7 +101,8 @@
 	OPENED_PACKAGE(MODULE_DIRECTIVE),
 	SERVICE_TYPE(MODULE_DIRECTIVE),
 	IMPLEMENTATION_TYPE,
-	PROVIDED_SERVICE(MODULE_DIRECTIVE);
+	PROVIDED_SERVICE(MODULE_DIRECTIVE),
+	INFERRED_VARIABLE;
```
monperrus: I propose to name it IS_INFERRED

src/test/java/spoon/reflect/ast/AstCheckerTest.java
```diff
@@ -158,6 +161,18 @@ private boolean isNotCandidate(CtMethod<?> candidate) {
 					|| notCandidates.contains(candidate.getDeclaringType().getSimpleName() + "#" + candidate.getSimpleName());
 		}
 
+		private boolean hasPrimitiveTypeForParameters(List<CtParameter<?>> parameters) {
```
monperrus: I don't see the relation with this PR. Why was it working before?

src/test/java/spoon/test/comment/CommentTest.java
```diff
@@ -749,6 +749,7 @@ public void testDocumentationContract() throws Exception {
 		final Launcher launcher = new Launcher();
 		launcher.getEnvironment().setNoClasspath(true);
 		launcher.getEnvironment().setCommentEnabled(true);
+		launcher.getEnvironment().setComplianceLevel(10);
```
monperrus: why is this required here?

src/test/java/spoon/reflect/ast/AstCheckerTest.java
surli: I'm not sure. I suspect that on almost all other setters we were not using primitive types. I can extract it in another PR if you want: the relation here is that it fix a broken test resulting by the new setter.

src/test/java/spoon/test/comment/CommentTest.java
surli: I added a new snippet in the code with the following code:
```
var x = 10;
```

This won't compile if I do not use a compliance level to 10.

src/test/java/spoon/reflect/ast/AstCheckerTest.java
monperrus: Bizarre. We have plenty of other setters with boolean (eg setImplicit, setShadow, etc). It would be much better to understand what happens here.

surli: I'll have a look

surli: Good thing you insisted on that one: I forgot to call the model change listener... I updated the contract here because it was not clear enough.

monperrus:
That's great, thanks Simon!

PR #2057, opened 2018-06-12
feat(launcher): handles range version in MavenLauncher

tdurieux:


src/main/java/spoon/MavenLauncher.java
```diff
@@ -407,6 +411,139 @@ private String extractVariable(String value) {
...
+			});
+			if (fileVersion == null) {
+				return Collections.emptyList();
+			}
+			List<Version> versions = Arrays.stream(fileVersion).map(f -> new Version(f.getName())).collect(Collectors.toList());
+			versions.sort(Comparator.reverseOrder());
+			return versions;
+		}
+
+		class Version implements Comparable<Version> {
```
surli: What about extracting the inner classes: the MavenLauncher seems now composed almost entirely of InheritanceModel. It seems easier to me to maintain if it's extracted.

tdurieux: I don't think it is a good idea InheritanceModel, is completely useless outside MavenLauncher.
I agree that the architecture is not really nice but it is not the purpose of this PR.

PR #2058, opened 2018-06-13
Upgrade license to 2018

surli:
This PR intends to upgrade the license date but also to put back the plugin to manage licenses.


monperrus:
thanks!

PR #2065, opened 2018-06-14
review feat: Enable comments by default

surli:
This PR is a proposal to change a default behaviour in Spoon: until now, the default was to **not** take into account the comments.
I assume this behaviour was used because the comments were not properly parsed then. But I think it's not the case anymore and we got more users using Spoon for parsing comments: very recently we got a question on Stackoverflow linked to that: https://stackoverflow.com/questions/50780598/how-to-get-comments-using-spoon/ and I know we also got issues related to that.

So in the same way we decided to use `noclasspath` mode by default, I propose that we parse comments by default. WDYT?

### DIFF
--- a/src/main/java/spoon/Launcher.java
+++ b/src/main/java/spoon/Launcher.java
@@ -378,7 +378,14 @@ public class Launcher implements SpoonAPI {
 			sw1 = new Switch("enable-comments");
 			sw1.setShortFlag('c');
 			sw1.setLongFlag("enable-comments");
-			sw1.setHelp("Adds all code comments in the Spoon AST (Javadoc, line-based comments), rewrites them when pretty-printing.");
+			sw1.setHelp("[DEPRECATED] Adds all code comments in the Spoon AST (Javadoc, line-based comments), rewrites them when pretty-printing. (deprecated: by default, the comments are enabled.)");
+			sw1.setDefault("false");
+			jsap.registerParameter(sw1);
+
+			// Disable generation of javadoc.
+			sw1 = new Switch("disable-comments");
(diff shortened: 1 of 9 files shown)


tdurieux:
The comment was not enable for performance reasons.

surli:
> The comment was not enable for performance reasons.

Do we have an idea about the perf difference on a large code base?

tdurieux:
We need to read again all the Java files.
It is no terribly slow but it has an impact.

monperrus:
I'm rather OK for this new default behavior.

monperrus:
THanks. We have to prepare a really good changelog with all the non-bacward compatible changes such as this one.

src/main/java/spoon/Launcher.java
```diff
@@ -425,7 +432,19 @@ protected void processArguments() {
 		environment.setTabulationSize(jsapActualArgs.getInt("tabsize"));
 		environment.useTabulations(jsapActualArgs.getBoolean("tabs"));
 		environment.setCopyResources(!jsapActualArgs.getBoolean("no-copy-resources"));
-		environment.setCommentEnabled(jsapActualArgs.getBoolean("enable-comments"));
+
+		if (jsapActualArgs.getBoolean("enable-comments")) {
+			Launcher.LOGGER.warn("The option --enable-comments (-c) is deprecated as it is now the default behaviour in Spoon.");
+		} else {
+			Launcher.LOGGER.warn("Spoon now parse by default the comments. Consider using the option --disable-comments if you want the old behaviour.");
```
seintur: Typo: parse -> parses

Issue #2063, opened 2018-06-14
bug: CtGenerationTest#testGenerateRoleHandlers relies on ModelHandler which is generated by the test


surli:
There is an egg/chicken issue with `CtGenerationTest#testGenerateRoleHandlers`.
I created in #2054 a new role to manage inferred variables and I use it in `CtLocalVariable`. It seems that now I cannot generate the `ModelRoleHandlers` using this test, because the model role handlers is not defined...

See:
> spoon.SpoonException: The element of class interface spoon.reflect.code.CtLocalVariable does not have CtRole.INFERRED_VARIABLE

Any idea @pvojtechovsky @monperrus ?

surli:
Fixed in #2068

PR #2068, opened 2018-06-14
fix: init RoleHandler lazily to pass when RoleHandler is generated

pvojtechovsky:
fix of #2063

### DIFF
--- a/src/main/java/spoon/metamodel/MetamodelProperty.java
+++ b/src/main/java/spoon/metamodel/MetamodelProperty.java
@@ -74,7 +74,7 @@ public class MetamodelProperty {
 	 */
 	private CtTypeReference<?> itemValueType;
 
-	private final RoleHandler roleHandler;
+	private RoleHandler roleHandler;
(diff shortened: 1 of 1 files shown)


surli:
Thanks @pvojtechovsky for the fix! I just tried it with #2054 the test now works well, however the generated ModelRoleHandlers contains the following wrong imports:

```java
import spoon.generating.meta.spoon.reflect.meta.impl.ListHandler;
import spoon.generating.meta.spoon.reflect.meta.impl.MapHandler;
import spoon.generating.meta.spoon.reflect.meta.impl.SetHandler;
import spoon.generating.meta.spoon.reflect.meta.impl.SingleHandler;
```

I had to remove them manually. I don't know if it's related with your changes here...

pvojtechovsky:
I cannot imagine that this PR would cause the problems with import. I guess it is some other import related already merged PR.

surli:
Ok I will open an issue then.

PR #2069, opened 2018-06-14
fix: Legacy template TypeReference matches Type optionally

pvojtechovsky:
> There is a regression on the template side, in https://github.com/SpoonLabs/spoon-examples, see https://ci.inria.fr/sos/job/spoon-examples/507/console

It fixes problem reported in #1686


monperrus:
thanks

PR #2070, opened 2018-06-15
review fix(module): Spoon is now able to build module in full classpath

### LINKED ISSUE #1769: bug: Project with modules cannot be spooned in full classpath mode yet

surli:
Fix #1769 

This PR changes the internal mechanism of `JDTBatchCompiler` to process CU in order to attach the information of modules.
The process might be encapsulated in a condition to check if we are using Java 9 or more, but I want first to check that everything works with it. 

And kudos @abourdon for the valuable co-debugging with me on this one! 👏


monperrus:
That's good stuff! Thanks @surli & @abourdon!

PR #2071, opened 2018-06-15
review chore: Coverage should use JDK9 else some tests could not be executed

surli:


monperrus:
Coveralls does not work on JDK9, see 

* https://github.com/trautonen/coveralls-maven-plugin/issues/112
* https://coveralls.io/github/INRIA/spoon (no more data since this PR was merged)

We need to switch back to JDK8, #2469

PR #2079, opened 2018-06-16
feat(Pattern): report missing Pattern parameter value

pvojtechovsky:
If Pattern parameter is mandatory and it is missing, then it throws SpoonException with better explanation


monperrus:
thanks

PR #2080, opened 2018-06-18
review fix(comment): Creating an inline comment with line separators throws an exception

surli:
This PR is a proposal for the following problem: when trying to comment all statements of a method, I created an inline comment for a CtIf, and it produces an uncompilable source code, before it only added a `//` before the content.

I propose that we detect automatically the presence of new lines in the content of a comment when trying to create it and to apply an action. 
Here my proposed action is to change the comment type to create a block comment.

Another possibility is to throw an exception in that kind of case. WDYT?

### DIFF
--- a/src/main/java/spoon/reflect/factory/CodeFactory.java
+++ b/src/main/java/spoon/reflect/factory/CodeFactory.java
@@ -16,6 +16,7 @@
  */
 package spoon.reflect.factory;
 
+import spoon.SpoonException;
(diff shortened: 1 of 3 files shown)


monperrus:
Silently ignoring an explicit parameter given by the user is not a good practice (hard to understand, hard to debug).

I would rather suggest to throw an exception, or to create a method createComment(String) where the comment type is automatically inferred.

surli:
I agree.

PR #2083, opened 2018-06-19
review: fix: the imports should be properly computed after processing

surli:
Current import mechanism is the following:
 1. We get existing imports from the source and put them in the compilation unit
 2. When pretty printing, we compute again the imports and add the new ones

We never delete in step 2 an import that has been added in step 1 because it's really tedious.


surli:
This PR is ready to be reviewed, but please wait #2096 and #2090 to be merged and the CI launched again, before merging it.

PR #2090, opened 2018-06-21
review fix: EnumValues constructor call should be marked as implicit

surli:


surli:
@pvojtechovsky I got an assertion error with `PrinterTest#testPrinterTokenListener` because I added a line separator in a printList call:

```java
elementPrinterHelper.printList(ctEnum.getEnumValues(),
					null, false, null, false, false, "," + DefaultJavaPrettyPrinter.LINE_SEPARATOR, false, false, ";",
					enumValue -> scan(enumValue));
```

I don't really know where to start for fixing that one... Do you have an idea?

src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
```diff
@@ -728,7 +728,7 @@ public void visitCtDo(CtDo doLoop) {
 			printer.writeSeparator(";").writeln();
 		} else {
 			elementPrinterHelper.printList(ctEnum.getEnumValues(),
-					null, false, null, false, false, ",", true, false, ";",
+					null, false, null, false, false, "," + DefaultJavaPrettyPrinter.LINE_SEPARATOR, false, false, ";",
 					enumValue -> scan(enumValue));
```
pvojtechovsky: it is not actually allowed to send token which contains two different token types: "," and EOL

Solution:
replace
```java
enumValue -> scan(enumValue)
```
with 
```java
enumValue -> {
   printer.writeln();
   scan(enumValue)
}
```

src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
```diff
@@ -763,21 +763,14 @@ public void visitCtDo(CtDo doLoop) {
...
-					} else {
-						printer.writeSeparator(",");
-					}
-					scan(ctexpr);
-				}
-				printer.writeSeparator(")");
+			if (!constructorCall.isImplicit()) {
+				this.getPrinterTokenWriter().writeSeparator("(");
+				elementPrinterHelper.printList(constructorCall.getArguments(), null, false, null, false, false, ",", true, false, null, expr -> scan(expr));
+				this.getPrinterTokenWriter().writeSeparator(")");
```
pvojtechovsky: remove both `writeSepratator` calls and use instead:
```java
elementPrinterHelper.printList(constructorCall.getArguments(), null, false, "(", false, false, ",", true, false, ")", expr -> scan(expr));
```

src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
surli: Great thanks!

src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
surli: Hmmm I tried with "(" for the first parameter, but it works indeed better at the right place :)

monperrus:
should only be merged after #2096?

surli:
No it can be already merged

PR #2098, opened 2018-06-22
review refactor: Remove deprecated from last release of Spoon

surli:


src/main/java/spoon/legacy/NameFilter.java
```diff
@@ -1,61 +0,0 @@
...
- *
- * <pre>
- * CtMethod&lt;?&gt; normalFor = type.getElements(
- * 		new NameFilter&lt;CtMethod&lt;?&gt;&gt;(&quot;normalFor&quot;)).get(0);
- * </pre>
- *
- * @deprecated Use {@link NamedElementFilter} instead: the actual NameFilter could return wrongly typed results. NamedElementFilter explicit the use of a type.
- */
-@Deprecated
-public class NameFilter<T extends CtNamedElement> implements Filter<T> {
```
monperrus: The rule is that we keep all classes that are in package `legacy`. So we should simply remove the @Deprecated.

surli: My bad. I will put it back.

PR #2100, opened 2018-06-22
WiP feat: computeImports at the level of CU (which were before in DJPP)

surli:
This PR follows the work started in #2083. 
The idea is to be able to compute imports in CU, to allow the user to call that method properly after their transformations: then I'm trying to extract everything regarding the imports from the DJPP to the CU.


src/main/java/spoon/experimental/modelobs/EmptyModelChangeListener.java
```diff
@@ -33,6 +36,7 @@
 	@Override
 	public void onObjectUpdate(CtElement currentElement, CtRole role,
 			CtElement newValue, CtElement oldValue) {
+		this.changeCu(currentElement);
```
pvojtechovsky: Let's rename this type from EmptyModelChangeListener to DefaultModelChangeListener ... because it is no more empty
WDYT?

src/main/java/spoon/experimental/modelobs/EmptyModelChangeListener.java
```diff
@@ -111,10 +126,24 @@ public void onSetDelete(CtElement currentElement, CtRole role, Set field, Collec
...
 	@Override
 	public void onSetDeleteAll(CtElement currentElement, CtRole role, Set field,
 			Set oldValue) {
+		this.changeCu(currentElement);
+	}
+
+	private void changeCu(CtElement element) {
+		SourcePosition position = element.getPosition();
+		if (position != null && !(position instanceof NoSourcePosition)) {
+			position.getCompilationUnit().setIsChanged(true);
```
pvojtechovsky: If spoon needs that call to keep internal consistency, then it should be part of some listener, which is always used - cannot be replaced by a different custom listener. I guess spoon should allow to register more then one listener.
WDYT?

tdurieux:
I m afraid of the negative performance impact of this change

src/main/java/spoon/experimental/modelobs/EmptyModelChangeListener.java
surli: Yes I did not plan to keep it like that ;)

src/main/java/spoon/experimental/modelobs/EmptyModelChangeListener.java
surli: Not necessarily.
It should be part of a standard listener that the user should be able to extend but it also could create another listener and never call this method: he would have to manage manually the flags to compute back the new imports then.

surli:
> I m afraid of the negative performance impact of this change

Actually it should not impact much the users: the imports are only computed back in case of pretty-printing after a change in the model...

tdurieux:
> Actually it should not impact much the users: the imports are only computed back in case of pretty-printing after a change in the model... and this was already the case before

My issue is not this one. You are calling getParent(CtType.class) a lot of time and this method is not a light one. This will be call frequently during the creation of the model...

surli:
> You are calling getParent(CtType.class)

Actually I do this call only if the element has no attached source position to give me the CU.
So it should only happen for the newly added element on the AST if I'm correct.

surli:
> And I m pretty sure that all the compilation unit will be marked as changed with the current change.

Then I should only activate this listener when the model has been built. Is there a property somewhere that is set when the model has been built for the first time? Do you think it might be interesting to add this feature if it does not exist?

pvojtechovsky:
> Do you think it might be interesting to add this feature if it does not exist?

yes! I need that for Sniper mode too ... 
I need something what avoids sending of change events at model creation time.

monperrus:
close?

pvojtechovsky:
The sniper mode still needs improvements in area of computation of imports, so this PR is probably (I haven't tried to understood it yet) still relevant. Simon has to decide.

> I'm trying to extract everything regarding the imports from the DJPP to the CU.

this sounds like a very good way  👍

monperrus:
@surli ok to close?

surli:
I think it will be directly conflicting with Pavel's work on the subject in #2683, so ok to close it for now. I'll see if something in the same idea is still needed after #2683 and if I have more time to implement it :)

Issue #2094, opened 2018-06-21
performance: serialization should use GZIPInputStream to take less space on disk


monperrus:
see https://github.com/INRIA/spoon/issues/1983#issuecomment-399092488

easy good first issue!

surli:
Fixed in #2103

PR #2103, opened 2018-06-23
perf: Gzip model serialization

mehdi-kaytoue:
See #2094. ( my first PR, please be nice with me ;) ) 
Mehdi


src/main/java/spoon/reflect/ModelStreamer.java
```diff
@@ -35,20 +35,24 @@
 	 * 		the factory to be save
 	 * @param out
 	 * 		the used output stream
+	 * @param zipIt
+	 * 		true if output file has to be compressed
 	 * @throws IOException
 	 * 		if some IO error occurs
 	 */
-	void save(Factory f, OutputStream out) throws IOException;
+	void save(Factory f, OutputStream out, boolean zipIt) throws IOException;
```
surli: I wonder if it wouldn't be interesting to propose the compression by default.
Then to keep the old signature of those methods, and to add a boolean `zipIt` and a new setter in the API for deactivating it.

WDYT?

surli:
Could you at least add a new unit test where you call the serializer with the compression: I don't see in your code any usage of save() or load() with the true value.

src/main/java/spoon/reflect/ModelStreamer.java
mehdi-kaytoue: Good idea! I'll try to do it tonight.

mehdi-kaytoue:
Sure, I'll see how to do that.

src/main/java/spoon/reflect/ModelStreamer.java
monperrus: Excellent! In Spoon, we care a lot about backward compatibility, so we generally prefer to add new methods than changing the signature of existing ones.

msteinbeck: I propose to use an enum rather than a boolean. This allows to add further compression types (e.g. lzma) in future.

monperrus:
+1, excellent idea. an enum is also better for readability.

pvojtechovsky:
> However, for SerializationModelStreamer#load, we don't have access to the environment. Is there a way to get it as a singleton?

Would it be possible to detect the type of serialized stream by some header? `save` would first add the header and then would continue with binary data depending on the SerializationType. If yes, then load method doesn't need extra parameters and will automatically use the deserialization type which fits to the current stream content.

mehdi-kaytoue:
The idea is good perse @pvojtechovsky, but I was not able to find a way for now. For Gzip, the few first bit of the inputstream can tell you, but I was not able to make it work for the moment (see e.g., https://www.programcreek.com/java-api-examples/?class=java.util.zip.GZIPInputStream&method=GZIP_MAGIC ).

For now I just try/catch ZipException, as we have just two different types of serialization. 

Checking just the header is OK if we change the compression technique, but if we propose also to use other serialization libraries, try/catch should be ok, but this is not very formal.

pvojtechovsky:
I guess it should be possible to create buffered input stream then to read e.g. first 8 bytes and detect what kind of stream it is. Then reset the content of the stream and to create appropriate stream wrapper (e.g. GZip) and then do it again.
Note: the header might be for example 'SpoonGZip` or anything shorter ... then it can continue with compressed part

src/main/java/spoon/support/SerializationType.java
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * The fact that you are presently reading this means that you have had
+ * knowledge of the CeCILL-C license and that you accept its terms.
+ */
+package spoon.support;
+
+/**
+ *	Different types of serialization.
+ */
+public enum SerializationType {
```
msteinbeck: There is a difference between serialization and compression. Serialization is a mapping to another format, for instance, JSON, XML, and so on whereas compression is a method to reduce the number of bytes requirevd to represent the same entity. The name of you Enum suggests different serialization strategies, though, it also contains different compression techniques. I don't think that Spoon will ever use a serialization technique other than the Java default serialization as, due to the complexity of the model, it is hard to develop and even harder to maintain. The compression technique, on the other hand, may vary.

tl;dr
I suggest to rename this Enum to CompressionType with: NONE and GZIP as values.

mehdi-kaytoue: After looking to a lot of other serialization techniques I arrived to the same conclusion. Keep basic java serialization; propose different compression types. Given my own (yet short) experience of spoon (working with >5GB uncompressed serialized factory), it's enough. Other types of serialization require a lot of work, and as you said, very hard to maintain.

It did as suggested, it should also be easy to add new compression types such as LZMA.

mehdi-kaytoue:
As suggested, we now have:
- A CompressionType (NONE, GZIP) enum
- A getter/setter in StandardEnvironment for the compression type
- SerializationModelStreamer#save compress as chosen in StandardEnvironment
- SerializationModelStreamer#load decompress if needed (look the first bit of the stream to see if its GZIP)
- A junit test for both NONE and GZIP compression

src/test/java/spoon/test/serializable/ModelStreamerTest.java
```diff
@@ -0,0 +1,80 @@
...
+		launcher.addInputResource("./src/main/java/spoon/reflect/declaration");
+		launcher.buildModel();
+		this.factory = launcher.getFactory();
+		file = new File(filename);
+		file.deleteOnExit();
+    }
+
+	@Test
+	public void testGZipCompression() throws IOException {
+
```
surli: I would suggest to specify explicitely that you're using GZIP compression type: the default value might evolve in the future.

surli:
It's a really good job @mehdi-kaytoue! I only have a small comment, but I think we should be able to include it in the next release ;)

src/test/java/spoon/test/serializable/ModelStreamerTest.java
mehdi-kaytoue: You're right! I'll do that tonight after work.

mehdi-kaytoue:
Thanks! I like your comments, very instructing!

monperrus:
@surli, do you proceed with merging?

monperrus:
Congrats and thank you for this first PR!

PR #2107, opened 2018-06-25
WIP: refactor: Remove redundant toString() methods

zielint0:
There is no need to keep redundant toString() methods.
Some of them are changed to 'this' for clarity.
Two typos fixed.


zielint0:
Redundancies are a [code smell](http://wiki.c2.com/?CodeSmell=) and should be removed without any mercy.

Paper [Using Redundancies to Find Errors](https://web.stanford.edu/~engler/p401-xie.pdf) claims that **redundancies correlate with errors**.
> We then use these errors to demonstrate that redundancies,
even when harmless, strongly correlate with the presence of traditional hard errors (...)

Static analysis tools often provides checkers/inspections to detect redundancies
- [**SonarSource**](https://rules.sonarsource.com/java/tag/clumsy/RSPEC-1858) - "toString()" should never be called on a String object - 
> Invoking a method designed to return a string representation of an object which is already a string is a waste of keystrokes. This redundant construction may be optimized by the compiler, but will be confusing in the meantime.
- [**Semmle**](https://help.semmle.com/wiki/display/JAVA/Useless+toString+on+String) - Useless toString on String
> Description: Calling 'toString' on a string is redundant.
> Recommendation
> Do not call toString on a String object.
- [**Spotbugs**](http://spotbugs.readthedocs.io/en/latest/bugDescriptions.html#dm-method-invokes-tostring-method-on-a-string-dm-string-tostring) - Dm: Method invokes toString() method on a String (DM_STRING_TOSTRING)
> Calling String.toString() is just a redundant operation. Just use the String.

Moreover if the reference is null, toString() will generate a NPE. Without toString(), null is printed (no crash!)
```
public class Main {
        String s1 = null;
        String s2 = null;
        System.out.println(s1); // null
        System.out.println(s2.toString()); // NPE
    }
}
```
Conclusion: in nearly every case we should avoid code redundancy, including redundant toString() method calls. I strongly recommend to call toString() ONLY when it is necessary to compile.

> because I don't see immediately the concatenation, and then I don't understand immediately why we would return a this.getType()

I understand your feeling. In the past I had it too. But after some reading of code I lost any appetite for redundancies as they make code longer. In most cases shorter code is easier to: read, fix and extend.

surli:
Actually most of your pointers here are not recommending to call `String.toString()` on which I totally agree. 

But here we mostly call the explicit `toString` method on different kind of objects, like in my example for the `this.getType()` which returns a `CtType` I think. So for me it's not a code redundancy: just an explicit call which might be useful for understanding.

Now where I agree with you is on a avoiding NPE.

monperrus:

However, as @surli, I'm not in favor of this change, because it adds implicitness. In Spoon, we always favor explicit coding practices for better understandability and better maintainability.

zielint0:
@surli 
@monperrus 
Thanks for feedback guys :-)
I will take it into account while creating out ruleset and quality document.
String s = "x"; s.toString(); -> bad
others -> ok

But how about removing toString() from StringBuilder/Buffer? Isn't it clear enough?

Btw. Any other opinions?

surli:
> But how about removing toString() from StringBuilder/Buffer? Isn't it clear enough?

Even there I prefer to see the explicit call.
I propose that we close this PR.

zielint0:
@surli 
> I propose that we close this PR.

Please give me some time to discuss some things here to avoid poluting other places.
I will close this PR later.
Thanks!

PR #2111, opened 2018-06-26
review test: fix line separators in test

surli:
This PR is a first one in order to fix test on windows: when we set `line.separator` we need to reset it to its old value afterwards.

This alone won't fix all tests on windows: we also need to remove all `\n` usage and to use systematically `System.getProperty("line.separator");`. This will be done in another PR.

### DIFF
--- a/src/test/java/spoon/processing/CtGenerationTest.java
+++ b/src/test/java/spoon/processing/CtGenerationTest.java
@@ -1,5 +1,7 @@
 package spoon.processing;
 
+import org.junit.After;
+import org.junit.Before;
 import org.junit.ComparisonFailure;
 import org.junit.Test;
 import spoon.Launcher;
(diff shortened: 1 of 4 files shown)


pvojtechovsky:
>This PR is a first one in order to fix test on windows: when we set line.separator we need to reset it to its old value afterwards.

why to reset it back? We can assume that all spoon tests will use linux separator `\n`

I think it is not main contract of Spoon tests to "play" with OS specific separators. So I suggest to keep spoon tests as simple as possible and to always expect linux EOL.

We can make one test which checks that system property "line.separator" is taken into account by DJPP and that's all. This test should try all possible EOLs of all possible OSes.

> This alone won't fix all tests on windows: we also need to remove all \n usage and to use systematically System.getProperty("line.separator");

I guess it makes spoon tests more complicated and brings no advantage - in worst case it will still pass on some OS and fail on other OS... so we get OS specific tests. I guess it is not good.

WDYT?

surli:
> We can assume that all spoon tests will use linux separator \n

That's not the case: I saw that we're using `System.getProperty("line.separator")` instead of `\n` in some tests.

I have the feeling that keeping `\n` in tests is a bad idea: we are using the line separator property in DJPP, which means that in all tests that are using DJPP we should reset the property to `\n`.

It seems more natural to me to force the usage of the get property everwhere in the tests: then 1. we'll test properly the expected behaviour and 2. we don't risk to miss any call to properly set the line separator property.

I agree with you on the fact that we should keep test as simple as possible but for me it seems easier to understand a test with a line separator get by the property, than a test with a before annotation somewhere which reset this property.

pvojtechovsky:
ok, I do not persist on my opinion.

Just do not modify tests related to content of **comments**. Note that comment values are always separated by `\n` and it is independent on OS. I think it is a nice feature. When DJPP prints such comments back to sources then it can use OS specific EOL, but the code which works with comments is much simpler if it is not OS specific.

PR #2112, opened 2018-06-26
review fix(MavenLauncher): Improving MavenLauncher dependancy resolution

nharrand:
Based on @tdurieux 's work. (WiP)
 * Handles cases where a dependancy's version is not not specified in the targeted pom, but instead is defined earlier or in parent pom.
 * Handles version with weird name
 * Handles the properties pom.groupId, pom.artifactId, pom.version
This could probably benefit some more testing and/or some refactoring.
Help welcome!


surli:
is it ready for review ?

tdurieux:
This class needs a major refactoring because it is a real mess.

surli:
OK it's a WIP then, I edited the title.

nharrand:
I propose to change MavenLauncher's behavior: Instead of implementing every single rules of maven dependency resolution to build a classpath, I  propose to build it by invoking maven. 
 + If maven can run the project, then spoon should at least find the bytecode for every classes used by the project.
 - This requires the user to specify the path to maven home (if it is not a common one). So it might break retro compatibility.

src/main/java/spoon/MavenLauncher.java
```diff
@@ -109,657 +120,99 @@ public MavenLauncher(String mavenProject, String m2RepositoryPath, SOURCE_TYPE s
...
-					}
-				}
+		return classpath;
+	}
 
-			} catch (Exception e) {
-				LOGGER.log(Level.ERROR, "Unable to read the pom of the dependency:" + dependence.toString(), e);
-			}
-			return dependence;
+	private File getDefaultMavenHome() throws SpoonException {
```
surli: outch! This method looks really ugly!
Can't you rely on environment variable instead? The `M2_HOME` was considered as a standard one, but it's not heavily used.

surli:
> Instead of implementing every single rules of maven dependency resolution to build a classpath

Maven is written in Java. Shouldn't we get the implementation of Maven concerning their dependency resolution?

src/main/java/spoon/MavenLauncher.java
nharrand: Oh you saw that...Fine, I'll correct it.

nharrand:
I agree, but I could not find the documention to do that. I'll look better.

surli:
It's maven: the documentation is the source ;) https://github.com/apache/maven-dependency-plugin/tree/maven-dependency-plugin-3.1.1

nharrand:
I've tried to implement a version that uses directly MavenCli programatically (No calls to MavenInvoker). But to be honest, I am unsure whether it is a good idea. I relies on maven packages that I cannot declare (or at least did not succeed to) in the pom because it generates some version conflict (as far as I understand). I think it works on my machine, because it happens that in my classpath the maven classes follow an api that is the same as the one I expect, but I am unsure if it will work on a machine with a different maven version installed, or if spoon is used in other circumstances.

I am starting to think that there is a reason why most people seem to use MavenInvoker. I think it is more stable because it isolate the nested call to maven.

What do you think? (Am I missing something?)

surli:
So if I sum up: the problem is Maven dependencies are difficult to resolve, so the good way to do it is to use Maven directly.

One solution is then to use a Maven Invoker, but clients would have to specify their Maven path.
Another solution is to use Maven dependency resolver but then we cannot guarantee a good behaviour, depending on the version of Maven the clients use.

Then yeah maybe it makes sense to use your first solution...
I propose that you change the API: 
 - by default MavenLauncher checks for the env var `M2_HOME` and try to use it
 - another constructor is created with an argument for the Maven executable path
 - if `M2_HOME` is not available and the argument is not used an expection is thrown

WDYT?

nharrand:
> One solution is then to use a Maven Invoker, but clients would have to specify their Maven path.
Another solution is to use Maven dependency resolver but then we cannot guarantee a good behaviour, depending on the version of Maven the clients use.

Yes, that is at least how I see it.

> by default MavenLauncher checks for the env var M2_HOME and try to use it
    another constructor is created with an argument for the Maven executable path
    if M2_HOME is not available and the argument is not used an expection is thrown

I think we could also try to exec directly the command mvn -version to determine the path to maven home, before throwing an error.

nharrand:
Could someone who has access to travis CI add the following environment variable `M2_HOME="/usr/local/maven"` ?

tdurieux:
You can add it in travis.yml

nharrand:
This code is ready for review!
(Please don't judge the last ten commits of struggle ending up on on just adding `-c` in one travis script...)

chore/travis/travis-jdk10.sh
```diff
@@ -8,6 +8,6 @@ wget https://raw.githubusercontent.com/sormuras/bach/master/install-jdk.sh
 chmod +x install-jdk.sh
 
 export JAVA_HOME=$HOME/openjdk8
-source ./install-jdk.sh -f 10
+source ./install-jdk.sh -f 10 -c
```
surli: Could you explain why you needed this argument?

pom.xml
```diff
@@ -284,6 +289,11 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
+      <groupId>org.apache.maven.shared</groupId>
+      <artifactId>maven-invoker</artifactId>
+      <version>2.2</version>
```
surli: So you have one version 3.1.1 of maven-invoker and one version 2.2? Does it make sense?

src/main/java/spoon/MavenLauncher.java
```diff
@@ -108,656 +125,144 @@ public MavenLauncher(String mavenProject, String m2RepositoryPath, SOURCE_TYPE s
...
+		request.setPomFile(pom);
+		request.setGoals(Arrays.asList("dependency:build-classpath"));
+		Properties properties = new Properties();
+		if (sourceType == SOURCE_TYPE.APP_SOURCE) {
+			properties.setProperty("includeScope", "runtime");
+		}
+		properties.setProperty("mdep.outputFile", "spoon.classpath.tmp");
+		request.setProperties(properties);
+
+		//FIXME Should the standard output made silent and error verbose?
```
surli: It might be interesting to use a logger debug for this one? WDYT?

src/main/java/spoon/MavenLauncher.java
```diff
@@ -108,656 +125,144 @@ public MavenLauncher(String mavenProject, String m2RepositoryPath, SOURCE_TYPE s
...
-		private List<InheritanceModel> modules = new ArrayList<>();
-		private Model model;
-		private InheritanceModel parent;
-		private File directory;
-		private Map<String, String> dependencyManagements = new HashMap<>();
+	/**
+	 *
+	 * @param classPathFiles File[] containing the classpath elements separated with ':'
+	 */
+	public static String[] readClassPath(File ... classPathFiles) {
```
surli: Why does it take an array of files here? You should create only one classpath file, don't you?

src/main/java/spoon/MavenLauncher.java
```diff
@@ -108,656 +125,144 @@ public MavenLauncher(String mavenProject, String m2RepositoryPath, SOURCE_TYPE s
...
+				e.printStackTrace();
+				throw new SpoonException("Failed to read classpath written in temporary file " + classPathFile.getAbsolutePath() + ".");
+			} finally {
+				if (br != null) {
 					try {
-						this.incremental = Integer.parseInt(splitVersion[2]);
-					} catch (NumberFormatException ignore) {
+						br.close();
+					} catch (IOException e) {
+						e.printStackTrace(); //These Exceptions occur after the classpath has been read, so it should not prevent normal execution.
```
surli: Throw instead a SpoonException maybe?

src/main/java/spoon/MavenLauncher.java
```diff
@@ -108,656 +125,144 @@ public MavenLauncher(String mavenProject, String m2RepositoryPath, SOURCE_TYPE s
...
-				}
-				if (splitVersion.length > 1) {
-					try {
-						this.minor = Integer.parseInt(splitVersion[1]);
-					} catch (NumberFormatException ignore) {
-					}
-				}
-				if (splitVersion.length > 2) {
+			} catch (Exception e) {
+				e.printStackTrace();
```
surli: Remove that one.

src/main/java/spoon/MavenLauncher.java
```diff
@@ -108,656 +125,144 @@ public MavenLauncher(String mavenProject, String m2RepositoryPath, SOURCE_TYPE s
...
+			String line;
 
-					if (dependency.getExclusions() != null) {
-						for (int i = 0; i < dependency.getExclusions().size(); i++) {
-							Exclusion exclusion = dependency.getExclusions().get(i);
-							dependence.removeDependency(exclusion.getGroupId(), exclusion.getArtifactId());
-						}
-					}
+			while ((line = output.readLine()) != null) {
+				if (line.contains("Maven home: ")) {
```
surli: Could you use a constant for "Maven home: " please?

src/main/java/spoon/MavenLauncher.java
```diff
@@ -108,656 +125,144 @@ public MavenLauncher(String mavenProject, String m2RepositoryPath, SOURCE_TYPE s
...
-					if (module.model.getVersion() == null) {
-						module.model.setVersion(version);
-					}
-					dependence.addDependence(module.getTreeDependency(isLib, hierarchy));
-				}
-			}
-			return dependence;
+			p.waitFor();
+		} catch (IOException e) {
+			e.printStackTrace();
```
surli: Here again: declaring two exception for the same behaviour is useless. You can catch both in the same block and throw a SpoonException here.

src/main/java/spoon/MavenLauncher.java
```diff
@@ -108,656 +125,144 @@ public MavenLauncher(String mavenProject, String m2RepositoryPath, SOURCE_TYPE s
...
-				}
-			}
-			sb.append("}");
-			return sb.toString();
-		}
+		List<File> classPathPrints = null;
+		try {
+			classPathPrints = Files.find(Paths.get(pom.getParentFile().getAbsolutePath()),
+					Integer.MAX_VALUE,
+					(filePath, fileAttr) -> filePath.endsWith("spoon.classpath.tmp"))
```
surli: Here again: use a constant for the name of the file.

src/main/java/spoon/MavenLauncher.java
```diff
@@ -108,656 +125,144 @@ public MavenLauncher(String mavenProject, String m2RepositoryPath, SOURCE_TYPE s
...
+					.map(p -> p.toFile())
+					.collect(Collectors.toList());
+		} catch (IOException e) {
+			e.printStackTrace();
```
surli: No printstacktrace

src/test/java/spoon/MavenLauncherTest.java
```diff
@@ -17,15 +18,15 @@ public void spoonMavenLauncherTest() {
...
 
 		// 56 because of the sub folders of src/main/java
-		assertEquals(58, launcher.getModelBuilder().getInputSources().size());
+		assertEquals(59, launcher.getModelBuilder().getInputSources().size());
 
 		// with the tests
 		launcher = new MavenLauncher("./", MavenLauncher.SOURCE_TYPE.ALL_SOURCE);
 
-		assertEquals(30, launcher.getEnvironment().getSourceClasspath().length);
+		assertEquals(194, launcher.getEnvironment().getSourceClasspath().length);
```
surli: Could you add a new test to check this new behaviour: you did this PR because on some projects dependency resolution was not working. So could you put pom.xml files that wasn't working previously and that is working now, as a test?

chore/travis/travis-jdk10.sh
nharrand: Yes! So basically it tells `install-jdk.sh` to use the default (the one present in the base conatiner) collection of trusted certificate authority (CA) certificates for java. Without this option, calling maven from java results in SSL errors.

pom.xml
nharrand: Probably not. This is a leftover from previous trials. I'll fix this ASAP.

src/main/java/spoon/MavenLauncher.java
nharrand: Yes. I'll do that.

src/main/java/spoon/MavenLauncher.java
nharrand: Ok. I'll add it ASAP.

src/main/java/spoon/MavenLauncher.java
nharrand: ok

src/main/java/spoon/MavenLauncher.java
nharrand: Yes, indeed.

src/main/java/spoon/MavenLauncher.java
nharrand: Ok.

src/main/java/spoon/MavenLauncher.java
nharrand: So I thought too. But this is the behavior of maven-dependency-plugin for multi module projects. It creates one classpath file per module.

chore/travis/travis-jdk10.sh
surli: ok, can you add a comment on the script please?

src/main/java/spoon/MavenLauncher.java
surli: OK it worthes a comment then

surli: Moreover I'm not sure the method should be public here?

src/main/java/spoon/MavenLauncher.java
surli: Actually you can use a try with resources to be sure it's closed at the end.
Something like:
```
try (br = new BufferedReader(new FileReader(classPathFile));) {
...
}
```

surli: And maybe add the throws IOException to the method itself if it's not public anymore. You'll manage those elsewhere.

src/test/java/spoon/MavenLauncherTest.java
nharrand: It's not that it used to not work on certain poms but more than some types were not resolved because not existing in the classpath where they could have been if MavenLauncher generated a classpath more similar to Maven.
I am unsure it is testable just with a pom. The problem appear when analyzing an AST. 
I could add a test that check that for a maven project (may be spoon itself), all type are resolvable, even those defined in dependencies. WDYT?

chore/travis/travis-jdk10.sh
nharrand: Done.

src/main/java/spoon/MavenLauncher.java
nharrand: Done.

src/test/java/spoon/MavenLauncherTest.java
surli: > I could add a test that check that for a maven project (may be spoon itself), all types are resolvable, even those defined in dependencies. WDYT?

Sounds good yes.

src/main/java/spoon/MavenLauncher.java
surli: > for the user to manually add its classpath

You can still use the `environment.setSourceClasspath`

nharrand: Yes, but his method gives you what content to feed in `environment.setSourceClasspath` should you choose to generate it with maven.

surli: Yeah but the purpose of this method is to be called internally to feed automatically the environment. Then you can use it to improve the classpath by adding/removing some libs.
So I'm still not sure there's a strong reason to let this one open.

nharrand: Ok, I made it private.

src/main/java/spoon/MavenLauncher.java
nharrand: Ok this should be fixed now.

src/test/java/spoon/MavenLauncherTest.java
nharrand: Done.

surli:
thanks @nharrand :)

Issue #2055, opened 2018-06-11
Source position and type cast and brackets


pvojtechovsky:
This problem was found by #1927

In the code like
```java
someFnc(((File) (null)));
```
The source position of `null` literal contains only sources "null", but may be there should be type case `(File)` and may be also brackets. So source position would contain this: `((File) (null))`.

But then I would prefer to have more details in source position. So may be we should introduce `ExpressionSourcePosition`, which would be similar to `DeclarationSourcePosition`. It means  it would have these parts
* start/end -> "((File) (null))"
* typeCastStart/End -> "(File)"
* expressionStart/End -> "(null)"

@tdurieux Do you think it would make sense to introduce new `ExpressionSourcePosition` or we can use `DeclarationSourcePosition` instead here too? Or may be we do not need that at all and source position of expressions is just simple SourcePosition for some reason? 

And what about brackets? Are they part of expression or not?

tdurieux:
> Do you think it would make sense to introduce new ExpressionSourcePosition

That is a good question.
What is the difference with the normal SourcePosition?


> And what about brackets? Are they part of expression or not?

I would say yes, otherwise there are part of nothing, no?

pvojtechovsky:
> What is the difference with the normal SourcePosition?

normal SourcePosition has only 
* start/end of element

while ExpressionSourcePosition will have
* start / end of whole expression including type cast
* typeCastStart/End the type cast part
* expressionStart/End the pure expression part

>> And what about brackets? Are they part of expression or not?

> I would say yes, otherwise there are part of nothing, no?

I agree that braces should be part of expression too (they are actually not it's part.)

pvojtechovsky:
May be I will have time this evening to have a look at this issue. But I am still not sure about concept of the solution.

I will probably create new class `CompoundSourcePositionImpl extends SourcePositionImpl` and will change `DeclarationSourcePositionImpl extends CompoundSourcePositionImpl`.

Where new class `CompoundSourcePositionImpl` will contain `declarationSourceStart` and `declarationSourceEnd`.
So 
CompoundSourcePositionImpl will have these fields:
* sourceStart/End - start end of whole expression including type cast and brackets
* declarationSourceStart/End - start end of expression itself excluding type cast or brackets

WDYT?

tdurieux:
It seems good for me. Thanks

PR #2113, opened 2018-06-26
review2: fix: expression type cast source positions

pvojtechovsky:
fix of #2055

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilder.java
+++ b/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilder.java
@@ -152,6 +152,7 @@ import spoon.reflect.reference.CtArrayTypeReference;
 import spoon.reflect.reference.CtReference;
 import spoon.reflect.reference.CtTypeReference;
 import spoon.reflect.reference.CtUnboundVariableReference;
+import spoon.support.compiler.jdt.ContextBuilder.CastInfo;
 import spoon.support.reflect.CtExtendedModifier;
 
 import java.util.HashSet;
@@ -935,7 +936,10 @@ public class JDTTreeBuilder extends ASTVisitor {
 
 	@Override
 	public boolean visit(CastExpression castExpression, BlockScope scope) {
-		context.casts.add(this.references.buildTypeReference(castExpression.type, scope));
+		CastInfo ci = new CastInfo();
+		ci.nrOfBrackets = ((castExpression.bits >>> 21) & 0xF);
+		ci.typeRef = this.references.buildTypeReference(castExpression.type, scope, true);
+		context.casts.add(ci);
 		castExpression.expression.traverse(this, scope);
 		return false;
 	}
--- a/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilder.java
+++ b/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilder.java
@@ -937,6 +937,7 @@ public class JDTTreeBuilder extends ASTVisitor {
 	@Override
 	public boolean visit(CastExpression castExpression, BlockScope scope) {
 		CastInfo ci = new CastInfo();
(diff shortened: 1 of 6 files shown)


src/main/java/spoon/support/compiler/jdt/JDTTreeBuilder.java
```diff
@@ -935,7 +936,10 @@ public boolean visit(BreakStatement breakStatement, BlockScope scope) {
 
 	@Override
 	public boolean visit(CastExpression castExpression, BlockScope scope) {
-		context.casts.add(this.references.buildTypeReference(castExpression.type, scope));
+		CastInfo ci = new CastInfo();
+		ci.nrOfBrackets = ((castExpression.bits >>> 21) & 0xF);
```
surli: ok this is a low level one ;) Could you explain the operation in a comment please?

src/test/java/spoon/test/position/PositionTest.java
```diff
@@ -768,4 +768,28 @@ public void testArrayArgParameter() throws Exception {
 			assertEquals("String arg[]", contentAtPosition(classContent, param.getType().getPosition()));
 		}
 	}
+	
+	@Test
+	public void testExpressions() throws Exception {
+		//contract: the expression parameter declared like `String arg[]`, `String[] arg` and `String []arg` has correct positions
```
surli: It seems to be a cut/paste contract not related to what you test

monperrus:
do we really to need a new interface? or would it be ok to only add a new implementation class?

pvojtechovsky:
New interface is consistent with other SourcePosition based interfaces. And yes, I will write a code which will check whether source position implements `CompoundSourcePosition`. It is not nice to ask if it implements `CompoundSourcePositionImpl`.

monperrus:
Sometimes, it seems to me that we create too many interfaces, that have no essential semantic
meaning to be modeled, and that only an implementation class and proper polymorphism would be
sufficient.
 
> And yes, I will write a code which will check whether source position implements
> |CompoundSourcePosition|. It is not nice to ask if it implements |CompoundSourcePositionImpl|.
one option would be to add a method "isCompound" in SourcePosition.

pvojtechovsky:
I agree with you, but let's refactor all source position interfaces to keep whole SourcePosition model consistent.... but then it is a lot of changes :-}

monperrus:
OK, so you would go with this one unmodified?

pvojtechovsky:
> OK, so you would go with this one unmodified?

yes, I prefer to keep one same pattern (iface + class) for all SourcePosition interfaces first. If we decide that this pattern is wrong in this case, then we should refactor it on all places where it occurs. I agree with this refactoring, but it should be different PR.

monperrus:
rebase?

pvojtechovsky:
rebase done

src/main/java/spoon/support/compiler/jdt/JDTTreeBuilder.java
```diff
@@ -935,7 +936,11 @@ public boolean visit(BreakStatement breakStatement, BlockScope scope) {
 
 	@Override
 	public boolean visit(CastExpression castExpression, BlockScope scope) {
-		context.casts.add(this.references.buildTypeReference(castExpression.type, scope));
+		CastInfo ci = new CastInfo();
+		//the 8 bits from 21 to 28 represents number of enclosing brackets
```
surli: Just for my information, because I'm curious ;) It's something you obtained by reverse engineering or there's a documentation somewhere with that info?

surli:
LGTM

src/main/java/spoon/support/compiler/jdt/JDTTreeBuilder.java
pvojtechovsky: There is a documentation of these bits in source code of compiler ... and I needed some luck to found it ;-)

PR #2115, opened 2018-06-26
feat(position): CompoundSourcePosition extends SourcePosition

### LINKED ISSUE #2055: Source position and type cast and brackets

pvojtechovsky:
first baby step to fix #2055 

* CompoundSourcePosition extends SourcePosition
* DeclarationSourcePosition extends CompoundSourcePosition
* part of fields and code of DeclarationSourcePositionImpl was moved to CompoundSourcePositionImpl

### DIFF
--- a/src/main/java/spoon/reflect/cu/position/CompoundSourcePosition.java
+++ b/src/main/java/spoon/reflect/cu/position/CompoundSourcePosition.java
@@ -19,7 +19,9 @@ package spoon.reflect.cu.position;
 import spoon.reflect.cu.SourcePosition;
 
 /**
- * This interface represents the position of a program element in a source file.
+ * This interface represents the position of a program element like an expression in a source file.
+ * The start/end represents range of whole program element including children element and comments
+ * The nameStart/End represents range of core part of program element.
  */
 public interface CompoundSourcePosition extends SourcePosition {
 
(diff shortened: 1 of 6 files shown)


surli:
I assume this is the first one to review before #2113 right? is it ready?

src/main/java/spoon/reflect/cu/position/CompoundSourcePosition.java
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * The fact that you are presently reading this means that you have had
+ * knowledge of the CeCILL-C license and that you accept its terms.
+ */
+package spoon.reflect.cu.position;
+
+import spoon.reflect.cu.SourcePosition;
+
+/**
+ * This interface represents the position of a program element in a source file.
```
surli: The javadoc here is exactly the same that for `DeclarationSourcePosition`. Could you edit one of those Javadoc to explicit the differences between the interfaces?

pvojtechovsky:
> is it ready?

yes, I am done here.

PR #2116, opened 2018-06-26
refactor: make checkstyle happy

zielint0:
Fixed plenty checkstyle violations.


pvojtechovsky:
please discuss with @monperrus first whether we want to have checkstyle in tests too. Actually we enforce checkstyle rules only in runtime code. Test are not checked.
... it might be a lot of effort to fix that ... and it would be shame to waste your effort then.

zielint0:
@pvojtechovsky
@monperrus 
No worries, I can deal with it quickly! :)
The effort is really small and I like it. :)

surli:
Yes actually we don't really care about the checkstyle of the tests: so if you want to take time to fix this, good, but we won't enforce it.
 as @pvojtechovsky stated: 
> ... it might be a lot of effort to fix that ... and it would be shame to waste your effort then.

But here again, as you wish.

monperrus:
conflict here

zielint0:
@monperrus 
Resolved.

PR #2128, opened 2018-06-27
test: CtCatch source position

pvojtechovsky:
@tdurieux do you agree that the test added by this PR should pass?

```java
CtTry tryStatement = ...;
assertEquals("catch (IOException e) {\n" + 
		"			throw new RuntimeException(e);\n" + 
		"		}", contentAtPosition(classContent, tryStatement.getCatchers().get(0).getPosition()));
```

note: actually CtCatch source position contains only from parameters. E.g.: `IOException e`, which is not correct.

WDYT?

I will try to fix it in next days if you agree.


tdurieux:
Yes, I agreed.

to get the catch position you should look at the jdt parent element.
I kind of did it for the catch variables

tdurieux:
It seems good to me. Thanks @pvojtechovsky

pvojtechovsky:
I am done here.

src/main/java/spoon/support/compiler/jdt/PositionBuilder.java
```diff
@@ -156,18 +162,45 @@ SourcePosition buildPositionCtElement(CtElement e, ASTNode node) {
...
+				int endOfTry = tryStatement.getPosition().getSourceEnd();
+				//offset of the bracket before catch
+				int lastBracket = getEndOfLastTryBlock(tryStatement, 0);
+				int catchStart = findNextNonWhitespace(contents, endOfTry, lastBracket + 1);
+				if (CATCH.equals(new String(contents, catchStart, CATCH.length())) == false) {
+					throw new SpoonException("Unexpected beginning of catch statement on offset: " + catchStart);
+				}
+				int bracketStart = findNextNonWhitespace(contents, endOfTry, catchStart + CATCH.length());
+				if (bracketStart < 0) {
+					throw new SpoonException("Unexpected end of file instead of \'(\' after catch statement on offset: " + catchStart);
```
tdurieux: I'm afraid of this exception because there is a chance that we forget to handle one specific case and it will break the construction of the model.

pvojtechovsky: What do you suggest? To silently ignore our mistake ("we forget to handle one specific case") and to produce invalid source position?

tdurieux: I don't know what is the best solution.

tdurieux: I meant skipSelfChecks

pvojtechovsky: ok, I will do so

PR #2136, opened 2018-06-28
[wip] Refactor: fix checkstyle violations

zielint0:
I fixed some checkstyle violations.

Also small additional fix - take a look on:
https://github.com/INRIA/spoon/pull/2136/commits/89e053fea0791943d2fd1b6062a3d5eaa7c6d0bb
I refactored
```
    if (ac == Override.class || ac == SuppressWarnings.class || ac == Root.class) {
        return false;
    }
    return true;
```

to
```
return ac != Override.class && ac != SuppressWarnings.class && ac != Root.class;
```

to improve its readability. Previous version is unwieldy, long and hard to read.

### DIFF
--- a/src/test/java/spoon/MavenLauncherTest.java
+++ b/src/test/java/spoon/MavenLauncherTest.java
@@ -3,7 +3,6 @@ package spoon;
 import org.apache.commons.lang3.StringUtils;
 import org.junit.Test;
 
-import java.io.File;
 import java.nio.file.Paths;
 import java.util.Arrays;
 import java.util.List;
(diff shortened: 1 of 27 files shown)


surli:
One reason why we don't apply checkstyle on the tests, is that a lot of our classes in the tests are actually code resources.
We do not put everything in resources because we generally want to ensure that those resources can be compiled, and also by lazyness let's be honest here.

My point is, those resources shouldn't be refactored, even for checkstyle: some unused imports we might have add on those files, or unused methods/fields are there purposely. We generally place those kind of resources in a package `testclasses` but it's not always the case.

So here, if you really want to refactor the tests, I think there should be a discussion first on how to do it. I will open an issue on that direction.

zielint0:
Thank you for this insightful comment. Looks like I am going too fast. Sorry for that!

Did I break anything in
- https://github.com/INRIA/spoon/pull/2127
- https://github.com/INRIA/spoon/pull/2116
?

surli:
I did check them back before commenting the issue, and it's fine as you were really modifying test suites, not test resources :)

zielint0:
Tension released.

zielint0:
I did it.
In my first PR I broke a test class!
https://github.com/INRIA/spoon/pull/2102
The test error was so strange that I was sure that something went horribly wrong and I closed my PR myself quickly.

I appreciate our CI+test suite :-)

surli:
So I close this one for now.

PR #2141, opened 2018-06-29
review fix: launcher#setArgs cannot be called twice

surli:
This PR shows that the current behaviour is to never reset the model, even if the same launcher is called twice with different arguments.

My current test shows that `model` and `model2` are the same and they contain 2 types at the end of the test.

I don't know if it's an expected behaviour or not, but it looks very dangerous to me. WDYT?

### DIFF
--- a/src/main/java/spoon/Launcher.java
+++ b/src/main/java/spoon/Launcher.java
@@ -99,6 +99,8 @@ public class Launcher implements SpoonAPI {
 	private List<String> processorTypes = new ArrayList<>();
 	private List<Processor<? extends CtElement>> processors = new ArrayList<>();
 
+	private boolean processed;
+
 	/**
 	 * A default program entry point (instantiates a launcher with the given
 	 * arguments and calls {@link #run()}).
@@ -123,6 +125,11 @@ public class Launcher implements SpoonAPI {
 
 	public void setArgs(String[] args2) {
 		this.commandLineArgs = args2;
+		if (processed) {
+			throw new SpoonException("You cannot process twice the same launcher instance.");
+		}
+		processed = true;
+
 		processArguments();
 	}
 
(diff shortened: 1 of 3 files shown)


surli:
By the way, the only way to make this case working is by using `setArgs()` because then we call `processArguments` which create a new `JDTBasedSpoonCompiler`. If we try doing that using the API (`addInputResource` and so one), it will never work because we don't create back the `JDTBasedSpoonCompiler`. 

Personally what I would want as a user is:
  1. To obtain a new model each time I call `buildModel()`
  2. To be able to call `buildModel()` several times for the same launcher

pvojtechovsky:
I would say they spoon model is not ready to coexist with another spoon model made by same launcher. Note that each model produced by launcher actually MUST share `Factory` and  `StandardEnvironment` and it causes at least problems with change event handling ... 

So it looks like that idea needs deeper refactoring

surli:
> I would say they spoon model is not ready to coexist with another spoon model made by same launcher.

Ok then we should prevent calling twice `setArgs` on the same launcher by throwing an exception for example. Do you agree?

pvojtechovsky:
> we should prevent calling twice setArgs on the same launcher by throwing an exception

I agree ... and later we might think about that refactoring. It is not nice when model specific things are in StandartEnvironment...

src/main/java/spoon/Launcher.java
```diff
@@ -99,6 +99,8 @@
 	private List<String> processorTypes = new ArrayList<>();
 	private List<Processor<? extends CtElement>> processors = new ArrayList<>();
 
+	private boolean processed;
```
monperrus: explicitly set it to false?

monperrus: add a comment about the meaning of the field?

PR #2142, opened 2018-06-30
fix: position of nested type arguments like List<List<?>>

pvojtechovsky:
fix of another problem found by Sniper mode tests

### DIFF
--- a/src/test/java/spoon/test/position/PositionTest.java
+++ b/src/test/java/spoon/test/position/PositionTest.java
@@ -111,8 +111,7 @@ public class PositionTest {
 	@Test
 	public void testPositionParameterTypeReference() throws Exception {
 		//contract: the parameterized type reference has a source position which includes parameter types, etc.
-		final Factory build = build(new File("src/test/java/spoon/test/position/testclasses/"));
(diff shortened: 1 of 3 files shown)


pvojtechovsky:

src/test/java/spoon/test/position/PositionTest.java
```diff
@@ -126,6 +125,10 @@ public void testPositionParameterTypeReference() throws Exception {
 		CtTypeReference<?> field3Type =  foo.getField("field3").getType();
 		//this probably points to an bug in JDT. But we have no workaround in Spoon, which handles spaces and comments too
 		assertEquals("List<T // */ >\n\t/*// */>", contentAtPosition(classContent, field3Type.getPosition()));
+
+		CtTypeReference<?> field4Type =  foo.getField("field4").getType();
+		//this probably points to an bug in JDT. But we have no workaround in Spoon, which handles spaces and comments too
```
tdurieux: Did you copy the comment on purpose?

pvojtechovsky: thanks for pointing to that. It was not intentional

PR #841, opened 2016-09-21
feat(ast): Creates CtTypeMember list in CtType.

GerardPaligot:
Closes #681

PR #2143, opened 2018-06-30
fix(performance): skip check in building of model CtTypeImpl#addMethod

pvojtechovsky:
This check takes about 25% of spoon model build time. Does it really makes sense to check it???

Or may be it should be by default disabled and we can enable it only in our tests?

WDYT?

### DIFF
--- a/src/main/java/spoon/support/reflect/declaration/CtTypeImpl.java
+++ b/src/main/java/spoon/support/reflect/declaration/CtTypeImpl.java
@@ -624,9 +624,9 @@ public abstract class CtTypeImpl<T> extends CtNamedElementImpl implements CtType
 					removeTypeMember(m);
 				} else {
 					// checking contract signature implies equal
-					if (!factory.getEnvironment().checksAreSkipped() && m.equals(method)) {
-						throw new AssertionError("violation of core contract! different signature but same equal");
-					}
+//					if (!factory.getEnvironment().checksAreSkipped() && m.equals(method)) {
+//						throw new AssertionError("violation of core contract! different signature but same equal");
+//					}
 				}
 			}
 		}


zielint0:
Tests are our friends. Don't kill friends ;-)

surli:
> Or may be it should be by default disabled and we can enable it only in our tests?

Keeping commented code seems useless: it might make sense to keep it deactivated by default, and to allow activating it by an internal flag.

surli:
@monperrus @tdurieux ok to merge?

tdurieux:
I don't know if the environment is the good place for SKIP_CHECK_CONTRACT_SIGNATURE

surli:
You'd prefer in CtTypeImpl directly?

pvojtechovsky:
The commented code should be removed or uncommented and wrapped in an if with an switch. So this PR is actually here just to point to the problem. It is not best for direct merging.

I vote for removing, because I did not understood the purpose/usefulness of this check at all. What kind of error it might found?

surli:
> or uncommented and wrapped in an if with an switch

Actually I edited the branch to do that ;)

> I vote for removing, because I did not understood the purpose/usefulness of this check at all. What kind of error it might found?

Apparently it has been added in #841 which was a huge refactoring: not sure why exactly it has been introduced. To me it looks more as something to add in our unit test than something to keep in the code. @tdurieux @monperrus WDYT on this one?

tdurieux:
If I correctly remember, we had this test because we had some collision bug in the signature.
But apparently, we don't face this problem anymore. I think it is safe to remove it if after two years nobody enters it.

pvojtechovsky:
@surli do you agree to remove that code too? Martin, Thomas and me seems to agree with remove of this code.

surli:
Yes ok for me to remove it.

pvojtechovsky:
This PR is ready for merge. I have simply deleted the method not equals checking code

PR #2150, opened 2018-07-01
fix(position): position of CtCase includes it's statements

pvojtechovsky:


pvojtechovsky:

PR #2156, opened 2018-07-02
fix(position): CtCatch has no modifiers, they are in CatchVariable

pvojtechovsky:
The modifiers were on two places before
1) CtCatch - wrong
2) CtCactchVariable - OK

So this PR removes them from CtCatch

Note: CtCatch would need a bettet source position structure. I actually I used BodyHolderSourcePosition, but it is may be not fitting well here. WDYT? Can we live with that or we make new structure for that?

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/PositionBuilder.java
+++ b/src/main/java/spoon/support/compiler/jdt/PositionBuilder.java
@@ -445,13 +445,15 @@ public class PositionBuilder {
 		CtTry tryElement = catcher.getParent(CtTry.class);
 		//offset after last bracket before catch
 		int declarationStart = getEndOfLastTryBlock(tryElement, 1) + 1;
-		DeclarationSourcePosition oldCatcherPos = (DeclarationSourcePosition) catcher.getParameter().getPosition();
+		DeclarationSourcePosition paramPos = (DeclarationSourcePosition) catcher.getParameter().getPosition();
 		int bodyStart = catcher.getBody().getPosition().getSourceStart();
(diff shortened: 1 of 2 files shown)


tdurieux:
Which position information you would like to add to CtCatch?

monperrus:
LGTM

pvojtechovsky:
> Which position information you would like to add to CtCatch?

catch has two child roles: parameter and body. So in ideal situation I need a source position with three pairs of start/end
1) whole catch element
2) parameter related part
3) body related part

But may be we might be able to detect these child fragments from positions of children. In that case the catch might be represented by simple SourcePositionImpl with one start/end pair, which represents whole catch element.

WDYT?

We can improve/change it later

tdurieux:
> 1. whole catch element
> 1. parameter related part
> 1. body related part

I think that these three elements will be useful for the sniper mode.
I think it is better to have those positions.
Do you think we should create a ParameterHolderSourcePosition that will be used for the catch and the try with resources?

pvojtechovsky:
> I think that these three elements will be useful for the sniper mode.

I don't know. May be. Let's play with sniper mode code for some time and gather some experience and then we will see what we really need. Note sniper mode uses SourceFragments which are based on information from SourcePosition. And SourceFragments might probably easily extract some source positions from child elements too.

But may be we should collect the list of Spoon elements whose source position is not fitting well to current SourcePosition model. I see these:
* try with resource
* catch
* switch and case and default
* ... ?
Then we will see what SourcePosition structures are missing ... if any

PR #2159, opened 2018-07-03
fix: Invalid end column

Egor18:
Hi! I've found out that `getEndColumn()` from `CompoundSourcePositionImpl` works incorrectly.
Here are 2 screenshots to illustrate the problem.    
Now:
not_ok
Expected:
ok

The problem is that only `getEndLine()` has been overriden, but not `getEndColumn()`.
So this PR contains fix and additional test.


Egor18:
Without this fix `assertEquals(2, pos.getEndColumn());` fails, because `pos.getEndColumn()` returns 14 (wrong).

surli:
LGTM @tdurieux ok to merge?

tdurieux:
yes

PR #2161, opened 2018-07-03
WiP fix: StackOverflow when checking overriding with generics

### LINKED ISSUE #2160: Bug: Overriding a method with self referential generic leads to StackOverflowException
Given:
```
public class A {
	<T extends Enum<T>> void m6(T t) {
	}
}
public class B extends A {
	@Override
	<S extends Enum<S>> void m6(S s) {
	}
}
```
Checking if `B#m6` overrides `A#m6` leads to a `StackOverflowException`.
I could manage to fix that in `spoon.support.visitor.MethodTypingContext#isSameMethodFormalTypeParameter` but there seams to be another issue when adopting the type (called in `spoon.support.visitor.ClassTypingContext.ExecutableContext#isSameSignatureLikeScopeMethod`, line 696). 

May be it is due to the fact, that a type definition like `S extends Enum<S>` is stored in the model like this:

0: CtTypeParameter (`S extends Enum<S>`)
1: --> superClass: CtTypeReference (`Enum<S>`)
2: ----> actualTypeArguments[0]: CtTypeParameterReference (`S`)
3: ------> superType: CtTypeReference (`Enum<S>`)
4: --------> actualTypeArguments[0]: CtTypeParameterReference (`S`)
5: ----------> superType: CtTypeReference (`java.lang.Object`)

The `spoon.support.visitor.AbstractTypingContext#adaptType` method adapts a "level 0" type to a "level 2" type, which then not equals.

surli:
fix #2160

### DIFF
--- a/src/main/java/spoon/support/visitor/MethodTypingContext.java
+++ b/src/main/java/spoon/support/visitor/MethodTypingContext.java
@@ -229,73 +229,53 @@ public class MethodTypingContext extends AbstractTypingContext {
 		//the methods has same count of formal parameters
 		//check that bounds of formal type parameters are same after adapting
 		for (int i = 0; i < thisTypeParameters.size(); i++) {
-			if (isSameMethodFormalTypeParameter(thisTypeParameters.get(i), thatTypeParameters.get(i)) == false) {
+			if (isSameFormalTypeParameter(getBound(thisTypeParameters.get(i)), getBound(thatTypeParameters.get(i))) == false) {
 				return false;
 			}
 		}
 		return true;
 	}
 
-	private boolean isSameMethodFormalTypeParameter(CtTypeParameter scopeParam, CtTypeParameter superParam) {
-		CtTypeReference<?> scopeBound = getBound(scopeParam);
-		CtTypeReference<?> superBound = getBound(superParam);
-		int idxOfScopeBoundTypeParam = getIndexOfTypeParam(scopeParam.getTypeParameterDeclarer(), scopeBound);
-		if (idxOfScopeBoundTypeParam >= 0) {
-			int idxOfSuperBoundTypeParam = getIndexOfTypeParam(superParam.getTypeParameterDeclarer(), superBound);
-			if (idxOfSuperBoundTypeParam >= 0) {
-				/*
-				 * Both type parameters have bound defined as sibling type parameter.
-				 * Do not try to adaptType, because it would end with StackOverflowError
-				 * They are same formal type parameters if index is same
-				 */
-				return idxOfScopeBoundTypeParam == idxOfSuperBoundTypeParam;
-			}
-		}
-		if (scopeBound.getActualTypeArguments().size() != superBound.getActualTypeArguments().size()) {
-			return false;
-		}
-
-		for (int i = 0; i < scopeBound.getActualTypeArguments().size(); i++) {
-			CtTypeReference scopeBoundATArg = scopeBound.getActualTypeArguments().get(i);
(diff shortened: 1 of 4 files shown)


surli:
@pvojtechovsky this one is related to `adaptType` do you think you can fix it quickly? 

It looks complicated to me and I want to know if I wait for a fix or if I release Spoon before the fix.

src/main/java/spoon/support/visitor/MethodTypingContext.java
```diff
@@ -267,6 +268,25 @@ private boolean isSameMethodFormalTypeParameter(CtTypeParameter scopeParam, CtTy
 				if (scopeBoundWildcard.getBoundingType().equals(scopeMethod.getFactory().Type().getDefaultBoundingType())) {
 					return true;
 				}
```
vidl: Shouldn't the check above (line 241, idxOfScopeBoundTypeParam...) handle this already? The comment seems to indicate that (but it does not work for this case, though).

vidl: ok - I had a quite similar solution but trapped then in the "adaptType" issue, which comes later, when the StackOverflow is fixed.

pvojtechovsky:
I will have a look at it tomorrow morning. Quickly? I do not know yet.

pvojtechovsky:
I did some improvement ... I am not sure it is correct. It needs to be checked by automated tests... and there is still a conceptual problem. I see no simple and clear solution for that.

The new test you added still fails on `reference.equals(reference)`, where reference is
A) S extends List<S extends List<S extends Object>>
B) S extends List<S extends Object>

the problem is on conceptual level. I think that Spoon must never fill superType on reference, when this superType is not declared on reference but is declared on definition of the type

Example
A) void method(List<? extends Double> p)
B) <T extends Double> void method(T p)

In case A) the supertype is on reference so it must be filled in
In case B) the supertype of type of parameter P must by not set, because it is define somewhere else.
... or something like that ;-)

pvojtechovsky:
@surli it is about year ago when you added the concept that **each** CtTypeParameterReference has filled in super type. I never liked it, but I did not had good argument why it is wrong. Now that circular reference problems shows that this concept is wrong.

In this example
A) `void method(List<? extends Double> p)`
B) `<T extends Double> void method(T p)`

the CtTypeParameterReference in case of A is needed. It is primarily defined on that reference.
In case B the type parameter reference of type of parameter `p` has also superType, but it SHOULD be DERIVED property in this case. It is bad idea to duplicate this information and to physically store it into superType property.

WDYT?

surli:
> the CtTypeParameterReference in case of A is needed. It is primarily defined on that reference.
In case B the type parameter reference of type of parameter p has also superType, but it SHOULD be DERIVED property in this case. It is bad idea to duplicate this information and to physically store it into superType property.

I see your point but if I'm right (not sure there...) in both cases we are dealing with `CtTypeParameterReference` so currently we don't have a proper way to know that it should be derived or not.

So what you seem to suggest is to distinguish those cases in dealing with two different kind of metamodel elements? Or do you suggest something different?

pvojtechovsky:
> to distinguish those cases 
yes we should clearly distinguish between these two cases

> in dealing with two different kind of metamodel elements
it is one of the ways how to implement it. 
Yes, I like the idea of new spoon model element which would model that case.

pvojtechovsky:
I will try to solve it in next days

surli:
I'm closing this one since I think @pvojtechovsky fixed it in #2406

PR #2165, opened 2018-07-03
review: Use Apache Common Compressor API to [de]compress serialized models

mehdi-kaytoue:
Since #2094 and #2103 we can [de]serialize the model (CtFactory) with standard java serialization in a file, or a gzip file to save space. Choice was made to write things so that we can add new compression types easily. 

In this PR, I propose to use Apache Common Compressor API (+other files needed for some compression types such as LZMA).

Main changes in src/main/java/spoon/support/SerializationModelStreamer.java and its junit test case src/test/java/spoon/test/serializable/ModelStreamerTest.java and of course pom.xml.

I don't know what you think about adding a new lib, but this lib allows very easily (i) to test what compressor is used for an inputstream (except 7zip btw), (ii) to decompress, (iii) to compress.

Looking forward discussing it with you guyz :)
Cheers,
Mehdi


pvojtechovsky:
it looks like you added sources (org.tukaani.xz.**) of this library into Spoon sources? It is not good idea. It would be very hard to maintain Spoon then. It should be enough to add dependency on that library.  Then we can update version of org.tukaani.xz just by change of version number

mehdi-kaytoue:
Sorry, for some obscure reason I thought it was not on maven central ... done!

surli:
Could you tell us how much the new libraries changes the jar file size: is it adding few Kb to the final jar with dependencies or is it adding some Mb? 

I don't really know if there's a real need for that feature, so we should be careful when adding new libraries about the impact on Spoon size.

monperrus:
@mehdi-kaytoue could you fix Travis so that we can merge? Thanks.

mehdi-kaytoue:
Sorry for (very) late reply. The activity lately woke me up! 
I let it aside as I did not know exactly how to do it @surli .
What did you decide? Do you think this feature is interesting?

msteinbeck:
Regarding external libraries. FST might be interesting: https://github.com/RuedigerMoeller/fast-serialization It speeds up serialization and is extremely helpful when dealing with very large models.


@mehdi-kaytoue Did you see the note about multi-threading in org.tukaani.xz: `Single-threaded streamed compression and decompression and random access decompression have been fully implemented. Threading is planned but it is unknown when it will be implemented.`. So I guess it performs very bad in terms of speed. My experience with BZIP2 is that it produces noticeably smaller files, but takes considerably more time to compress the byte stream. However, unlike LZMA (due to missing multi-threading) it might be worthwhile.

monperrus:
> Regarding external libraries. FST might be interesting: RuedigerMoeller/fast-serialization

Yes, it's interesting, probably in a future PR.

With no further comments, this PR seems ready to merge.

surli:
Thanks @mehdi-kaytoue for this PR

PR #2169, opened 2018-07-04
Sonarqube-repair - add no-argument constructor in SubFactory

ashutosh1598:
This PR repairs [three sonarqube bugs](https://sonarqube.ow2.org/project/issues?id=fr.inria.gforge.spoon%3Aspoon-core&resolved=false&rules=squid%3AS2055&types=BUG). Classes ModuleFactory, PackageFactory, and DefaultCoreFactory extend SubFactory and are Serializable. But SubFactory is not Serializable and should contain a no-argument constructor according to [Sonar Rule 2055](https://rules.sonarsource.com/java/type/Bug/RSPEC-2055).

Even though I enabled comments, spoon did not pick up the commented out code and hence this diff. But that block is commented since 5 years and is itself marked as a [code smell](https://sonarqube.ow2.org/project/issues?id=fr.inria.gforge.spoon%3Aspoon-core&issues=AWIvav2IRIyoeYNsnf03&open=AWIvav2IRIyoeYNsnf03) by Sonarqube.

### DIFF
--- a/src/main/java/spoon/reflect/factory/SubFactory.java
+++ b/src/main/java/spoon/reflect/factory/SubFactory.java
@@ -23,7 +23,10 @@ package spoon.reflect.factory;
  */
 public abstract class SubFactory {
 
-	protected Factory factory;
+	public SubFactory() {
+}
+
+protected Factory factory;
 
 	/**
 	 * The sub-factory constructor takes an instance of the parent factory.
@@ -32,9 +35,6 @@ public abstract class SubFactory {
 		super();
 		this.factory = factory;
 	}
-
-public SubFactory() {
-}
 
 }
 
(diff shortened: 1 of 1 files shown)


pvojtechovsky:
I would say that it is a bug that some Spoon SubFactory classes are serializable. We never serialize them, but we attach new instance of these factories to deserialized Factory.

monperrus:
Is that exactly the sniper-mode pretty-printing?

(it is interesting that the commented method is completely removed by the sniper)

ashutosh1598:
Yes this was the exact file printed by sniper mode.

monperrus:
Could you prepare an alternative PR with the other solution (removing "implements Serializable") so that we can compare the impact?

ashutosh1598:
Yes I'm doing that right now.

ashutosh1598:
Fixed in https://github.com/INRIA/spoon/pull/2173

PR #2172, opened 2018-07-05
WIP feat: javadoc {@link} can be made fully-qualified

### LINKED ISSUE #2170: regression: in FQN mode comments are printed as the original ones which might cause compilation errors
See: https://ci.inria.fr/sos/job/JUnit/1081/console
We got a class like this: 
```java
/**
 *
 *
 * @deprecated Included for backwards compatibility with JUnit 4.4. Will be
 * removed in the next major release. Please use
 * {@link BlockJUnit4ClassRunner} in place of {@link JUnit4ClassRunner}.
 */
@java.lang.Deprecated
public class MethodValidator {
```

instead of:
```java
/**
 *
 *
 * @deprecated Included for backwards compatibility with JUnit 4.4. Will be
 * removed in the next major release. Please use
 * {@link org.junit.runners.BlockJUnit4ClassRunner} in place of {@link JUnit4ClassRunner}.
 */
@java.lang.Deprecated
public class MethodValidator {
[...]
```

surli:
Requirements implemented:
* <s>R1: add support for getting the original comments</s> done in #2746 
* <s>R2: the original comment can be analyzed from the model and from the serialized model, even if the Java files don't exist anymore on disk)</s> abandoned 
* <s>R3: the @link fragments can be analyzed done</s> in #2802
* R4: javadoc {@link} can be made fully-qualified here

fix #2170

### DIFF
--- a/src/main/java/spoon/support/reflect/code/CtJavaDocImpl.java
+++ b/src/main/java/spoon/support/reflect/code/CtJavaDocImpl.java
@@ -19,13 +19,16 @@ package spoon.support.reflect.code;
 import spoon.javadoc.internal.Javadoc;
 import spoon.javadoc.internal.JavadocBlockTag;
 import spoon.javadoc.internal.JavadocDescriptionElement;
+import spoon.javadoc.internal.JavadocInlineTag;
 import spoon.reflect.annotations.MetamodelPropertyField;
 import spoon.reflect.code.CtComment;
 import spoon.reflect.code.CtJavaDoc;
 import spoon.reflect.code.CtJavaDocTag;
 import spoon.reflect.declaration.CtElement;
+import spoon.reflect.declaration.CtType;
 import spoon.reflect.path.CtRole;
 import spoon.reflect.visitor.CtVisitor;
+import spoon.reflect.visitor.filter.NamedElementFilter;
 import spoon.support.util.ModelList;
 
 import java.util.List;
@@ -126,6 +129,9 @@ public class CtJavaDocImpl extends CtCommentImpl implements CtJavaDoc {
 			addTag(getFactory().createJavaDocTag(tag.getContent().toText(), CtJavaDocTag.TagType.tagFromName(tag.getTagName())));
 		}
 
+		// if we work in fully-qualified mode, we also fully qualify the {@link}
+		updatedInlineLinksInJavadoc();
+
 		// we cannot call super.setContent because it calls cleanComment (which has already been done above)
 		// and we don't want to clean the comment twice
 		String contentWithTags = javadoc.getDescription().toText().trim(); // trim is required for backward compatibility
@@ -137,6 +143,37 @@ public class CtJavaDocImpl extends CtCommentImpl implements CtJavaDoc {
 	}
 
 
+	/** transforms all {@link XXX} into fully qualified name */
+	private void updatedInlineLinksInJavadoc() {
+		if (this.getFactory().getEnvironment().isAutoImports() == true) {
+			return;
+		}
+		for (JavadocDescriptionElement fragment : getJavadocElements()) {
+			if (fragment instanceof JavadocInlineTag) {
+				JavadocInlineTag tag = (JavadocInlineTag) fragment;
+				if (tag.getType().equals(JavadocInlineTag.Type.LINK)) {
+					String stype;
+					String suffix= "";
+					if (tag.getContent().contains("#")) {
+						// link to a method
+						String[] split = tag.getContent().split("#");
+						stype = split[0];
+						suffix = "#" + split[1];
+					} else {
+						// link to a class
+						stype = tag.getContent();
+					}
+
+					// is there such a type in the model?
+					CtType type = this.getFactory().getModel().filterChildren(new NamedElementFilter<>(CtType.class, stype)).first(CtType.class);
+					if (type != null) {
+						// we write it fully qualified
+						tag.setContent(type.getQualifiedName() + suffix);
+					}
+				}
+			}
+		}
+	}
 
 	@Override
 	public String getLongDescription() {
--- a/src/main/java/spoon/support/reflect/code/CtJavaDocImpl.java
+++ b/src/main/java/spoon/support/reflect/code/CtJavaDocImpl.java
@@ -153,7 +153,7 @@ public class CtJavaDocImpl extends CtCommentImpl implements CtJavaDoc {
 				JavadocInlineTag tag = (JavadocInlineTag) fragment;
 				if (tag.getType().equals(JavadocInlineTag.Type.LINK)) {
 					String stype;
-					String suffix= "";
+					String suffix = "";
 					if (tag.getContent().contains("#")) {
 						// link to a method
 						String[] split = tag.getContent().split("#");
--- a/src/test/java/spoon/test/comment/CommentTest.java
+++ b/src/test/java/spoon/test/comment/CommentTest.java
@@ -58,10 +58,14 @@ import spoon.reflect.declaration.CtType;
 import spoon.reflect.declaration.ModifierKind;
 import spoon.reflect.factory.Factory;
 import spoon.reflect.factory.FactoryImpl;
+import spoon.reflect.path.CtPath;
+import spoon.reflect.path.CtPathStringBuilder;
 import spoon.reflect.visitor.filter.AbstractFilter;
+import spoon.reflect.visitor.filter.NamedElementFilter;
 import spoon.reflect.visitor.filter.TypeFilter;
 import spoon.support.DefaultCoreFactory;
 import spoon.support.JavaOutputProcessor;
+import spoon.support.SerializationModelStreamer;
 import spoon.support.StandardEnvironment;
 import spoon.support.compiler.jdt.JDTSnippetCompiler;
 import spoon.test.comment.testclasses.BlockComment;
@@ -158,6 +162,7 @@ public class CommentTest {
 		Factory f = getSpoonFactory();
 		CtClass<?> type = (CtClass<?>) f.Type().get(OtherJavaDoc.class);
 		CtJavaDoc classJavaDoc = (CtJavaDoc) type.getComments().get(0);
+		assertEquals("A short description without a proper end", classJavaDoc.getContent());
 		assertEquals("A short description without a proper end", classJavaDoc.getShortDescription());
 		assertEquals("A short description without a proper end", classJavaDoc.getLongDescription());
 	}
@@ -1024,6 +1029,34 @@ public class CommentTest {
 	}
 
 	@Test
+	public void testCommentWithLinkInFQNMode() {
+		// contract: in FQN mode links should be written in FQN too
+		Launcher launcher = new Launcher();
+		launcher.addInputResource("./src/test/java/spoon/test/comment/testclasses/JavadocLinkComment.java");
+		launcher.addInputResource("./src/main/java/spoon/support/reflect/reference/CtTypeMemberWildcardImportReferenceImpl.java");
+		launcher.addInputResource("./src/main/java/spoon/reflect/declaration/CtElement.java");
+		launcher.getEnvironment().setCommentEnabled(true);
+		launcher.getEnvironment().setAutoImports(false);
+		CtModel model = launcher.buildModel();
+
+		CtClass<?> javadocLinkComment = model.getElements(new NamedElementFilter<>(CtClass.class, "JavadocLinkComment")).get(0);
+
+		List<CtComment> comments = javadocLinkComment.getComments();
+		assertEquals(1, comments.size());
+
+		CtComment comment = comments.get(0);
+		String commentString = comment.toString();
+
+		assertTrue(commentString, commentString.contains("{@link spoon.support.reflect.reference.CtTypeMemberWildcardImportReferenceImpl}"));
+
+		assertTrue(commentString, commentString.contains("{@link spoon.reflect.declaration.CtElement#toString()}"));
+
+		CtMethod<?> method = javadocLinkComment.getMethodsByName("method").get(0);
+		comments = method.getComments();
+		assertEquals(1, comments.size());
+
+	}
+
 	public void testStatementComments() {
 		// contract: the statements have their comment even if they are nested in another block
 		Launcher launcher = new Launcher();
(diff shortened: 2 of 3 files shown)


monperrus:
first version with parsing code from JavaParser.

pvojtechovsky:
@monperrus, It is great that you are working on model of javadoc comment content!

monperrus:
Work:
* add support for parsing Javadoc comments, with a mix of Thomas and JavaParser code
  * Thomas code to clean comments (remove the prefix " * "), this maximizes backward compatibility (esp in WildComments, which is indeed a wild class)
  * Javaparser code to parse tags and descriptions, this is mature code which has already been debugged.
* add method "getRawComment()" for clients who want to do their own parsing (this was impossible before)
* simplication: removed brittle and hard to maintain  post-processing of comments (many bugs were due to things inconsistently handled between pre-parsing or post-parsing). We now have the sniper mode for those who want to keep the comments "as is". 
* unification: `CtElementImpl#getDocComment()` and `CommentHelper#printComment()` is now based on the same code.
* in fqn mode, rewrite `@link`, fix #2170

monperrus:
merged. CI OK.

pom.xml
```diff
@@ -622,6 +622,10 @@
               <!-- We also want to check the license in templates to generate files with the proper header -->
               <include>test/java/spoon/generating/clone/*</include>
             </includes>
+            <excludes>
+                <!--  excluding the code coming from Javaparser -->
+                <exclude>main/java/spoon/javadoc/**</exclude>
```
pvojtechovsky: why the source in javadoc dir are excluded from checking of license header?

src/main/java/spoon/javadoc/internal/Javadoc.java
```diff
@@ -0,0 +1,280 @@
...
+
+	/**
+	* The structured content of a single Javadoc comment.
+	*
+	* <p>It is composed by a description and a list of block tags.
+	*
+	* <p>An example would be the text contained in this very Javadoc comment. At the moment of this
+	* writing this comment does not contain any block tags (such as <code>@see AnotherClass</code>)
+	*/
+	public class Javadoc {
```
pvojtechovsky: Why all the lines above have TAB after EOL? Why the last line of this file has TAB?

pvojtechovsky:
I see the problem with line separators. The new javadoc parsing code often uses `System.lineSeparator()`, but it is usually not fitting to the EOL of processes java source file.

 Note that Spoon CtComment is always using `\n` as line separator for the value returned by `CtComment.getContent()`. See javadoc of `CtComment.LINE_SEPARATOR`.
```java
	/**
	 * This line separator is used in comments returned by {@link #getContent()}.
	 * It is OS independent.
	 * It has no influence to pretty printed comments, which uses by default OS dependent line separator
	 */
	String LINE_SEPARATOR = "\n";
```
It is much simpler to analyze comments when you can be sure that EOL is always `\n`, independent on the EOLs used in source file. Note that java sources are sometime messy concerning different kinds of EOLs ... which can differ even in files of the same project.

So the idea is: When Spoon parses comments it detects all possible kinds of EOL and replaces them by `\n`. 
When Spoon prints comments back to file, then 
A) it uses EOL defined by Environment
B) it uses EOL which was used in origin source file (SniperPrinter)

So I guess that code of this PR will work well on Linux and will always fail on other OSes.

src/main/java/spoon/javadoc/internal/Javadoc.java
```diff
@@ -0,0 +1,280 @@
...
+	}
+
+	private static String trimRight(String string) {
+		while (!string.isEmpty() && Character.isWhitespace(string.charAt(string.length() - 1))) {
+			string = string.substring(0, string.length() - 1);
+		}
+		return string;
+	}
+
+	private static List<String> cleanLines(String content) {
```
pvojtechovsky: this method is not used. Should be deleted

src/main/java/spoon/javadoc/internal/Javadoc.java
```diff
@@ -0,0 +1,280 @@
...
+			cleanedLines = cleanedLines.subList(1, cleanedLines.size());
+		}
+		while (cleanedLines.size() > 0 && cleanedLines.get(cleanedLines.size() - 1).trim().isEmpty()) {
+			cleanedLines = cleanedLines.subList(0, cleanedLines.size() - 1);
+		}
+		return cleanedLines;
+	}
+
+	// Visible for testing
+	static int startsWithAsterisk(String line) {
```
pvojtechovsky: this function is not used. Should be deleted

pom.xml
pvojtechovsky: ok, I see that license headers are different ... from javaparser

monperrus: Yes, we want to keep the original Javaparser double-license scheme.

src/main/java/spoon/javadoc/internal/Javadoc.java
monperrus: I had to use an automated formatter to comply with our tough checkstyle. It did that, I don't know why. But not that it's checkstyle compliant :)

monperrus:
Addressed all your comments.

src/main/java/spoon/javadoc/internal/Javadoc.java
```diff
@@ -0,0 +1,222 @@
...
+
+	/** parses the Javadoc content (description + tags) */
+	public static Javadoc parse(String commentContent) {
+		List<String> cleanLines;
+
+		// In JavaParse, ,they had their own version of cleaning
+		// not backward compatible with Spoon's one
+		// so we keep Spoon's one
+		// cleanLines = cleanLines(commentContent);
+		cleanLines = Arrays.asList(commentContent.split(EOL));
```
pvojtechovsky: > I took care of not writing anything OS-dependent.

The input string actually always contains `\n`. But constant `EOL` on this line is OS dependent EOL == System.lineSeparator();

There is more uses of `System.lineSeparator()`, I suggest to replace them by "\n", so we keep already established concept that lines in Javadoc values in Spoon are always separated by `\n` ... as long as they are in form of String values of the spoon model. Once they are printed, the EOL vary depending on OS or Spoon configuration.

pvojtechovsky:
> How could you see this one hour ago? Mistery!

No mistery ... I am fighting with github comment approval process. I made this comment weeks ago. But I just found that I have to approve them before anybody else can see them...

src/main/java/spoon/reflect/code/CtComment.java
```diff
@@ -59,12 +60,21 @@
...
+	 * @see #getRawContent() for the raw comment.
+	 * @return the parsed content of the comment
 	 */
-	@PropertyGetter(role = COMMENT_CONTENT)
+	@DerivedProperty
 	String getContent();
 
+	/** The exact string that appears in source code, with delimiters */
+	@PropertyGetter(role = COMMENT_CONTENT)
+	String getRawContent();
```
pvojtechovsky: Do we really need to get access to origin raw content? I have bad feeling that it will cause performance problems ... spoon is already often calls `CtElement#equals`, which does a deep compare of whole AST and now each `CtCommentImpl#equals` calls `getContent()` twice ... and getContent always
* cleans raw comment
* parses cleaned comment
* makes a model of comment
* prints text of comment from model

Do we really need raw comment? Cannot we simply store comment in an cleaned and parsed form?

src/main/java/spoon/reflect/code/CtComment.java
```diff
@@ -59,12 +60,21 @@
...
+
 	@PropertySetter(role = COMMENT_CONTENT)
 	<E extends CtComment> E setContent(String content);
```
pvojtechovsky: If you keep `getRawComment`, then you should provide `setRawContent` too.

src/main/java/spoon/javadoc/internal/Javadoc.java
monperrus: Here you mix two things: the line separator of the input source (here), and the line separator in `getContent()`

`getContent()` uses `\n`, see `ret.append(CtComment.LINE_SEPARATOR)`

src/main/java/spoon/support/compiler/jdt/JDTCommentBuilder.java
```diff
@@ -583,7 +522,11 @@ static CtElement getBody(CtElement e) {
 	 * @return the content of the comment
 	 */
 	private String getCommentContent(int start, int end) {
-		return cleanComment(new CharArrayReader(contents, start, end - start));
+		try {
+			return IOUtils.toString(new CharArrayReader(contents, start, end - start));
```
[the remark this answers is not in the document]
monperrus: done, thanks!

src/main/java/spoon/support/compiler/jdt/JDTCommentBuilder.java
```diff
@@ -598,8 +541,9 @@ private static String cleanComment(Reader comment) {
 		StringBuilder ret = new StringBuilder();
 		try (BufferedReader br = new BufferedReader(comment)) {
 			String line = br.readLine();
-			if (line.length() < 2 || line.charAt(0) != '/') {
-				throw new SpoonException("Unexpected beginning of comment");
+			// nothing in the first line
+			if (line == null || line.length() == 0) {
```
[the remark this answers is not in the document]
monperrus: the exception has to be removed because we use cleanComments in more cases now, see https://github.com/INRIA/spoon/pull/2172#issuecomment-429629730

src/main/java/spoon/reflect/code/CtComment.java
pvojtechovsky: We might get raw content as derived property from the origin sources - using SourcePosition ... and to keep model nice and fast with cleaned comment. WDYT?

monperrus: > CtCommentImpl#equals calls getContent()

I am aware of this, I'm not against to undo this change. We  would loose a little bit of backward compatibility (this will require more test changes), but we will gain performance. Do you prefer to go this way?

src/main/java/spoon/reflect/code/CtComment.java
monperrus: added doc

monperrus: added documentation

src/main/java/spoon/javadoc/internal/Javadoc.java
pvojtechovsky: no. I speak about the line separator of the input source (here). This method is called here 
```java
javadoc = Javadoc.parse(cleanComment(content));
```
and `cleanComment` always returns `\n` as line separator, therefore input of this `parse` method always has `\n` as line separator ... but `EOL` constant defined above is actually `\r\n` on Windows so it cannot work.

src/main/java/spoon/support/compiler/jdt/JDTCommentBuilder.java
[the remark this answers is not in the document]
pvojtechovsky: ok, thanks

monperrus:
Working on this, based on @pvojtechovsky comments in order to make progress towards merge

* removing the call to `getConcent` in `#equals` for sake of performance.
* adding `setRawContent`

pvojtechovsky:
Spoon model is "semantic java model" not source code java model. Therefore I am not happy with storing rawComment in model which may cause performance problems. 
Cannot we store cleaned comment and to implement `getRawComment` by getting of String from origin sources using Comment position start/end?

monperrus:
> which may cause performance problems.
I disagree here, we have no measurement yet to claim anything or take a decision.

Are you talking about memory performance or runtime performance?

monperrus:
> I agree to continue like you suggest.

cool :)

> We can always change it.

yes, software is soft!

src/main/java/spoon/reflect/code/CtComment.java
```diff
@@ -36,19 +37,35 @@
...
+		public String getPrefix() {
+			return prefix;
+		}
+
+		public String getSuffix() {
+			return suffix;
+		}
+
+		private String prefix = "";
+		private String suffix = "";
```
pvojtechovsky: `private final String suffix;` WDYT?

src/main/java/spoon/reflect/visitor/CommentHelper.java
```diff
@@ -37,89 +36,56 @@
...
-				break;
-			case JAVADOC:
-				printer.write(DefaultJavaPrettyPrinter.BLOCK_COMMENT_END);
-				break;
+	}
+
+	/** Utility method to print a complete comment */
+	public static void printComment(PrinterHelper printer, CtComment comment) {
+		CtComment.CommentType commentType = comment.getCommentType();
+		// we prefer the raw version, if it's in sync with the comment type
```
pvojtechovsky: > we prefer the raw version

why?

pvojtechovsky:
Printing of raw comment has these bad side effects:
* the indentation of the comments will not fit with indentation of context
* the EOL in comment body may be different to other EOLs in the file
* the setComment("first line\nsecond line"), will be printed wrong now

Looking at all necessary changes I would say that storing of raw content is not good.

> add method "getRawComment()" for clients who want to do their own parsing (this was impossible before)

@monperrus Is it the only reason why we store raw content and print wrong aligned comments with inconsistent EOLs now? It is quite high price for that :-(

pvojtechovsky:
See alternative implementation of `getRawContent` in #2746.
WDYT?

monperrus:
>   * the indentation of the comments will not fit with indentation of context
>
Yes, I agree with you, we want to keep this behavior.
>
>   * the EOL in comment body may be different to other EOLs in the file
>
Correct, this is not good.
>
>   * the setComment("first line\nsecond line"), will be printed wrong now
>
What do you mean?
>
> Looking at all necessary changes I would say that storing of raw content is not good.
>
I disagree. The problem is not about storing it, it's about using it in the way I've explored
yesterday , which is not satisfactory, I agree.

> See alternative implementation of |getRawContent| in #2746 <https://github.com/INRIA/spoon/pull/2746>.
It does not comply with requirement R2 in https://github.com/INRIA/spoon/pull/2172#issue-199407714.
But the more I think about it, the more I prefer this alternative implementation you propose. It
would still enable client code to do what the they want in most cases.

WDYT?

pvojtechovsky:
> R2: the comments (incl. the original one), can be analyzed from the model and the serialized model, even if the Java files don't exist anymore on disk)

I did not noticed that requirement. What is the use case behind that: "the Java files don't exist anymore on disk"? By the way Sniper printer will not work in such case too.

>>  the setComment("first line\nsecond line"), will be printed wrong now

> What do you mean?

if printing would use original tabs and EOLs then it would print
```java
/* first line
second line */
   someElementWithIndent;
```
instead of current
```java
   /*
    * first line
    * second line
    */
   someElementWithIndent;
```

> But the more I think about it, the more I prefer this alternative implementation you propose. It
would still enable client code to do what the they want in most cases.

I liked the origin cleaned/semantic comments. They are more fitting to concept of semantic Java AST, then rawComments. So I would prefer that simple alternative solution too.

monperrus:
> I liked the origin cleaned/semantic comments
Now that we're introducing an object tree for comment comments, we have to basically give away the
String field, and only maintain an object field. (note that we already had a proto-tree with the
tags before ... and actually some hidden synchronization problems already)

The pretty-printed string version of the comment has to be computed on the fly (and possibly cached
... to the price of maintaining a cache infrastructure).

WDYT?

pvojtechovsky:
I do not know all side effects yet, but I agree that we should get rid of String representation of comment and switch to the `object tree for comment content`... but that is longer way ... because we need comment AST nodes with same capabilities like CtElement node has. I need mainly access to SourcePosition to be able build source fragments and handle sniper pretty printing correctly. Also the change listeners are needed ... but all that can be done later. For now I like current PR, but without that rawContent stuff. Just parsing of comments as internal invisible stuff.

pvojtechovsky:
Please maximize compatibility - most of old tests should pass without all these changes.

monperrus:
OK, we're along the same line. I agree, we can add support for position and change listeners later.

monperrus:
And you know me, I take backward compatibility super seriously.

monperrus:
the diff is clean and CI is OK ... but the performance cost is huge: (11 minutes versus 7 minutes per job) https://travis-ci.org/INRIA/spoon/builds/462260778

What are the solutions?

pvojtechovsky:
If I understood it well, then `updatedInlineLinksInJavadoc` method needs that time.

There is also another reason why we should do it different - new concept of handling of imports, which I discussed with Simon (#2683). One of the ideas of this concept is that model is normally not modified to show/hide fully qualified names. Only just before printing, there are started some model procesors/validators, which changes model to apply imports or FQN mode.

So may be `updatedInlineLinksInJavadoc` should be called from such model validator before printing only. While actually it is called all the time even during building of the model.

So I suggest to wait with this PR after #2683 is merged.

monperrus:
Yes the super slow call is `this.getFactory().getModel().filterChildren(new NamedElementFilter<>(CtType.class, stype)).first(CtType.class)`

OK for me, let's wait. I WIP this one again.

monperrus:
#3160 is a better alternative, way faster (no call to `Type().get()`).

PR #2177, opened 2018-07-05
refactor: move resources to testclasses in generics

zielint0:
@surli 
I am ready to merge.


zielint0:
Why this spoon-bot said that?

surli:
> Why this spoon-bot said that?

Short version:
There was a kind of race condition when checking the API.

Long version:
Here's the process when pushing on master (or when merging a PR):
a Jenkins job will get the last changes from master and deploy automatically a new snapshot to the repository.

Here's the process for checking the API:
it reproduce the merge commit between the PR and the master branch, then it check the API against the last downloaded snapshot. 

So let's say that your PR was created just when another PR is merged: then the check of your PR will use the new API, but the new snapshot is not yet available. So it shows some warning which in fact are about the previously merged PR.

zielint0:
@surli 
Thank you very much for this detailed explanation! Mercii! :-)

PR #2179, opened 2018-07-05
refactor: rename test file

zielint0:
@surli 
Should "TestInterfaceWithoutSetup" be renamed to "InterfaceWithoutSetupTest"?


surli:
Actually it does change anything for us.

surli:
> But for consistency we should have all names with the same convention.

The current convention is to have `Test` in the fully qualified name of the test:
> // contract: all test classes, as well as template/processors/etc used in tests have "test" in their fully qualified name

This is expressed in `MainTest#testTest` and for example we have `TestCompilationUnit`, `TestCtBlock` etc.

zielint0:
@surli 
Thank you very much for this explanation and examples.
Could I unify test naming to CtBlockTest, CompilationUnitTest etc. as well?
Perhaps I am able to improve consistency a bit.
Do you see any obstacles?

PR #2190, opened 2018-07-08
refactor: move resources to testclasses in strings

zielint0:
I need some time to think about it.
Could you guide me, please?


monperrus:
why is `sourceJar/test.jar` a changed file?

zielint0:
@monperrus 
Because it contains a Java file that was modified (I added .testclasses suffix).
I had to modify Java code (change package) to pass tests.

src/test/java/spoon/test/jar/JarTest.java
```diff
@@ -27,7 +27,7 @@ public void testJar() throws Exception {
 				SpoonResourceHelper.resources("./src/test/resources/sourceJar/test.jar"));
 		Assert.assertTrue(compiler.build());
 		assertEquals(1, factory.getModel().getAllTypes().size());
-		assertEquals("spoon.test.strings.Main", factory.getModel().getAllTypes().iterator().next().getQualifiedName());
+		assertEquals("spoon.test.strings.testclasses.Main", factory.getModel().getAllTypes().iterator().next().getQualifiedName());
```
surli: I don't agree with changing this equals and the related jar file: the purpose of moving the resource file is to ease the maintaining by avoid polluting test classes with resources at the same place. But here the file is already in a jar file, so it does not pollute anything and changing the package name does not change anything to the test.

zielint0:
@surli 
@monperrus 

I am not sure how to explain but will try:
1. create testclasses
2. move Main
3. so I must change package name (add testclasses suffix)
4. fix java code in jar to pass test in JarTest

The result is:
- Main moved to testclasses
- tests pass

What I am doing wrong?
What is incorrect in this try?
Can you move Main using correct way?

PR #2191, opened 2018-07-08
refactor: move resources to testclasses in secondaryclasses

zielint0:


monperrus:
per the naming convention, it seems that we should rather rename secondaryclasses to testclasses.

zielint0:
There is already package *testclasses* (with: SampleClass, SampleImportClass).
I cannot rename *secondaryclasses* to *testclasses* because of that.
Should I move content of *secondaryclasses* to *testclasses*?

surli:
> per the naming convention, it seems that we should rather rename secondaryclasses to testclasses.

I don't agree actually: it follows the usual convention.
`test root package / test sub package / testclasses`

> Should I move content of secondaryclasses to testclasses?

I'd say no.

zielint0:
@surli
So can you merge this pr as is?

PR #2193, opened 2018-07-08
refactor: fix checkstyle violations in tests api

zielint0:
Step by step, oh, checkstyle, you're always on my mind
Step by step, oh, checkstyle, I really think it's just a matter of time

- MetamodelTest.java
- APITest.java

I am ready to merge. With style!


pvojtechovsky:
in Eclipse there is a checkstyle plugin, which will mark all checkstyle problems immediately in editor, so I usually fix them before commit. Try it ;-)

And in the root of spoon project there is an xml which can be used to configure it

zielint0:
@pvojtechovsky 
Zdravim!
Thank you for this suggestion, however I use IntelliJ IDEA (even the free "community" version is smarter than Eclipse). I tried to use checkstyle plugin for IntelliJ IDEA, but I am experiencing problems (I suspect this is because of I am using beta version of IDE, it's not compatible with stable checkstyle plugin). I wholeheartedly recommend IDEA because of its editor, code inspection and *many more*.

Having said that, I must confess that I use cli very often. Despite IDE's git integration, I use git in command line. I also use checkstyle in command line:
```
java -jar .\checkstyle-8.10.1-all.jar -c .\checkstyle.xml src > .\result.txt
```

Once again, thanks for tip! I very appreciate your feedback!
Děkuji vám moc! :-)

PR #2206, opened 2018-07-11
refactor: occurences -> occurrences

zielint0:
Fixed naming in names.

### DIFF
--- a/src/main/java/spoon/pattern/PatternParameterConfigurator.java
+++ b/src/main/java/spoon/pattern/PatternParameterConfigurator.java
@@ -132,14 +132,14 @@ public class PatternParameterConfigurator {
 	}
 
 	public PatternParameterConfigurator setMinOccurence(int minOccurence) {
-		currentParameter.setMinOccurences(minOccurence);
+		currentParameter.setMinOccurrences(minOccurence);
 		return this;
 	}
 	public PatternParameterConfigurator setMaxOccurence(int maxOccurence) {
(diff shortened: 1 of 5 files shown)


surli:
The refactoring you propose here might break a lot of API and does not really bring a value for the developers, so I won't merge it.

monperrus:
this is a brand new, experimental API, so I would be OK to fix it right now. But I agree, in six months, it will be too late.

@pvojtechovsky ?

pvojtechovsky:
> this is a brand new, experimental API, so I would be OK to fix it right now.

I agree with all of you Simon, Martin, Tomasz ;-)

I suggest to merge it now with the request to make new minor Spoon release 7.1 soon, then there will be not many clients influenced by this change. WDYT?

zielint0:
@pvojtechovsky 
> ... it is interesting how much work might be to fix one missing r ... :-(

Och no, it was quick (Shift+F6).
than vs then was much longer!

zielint0:
By mistake I deleted branch.
I re-opened PR.
Sorry for confusion.

surli:
could you create an empty commit to trigger back the CI please? And same for the other PRs

zielint0:
@surli 

> could you create an empty commit to trigger back the CI please? And same for the other PRs

All Travis green for:
- https://github.com/INRIA/spoon/pull/2212
- https://github.com/INRIA/spoon/pull/2213
- https://github.com/INRIA/spoon/pull/2214

PR #2213, opened 2018-07-13
Refactor fix naming in ct named path element.java (re-introduced)

zielint0:


zielint0:
@monperrus 
> Thanks a lot!

My pleasure :-)

PR #2216, opened 2018-07-13
Fix #2211: Clone metadata

### LINKED ISSUE #2211: Clone the metadata of nodes
Hello all,

I would know if you agree to clone also the metadata when we clone a CtElement.

The use case is the following:

I find a specific element inside a method, _e.g._ a literal. According to its value, I generate multiple variants of the same method with different values for this literal.

To do this, I apply the following algorithm:
1. replace the value of the original literal with one of the generated
2. clone the CtMethod
3. restore the original value in the original method.

And I obtain a set of variants of the original method

This allows me to not execute the same query on the variants, _i.e_ the query to find specifc nodes.

However, I would like to add metadata on the literal, with the new values(_i.e._ before the step 2) but currently, it cannot be done using the `putMetadata()` method, since the metadata will be lost when cloning the method (at the step 2).

WDYT?

Best!

danglotb:


src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
```diff
@@ -469,6 +469,17 @@ public void replace(CtElement element) {
...
 	}
 
+	@Override
+	public <E extends CtElement> E setMetadata(Map<String, Object> metadata) {
+		if (metadata != null) {
+			if (this.metadata == null) {
+				this.metadata = new HashMap<>();
+			}
+			this.metadata.putAll(metadata);
+		}
```
pvojtechovsky: `else this.metadata = null`

src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
```diff
@@ -486,6 +497,11 @@ public Object getMetadata(String key) {
 		return metadata.get(key);
 	}
 
+	@Override
+	public Map<String, Object> getMetadata() {
+		return this.metadata;
```
pvojtechovsky: `return Collections.unmodifiableMap(this.metadata)`
WDYT?

danglotb:
Thank you @pvojtechovsky for the review. 

I excluded `CtElement#setMetadata(Map<String, Object>)` from the test `AstCheckerTest#testPushToStackChanges()`.

WDYT?

src/test/java/spoon/generating/CloneVisitorGenerator.java
```diff
@@ -212,7 +212,7 @@ private CtInvocation createFactoryInvocation(CtVariableAccess<CtElement> element
 					"spoon.support.reflect.declaration.CtModifiableImpl", "spoon.support.reflect.declaration.CtMultiTypedElementImpl", //
 					"spoon.support.reflect.declaration.CtTypeMemberImpl", "spoon.support.reflect.code.CtRHSReceiverImpl",
 					"spoon.support.reflect.declaration.CtShadowableImpl", "spoon.support.reflect.code.CtBodyHolderImpl", "spoon.support.reflect.declaration.CtModuleDirectiveImpl");
-			private final List<String> excludesFields = Arrays.asList("factory", "elementValues", "target", "metadata");
+			private final List<String> excludesFields = Arrays.asList("factory", "elementValues", "target");//, "metadata");
```
pvojtechovsky: why that comment? It makes no sense for future. I would prefer to have the sources clean. WDYT?

pvojtechovsky:
It looks good to me.
@monperrus, do you agree to add `CtElement#get/setMetadata(Map)` method?

monperrus:
I propose to call them setAllMetadata/getAllMetadata

Otherwise, it's OK for me.

src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
```diff
@@ -469,6 +469,19 @@ public void replace(CtElement element) {
 		ReplacementVisitor.replace(this, elements);
 	}
 
+	@Override
+	public <E extends CtElement> E setMetadata(Map<String, Object> metadata) {
+		if (metadata != null) {
+			if (this.metadata == null) {
+				this.metadata = new HashMap<>();
+			}
+			this.metadata.putAll(metadata);
```
surli: You should have cleared the map first, else it's not a setter.

src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
```diff
@@ -469,6 +469,19 @@ public void replace(CtElement element) {
 		ReplacementVisitor.replace(this, elements);
 	}
 
+	@Override
+	public <E extends CtElement> E setMetadata(Map<String, Object> metadata) {
```
surli: question for @monperrus and @pvojtechovsky : shouldn't we call the model change listener here? Do we consider that metadata are not part of the model and the change on them shouldn't be notified?

pvojtechovsky: I would say change listener doesn't have to call change listener. But if one will needed. It is no problem to implement it later.

src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
```diff
@@ -486,6 +501,14 @@ public Object getMetadata(String key) {
 		return metadata.get(key);
 	}
 
+	@Override
+	public Map<String, Object> getAllMetadata() {
+		if (this.metadata == null) {
+			return null;
```
pvojtechovsky: what about `return Collections.emptyMap()`?

src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
```diff
@@ -469,6 +469,21 @@ public void replace(CtElement element) {
 		ReplacementVisitor.replace(this, elements);
 	}
 
+	@Override
+	public <E extends CtElement> E setAllMetadata(Map<String, Object> metadata) {
+		if (metadata != null) {
```
pvojtechovsky: `if (metadata != null && metadata.size() > 0) ...`
WDYT?

src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
surli: ok then I just wanted to be sure it was not something missing.

src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
surli: or this other way:
```
if (metadata == null) {
  // empty map
  return this;
}
this.metadata.clear();
this.metadata.putAll(metadata);
```

src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
```diff
@@ -469,6 +469,21 @@ public void replace(CtElement element) {
 		ReplacementVisitor.replace(this, elements);
 	}
 
+	@Override
+	public <E extends CtElement> E setAllMetadata(Map<String, Object> metadata) {
+		if (metadata == null || metadata.isEmpty()) {
+			this.metadata = new HashMap<>();
```
pvojtechovsky: empty hashmap has some size in memory and if we would have such empty hash maps more often (for each CtElement) then Spoon model would be n-times bigger - for nothing. 

I would prefer to use null to represent "no metadata attribute". WDYT?

src/main/java/spoon/reflect/declaration/CtElement.java
```diff
@@ -305,6 +311,11 @@
 	 */
 	Object getMetadata(String key);
 
+	/**
+	 * Retrieves all metadata stored in an element. Returns null if it does not exist.
```
pvojtechovsky: Now the comment needs update. It doesn't return null

pvojtechovsky:
@surli, I am sorry, but I by mistake pressed ENTER during editing of commit message comment ... so I committed wrong message ... :-(

surli:
No problem @pvojtechovsky ;)

PR #2221, opened 2018-07-14
refactor: fix code inspection - redundant throws clause

zielint0:
If the method does NOT throw a specific checked exception, the exception should NOT be declared in throws clause.

To make code compilable, I removed checking for non-existing exceptions in AbstractProcessor.java

I am ready to merge.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,6 @@ env:
   - SCRIPT=travis-coverage.sh
   - SCRIPT=travis-verify.sh
   - SCRIPT=travis-maven339-jdk8.sh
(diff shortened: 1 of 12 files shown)


monperrus:
I disagree with your statement `If the method does NOT throw a specific checked exception, the exception should NOT be declared in throws clause.`, because a throws clause may be here to warn clients that the method may be thrown **in the future**.

But I agree with your PR because the whole exception-handling design of Spoon is to only use unchecked exception, in particular SpoonException and its subclasses (see discussion in #1947).

Thanks!

PR #2222, opened 2018-07-15
Refactor move resources to testclasses in filters

zielint0:


zielint0:
@monperrus 

> According to the baby PR principle, it should be two different PRs

So I created refactor: fix code inspection manual array to collection copy #2224 

> Some assertions have changed. It may signal a behavioral change.

I moved a Foo.java to testclasses, so changing assertions was unavoidable (without it, tests were not passing).

PR #2224, opened 2018-07-15
refactor: fix code inspection manual array to collection copy

zielint0:


src/main/java/spoon/reflect/visitor/chain/CtQueryImpl.java
```diff
@@ -80,9 +80,7 @@ public CtQueryImpl addInput(Object... input) {
 			this.inputs = new ArrayList<>();
 		}
 		if (input != null) {
-			for (Object in : input) {
-				this.inputs.add(in);
-			}
+			Collections.addAll(this.inputs, input);
```
surli: Why using the static call instead of using `this.inputs.addAll(input)`?

zielint0: https://m.youtube.com/watch?v=15X0qFtBqiQ

zielint0: And I think that static call shows what is added to what in a clearer way.

surli: I agree with using `addAll` instead of the loop: my question was only about why not using the invocation from the instance directly.
IMHO it's more readable than calling `Collections.addAll` with two arguments.

zielint0: I tried to refactor from **Collections.addAll()** to **this.input.addAll(input)**, but second variant cannot be compiled because of argument type which is: addInput(Object... input)

**Collections.addAll()** is:
```addAll(Collection<? super T> c, T... elements)```

**this.input.addAll()** is:
```addAll(Collection<? extends E> c)```

The exact mismatch in **this.input.addAll(input)** is:
```
addAll in List (java.util.Collection<?>) cannot be applied to (java.lang.Object[])
```

In other words: **Object... input** is an array, so it can be used only in Collection<? super T> c, **T... elements**)

So the only way in this context is static method Collections.addAll().

surli: OK for this one then. But for the other ones you could have use the form instance.addAll() no?

zielint0: I used explicit and simple way in all cases: x.addAll(Arrays.asList(y)) :-)
No need for Collections.addAll()
I am ready to merge.

surli:
My mistake: I though that except the first change, the other ones were about adding list in list, not array in list: that was my original question. So in this case, yeah I agree its better using `Collections.addAll` sorry for the bothering @zielint0

pvojtechovsky:
Thank you Tomasz

zielint0:
@pvojtechovsky 
>Thank you Tomasz

Děkuji za dobré slovo!

PR #2233, opened 2018-07-16
refactor: remove unnecessary 'return' statement

zielint0:
Verbose or redundant code constructs
- Unnecessary return statement

Rationale
Simpler, more readable code. Less clutter.


surli:
> more readable code

I would not say that: it took me some time to check in `PartialEvaluator` that removing the `return` was correct because there was not another statement after it. On long method it sometimes make sense to improve the readibility. Nevertheless I merge it.

PR #2239, opened 2018-07-17
WiP fix: wrong identifiers according to JLS are not according anymore in setSimpleName

### LINKED ISSUE #2237: bug: should reject invalid names according to the Java specification
Hi all.

I am opening this thread to discuss the possibility of checking the name against the specification names.

Currently in Spoon, I can do the following:

```java
CtVariableReference reference = factory.createLocalVariableReference().setSimpleName("tacos.EatIt()")
factory.createVariableRead(reference, false);
```

This can lead to an inconsistency of the spoon model because the invocation is represented by a local variable.

What can be done, is a check on the provided name, using a regex.

WDYT?

EDIT: I fixed the snippet since it was incorrect. However, the discussion remains the same and open.

surli:
fix #2237


surli:
I see at least two cases that are difficult to handle here:
  - when calling an array type reference such as `String[]`, `setSimpleName` is called with the string `String[]`: IMHO it's wrong, that's why I opened #2240 
  - when invoking something from an anonymous class, the name of the type reference is just a number (e.g. `SampleClass$1` the class simple name is `1`): it's normally not authorized as an identifier in Java. I wonder if we should renamed them using a dedicated convention. WDYT?

src/main/java/spoon/reflect/factory/Factory.java
```diff
@@ -130,6 +131,67 @@
  * Otherwise FactoryImpl is a default implementation.
  */
 public interface Factory {
+	String[] keywords = {
```
pvojtechovsky: Why do you add that internal code into Factory interface? I suggest to move it to an utility class in a ...internal... package.
WDYT?

src/main/java/spoon/reflect/factory/Factory.java
```diff
@@ -130,6 +131,67 @@
...
+			"do", "goto", "private", "this",
+			"break", "implements", "protected", "throw",
+			"else", "import", "public", "throws",
+			"case", "enum", "instanceof", "return", "transient",
+			"catch", "extends", "try",
+			"final", "interface", "static",
+			"class", "finally", "strictfp", "volatile",
+			"const", "native", "super", "while", "_",
+			"true", "false", "null"
+			// primitive types are removed as they can be used in type reference
```
pvojtechovsky: I needed in sniper mode algorithms the list of java keywords too. But I needed complete list. So if this list is not complete then I suggest to make it even private - because it makes sense only for this one algorithm.
WDYT?

src/main/java/spoon/reflect/factory/Factory.java
surli: Yes no problem about moving it: I was wondering about putting it in `support.util` but then I'd have to create a new class for it. Unless you have a better idea?

src/main/java/spoon/reflect/factory/Factory.java
surli: Another way to do it is to keep a complete list of keywords, and to add the primitive types as the exceptions below. Maybe it would bring also a better semantic. WDYT?

src/main/java/spoon/support/reflect/declaration/CtNamedElementImpl.java
```diff
@@ -43,6 +43,10 @@ public String getSimpleName() {
 
 	@Override
 	public <T extends CtNamedElement> T setSimpleName(String simpleName) {
+		if (simpleName == null) {
```
pvojtechovsky: just a question.... is there a valid use case which calls `setSimpleName(null)`?

surli: I don't think so: I think it might be part of another PR to specify the behaviour of Spoon when calling `setSimpleName(null)`.

pvojtechovsky:
> when invoking something from an anonymous class, the name of the type reference is just a number (e.g. SampleClass$1 the class simple name is 1): it's normally not authorized as an identifier in Java. I wonder if we should renamed them using a dedicated convention. WDYT?

I am not using annonymous classes as client of spoon, but I would say too, that we should rename them as Simon suggest.

src/main/java/spoon/reflect/factory/Factory.java
pvojtechovsky: I am personally OK with new utility class... but may be there is someone who has better idea... utility classes are not the best practice.

src/main/java/spoon/reflect/factory/Factory.java
pvojtechovsky: yes, sounds good. Note: sniper mode needs also list of operators, separators, ... and some kind of java parser utilities ... I would say all of them (including keywords) should stay internal in spoon. It is probably not target of spoon to publish this java parser information in API

src/main/java/spoon/support/util/IdentifierHelper.java
```diff
@@ -0,0 +1,149 @@
...
+                identifierExceptions = otherReferenceIdentifierExceptions;
+                break;
+
+            default:
+                identifierExceptions = new String[0];
+                break;
+        }
+
+        // in case of exception: it's accepted directly
+        for (String identifierException : identifierExceptions) {
```
pvojtechovsky: I suggest to create several static `HashSet<String>` - one for each identifier type - and then simply call `Set.contains(str)`. I would say it will be faster, then current algorithm.
WDYT?

surli: I'll change that.

monperrus:
@surli any chance to finish this one?

surli:
I can't say when but yeah I want to finish it.

monperrus:
lots of conflicts, close?

surli:
> lots of conflicts, close?

this still looks like an important one to me, but definitely I cannot say I worked on it lately, and not sure to take it back soon. Close it if you want, I'll reopen if needed.

monperrus:
#2237 stays the reference for this.

PR #2240, opened 2018-07-17
review refactor: Change the way simple name is used in ArrayTypeReference

surli:
In `CtArrayTypeReference#getSimpleName` it returns a name based on its component type: IMHO we should mark the method as a derived property. Moreover, the setter is then useless and potentially harmful for the user. So I added the annotation and an empty overridden method in the implementation.


surli:
@pvojtechovsky there is one remaining failing test in `TemplateTest` maybe you have an idea how to fix it?

pvojtechovsky:
I will have a look

pvojtechovsky:
I am finished with the Template fix. It is ready for merge.

I agree with Simon's changes too. 

@monperrus please review and merge this one.

danglotb:
I merged this one since @monperrus is not available. Thank you!

PR #2242, opened 2018-07-17
WIP refactor: code inspection - Control flow issues: Pointless boolean ex…

zielint0:
…pression

Code inspection
Control flow issues
- Pointless boolean expression

Fixes also SonarQube - Boolean literals should not be redundant

I am ready to merge.


pvojtechovsky:
I personally prefer `if (x == false)` over `if(!x)`, because with longer expressions it is more readable for me. And I am sure, I am not the only one. 
@surli , @monperrus, WDYT?

surli:
Personally I never use the form `== false`: I understand the readibility argument, but here it's not a problem for me.

zielint0:
[Make a big deal out of == true?](https://softwareengineering.stackexchange.com/questions/12807/make-a-big-deal-out-of-true)

> When I see someBool == true, I can't help but feel like the programmer hasn't internalized the idea of evaluation, which is a pretty fundamental deficiency.
However, my perspective is skewed because I spent several summers in college teaching programming to kids, who frequently wrote expressions like this because they genuinely hadn't mastered the mental exercise of evaluating an expression. Once they grokked the concept, the redundancy became obvious.
For an otherwise competent professional programmer, this is probably not the case. It's probably just a bad habit they developed in their early days of programming and never quite shook. But it would still scare me a bit if it was the first thing I saw someone do in an interview.

[Why Use !boolean_variable Over boolean_variable == false](https://softwareengineering.stackexchange.com/questions/136908/why-use-boolean-variable-over-boolean-variable-false)

> When I see a line like  if (!lateForMeeting()), I read that as "If not late for meeting", which is quite straight-forward to understand, as opposed to if (lateForMeeting() == false) which I'd read as "If the fact that I'm late for meeting is false".
They're identical in meaning, but the former is closer to how the equivalent English sentence would be constructed.

> Writing == false and == true is redundant. It can be taken to arbitrary extremes, too. If you start writing
```if (condition == false) { ... }```
Then why not
```if ((condition == false) == true) { ... }```
Or why not
```if ((someExp == anotherExp) == true) { ... }```
The moral of this story is that if condition is a boolean expression, then you don't need to add == false; that's what operator ! is for ;)

---

So many analyzers detect this code smell
- http://checkstyle.sourceforge.net/config_coding.html#SimplifyBooleanExpression
- SonarQube (Boolean literals should not be redundant) - enables on Spoon's SonarQube

**I think it should be enforced by Checkstyle.**

**Comment by pvojtechovsky** [comment]
I do not want to fight for my opinion in this case. I can live with `!` - no problem. We just have in our company project the rule that `x == false` should be used, because it cann't be overseen like little `!`. 
But as I mentioned - no need to fight for that. I have no problem with `!`. Owner of Spoon project is the one who can decide ;-)

Tomasz, I welcome your cleaning effort a lot 👍

**Comment by zielint0** [comment]
@pvojtechovsky 
@surli 
@monperrus 
I fixed this because it is [detected](https://sonarqube.ow2.org/project/issues?id=fr.inria.gforge.spoon%3Aspoon-core&resolved=false&rules=squid%3AS1125) by our (official) SonarQube instance. I assume that this is not a random set of rules, isn't it? BTW I think that some checkers should be added.

If it is decided NOT to merge this PR, I would like to ask Simon (or other responsible person) to remove this checker from our SonarQube instance.

BTW. If we compare two versions:
```
if (!isLevelHigh()) {
...
}

```
```
if (isLevelHigh() == false) {
...
}
```
Isn't the first version more readable, because we can understand it as: Is Level High? No!

@pvojtechovsky 
> Tomasz, I welcome your cleaning effort a lot :+1: 

Thank you Pavel so much. I also appreciate your open mind, support and friendliness :-)

**Comment by surli** [comment]
On my side I understand the usage of `== false` for readibility reason for condition like:

```java
String.class.getName().equals(valueType.getActualTypeArguments().get(0).getQualifiedName()) === false
```

So for me, the solution is not to remove the `== false` there, but to break the condition in something more readable, such as:

```java
String firstArgumentFQN = valueType.getActualTypeArguments().get(0).getQualifiedName());
boolean isFirstArgumentFQNString = String.class.getName().equals(firstArgumentFQN);
if (!isFirstArgumentFQNString)
```

So I would consider this PR as not finished, as the condition are still hard to understand. WDYT?

zielint0:
@surli 
I will check if I can do something like you suggest.

monperrus:
No activity for one month, closing it for reducing the WIP load. Don't hesitate to reopen. Thanks for your proposal!

PR #2244, opened 2018-07-18
review: fix: comment on nested else if statement

pvojtechovsky:
This PR contains a text which reproduces this problem:
```java
    	//c1
        if (value == null) {
        	//c2 belongs to toto
            value = "toto";
        } else if (value.equals("x")) {
        	//c3 belongs to getClass
        	this.getClass();
        }
```
The comment `//c3` is incorectly placed to root if statements

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/JDTCommentBuilder.java
+++ b/src/main/java/spoon/support/compiler/jdt/JDTCommentBuilder.java
@@ -516,7 +516,7 @@ class JDTCommentBuilder {
 				if (element == null) {
 					return;
 				}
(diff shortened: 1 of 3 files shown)


surli:
@pvojtechovsky tell me if this fix seems correct for you

pvojtechovsky:
@surli, the solution and change in the test is OK. Thank You for that fix!
I am ok with merge of this PR

pvojtechovsky:
@monperrus, @tdurieux could you please review and merge this one? Simon and me are the authors so we cannot do that ;-)

PR #2245, opened 2018-07-18
refactor: Anonymous class now have an explicit name

surli:


pvojtechovsky:
> it's a bad idea.

@surli, could you explain it? Please share your experience ;-)

surli:
The problem is that if the simple name is use everywhere in Spoon to output classes: then if I put an anonymous class in `Foo`, its name for compilation must be `Foo$1`. If I rename it in the model for something such as: `Foo$<anonymous>1` I would have to remove the anonymous everywhere for the output.

Moreover, it's not something which would be expected by the client: if she's looking for the second anonymous class, for example, she might create a name filter to `Foo$2` and would be surprised to not find anything. 

So all in all I now think it's a bad idea to change their name.

PR #2246, opened 2018-07-18
review: refactor: print statement's semicolon and after comment in scope of scan

pvojtechovsky:
The sniper printing mode algorithm expects that all source code of an element is printed in scope of it's `DJPP#scan` method.
But it was not the case of CtStatement, which printed suffix semicolon and after comment later... after living the scan method.
I hope this kind of DJPP refactoring is acceptable.

Advantage: the CtStatement#toString should now print after comment too - as expected.

### DIFF
--- a/src/main/java/spoon/reflect/visitor/PrintingContext.java
+++ b/src/main/java/spoon/reflect/visitor/PrintingContext.java
@@ -44,6 +44,9 @@ public class PrintingContext {
 	public boolean isFirstForVariable() {
 		return (state & FIRST_FOR_VARIABLE) != 0L;
 	}
+	/**
+	 * {@link #isNextForVariable()}
+	 */
 	@Deprecated
 	public boolean noTypeDecl() {
 		return isNextForVariable();
--- a/src/main/java/spoon/reflect/visitor/PrintingContext.java
+++ b/src/main/java/spoon/reflect/visitor/PrintingContext.java
@@ -17,6 +17,7 @@
 package spoon.reflect.visitor;
 
 import spoon.reflect.code.CtExpression;
+import spoon.reflect.code.CtStatement;
 import spoon.reflect.declaration.CtElement;
 import spoon.reflect.declaration.CtType;
 import spoon.reflect.reference.CtTypeReference;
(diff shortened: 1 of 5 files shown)


src/main/java/spoon/reflect/visitor/PrintingContext.java
```diff
@@ -26,17 +27,22 @@
...
 	private long SKIP_ARRAY 			= 1 << 2;
 	private long IGNORE_STATIC_ACCESS   = 1 << 3;
 	private long IGNORE_ENCLOSING_CLASS = 1 << 4;
 	private long FORCE_WILDCARD_GENERICS = 1 << 5;
+	private long FIRST_FOR_VARIABLE 	= 1 << 6;
 
 	private long state;
+	private CtStatement statement;
 
-	public boolean noTypeDecl() {
```
surli: instead of removing it, couldn't you just deprecate it with checking the flag: `NEXT_FOR_VARIABLE`? I imagine the semantic of the flag did not really change did it?

src/main/java/spoon/reflect/visitor/PrintingContext.java
```diff
@@ -53,20 +59,30 @@ public boolean ignoreEnclosingClass() {
...
 			oldState = state;
+			oldStatement = statement;
 		}
 		@Override
 		public void close() {
 			state = oldState;
+			statement = oldStatement;
 		}
 
-		public <T extends Writable> T noTypeDecl(boolean v) {
```
surli: Same question here

src/main/java/spoon/reflect/visitor/PrintingContext.java
pvojtechovsky: semantic of the flag is same. It just have better name. OK. So I will deprecate old name and create new name with same internal property

pvojtechovsky:
the required changes were applied. It is ready for review and merge from my point of view.

src/main/java/spoon/reflect/visitor/PrintingContext.java
```diff
@@ -26,17 +27,32 @@
...
 	private long IGNORE_ENCLOSING_CLASS = 1 << 4;
 	private long FORCE_WILDCARD_GENERICS = 1 << 5;
+	private long FIRST_FOR_VARIABLE 	= 1 << 6;
 
 	private long state;
+	private CtStatement statement;
 
+	/**
+	 * @return true if we are printing first variable declaration of CtFor statement
+	 */
```
surli: Could you also add some documentation for the deprecated?

surli:
Apparently we got a regression on windows machine after this one, see: https://ci.inria.fr/sos/job/spoon-windows/75/

PR #2248, opened 2018-07-18
fix: sonarqube remove commented code

zielint0:
SonarQube rationale
> Programmers should not comment out code as it bloats programs and reduces readability.
Unused code should be deleted and can be retrieved from source control history if required.
See
MISRA C:2004, 2.4 - Sections of code should not be "commented out".
MISRA C++:2008, 2-7-2 - Sections of code shall not be "commented out" using C-style comments.
MISRA C++:2008, 2-7-3 - Sections of code should not be "commented out" using C++ comments.
MISRA C:2012, Dir. 4.4 - Sections of code should not be "commented out"

Commented code is always availabe in repository.
There is no need to keep it (*waiting for better times*).


pvojtechovsky:
This PR is OK for me. Just with exception of one comment mentioned above

PR #2249, opened 2018-07-18
fix typo: Occurence->Occurrence (important for consistency, see previous changes!)

zielint0:
@monperrus 

This PR is follow up to refactor: fix naming occurences (re-introduced) #2214
As you merged #2214, this one should be too.
This PR finishes partial fix from #2214

Previous discussion: https://github.com/INRIA/spoon/pull/2206
@pvojtechovsky 
> ... it is interesting how much work might be to fix one missing r ... :-(

It is even more - it looks like I forgot to rename:
- setMinOccurence method and its argument
- setMaxOccurence
- patterns

so I created this PR.

Because we cannot have pair:
- setMinOccurence (single *r*)
- setMaxOccurrence (double *r*)

this PR is required for consistency and to clean a mess. Sorry!

I think now we have consistent min/max methods & their arguments & correct patterns. Someone please take a look.

I am ready to merge.


surli:
It still concerns typo in patterns, @pvojtechovsky are you ok to merge?

pvojtechovsky:
I am OK with merge. May be we should release Spoon then again to minimize amount of clients of that old wrong API.

surli:
> May be we should release Spoon then again to minimize amount of clients of that old wrong API.

I might release Spoon at the end of august if needed. Don't hesitate to open an issue if you feel the need.

zielint0:
@surli 
@pvojtechovsky 
Thank you very much!

Issue #2254, opened 2018-07-21
Spoon CompilationUnit improvements


pvojtechovsky:
This issue is related to #1785

Sniper mode printer needs to handle `package reference` , `import x.y.z;` and list of types of compilation unit.
It brings several requirements on compilation unit.

- [x] The compilation unit should implement `SourcePositionHolder` so it may act as root of the tree of source fragments (see #2255)
- [x] The compilation unit should have new property `CtPackageDeclaration packageDeclaration`
- [x] `packageDeclaration` should hold `SourcePosition` of the line `package x.y.z;`. (see #2811) 
- [x] `CtCompilationUnitImpl#declaredTypeRefs` should be declared as @MetamodelPropertyField(DECLARED_TYPE_REF) ... including @PropertyGetter and @PropertySetter on interface and appropriate `RoleHandler`
 - [x] `CtCompilationUnitImpl#imports` should be declared as @MetamodelPropertyField(DECLARED_IMPORT) ... including @PropertyGetter and @PropertySetter on interface and appropriate `RoleHandler`
 - [x] `CtCompilationUnitImpl#packageDeclaration` should be declared as @MetamodelPropertyField(PACKAGE_DECLARATION) ... including @PropertyGetter and @PropertySetter on interface and appropriate `RoleHandler`
- [x] `CtCompilationUnit` should implement `CtElement` (which extends CtVisitable) and `CtScanner` should visit package reference, imports and all declared type references.
- [x] `CtImport.getParent()` should return instance of `CompilationUnit` (therefore it must extend `CtElement`)

**Concept related to Spoon AST model**
`CtCompilationUnit` is not accessible from `CtModel`. It is not part of it's parent/child tree.
Each `CtCompilationUnit` is a root of own model, which has no parent, and has these children:
* package reference
* module reference
* imports
* references to included types.

WDYT? Do you see any problem? Drop here a note if you would have time to implement some part of that.

tdurieux:
`CompilationUnit should implement CtElement` is a big change, and kind of broke the tree representation of the AST. Honestly, I don't fully understand the impact that this change can have on existing clients.

But I think it is important.

It also solves an issue that we have with comments, such as license, that are more associated with the file and not really with a class.

For the moment, I don't see any problem.

pvojtechovsky:
Last changes suggested here were merged in last days. So I am closing this issue.

PR #2255, opened 2018-07-21
review: feat: CompilationUnit extends SourcePositionHolder

pvojtechovsky:
`CompilationUnit` extends `SourcePositionHolder` now, so it can act as root of source fragments - as needed for sniper mode pretty printer as discussed in #2254.


tdurieux:
Thanks, I will merge when I will be home.

PR #2256, opened 2018-07-21
Sonarqube-repair : remove useless assignments in ReferenceBuilder.java

ashutosh1598:
Removes 2 useless assignments

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
+++ b/src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
@@ -16,15 +16,28 @@
  */
 package spoon.support.compiler.jdt;
 
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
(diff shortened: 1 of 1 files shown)


ashutosh1598:
@pvojtechovsky , the copyright comment at the top was being printed twice - before the imports and after them. So I have removed one of it manually.

pvojtechovsky:
> the copyright comment at the top was being printed twice

yes I know. All the source code at beginning of CompilationUnit is problem for sniper mode printer. Therefore I opened #2254 which are needed to fix that.

Anyway thank you for experimenting with that and reporting that problem.

pvojtechovsky:
The algorithm of `ReferenceBuilder` is quite complicated. It is hard to say whether your changes are correct or not. I have feeling like the best solution would be to refactor field `bounds` to a parameter of `getTypeReference` function. 

@surli, WDYT?

ashutosh1598:
Ok. My bot does not do any bug detection. It simply asks SonarQube api for the location of the bug. So I did not think about whether it was correct or not.

surli:
The removal of the bounds assignment looks good to me here: it's a piece of code I wrote, and I might have forget to remove and old statement.
For the second removal, it looks like a local variable so it shouldn't have any impact on the remaining of the code.
So I'll merge it.

However I'm ok with you @pvojtechovsky: we should refactor it to use `bound` as a parameter.

PR #2257, opened 2018-07-21
fix(position): CtImport has source position now

pvojtechovsky:
CtImport has source position now. It is needed for Sniper mode printer.

... and I improved test `ModelUtils` here too, to allow configuration of Launcher before model is built. I guess it doesn't need separate PR. Or?

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/JDTImportBuilder.java
+++ b/src/main/java/spoon/support/compiler/jdt/JDTImportBuilder.java
@@ -26,6 +26,7 @@ import spoon.reflect.declaration.CtPackage;
 
 import spoon.reflect.declaration.CtType;
 import spoon.reflect.factory.Factory;
+import spoon.reflect.reference.CtReference;
 import spoon.reflect.declaration.CtImport;
 import spoon.reflect.visitor.filter.NamedElementFilter;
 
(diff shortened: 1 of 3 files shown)


tdurieux:
For me it is ok.
 
Is it ready to merge?

pvojtechovsky:
Yes, it is ready for merge

src/main/java/spoon/support/compiler/jdt/JDTImportBuilder.java
```diff
@@ -108,6 +109,13 @@ void build() {
 		spoonUnit.setImports(this.imports);
 	}
 
+	private CtImport createImportWithPosition(CtReference ref, ImportReference importRef) {
+		CtImport imprt = factory.Type().createImport(ref);
+		imprt.setPosition(factory.Core().createCompoundSourcePosition(spoonUnit, importRef.sourceStart(), importRef.sourceEnd(), importRef.declarationSourceStart, importRef.declarationEnd, spoonUnit.getLineSeparatorPositions()));
+		//TODO initialize source position of ref
```
tdurieux: I think the position of the reference should be `importRef.declarationSourceStart, importRef.declarationEnd`, no?

pvojtechovsky: probably yes, but I do not need it now, so the position of whole import was enough for this PR and sniper mode. I suggest to set position of reference in a later PR.

tdurieux: Ok for me.

PR #2258, opened 2018-07-21
fix: ReplacementVisitor doesn't eat exceptions

pvojtechovsky:
I had hard time to found a reason for a strange problem ... until I found that ReplacementVisitor eats exceptions. 

May be before there was not much code running in scope of this exception block, but now with integrated Change collector and sniper mode, there makes sense to report exceptions, which happens here.

@monperrus, do you have any idea why it was caught and ignored?


tdurieux:
Personally I don't remember why. It was probably because of invalid call but we should definitely handle this case

PR #2263, opened 2018-07-21
refactor: fix code inspection - remove unused assignment

zielint0:
Probable bugs
- Unused assignment

Unused assignment are confusing because we lost value later (it is overwritten line down or in ctr).
Example
int a = 1;
...
a = 0;

If the value is lost/overwritten, it is not needed.

The logic in src/main/java/spoon/support/template/Parameters.java is problematic. Should it be fixed?

I am ready to merge.


src/main/java/spoon/pattern/internal/node/ForEachNode.java
```diff
@@ -54,7 +54,6 @@ public ForEachNode() {
 	@Override
 	public boolean replaceNode(RootNode oldNode, RootNode newNode) {
 		if (iterableParameter == oldNode) {
-			oldNode = newNode;
```
pvojtechovsky: I have to fix this one in another PR. The correct code is something like `itetrableParameter = newNode` ... but it needs a test and some type cast ...

PR #2264, opened 2018-07-21
refactor: fix SonarQube: A NullPointerException could be thrown; o is…

zielint0:
… nullable here.

@surli 
@monperrus 
@pvojtechovsky 

Should not it be something like below?
```
if (o instanceof CtElementImpl) {

}
```

BTW. Is it good practice to throw in equals?
```
throw new IllegalStateException("violation of equal/hashcode contract between \n" + this.toString() + "\nand\n" + o.toString() + "\n");
```

I am ready to merge.


pvojtechovsky:
> if (o instanceof CtElementImpl)

you are right that would be better. You can use it even as subsitution for `o != null`. So could you change it in this PR?

PR #2266, opened 2018-07-23
review: fix(pattern): ForEachNode#replaceNode

pvojtechovsky:


zielint0:
Cann't -> Cannot

pvojtechovsky:
Tomasz pointed at this problem in another PR, so I noticed that there is wrong code here. This PR contains a fix, but I do not know how to test it. The test would have to 
1) define a pattern parameter
2) to use it as inline forEach 
3) to replace that parameter again
... does this use case makes sense? Hard to say...

zielint0:
Thank you for deeper inspection than in my PR :)

surli:
I'm ok to merge it without a test: but can you at least add some comments on the method to explain that specific situation?

surli:
Apparently you broke the checkstyle when pasting the comment I think. Could you fix it?

Issue #2268, opened 2018-07-24
regression: CommentTest.testStatementComments is failing on windows


surli:
Since the merge of #2246 there's a regression in `CommentTest.testStatementComments` on windows machine. See: https://ci.inria.fr/sos/job/spoon-windows/80/

pvojtechovsky:
Here is the fix #2269

PR #2269, opened 2018-07-24
fix: CommentTest#testStatementComments uses OS specific EOL

pvojtechovsky:
fix of #2268


surli:
Thanks Pavel it's green again :)

PR #2278, opened 2018-07-25
refactor: delete unused testclass - VisitorFake

zielint0:


pvojtechovsky:
This class is a test itself. If CtAbstractVisitor misses some method from CtVisitor, then `VisitorFake` class compilation will fail - test fails. So we should keep this file even if it looks like unused.

zielint0:
@pvojtechovsky 
Good to know! Thank you Pavel! It was tricky! ;-)
Could you add some comment to prevent such situation in future?

pvojtechovsky:
> Do you see any problem if I move it to "testclasses" directory?

it is special kind of test class. It is not analyzed by Spoon. So it probably doesn't belong to test classes.

> add some comment

there is already a comment which explains the purpose of that class.

zielint0:
@pvojtechovsky 

> there is already a comment which explains the purpose of that class.

Yes! I meant something like: *This class is not explicitly used by other classes but required. Don't delete it!*

pvojtechovsky:
It is good idea. Please make a PR which adds such comment.

zielint0:
@pvojtechovsky 

> It is good idea. Please make a PR which adds such comment.

add warning notive to Visitor #2287

PR #2283, opened 2018-07-26
review: feat(sniper): detailed tree of source fragments of compilation unit

pvojtechovsky:
this PR introduces new types:
* `SourceFragment`
* `ElementSourceFragment implements SourceFragment`
* `TokenSourceFragment implements SourceFragment`
* `CollectionSourceFragment implements SourceFragment`

`ElementSourceFragment` is a wrapper around `CtElement` and it's `SourcePosition`. It also builds hierarchy of `ElementSourceFragment`s, which mirrors hierarchy of fragments of source code of compilation unit.

It provides 3 views at tree of source fragments
1) Tree of `ElementSourceFragment`s. It consists of fragments defined by SourcePosition.start/end of `CtElement`s only. It is basic structure, which must be build after Spoon model is created and before it is modified.
2) Each `ElementSourceFragment` can be optionally expanded to **detailed tree** which contains `TokenSourceFragment`s next to child `ElementSourceFragment`s. See `ElementSourceFragment#getChildrenFragments()`.
3) Each **detailed tree** can be reorganized into so called **grouped tree**, which contains same information like **detailed tree**, but additionally SourceFragments, which belongs to same parent CtRole are grouped together and represented by `CollectionSourceFragment`

Notes:
* The detailed tree consists of small parser tokens, which are equal to tokens produced by DJPP and TokenWriter. 
* The detailed tree is based on simple SourcePositionImpl only. It doesn't need any details provided e.g. by `DeclarationSourcePosition`
* The detailed tree uses simple parser to detect spaces, separators, keywords and identifiers.


src/main/java/spoon/support/reflect/cu/CompilationUnitImpl.java
```diff
@@ -301,6 +306,31 @@ public SourcePosition getOrCreatePartialSourcePosition() {
 		return myPartialSourcePosition;
 	}
 
+	@Override
+	public ElementSourceFragment getRootSourceFragment() {
+		if (rootFragment == null) {
+			String originSourceCode = getOriginalSourceCode();
```
surli: `getOriginalSourceCode()` might return null if `getFile()` returns null. You should check the result to avoid NPE here.

src/main/java/spoon/support/reflect/cu/CompilationUnitImpl.java
```diff
@@ -301,6 +306,31 @@ public SourcePosition getOrCreatePartialSourcePosition() {
...
+			rootFragment = new ElementSourceFragment(this, null);
+			for (CtImport imprt : getImports()) {
+				rootFragment.addChild(new ElementSourceFragment(imprt, null /*TODO role for import of CU*/));
+			}
+			for (CtType<?> ctType : declaredTypes) {
```
surli: The CU can also be used for package declaration or module declaration: in those cases, you don't create the tree?

src/test/java/spoon/test/main/MainTest.java
```diff
@@ -474,6 +477,39 @@ public void scan(CtRole role, CtElement element) {
 		});
 	}
 
+	@Test
+	public void testSourcePositionTreeIsCorrectlyOrdered() {
+		List<CtType> types = rootPackage.filterChildren(new TypeFilter<>(CtType.class)).filterChildren((CtType t) -> t.isTopLevel()).list();
```
surli: Could you specify the contract?

src/main/java/spoon/support/reflect/cu/CompilationUnitImpl.java
pvojtechovsky: I forgot that cases. As fast solution: I will check them and `throw SpoonException(...not supported...)`. Later we can check if we can create the tree for them too - I have no idea yet if it is easily possible or  needs more effort.

surli: It looks ok for me. Maybe it worth opening an issue to avoid forgetting it.

monperrus:
It looks interesting. What's the final goal of this new feature?

What do you mean by "ElementSourceFragment is a wrapper around CtElement and its SourcePosition"?

pvojtechovsky:
> It looks interesting. What's the final goal of this new feature?

It is core feature need by #1927 - sniper mode.

> What do you mean by "ElementSourceFragment is a wrapper around CtElement and its SourcePosition"?

yes,
1) it connects element and source position
2) it knows children and sibling fragments, so it builds tree of source fragments

It is not enough to link `element` only and to use `CtElement#getPosition`, because 
* the position might be changed during refactoring and we need origin elements and origin source positions which maps to origin sources (before refactoring)

monperrus:
> It is core feature need by #1927 - sniper mode. 

Excellent!

> we need origin elements and origin source positions which maps to origin sources (before refactoring)

Do you mean something like  `CtElement#getOriginalPosition` and `CtElement#getOriginalElement`?

pvojtechovsky:
>> we need origin elements and origin source positions which maps to origin sources (before refactoring)

> Do you mean something like CtElement#getOriginalPosition and CtElement#getOriginalElement?

I would say no. But may be I do not understand your idea.

The concept of this PR and core feature for sniper mode is: 
* The compilation unit has origin source code
* The compilation unit has tree of `SourceFragment`s, which mirrors the tokens from origin source code
* There are different kinds of  `SourceFragment`s: ElementSourceFragment, TokenSourceFragment, ...
* ElementSourceFragment represents a fragment which mirrors the `CtElement`
* TokenSourceFragment represents a fragment of primitive tokens, like separators, newlines, etc. It is not mapped to any child `CtElement`.

So from this concept point of view there is relation from ElementSourceFragment to the CtElement, which was made during building of origin Spoon model. Opposite relation is not needed.

When the spoon model is changed later (some CtElements are cloned, added, modified, moved, removed, ... ) then these new CtElements doesn't have source position, etc. So it makes no sense to ask these elements for origin position or origin element.

monperrus:
> * The compilation unit has origin source code

Do you mean getRootSourceFragment?

> * The compilation unit has tree of `SourceFragment`s, which mirrors the tokens from origin source code

I would propose to rename ElementSourceFragment to CtElementMirror. WDYT?

> * ElementSourceFragment represents a fragment which mirrors the `CtElement`

I expect a field `CtElement mirroredElem` there. But there is no such field. What's the reason?

pvojtechovsky:
> The compilation unit has origin source code

I meant `String getOriginSourceCode()` method.

> I would propose to rename ElementSourceFragment to CtElementMirror. WDYT?
> I expect a field CtElement mirroredElem there. But there is no such field. What's the reason?

It looks like you look at the concept/feature from wrong direction...
The primary concept of this feature is to split source code (represented as one long String) to the small Strings which are semantically equal to TokenWriter tokens.
Then these tokens are organized into tree structure, which is represented by `SourceFragment`s.
Then some source fragments can be mapped to CtElements (see ElementSourceFragment).

There is no reason why EVERY CtElement should know it's ElementSourceFragment, but there is algoritm in CompilationUnit which can search and return ElementSourceFragment for CtElement - if there exist such ElementSourceFragment...

@monperrus Is it more clear now?

monperrus:
It's not clear but it's better. It looks interesting. 

As we've done in the past, are you OK if I propose some renaming and refactoring by pushing in your branch?

pvojtechovsky:
>  pushing in your branch?

Yes sure, I am ok with that. But it would be better for me if you would do the refactoring directly in #1927 - branch `refDJPP`. It contains code which uses classes from this PR. After you refactor it, I can rebase interactive your changes by way the part will become part of `feaSourceFragment` branch and another part will become part of `refDJPP`. 

Otherwise you will refactor only `feaSourceFragment` and I will have to then fix compilation problems in `refDJPP` manually without refactoring tool :-(

WDYT?

By the way I am open to discuss merging of new SourceFragment and old SourcePosition somehow - but before we can discuss it it needs your understanding of SourceFragment concept and implementation of #1927.

monperrus:
In #2419 this the same code, with a refactoring:

* removed getRootSourceFragment and getSourceFragment from CompilationUnit
* added getOriginalSourceFragment in SourcePositionHolder

Does it make sense to you?

pvojtechovsky:
I like it a lot!

I just do not know how it will behave it this method is called on the brand new CtElement, which was made by cloning or Factory#createXxx call. We should make a test and define some expected behavior.

monperrus:
> I like it a lot!
Super.

> We should make a test and define some expected behavior.
Just added a test with contract "on a new element, getOriginalSourceFragment returns the actual
element toString"

pvojtechovsky:
> Just added a test with contract "on a new element, getOriginalSourceFragment returns the actual
element toString"

May be it would be better to return some constant `SourceFragment NO_SOURCE_FRAGMENT` ... because
* there is no original source fragment in new CtElement
* the returned source fragment made by toString, doesn't conform to source fragment concept: source fragment consists of tree of smalled source fragments which represents TokenWriter tokens.

WDYT?

monperrus:
> May be it would be better to return some constant SourceFragment NO_SOURCE_FRAGMENT ... because

You're right, it would be  better.

monperrus:
do you continue based on the design we discussed in #2419?

pvojtechovsky:
OK, I can integrate these change into main PR today or tomorrow evening.

monperrus:
looking forward to it!

pvojtechovsky:
I do not understand why CI failed here. Is it problem of CI or this PR?

I am finished with my changes here.

monperrus:
Thanks Pavel. 

I've put the implementation classes in an `internal` package.

Let me think about it and add some more documentation.

monperrus:
and I want to have a look at the tests, which look super interesting!!

pvojtechovsky:
Why this change?
```java
ElementSourceFragment getOriginalSourceFragment() {
```
changed to 
```java
SourceFragment getOriginalSourceFragment()
```
It can never return SourceFragment. It is always ElementSourceFragment. ElementSourceFragment is not just implementation of SourceFragment interface. It means a special Source fragment which is linked to an CtElement. Other SourceFragments might be for example linked to a token.

monperrus:
Because of the design best practice that interface should only depend on other interfaces, and not
on concrete implementations.

One solution is to extract an interface out of ElementSourceFragment. WDYT?

pvojtechovsky:
> One solution is to extract an interface out of ElementSourceFragment

I understand, but If I remember well, you already agreed that in some cases Spoon uses interfaces too much. I do not think that it is helpful to have interface `ElementSourceFragment` and exactly one class `ElementSourceFragmentImpl` ... nobody will ever write different implementation for `ElementSourceFragment`
WDYT? ;-)

monperrus:
Yes, you're right, sometimes there are several design principles that contradict :-)

We can go back to |ElementSourceFragment.

We need anyhow to put them @Experimental.
|

monperrus:
Pavel, to your opinion, is ElementSourceFragment meant to be used by clients? or is it more an internal class for us?

For clients, what about a simple method in the interface:

	String getOriginalSourceFragment()

WDYT?

pvojtechovsky:
> String getOriginalSourceFragment()

short answer: No

long answer: `ElementSourceFragment getOriginalSourceFragment()` returns root ElementSourceFragment of current element. That root ElementSourceFragment contains subtree of all child fragments, which is the core feature needed to be able to print sources of element using origin sources as much as possible -> sniper mode.

>  is ElementSourceFragment meant to be used by clients? 

I would say, normally clients doesn't need that, same like they doesn't need SourcePosition. But these clients who wants to play with origin sources, will highly appreciate that detailed source tree.

monperrus:
then, I moved SourceFragment to package "internal" and we're ready to merge.

PR #2297, opened 2018-07-29
refactor: remove commented code from tests

zielint0:
It looks like that is is fixed/obsolete
```
		// for now the type of CtCatchVariable is not the same
		// this should be fix in the future (see: https://github.com/INRIA/spoon/issues/1420)
		//assertEquals(variable2, variable1);
```
But if it is needed, I will restore it.

@surli 
?

I am ready to merge.

### DIFF
--- a/src/test/java/spoon/test/exceptions/ExceptionTest.java
+++ b/src/test/java/spoon/test/exceptions/ExceptionTest.java
@@ -123,10 +123,5 @@ public class ExceptionTest {
 		CtCatchVariable variable2 = catches.get(1).getParameter(); // outside the lambda
 
 		assertEquals(variable1.getMultiTypes(), variable2.getMultiTypes());
-
-		// for now the type of CtCatchVariable is not the same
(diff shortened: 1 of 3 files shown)


src/test/java/spoon/test/exceptions/ExceptionTest.java
```diff
@@ -123,10 +123,5 @@ public void testUnionCatchExceptionInsideLambdaInNoClasspath() {
 		CtCatchVariable variable2 = catches.get(1).getParameter(); // outside the lambda
 
 		assertEquals(variable1.getMultiTypes(), variable2.getMultiTypes());
-
-		// for now the type of CtCatchVariable is not the same
-		// this should be fix in the future (see: https://github.com/INRIA/spoon/issues/1420)
-		//assertEquals(variable2, variable1);
```
surli: As it looks fixed we should instead uncomment the assertEquals.

zielint0: @surli 
Thank you for this hint!
I restored an assertion. All tests pass.

PR #2300, opened 2018-07-29
refactor: fix javadoc

zielint0:
@surli 
I corrected some Javadoc errors. Please review it deeply and suggest missing information. Thanks!

- ~~PatternParameterConfigurator.java -> I removed non-existing parameters. What is *pep*?~~
- ~~DefaultGenerator.java -> There is no class GeneratorImpl. Should it be a *Generator*?~~ -> yes
- ~~PatternPrinter.java -> I removed non-existing parameters. What is *parameterInfo*?~~
- ~~ElementNode.java -> *roleHandler* changed to *role*. Check full doc of this method please.~~
- ~~SingleHandler.java -> I removed non-existing *V*. Added *U*. Should I add one more *U*?~~

I am ready to merge.


src/main/java/spoon/reflect/meta/impl/SingleHandler.java
```diff
@@ -27,7 +27,7 @@
 /**
  * implementation of {@link RoleHandler}, which handles attributes of type &lt;V&gt;
```
surli: You missed a V here.

surli:
> - PatternParameterConfigurator.java -> I removed non-existing parameters. What is pep?
>  - PatternPrinter.java -> I removed non-existing parameters. What is parameterInfo?

@pvojtechovsky could you answer for those ones?

>  - DefaultGenerator.java -> There is no class GeneratorImpl. Should it be a Generator?

yes

>  - ElementNode.java -> roleHandler changed to role. Check full doc of this method please.
>  - SingleHandler.java -> I removed non-existing V. Added U. Should I add one more U?

Not sure to understand the question: what do you mean about adding one more U?

pvojtechovsky:
> PatternParameterConfigurator.java -> I removed non-existing parameters. What is pep?

@param pep pair of parameter and element which has to be transformed

> PatternPrinter.java -> I removed non-existing parameters. What is parameterInfo?

@param parameterInfo describes a to be printed pattern parameter 

> ElementNode.java -> roleHandler changed to role. Check full doc of this method please.

new comment is correct.

> SingleHandler.java -> I removed non-existing V. Added U. Should I add one more U?

No, one U is enough.

src/main/java/spoon/reflect/meta/impl/SingleHandler.java
zielint0: ?

zielint0:
@pvojtechovsky 
Thank you for all explanations.

zielint0:
What is `@param type` in spoon/pattern/internal/PatternPrinter.java 
`generatePatternParameterElement(ParameterInfo parameterInfo, Class<T> type)?`

pvojtechovsky:
> What is @param type in spoon/pattern/internal/PatternPrinter.java

it is a type/class of to be generated element

zielint0:
@pvojtechovsky 
Thank you!

src/main/java/spoon/reflect/meta/impl/SingleHandler.java
surli: It says that it handles attributes of type V but there is no type V defined in the class signature now.

zielint0: Class signature is: abstract class SingleHandler<T, U> extends AbstractRoleHandler<T, U, U>
so why 'V' should be documented?

surli: It should not but currently the comment is the following:
>[...] which handles attributes of type &lt;V&gt;

So V or `&lt;V&gt;` is currently documented and should not be.

zielint0: Ok, not it is not documented.

PR #1186, opened 2017-02-14
feature: remove FieldScopeFunction

pvojtechovsky:
removes FieldScopeFunction - issue #1167

PR #2331, opened 2018-08-02
refactor: uncomment a test in VariableReferencesTest.java

zielint0:
I uncommented it and it worked. Should it be permanently enabled?
@surli 
WDYT?

I am ready to merge technically. But politically?


surli:
@pvojtechovsky apparently you commented this test in #1186 so WDYT about uncommenting back?

zielint0:
@pvojtechovsky 
How about uncommenting this test?

pvojtechovsky:
you can uncomment it. I do not remember why it is commented. I just remember that FieldScope function was hard to implement and is not 100% correct in some cases. But these cases are probably not covered by that uncommented test.

zielint0:
@pvojtechovsky 
Thanks!

PR #2337, opened 2018-08-02
refactor: add override annotation in CtScanner.java

zielint0:

PR #2341, opened 2018-08-02
refactor: add override annotation CloneVisitor.java

zielint0:


monperrus:
same explanation as #2337

PR #2343, opened 2018-08-02
refactor: fix code inspection - collection to array

zielint0:
Use modern way to convert collection to array.

old: (mSourceClasspath.toArray(new String[mSourceClasspath.size()]));
new: (mSourceClasspath.toArray(new String[0]));

benefits:
- more readable
- faster to understand
- a bit faster

I am ready to merge.

### DIFF
--- a/src/main/java/spoon/IncrementalLauncher.java
+++ b/src/main/java/spoon/IncrementalLauncher.java
@@ -241,7 +241,7 @@ public class IncrementalLauncher extends Launcher {
 			setBinaryOutputDirectory(mClassFilesDir);
 		}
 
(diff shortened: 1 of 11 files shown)


surli:
Here again I'm ok to merge but I want some more sources: I'm not entirely sure it's more readable as you loose the concept of size of the array. A beginner might think the array will be empty at the end. But ok let's say now everyone knows that the array won't be empty.

What about the performance?
I'd think that as the method doesn't have to compute the size of the array it would be faster when passing an array with the proper size: so why would it be faster this way?

surli:
ok thanks

PR #2346, opened 2018-08-03
refactor: fix code inspection - split declarations

zielint0:
Code style issues
- Multiple variables in one declaration


surli:
I'm ok to merge it but what's the rationale behind it?
I understand the problem when it's a declaration and an assignation at the same time: it's not that clear, but when it's only a declaration I find this notation quite useful.

Like for: 
```java
boolean srcSuccess, templateSuccess;
```

I find it more concise.

zielint0:
@surli 
It is easy to overlook a variable.
One variable in one line is better for maintenance because of higher readability.

It was said by me, person that very values short code and low redundancy ;-)

PR #2353, opened 2018-08-07
refactor: move test outside testclasses

zielint0:
VariableReferencesModelTest.java is not a *testclass*. So I moved it out of testclasses directory. 

@surli 
Is it correct?


surli:
I don't think I agree with you for moving that one.
Apparently it has been put in `testclasses` in #1924 because it's not a unit test to be directly executed, but a file to be interpreted by Spoon.
In fact we even have a comment that says that: https://github.com/INRIA/spoon/blob/d07f50af94f8e9b25d50f6b1afa060950b0afbe6/src/test/java/spoon/test/query_function/VariableReferencesModelTest.java#L13

So IMHO it should remain where it is.

zielint0:
@surli
Thanks!

PR #2354, opened 2018-08-08
Convert from List to Set to speed up contains() logic.

aggie05engineer:
Using a List here results in extremely long runtimes when the source
folder is large because the contains() call is O(n). The logic
ends up placing this List into a Set anyway so just use a Set here
and make contains() checks cheap.

### DIFF
--- a/src/main/java/spoon/support/compiler/VirtualFolder.java
+++ b/src/main/java/spoon/support/compiler/VirtualFolder.java
@@ -45,7 +45,7 @@ public class VirtualFolder implements SpoonFolder {
 
 	@Override
 	public List<SpoonFile> getAllFiles() {
-		List<SpoonFile> result = new ArrayList<>();
+		Set<SpoonFile> result = new HashSet<>();
 
 		for (SpoonFile f : getFiles()) {
 			// we take care not to add a file that was already found in a folder
@@ -53,7 +53,9 @@ public class VirtualFolder implements SpoonFolder {
 				result.add(f);
 			}
 		}
-		return result;
+
+		List<SpoonFile> resultAsList = new ArrayList<>(result);
+		return resultAsList;
 	}
 
 	@Override
--- a/src/main/java/spoon/support/compiler/VirtualFolder.java
+++ b/src/main/java/spoon/support/compiler/VirtualFolder.java
@@ -45,17 +45,9 @@ public class VirtualFolder implements SpoonFolder {
 
 	@Override
 	public List<SpoonFile> getAllFiles() {
-		Set<SpoonFile> result = new HashSet<>();
-
-		for (SpoonFile f : getFiles()) {
-			// we take care not to add a file that was already found in a folder
-			if (!result.contains(f)) {
-				result.add(f);
-			}
-		}
-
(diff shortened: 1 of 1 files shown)


src/main/java/spoon/support/compiler/VirtualFolder.java
```diff
@@ -45,15 +45,17 @@ public void addFolder(SpoonFolder o) {
...
 		for (SpoonFile f : getFiles()) {
 			// we take care not to add a file that was already found in a folder
 			if (!result.contains(f)) {
 				result.add(f);
 			}
 		}
-		return result;
+
+		List<SpoonFile> resultAsList = new ArrayList<>(result);
+		return resultAsList;
```
zielint0: I recommend to avoid unnecessary local variable:

```
List<SpoonFile> resultAsList = new ArrayList<>(result);
return resultAsList;
```

Just return what is needed:
```
return new ArrayList<>(result);
```
It is shorter and easier to understand.

src/main/java/spoon/support/compiler/VirtualFolder.java
```diff
@@ -45,15 +45,17 @@ public void addFolder(SpoonFolder o) {
 
 	@Override
 	public List<SpoonFile> getAllFiles() {
-		List<SpoonFile> result = new ArrayList<>();
+		Set<SpoonFile> result = new HashSet<>();
 
 		for (SpoonFile f : getFiles()) {
```
zielint0: f -> file

src/main/java/spoon/support/compiler/VirtualFolder.java
```diff
@@ -45,15 +45,17 @@ public void addFolder(SpoonFolder o) {
...
 			// we take care not to add a file that was already found in a folder
 			if (!result.contains(f)) {
```
surli: This test seems not needed anymore

src/main/java/spoon/support/compiler/VirtualFolder.java
surli: Actually, in order to improve the perf, it looks interesting to store the results locally in the instance to avoid computing it again each time. WDYT?

surli:
Thanks @aggie05engineer for your proposal!

src/main/java/spoon/support/compiler/VirtualFolder.java
aggie05engineer: I do not understand the code well enough unfortunately. Yesterday was my first day playing with spoon.

files appears to be mutable given addFile() and addFolder() methods. There is no need to call getFiles() here at all since getFiles() just returns a List representation of the Set<SpoonFile> files anyway. If there are thread concerns then there are already issues with non-protected blocks _so_ I think just using the files as is works ok? Submitting new change today.

surli: Yeah actually I had second thought about my comment after posting it. So let's forgot this idea for now.

pvojtechovsky:
I suggest to return immutable list. WDYT?

monperrus:
Thanks a lot for your contribution

PR #2355, opened 2018-08-08
WIP refactor: add 'final' keyword to utility classes

zielint0:
Prevent creating instances of utility classes.

Class structure
- Utility class is not 'final'


surli:
I'm mixed on this one: it's potentially breaking because some clients might have maybe extended those classes by adding a new class in the right package name, but I'm really not sure anyone did that.

So I'm in favor with this PR. Let's see what the others think: @pvojtechovsky @monperrus @tdurieux WDYT?

zielint0:
It is technically possible to subclass a static class. However it is not possible to override a static method, so no sane person extends static class. Extending utility class should be prevented by: making it final and having a private constructor.

https://stackoverflow.com/questions/2223386/why-doesnt-java-allow-overriding-of-static-methods

zielint0:
Ok. I am closing.

PR #2366, opened 2018-08-10
review fix: ignore a test for memory issues on the CI

surli:
One test has been added by @nharrand in #2112 to show the usage of the new dependency resolver. However this test consumes too much memory on our CI, on the job used to deploy maven artefacts: 
> [ERROR] Tests run: 7, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 68.03 s <<< FAILURE! - in spoon.MavenLauncherTest
[ERROR] testTypeResolution(spoon.MavenLauncherTest)  Time elapsed: 38.801 s  <<< ERROR!
java.lang.OutOfMemoryError: GC overhead limit exceeded
	at spoon.MavenLauncherTest.testTypeResolution(MavenLauncherTest.java:25)

https://ci.inria.fr/sos/job/Spoon-Snapshot-Deployer/1689/console

I propose to ignore the test for now.

### DIFF
--- a/src/test/java/spoon/MavenLauncherTest.java
+++ b/src/test/java/spoon/MavenLauncherTest.java
@@ -10,6 +10,7 @@ import spoon.reflect.visitor.CtScanner;
 import java.nio.file.Paths;
 import java.util.Arrays;
 import java.util.List;
+import org.junit.Ignore;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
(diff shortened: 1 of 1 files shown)


tdurieux:
Maybe we should check why the memory consumption is so high

surli:
> Maybe we should check why the memory consumption is so high

yes that's what I want to do, but right now the snapshots are not deployed anymore because of that test. So I propose to ignore it for fixing the deployment, and then to look what happens in the test concerning memory consumption.

surli:
So ok to merge it @tdurieux ?

surli:
@monperrus would you merge it? We don't have any snapshot since 10 days because of that issue...

PR #2377, opened 2018-08-15
WIP refactor: code inspection - fix redundant local variable

zielint0:
WIP
DO NOT MERGE

### DIFF
--- a/src/main/java/spoon/refactoring/CtRenameLocalVariableRefactoring.java
+++ b/src/main/java/spoon/refactoring/CtRenameLocalVariableRefactoring.java
@@ -84,11 +84,8 @@ public class CtRenameLocalVariableRefactoring extends AbstractRenameRefactoring<
 
 		@Override
 		public ScanningMode enter(CtElement element) {
-			CtElement ele = element;
-			if (ele.hasParent(ignoredParent)) {
-				if (element.hasParent(ignoredParent)) {
-					return ScanningMode.SKIP_ALL;
-				}
+			if (element.hasParent(ignoredParent)) {
+				return ScanningMode.SKIP_ALL;
 			}
 			if (element instanceof CtType) {
 				nrOfNestedLocalClasses++;


src/main/java/spoon/refactoring/CtRenameLocalVariableRefactoring.java
```diff
@@ -85,8 +85,7 @@ public void accept(CtReference t) {
 		@Override
 		public ScanningMode enter(CtElement element) {
 			if (ignoredParent != null && element instanceof CtElement) {
-				CtElement ele = (CtElement) element;
-				if (ele.hasParent(ignoredParent)) {
+				if (((CtElement) element).hasParent(ignoredParent)) {
```
surli: I really don't agree with this change: generally I'm the one extracting those casts in a local variable. I find it really more readable, and it's really useful when debugging. 

Now here I even don't understand why a cast is needed: element is already a CtElement, so we just need to remove the cast, don't we?

src/main/java/spoon/support/compiler/SnippetCompilationHelper.java
```diff
@@ -50,13 +50,12 @@ private SnippetCompilationHelper() { }
 
 	public static void compileAndReplaceSnippetsIn(CtType<?> c) {
 		Factory f = c.getFactory();
-		CtType<?> workCopy = c;
 		Set<ModifierKind> backup = EnumSet.noneOf(ModifierKind.class);
-		backup.addAll(workCopy.getModifiers());
-		workCopy.removeModifier(ModifierKind.PUBLIC);
+		backup.addAll(c.getModifiers());
```
surli: Ok with the remove of workCopy but then could you rename the argument of the method: `c` is really not something explicit... `workType` maybe or whatever.

src/main/java/spoon/refactoring/CtRenameLocalVariableRefactoring.java
zielint0: @surli 
> I really don't agree with this change: generally I'm the one extracting those casts in a local variable. I find it really more readable, and it's really useful when debugging.

Restored.

> Now here I even don't understand why a cast is needed: element is already a CtElement, so we just need to remove the cast, don't we?

We do :-)

src/main/java/spoon/refactoring/CtRenameLocalVariableRefactoring.java
```diff
@@ -84,9 +84,9 @@ public void accept(CtReference t) {
 
 		@Override
 		public ScanningMode enter(CtElement element) {
-			if (ignoredParent != null && element instanceof CtElement) {
-				CtElement ele = (CtElement) element;
-				if (ele.hasParent(ignoredParent)) {
+			CtElement ele = element;
```
surli: I don't understand the purpose of local variable ele here: IMHO you can just remove this variable and the first if.

zielint0: @surli 
Done.

zielint0:
WIP
DO NOT MERGE

PR #2387, opened 2018-08-17
refactor: fix SonarQube - make field 'private', use accessor

zielint0:
Change access level from 'public' to 'private'.
Use accessor in other places.

I even did not add an accessor - it existed so I used it. I assume that this public accessor (getContext()) was added to have this field private. Should not be a compatibility issue.


surli:
> I assume that this public accessor (getContext()) was added to have this field private. Should not be a compatibility issue.

I agree this need a change however it could be a compatibility issue for users that inherits from `DefaultJavaPrettyPrinter` and uses the field context. 
I propose that you add a deprecated annotation on it with a since 7.1.0. And add some javadoc to explain to use the getter instead. It will be removed in the next two releases.

PR #2389, opened 2018-08-18
refactor: rename getChilds to getChildren

zielint0:
Correct spelling.


surli:
Here it's package protected. It might be a compatibility problem if a client decided to inherits the class in a package with the same name.
I propose that you keep your change, but you add a `getChilds` method that uses `getChildren` and you add a deprecated annotation with since 7.1.0 and Javadoc explanation to use `getChildren`. It will be definitely removed in two releases.

zielint0:
@surli 
> I propose that you keep your change, but you add a getChilds method that uses getChildren and you add a deprecated annotation with since 7.1.0 and Javadoc explanation to use getChildren. 

Done.

PR #2395, opened 2018-08-19
refactor: remove unused variable from JDTTreeBuilderHelper.java

zielint0:
@surli 
It looks that local variable *parameterTypesOfExecutable* is unused.
WDYT?


monperrus:
Thanks!

PR #2398, opened 2018-08-20
refactor: split variables declaration in test

zielint0:
I am ready to merge.

### DIFF
--- a/src/test/java/spoon/test/main/MainTest.java
+++ b/src/test/java/spoon/test/main/MainTest.java
@@ -296,7 +296,7 @@ public class MainTest {
 		class Counter {
 			int scan;
 			int enter;
-			int exit = 0;
+			int exit;
 		}
 
 		final Counter counter = new Counter();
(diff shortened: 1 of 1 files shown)


src/test/java/spoon/test/main/MainTest.java
```diff
@@ -294,7 +294,9 @@ public void test() {
 
 	private void checkContractCtScanner(CtPackage pack) {
 		class Counter {
-			int scan, enter, exit = 0;
+			int scan;
```
monperrus: also explicitly initialize them to 0?

zielint0: @monperrus 
No. Instead of adding 2 redundant initializers, I removed 1 redundant initializer. So the class is now:
```
class Counter {
			int scan;
			int enter;
			int exit;
		}
```
Class' fields have default initialization to 0 (or: "", null) so adding redundant initializer is redundant and confusing. Developer may wonder: why this initializer is added here, if this value is overrided by JVM?

Also it is confusing to trick user by override initializer later:
```
int x = 0;
... // some non x-activity
x = 2;
```

I removed such things here: https://github.com/INRIA/spoon/pull/2263/files

Please note that with removing such initializers, I removed other mistakes here. So some analysis paid off ;-)

PR #2405, opened 2018-08-22
refactor(printer): simplify printing of CtTypeParameter

pvojtechovsky:
CtTypeParameter can never have qualified name and it can never use `super` as a bound type. It is also not needed to create reference to print CtTypeParameter.


monperrus:
Thanks Pavel

Issue #2401, opened 2018-08-21
Discussion: Meaning of CtTypeParameterReference#upper and #bounds


pvojtechovsky:
I am thinking about solution of #2161 and it looks like Spoon model is not fitting to java specification concerning some attributes of CtTypeParameterReference.

There are these problems
> P1 `CtTypeParameterReference#upper` makes sense only for `CtWildcardReferenceImpl`. 

The attribute `upper` of pure `CtTypeParameterReference` makes no sense or it should be unsettable and should always return true.

> P2 `CtTypeParameterReference#superType` should be settable only for `CtWildcardReferenceImpl`.

The attribute `superType` should be derived for `CtTypeParameterReferenceImpl`, because it can be always computed from `CtTypeParameter#superClass`.
The attribute `superType` should be settable (not derived) for `CtWildcardReferenceImpl` only.

WDYT?

pvojtechovsky:
It seems to me that the most clean solution would be 
1) to replace
`CtWildcardReference extends CtTypeParameterReference`
by
`CtWildcardReference extends CtTypeReference`

2) to remove CtTypeParameterReference methods `isUpper`, 'setUpper`, `setBounds`, `addBound`, `removeBound`, 'setBoundingType', because they are completely useless - they mirrors no java language feature, because type parameter reference has no bounds and no upper.

3) may be we can keep `CtTypeParameterReference#getBoundingType`, but it should be derived with implementation `getDeclaration().getSuperclass()` ... it would be also more clear if `CtTypeParameterReference#getSuperclass()` would return `getDeclaration().getSuperclass()` too

4) to add CtWildcardReference methods: `isUpper`, 'setUpper`, `getBoundingType`, `setBoundingType`. Note: `add/removeBound` or `setBounds` makes no sense here too, because wildcard reference doesn't support multiple bounds.

WDYT? 
Will you accept such deep changes in Spoon model? ... I guess they are needed, because otherwise the API is confusing with useless methods, whose calling has no influence to printed java source or causes not compilable sources.

I am still searching for a reason which explains why that "confusing" design is needed ... do you have any idea?

monperrus:
> I am still searching for a reason which explains why that "confusing" design is needed ... do you have any idea?

AFAIK, because CtTypeParameter has been in and out over time.

> Will you accept such deep changes in Spoon model?

In principle yes, unless we realize during the refactoring that we're confused once more :-)

pvojtechovsky:
[...]

Issue #2378, opened 2018-08-16
BUG: stack over flow when getAllMethods()


danglotb:
Hi all, I am reporting an error:

Spoon throws a `StackOverFlowError` when I am trying to call `getAllMethods()` on the following object:

```java
public class Matriochka extends SubMatriochka {
     public void aMethod()  {
     }
 }
```
```java
public class SubMatriochka {
     public <F extends Enum<F> & Factory> int aMethod(Class<F> object) {
        return 23;
    }
 }
```

_i.e._ the following snippet is throwing a  `StackOverFlowError`
```java
final Class<?> aClass = Matriochka.class;
final CtType<?> ctType = createFactory().Class().get(aClass);
ctType.getAllMethods();
```

You may find a test, with all the resources to reproduce the error on my [fork](https://github.com/danglotb/spoon/tree/stackoverflow-error) on the branch **stackoverflow-error**.

Thank you!

pvojtechovsky:
I guess it is related to known problem discussed in #2161. The solution needs some changes in Spoon model - so it needs some time.

monperrus:
fixed in #2406. don't hesitate to reopen if it does not work the the latest master.

PR #2406, opened 2018-08-22
review: refactor: CtTypeParameterReference

pvojtechovsky:
an experiment related to #2401


pvojtechovsky:
All the tests should pass now - I just did a rebase to minimize amount of changes in this PR.

This PR implements ideas from #2401 and fixes #2161 (the failing test is already part of this PR) and probably fixes also #2378 (I have not added that test here).

The main conceptual changes are:
* CtTypeParameterReference has no bounds and no super class. It holds only `simpleName` of type parameter - the methods were moved from CtTypeParameterReference to CtWildcardReference and some of them were even removed, because CtWildcardReference doesn't support `CtIntersectionTypeReference`. So both types now mirrors java features.
* CtExecutableReference#parameters must not contain `CtTypeParameterReference` (it will fail with SpoonException now), but they must contain their bounds/superClass instead.

Q1) WDYT?

I suggest another change (which is not part of this PR): `CtWildcardReference extends CtTypeReference<Object>` ... actually it extends `CtTypeParameterReference` and it seems to me confusing and makes check
```java
ref instanceof CtTypeParameterReference` 
```
tricky, because it usually have to look like this
```java
ref instanceof CtTypeParameterReference && !(ref instanceof CtWildcardReference)
```

Q2) WDYT?

And I am not sure what should we use for return type of `CtExectutableReference`
A) CtTypeParameterReference - same like before - current implementation of this PR
B) CtTypeReference with bounds/superClass of `CtTypeParameterReference` - same like CtExectutableReference parameters.

Q3) WDYT?

src/test/java/spoon/test/parent/ParentTest.java
```diff
@@ -212,10 +212,10 @@ public void testParentOfPrimitiveReference() throws Exception {
 	}
 
 	public static void checkParentContract(CtPackage pack) {
-		for(CtElement elem: pack.getElements(new TypeFilter<>(CtElement.class))) {
+		pack.filterChildren(null).forEach((CtElement elem) -> {
 			// there is always one parent
-			Assert.assertNotNull("no parent for "+elem.getClass()+"-"+elem.getPosition(), elem.getParent());
```
monperrus: AFAIU we can have both assertions there

monperrus:
Q1) It looks good! There are very few changes in the tests, which indeed shows that this was unnecessary

Q2) Maybe, but it seems we can discuss this separately in another PR #babypr 

Q3) Return type of which method?

pvojtechovsky:
> Q3) Return type of which method?

```java
interface A<T extends String> {
 T method();
}
```
and now there is question what `CtExecutableReference.getType()` should return:
A) CtTypeParameterReference -> `T`
B) CtTypeReference -> `String`

monperrus:
I would say we keep the current behavior for backward compatibility and we add a method "getResolvedType" or "getInferredType" with the new behavior.

monperrus:
I guess that we're ready for merge?

pvojtechovsky:
Yes this PR is ready for merge

monperrus:
Thanks Pavel for this conceptual cleaning

PR #2410, opened 2018-08-26
review: refactor: SignaturePrinter prints type parameter bounds, no name

pvojtechovsky:
Signature shouldn't contain name of type parameter reference, because these three methods declarations:
```java
<T extends String> void m(T a);
<S extends String> void m(S b);
void m(String c)
```
Should have the same signature: `m(String)`

WDYT?

### DIFF
--- a/src/main/java/spoon/support/visitor/SignaturePrinter.java
+++ b/src/main/java/spoon/support/visitor/SignaturePrinter.java
@@ -25,6 +25,7 @@ import spoon.reflect.reference.CtExecutableReference;
 import spoon.reflect.reference.CtIntersectionTypeReference;
 import spoon.reflect.reference.CtTypeParameterReference;
 import spoon.reflect.reference.CtTypeReference;
+import spoon.reflect.reference.CtWildcardReference;
 import spoon.reflect.visitor.CtScanner;
 
 /**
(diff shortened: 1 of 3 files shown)


surli:
What if the type parameter is defined at the class level? Such as:

```java
class Foo<T> {
 void m(T a);
}
```

or 
```java
class Bar<T extends String> {
 void m(T a);
}
```

surli:
> It should be same - the super class of the type parameter should be used.

But in case of `Foo` you don't know that T extends String...

pvojtechovsky:
in case of 
```java
class Foo<T> {
 void m(T a);
}
```
`m` has signature `m(Object)`

pvojtechovsky:
I see now, more cases, which are probably not handled well too.
```java
<T extends List<String>> void m(T a);
<T extends List<Integer>> void m(T b);
<T extends List> void m(T c);
void m(List<String> d);
void m(List<?> e);
void m(List f);
```
all of them should have same signature: `m(List)`

WDYT?

If you agree than we should use `CtTypeReference#getTypeErasure()` during signature printing too. May be it is task for another small PR.

surli:
I'm ok with the other changes you propose. For me they could be implemented in the same PR or in another one, as you prefer.

PR #2411, opened 2018-08-27
WiP: fix: CtType#getMethod, CtType#getConstructor

pvojtechovsky:
1) `CtType#getMethod` and `CtType#getConstructor` are now using same algorithm `CtTypeImpl#hasSameParameters` for matching of parameter types. 

2) CtTypeImpl#hasSameParameters is using *type erasure* to check if parameters are matching.


pvojtechovsky:
Looks like this change cannot be done as basic step without other changes introduced by #2406. The change in concepts of ExecutableReference introduced by #2406 are needed too.
So I am closing this PR

monperrus:
This is CtClass

msteinbeck:
`CtType` may also be an interface which, by design, has no constructor.

PR #2414, opened 2018-08-29
review: fix: CtArrayTypeReference type erasure

pvojtechovsky:
fixes type erasure of array in methods like: `public <I extends String> void methodWithArray(I[] param)` where type erasure should be `String[]`

### DIFF
--- a/src/test/java/spoon/test/ctType/CtTypeParameterTest.java
+++ b/src/test/java/spoon/test/ctType/CtTypeParameterTest.java
@@ -86,7 +86,7 @@ public class CtTypeParameterTest {
 		}
 		Class<?> paramType = declExec.getParameterTypes()[paramIdx];
 		// contract the type erasure given with Java reflection is the same as the one computed by spoon
-		assertEquals("TypeErasure of executable param " + getTypeParamIdentification(typeParam), paramType.getName(), typeParam.getTypeErasure().toString());
+		assertEquals("TypeErasure of executable param " + getTypeParamIdentification(typeParam), paramType.getTypeName(), param.getType().getTypeErasure().toString());
 	}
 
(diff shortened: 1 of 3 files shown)


src/test/java/spoon/test/ctType/CtTypeParameterTest.java
```diff
@@ -86,7 +86,7 @@ private void checkTypeParamErasureOfExecutable(CtTypeParameter typeParam) throws
 		}
 		Class<?> paramType = declExec.getParameterTypes()[paramIdx];
 		// contract the type erasure given with Java reflection is the same as the one computed by spoon
-		assertEquals("TypeErasure of executable param " + getTypeParamIdentification(typeParam), paramType.getName(), typeParam.getTypeErasure().toString());
+		assertEquals("TypeErasure of executable param " + getTypeParamIdentification(typeParam), paramType.getTypeName(), param.getType().getTypeErasure().toString());
```
monperrus: it's strange to change both the expected value and the tested one.

what is the new value of `typeParam.getTypeErasure()` after the fix?

src/test/java/spoon/test/ctType/CtTypeParameterTest.java
```diff
@@ -103,7 +103,7 @@ private void checkParameterErasureOfExecutable(CtParameter<?> param) {
 		Class<?> paramType = declExec.getParameterTypes()[paramIdx];
 		assertEquals(0, typeErasure.getActualTypeArguments().size());
 		// contract the type erasure of the method parameter given with Java reflection is the same as the one computed by spoon
-		assertEquals("TypeErasure of executable " + exec.getSignature() + " parameter " + param.getSimpleName(), paramType.getName(), typeErasure.getQualifiedName());
```
monperrus: same thing

PR #2419, opened 2018-08-30
WIP: brainstorming around #2283

monperrus:


src/main/java/spoon/support/reflect/CtExtendedModifier.java
```diff
@@ -87,4 +88,14 @@ public int hashCode() {
 		result = 31 * result + (kind != null ? kind.hashCode() : 0);
 		return result;
 	}
+
+	@Override
+	public SourceFragment getOriginalSourceFragment() {
+		return new SourceFragment() {
```
pvojtechovsky: I suggest to use same/very similar implementation like `CtElement#getOriginalSourceFragment`. It looks like it should be possible.

pvojtechovsky: And then all `getOriginalSourceFragment` may return `ElementSourceFragment`

pvojtechovsky:
All ideas from this PR were moved to #2283. So closing this one

PR #2432, opened 2018-09-01
review: fix(position): labeled statement

pvojtechovsky:


src/main/java/spoon/support/compiler/jdt/JDTTreeBuilder.java
```diff
@@ -1175,9 +1175,26 @@ public boolean visit(IntLiteral intLiteral, BlockScope scope) {
...
+		//context.enter(factory.Core().createBlock(), block);
+		ContextBuilder.LabeledStatementHolder outerLabelHolder = context.label.peek();
+		if (outerLabelHolder != null && outerLabelHolder.wasAssignedToAStatement == false) {
+			/*
+			 * It is the case of nested label like:
+			 * label1: label2: while(x) {}
+			 * We have to add implicit block
+			 * label1: {label2: while(x) {}}
+			 */
+			how to create implicit CtBlock? Does it belong here or somewhere else?
```
pvojtechovsky: how to create implicit CtBlock? Does it belong here or somewhere else?
@tdurieux Could you implement it or give me a tip?

pvojtechovsky:
@tdurieux I do not know how to create implicit block in JDTTreeBuilder for this case
```java
label1: label2: while(x);
```
the required behavior is this result (with implicit block)
```java
label1: {label2: while(x);}
```

tdurieux:
I have to check the visiting order in the case of the label.

It is a mess to add an element that is not present in the JDT AST. I think the easiest solution would be to add the block and the label here:

https://github.com/INRIA/spoon/blob/c74c80ee7aa6ddafbf8904936d37cbf066bd550f/src/main/java/spoon/support/compiler/jdt/ParentExiter.java#L440

but I don't know if all the information is available at that location :/

tdurieux:
I succeed to make the tests pass. I'm not a big fan of my implementation but I don't know how to improve it.
@pvojtechovsky maybe you will find a way to make it better.

pvojtechovsky:
@tdurieux, thank you for that solution. Yes, it is tricky to understand it, but after debugging and thinking about that I cannot see a test case where it will fail. So that solution is OK for me.

I had an idea how to do it different:
* to create implicit CtBlock for each LabeledStatement directly in `JDTTReeBuilder#visit(LabeledStatement,BlockScope)`.
* to implement ParentExiter by way which would remove implicit CtBlock and apply it's label to the child statement of the CtBlock, in the case when this child statement has no label yet.

That seems to me as simpler and less tricky solution, but I have not tried it and I am not sure whether it is really simpler ;-)
WDYT, should we give it a try or will we merge PR as it is now?

tdurieux:
> I had an idea how to do it different

I think it should work, it is probably better because it will be the same algo for all the techniques.

pvojtechovsky:
So I will try to implement that second alternative in another PR - so we see whether it works and more maintenable.

tdurieux:
Thanks!

pvojtechovsky:
This PR contains the 3rd alternative. It seems to be the less code and easier to understand. 
@tdurieux  WDYT?

tdurieux:
It looks good for me!

PR #2435, opened 2018-09-01
refactor: inline local variable - CtRenameLocalVariableRefactoring.java

zielint0:


src/main/java/spoon/refactoring/CtRenameLocalVariableRefactoring.java
```diff
@@ -84,9 +84,8 @@ public void accept(CtReference t) {
 
 		@Override
 		public ScanningMode enter(CtElement element) {
-			if (ignoredParent != null && element instanceof CtElement) {
-				CtElement ele = (CtElement) element;
-				if (ele.hasParent(ignoredParent)) {
+			if (ignoredParent != null && element != null) {
+				if (element.hasParent(ignoredParent)) {
```
pvojtechovsky: Thank you, it is really better now. Would not it be even easier, if there would be only one `if` and longer condition?

zielint0:
@pvojtechovsky 
You are right. I combined 'if's. Thanks! :-)

PR #2437, opened 2018-09-03
refactor(MavenLauncher)

nharrand:
The goal of this PR is to refactor MavenLauncher by:
 * Not regenerating classpath files if changes are younger than an hour.
 * Removing classes that were not used anymore (`spoon.internal.mavenlauncher.Version`, `spoon.internal.mavenlauncher.RangeVersion`, `spoon.internal.mavenlauncher.TreeDependency`)
 * Improving `MavenLauncherTest`
 * Fixing a bug where additional classpath temporary file where present in ressources and therefor incorrectly accounted for.
	Now, `MavenLauncher#buildClasspath` follow maven module structure.
 * Fixing a bug where MavenLauncher crashed for pom that contained a maven-compiler-plugin with no configuration.
 * Adding a feature: MavenLauncher now also looks into the default profile to find source version compliance.
 * Fixing a bug which made InheritanceModel incorectly read properties composed of the concatenation of a string and another property

:warning: This changes some of the MavenLauncher constructors (by removing a now obsolete parameter (m2RepositoryPath)), but it does not change the simpliest one (`MavenLauncher(String mavenProject, SOURCE_TYPE sourceType)`).


nharrand:
This PR is now ready for code review!

src/main/java/spoon/MavenLauncher.java
```diff
@@ -58,50 +61,50 @@
...
-
 	/**
+	 * MavenLauncher constructor assuming either an environment
+	 * variable M2_HOME, or that mvn command exist in PATH.
 	 *
 	 * @param mavenProject the path to the root of the project
-	 * @param m2RepositoryPath the path to the m2repository
 	 * @param sourceType the source type (App, test, or all)
 	 */
-	public MavenLauncher(String mavenProject, String m2RepositoryPath, SOURCE_TYPE sourceType) {
```
surli: Actually instead of breaking the compatibility by removing this constructor, you could just deprecate this method and document the fact that m2repositoryPath is not using anymore. 
Same for the other constructor: it would be removed in the next release +1

surli:
> Not regenerating classpath files if changes are younger than an hour.

Is it possible to force the regeneration even if changes were made less than an hour ago?

surli:
> Not regenerating classpath files if changes are younger than an hour.

Is it possible to force the regeneration even if changes were made less than an hour ago?

src/main/java/spoon/MavenLauncher.java
nharrand: Yes, that makes sense. I'll do that ASAP.

nharrand:
Currently, you need to delete `spoon-classpath.tmp` files. But I can add a constructor with a boolean parameter to force regeneration. WDYT?

monperrus:
@tdurieux do you merge?

PR #2438, opened 2018-09-03
WiP: alternative fix(position): labeled statement

pvojtechovsky:
This is alternative solution to #2432. I would prefer this solution. It needs less tricks.

The only tricky thing is the way how helper CtBlock is removed from the Spoon model in ParentExiter - see call of `jdtTreeBuilder.getContextBuilder().replaceContextElement(block, child, childJDT);`

@tdurieux do you see a better place/way, which allows to merge helper CtBlock with it's child statement?


tdurieux:
It is better than my solution.

(still not easy to understand :/)

PR #2447, opened 2018-09-05
review: fix bug in autoimports

### LINKED ISSUE #2369: Auto-imports of java.lang classes fail to use qualified names if class of same name is defined in file.
Hello,

If I try to run an Launcher (with auto imports) on the following java code:

~~~
package TestJava;

public class JavaLongUse {

    public class Long {

    }

    public static long method() {
        return java.lang.Long.parseLong("10000");
    }

        System.out.println(method());
    }
}
~~~

I get the following output:

~~~
package TestJava;


public class JavaLongUse {
    public static class Long {}

    public static long method() {
        return Long.parseLong("10000");
    }

        System.out.println(JavaLongUse.method());
    }
}
~~~

Which leads to the compilation error "The method parseLong(String) is undefined for the type JavaLongUse.Long", because "java.lang.Long" disappears in the output file, and "Long" now refers to the JavaLongUse.Long class.

Is there a way to force qualified names in specific places?

monperrus:
fix #2369 

plus some boy-scout cleaning and documentation


monperrus:
Thanks @creichen for reminding us about the importance of this bug

PR #2452, opened 2018-09-06
refactor: simplify value check - SwitchNode.java

zielint0:


monperrus:
IMHO, the replaced code is more explicit, more understandable. I would not go in this direction.

pvojtechovsky:
I agree with Martin, The code of this PR is not easily understandable for me

Issue #675, opened 2016-05-24
feature: add support for analysis of binary code with a decompiler and shadow classes


msteinbeck:
It would be nice to have some sort of decompiler in spoon. This way visitors that gather metrics for entities (for instance the ATFD metric of particular classes) could access functions of libraries.

monperrus:
I agree this would be nice,  this point is indeed already in the roadmap. if you'd like to work on it soon, we can give you some hints.

retuxx:
>  if you'd like to work on it soon, we can give you some hints.

What do you mean by that. Is there some code available already?

tdurieux:
I used several month ago a decompiler in order to improve the usage of the templates:
https://github.com/INRIA/spoon/pull/464/files

@GerardPaligot checks several java decompiler but none are perfect.

GerardPaligot:
After my researches, the best one seems to be CFR (http://www.benf.org/other/cfr/) but yes, it isn't perfect.

retuxx:
> [...] but yes, it isn't perfect.

It doesn't have to be perfect IMHO.

retuxx:
So do you have plan to integrate one of the mentioned decompiler into spoon?

GerardPaligot:
We made some researches but we don't have any plan to integrate it at the short term.

monperrus:
Reopening this issue, because it's relevant and @nharrand is interested in it

monperrus:
as decompiler, I had good results in the past with CFR: http://www.benf.org/other/cfr/

nharrand:
We might also want to consider these:
[Krakatau](https://github.com/Storyyeller/Krakatau)
[Procyon](https://bitbucket.org/mstrobel/procyon/wiki/Java%20Decompiler)
[fernflower](https://github.com/fesh0r/fernflower)

monperrus:
Actually, @gluckzhang may use such a feature also!

nharrand:
I guess this issue could be closed now that we have a `spoon-decompiler` module.

monperrus:
yes!

monperrus:
Food for thought:
Beyond Binary Program Transformation
https://pdfs.semanticscholar.org/e5c2/9d46c1300ab5353bb4df08ed04a0ea9ec80a.pdf

SeanCoek:
It seems JarLauncher only support jar files analysis. Is other kind of format be support? say "war" file or just the original bytecode files with project directory.

[...]

PR #2455, opened 2018-09-07
feature: add support for Jar files analysis with JarLauncher (decompile then build AST with spoon)

nharrand:
This PR aims at providing support for JAR annalysis. (Related to #675 ).
My typical usecase is to be able to download an artifact's jar from maven central (and optionally its pom to get dependency information) to decompile it  and build a spoon AST.

This code provide a small wrapper named JarLauncher that does just that.

Still lack:
 * (fixing git commits)
 * some proper testing
 * documentation
 * an option to not decompile but instead build only shadow types?
 * support for alternative decompiler?
 * use maven install plugin for cleaner deployement of CFR?

As the decompiler I use ( [CFR](http://www.benf.org/other/cfr/) ) is not avilable on maven central I propose to add its jar in /libs.
WDYT?


src/main/java/spoon/decompiler/CFRDecompiler.java
```diff
@@ -0,0 +1,11 @@
+package spoon.decompiler;
+
+import java.io.File;
+import org.benf.cfr.reader.Main;
+
+public class CFRDecompiler implements Decompiler {
+	@Override
+	public void decompile(File jar, File outputDir) {
+		Main.main(new String[]{jar.getAbsolutePath(), "--outputdir", outputDir.getAbsolutePath()});
```
monperrus: is their a proper Java API?

src/main/java/spoon/decompiler/Decompiler.java
```diff
@@ -0,0 +1,16 @@
...
+
+	/**
+	 * Sets the output directory for source generated.
+	 *
+	 * @param jar
+	 * 		Path to jar to be analyzed.
+	 * @param outputDir
+	 * 		Path for the output directory for decompilation.
+	 */
+	void decompile(File jar, File outputDir);
```
monperrus: What about

 	List<SpoonFile> decompile(String jarPath); 
 	List<SpoonResource> decompile(String jarPath); 
 	CtModel decompile(String jarPath);

src/main/java/spoon/JarLauncher.java
```diff
@@ -0,0 +1,122 @@
...
+import spoon.decompiler.Decompiler;
+import spoon.internal.mavenlauncher.InheritanceModel;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+
+import static java.nio.file.StandardCopyOption.REPLACE_EXISTING;
+
+public class JarLauncher extends Launcher {
```
monperrus: @Experimental

pom.xml
```diff
@@ -295,6 +296,13 @@
       <artifactId>maven-invoker</artifactId>
       <version>3.0.1</version>
     </dependency>
+    <dependency>
+      <groupId>org.benf</groupId>
+      <artifactId>cfr</artifactId>
+      <version>${version.cfr}</version>
+      <scope>system</scope>
```
msteinbeck: Scope `system` will not package the dependency into the resulting jar. You need to create a local maven repository and add it to git. See: https://sookocheff.com/post/java/local-maven-repository/

nharrand: Thanks. I'll do that.

src/main/java/spoon/decompiler/Decompiler.java
nharrand: I'm not convinced by either of this solutions. My plan was to call the decompiler only if the input jar has been modified after last decompilation or if user forces it. I would prefer having directory declaration in JarLauncher than in Decompiler.

src/main/java/spoon/decompiler/CFRDecompiler.java
nharrand: Not as far as I know.

nharrand:
Ready for review.

monperrus:
LGTM.

Could you add the corresponding documentation in `doc`?

pom.xml
```diff
@@ -306,6 +306,9 @@
           <groupId>org.benf</groupId>
           <artifactId>cfr</artifactId>
           <version>0.132.0</version>
+          <!-- since it is not on Maven central, we don't package it -->
+          <!-- if you need it, download it at http://www.benf.org/other/cfr/ and put the jar in the classpath yourself -->
+          <scope>provided</scope>
```
nharrand: It is not on maven central, but I've upload it on `maven.inria.fr`, so it is still publicly available. (So we could package it if we wanted it.)

monperrus:
What's the license of CFR? Is it compatible with LGPL?

nharrand:
It's MIT.

PR #2459, opened 2018-09-10
refactoring(Adding a SpoonPom class)

nharrand:
The aim of this PR is to regroup all pom manipulating utilities into a class named `SpoonPom`.
Hopefully this will become handy when working on poms in `JarLauncher`.
This consists only in renaming and moving around a couple of methods.


nharrand:
Ready for review!

PR #2461, opened 2018-09-11
review: test: MavenLauncherTest should not depend on spoon as a test resource

monperrus:
per our discussion with @nharrand


nharrand:
Would it be possible that the classpath depends on the content of your .m2 directory? (I mean in case the dependencies declared have non fixed versions and for some of it their dependencies change from one version to another.)

monperrus:
@tdurieux do you merge?

tdurieux:
@monperrus what do you mean by MavenLauncher should not depend on spoon itself.
I don't understand the purpose of this PR

monperrus:
Sorry, I mean "MavenLauncher**Test** should not depend on spoon as a test resource"

src/test/java/spoon/MavenLauncherTest.java
```diff
@@ -42,39 +42,36 @@ public void scan(CtElement element) {
...
-		// number of the sub folders of src/main/java and src/test/java
-		assertTrue("size: " + launcher.getModelBuilder().getInputSources().size(), launcher.getModelBuilder().getInputSources().size() >= 220);
+		assertEquals(8, launcher.getEnvironment().getComplianceLevel());
+		assertEquals(0, launcher.getModelBuilder().getInputSources().size());
+		// more with ALL_SOURCE
+		assertEquals(166, launcher.getEnvironment().getSourceClasspath().length);
 
 		// specify the pom.xml
-		launcher = new MavenLauncher("./pom.xml", MavenLauncher.SOURCE_TYPE.APP_SOURCE);
+		launcher = new MavenLauncher(testProject, MavenLauncher.SOURCE_TYPE.APP_SOURCE);
```
pvojtechovsky: this line looks same like line 49 above. 
A) delete it
B) let it use pom.xml as says contract of the test

src/test/java/spoon/MavenLauncherTest.java
```diff
@@ -42,39 +42,36 @@ public void scan(CtElement element) {
...
 		assertEquals(8, launcher.getEnvironment().getComplianceLevel());
 
 		// without calling maven to generate classpath
-		launcher = new MavenLauncher("./pom.xml", MavenLauncher.SOURCE_TYPE.APP_SOURCE, new String[]{});
+		launcher = new MavenLauncher(testProject, MavenLauncher.SOURCE_TYPE.APP_SOURCE, new String[]{});
```
pvojtechovsky: shouldn't be there pom.xml somehow too? I do not understand that use case

pvojtechovsky:
rebase?

monperrus:
@nharrand would you take over this one? you're now an (the?) expert of the MavenLauncher behavior. Thanks!

nharrand:
@monperrus Yes I'll try.
If I sum up a little why we're having this issue:
 * The new implementation of MavenLauncher get the list of class path element from Maven.
 * The original motivation for this was to obtain the correct classpath and therefor beeing able to resolve all classes of a maven project (even only as a shadow class).
 * Testing the motivating property (resolution of all type) on spoon itself is too resource expensive #2366
 * Testing the non regression of the number of classpath element on spoon itself is annoying as this number changes frequently

I think the best solution is to test that MavenLauncher indeed call maven to create non empty files containing classpath element, that those element are valid, and trust Maven for the completeness.

If there are no objections, I'll remove the check on the number of element (on spoon) and implement the proposed changes.

monperrus:
closed per #2494

PR #2362, opened 2018-08-09
refactor: use isEmpty() in RangeVersion.java

zielint0:
Improve readability.

I am ready to merge.

PR #2328, opened 2018-08-01
refactor: use isEmpty() for better readability also in tests

zielint0:
I am ready to merge.

Issue #2105, opened 2018-06-24
refactor: fix all IDEA code inspection warnings


zielint0:
**Please do not close this issue before 2019-01-01**

Proposition of IntelliJ IDEA Code Inspection. I will create a ruleset from this task.

Class structure
- Utility class is not 'final'
   - #2355 (8 warnings) 2018-08-09

Code style issues
- 'expression.equals("literal")' rather than '"literal".equals(expression)' :+1: 
   - ~~#2365 (24 warnings) 2018-08-10~~
- C-style array declaration :+1: 
   - ~~#2229 (3 warnings) 2018-07-16~~
- 'size() == 0' replaceable with 'isEmpty()' :+1: 
   - ~~#2232 (68 warnings) 2018-07-16~~
- Multiple variables in one declaration :+1: 
   - ~~#2346 (4 warnings) 2018-08-04~~
- Redundant 'new' expression in constant array creation :+1: 
   - ~~#2344 (3 warnings) 2018-08-03~~
- Unnecessary call to 'super()' :+1: 
   - ~~#2274 (76 warnings) 2018-07-24~~

- Java language level migration aids
   - Java 5
      - 'StringBuffer' may be 'StringBuilder' :+1: 
         - ~~#2230 (1 warning) 2018-07-16~~
      - 'while' loop replaceable with 'foreach' :+1: 
         - ~~#2231 (2 warnings) 2018-07-16~~
   - Java 7
      - Explicit type can be replaced with <> :+1: 
         - ~~#2227 (11 warnings) 2018-07-16~~
- Unnecessary semicolon :+1: 
  - ~~#2226 (25 warnings) 2018-07-15~~

Control flow issues
- Pointless boolean
   - #2242 (102 warnings) 2018-07-18
- Unnecessary 'null' check before 'instanceof' expression or call :+1: 
   - ~~#2267 (3 warnings) 2018-07-23~~

Data flow
 - Redundant local variable
   - ~~#2275 (10 warnings) 2018-07-24~~

Declaration redundancy
- Redundant throws clause :+1:
  - ~~#2221 (6 warnings) 2018-07-15~~ 

Imports
- Unused import :+1:
  - ~~#2104 (105 warnings) 2018-06-24~~ 

JUnit
- Misordered 'assertEquals()' arguments
   - ~~#2314 (14 warnings) 2018-07-31~~

Performance
- 'Collection.toArray()' call style
   - ~~#2343 (11 warnings) 2018-08-02~~
[...]

PR #2468, opened 2018-09-15
refactor: use isEmpty to make intention clear

zielint0:


pvojtechovsky:
I personally prefer positive `size() > 0`. so it is ok for me like it is now. If others thinks that `!isEmpty()` is better, them I can live with that too :-)

zielint0:
I have fixed previous defects like this.
This PR fixes some new ones.

monperrus:
`size() > 0` is slightly more explicit (you don't have to know the meaning and contracts of isEmpty).

If Pavel prefers it, let's keep it.

Thanks @zielint0 for the proposal.

zielint0:
@pvojtechovsky 
@monperrus 
@surli 

I fixed a lot of them in:
https://github.com/INRIA/spoon/pull/2362
https://github.com/INRIA/spoon/pull/2348
https://github.com/INRIA/spoon/pull/2328
https://github.com/INRIA/spoon/pull/2282

These are the last ones to fix.
It was merged by @surli so I think that this is useful.

Please consider reopening this pull request.

zielint0:
@monperrus 
To make reviewers life easy, I created a list of merged PRs.
If you want to see rationale or fact if something similar was merged or not, please take a look:
https://github.com/INRIA/spoon/issues/2105

pvojtechovsky:
I checked #2362 #2348 #2328 #2282 and I agree that it is more readable only in this case:
```
		if (splitRange.length == 1 || splitRange[1].length() == 0) {
change to
		if (splitRange.length == 1 || splitRange[1].isEmpty()) {
```
because it is positive expression: "If empty then do something", but in all other cases I need more time to understand new expressions, because they often contain negative condition. E.g.
`if (!clone.getBody().getStatements().isEmpty()) {` is not better then before

and the worst case is:
```
		assertTrue(factory.Type().getAll().size() > 0);
changed to
 		assertFalse(factory.Type().getAll().isEmpty());
```
Before I read it like "Type getAll contains something". Now I have to read: "It is not true that type get all is empty". My mind needs three time more effort to understand that ... 
So I would not merge these other PRs.
Thomas I very appreciate your cleaning effort, but in this particular case, I would not agree with these changes.

monperrus:
reopened per @zielint0 's request in #2475

zielint0:
@monperrus 
Mercii!

@pvojtechovsky 
Thank you for this explanation. I will consider reworking it :-)

PR #2472, opened 2018-09-16
refactor: remove redundant initializer

zielint0:
One more redundant initializer fixed.

### DIFF
--- a/src/main/java/spoon/reflect/visitor/printer/internal/CollectionSourceFragment.java
+++ b/src/main/java/spoon/reflect/visitor/printer/internal/CollectionSourceFragment.java
@@ -63,7 +63,7 @@ public class CollectionSourceFragment implements SourceFragment {
 	 */
 	public boolean isOrdered() {
 		CtRole role = null;
-		ContainerKind kind = null;
+		ContainerKind kind;
 		for (SourceFragment childSourceFragment : items) {
 			if (childSourceFragment instanceof ElementSourceFragment) {
 				ElementSourceFragment esf = (ElementSourceFragment) childSourceFragment;
--- a/src/main/java/spoon/reflect/visitor/printer/internal/CollectionSourceFragment.java
+++ b/src/main/java/spoon/reflect/visitor/printer/internal/CollectionSourceFragment.java
@@ -63,13 +63,12 @@ public class CollectionSourceFragment implements SourceFragment {
 	 */
 	public boolean isOrdered() {
 		CtRole role = null;
-		ContainerKind kind;
 		for (SourceFragment childSourceFragment : items) {
 			if (childSourceFragment instanceof ElementSourceFragment) {
 				ElementSourceFragment esf = (ElementSourceFragment) childSourceFragment;
 				if (role == null) {
 					role = esf.getRoleInParent();
-					kind = esf.getContainerKindInParent();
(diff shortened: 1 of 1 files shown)


zielint0:
@monperrus 
We need to analyze this issue in context.
The problem solved here is not explicit vs implicit initialization.

The problem solved here is confusing overriding of initialized value.
example:
a) int x = 2;
b) x = 1;
c) System.out.println(x);

in line b) the initialized value from line a) is lost (overriden). It is confusing for developer.

In C/C++ if we create a variable (in most cases) it has a value that is in memory cell.
In Java need for initialization is greatly reduced. We have:
- classes fields initialized to 0 ("" for string, 0 for int, 0.0 for double, null for objects)
- uninitialized reference error if we allocate object and try to access it without initialization
so redundant initalization often confusing and sometimes dangerous.

Now let's take a look on whole method:
```
public boolean isOrdered() {
		CtRole role = null;
		ContainerKind kind = null;
		ContainerKind kind;
		for (SourceFragment childSourceFragment : items) {
			if (childSourceFragment instanceof ElementSourceFragment) {
				ElementSourceFragment esf = (ElementSourceFragment) childSourceFragment;
 				if (role == null) {
 					role = esf.getRoleInParent();
 					kind = esf.getContainerKindInParent();
 					if (kind != ContainerKind.LIST) {
 						return false;
 					}
 				} else {
 					if (role != esf.getRoleInParent()) {
 						//the collection contains elements of different roles. It cannot be ordered
 						return false;
 					}
 					//else there is another element of the same role - ok
 				}
 			}
 		}
 		//there are only elements of one role of container kind LIST
 		return true;
 	}
 }
```
First *kind* is initalized to null, then null is overridden by *kind = esf.getContainerKindInParent();*
This is problem to be fixed, not redundant initalization per se.

I fixed similar one in https://github.com/INRIA/spoon/pull/2372
Merged by @surli :-)

@monperrus 
@pvojtechovsky 
@surli 
Please consider reopeling this PR. Thanks!

zielint0:
@monperrus 
To make reviewers life easy, I created a list of merged PRs.
If you want to see rationale or fact if something similar was merged or not, please take a look:
https://github.com/INRIA/spoon/issues/2105

monperrus:
reopened per @zielint0 's request in #2475

pvojtechovsky:
I agree with Tomasz, that there is place for improvement in this case. I suggest to make it even more clear and to move declaration of variable to place where it is really needed.
```java
ContainerKind kind = esf.getContainerKindInParent();
```
WDYT?

monperrus:
In this case, kind is assigned if 1) one enters in the loop and 2) one enters in the if

For all the other cases, I would still be in favor of an explicit initialization.

monperrus:
> `ContainerKind kind = esf.getContainerKindInParent();`

That would be good, and even better if kind can be made final.

zielint0:
@pvojtechovsky 
@monperrus 
Thank you! I implemented both suggestions :-)

zielint0:
I am ready to merge.

pvojtechovsky:
It looks like this will be in conflict with #1927 too, because the package was renamed.

zielint0:
Reintroduced in https://github.com/INRIA/spoon/pull/2480

Issue #2483, opened 2018-09-19
bugs: bugs found with static analyzer PVS-Studio Java


Egor18:
Hello!
I found several bugs in Spoon with our PVS-Studio Java static analyzer prototype.

1.
https://github.com/INRIA/spoon/blob/6b07da89021260f2222212ec1cd523ae9f4ec72f/src/main/java/spoon/support/visitor/ClassTypingContext.java#L218-L227 **Warning:** V6008 Null dereference of 'type'. ClassTypingContext.java:225

2.
https://github.com/INRIA/spoon/blob/97f6af6a14ba948050f087e205cc43d9b1516460/src/main/java/spoon/support/compiler/jdt/PositionBuilder.java#L691-L696 **Warning:** V6007 Expression 'content[off] == '\n'' is always false. PositionBuilder.java:694

Probably forgot to increment `off` here.

3.
https://github.com/INRIA/spoon/blob/8f14f3d0af5a21eef32a061224abbc786e7d3484/src/test/java/spoon/support/visitor/java/JavaReflectionTreeBuilderTest.java#L244-L245 **Warning:** V6009 Function 'equals' receives an odd argument. An object 'removeModifiers(elementModifiers, ModifierKind.PUBLIC, ModifierKind.ABSTRACT)' is used as an argument to its own method. JavaReflectionTreeBuilderTest.java:244

https://github.com/INRIA/spoon/blob/8f14f3d0af5a21eef32a061224abbc786e7d3484/src/test/java/spoon/support/visitor/java/JavaReflectionTreeBuilderTest.java#L253-L254 **Warning:** V6009 Function 'equals' receives an odd argument. An object 'removeModifiers(elementModifiers, ModifierKind.PUBLIC)' is used as an argument to its own method. JavaReflectionTreeBuilderTest.java:253

I'm pretty sure that `otherModifiers` (which is unused now) should be used here.

4.
https://github.com/INRIA/spoon/blob/88fe9031bca9d7c9cbb2103cb9e66ab5936760a9/src/main/java/spoon/metamodel/MMMethodKind.java#L114-L117 **Warning:** V6007 Expression 'result.level == k.level' is always false. MMMethodKind.java:115

We've already checked that `result.level < k.level`, how could it be `result.level == k.level`?

5.
[...]

PR #2488, opened 2018-09-20
fix: Unreachable code

Egor18:
Fix unreachable code (#2483 Warning 2)

I'm not quite sure what to do with that one, but there is definitely an error here (unreachable code):
```
if (content[off] == '\r') {
    if (content[off] == '\n') {

```

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/PositionBuilder.java
+++ b/src/main/java/spoon/support/compiler/jdt/PositionBuilder.java
@@ -691,7 +691,7 @@ public class PositionBuilder {
 					if (content[off] == '\r') {
 						//we have found end of this comment
 						//skip windows \n too if any
-						if (content[off] == '\n') {
+						if (off < maxOff && content[off + 1] == '\n') {
 							off++;
 						}
 						return off;


src/main/java/spoon/support/compiler/jdt/PositionBuilder.java
```diff
@@ -691,7 +691,7 @@ private int getEndOfComment(char[] content, int maxOff, int off) {
 					if (content[off] == '\r') {
 						//we have found end of this comment
 						//skip windows \n too if any
-						if (content[off] == '\n') {
+						if (off < maxOff && content[off + 1] == '\n') {
```
pvojtechovsky: Yes, it needs + 1. But also in first part of condition. Like this:
`if (off + 1 < maxOff && content[off + 1] == '\n')`
OK?

Egor18: But why? The maximum value of `maxOff` is `content.length - 1` (see the upper lines), so it seems like a valid bounds check to me.

pvojtechovsky: You are right! So it is OK. Thank you.

PR #2493, opened 2018-09-20
fix: Checks for null

Egor18:
Fix invalid checks for null (dereferenced before checked) (#2483 Warnings 5)

### DIFF
--- a/src/main/java/spoon/pattern/PatternBuilder.java
+++ b/src/main/java/spoon/pattern/PatternBuilder.java
@@ -112,11 +112,11 @@ public class PatternBuilder {
 	}
 
 	protected PatternBuilder(List<CtElement> template) {
-		this.templateTypeRef = getDeclaringTypeRef(template);
-		this.patternModel = Collections.unmodifiableList(new ArrayList<>(template));
 		if (template == null) {
 			throw new SpoonException("Cannot create a Pattern from an null model");
 		}
+		this.templateTypeRef = getDeclaringTypeRef(template);
+		this.patternModel = Collections.unmodifiableList(new ArrayList<>(template));
 		this.valueConvertor = new ValueConvertorImpl();
 		patternNodes = ElementNode.create(this.patternModel, patternElementToSubstRequests);
 		patternQuery = new PatternBuilder.PatternQuery(getFactory().Query(), patternModel);
--- a/src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
+++ b/src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
@@ -307,9 +307,12 @@ public class ReferenceBuilder {
 	 */
 	CtReference getDeclaringReferenceFromImports(char[] expectedName) {
 		CompilationUnitDeclaration cuDeclaration = this.jdtTreeBuilder.getContextBuilder().compilationunitdeclaration;
+		if (cuDeclaration == null) {
+			return null;
+		}
 		LookupEnvironment environment = cuDeclaration.scope.environment;
 
-		if (cuDeclaration != null && cuDeclaration.imports != null) {
+		if (cuDeclaration.imports != null) {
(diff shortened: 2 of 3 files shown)


src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
```diff
@@ -307,9 +307,12 @@ private boolean isTypeArgumentExplicit(TypeReference[][] typeArguments) {
 	 */
 	CtReference getDeclaringReferenceFromImports(char[] expectedName) {
 		CompilationUnitDeclaration cuDeclaration = this.jdtTreeBuilder.getContextBuilder().compilationunitdeclaration;
+		if (cuDeclaration == null) {
```
monperrus: what's the added value of this change?

src/main/java/spoon/support/reflect/declaration/CtEnumImpl.java
```diff
@@ -113,8 +113,12 @@ public boolean removeEnumValue(CtEnumValue<?> enumValue) {
 
 	@Override
 	public <C extends CtEnum<T>> C setEnumValues(List<CtEnumValue<?>> enumValues) {
+		if (enumValues == null) {
```
monperrus: Why does this remove the static analysis warning?

src/main/java/spoon/pattern/PatternBuilder.java
```diff
@@ -112,11 +112,11 @@ public static PatternBuilder create(CtElement... elems) {
 	}
 
 	protected PatternBuilder(List<CtElement> template) {
-		this.templateTypeRef = getDeclaringTypeRef(template);
-		this.patternModel = Collections.unmodifiableList(new ArrayList<>(template));
 		if (template == null) {
 			throw new SpoonException("Cannot create a Pattern from an null model");
 		}
+		this.templateTypeRef = getDeclaringTypeRef(template);
```
monperrus: this is easily mergeable, what about putting it in a separate PR? 

(for changes that look quite different like this one, the discussion and merge is easier in separate PRs, thanks!)

src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
Egor18: The previous check `cuDeclaration != null` makes no sense. It was performed too late:

```
LookupEnvironment environment = cuDeclaration.scope.environment; // <= cuDeclaration dereference (if cuDeclaration is null, we will get NullPointerException)
if (cuDeclaration != null && cuDeclaration.imports != null) { ... } // <= cuDeclaration could not be null here
```

Alternatively, I can remove that check at all, if `this.jdtTreeBuilder.getContextBuilder().compilationunitdeclaration` could not be null.

src/main/java/spoon/support/reflect/declaration/CtEnumImpl.java
Egor18: The same pattern here:
```
.... new ArrayList<>(enumValues) // <= if enumValues is null, we will get NullPointerException
if (enumValues == null || enumValues.isEmpty()) { .... } // <= enumValues could not be null here
```

We detect such errors with data-flow analysis. 
When I add that check for null with unconditional return in 'then' branch, the analyzer understands that in the code, following the check, the `enumValues` reference is not null.

src/main/java/spoon/pattern/PatternBuilder.java
Egor18: Would you like me to make 3 separate PR's for this issue again?

src/main/java/spoon/support/reflect/declaration/CtEnumImpl.java
monperrus: Thanks. I was not aware that `new ArrayList<>(null)` throws an NPE

PR #2494, opened 2018-09-20
fix(MavenLauncherTest): Remove fixed numbers and improves classpath testing

nharrand:
Continuation of #2461


nharrand:
Ready for review.

PR #2499, opened 2018-09-20
refactor: remove unused exceptions

zielint0:


zielint0:
@monperrus 
The last fix is here refactor: remove unused exceptions from tests #2508

PR #2503, opened 2018-09-21
review: fix: compiled snippets have no parent

### LINKED ISSUE #2318: bug? About the usage of CtCodeSnippetStatement
I just found in https://github.com/SpoonLabs/astor/pull/112 that @martinezmatias was using `CtCodeSnippetStatement` in given more that one statement at a time, and it was working in 6.2.0 but apparently no more in 7.0.0. 

Here's an example of the test and its behaviour in Spoon 7.0.0:

```java
String snippet = "int n=0; n += 2";
Factory factory = new FactoryImpl(new DefaultCoreFactory(), new StandardEnvironment());
CtStatement st = factory.Code().createCodeSnippetStatement(snippet).compile();

System.out.println(st.toString()); // n += 2;
CtElement fix = ((CtBlock) st.getParent()).getStatement(1); // indexOutOfBoundsException : there's only one statement
```

So apparently there's a bug here because the second statement indeed have the block as parent, but is not included as a statement of the block. 

Now the question is: should a `CtCodeSnippetStatement` accept to get more that one statement? Shouldn't we create a new element `CtCodeSnippetBlock` to those cases? 

I'm really not a user of those features in Spoon so I need your vision on this one.

monperrus:
fix #2318

### DIFF
--- a/src/test/java/spoon/test/serializable/SerializableTest.java
+++ b/src/test/java/spoon/test/serializable/SerializableTest.java
@@ -25,39 +25,34 @@ public class SerializableTest {
 
 	@Test
 	public void testSerialCtStatement() throws Exception {
+		// contract: one can also serialize snippets
 		Factory factory = new FactoryImpl(new DefaultCoreFactory(), new StandardEnvironment());
 		CtStatement sta2 = (factory).Code()
 				.createCodeSnippetStatement("String hello =\"t1\"; System.out.println(hello)").compile();
 
 		byte[] ser = ByteSerialization.serialize(sta2);
-		CtStatement des = (CtStatement) ByteSerialization.deserialize(ser);
+		CtStatement deserializedSta2 = (CtStatement) ByteSerialization.deserialize(ser);
 
 		String sigBef = sta2.getShortRepresentation();
-		String sigAf = des.getShortRepresentation();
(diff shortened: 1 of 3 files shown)


src/test/java/spoon/test/serializable/SerializableTest.java
```diff
@@ -25,39 +25,34 @@
...
-		assertNotNull(typeDes);
-		//After deserialization, getDeclaringType throws an exception
-		CtType<?> decl =  typeDes.getDeclaringType();
-		assertNull(decl);
-
-		CtPackage parentOriginal = (CtPackage) typeBef.getParent();
-		CtPackage parentDeser = (CtPackage) typeDes.getParent();
-
-		assertEquals(CtPackage.TOP_LEVEL_PACKAGE_NAME,parentOriginal.getSimpleName());
+		CtType<?> typeDes = deserializedSta2.getParent(CtType.class);
```
pvojtechovsky: `assertFalse(deserializedSta2.isParentInitialized());` WDYT?

monperrus: done

pvojtechovsky:
Thank you Martin

PR #2506, opened 2018-09-22
feature: add support for relative path with CtPath#relativePath

monperrus:
add support for relative path with CtPath#relativePath with a functional style


monperrus:
done

PR #2513, opened 2018-09-22
refactor: remove redundant variable

zielint0:


src/main/java/spoon/pattern/PatternParameterConfigurator.java
```diff
@@ -411,59 +411,58 @@ private void configureByTemplateParameter(CtType<?> templateType, Map<String, Ob
 				String fieldName = typeMember.getSimpleName();
 				String stringMarker = (param.value() != null && !param.value().isEmpty()) ? param.value() : fieldName;
 				//for the compatibility reasons with Parameters.getNamesToValues(), use the proxy name as parameter name
-				String parameterName = stringMarker;
```
pvojtechovsky: this new parameter improves readability and expresses my intention - see comment.
Meaning of `stringMarker` and `parameterName` variables are different. Even if they have actually same value. So please rollback this and related changes.

zielint0:
https://github.com/INRIA/spoon/pull/2549

PR #2519, opened 2018-09-22
refactor: simplify ElementSourceFragment.java

zielint0:


src/main/java/spoon/support/sniper/internal/ElementSourceFragment.java
```diff
@@ -485,7 +486,6 @@ public SourcePositionHolder getElement() {
 			if (child instanceof TokenSourceFragment) {
 				result.add(child);
 				i++;
-				continue;
```
pvojtechovsky: I agree that this statement is useless from compiler point of view, but it makes it more readable and safer for refactoring, because it explicitly says the next iteration must start. I suggest to keep `continue` here

zielint0:
Thank you Pavel :-)

PR #2520, opened 2018-09-22
refactor: simplify 'if' condition in ImportScannerImpl.java

zielint0:


src/main/java/spoon/reflect/visitor/ImportScannerImpl.java
```diff
@@ -212,12 +212,7 @@ private boolean matchesTypeName(String javadoc, CtTypeReference<?> typeRef) {
 
 	private boolean isTypeMatching(String typeName, CtTypeReference<?> typeRef) {
 		if (typeName != null) {
-			if (typeName.equals(typeRef.getQualifiedName())) {
-				return true;
-			}
-			if (typeName.equals(typeRef.getSimpleName())) {
-				return true;
-			}
+			return typeName.equals(typeRef.getQualifiedName()) || typeName.equals(typeRef.getSimpleName());
```
pvojtechovsky: @zielint0 Do we really want such change? I wrote more letters in origin code, because it is better readable **for me**.
I can live with your shorter expression, ... it still readbale too **for me** and may be it is better readable **for you**. ...
... do we really want to spent our precious free time on topics like this? ;-)

zielint0: @pvojtechovsky 
I thought it would be better ;-)

PR #2527, opened 2018-09-23
checkstyle: ReplacementVisitor.java

zielint0:
@monperrus 
@surli 
If checkstyle is enforced in spoon/src, why this non-conformant code landed in repository?

### DIFF
--- a/src/main/java/spoon/support/visitor/replace/ReplacementVisitor.java
+++ b/src/main/java/spoon/support/visitor/replace/ReplacementVisitor.java
@@ -1302,10 +1302,10 @@ public class ReplacementVisitor extends spoon.reflect.visitor.CtScanner {
 				if (val != null) {
 					map.put(key, val);
 					val.setParent(shouldBeDeleted.getParent());
-				}else {
+				} else {
 					map.remove(key);
 				}
-			}else {
+			} else {
 				map.remove(key);
 			}
 			listener.set(map);


pvojtechovsky:
because this class is generated automatically by the processor.

... and because our generator doesn't conforms to checkstyle, we released checkstyle rules there. It would be big effort to make generator checkstyle compatible now.

zielint0:
@pvojtechovsky 
Thank you so much for this clarification.

What would be some steps to make it conformant?

pvojtechovsky:
> What would be some steps to make it conformant?

if the only problem is this `}else` separator, then it might be fixable... problem somewhere in DefaultJavaPrettyPrinter.

zielint0:
> if the only problem is this }else separator, then it might be fixable... problem somewhere in DefaultJavaPrettyPrinter.

Thanks! Currently yes, I have not found other checkstyle violations in production code (but there are many violations in checkstyle report and I have not analyzed all yet).

pvojtechovsky:
I suggest to close this one, because this file is generated. It makes sense to open new issue to fix the generator ... the `}else` problem.

monperrus:
> I suggest to close this one, because this file is generated.

I agree, @zielint0 could you close it?

> It makes sense to open new issue to fix the generator ...

Agree. And to fix it. @zielint0 could you see how to change DefaultJavaPrettyPrinter so that the output is compliant with checkstyle? Thanks!

zielint0:
@monperrus 
> Agree. And to fix it. @zielint0 could you see how to change DefaultJavaPrettyPrinter so that the output is compliant with checkstyle? Thanks!

Will take a look but no promise ;-)

PR #2531, opened 2018-09-23
WIP: feature: Pattern detector

pvojtechovsky:
It is reverse operation for the "Matching code by Pattern".
The set of Spoon ASTs can be processed by PatternDetector, which produces "minimal" set of Patterns, which matches processed ASTs.

What it is good for? If client wants to automatically replace some algorithms, then s/he has to 
1) detect which algorithms exist in target area (using PatternDetector)
2) manually check all detected patterns, and optionally merge some of them into more generic pattern
3) for each existing pattern, write a new pattern 
4) match code for existing patterns and for each match generate code using new pattern and replace old code by new code
... and complex refactoring is done in safe way ... at least it is an idea how it might work ;-)

For example I plan to use that to refactor all Spoon model property accessor methods to fix all problems found by #1922.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -135,6 +135,11 @@
       <artifactId>maven-invoker</artifactId>
       <version>3.0.1</version>
     </dependency>
+    <dependency>
+      <groupId>com.github.gumtreediff</groupId>
+      <artifactId>core</artifactId>
+      <version>2.1.0</version>
+    </dependency>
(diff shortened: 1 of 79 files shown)


monperrus:
Hi Pavel, what about closing this PR to clean the PR dashboard?

monperrus:
Too many conflicts, closing

PR #2533, opened 2018-09-23
checkstyle: VisitorTest.java

zielint0:


pvojtechovsky:
@monperrus do we want such changes? Can we enforce checkstyle on all tests? If not does it makes sense to continue this way?

zielint0:
There are some files that cannot be checkstyled.
One contains non ascii characters (checkstyle supports only ascii). Another one is reported - https://github.com/checkstyle/checkstyle/issues/6064 because of checkstyle bug.

I think it is good to make tests readable, effort is very small.

Next month or two everything could be fixed. If we blacklist few files (2-3?), we could enforce checkstyle in tests as well! :-)

zielint0:
Also I enjoy making tests more readable :-)

PR #2536, opened 2018-09-24
can be closed: fix: shadow elements are built using correct factory

pvojtechovsky:
I do not know details now, but it happened that shadow type contains some elements, which points to wrong (temporary) factory instead to `Factory` of client's main `Launcher`.

So I made this PR to create shadow elements directly with normal Factory. It causes that they are consistently added into that Factory and that they are accessible from that Factory - of course. 
May be it is not wanted (see question #2546) ...

This PR experiments with idea of having shadow classes as part of normal Factory (after they are built once)

### DIFF
--- a/src/main/java/spoon/reflect/factory/TypeFactory.java
+++ b/src/main/java/spoon/reflect/factory/TypeFactory.java
@@ -528,19 +528,10 @@ public class TypeFactory extends SubFactory {
 			if (shadowClass == null) {
 				CtType<T> newShadowClass;
 				try {
-					newShadowClass = new JavaReflectionTreeBuilder(createFactory()).scan((Class<T>) cl);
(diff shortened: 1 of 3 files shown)


pvojtechovsky:
This PR now fails on many tests which expects that spoon model contains exactly only the classes which were added as input ... but with this PR spoon model contains all loaded shadow classes too.

Therefore `launcher.getFactory().getModel().getAllTypes()` (and similar code) returns shadow classes too, which is not expected by tests ... and may be also not by Spoon clients.

So I am ok to close this PR, because it is probably not acceptable that shadow types are part of model ... **even if it seems to me like a more consistent approach** ... 

WDYT?

monperrus:
> if it seems to me like a more consistent approach

I agree with you.

Yet, I agree to close: `Tests run: 1570, Failures: 18, Errors: 4`

We may reconsider this in the future.

PR #2537, opened 2018-09-24
feature: CtRole#getSubRole(Object)

pvojtechovsky:
It is often needed to detect type of sub role. So CtRole can do it now.

### DIFF
--- a/src/main/java/spoon/reflect/declaration/CtElement.java
+++ b/src/main/java/spoon/reflect/declaration/CtElement.java
@@ -290,6 +290,8 @@ public interface CtElement extends FactoryAccessor, CtVisitable, Cloneable, CtQu
 
 	/**
 	 * @return the {@link CtRole} of the parent's attribute where this element is used.
+	 * It returns the primary role. For example ((CtMethod) method).getRoleInParent() returns {@link CtRole#TYPE_MEMBER}.
+	 * <br>
 	 * Returns null if parent doesn't contain this element as direct children or if this element has no parent.
 	 */
 	CtRole getRoleInParent();
--- a/src/main/java/spoon/reflect/path/CtRole.java
+++ b/src/main/java/spoon/reflect/path/CtRole.java
@@ -19,6 +19,18 @@ package spoon.reflect.path;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.function.Predicate;
+
+import spoon.SpoonException;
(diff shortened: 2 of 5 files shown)


monperrus:
I agree with the general idea, this will be powerful.

I'm not sure about the API. Some questions to make progress:

* what is `CtElement#getRoleInParent()` expected to do? returning the least specific (eg TYPE_MEMBER) or the most specific role (eg METHOD)? (we should clarify the Javadoc)
* can it happen that there are several sub roles?
* what about naming it `getMostSpecificRole()`?

monperrus:
OK, thanks for the explanation.

>what about naming it getMostSpecificRole()?

>>Is it still needed if we have only one level of sub roles?

Super explicit names are always better :)

>  Could you make a PR with such comment?

It would be great and meaningful to integrate this comment change in this PR.

What about to make getSubRole() static? (`CtRole.TYPE_MEMBER.getSubRole(null);` is not good because the receiver is never used.)

pvojtechovsky:
>> what about naming it getMostSpecificRole()?

>Super explicit names are always better :)

note that there already exists `CtRole#getSubRoles()`, which returns all possible sub roles of main role. The new method name `getSubRole(Object)` matches with existing method ...
WDYT? Will we rename also `getSubRoles` to `getMostSpecificRoles` ? ... and deprecate old name?

monperrus:
Yes, we should have a unified API.

Currently, the API is spread over two classes:

* `CtElement#getRoleInParent()`
* `CtRole#getSubRoles()`
* `CtRole#getSuperRole()`

For the new one, instead of putting it in CtRole, I would propose to put it in CtElement: `CtElement#getRoleInParentSpecific()`.

The we could write `el.getRoleInParentSpecific()` as well as `el.getRoleInParentSpecific().getSuperRole()`

WDYT?

monperrus:
> * `CtRole#getSubRoles(Object)` works on unassigned CtElement too. And note that this method is used before the element is assigned to parent and before it becomes part of it's children.

It seems to me that we at least need a parent. Do you mean that getSubRole works when you have the parent, but the object is not yet in the children?

pvojtechovsky:
> It seems to me that we at least need a parent. 

no, parent is useless for getSubRole. Or I did not understood your idea.

monperrus:
Now, I understand, did a small refactoring (name + type).

OK for you?

pvojtechovsky:
It is OK for me and ready for merge.

PR #2538, opened 2018-09-24
feature: MetamodelProperty#getMethodBySignature(String)

pvojtechovsky:
It helps to detect metadata of spoon model implementation method.


monperrus:
thanks!

PR #2540, opened 2018-09-24
refactor: improve readability by removing static imports from test

zielint0:


pvojtechovsky:
@monperrus do we want such changes? I personally see nearly no benefit.

pvojtechovsky:
I will merge #2541, because static import of enums is probably not good for readability. But in this case I think that each Spoon developer knows ModuleUtils, which are a spoon test util, which is used nearly everywhere. So it makes sense to import it statically. Same like static import ...Assert.*
WDYT?

zielint0:
I would like to suggest to use static imports only for assertXYZ to avoid any source of confusion for any developers, especially for new ones. But even if one is experienced, it is helpful to have clear picture without any doubt. Spoon is getting bigger and more complex everyday (think about supporting new JDKs and new Spoon features like sniper mode) so readability and maintainability should be considered seriously ;-)

monperrus:
I have no strong opinion in this case.

Pavel, I'm happy with both closing or merging, as you prefer.

zielint0:
@pvojtechovsky 
@monperrus 
Thank you for your review. I will analyze it deeper and then decide if I propose it or not.

pvojtechovsky:
I also do not persist on any solution. I can live with both. But if merging causes that Tomasz will suggest another 50 PRs ;-) with similar changes, then I would say that we all "pay a lot for nothing". 

... It costs our time, but it brings no real value ...

I have many ideas which might bring new features to Spoon, but I do not have time for them :-(

zielint0:
@pvojtechovsky 
@monperrus 
Oh no, I can do it in more coarse grained few PRs. Please don't worry about number of PRs! I can fix test from time to time to avoid flood of PRs! ;-)

zielint0:
@pvojtechovsky 
@monperrus 
And a general remark, one has to pay a technical debt to eliminate majority of it, without constant fixes, codebase could became unmaintainable ;-)

PR #2541, opened 2018-09-24
refactor: remove static imports from CtAnnotationImpl.java

zielint0:
Some elements are statically imported (import static X.Y), then full name is used: X.Y.
I removed static imports and added missing classes.
Consistent syntax is used: CtRole.[something].

Previously there was mixed: CtRole.Value and Value (while 'Value' was statically imported).


pvojtechovsky:
@monperrus do we want such changes? I personally see nearly no benefit.

pvojtechovsky:
Thank you Tomasz

PR #2545, opened 2018-09-25
Check that tests pass on master

pvojtechovsky:
Looks like we have a problem ...


pvojtechovsky:
Everything is OK. It was problem in my local installation. Some spoon tests remembers previous state (in file system) and therefore fails even after checkout of master :-)

PR #2550, opened 2018-09-25
fix(test): remove ignored # from the path

pvojtechovsky:
I checked the CtPath string parser and duplicated ## is understood as #. I think that it is mistake to have ## there.
@nharrand WDYT?


nharrand:
Yes indeed! It's a typo.

monperrus:
thanks.

PR #2575, opened 2018-09-27
refactor: fix documentation errors in JarLauncher.java

zielint0:
Additional question:

Should lines
- (117) throw new SpoonException("Jar " + jar.getPath() + "not found.");
- (136) throw new SpoonException("Pom " + srcPom.getPath() + "not found.");

be changed to:

- (117) throw new SpoonException("Jar " + jar.getPath() + **" not found."**);
- (136) throw new SpoonException("Pom " + srcPom.getPath() + **" not found."**);

?

PR #2577, opened 2018-09-29
sonarqube: remove dead condition jarlauncher.java

zielint0:
In Java constructors never return "null" value so SonarQube reported this 'if' condition.
I also corrected string message, confirmed by @nharrand here: https://github.com/INRIA/spoon/pull/2575


pvojtechovsky:
@nharrand what was the intention here?

nharrand:
That's a left over from refactoring. (This constructor use to be a static method returning null in case of error.)

This can indeed be removed.

PR #2580, opened 2018-09-29
sonarqube: make field local in MavenLauncher.java

zielint0:
Fix SonarQube: *Remove the "model" field and declare it as a local variable in the relevant methods.*


pvojtechovsky:
@nharrand is that OK for you? Should not we add getter instead? WDYT?

pvojtechovsky:
> I would rather have a getter, or make it protected. I think it make sense to keep the information from the pom in a MavenLauncher instance.

I think so.
@nharrand, will you make a PR for that?

@zielint0 I suggest to close this one, OK?

nharrand:
Yes sure!

zielint0:
Thank you for clarification guys!

PR #2591, opened 2018-09-30
review: perf: cache result of CtTypeReference#findClass()

pvojtechovsky:
`ClassLoader#loadClass` is slower then reading from Map.

I am using internal `MapUtils` class for PatternDetector ... Can we add `MapUtils` now, or you want special PR for that internal class?

### DIFF
--- a/src/main/java/spoon/support/reflect/reference/CtTypeReferenceImpl.java
+++ b/src/main/java/spoon/support/reflect/reference/CtTypeReferenceImpl.java
@@ -40,13 +40,16 @@ import spoon.reflect.reference.CtTypeReference;
 import spoon.reflect.visitor.CtVisitor;
 import spoon.support.SpoonClassNotFoundException;
 import spoon.support.reflect.declaration.CtElementImpl;
+import spoon.support.util.internal.MapUtils;
 import spoon.support.visitor.ClassTypingContext;
 
 import java.lang.reflect.AnnotatedElement;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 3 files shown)


src/main/java/spoon/support/util/internal/MapUtils.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.function.Supplier;
+
+public abstract class MapUtils {
+
+	private MapUtils() {
+	}
+
+	/**
+	 * @return existing value of `key` from `map`. If value doesn't exist yet for `key` yet,
+	 * than `valueCreator` is used to create new value, which is than assigned to `key` and returned
```
zielint0: than -> then

src/main/java/spoon/support/util/internal/MapUtils.java
```diff
@@ -0,0 +1,51 @@
...
+	 */
+	public static <K, V> V getOrCreate(Map<K, V> map, K key, Supplier<V> valueCreator) {
+		return getOrCreate(map, key, valueCreator, null);
+	}
+	/**
+	 * @param initializer is called immediately after the value is added to the map
+	 * @return existing value of `key` from `map`. If value doesn't exist yet for `key` yet,
+	 * than `valueCreator` is used to create new value, which is than assigned to `key` and returned
```
zielint0: than -> then (2x)

src/main/java/spoon/support/reflect/reference/CtTypeReferenceImpl.java
```diff
@@ -140,23 +143,33 @@ public void accept(CtVisitor visitor) {
 		return findClass();
 	}
 
+	private static Map<String, Class> classByQName = Collections.synchronizedMap(new HashMap<>());
```
monperrus: static field = global state = evil

is that required? is there an alternative?

pvojtechovsky: > alternative

an LRUMap identity map with key = ClassLoader and value  = Map<ClassName, Class>

pvojtechovsky: Both solutions will improve performance and break nothing so I am OK to merge it.

PR #2593, opened 2018-09-30
perf: optimize get of nested and anonymous type declaration

pvojtechovsky:
I started to use `AllMethodsSameSignatureFunction`, which checks all types whether they are sub type of input type. This check often calls `TypeFactory#get(String)`, which is very slow in case of anonymous classes ... and there is many anonymous class in spoon model.
1) So the cache introduced in this PR makes it faster.

2) throwing of exception in `Integer.parseInt(className)` is slower then new call of `isNumber(String)`.


pvojtechovsky:
Note: I am using internal MapUtils class for PatternDetector ... Can I add MapUtils now, or you want special PR for that internal class?

PR #2604, opened 2018-10-01
refactor: unify syntax of asserts in tests (Assert.assert* -> assert*)

zielint0:


monperrus:
I like this. Could you add and configure a checkstyle rule that enforces this? Thanks!

zielint0:
> I like this. Could you add and configure a checkstyle rule that enforces this? Thanks!

I added task for myself :-) https://github.com/INRIA/spoon/issues/2609

PR #2605, opened 2018-10-01
refactor: use contains() to improve readability

zielint0:
- use contains() to clearly say intentions (both commits)
- simplify method *canBeRegExpPattern* (one commit)

### DIFF
--- a/src/main/java/spoon/reflect/path/CtPathStringBuilder.java
+++ b/src/main/java/spoon/reflect/path/CtPathStringBuilder.java
@@ -88,7 +88,7 @@ public class CtPathStringBuilder {
 			String kind = token;
 			CtPathElement pathElement = null;
 			token = tokenizer.getNextToken(PATH_DELIMITERS);
-			if (token != null && token.length() == 1 && PATH_DELIMITERS.indexOf(token) >= 0) {
+			if (token != null && token.length() == 1 && PATH_DELIMITERS.contains(token)) {
 				//nextToken is again path delimiter. It means there is no token value in between
 				throw new CtPathException("Path value is missing");
 			}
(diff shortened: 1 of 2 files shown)


src/main/java/spoon/reflect/path/impl/CtNamedPathElement.java
```diff
@@ -62,12 +62,9 @@ public CtNamedPathElement(String pattern, boolean canBeRegexp) {
 	}
 
 	private boolean canBeRegExpPattern(String str) {
-		if (str.indexOf("()") >= 0) {
-			//this is acceptable by RegExp, but is in conflict with method signature
-			//and it makes no sense for regexp
-			return false;
-		}
-		return true;
+		//this is acceptable by RegExp, but is in conflict with method signature
```
monperrus: here the meaning of the comment is less clear. The previous version is better.

src/main/java/spoon/reflect/path/CtPathStringBuilder.java
```diff
@@ -88,7 +88,7 @@ public CtPath fromString(String pathStr) throws CtPathException {
 			String kind = token;
 			CtPathElement pathElement = null;
 			token = tokenizer.getNextToken(PATH_DELIMITERS);
-			if (token != null && token.length() == 1 && PATH_DELIMITERS.indexOf(token) >= 0) {
+			if (token != null && token.length() == 1 && PATH_DELIMITERS.contains(token)) {
```
monperrus: ok for this one.

src/main/java/spoon/reflect/path/impl/CtNamedPathElement.java
zielint0: How about this one?
```
	private boolean canBeRegExpPattern(String str) {
		//this is acceptable by RegExp, but is in conflict with method signature
		//and it makes no sense for regexp
		return str.indexOf("()") < 0;
	}
```

monperrus:
updated comment and merged.

Issue #2101, opened 2018-06-22
Create QUALITY.md document


zielint0:
Propose
- List of IntelliJ IDEA code inspections (included in both free and commercial editions)
- List of software quality guidelines
- List of suggested books

zielint0:
Hello Martin
I would like to suggest some workflow, best practices and resources helpful to increase quality of spoon.
I enumerated some things that came to my mind first in issue's description above.
I plan to finish this task (first version) this weekend.

monperrus:
Thanks for this effort. Before adding a new file in the repo, I propose that we discuss about this very interesting topic in this thread.

zielint0:
Fixing defects from IDEA code inspection is in progress.
I suggest to remove unused imports.

https://github.com/INRIA/spoon/issues/2105

More suggestions will come in 2105.

surli:
Hi @zielint0 

thanks for your interest in Spoon and software quality.

FYI in order to improve Spoon quality we also have a SonarQube dashboard at this URL: there's plenty of bad smells to fix there! -> https://sonarqube.ow2.org/dashboard?id=fr.inria.gforge.spoon%3Aspoon-core

On a related subject, @ashutosh1598 is working on a repair bot to help fixing those bad smells from Sonar: any help would be appreciated on that subject. You can find more about that here: https://github.com/kth-tcs/sonarqube-repair

zielint0:
Hello Simon
Thank you for this information. It is very valuable. I will check them and try to exploit (or fix bot per se) :-)

ashutosh1598:
Currently the main problem with the bot is that sniper mode of spoon isn't working properly. Sniper mode is available at [refDJPP](https://github.com/pvojtechovsky/spoon/tree/refDJPP). The problem is that source positions in spoon are inconsistent. It can't be used on any code that has a type cast. Any help with https://github.com/INRIA/spoon/issues/2055 will be very beneficial.

monperrus:
Hi Tomasz, could you see if some of your old WIP or failing PRs can be closed? Thanks!

zielint0:
@monperrus 
I closed some of them.
Partially I fix them or reintroduce their changes in new PRs.

I am going to fix them all and close them.
They are not abandoned! ;-)

PR #2606, opened 2018-10-01
refactor: fix javadoc errors

zielint0:


src/main/java/spoon/reflect/path/CtRole.java
```diff
@@ -202,10 +202,10 @@ public CtRole getSuperRole() {
 	/**
 	 * @return sub role of this role, which match `item`.
 	 *
-	 * <code><pre>
```
monperrus: why this change? this is strictly equivalent.

zielint0: The error message from IDEA code inspection is: `Block element not allowed within inline element <code>: pre`

I think it should be like here https://stackoverflow.com/questions/4611591/code-vs-pre-vs-samp-for-inline-and-block-code-snippets

```
Use <code> for inline code that can wrap and <pre><code> for block code that must not wrap. 
```

monperrus:
OK. In the future, please state the IDEA code inspection message in the PR body.

zielint0:
> OK. In the future, please state the IDEA code inspection message in the PR body.

I am goint to prepare list of inspections (ruleset) + rationale as promised in
- https://github.com/INRIA/spoon/issues/2105
- https://github.com/INRIA/spoon/issues/2101

Issue #2609, opened 2018-10-01
Modify checkstyle to enforce unqualified asserts


zielint0:
Instead of *Assert.assertTrue(...)*, we want to enforce: *assertTrue(...)*.

Idea born in https://github.com/INRIA/spoon/pull/2604

monperrus:
cleaning pass, closing stale issues :-)

PR #2611, opened 2018-10-01
review: chore: prepares checkstyle checking for src/test/java

monperrus:
the idea is that we would add rules, one after the other, together with the required changes to make it pass, starting with #2609


monperrus:
@pvojtechovsky ready for merge

pvojtechovsky:
I do not understand travis build process. @surli can you check and merge it?

PR #2615, opened 2018-10-02
WiP: refactor: return types of Spoon model methods

pvojtechovsky:
Alternative of #1852 as fix of #1846
The code of this PR was generated by SniperPrinter. The sources of generator are in this [branch](https://github.com/pvojtechovsky/spoon/tree/refReturnTypeMM).

There are following problems, which has to be solved before we can merge it:
* the order of imports is changed by printer (see # #2638). Note: it is not using sniper mode for imports. @surli do you think we can modify import printer by way it keeps the order from origin sources?
* each file contains extra spaces between imports and class definition. I will try to fix that in Sniper printer.
* ImportScanner still adds 20 unwanted imports.
image
I can remove them manually before commit. Or it needs a fix in import scanner.
* 41 problems caused by change in return types. See this [discussion](https://github.com/INRIA/spoon/pull/1852#issuecomment-426072727) in #1852

### DIFF
--- a/src/main/java/spoon/reflect/code/BinaryOperatorKind.java
+++ b/src/main/java/spoon/reflect/code/BinaryOperatorKind.java
@@ -16,6 +16,9 @@
  */
 package spoon.reflect.code;
 
+
+
+
 /**
  * This enumeration defines all the kinds of binary operators.
  */
--- a/src/main/java/spoon/reflect/code/CtAbstractInvocation.java
+++ b/src/main/java/spoon/reflect/code/CtAbstractInvocation.java
@@ -16,15 +16,18 @@
  */
 package spoon.reflect.code;
 
-import spoon.reflect.declaration.CtElement;
-import spoon.reflect.reference.CtExecutableReference;
-import spoon.reflect.annotations.PropertyGetter;
-import spoon.reflect.annotations.PropertySetter;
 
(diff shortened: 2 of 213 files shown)


monperrus:
I love this first massive refactoring with the sniper mode. Eat your own dog food!

monperrus:
> 41 problems caused by change in return types

can you post them so as to inform our decision on this?

pvojtechovsky:
> can you post them so as to inform our decision on this?

here you are CSV with error descriptions: [errors.csv](https://github.com/INRIA/spoon/files/2457373/errors.txt)

pvojtechovsky:
> I think I use HashSet everywhere for the imports, so we should change that to use LinkedHashSet and it should works almost immediately.

It is true that HashSet is used, but then the imports are somehow "sorted", so any previous order is ignored anyway.

I think only sniper mode should try to keep origin order of imports. The normal mode is just responsible for providing of correct set of imports, which can be then printed in somehow defined order.

Actually the problem is that some imports are missing ... E.g. in CtElement the static CtRole.COMMENT. I made #2621 to fix it, but I run into bigger problems, so I need your feedback in #2624

monperrus:
Pass to close staled PRs. Feel free to reopen if appropriate.

PR #2619, opened 2018-10-03
fix: CtImportScanner shouldn't import type when static field is imported

pvojtechovsky:
When printer prints the code like
```java
import static spoon.reflect.path.CtRole.NAME;
...
String field = NAME.toString();
```
then it shouldn't add `import spoon.reflect.path.CtRole;`


monperrus:
Thanks Pavel!

PR #2627, opened 2018-10-04
refactor: make sure stream is closed

zielint0:


monperrus:
what about using a try-with-resource?

zielint0:
> what about using a try-with-resource?

I tried but it broke tests. I recommend to use this simpler solution now and consider potential improvements in future.

pvojtechovsky:
> I tried but it broke tests

So there is some deeper problem in original implementation. 
It is really not good idea to create a BufferedInputStream on the stream which is not under control of this function. Because BufferedInputStream may read much more data then needed by this function.
The correct solution needs a change in context of that function

PR #2630, opened 2018-10-05
bug: Invalid enum members modifiers

Egor18:
Hi. I found out that Spoon is probably wrong about the 'final' and 'public' modifiers of enum members.
Consider the following code:
```
public enum MyEnum {
    ONE,
    TWO,
    THREE;

    private static int len = -1; // Spoon says that len is final and public, but it is not!

    public static void f() {
        len = 44;
    }
}
```
BTW, If I do the same in a class it works fine:
```
public class MyClass {

    private static int len = -1; // len is not final and is not public (as expected)

    public static void f() {
        len = 44;
    }
}
```

This PR contains failing test cases (see `lenField.isFinal()` and `lenField.isPublic()`).
Any ideas how to fix this?
Thanks.

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilder.java
+++ b/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilder.java
@@ -1140,7 +1140,9 @@ public class JDTTreeBuilder extends ASTVisitor {
 		}
 		field.setSimpleName(CharOperation.charToString(fieldDeclaration.name));
 		if (fieldDeclaration.binding != null) {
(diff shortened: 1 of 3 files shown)


tdurieux:
It seems to be related to this, I will double check:
https://github.com/INRIA/spoon/blob/c5cc06e55a199c2e8fd50a800cf1a14148fd8351/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilder.java#L1143-L1149

monperrus:
LGTM. Ready for merge?

Egor18:
Yes, I hope this fix helps.

pvojtechovsky:
Thank You Egor ;-)

PR #2634, opened 2018-10-06
fix: use correct classloader during shadow type building

pvojtechovsky:
JavaReflectionTreeBuilder used factory with default environment instead of current environment, therefore it used wrong class loader - it was not possible to define the class loaded of JavaReflectionTreeBuilder


monperrus:
thanks Pavel

Issue #2482, opened 2018-09-19
bug: bug in comment association and pretty-printing


tempuser1909:
Hi, I am not sure if this is a bug for printing the source code with the comments.

Original:
```
public final class test{

    
// ------------------------------------ //
// test / test test     //
// ------------------------------------ //

    /**
     * test2
     *
     */   
    public interface testInterface {
        /**
         * test3
         * 
         */
        public void mytest(short a, short b);
    }
    private static class myInterface implements testInterface {
        public void mytest(short a, short b){
            
        }
    }
    
}//end class test

```
Pretty-printed by Spoon Launcher:
```


public final class test {
    // ------------------------------------ //
    public interface testInterface {
        // test / test test     //
        // ------------------------------------ //
        /**
         * test2
         */
        /**
         * test3
         */
        public void mytest(short a, short b);
    }

    private static class myInterface implements test.testInterface {
        public void mytest(short a, short b) {
        }
    }
}


```

So the problem is:
1. the test comment is printed inside of the interface class instead of being outside
2. the comment that marks the end of the class is gone

tempuser1909:
I checked via the GUI mode too.

It seems that the "end class" comment is detected but not printed.

And the two comments that were printed wrongly inside the interface class were detected wrongly as part of the method that is within the interface class.

![image](https://user-images.githubusercontent.com/16776064/45742250-229b8300-bc2c-11e8-9cf5-aaebb4968f8b.png)

pvojtechovsky:
> Do you mean to include a unit test for this with a pull request?

yes

> Is it under src/test/java/spoon/test/comment/CommentTest.java

yes 👍

tempuser1909:
Sorry for the late pull request. This is my first time writing a unit test. 

Please let me know if writing a more specific unit test is required. I just compared the original file with the prettyprint-ed file.

monperrus:
closed by #2636 

Thanks a lot @tempuser1909 for the bug report and the failing test case!

rihou:
The comments in the lambda expression are reported to have errors.

An example code:

import java.util.Arrays;
import java.util.List;
import java.util.stream.Collectors;

public class CommentInLambdaExpressionTest {

	public static void main(String[] args) {
		CommentInLambdaExpressionTest test = new CommentInLambdaExpressionTest();
		List<Integer> lst = test.setup();
		System.out.println(lst);
	}

    public List<Integer> setup() {
[...]

PR #2636, opened 2018-10-06
added a test for Comment Association and Printing

tempuser1909:
For Issue #2482. Added a test class for the following issues:
1. Comment association are mixed up
2. And the comment at the end of the file is not printed

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/JDTCommentBuilder.java
+++ b/src/main/java/spoon/support/compiler/jdt/JDTCommentBuilder.java
@@ -39,6 +39,7 @@ import spoon.reflect.code.CtSwitch;
 import spoon.reflect.cu.CompilationUnit;
 import spoon.reflect.cu.SourcePosition;
 import spoon.reflect.cu.position.BodyHolderSourcePosition;
+import spoon.reflect.cu.position.DeclarationSourcePosition;
 import spoon.reflect.declaration.CtAnonymousExecutable;
 import spoon.reflect.declaration.CtClass;
 import spoon.reflect.declaration.CtConstructor;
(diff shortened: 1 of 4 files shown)


tempuser1909:
Hi @pvojtechovsky, just wanted to confirm my understanding of comment association.

So for the comment associated with class `TestClassWithComments`, it is
```
//end class test
```

For comments associated with interface `testInterface`, they are
```
// ------------------------------------ //
// test / test test     //
// ------------------------------------ //
/**
   * test2
   */ 
```

And finally, the comment associated with method `mytest` under `testInterface`, it is
```
/**
   * test3
   */ 
```

Am I right? Also, I would like to ask why is
```
// ------------------------------------ //
// test / test test     //
// ------------------------------------ //
```
considered to be to associated with `testInterface` instead of being comments associated as part of class `TestClassWithComments`?

pvojtechovsky:
>  just wanted to confirm my understanding of comment association.

you are right

> why is `// test / test test     //` considered to be to associated with testInterface instead of being comments associated as part of class TestClassWithComments?

Because in java there is usually true that comment belongs to the element which follows after the comment. But note... there are some rare exceptions from this rule.

tdurieux:
Sorry @pvojtechovsky, I was not available this weekend.

Your fix seems good, ready for merge?

tdurieux:
> considered to be to associated with testInterface instead of being comments associated as part of class TestClassWithComments?

It is really difficult to find a rule that will work in all the cases. You should probably take into account the number of new lines before and after the comments but I don't know what is the limit.

pvojtechovsky:
>  ready for merge?

yes, I just added test contract comment. So then it is ready for merge

monperrus:
Thanks a lot!

PR #2640, opened 2018-10-08
refactor: remove redundant cast - CtRolePathElement.java

zielint0:
Remove redundant cast from *CtElement* to *CtElement*.

Checker: Redundant type cast
*Reports unnecessary cast expressions.*

This is also SonarQube fix - Remove this unnecessary cast to "CtIf".


monperrus:
@zielint0  could you add a checkstyle rule so that redundant casts now fail the build? Thanks.

zielint0:
> @zielint0 could you add a checkstyle rule so that redundant casts now fail the build? Thanks.

I briefly checked a list of possible Checkstyle violations and it looks that is it beyond Checkstyle.

I found it using IDEA Code Inspection (*Redundant type cast*).

PR #2641, opened 2018-10-08
refactor: constructor Metamodel(Factory) is protected

pvojtechovsky:
Now one can create Metemodel using own Factory and then refactor the Spoon itself - this is needed in #2615


monperrus:
I love your baby PRs :)

PR #2645, opened 2018-10-09
feat:Spoon agent

nharrand:
With the help of `IncrementalLauncher` and `Decompiler`, we can propose an abstraction of java standard `ClassFileTransformer` interface that let an user define Spoon transformations applied on bytecode at load time.

Every time a class is loaded:
 * if it matches a filter
 * it is decompiled
 * its decompiled sources are added to the Spoon Model
 * users transformation are applied on the new CtType
 * the type is recompiled
 * loading resumres


monperrus:
This is a pretty neat and unconventional usage of Spoon, I like it :-)

zielint0:
I think it may have an enormous potential of practical uses!

nharrand:
Currently this only includes a `ClassFileTransformer` abstraction, which allows user to implements agent. But may be we should add something similar to `Launcher` with a real premain that actually load `Processor`s and apply them. A `RuntimeLauncher` if you will.

WDYT?

src/main/java/spoon/decompiler/SpoonClassFileTransformer.java
```diff
@@ -0,0 +1,169 @@
...
+public abstract class SpoonClassFileTransformer implements ClassFileTransformer {
+
+	Predicate<String> classNameFilter;
+	String pathToDecompiled;
+	String pathToRecompile;
+	String pathToCache;
+	File cache;
+	Set<String> classPath;
+	public Decompiler decompiler;
+	Set<File> inputSources;
```
pvojtechovsky: I guess all the members above should be private or at least protected. WDYT?

pvojtechovsky:
Nice idea!

I have few comments to design. Actually client has to `extend` `SpoonClassFileTransformer` and must implement two methods. 
Problems:
* client has to implement accept even if s/he doesn't need filter.
* if client extends wrong methods then it will not work

May be it would be nicer if
* `SpoonClassFileTransformer` would have method `setTransformer(ClassTransformer)` ... or `ClassTransfomer` might be a new constructor parameter.
* `ClassTransformer` would be interface with 2 methods: `transform(CtType)` and `boolean accept(CtType)`
* `boolean accept(CtType)` would be marker `default` with body `return true;`
* optionally there might be method `SpoonClassFileTransformer#setFilter(Filter<? super CtType>)`, which would allow to define extra type filter too.

Then all the implementation can be done by lambda expressions

src/main/java/spoon/decompiler/SpoonClassFileTransformer.java
nharrand: Yes definitly.

nharrand:
@pvojtechovsky Seems like good ideas. I'll try to implement them.

monperrus:
Thanks a lot for this potentially game-changing PR.

PR #2647, opened 2018-10-09
contributing: explain the different kinds of pull-requests

monperrus:
We need to better explain the different kinds of pull-requests in the project. Here is a first draft. WDYT?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,8 +31,12 @@ Guidelines for all pull-requests:
 * The pull request does a single thing (eg a single bug fix or a single feature). 
 * The pull request must pass all continuous integration checks (incl. formatting rules).
 * The pull request must have an explicit and clear explanation.
-* The pull-request title starts with a prefix stating its kind: "fix:", "feature:", "refactor:", "perf:", "checkstyle:"
-
+* Pull-request title:
+  * The pull-request title starts with a prefix stating its kind: "fix:", "feature:", "refactor:", "perf:", "checkstyle:"
+  * Pull-requests that are in progress are prefixed by "WIP".
+  * Pull-requests that are ready for review are prefixed by "review", or labeled as "[review](https://github.com/INRIA/spoon/labels/review)".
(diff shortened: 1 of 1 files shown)


zielint0:
> There are different kinds of pull-requests, in particular bug-fix, new features, refactoring and performance pull-requests.

Refactoring of source code is changing its non-functional characteristics (make it faster, make it more readable, make it conform to a new rule, ...) WHILE preserving functional characteristics (i.e. a feature is not changed, e.g. if we refactor feature *save a file*, we do NOT change functionality, we (for example) improve performance of saving the file.

Having said that, perhaps 3 types of PRs are enough: bug-fix, new feature, refactoring. Of course for practical reasons, it is possible to enumerate special type of refactoring - performance PR. But it is really needed?

Just wondering ;-)

CONTRIBUTING.md
```diff
@@ -24,26 +24,29 @@ Current integrators (alphabetical order):
 Guidelines for pull requests
 ----------------------------
```
pvojtechovsky: Problem: when I started with Spoon contribution a year ago, I read these rules before I made my first PR to Spoon. But I understood it wrong - I though that PR must fulfill all these rules since the beginning, therefore I had a fear to make the first PR here ;-)

Solution: I suggest to add here something like: 

> This guideline is mandatory for PRs before they are merged to master. If you have anything interesting for Spoon community, then we welcome your PR even if it is not ideal since beginning. Spoon integrators will help You to fix remaining problems - if any. Your contribution is highly welcome! ;-)

pvojtechovsky:
@monperrus, it looks good to me. I just suggest to add comment, which will lower the wall between spoon and new contributors. (See comment above)

> Of course for practical reasons, it is possible to enumerate special type of refactoring - performance PR. But is it really needed?

I think it makes sense for clients, when they see in the Spoon release change log, the we improved a performance somewhere. Yes it is refactoring of special type.

pvojtechovsky:
LGTM. Is it ready for merge? Just CI had some problem ...

monperrus:
ready for merge. restarted CI job (failed because of slow MainTest).

PR #2649, opened 2018-10-10
remove deprecated methods after release 

monperrus:
@pvojtechovsky could you look how to remove the deprecated methods in `PatternParameterConfigurator`? Thanks!


monperrus:
ping @pvojtechovsky ?

monperrus:
> the only way how remove these methods is to remove support of old Templates.

Then, I support to remove the `@Deprecated` and instead point to the better option in the Javadoc. Could you do that?

(and then we have no `@Deprecated` anymore, which is the goal)

PR #2652, opened 2018-10-11
refactor: remove useless condition (always true)

zielint0:
Field *processors* is never null. Condition is always true. Should be removed.
```
public abstract class AbstractAssert<T extends AbstractAssert<T, A>, A> {
	protected final LinkedList<Processor<?>> processors = new LinkedList<>();
```

**Constant conditions & exceptions inspection**
*This inspection analyzes method control and data flow to report possible conditions that are always true or false, expressions whose value is statically proven to be constant, and situations that can lead to nullability contract violations.
Variables, method parameters and return values marked as @Nullable or @NotNull are treated as nullable (or not-null, respectively) and used during the analysis to check nullability contracts, e.g. report NullPointerException (NPE) errors that might be produced.
More complex contracts can be defined using @Contract annotation, for example:
@Contract("_, null -> null") — method returns null if its second argument is null @Contract("_, null -> null; _, !null -> !null") — method returns null if its second argument is null and not-null otherwise @Contract("true -> fail") — a typical assertFalse method which throws an exception if true is passed to it 
The inspection can be configured to use custom @Nullable
@NotNull annotations (by default the ones from annotations.jar will be used)*

### DIFF
--- a/src/main/java/spoon/testing/AbstractCtPackageAssert.java
+++ b/src/main/java/spoon/testing/AbstractCtPackageAssert.java
@@ -47,7 +47,7 @@ public abstract class AbstractCtPackageAssert<T extends AbstractCtPackageAssert<
 			throw new AssertionError(String.format("The actual package named %1$s isn't equals to the expected package named %2$s", actual.getSimpleName(), expected.getSimpleName()));
 		}
 
-		if (processors != null && !processors.isEmpty()) {
+		if (!processors.isEmpty()) {
 			process(actual.getFactory(), processors);
 		}
 


zielint0:
Somebody merge me! :-)

zielint0:
> Somebody merge me! :-)

I provided a fragment of code to explain it in PR. It is:
```
protected final LinkedList<Processor<?>> processors = new LinkedList<>();
```

LinkedList ctr (like any other ctr) does not return null so the conditional is not needed.

monperrus:
Assume that in the future we remove the "final" field. In that case, the NPE which is currently avoided would crash Spoon. WDYT?

zielint0:
https://www.innoq.com/de/blog/do-we-worship-complexity/

pvojtechovsky:
> The proper way to check for NPE is automatized static analysis. I will try to implement it.
that would be really great to have NPE solved/detected this way.

But until we have a tool which can assure that there is no code which can produce NPE we should keep redundant null checks there, because it is safer to have them... as long as we are still changing the spoon code.

zielint0:
> But until we have a tool which can assure that there is no code which can produce NPE we should keep redundant null checks there

But it would result in a huge number of not needed code. Do you think it is really worth to check everything for NPE? Even if we know that some code is safe?

As I said, currently there are many places that can produce NPE (misssing checks for null). Shouldn't they be fixed first?

I will provide them later.

pvojtechovsky:
> But it would result in a huge number of not needed code. 
I do not suggest to change anything. I suggest to keep it in the state as it is not now until solution based on static analysis is available

zielint0:
> I suggest to keep it in the state as it is not now until solution based on static analysis is available

Wait, I forget about SonarQube. It catches NPEs. (However the list of inspections could be improved a bit, but this is another issue).

What I meant is that I would like to add additional static analysis server.

monperrus:
In software engineering, there are many decisions that are not binary.

In this case, on the one hand, we have a redundant check (negative side), on the other hand we have an insurance for future NPEs, even if the code changes (positive side).

Both @pvojtechovsky and myself agree that the positive side surpasses the negative one, we can close this PR.

Thanks a lot @zielint0 for the proposal and the opportunity to discuss one of those core software engineering problems.

PR #2654, opened 2018-10-11
refactor: remove redundant transient keyword (SonarQube fix)

zielint0:
Fix SonarQube warning: *Remove the "transient" modifier from this field.*

*transient is used to mark fields in a Serializable class which will not be written out to file (or stream). In a class that does not implement Serializable, this modifier is simply wasted keystrokes, and should be removed.*

### DIFF
--- a/src/main/java/spoon/reflect/factory/CompilationUnitFactory.java
+++ b/src/main/java/spoon/reflect/factory/CompilationUnitFactory.java
@@ -41,7 +41,7 @@ public class CompilationUnitFactory extends SubFactory {
 		super(factory);
 	}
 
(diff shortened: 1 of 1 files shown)


zielint0:
Should something be changed or close this PR?

pvojtechovsky:
Looks like You found a not nice hack. A spoon factory has an unwanted state ... 

1) This cache of compilation units should be moved to CtModel.
2) The access to the cache content should be restricted - now everybody can modify content of that cache.

@monperrus, @surli  WDYT?

monperrus:
I agree.

zielint0:
I opened an issue *A spoon factory has an unwanted state #2703*.

PR #2665, opened 2018-10-14
refactor: add deprecated annotation (sonarqube)

zielint0:


monperrus:
the solution is rather to remove the @Deprecated, I've changed it here.

PR #2667, opened 2018-10-15
fix(MaintTest): quickfix of ci timeout issues

nharrand:
This should fix timout issues on travis. It is only a quick fix that print and flush() a message after each package testing for the one test that takes a long time.

The offending test is `MainTest#testElementToPathToElementEquivalency()` and we might want to rework it. _(Keeping in mind that we should not blame whoever wrote it... :innocent:)_

I suggest that we select a meaningful subset of the sources on which it runs. (So far it runs on the complete spoon AST).

### DIFF
--- a/src/test/java/spoon/test/main/MainTest.java
+++ b/src/test/java/spoon/test/main/MainTest.java
@@ -533,31 +533,40 @@ public class MainTest {
 
 	@Test
 	public void testElementToPathToElementEquivalency() {
+		System.out.println("Start testElementToPathToElementEquivalency...");
+		System.out.flush();
+		for (CtPackage ctPackage: rootPackage.getPackage("spoon").getPackages()) {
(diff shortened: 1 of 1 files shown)


monperrus:
Thanks a lot.

LGTM. Will merge it.

> I suggested that we select a meaningful subset of the sources on which it runs. (So far it runs on the complete spoon AST).

Yes! Let's do this in a separate PR. Could you open an issue to keep track of it?

nharrand:
Done!

monperrus:
Thanks!

monperrus:
seems like he quickfix is not enough: https://travis-ci.org/INRIA/spoon/jobs/441787940

nharrand:
That's really weird because the output I just added is indeed printed... Was it the wrong print stream? Well anyway I'll try to reduce the duration of the test overall.

PR #2668, opened 2018-10-15
feat(FernflowerDecompiler)

nharrand:
As I am unsure how reliable is `CFRDecompiler`, and that it might make `SpoonClassFileTransformer` a little tricky to use, I suggest to add at least a second option. [Fernflower](https://github.com/JetBrains/intellij-community/tree/master/plugins/java-decompiler/engine) is IntelliJ's decompiler, which is open source (Apache v2) but not distributed as a standalone project. But JBoss is distributing it as a Maven dependency.

This change only increase slightly the jar-with-dependencies (16M -> 17M).

### DIFF
--- a/doc/launcher.md
+++ b/doc/launcher.md
@@ -113,6 +113,11 @@ JarLauncher launcher = new JarLauncher("<path_to_jar>", "<path_to_output_src_dir
 });
 ```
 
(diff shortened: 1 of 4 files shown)


monperrus:
explain in the doc how to use it?

monperrus:
Thanks!

PR #2657, opened 2018-10-11
checkstyle: AccessTest.java

zielint0:

PR #2670, opened 2018-10-15
refactor: remove statically imported enums

zielint0:


monperrus:
This PR is related to the comment in #2657. 

We cannot afford spending valuable review time on aspects that could be automated.

We will merge the static import pull-requests once the policy is enforced in the CI.

Thanks.

PR #2674, opened 2018-10-16
refactor: flip order of comparison - put literal first to lower risk of NPE

zielint0:
Code inspection
- 'expression.equals("literal")' rather than 'literal.equals("expression")'

*Reports calls to .equals() whose arguments are String literals. Some coding standards specify that String literals should be the target of .equals(), rather than argument, thus minimizing NullPointerExceptions.*


monperrus:
thanks!

PR #2675, opened 2018-10-16
refactor: remove redundant type (extending Object)

zielint0:
Code inspection
- Type parameter explicitely extends 'java.lang.Object' inspection

Every Java class extends Object class OR another class.
Extending Object does not provide any value (it is like invoking "super()" in ctr) and makes code longer.


monperrus:
LGTM. @pvojtechovsky do you merge?

Issue #2669, opened 2018-10-15
perf(MainTest): improve performances of MainTest##testElementToPathToElementEquivalency()


nharrand:
I propose we reduce the set of sources on which this test operates (while trying to keep the actual coverage) to improve performances. (This test alone is responsible of a significant part of the total testing time.)

Related to #2667

monperrus:
cleaning pass, don't hesitate to comment or reopen.

PR #2681, opened 2018-10-18
perf(MainTest): Minify MainTest#testElementToPathToElementEquivalency keeping branch and statement coverage

### LINKED ISSUE #2662: MainTest is too slow for Travis
MainTest sometimes takes more than 10 minutes without any output and the build fails because of this, see https://travis-ci.org/INRIA/spoon/jobs/440920311

```
[INFO] Running spoon.test.main.MainTest
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated
```

nharrand:
Hopefully, this should preserve test coverage while improving performances.
Fixes #2662 #2669


pvojtechovsky:
Hi Nicolas,

previous version of test checked that we can create path which uniquely identifies element. Then we searched for that element using that path. 

Now the first part of the test is missing. I would say it is not good idea.

Second problem is that we have to maintain list of paths now, so future maintenance effort is higher.

What about using multiple CPU cores to check some paths parallel?

nharrand:
I admit this is not a perfect solution.
> previous version of test checked that we can create path which uniquely identifies element. Then we searched for that element using that path.

We used to check than element to path to element lead to equivalency, I transformed it (in this PR) to Path to element to Path equivalency, doesn't it also assess path to element uniqueness? Am I missing something out?

> Second problem is that we have to maintain list of paths now, so future maintenance effort is higher.

On that matter I agree. Plus it only provides non regression testing not really anything more.

>What about using multiple CPU cores to check some paths parallel?

Sure, I can try that too. We still can decide once we see how much perf it make us gain.

nharrand:
Although, to be more clear about where does this list of path comes from: it is actually generated by tracing test execution of the previous version of the test (on the whole spoon project model) and keeping only elements that actually leads to the ececution of a path that is not covered by previous elements.
So about:
>Second problem is that we have to maintain list of paths now, so future maintenance effort is higher.

It could actually be regenerated regularly. (I can also add the code for that.)

nharrand:
So #2694 is helpful when disposing of many cores, but is not really helpful with travis.

Build for this solution:
```
[INFO] Running spoon.test.main.MainTest
[INFO] Tests run: 15, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 123.778 s - in spoon.test.main.MainTest
```

[Build from alternative solution](https://travis-ci.org/INRIA/spoon/jobs/443445277)
```
[INFO] Running spoon.test.main.MainTest
[INFO] Tests run: 15, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 430.326 s - in spoon.test.main.MainTest
```

surli:
> It could actually be regenerated regularly. (I can also add the code for that.)

you should! :)

Actually you could even provide a new test that would recompute that list and check it against the given list: if it changes, that would mean the test is not up-to-date and we need to provide the new list. Almost what we do for the generated parts of Spoon.

On the overall it means we won't gain any time (because the time gain in this test, will be take again to compute the list and to the check) but it will be done in another test: so shouldn't be a problem for the time limitation in Travis I think. WDYT?

monperrus:
I like this proposal.

@pvojtechovsky I agree with you, the list of paths is not maintainable manually, **but** it is meant to be generated.

@surli the generation itself is complex and slow, so it cannot be in main CI, but in a separate CI task, such as the other "level-2" CI tasks on our Jenkins (see #2708)

WDYT?

monperrus:
I like this proposal.

@pvojtechovsky I agree with you, the list of paths is not maintainable manually, **but** it is meant to be generated.

@surli the generation itself is complex and slow, so it cannot be in main CI, but in a separate CI task, such as the other "level-2" CI tasks on our Jenkins (see #2708)

WDYT?

monperrus:
I like this proposal.

@pvojtechovsky I agree with you, the list of paths is not maintainable manually, **but** it is meant to be generated.

@surli the generation itself is complex and slow, so it cannot be in main CI, but in a separate CI task, such as the other "level-2" CI tasks on our Jenkins (see #2708)

WDYT?

monperrus:
I like this proposal.

@pvojtechovsky I agree with you, the list of paths is not maintainable manually, **but** it is meant to be generated.

@surli the generation itself is complex and slow, so it cannot be in main CI, but in a separate CI task, such as the other "level-2" CI tasks on our Jenkins (see #2708)

WDYT?

monperrus:
I like this proposal.

@pvojtechovsky I agree with you, the list of paths is not maintainable manually, **but** it is meant to be generated.

@surli the generation itself is complex and slow, so it cannot be in main CI, but in a separate CI task, such as the other "level-2" CI tasks on our Jenkins (see #2708)

WDYT?

monperrus:
I like this proposal.

@pvojtechovsky I agree with you, the list of paths is not maintainable manually, **but** it is meant to be generated.

@surli the generation itself is complex and slow, so it cannot be in main CI, but in a separate CI task, such as the other "level-2" CI tasks on our Jenkins (see #2708)

WDYT?

monperrus:
I like this proposal.

@pvojtechovsky I agree with you, the list of paths is not maintainable manually, **but** it is meant to be generated.

@surli the generation itself is complex and slow, so it cannot be in main CI, but in a separate CI task, such as the other "level-2" CI tasks on our Jenkins (see #2708)

WDYT?

nharrand:
I agree with all 6 comments from @monperrus !

monperrus:
> I agree these paths look weird, I'll look into that.

Good. Then we can proceed with merge.

nharrand:
I'll come back to this with a fresh attempt I think (Based on more recent change on contract versifiers.). The current one is not completely satisfying.

Issue #2624, opened 2018-10-04
Discussion: Concept of printing imports and qualified names


pvojtechovsky:
I am trying to fix tests in #2621, which marked a CtTypeAccess of enum field as implicit, when such field is statically imported.

Now I would need a concept/specification of how the spoon printer should behave in
A) autoImport = false -> MinimalImportScanner
B) autoImport = true -> ImportScannerImpl 
C) snipermode

Example of source code:
```java
import static spoon.reflect.path.CtRole.NAME;
import spoon.reflect.path.CtRole;
...
CtRole r1 = NAME;              //CtTypeAccess.isImplicit == true
CtRole r2 = CtRole.NAME;  //CtTypeAccess.isImplicit == false
CtRole r3 = spoon.reflect.path.CtRole.NAME;
```

In my understanding the `autoImport = false` means that **all what can use fully qualified name will use it.** So it should print:
```java
// no import here - if possible
...
CtRole r1 = spoon.reflect.path.CtRole.NAME;
CtRole r2 = spoon.reflect.path.CtRole.NAME;
CtRole r3 = spoon.reflect.path.CtRole.NAME;
```
It means isImplicit is ignored and always understood as false.

Then the `autoImport = true` means that 
B1) we try to print the sources **in the way how they are modeled**. It means we print or not print qualified name depending on implicit attribute
B2) we try to automatically **import as much as possible**. We import (including static import?) as much as possible

And the `sniper mode` means we try **to print the sources in the way how they are modeled**. It means we print or not print qualified name depending on implicit attribute. It expects that model (implicit=true/false) reflects origin sources.

monperrus:
Initially, autoimports meant B1.

Now that we have the sniper mode, which is basically B1, we can more autoimport to B2 and make autoimport independent of implicit values. This will likely simplify the autoimport code quite a lot.

pvojtechovsky:
> Now that we have the sniper mode, which is basically B1, we can more autoimport to B2

It will not work, because Sniper mode uses internally standard printing (auto import) and just wraps it to minimize changes. **So the standard printing should try to produce the origin code as much as possible**, then also Sniper mode will print source more matching to origin code.

So I suggest to Use B1 in standard auto import mode too.

> This will likely simplify the autoimport code quite a lot.

[...]

PR #2683, opened 2018-10-18
refactor: handling of imports in pretty printer

pvojtechovsky:
This PR implements #2624


pvojtechovsky:
Hallelujah! It passed all tests :-)

So, I started to split this PR to smaller parts. Please see, review and merge other PRs, which comes from this one.

pvojtechovsky:
@monperrus I need your help to fix test `ImportTest#testImportOfAnInnerClassInASuperClassPackageAutoImport`. It fails on implementation of #3060. I do not know that concept, so I do not know what is correct solution of this case.

pvojtechovsky:
There is also bug #3087 which causes failing test here. So it blocks progress here.

pvojtechovsky:
note: in case of type `spoon.test.imports.testclasses.internal.SuperClass.InnerClassProtected`
whose `SuperClass` is invisible and therefore the import statement must be access path like: `spoon.test.imports.testclasses.internal.ChildClass.InnerClassProtected`
there is no way in the spoon model how to declare which part of such computed access path is implicit.

pvojtechovsky:
May be we can add that isParentImplicit check, which was part of your origin solution ... anyway I would like to analyze failing tests, because I think that there might be something interesting ...

pvojtechovsky:
by the way, you can push to my branch. I will not force push there

tdurieux:
I pushed a change that makes the test pass but I am not 100% sure if it is correct

pvojtechovsky:
That fix looks good to me.

tdurieux:
Should we not scan the getAccessType()?

pvojtechovsky:
Why? If you think about performance, then it might be enough to check whether declaring type is visible in current compilation unit. That code might be extracted from getAccessType method.

tdurieux:
I meant that may be to compute the imports we need to scan the content of `getAccessType()`

pvojtechovsky:
Now I understand your point. 
I am not sure, but there is no way how to define,which part of access path is implicit... Therefore I guess it is always explicit and never imported. So we can probably ignore them when computing imports .
But I can be wrong.

pvojtechovsky:
Thanks to @tdurieux now the tests pass, so it is time to discuss the implementation and the concepts which are in it. Try to read #2624 . Let me know what is not clear. I can write some notes about concepts, which are there if needed for understanding and future maintenance. I will probably have no time to maintain Spoon in future, so remaining maintainers should understand what is here in this PR.

monperrus:
Thanks a lot Pavel and Thomas. I will have a deep look into it, for which I need to allocate real time.

monperrus:
I'm on it

src/main/java/spoon/reflect/declaration/CtElement.java
```diff
@@ -391,4 +392,18 @@
 	 */
 	List<CtElement> getDirectChildren();
 
+	/**
+	 * @return the source code of this element, with fully-qualified types.
+	 *
```
pvojtechovsky: @monperrus, I suggest to add there note like: "Call of toString() method will not modify the spoon model"
Note: It is new and critical contract, which was not true until this PR. But as developer I always expected that toString modifies nothing ... and then I wondered why debugging and looking into model, changes model... :-)

src/main/java/spoon/reflect/declaration/CtElement.java
```diff
@@ -391,4 +392,18 @@
...
+	 * {@link Environment#getToStringMode()} is not taken into account
+	 *
+	 * Use {@link #prettyprint()} for a nice source code.
+	 */
+	String toString();
+
+	/**
+	 * @return the source code of this element accorfing to {@link Environment#getToStringMode()}.
+	 */
```
pvojtechovsky: @monperrus , I suggest to add there note like: "Call of prettyprint() method can modify spoon model! It will apply model import validators, which will try to detect correct import statements, detect conflicts, change implicitness of types and packages, etc."

pvojtechovsky:
@monperrus, I see you gave it a lot of time. Thank you for having look!
I will have a look at your changes and add comments here, so we can discuss and reach common understanding.

C1) You removed CtElement#print(boolean), whose purpose was to print Spoon model in form in which it really actually is - without any further modifications of model. Are you sure Spoon should not offer that possibility?

C2) I like new `prettyprint()` method, which is kind of higher level API from my point of view and which is needed too.

src/main/java/spoon/compiler/Environment.java
```diff
@@ -43,6 +43,10 @@
 	 */
 	void setComplianceLevel(int level);
 
+	TO_STRING_MODE getToStringMode();
```
pvojtechovsky: I suggest to rename `toString mode` to `pretty printing mode`, because CtElement#toString will always use VANILA model independent on that setting. 
In other words: setToStringMode actually doesn't influence behavior of CtElement#toString, but it influence how pretty printing works.
WDYT?

monperrus:
Hi Pavel,

Yes, we're almost ready to merge. You did a fantastic job.

> C1) You removed CtElement#print(boolean), whose purpose was to print Spoon model in form in which it really actually is - without any further modifications of model. Are you sure Spoon should not offer that possibility?

This is the goal of `toString`, toString is a debug method. However, you're right, it still hides some implicit elements such as `this.`. We can change this easily so as to show all implicit elements but: 1) we would have to change tons of assertions 2) we would break backward compatibility and break many tests in client code. So, I prefer to stay as is.

> C2) I like new prettyprint() method, which is kind of higher level API from my point of view and which is needed too.

thanks :-) it's still experimental though (we sometimes create a fake compilation unit on the fly)

src/main/java/spoon/reflect/declaration/CtElement.java
```diff
@@ -391,4 +393,18 @@
 	 */
 	List<CtElement> getDirectChildren();
 
+	/**
+	 * @return a debug version of the source code of this element, with fully-qualified types. Warning, some implicit elements are not shown in the string.
```
pvojtechovsky: > Warning, some implicit elements are not shown in the string.

this warning is probably not true, because DJPP is run in ignoreImplicit=true mode. So everything should be fully qualified when toString is called.
Or do you know a case when implicitness is not ignored?

pvojtechovsky:
> it still hides some implicit elements such as this.

I agree with you, we should keep it hidden. So no change is needed.

src/main/java/spoon/reflect/declaration/CtElement.java
pvojtechovsky: ok, the `this.` qualifier is an example of implicit element. So the warning is correct for me.

pvojtechovsky:
The status of this PR after commit https://github.com/INRIA/spoon/pull/2683/commits/2dd73cd9f8d2515eb66c36b08e85b6bee9834b86
is OK for me. I have no more comments.

monperrus:
Let's cool it down a little bit. @nharrand I guess you would be the final one to merge.

monperrus:
Big kudos to Pavel @pvojtechovsky 
Spoon loves you :heart::heart:!

src/main/java/spoon/reflect/visitor/ImportAnalyzer.java
```diff
@@ -0,0 +1,236 @@
...
+						 *
+						 * ... but in other normal cases, I guess the declaring type is used and needed for import!
+						 * ... so I don't understand why SKIP_ALL works in all cases. May be there is missing test case?
+						 */
+						if (!((CtTypeReference) parent).getAccessType().equals(element)) {
+							return ScanningMode.SKIP_ALL;
+						}
+					} else {
+						//May be this can never happen
+						throw new SpoonException("Check this case. Is it relvant or not?");
```
seintur: Typo: relvant -> relevant

monperrus:
As expected we break most client code:

* https://ci.inria.fr/sos/job/npefix/954/display/redirect
* https://ci.inria.fr/sos/job/nopol/595/display/redirect
* https://github.com/STAMP-project/AssertFixer/issues/42
* https://ci.inria.fr/sos/job/gumtree-spoon-ast-diff/

monperrus:
It seems that the reason is writing fully-qualified `java.lang.*`. 

To solve this, I propose to never write `java.lang.*` in `toString()`.

WDYT?

src/main/java/spoon/reflect/visitor/ImportAnalyzer.java
monperrus: thanks @seintur but it's already  merged, would you edit the file directly on Github?

nharrand:
So, you would like to always write import for classes part of the standard library? (like everything prefixed with `java.lang` or `java.util` ?) Or just `java.lang`?

monperrus:
I'm talking about `toString` which does not involve imports. I want that toString never writes
`java.lang`.

nharrand:
My bad for import. Still why just `java.lang` and not all classes of the standard library?

monperrus:
to be backward-compatible and get the client builds passing again.

nharrand:
Ok my bad, after some clarification, `java.lang` is the only package imported by default, so it makes sense to simplify its fully qualified name all the time.
SGTM.

seintur:
And broke also https://ci.inria.fr/sos/job/juliac/1056/

Don't know for sure this is only about java.lang.* stuff. Still investigating the issue...

monperrus:
Started to analyze. We have three kinds of regressions:

* R1 before toString was dependent on autoimport mode
* R2 some changes in autoimport mode
* R3 some changes in full-qualified autoimport mode

I'm on it.

PR #2684, opened 2018-10-18
refactor: simplify logic in CtNamedPathElement.java (SonarQube fix)

zielint0:
SonarQube fix: *Remove this conditional structure or edit its code blocks so that they're not all the same.*

*Having all branches in a switch or if chain with the same implementation is an error. Either a copy-paste error was made and something different should be executed, or there shouldn't be a switch/if chain at all. Note that this rule does not apply to if chains without else-s, or to switch-es without default clauses.*


zielint0:
@monperrus 
Is it useful? I feel that `scanCtElement` should be analyzed and perhaps refactored.

zielint0:
Thank you for feedback!

PR #2688, opened 2018-10-18
doc: fix javadoc in TypeFactory.java

zielint0:
SonarQube: *Remove this unnecessary cast to "CtElement".*

### DIFF
--- a/src/main/java/spoon/reflect/factory/TypeFactory.java
+++ b/src/main/java/spoon/reflect/factory/TypeFactory.java
@@ -678,7 +678,7 @@ public class TypeFactory extends SubFactory {
 			}
 		}
 		Visitor visitor = new Visitor();
-		formalTypeDeclarer.accept(visitor);
+		((CtElement) formalTypeDeclarer).accept(visitor);
 		return visitor.adapter;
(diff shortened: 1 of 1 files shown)


monperrus:
We have two different PRs here (typo and cast). They should be separated in two.

src/main/java/spoon/reflect/factory/TypeFactory.java
```diff
@@ -282,7 +282,7 @@ public TypeFactory() {
 	}
 
 	/**
-	 * Creates a reference to a one-dimension array of given type.
```
monperrus: "a one" is correct, see https://www.quora.com/Which-is-correct-a-one-or-an-one

zielint0:
> We have two different PRs here (typo and cast). They should be separated in two.

Cast is in https://github.com/INRIA/spoon/pull/2688

src/main/java/spoon/reflect/factory/TypeFactory.java
zielint0: @monperrus 
That's right. Thank you for your careful review!

PR #2693, opened 2018-10-18
refactor: add loggers to SpoonClassFileTransformer.java (fix SonarQube)

zielint0:
SonarQube: *Use a logger to log this exception.*

@pvojtechovsky 
Please tell me if I used logging correctly (according to Spoon rules and according to Java best practices).

### DIFF
--- a/src/main/java/spoon/decompiler/SpoonClassFileTransformer.java
+++ b/src/main/java/spoon/decompiler/SpoonClassFileTransformer.java
@@ -16,7 +16,6 @@
  */
 package spoon.decompiler;
 
-import org.apache.log4j.Logger;
 import org.benf.cfr.reader.Main;
 import spoon.IncrementalLauncher;
 import spoon.SpoonModelBuilder;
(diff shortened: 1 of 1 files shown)


nharrand:
Hello @zielint0 

Maybe `launcher.getEnvironment().debugMessage` calls could actually be transformed into Logger calls too actually...

monperrus:
We don't use one logger per class in Spoon. 

But you're right `e.printStackTrace()` is bad practice.

Could you update this PR to use @nharrand's solution?

zielint0:
> We don't use one logger per class in Spoon.

Please explain motivation.

> Could you update this PR to use @nharrand's solution?

Take a look. Maybe this is what you meant.

src/main/java/spoon/decompiler/SpoonClassFileTransformer.java
```diff
@@ -136,6 +133,7 @@ public SpoonClassFileTransformer(Predicate<String> classNameFilter,
 			ProtectionDomain protectionDomain,
 			byte[] classfileBuffer
 	) throws IllegalClassFormatException {
+		IncrementalLauncher launcher = new IncrementalLauncher(inputSources, classPath, cache);
```
nharrand: I think it is better to build the model only after filtering classes the user does not want to transform. (And after decompilation has been tried with succes, althought this last one is less important).

zielint0: If it is not a low hanging fruit, I leave it as if for now.

Issue #2701, opened 2018-10-20
feat: add principled support for models validators


surli:
I start this issue as a general topic to discuss the concept of model *validators*. We proposed this concept while discussing with @pvojtechovsky in #2624 about moving imports, but I already had something like that in mind when proposing #1598 a while ago.

The general idea is to allow the execution on validations on a Spoon model. Those validations can be executed only to report some information, but also to change the model in order to fix it.
Right now, the first idea of such validator would be to properly manage the imports in a model.

Still several questions have to be clarified from my point of view:
  1. What's the difference between a "validator" and a "processor"? At first I'd say the difference is only on the step when a validator is executed but I'm not so sure anymore, see why in 2.
  2. When the validators needs to be triggered? At first, we only discussed about executing the validator before pretty-printing, but maybe it needs to also be executed before executing the processors, after executing the processors, before pretty-printing...? 
  3. If a validator aims at fixing a model, then it can transform it, so how do we check the validations of our validators? :) 

So, here it is, just few questions to start the discussion.

pvojtechovsky:
> 1. What's the difference between a "validator" and a "processor"?

It would be probably easier if we would merge these terms and to have only processor.
Problems:
* After CompilationUnit is CtElement (will be soon), then Processor might be called on CtCompilationUnit
* The CtCompilationUnit is actually not accessible from `factory.Package().getRootPackage()`, so compilation unit processor have to be called different - or compilation units must be somehow accessible from the root package
* The processor is always called on all nodes of model, while pretty printer validator needs to be called on to be printed compilation unit(s) and all it's types

> 2. When the validators needs to be triggered?

printer is actually processor too (well ... depending on Launcher configuratiom :-( ...), but it's processor queue is created individually in `JDTBasedSpoonCompiler#generateProcessedSourceFilesUsingTypes`.

May be we should have processors for everything and client will simply define list of processors, which has to be executed.... e.g. ["buildModel", "modifyModel", "validateModel", "printModel", "compileModel"]
And we should also support more flexible alternative solution where client can call processors one by one ... as his/her algorithm needs.... but that needs more coding on client's side.

> 3. If a validator aims at fixing a model, then it can transform it, so how do we check the validations of our validators?

I do not understand this question.

monperrus:
closed per #2683

PR #2702, opened 2018-10-20
feat: CtCompilationUnit extends CtElement

pvojtechovsky:
The refactoring of compilation unit based on discussions in #2254 and #2701


surli:
Why exactly this new element contains references for types? The type were belonging to CompilationUnit because in the reality the types are declared in their CU. Not sure to understand the move, especially if you keep an inheritance between CompilationUnit and CtCompilationUnit. (Not sure either why to keep the old CompilationUnit class, though except maybe for backward compatibility)

pvojtechovsky:
>  an inheritance between CompilationUnit and CtCompilationUnit. 

it is here only for backward compatibility. I will gladly remove old Compilation unit if @surli and @monperrus  agrees that.



> Why exactly CtCompilationUnit element contains references for types? The type were belonging to CompilationUnit because in the reality the types are declared in their CU.

The `CtType`s belong to
A) CtPackage
B) CtCompilationUnit

But spoon model is a Tree by definition, so it is not allowed for `CtType` to have two parents.

First I started to implement CtCompilationUnit by way it contained CtType directly - saying it is OK that parent/child relationship is not correct in that case. But there were several problems
1) model consistency the model checking tests failed - OK they can be modified
2) removing of type from CtCompilationUnit did not worked
3) `CtScanner.visitCompilationUnit(cu)` visited CtType and all it's children twice. 1st because it contains declared CtPackage which contains CtType, and second because CtCompilationUnit contains declared types too.
4) cloning of CtCompilationUnit cloned whole subtree of CtPackage (including CtType) and whole sub tree of CtType. I as a client would expect that only CtCompilation unit is cloned and reused the existint CtType, but instead it cloned each CtType twice.

So then I found that it is much easier and correct too, when CtCompilationUnit has reference to CtPackage and reference to CtType. All the problems above are solved by that change.
And there is still method `CtCompilationUnit.getDeclaredTypes`, which is derived now and looks for CtTypes using internal references.

As a next step we can even add a collection of CtCompilationUnits somewhere into CtModel or into CtPackage and then CtCompilationUnit becomes normal member of spoon model. But that is again possible only if CtCompilationUnit contains references to types and package.

WDYT?

surli:
So trying to understand the new hierarchy of Spoon concepts with your change. You'll have something like:

```
CtModel
|
 -- CtRootPackage
   |
    -- CtPackage
       |
        -- CtCompilationUnit (package declaration)
       |
        -- CtType
          |
           -- CtCompilationUnit (type declaration)
```

am I right? Isn't it a bit odd that it's the contrary of the reality: the compilation units actually precedes the declaration of package/types.

pvojtechovsky:
This PR doesn't puts CtCompilationUnit into any parent. It is actually without parent.
But in future we should put CtCompilationUnit somewhere, so it makes sense to discuss it.

> The compilation units actually precedes the declaration of package/types.

This is philosophical discussion ... and might be long. I do not want to attend ;-)

Do you have an alternative? Where CtCompilationUnit belongs to from your point of view? ... and I suggest to keep Spoon the model a "tree" do not switch it to something else ... it would cause many deep changes.

pvojtechovsky:
This PR doesn't puts CtCompilationUnit into any parent. It is actually without parent.
But in future we should put CtCompilationUnit somewhere, so it makes sense to discuss it.

> The compilation units actually precedes the declaration of package/types.

This is philosophical discussion ... and might be long. I do not want to attend ;-)

Do you have an alternative? Where CtCompilationUnit belongs to from your point of view? ... and I suggest to keep Spoon the model a "tree" do not switch it to something else ... it would cause many deep changes.

surli:
The way I see it, the package contains a list of CtCompilationUnit, which themselves contains the CtType. The package only contains CtTypeReference.

Now if a CtPackage is declared in a compilation unit (package-info.java) then its parent is a CtCompilationUnit.
So we have something like
```
CtModel
|
--- CtRootPackage
 |
  --- CtCompilationUnit
    |
     --- CtPackage
       |
        --- CtCompilationUnit
          |
           --- CtType
       |
        --- CtCompilationUnit
          |
           --- CtType
       |
        --- CtPackage
           |
            --- CtCompilationUnit
               |
                --- CtType
               |
                --- CtType
```

pvojtechovsky:
We speak about two models:
* java runtime model - it doesn't contain CompilationUnit at all. Package contains Types. That model is original model of spoon and is good for analyzing and refactoring
* java file system model - it maps types to CompilationUnits, which represents files in file system. That model is needed when spoon tries to print sources back into file system.

The `java runtime model` is primary target of Spoon library. So current Spoon model design Package has Types is correct from that point of view.
The `java file system model` is needed too. I understand it is **mapping** of primary (runtime) model to file system. 

The `java file system model` is kind of extension of primary model. Therefore usage of type/package references to primary model makes sense.

surli:
I'm really mixed on this one.
I don't really agree with you when you say: 
> The java runtime model is primary target of Spoon library. 

I agree that the current model does not support directly the filesystem concepts (even though they can be retrieved using the API). 
Now I think we both agree we cannot maintain two different models in Spoon, we want only one. If we modify the existing I would really prefer something closer to the **source** code, as it's the primary target of Spoon, so a model which would contain the concepts of my Java files.

The rationale behind that is:
  - if we break something, let's break it in something we would want to keep and maintain on the long term view
  - keep the model close to the concepts that are meaningful for Java devs, not something that's easy for us to develop

pvojtechovsky:
@surli thank you for that open and constructive discussion. It is really helpful 👍 

> Now I think we both agree we cannot maintain two different models in Spoon, we want only one. 

I agree.

> if we break something, let's break it in something we would want to keep and maintain on the long term view

I agree with this sentence. But this sentence is not fitting to this PR, because this PR actually breaks nothing. All tests passed without need for bigger change.

The idea to have CtType contained in CtCompilationUnit is breaking. I do not think it is helpful and needed.

> keep the model close to the concepts that are meaningful for Java devs, not something that's easy for us to develop

I agree with this sentence too :-) 

But I think that CtCompilationUnit is secondary think, which should not bother Spoon clients who don't need it. Therefore I would prefer to have CtCompilationUnit somehow outside of primary spoon model.

Why I think it is not needed?
1) I am using java for about 17 years. And I never needed to think about compilation unit.
2) I am actively using Spoon for more then 1 year. And I had to learn what is compilation unit and how it works only when I was fixing internal spoon algorithms:
* the fixing of compilation process - here I touched compilation unit first time in my live
* the introducing CtCompilationUnit into model (this PR) to be able to finish sniper printer.
3) Now I understand what compilation unit is and I still think that I will not need it for next years of work with java and Spoon. 
4) I can imagine only one use case where compilation unit is relevant for spoon client - when s/he writes refactoring which moves types from one file to another. I guess it will be less then 1% of spoon clients. Actually 0%.

I only need to know that about CompilationUntis: Spoon should keep by default the types in origin files (compilation units). I really do not want to touch them all the time I am processing the spoon model.

So the compilation units are secondary stuff which should be hidden somewhere in background ... but of course well designed and well accessible when somebody really needs it.

**May be there is some misunderstanding** ... in one of previous comments you draw this picture - as your understanding of what I probably plan to do:
```
CtModel
|
 -- CtRootPackage
   |
    -- CtPackage
       |
        -- CtCompilationUnit (package declaration)
       |
        -- CtType
          |
           -- CtCompilationUnit (type declaration)
```
but it is wrong. It doesn't fit to this PR or concepts I would like to have in Spoon.

I suggest one of these models

**M1) compilation units in packages**, because packages represents file system directories
```
CtModel
|
 -- CtRootPackage
   |
    -- CtPackage P1
       |
        -- CtCompilationUnit (package declaration which refers P1)
       |
        -- CtCompilationUnit (refers T1)
       |
        -- CtType T1
```
... or one of following models, because compilation units are not fitting to java runtime model at all.

**M2) compilation units in root of model** (e.g. Set<CtCompilationUnit>). It would be very similar to what we already have in Spoon 7.1
```
CtModel
|
 -- CtRootPackage
|  |
|   -- CtPackage P1
|      |
|       -- CtType T1
 -- CtCompilationUnit (package declaration which refers P1)
|
 -- CtCompilationUnit (refers T1)
```

**M3) compilation units in special model**. Just an little alternative of M2.
```
CtModel
|
 -- CtRootPackage
|  |
|   -- CtPackage P1
|      |
|       -- CtType T1
CtCompilationUnitsModel
|
 -- CtCompilationUnit (package declaration which refers P1)
|
 -- CtCompilationUnit (refers T1)
```
**M4) compilation units in under new CtDirectory element**. That would be biggest change. And there would be duplicate information - rename of package would need change of package name and change of CtDirectory name. :-(
```
CtModel
|
 -- CtRootPackage
|  |
|   -- CtPackage P1
|      |
|       -- CtType T1
 -- CtRootDirectory
  |
  -- CtDirectory "p1"
    |
     -- CtCompilationUnit (CtPackageDeclaration which has package reference to P1)
    |
      -- CtCompilationUnit (with type reference to T1)
```

Before I started to write this comment I liked the most M1, but now I am sure that I do not want to touch  compilation units in future of my daily work so I prefer M2.

WDYT?

pvojtechovsky:
> So for M2 be careful that the CtModel could have several modules, so maybe the CtCompilationUnit are not attached to the CtModel directly, but to their CtModule. WDYT?

Thanks for pointing to that. I will thing about that later. I suggest to add CtCompilationUnit as child of CtModel in different PR. This PR is already big enough.... and tests actually pass, so it is good time to review and merge it ... after other related and already made PRs are merged.

pvojtechovsky:
> Why CompilationUnit extends CtCompilationUnit? Why CompilationUnit is not removed?

1) backward compatibility
2) there is 146 other references to CompilationUnit in spoon sources and tests. So it would make review of this PR much harder

src/test/java/spoon/test/compilationunit/TestCompilationUnit.java
```diff
@@ -90,11 +97,11 @@ public void testGetUnitTypeWorksWithCreatedObjects() {
 		CtPackage myPackage = launcher.getFactory().Package().getOrCreate("my.package");
 		CompilationUnit cu = launcher.getFactory().createCompilationUnit();
 		assertEquals(CompilationUnit.UNIT_TYPE.UNKNOWN, cu.getUnitType());
-
-		cu.setDeclaredPackage(myPackage);
```
monperrus: could we stay backward compatible here and keep setDeclaredPackage?

src/test/java/spoon/test/compilationunit/TestCompilationUnit.java
```diff
@@ -148,7 +155,7 @@ public void testAddDeclaredTypeInCU() throws IOException {
 		assertEquals(3, cu.getDeclaredTypes().size());
 
 		CtType typeBla = launcher.getFactory().Class().create("spoon.test.model.Bla");
-		cu.addDeclaredType(typeBla);
```
monperrus: could we stay backward compatible here and keep addDeclaredType?

Conceptually, we add types not references.

monperrus:
In general, it looks good.

There are very few non-backward compatible changes in the tests, and it seems we may even reduce them further. 

To be sure I understand well: what the final goal of this change?

src/test/java/spoon/test/compilationunit/TestCompilationUnit.java
pvojtechovsky: yes, this can be derived method which will internally create a package reference and package declaration. I will do so

pvojtechovsky:
> final goal of this change?

To have CtCompilationUnit, which extends CtElement and is CtVisitable. Then we can apply Processor on CtCompilationUnit. That processor will compute imports and manipulate isImplicit attribute to prepare for pretty printing.

src/main/java/spoon/reflect/cu/CompilationUnit.java
```diff
@@ -128,6 +33,7 @@
 	 * 		an arbitrary index in the source code
 	 * @return the index where the line starts
 	 */
+	@Deprecated
```
surli: Could you specify on the deprecated method what should be used instead with a deprecated annotation in the javadoc.

src/main/java/spoon/reflect/declaration/CtCompilationUnit.java
```diff
@@ -0,0 +1,221 @@
...
+import spoon.reflect.reference.CtTypeReference;
+import spoon.support.DerivedProperty;
+import spoon.support.Experimental;
+import spoon.support.UnsettableProperty;
+
+/**
+ * Defines a compilation unit. In Java, a compilation unit can contain only one
+ * public type declaration and other secondary types declarations (not public).
+ */
+public interface CtCompilationUnit extends CtElement {
```
surli: I would have put an `@Experimental` annotation: it still can change.

src/main/java/spoon/reflect/factory/CompilationUnitFactory.java
```diff
@@ -75,8 +76,9 @@ public CompilationUnit getOrCreate(CtPackage ctPackage) {
 				String path = file.getCanonicalPath();
 				CompilationUnit result = this.getOrCreate(path);
 				result.setDeclaredPackage(ctPackage);
-				ctPackage.setPosition(this.factory.createPartialSourcePosition(result));
-
+				if (ctPackage.getPosition() == SourcePosition.NOPOSITION) {
```
surli: Why those new checks?

src/main/java/spoon/reflect/factory/Factory.java
```diff
@@ -1008,9 +1008,4 @@
 	 * @see PackageFactory#createPackageDeclaration(CtPackageReference)
 	 */
 	CtPackageDeclaration createPackageDeclaration(CtPackageReference packageRef);
-
-	/**
-	 * @see PackageFactory#createPackageDeclaration(String)
-	 */
-	CtPackageDeclaration createPackageDeclaration(String name);
```
surli: Isn't it linked to #2707 ?

src/main/java/spoon/reflect/visitor/CtBiScannerDefault.java
```diff
@@ -983,6 +983,20 @@ public void visitCtUsedService(spoon.reflect.declaration.CtUsedService usedServi
...
 	}
 
+	// autogenerated by CtBiScannerGenerator
+	@java.lang.Override
+	public void visitCtCompilationUnit(spoon.reflect.declaration.CtCompilationUnit compilationUnit) {
+		spoon.reflect.declaration.CtCompilationUnit other = ((spoon.reflect.declaration.CtCompilationUnit) (this.stack.peek()));
+		enter(compilationUnit);
+		biScan(spoon.reflect.path.CtRole.COMMENT, compilationUnit.getComments(), other.getComments());
+		biScan(spoon.reflect.path.CtRole.ANNOTATION, compilationUnit.getAnnotations(), other.getAnnotations());
+		biScan(spoon.reflect.path.CtRole.PACKAGE_DECLARATION, compilationUnit.getPackageDeclaration(), other.getPackageDeclaration());
```
surli: Why the package is not also a ref? When we discussed about the model architecture, the package got a CtRootPackage as ancestor, not a CtCompilationUnit.

src/main/java/spoon/support/compiler/jdt/JDTCommentBuilder.java
```diff
@@ -558,7 +558,7 @@ public void scan(CtElement element) {
 
 		if (!spoonUnit.getDeclaredTypes().isEmpty()) {
 			findCommentParentScanner.scan(spoonUnit.getDeclaredTypes());
-		} else if (spoonUnit.getDeclaredModule() != null) {
+		} else if (spoonUnit.getDeclaredModuleReference() != null) {
 			findCommentParentScanner.scan(spoonUnit.getDeclaredModule());
```
surli: shouldn't it be changed to `spoonUnit.getDeclaredModuleReference().getDeclaration()`?

src/main/java/spoon/reflect/visitor/CtBiScannerDefault.java
pvojtechovsky: It is ref. Because CtCompilationUnits refers CtPackageDeclaration (new node type), which then has CtPackageReference.

src/main/java/spoon/reflect/cu/CompilationUnit.java
pvojtechovsky: there is no better alternative for that unused (useless?) method. But may be it is safer to keep it there - without deprecated. WDYT?

src/main/java/spoon/reflect/declaration/CtCompilationUnit.java
pvojtechovsky: ok

src/main/java/spoon/support/compiler/jdt/JDTCommentBuilder.java
surli: So you could actually keep `getDeclaratedModule()` both in the test `!= null` and after. I'd prefer a consistent solution: `declaredModule` or `declaredModuleReference` but not both.

src/main/java/spoon/reflect/visitor/CtBiScannerDefault.java
surli: OK this is then used only if the CtCompilationUnit contains java types.
What if the CtCompilationUnit is a package-info or a module-info? Shouldn't we use inherited compilation units to manage those now?

pvojtechovsky: > OK this is then used only if the CtCompilationUnit contains java types.

I think it is used for package-info too. But I have not checked it deeper. It should behave same like before this PR.

pvojtechovsky:
All the review comments should be solved. Just unused (actually Deprecated) methods in CompilationUnit needs, which belongs to some utils class are still deprecated.

src/main/java/spoon/reflect/visitor/CtBiScannerDefault.java
surli: OK maybe we should clarify that in a future PR.

PR #2705, opened 2018-10-21
fix: MetamodelProperty#containerKindOf knows Collection now

pvojtechovsky:
Little change needed by #2702

There is no way how to test it. The "test" comes with #2702 after spoon model starts use `Collection` in some API methods.

### DIFF
--- a/src/main/java/spoon/metamodel/MetamodelProperty.java
+++ b/src/main/java/spoon/metamodel/MetamodelProperty.java
@@ -22,6 +22,7 @@ import static spoon.metamodel.Metamodel.getOrCreate;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


src/main/java/spoon/metamodel/MetamodelProperty.java
```diff
@@ -600,7 +601,7 @@ private	static ContainerKind containerKindOf(Class<?> valueClass) {
 		if (Map.class.isAssignableFrom(valueClass)) {
 			return ContainerKind.MAP;
 		}
-		if (Set.class.isAssignableFrom(valueClass)) {
+		if (Set.class.isAssignableFrom(valueClass) || Collection.class.isAssignableFrom(valueClass)) {
```
monperrus: Set is a subtype of Collection so only the second part is necessary
```suggestion
		if (Collection.class.isAssignableFrom(valueClass)) {
```

PR #2706, opened 2018-10-21
fix: relax intercession test to fit #2615 - new fluent API return types

pvojtechovsky:
Relax contract of IntercessionTest#testSettersAreAllGood so the new Spoon API (e.g. #2702) can be made in correct way.

PR #2707, opened 2018-10-21
feat: CtPackageDeclaration

pvojtechovsky:
This is new Spoon model element, which represents package declaration - the line at top of compilation unit. Example:
```java
package your.nice.package.name;
...
```
It is needed to complete #2702 ... and at the sniper mode of compilation unit header.

Actually it is not used in CompilationUnit. I plan to add it to compilation unit in PR #2702, which changes CompilationUnit a lot so it makes no sense to add it now.


pvojtechovsky:
needs merge of #2706 first to pass tests

pvojtechovsky:
I rebased on master. Now the test should pass. It is ready for review.

monperrus:
LGTM. We need the corresponding tests to specify this new metamodel concept.

pvojtechovsky:
I made this PR as a minimal baby step, which adds new metamodel element, which is needed by already big enough #2702

monperrus:
Can we create the node in JDTTreeBuilder in this PR? Then we can test the presence of the node and
the pretty-printing.

pvojtechovsky:
It makes this PR more complicated ... (less baby)
* I have to modify CompilationUnit and to add there code, which will be again removed in #2707 ... so it is wasting of effort.
* I have to learn and modify JDTTreeBuilder (so I would need more time for that) - I would like to have it as the last step - first we need a model and then we can fill it in and use it.

... therefore less compatible with #2707, which is the main purpose of this PR.

PR #2716, opened 2018-10-24
review: Generic contracts are run only on **.testclasses.**

monperrus:
and the rest will move to Jenkins, see plan in #2662 

WDYT?


monperrus:
Good news, by working on the tests, we find bugs!

The `**.testclasses.**` choice was excellent to find corner cases that did not even exist in `src/main/java`. See all filtered classes in `MainTest`.

Those cases are discarded for now, we'll look at them once we're done with this refactoring that blocks #2172. 

(Still we fixed a shadow-related bug with a one-liner)

monperrus:
merge?

pvojtechovsky:
Thank you Martin. It is interesting idea to be able to run these contracts on any model 👍

PR #2726, opened 2018-10-29
fix: PatternMatcher ignores isImplicit

pvojtechovsky:
The `isImplicit` attribute is ignored by pattern matcher now. It means that these two lines:
```java
java.lang.String x;
String x;
```
are same from pattern matcher point of view.

It will be tested by #2683 which correctly sets isImplicit attributes and therefore two lines above has different values. Actually the values are same so it is not easily testable.

### DIFF
--- a/doc/pattern.md
+++ b/doc/pattern.md
@@ -4,6 +4,39 @@ title: Spoon Patterns
 
 Spoon patterns enables you to find code elements. A Spoon pattern is based on a one or several AST nodes, which represent the code to match, where some parts of the AST are pattern parameters. When a pattern is matched, one can access to the code matched in each pattern parameter.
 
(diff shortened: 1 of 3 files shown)


pvojtechovsky:
It ignores difference between implicit and non implicit blocks too. For example
```java
if (x) { 
  dosomething();
}
```
and
```java
if (x) 
  dosomething();
```
are same from pattern matcher point of view

monperrus:
LGTM.

monperrus:
I guess it's important for users. Could you update the documentation to explain this?

pvojtechovsky:
Documentation was added. Thanks for that idea ... I do not like writing documentation much, but I agree that it is critical to have good documentation. It is not enough to have good code ;-)

monperrus:
Thanks!

Issue #2714, opened 2018-10-23
feat: add support for ${java.version} in pom.xml


lodart:
Hi ! 
I have an issue creating a MavenLauncher for projects that have `${java.version}` in their `POM.xml`.

```
Exception in thread "main" java.lang.NumberFormatException: For input string: ""
	at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	at java.base/java.lang.Integer.parseInt(Integer.java:662)
	at java.base/java.lang.Integer.parseInt(Integer.java:770)
	at spoon.support.compiler.SpoonPom.getSourceVersion(SpoonPom.java:259)
	at spoon.MavenLauncher.init(MavenLauncher.java:171)
	at spoon.MavenLauncher.<init>(MavenLauncher.java:119)
	at spoon.MavenLauncher.<init>(MavenLauncher.java:60)
	at org.rlebouc.Main.main(Main.java:21)
```
Replacing it with `1.8` for example seems to work around.

Thank you for your help !

monperrus:
Do you use the spoon-maven-plugin or the java class MavenLauncher?

lodart:
Sorry, I got confused : I use the MavenLauncher java class. I'll edit the title of the issue.

monperrus:
That's probably not to hard to support. Would you try your first pull request?

lodart:
Yes, absolutely, but I'm not sure I can do it. I'll have a look.

surli:
@lodart as a first move, I'd check what happens in this method: https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/compiler/SpoonPom.java#L202-L211 with a breakpoint on the return.

lodart:
The problem seems to come from here : 
https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/compiler/SpoonPom.java#L259
If `${java.version}` is set to `1.X` it works well, the first chars `1.` being removed, it remains only the version number. But if it is set to `11` then it remains only an empty string.

nharrand:
Or just check if it contains a ".".

PR #2729, opened 2018-10-29
fix: java version not starting with "1." in pom.xml

lodart:
If the java version does not start with `1.` in pom.xml, a `NumberFormatException` is thrown when creating a new `MavenLauncher`. (issue #2714)
We check if the java version number contains a `.` before removing it.

I added an Assert in `MavenLauncherTest`. 

I haven't done many PRs so feel free to tell me if I did anything wrong.

### DIFF
--- a/src/main/java/spoon/support/compiler/SpoonPom.java
+++ b/src/main/java/spoon/support/compiler/SpoonPom.java
@@ -286,7 +286,7 @@ public class SpoonPom implements SpoonResource {
 		return environment.getComplianceLevel();
 	}
 
-	private int correctJavaVersion(String javaVersion){
+	private int correctJavaVersion(String javaVersion) {
 		String version = extractVariable(javaVersion);
 		return Integer.parseInt((version.contains(".") ? version.substring(2) : version));
 	}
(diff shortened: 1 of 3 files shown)


src/main/java/spoon/support/compiler/SpoonPom.java
```diff
@@ -256,36 +256,41 @@ public int getSourceVersion() {
...
 		javaVersion = getProperty("maven.compile.source");
 		if (javaVersion != null) {
-			return Integer.parseInt(extractVariable(javaVersion).substring(2));
+			return correctJavaVersion(javaVersion);
 		}
 		// return the current compliance level of spoon
 		return environment.getComplianceLevel();
 	}
 
+	private int correctJavaVersion(String javaVersion){
```
nharrand:
```suggestion
	private int correctJavaVersion(String javaVersion) {
```

nharrand: To please checkstyle. Otherwise LGTM.

lodart: Thank's !

lodart:
My bad, yes it is.

surli:
thanks for your contrib!

PR #2731, opened 2018-10-30
feat: new CtTypeMemberWildcardImportReference for `import package.Type.*`

pvojtechovsky:
Fixes implementation of CtWildcardStaticTypeMemberReferenceImpl to be a real spoon model element.

I also change the semantic of this element. Before it was kind of a CtTypeReference. Now it is only CtReference, because it references all static type members of the referred type.

It also have a new method (main reason why I needed it): `getTypeReference()`, which returns type reference to the type whose static type members are included.


monperrus:
I like this idea, but maybe not the name:

what do you think of `CtWildcardTypeMemberImport` or `CtTypeMemberWildcardImport`? 
(will we have `CtPackageWildcardImport` in the future?)

WDYT?

pvojtechovsky:
> name

All suggesttions ends with `*Import`, but it is not an **import**, it is really a kind of **reference** to all static type members. Note that we have `CtImport#setReference` method, which accepts reference ... (`CtWildcardStaticTypeMemberReference`) as an argument.

I would prefer to have a name which ends with `Reference`.

And note that name is already invented and committed by Simon. So any change causes little incompatibility, because we should change also `CoreFactory#createWildcardStaticTypeMemberReference` ...

monperrus:
>  it is really a kind of reference to all static type members

not really. The definition of a reference is something for which you can get a declaration (`getDeclaration`). There is no meaningful declaration in this case.

> And note that name is already invented and committed by Simon

That was the implementation class (which can stay as is, and, you're right, I'm less rigorous for naming of implementation classes than for naming of public interfaces), it's perfectly OK to improve the name for the promoted interface.

what about `CtTypeMemberWildcardImportReference`?

pvojtechovsky:
@monperrus I renamed to new name. It is ready for merge.

monperrus:
Thanks for the progress on import-sniper!

PR #2733, opened 2018-10-31
fix: fix invalid if model with empty statements in then/else branch

Egor18:
Hi! I found out that spoon is probably wrong about the following code:

```
if (x);
else System.out.println();
```

It says that `ifElement.getThenStatement()` here is `System.out.println()` and `ifElement.getElseStatement()` is null. But it's incorrect.
Then statement should be null and the else statement should be System.out.println();

This PR provides a failing test case.

### DIFF
--- a/src/main/java/spoon/support/compiler/SnippetCompilationHelper.java
+++ b/src/main/java/spoon/support/compiler/SnippetCompilationHelper.java
@@ -128,9 +128,7 @@ public class SnippetCompilationHelper {
 		CtMethod<?> wrapper = c.getMethod(WRAPPER_METHOD_NAME);
 
 		List<CtStatement> statements = wrapper.getBody().getStatements();
-
-
-		CtStatement ret = statements.get(0);
+		CtStatement ret = statements.get(statements.size() - 1);
 
 		// Clean up
 		c.getPackage().removeType(c);
--- a/src/main/java/spoon/support/compiler/SnippetCompilationHelper.java
+++ b/src/main/java/spoon/support/compiler/SnippetCompilationHelper.java
@@ -128,7 +128,9 @@ public class SnippetCompilationHelper {
 		CtMethod<?> wrapper = c.getMethod(WRAPPER_METHOD_NAME);
 
 		List<CtStatement> statements = wrapper.getBody().getStatements();
-		CtStatement ret = statements.get(statements.size() - 1);
+
+
+		CtStatement ret = statements.get(0);
 
 		// Clean up
 		c.getPackage().removeType(c);
--- a/src/main/java/spoon/support/compiler/jdt/ContextBuilder.java
+++ b/src/main/java/spoon/support/compiler/jdt/ContextBuilder.java
@@ -18,6 +18,7 @@ package spoon.support.compiler.jdt;
 
 import org.eclipse.jdt.internal.compiler.ast.ASTNode;
 import org.eclipse.jdt.internal.compiler.ast.CompilationUnitDeclaration;
+import org.eclipse.jdt.internal.compiler.ast.EmptyStatement;
 import org.eclipse.jdt.internal.compiler.ast.Expression;
 import org.eclipse.jdt.internal.compiler.ast.TypeDeclaration;
 import org.eclipse.jdt.internal.compiler.ast.TypeReference;
@@ -94,7 +95,8 @@ public class ContextBuilder {
 	void enter(CtElement e, ASTNode node) {
 		stack.push(new ASTPair(e, node));
 		if (!(e instanceof CtPackage) || (compilationUnitSpoon.getFile() != null && compilationUnitSpoon.getFile().getName().equals(DefaultJavaPrettyPrinter.JAVA_PACKAGE_DECLARATION))) {
-			if (compilationunitdeclaration != null && !e.isImplicit()) {
+			if (compilationunitdeclaration != null && !e.isImplicit()
+					|| node instanceof EmptyStatement) {
 				e.setPosition(this.jdtTreeBuilder.getPositionBuilder().buildPositionCtElement(e, node));
 			}
 		}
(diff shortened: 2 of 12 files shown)


monperrus:
working on this

monperrus:
Thanks a lot for the bug report and the failing test case.

An invalid model is one of most serious bugs we can have.

Fixed!

monperrus:
If you look at the test failures, handling empty statements while remaining fully backward compatible is a challenge :)

Suggestions welcome!

pvojtechovsky:
I will have a look at it now

pvojtechovsky:
Do we want to see Empty statements in Spoon model?
For example this java code
```java
;;;;;;;;;;
```
Should it be represented as
A) no statement
B) 10 empty statements

WDYT?

Current solution which represents empty statement as CtComment with implicit=true and non empty position with source code `;` is breaking some internal concepts like:
* implicit element has no source position

pvojtechovsky:
It seems to me that the cleaner solution is to introduce new element `CtEmptyStatement`. WDTY?

Egor18:
I've just discovered another bug with that kind of code. Spoon (master branch) fails to build a model when encounters the following code:

```
if (true) // comment
    ;
```

```
	at spoon.support.compiler.jdt.JDTCommentBuilder$1.visitCtIf(JDTCommentBuilder.java:435)
	at spoon.support.reflect.code.CtIfImpl.accept(CtIfImpl.java:45)
	at spoon.reflect.visitor.CtInheritanceScanner.scan(CtInheritanceScanner.java:183)
	at spoon.support.compiler.jdt.JDTCommentBuilder$1.scan(JDTCommentBuilder.java:300)
	at spoon.support.compiler.jdt.JDTCommentBuilder.insertCommentInAST(JDTCommentBuilder.java:498)
	at spoon.support.compiler.jdt.JDTCommentBuilder.buildComment(JDTCommentBuilder.java:159)
	at spoon.support.compiler.jdt.JDTCommentBuilder.build(JDTCommentBuilder.java:115)
	at spoon.support.compiler.jdt.JDTBasedSpoonCompiler.buildModel(JDTBasedSpoonCompiler.java:446)
	at spoon.support.compiler.jdt.JDTBasedSpoonCompiler.buildUnitsAndModel(JDTBasedSpoonCompiler.java:374)
	at spoon.support.compiler.jdt.JDTBasedSpoonCompiler.buildSources(JDTBasedSpoonCompiler.java:339)
	at spoon.support.compiler.jdt.JDTBasedSpoonCompiler.build(JDTBasedSpoonCompiler.java:122)
	at spoon.support.compiler.jdt.JDTBasedSpoonCompiler.build(JDTBasedSpoonCompiler.java:105)
	at spoon.Launcher.buildModel(Launcher.java:767)
	at spoon.Launcher.run(Launcher.java:718)
	at spoon.Launcher.run(Launcher.java:118)
	at spoon.test.comment.CommentTest.getSpoonFactory(CommentTest.java:104)
```

That's because here in JDTCommentBuilder.java if `e.getThenStatement()` is `null`, we will get NPE:
```
public void visitCtIf(CtIf e) {
    if (!(e.getThenStatement() instanceof CtBlock)) {
        if (comment.getPosition().getSourceEnd() <= e.getThenStatement().getPosition().getSourceStart()) {
```

We can simply discard that comment in that case, or probably do something better.

pvojtechovsky:
Hi Egor, thanks for these tests. It shows another reason why we need empty statement ... to be able to link comment to some entity.

I have found that `CtElement extends CtStatement`, but it is quite wild concept from my point of view, which cannot be used to describe something like:
```java
/*a*/;/*b*/;
```

monperrus:
> Do we want to see Empty statements in Spoon model?

I would say no. One core design principle of Spoon is to abstract over syntax, and empty statements are purely syntactic.

I propose to fix this bug and the comment one in two different PRs.
@pvojtechovsky could you create an issue with the comment bug?

However, we can have temporary empty statements. In this particular bug, we need a fake node just for the exiter, and we can remove it afterwards. This is the best solution I see so far.

WDYT?

pvojtechovsky:
> One core design principle of Spoon is to abstract over syntax, 

I like that design principle a lot. It allows powerful Patterns, Matchers, ... etc. 

> and empty statements are purely syntactic.

I am not sure here.

I agree that
```java
print();;;
```
should be simplified to 
```java
print();
```
by Spoon. Nobody wants to analyze empty statements.

But 
```java
if (x) ;
then print();
```
is may be something else. If we represent it as **empty implicit CtBlock**, then it is not consistent, because we have no AST node which represents that statement `;`... it will probably cause problems in sniper printer, because it expectes that implicit elements has no source code ... The nicer and may be better for analyze it would be to have implicit CtBlock with one `CtEmptyStatement` in it. I do not know. I never needed to analyze something like that ;-)

@Egor18 WDYT? What would you expect in your use case? Could you describe your use case of these empty statements so we can fix it best way?

Egor18:
@pvojtechovsky, I was trying to run Spoon over the [spotbugs](https://github.com/spotbugs/spotbugs) tests, just to see if it's ok with some tricky code like this one, and unfortunately, the model build failed.

From my point of view, the best possible solution is just to fix that case only (return null for the then part). 
A new `CtEmptyStatement` or implicit `CtBlock` in that would break backward compatibility. 
For example, in our analyzer I have the code, which relies on the fact that `ifElement.getThenStatement() == null`, to detect suspicious code like that (odd semicolon after if):
```
if (cond);
{
  /* ... */
}
```
It's not a big deal to change that code, but all the users would have to do that.

Regarding the comment: `if (cond) /* comment */ ;`
I believe we can simply attach it to the whole CtIfStatement (it's a rare and strange case anyway).

pvojtechovsky:
ok, thanks for feedback. I will do it like you suggest.

monperrus:
> I agree that
>
> print();;;
>
> should be simplified to
>
> print();
Good, I also agree.

> If we represent it as *empty implicit CtBlock*, then it is not consistent, because we have no AST
> node which represents that statement
This is the current definition of an implicit node in Spoon: something in the model that is not in
the source.

WDYT?

pvojtechovsky:
Looks like all tests passed. It is ready for review.

monperrus:
really elegant! I will merge.

pvojtechovsky:
Thanks for compliment :-)

PR #2735, opened 2018-11-01
fix(documentation): pattern.md

pvojtechovsky:


pvojtechovsky:
the space before `*` is needed to format bullet correctly. See the problem in documentation on web

surli:
Oh indeed. I had only checked on Github directly and the output was ok there.

PR #2750, opened 2018-11-07
test: remove long logs fof CtGenerationTest which makes CI log unreadable

monperrus:
Surefire displays ComparisonFailure ... which is not bad for short strings, but completely breaks the output if CtGenerationTest fails eg https://travis-ci.org/INRIA/spoon/jobs/452050077


pvojtechovsky:
We might remove logs produced by Query engine too. It would make test output even more readable.

WDYT?

monperrus:
Feel free to do so by pushing in this PR. I am not a fan of logs that nobody reads.

PR #2753, opened 2018-11-08
refactor: fix code inspection: type parameter extends Object

zielint0:
*Reports any type parameters and wildcard type arguments explicitly declared to extend java.lang.Object.*


pvojtechovsky:
Welcome back Tomasz ;-)
Thanks for continuing of cleaning!

PR #2755, opened 2018-11-08
refactor: sort modifiers in ModelStreamerTest.java

zielint0:
*Reports on declarations whose modifiers are not in the canonical preferred order (as stated in the Java Language Specification).*


pvojtechovsky:
Is it the only occurrence of that problem in tests? If there is more could we make one PR for more/all such changes?

zielint0:
> Is it the only occurrence of that problem in tests? If there is more could we make one PR for more/all such changes?

Only in test classes. Take a look here: https://github.com/INRIA/spoon/pull/2758

zielint0:
@pvojtechovsky 
I am ready to merge.

pvojtechovsky:
So if it is the only place where it is wrong in test case code (all other problems are in test models), then it makes sense to merge it as it is. Thank You Tomasz

PR #2757, opened 2018-11-08
refactor: split multiple variables in VariableReferencesModelTest.java

zielint0:
*Reports multiple variables being declared in a single declaration. Some coding standards prohibit such declarations.*

Multiple variables in one line are hard to read, and easy to overlook. Thing to avoid.


pvojtechovsky:
Do not touch `VariableReferencesModelTest.java` ;-)
It is the model which contains several wild cases ... we need them.
```java
 /**
  * The main purpose of this test is to be transfomed by Spoon into Spoon model, 
  * which is then used by {@link spoon.test.query_function.VariableReferencesTest}
  * 
  * The values of the fields must be a sequence starting from 1
  */
```

PR #2758, opened 2018-11-08
WIP refactor: fix order

zielint0:
The purpose of this PR is to show *missorted modifiers* in testing code (testclasses).

Perhaps (perhaps!) some of modifier sorting could be useful.

I have chose only these testclasses that I consider useful to *fix*. But there is high probability of introducing unwanted changes!

Guys take a look, perhaps some commits could be merged.

PR as a whole is rather informational. It is more detailed response for: https://github.com/INRIA/spoon/pull/2755

### DIFF
--- a/src/test/java/spoon/support/compiler/jdt/testclasses/ExtendedStringLiteralTestClass.java
+++ b/src/test/java/spoon/support/compiler/jdt/testclasses/ExtendedStringLiteralTestClass.java
@@ -2,6 +2,6 @@ package spoon.support.compiler.jdt.testclasses;
 
 public class ExtendedStringLiteralTestClass {
 
-	final public static String extendedStringLiteral = "hello "+"world!";
+	public static final String extendedStringLiteral = "hello "+"world!";
 	
 }
--- a/src/test/java/spoon/test/annotation/testclasses/Main.java
+++ b/src/test/java/spoon/test/annotation/testclasses/Main.java
@@ -23,17 +23,17 @@ public class Main {
 	public void m1() {
 	}
 
-	final public static int INTEGER = 42;
-	final public static String STRING = "Hello World!";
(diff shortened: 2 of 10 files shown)


pvojtechovsky:
The `WIP` prefix in the issue name is enough to stop us from merging. 
The `review` prefix in the issue name says you are finished - we can have a look.
No prefix + every test pass ... merge is possible.

zielint0:
> The `WIP` prefix in the issue name is enough to stop us from merging.

No, one of my WIP PR was merged in the past, so I am cautionus now ;-)

pvojtechovsky:
Are you aware that you modify `**/testclasses/**` which is intended for test models. What is purpose for cleaning there?

Note: if everything in Spoon will be "clean" - including models - then Spoon will be not tested on real life messy source code ;-)

I suggest o keep `**/testclasses/**` as it is.

@monperrus  WDYT?

zielint0:
> Are you aware that you modify `**/testclasses/**` which is intended for test models. What is purpose for cleaning there?

It is not cleaning. It is an answer for your question about missorted modifiers - https://github.com/INRIA/spoon/pull/2755

> Is it the only occurrence of that problem in tests? If there is more could we make one PR for more/all such changes?

zielint0:
@pvojtechovsky 
> Note: if everything in Spoon will be "clean" - including models - then Spoon will be not tested on real life messy source code ;-)

I have chosen only these testclasses where test case focuses on something different than keywords order (but there could be mistakes with choice).

The benefit is easier reading of test cases.

zielint0:
> > but there could be mistakes with choice
> 
> I do not think it makes sense to spent our free time to check whether this change is safe or not. I will rather live with arbitrary order of modifiers ;-)

Yeah, this PR was just answer for your question! ;-)

PR #2765, opened 2018-11-13
fix: fix incomplete behavior: Javadoc#setContent should parse tags

monperrus:
Javadoc#setContent should parse tags

for this, encapsulates tag parsing in CtJavaDocImpl

(baby PR to make progress on #2172)

### DIFF
--- a/src/main/java/spoon/support/reflect/code/CtCommentImpl.java
+++ b/src/main/java/spoon/support/reflect/code/CtCommentImpl.java
@@ -26,6 +26,8 @@ import spoon.reflect.visitor.CtVisitor;
 
 import java.util.Objects;
 
+import static spoon.support.compiler.jdt.JDTCommentBuilder.cleanComment;
+
 public class CtCommentImpl extends CtStatementImpl implements CtComment {
 	private static final long serialVersionUID = 1L;
 
@@ -63,7 +65,7 @@ public class CtCommentImpl extends CtStatementImpl implements CtComment {
 	@Override
 	public <E extends CtComment> E setContent(String content) {
 		getFactory().getEnvironment().getModelChangeListener().onObjectUpdate(this, CtRole.COMMENT_CONTENT, content, this.content);
-		this.content = content;
+		this.content = cleanComment(content);
 		return (E) this;
 	}
 
--- a/src/main/java/spoon/support/reflect/code/CtCommentImpl.java
+++ b/src/main/java/spoon/support/reflect/code/CtCommentImpl.java
@@ -32,7 +32,7 @@ public class CtCommentImpl extends CtStatementImpl implements CtComment {
 	private static final long serialVersionUID = 1L;
 
 	@MetamodelPropertyField(role = CtRole.COMMENT_CONTENT)
-	private String content;
+	protected String content;
 
 	@MetamodelPropertyField(role = CtRole.COMMENT_TYPE)
 	private CommentType type;
--- a/src/main/java/spoon/support/reflect/code/CtJavaDocImpl.java
+++ b/src/main/java/spoon/support/reflect/code/CtJavaDocImpl.java
@@ -116,6 +116,11 @@ public class CtJavaDocImpl extends CtCommentImpl implements CtJavaDoc {
 
 		tags.clear();
 
+		// avoiding NPE later
+		if (content == null) {
+			content = "";
+		}
+
 		String longDescription = "";
 		String currentTagContent = "";
 		CtJavaDocTag.TagType currentTag = null;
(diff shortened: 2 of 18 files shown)


src/main/java/spoon/support/reflect/code/CtCommentImpl.java
```diff
@@ -63,7 +65,7 @@ public String getRawContent() {
 	@Override
 	public <E extends CtComment> E setContent(String content) {
 		getFactory().getEnvironment().getModelChangeListener().onObjectUpdate(this, CtRole.COMMENT_CONTENT, content, this.content);
-		this.content = content;
+		this.content = cleanComment(content);
```
pvojtechovsky: Shouldn't model change listener receive cleaned comment? ... because it is the real new value
I am not sure. 
WDYT?

src/main/java/spoon/support/reflect/code/CtJavaDocImpl.java
```diff
@@ -96,6 +99,69 @@ public String getShortDescription() {
 		}
 	}
 
+	@Override
+	public <E extends CtComment> E setContent(String content) {
+		// in Javadoc we parse the tags
+		parseTags(cleanComment(content));
```
pvojtechovsky: parseTags adds tags internally. Sp we should clear old tags here. Or did I overlook something?

src/main/java/spoon/support/reflect/code/CtCommentImpl.java
monperrus: I thought about that too. I'm not sure either. The problem is that if we add an assignment as first statement, this will break the test verifying that the first statement calls the listener .

src/main/java/spoon/support/reflect/code/CtJavaDocImpl.java
monperrus: you're right! fixed in parseTags!

src/main/java/spoon/support/reflect/code/CtCommentImpl.java
pvojtechovsky: so let's keep it simple like it is now. May be just put there a comment that we are not sure and it is possible to move it up when somebody founds it useful later.

monperrus:
Interesting, the `tags.clear()` surfaces a very subtle bug in cloning (`testGenerateCloneVisitor` in CI).

So we first have to look at #2767 and #2768.

src/main/java/spoon/support/reflect/code/CtCommentImpl.java
monperrus: Agree. I mark this point as resolved.

src/main/java/spoon/support/reflect/code/CtJavaDocImpl.java
```diff
@@ -96,6 +99,75 @@ public String getShortDescription() {
...
+		// avoiding NPE later
+		if (content == null) {
+			content = "";
+		}
+
+		String longDescription = "";
+		String currentTagContent = "";
+		CtJavaDocTag.TagType currentTag = null;
+
+		String[] lines = cleanComment(content).split("\n|\r\n|\r");
```
pvojtechovsky: cleanComment always returns string with CtComment.LINE_SEPARATOR as line separator.
So it is enough to write `split("\n")

src/main/java/spoon/support/reflect/code/CtJavaDocImpl.java
```diff
@@ -96,6 +99,75 @@ public String getShortDescription() {
...
+	 * @param comment the comment
+	 * @param tagContent the tagContent of the tag
+	 * @param tagType the tag type
+	 */
+	private void defineCommentContent(String tagContent, CtJavaDocTag.TagType tagType) {
+		if (tagType != null) {
+			CtJavaDocTag docTag = this.getFactory().Code().createJavaDocTag(tagContent, tagType);
+			this.addTag(docTag);
+		} else if (!tagContent.isEmpty()) {
+			this.setContent(tagContent.trim());
```
pvojtechovsky: Is it recursive call to setContent which calls defineCommentContent, which calls setContent ... ?
I have not checked it deeper I just do not understand it on first sight.

pvojtechovsky:
Martin, it looks good to me.

src/main/java/spoon/support/reflect/code/CtJavaDocImpl.java
monperrus: fixed!

src/main/java/spoon/support/reflect/code/CtJavaDocImpl.java
monperrus: fixed

monperrus:
ready to merge.

pvojtechovsky:
Thank you Martin!

Issue #2747, opened 2018-11-06
Modularization/Splitting of Spoon


pvojtechovsky:
There are several reasons why it might make sense to split Spoon into several smaller pieces

R1) There are clients who are not using Maven, Decompiler, Eclipse JDK compiler, ... and complains about too many dependencies

R2) There might be communities who are thinking about development of java analyzer and they might appreciate a light (0 dependency) model of `Java Semantic AST`. 

R3) Then they might be interested:
* in printing of such AST
* in traversing, matching, refactoring, querying, ...
... all that might be quite light highly usable libraries with little dependencies.

For example a transpiler from C to Java does not need Eclipse compiler, but needs Semantic AST + PrettyPrinter.

**May be the Spoon community might grow faster** with such smaller products/modules. Because there would be a higher amount of use cases.

WDYT?

For example I am not using Eclipse modeling tools, because they had (I do not know current state) very heavy dependencies. I would need light modeler. ... even for maintenance of Spoon meta ... meta model the third party light modeling tool would be nice to have ...

nharrand:
Do you mean separated repositories, or just publishing multiple smaller artifacts?

pvojtechovsky:
It is a technical detail from my point of view.

Actually I tend to publishing multiple smaller artifacts, because it seems to be easier. But if there would be a good reason to separate repositories ...

monperrus:
There are three levels of modularity:
- repositories
- code modules
- deployed units

I'm in favor of keeping one unique repository, because I consider a repo as a unit of community and conversation, not as a unit of code.

In Maven, the general rule is that there is a one-to-one relation between code module and deployed unit (one Maven module = one jar), however this one-to-one relation can be questioned, either in the context of Maven, or outside Maven.

The questions to answer first are: what the maximum size of a jar we can accept? why? have we reached that limit already?

nharrand:
> It is a technical detail from my point of view.

I agree but:
> I'm in favor of keeping one unique repository, because I consider a repo as a unit of community and conversation, not as a unit of code.

Yes that was my main concern too.

pvojtechovsky:
> The questions to answer first are: what the maximum size of a jar we can accept? why? have we reached that limit already?

My primary reason for splitting of Spoon is not about size of JAR. It is about message like 

> The Spoon lib is not a monolith which is tightly bound to heavy Eclipse compiler etc.

but 

> The Spoon model and algorithms can be successfully used without any extra dependencies

[...]

Issue #2737, opened 2018-11-02
Missing dependencies: "org.benf:cfr:0.132.0" and "org.eclipse.platform:org.eclipse.equinox.app"


RichardBradley:
The published version 7.1.0 of spoon-core depends on a library that is not in the public maven repositories: `org.benf:cfr:0.132.0`

You can see this at e.g. https://mvnrepository.com/artifact/fr.inria.gforge.spoon/spoon-core/7.1.0 which lists it as a dep but does not have a hyperlink to a published version of this artefact.

When I try to compile a project against 7.1.0 spoon-core, I get the following error (using SBT/Ivy):

```
[info] Updating ...
[warn] 	module not found: org.benf#cfr;0.132.0
```

Similarly, it seems that 7.0.0 has a transitive dependency on `org.eclipse.platform:org.eclipse.equinox.app`, which I cannot resolve either, but things seem to work OK if I exclude that dep, i.e.

```
libraryDependencies += "fr.inria.gforge.spoon" % "spoon-core" % "7.0.0" exclude("org.eclipse.platform", "org.eclipse.equinox.app")
```

nharrand:
Yes the dependency is available on the inria maven repo:
```xml
<repository>
      <id>inria</id>
      <name>triskell-public-release</name>
      <url>http://maven.inria.fr/artifactory/triskell-public-release</url>
</repository>
```
Not sure how to add a repo to sbt though.

monperrus:
Does it mean that the `with-dependencies` is not anymore a true `with-dependencies`, esp. for non Maven users?

monperrus:
@RichardBradley did you take the with-dependencies version?

RichardBradley:
@monperrus, I did not use the with-dependencies jar, no. I think it is more conventional to use the standard jar, as it allows for de-duplication of shared dependencies and detection of conflicting dependencies etc.

RichardBradley:
@nharrand:  the docs need updating to reference this non-standard repo then. The front-page does not mention this at present: https://github.com/INRIA/spoon#getting-started-in-2-seconds

I don't know what spoon needs "cfr" or "eclipse.equinox.app" for, but the latter at least looks like a possible mistake to me.

RichardBradley:
I expect that adding the maven.inria.fr repository will work. I have not tried it, because I chose to downgrade to ver 7.0.0 rather than trust a repos I have never heard of.

[...]

PR #2766, opened 2018-11-15
refactor: Excluding decompiler related features from spoon-core

nharrand:
Attempt to implement #2747
Related to #2737


monperrus:
Looks really good to me. This will enable to have a clean Maven central release without any odd dependency.

We can also add a travis job to test it.

@surli with respect to our conversation, WDYT?

spoon-bytecode/pom.xml
```diff
@@ -0,0 +1,545 @@
...
+    <modelVersion>4.0.0</modelVersion>
+
+    <!--<parent>
+        <groupId>fr.inria.gforge.spoon</groupId>
+        <artifactId>spoon-core</artifactId>
+        <version>7.2.0-SNAPSHOT</version>
+    </parent>-->
+
+    <groupId>fr.inria.gforge.spoon</groupId>
+    <artifactId>spoon-bytecode</artifactId>
```
surli: I'm not sure about this name: spoon-decompiler, maybe?

spoon-bytecode/pom.xml
```diff
@@ -0,0 +1,545 @@
...
+            <name>triskell-public-release</name>
+            <url>http://maven.inria.fr/artifactory/triskell-public-release</url>
+        </repository>
+    </repositories>
+
+    <dependencies>
+        <dependency>
+            <groupId>fr.inria.gforge.spoon</groupId>
+            <artifactId>spoon-core</artifactId>
+            <version>7.2.0-SNAPSHOT</version>
```
surli: It would be really great/easier for our Maven config to have a spoon-pom, which would be our parent pom and in which we would define the different plugin configurations. Moreover it would allow us to release more easily.

README.md
```diff
@@ -48,19 +48,6 @@ CtClass l = Launcher.parseClass("class A { void m() { System.out.println(\"yeah\
 
 Some examples about the usage of Spoon can be found on [spoon-examples](https://github.com/SpoonLabs/spoon-examples).
 
-:warning: We introduced in Spoon 7.1.0 an external dependency that is not available on Maven Central.
```
surli: This warning is still true for 7.1.0. Let's wait the release before removing it.

doc/launcher.md
```diff
@@ -118,6 +118,8 @@ Spoon provides two out of the shelf decompilers, CFR by default, and Fernflower.
 JarLauncher launcher = new JarLauncher("<path_to_jar>", "<path_to_output_src_dir>", "<path_to_pom>", new FernflowerDecompiler(new File("<path_to_output_src_dir>/src/main/java")));
 ```
 
+:warning: The `JarLauncher` feature (and all features relying on decompilation) are not included in `spoon-core` but in `spoon-bytecode`. If you wnat to use them you should declare a dependency to `spoon-bytecode`.
```
surli: typo: wnat-> want. 

Side note: we might need to upgrade the release documentation.

README.md
nharrand: I really don't want to be annoying with this, so if you want I'll restore it. But, to my knowledge, the part about the need to insert the additional repository for user of spoon, is simply not true.

doc/launcher.md
nharrand: type -> thanks, I'll correct ASAP.
Side not -> Totally agree.

spoon-bytecode/pom.xml
nharrand: Sure, why not. (In my mind, it was a version of spoon the allowed to work on bytecode. Do you think, the "how it is done" is more important to the user?)

spoon-bytecode/pom.xml
nharrand: Yes I agree. I admit this was a bit lazy ;-)

nharrand: Does this means, that I should change the parent `<packaging>jar</packaging>` to  `<packaging>pom</packaging>` ? Do I need to create a sub module just for the spoon-core artifact?

surli: > Does this means, that I should change the parent <packaging>jar</packaging> to <packaging>pom</packaging> ?

yep

> Do I need to create a sub module just for the spoon-core artifact?

yep

Actually you might want to create a folder `spoon-pom` for this module and then don't forget the `<relativePath>` info on both spoon-core pom and spoon-bytecode

spoon-bytecode/pom.xml
surli: It was really a way to open the discussion, I really don't know what's better :)

spoon-bytecode/pom.xml
nharrand: Ok I got. Thanks!

README.md
surli: Ahah still a bit sensitive on that one I guess ;)
for me it's really more to avoid having user complaints in case they got problem because of that. I never manage to reproduce the error neither, so.... + I really want to release quickly your new module, to avoid this dependency issue. 

Now it's up to you

spoon-bytecode/pom.xml
nharrand: How about `spoon-full`?

README.md
nharrand: I have restored it with small modifications. If you don't like them, I'll just restore your version.

nharrand:
I deactivated travis builds for `spoon-bytecode` because it relies on resolving `spoon-core` which in turn has `spoon-pom` as a parent, which is not publish yet, so it fails. We ca just uncomment the two builds in a futur PR. Otherwise, this is ready for review.

nharrand:
Similarly  properties could be move into spoon-pom, but ... #2771

surli:
FTR I got an email from the CFR maintainer: 

> CFR 0.135 (damnit, I even changed my naming scheme to conform, ish. ;) ) is now available on maven central.
```
<dependency>
  <groupId>org.benf</groupId>
  <artifactId>cfr</artifactId>
  <version>0.135</version>
</dependency>
```

chore/travis/travis-spoon-bytecode-verify.sh
```diff
@@ -0,0 +1,21 @@
...
+# This script intends to be run on TravisCI
+# it runs verify and site maven goals
+# and to check documentation links
+
+# fails if anything fails
+set -e
+
+source /opt/jdk_switcher/jdk_switcher.sh
+
+pip install --user CommonMark==0.7.5 requests pygithub
```
surli: Why do you need that?

spoon-pom/pom.xml
```diff
@@ -0,0 +1,187 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <!--<parent>
```
surli: why do you put the parent in comment?

spoon-bytecode/pom.xml
surli: Let's keep spoon-bytecode for now.

spoon-bytecode/pom.xml
```diff
@@ -0,0 +1,504 @@
...
+    <groupId>fr.inria.gforge.spoon</groupId>
+    <artifactId>spoon-bytecode</artifactId>
+    <packaging>jar</packaging>
+    <version>0.1.0-SNAPSHOT</version>
+    <name>Spoon Bytecode</name>
+    <description>Extension of spoon for bytecode through decompilation</description>
+    <url>http://spoon.gforge.inria.fr/</url>
+    <inceptionYear>2007</inceptionYear>
+
+    <distributionManagement>
```
surli: There is still a lot of redundancy in this pom, now you created a spoon-pom.

nharrand: Do you want want migrate more?

chore/travis/travis-spoon-bytecode-verify.sh
nharrand: That's a copy from `chore/travis/travis-verify.sh`, shall I remove it?

spoon-bytecode/pom.xml
pvojtechovsky: > spoon-full

no please.

> spoon-decompiler or spoon-bytecode

I like `spoon-decompiler` more, because it express that feature better. Spoon can decompile class (bytecode) and generate AST for that.
But `bytecode` is more general term, which might mean also bytecode generation (= compiling), which is already possible with spoon thanks to included JDK compiler.

Let's wait for feedback of others - if any - before you start to rename it. I can live with both names ;-)

pvojtechovsky:
I do not use Maven and I don't try to understand that building stuff, but I like the idea of this PR a lot. 👍 Thank You Nicolas!

.travis.yml
```diff
@@ -16,6 +16,8 @@ env:
   - SCRIPT=travis-verify.sh
   - SCRIPT=travis-maven339-jdk8.sh
   - SCRIPT=travis-jdk10.sh
+#  - SCRIPT=travis-spoon-bytecode-jdk8.sh
```
surli: why is it commented? Not working?

chore/travis/travis-spoon-bytecode-verify.sh
surli: I think we needed it for check links but apparently it's removed from the script. So I think it can be removed yes.

spoon-bytecode/pom.xml
surli: +1 for spoon-decompiler then. Thanks @pvojtechovsky for managing to say what was bothering me :)

spoon-bytecode/pom.xml
```diff
@@ -0,0 +1,476 @@
...
+    <packaging>jar</packaging>
+    <version>0.1.0-SNAPSHOT</version>
+    <name>Spoon Bytecode</name>
+    <description>Extension of spoon for bytecode through decompilation</description>
+
+    <dependencies>
+        <dependency>
+            <groupId>fr.inria.gforge.spoon</groupId>
+            <artifactId>spoon-core</artifactId>
+            <version>7.2.0-SNAPSHOT</version>
```
surli: The version of spoon-core should be always the same as the one defined in spoon-pom. So you should use something like `${project.version}`.

spoon-bytecode/pom.xml
```diff
@@ -0,0 +1,476 @@
...
+        </dependency>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <version>4.12</version>
```
surli: The version of junit used should be the same in all subprojects of Spoon. So you should define the version in spoon-pom properties.

spoon-bytecode/pom.xml
```diff
@@ -0,0 +1,476 @@
...
+            <version>2.5.0.Final</version>
+        </dependency>
+        <dependency>
+            <groupId>org.benf</groupId>
+            <artifactId>cfr</artifactId>
+            <version>0.135</version>
+        </dependency>
+    </dependencies>
+
+    <build>
```
surli: why all this section is not defined in spoom-pom?

surli:
Sorry for asking lot of changes but the idea here is really to have a first external module of Spoon to start seeing how we can decompose some parts, so the most info we can put in an external pom and reuse, the better. 
By the way, a good thing you can try before we merge the release is a dry-run mvn release, on spoon-pom: `mvn release:prepare -DdryRun=true`, this will check that everything's ok to release all spoon artefacts at once. Thanks

nharrand:
> Sorry for asking lot of changes but the idea here is really to have a first external module of Spoon to start seeing how we can decompose some parts, so the most info we can put in an external pom and reuse, the better.
Don't be, it's actually helpful. But, to answer a lot of your comments, my problem is that spoon-pom is not yet published on maven central. So, locally on my machine I can move a lot of things in spoon pom, but if I do, I'll break CI. So I was trying to achieve an ugly middle ground where I add stuff to `spoon-pom`, planing to remove them from other pom after.
Do you have an other solution? (Should I not care to break CI, if everything run smoothly locally?)

spoon-bytecode/pom.xml
nharrand: You mean `<build>`?

nharrand: or `dependency`?

README.md
nharrand: @surli WDYT?

surli: IMO the message for 7.2.0 is useless, since the warning will be removed when Spoon 7.2.0 will be released.

pom.xml
```diff
@@ -20,204 +20,29 @@
...
-    <url>https://github.com/INRIA/spoon</url>
-    <connection>scm:git:https://github.com/INRIA/spoon.git</connection>
-    <developerConnection>scm:git:ssh://github.com/INRIA/spoon.git</developerConnection>
-    <tag>HEAD</tag>
-  </scm>
 
   <properties>
     <java.src.version>1.8</java.src.version>
+  </properties>
+  <!--<properties>
```
surli: you might want to remove them instead of commenting them

pom.xml
```diff
@@ -228,7 +53,7 @@
     <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
-      <version>4.12</version>
+      <version>${junit.version}</version>
       <scope>test</scope>
     </dependency>
     <dependency>
```
surli: I assume that all dependencies below are specific to spoon-core and useless for spoon-decompiler, am I right?

spoon-decompiler/pom.xml
```diff
@@ -0,0 +1,112 @@
...
+                        <id>make-assembly</id><!-- this is used for inheritance merges -->
+                        <phase>package</phase><!-- append to the packaging phase. -->
+                        <goals>
+                            <goal>attached</goal><!-- goals == mojos -->
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+
+            <plugin>
```
surli: you removed checkstyle from spoon-core, so I assume you put it in spoon-pom, so it means you don't need it here.

pom.xml
nharrand: Yes. But if one turns out to be needed for `spoon-decompiler` too, we'll move it to `spoon-pom`, if that's what you mean.

spoon-decompiler/pom.xml
nharrand: No. Both `spoon-core` and `spoon-decompiler` declare the checkstyle plugin but with differents option (exculsion of some sources for exemple).

pom.xml
nharrand: Yes. That's some leftover trial and errors.

README.md
nharrand: I agree. I'll remove it then.

nharrand:
Shall I add a `README.md` to the new modulde?

monperrus:
We did the last changes per our discussion, CI is green.

@surli do you merge?

spoon-decompiler/pom.xml
nharrand: But the checkstyle profile is now in `spoon-pom`.

PR #2775, opened 2018-11-22
refactor: getDocComment uses the same implementation as pretty-printer

monperrus:
(reduce code duplication)


pvojtechovsky:
Thank you Martin

Issue #1467, opened 2017-07-06
bug: FilterTest#testClassCastExceptionOnForEach seems to fail on CI


surli:
See: https://ci.inria.fr/sos/job/Spoon%20Install%20Commit%20Hook/224/console

> Tests run: 40, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 4.194 sec <<< FAILURE! - in spoon.test.filters.FilterTest
testClassCastExceptionOnForEach(spoon.test.filters.FilterTest)  Time elapsed: 0.109 sec  <<< ERROR!
spoon.SpoonException: Execution of query callback failed
	at spoon.test.filters.FilterTest.testClassCastExceptionOnForEach(FilterTest.java:884)
Caused by: java.lang.ClassCastException: spoon.reflect.CtModelImpl$CtRootPackage cannot be cast to spoon.reflect.declaration.CtType
	at spoon.test.filters.FilterTest.testClassCastExceptionOnForEach(FilterTest.java:884)

surli:
This bug seems to occur with a specific version of JDK, I manage to reproduce it with Oracle JDK 8.0.40: https://ci.inria.fr/sos/job/Spoon%20Install%20Commit%20Hook/229/consoleFull

The last version of JDK 8 seems to fix the bug, which is why we did not see the bug with Travis: https://ci.inria.fr/sos/job/Spoon%20Install%20Commit%20Hook/226/consoleFull

nharrand:
On travis it occurs with both openjdk11 and oracle jdk11.
See https://travis-ci.org/INRIA/spoon/jobs/459858096
It might be worth reopening this issue.

PR #2782, opened 2018-11-26
WIP: perf(.travis.yml): merge travis builds to fall under 5 builds

nharrand:
As travis builds are limited to 5 parallel builds for Spoon, I propose to merge the 3 fastest builds into one. (This 3 builds are just verification that are not tests, and are quite fast.) ~~Hopefully this new build is still faster than the slowest one~~, so it should be a net gain on ~~both time and~~ resources.


nharrand:
Ok, my bad `travis-coverage.sh` is actually comparably slow to `travis-jdk8.sh`. 
 * Would it make sense to make it failable and remove `travis-jdk8.sh`?
 * Or maybe can we merge former `travis-verify.sh`, and former travis-maven339-jdk8.sh` with `travis-spoon-decompiler.sh` as they are the actual 3 fastest?

surli:
I would personnally merge mvn339-jdk8 and jkd8 to keep only one. And I'd remove jdk9 since it won't be a LTS: we should only test against JDK8 and 11, since 11 will be the new LTS. For corner cases we should have checks for 7, 9 and 10 on another CI.

nharrand:
Sounds good to me.

Following your advice, I've replaced `travis-jdk9.sh` by `travis-jdk11.sh` and removed `travis-jdk8.sh` in favor of `travis-maven339-jdk8.sh`. 

As we are still one build short, I've left the merge between `travis-verify.sh` and `travis-spoon-decompiler.sh`. (Alternatively, we can unmerge them, and remove `travis-jdk10.sh`)

surli:
>  back to jdk9 while #1467 occurs with jdk11 

Actually that's interesting: I'd be more in favor of keeping jdk11 and fixing quickly #1467. As I said JDK11 is supposed to be the new LTS for Java. We need to fix that ASAP.

monperrus:
LGTM

monperrus:
Waiting for #2789 to be merged before rebasing.

nharrand:
OK, this finally works.
Ready for review.

PR #2785, opened 2018-11-26
bug: test to reproduce the error with wildcards

monperrus:


monperrus:
@pvojtechovsky you touched the imports recently, would you have a look at this one?

pvojtechovsky:
I will have a look

pvojtechovsky:
I rebased this new test to #2683 and it passes well. So I do not want to try to fix it using old import code, which I would like to get rid.

monperrus:
OK. Let's finish #2683 then.

Issue #2779, opened 2018-11-26
continuous delivery for Spoon


monperrus:
Hi all,

I'd like to propose to set up continuous delivery for Spoon. The idea is that we would do an automated release every month.

Notes:
* we would set up a project `spoon-continuous-delivery` on SpoonLabs, with a monthly Travis job responsible for making the actual release.
* by default, there would be the automated changelog, but we would have a way to extend it by committing to `spoon-continuous-delivery` before the due date (eg by committing to `spoon-continuous-delivery/changelog/7.3.0.md`)
* by default, the last digit of the version would be {year}{month}, eg `7.2.201812`. We would change manually major and minor with normal commits on master.

WDYT?

pvojtechovsky:
I like this idea

nharrand:
+1

nharrand:
If we stick with `{Major}.{Minor}.{Patch}`, we can still:
 * Change the **Major** version when breaking changes occur (on non `@Experimental`, and non `internal` api).
 * Change the **Minor** version when we feel like significant new features/bugfixes/modifications have accumulated as currently.
 * Change the **Patch** monthly automatically.

I think a user who updates his/her maven dependency to spoon when a new **Minor** is published wouldn't be affected.
Only a user who would update automatically for the latest would be affected. But I assume a user that does so is expressing that he/she wants the latest bugfixes ASAP. So I don't feel it is such a contract breach.

surli:
I completely agree on the versioning scheme. And the impact on the users, but I'm not sure to follow exactly what you mean. 

From what I understand, you propose that the release job infers automatically the new version number of the release based on what? The label info from the PRs? 
Or do you assume that each time we break something we will push the new version number for the release?

nharrand:
I assumed (but it's just a proposal) that if we leave it alone, the release job would only release a new **Patch** every month.
But we could manually release **Major** and **Minor** when we decide to.

surli:
> the release job would only release a new Patch every month.

yeah but it would be wrong if we'd merge a breaking PR during the month. So we would need to be extra careful on that case.

nharrand:
Yes for existing API. 
We could also progressively increase the use of `@Experimental` and internal packages, to avoid these problems.

monperrus:
Just did a manual release myself. Full automation won't be straightforward, and it will take us some releases to get the release script working perfectly :-) But it would really save time. I

[...]

PR #2786, opened 2018-11-26
doc: remove all hard coded version numbers to simplify release

monperrus:
baby pr for #2779

### DIFF
--- a/doc/_config.yml
+++ b/doc/_config.yml
@@ -18,9 +18,16 @@ include: [_includes]
 # variables
 
 sidebar_tagline: Spoon
-sidebar_version: version 7.0.0
-spoon_release: "7.1.0"
-spoon_snapshot: "7.2.0-SNAPSHOT"
+
+# this is replaced at deployment time by job https://ci.inria.fr/sos/job/Website Deployer/
+# LATESTVERSION=`curl -s "http://search.maven.org/solrsearch/select?q=g:%22fr.inria.gforge.spoon%22+AND+a:%22spoon-core%22" | jq .response.docs[0].latestVersion`
+# sed -i -e "s/^spoon_release: .*/spoon_release: $LATESTVERSION/" _config.yml
+spoon_release: "TO_BE_REPLACED"
+
+# this is replaced at deployment time by job https://ci.inria.fr/sos/job/Website Deployer/
+# SNAPSHOTVERSION=`xmlstarlet sel -t -v /_:project/_:version ../pom.xml`
+# sed -i -e "s/^spoon_snapshot: .*/spoon_snapshot: \"$SNAPSHOTVERSION\"/" _config.yml
+spoon_snapshot: "TO_BE_REPLACED"
 
 exclude:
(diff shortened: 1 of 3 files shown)


doc/_config.yml
```diff
@@ -18,9 +18,16 @@ include: [_includes]
 # variables
 
 sidebar_tagline: Spoon
-sidebar_version: version 7.0.0
```
surli: just for info, can't you define sidebar_version according to spoon_release?

doc/_config.yml
```diff
@@ -18,9 +18,16 @@ include: [_includes]
...
-spoon_release: "7.1.0"
-spoon_snapshot: "7.2.0-SNAPSHOT"
+
+# this is replaced at deployment time by job https://ci.inria.fr/sos/job/Website Deployer/
+# LATESTVERSION=`curl -s "http://search.maven.org/solrsearch/select?q=g:%22fr.inria.gforge.spoon%22+AND+a:%22spoon-core%22" | jq .response.docs[0].latestVersion`
+# sed -i -e "s/^spoon_release: .*/spoon_release: $LATESTVERSION/" _config.yml
```
surli: Is this command pushed somewhere?

tdurieux:
@GerardPaligot did before a script that automatically updates those version. I feel it is a better solution

doc/_config.yml
monperrus: that is another option, but I prefer to simply get rid of sidebar_version

doc/_config.yml
monperrus: yes, in the Jenkins config, which is mirrored here https://github.com/SpoonLabs/spoon-ci-config

monperrus:
You're right, the problem is that this script is a bash monster that hasn't been executed and maintained for more than 4 years.

tdurieux:
I just looked a the script (that is not really a monster), it seems that it is pretty simple to maintain and only minor changes are required.
see: 
https://github.com/INRIA/spoon/blob/master/doc/_release/deploy_production.sh

monperrus:
Thanks for the link.

Regardless of the script, I'm in favor of the simplification of this PR, on the ground that simpler means less error-prone today and in the future.

doc/_config.yml
surli: Good to know!

doc/_config.yml
surli: ok

GerardPaligot:
That could be interesting to migrate all these old scripts from bash to a better and more maintainable language. You always use Maven to package your project, that could be interesting to migration to Gradle and create better scripts in Groovy or Kotlin for all these tasks that could be automatic. :)

monperrus:
+100 for @GerardPaligot's comment. (good to see you Gérard :-)

For the next release, I'm considering writing some Python scripts.

monperrus:
With no other comments, ready for merge.

monperrus:
ready for merge

PR #2787, opened 2018-11-27
fix(CtQueryImpl#detectTargetClassFromCCE, FilterTest#testInvalidQueryStep): Adapt parsing of ClassCastException message which has changed in oracle jdk 11

nharrand:
See #2782 for test

Apparently, in jdk 11, ClassCastException message can be:
```
class java.lang.Integer cannot be cast to class spoon.reflect.declaration.CtType (java.lang.Integer is in module java.base of loader 'bootstrap'; spoon.reflect.declaration.CtType is in unnamed module of loader 'app')
```
instead of:
```
java.lang.Integer cannot be cast to spoon.reflect.declaration.CtType
```
So:
`CtQueryImpl#detectTargetClassFromCCE` has been changed to parse new messages as well. ( fix#1467 )
`FilterTest#testInvalidQueryStep()` has been changed for the same reason (an assert relied on the previous message).


src/test/java/spoon/test/fieldaccesses/FieldAccessTest.java
```diff
@@ -388,7 +389,7 @@ public void testFieldAccessOnUnknownType() {
 			@Override
 			public <T> void visitCtFieldWrite(CtFieldWrite<T> fieldWrite) {
 				visited++;
-				assertEquals("array", ((CtVariableWrite) fieldWrite.getTarget()).getVariable().getSimpleName());
+				assertEquals("array", ((CtVariableAccess) fieldWrite.getTarget()).getVariable().getSimpleName());
```
surli: Is this related to that PR?

surli: That's actually really since we are inside a `visitCtFieldWrite`. So it would mean the target of a field write is something we cannot write?

surli: I propose that you split in two PR: I'm ok to merge everything, but on this one we really should discuss and understand what happens.

nharrand: Sure. No problem.

Issue #2781, opened 2018-11-26
Set encoding per file


Egor18:
Hi!
There are some big legacy projects, which have mixed encodings like UTF-8 and Cp1251 for source files.
I know that we can set encoding in spoon with `launcher.getEnvironment().setEncoding(...)` for the whole project, however it seems like it's impossible to specify encoding for a specific file.
It cases some trouble. For example if we parse Cp1251 file as UTF-8:
In Cp1251:
```
String s1 = "Привет";
String s2 = "АБВГДЕ";
```

In UTF-8:
```
String s1 = "??????";
String s2 = "??????";

```
These two strings considered equal in AST, because of the encoding.

So what can I do here? I'd like to have some encoding auto detection (outside of spoon), but I'm not sure if it possible to specify encoding for each file separately in spoon.

pvojtechovsky:
It is not yet possible, but I guess there should be possible to make PR which allows to register some callback, which will be called to detect encoding of each file individually.
If I remember well it is Spoon code who reads bytes from file and converts them to characters, so Spoon can influence used encoding too.

surli:
I think an issue might be when using JDT directly: see https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/compiler/jdt/JDTBasedSpoonCompiler.java#L158. Currently we only give one encoding for the whole project to be parsed by JDT. I'm not sure we can give it multiple encodings.

pvojtechovsky:
I checked the code yesterday and it is doable. Yes, we give one encoding to JDK compiler, but internally JDK compiler calls Spoon back and spoon delivers `char[]`. So Spoon can read each file using different encoding.

Egor18:
Thanks, I've just managed to add a specific callback to the `Environment` for that purpose.    
So I can invoke it for each file in `initializeCompiler` method from `FileCompilerConfig`:
```
if (jdtCompiler.getEnvironment().getEncodingCallback() == null) {
    char[] content = IOUtils.toCharArray(inputStream, jdtCompiler.getEnvironment().getEncoding());
    cuList.add(new CompilationUnit(content, fName, jdtCompiler.getEnvironment().getEncoding().displayName()));
}
else {
    String enc = jdtCompiler.getEnvironment().getEncodingCallback().apply(new File(fName));
    char[] content = IOUtils.toCharArray(inputStream, enc);
    cuList.add(new CompilationUnit(content, fName, enc));
}
```


The api looks like this:

```
[...]

PR #2788, opened 2018-11-27
Add encoding detection callback

Egor18:
This PR provides ability to specify user-defined callback, which could be used to detect encoding for each file separately. See #2781.

For example I use juniversalchardet lib:
```
private Charset detectEncoding(byte[] fileBytes) {
   final int maxLength = 4096;
   UniversalDetector detector = new UniversalDetector(null);
   detector.handleData(fileBytes, 0, Math.min(fileBytes.length, maxLength));
   detector.dataEnd();
   String encoding = detector.getDetectedCharset();
   detector.reset();
   if (encoding != null) {
       return Charset.forName(encoding);
   }
   return Charset.forName("UTF-8");
}
```
Spoon api:
```
launcher.getEnvironment().setEncodingDetectionCallback(bytes -> detectEncoding(bytes));
// ...
```
It helps to have correct AST in projects with mixed encodings.

### DIFF
--- a/src/main/java/spoon/compiler/SpoonFile.java
+++ b/src/main/java/spoon/compiler/SpoonFile.java
@@ -16,8 +16,10 @@
  */
 package spoon.compiler;
 
+import org.apache.commons.io.IOUtils;
 import spoon.SpoonException;
 
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.charset.Charset;
@@ -54,8 +56,9 @@ public interface SpoonFile extends SpoonResource {
 		byte[] bytes;
 		try {
 			InputStream contentStream = getContent();
-			bytes = new byte[contentStream.available()];
-			contentStream.read(bytes);
+			ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+			IOUtils.copy(contentStream, outputStream);
+			bytes = outputStream.toByteArray();
 		} catch (IOException e) {
 			throw new SpoonException(e);
 		}
--- a/src/main/java/spoon/compiler/SpoonFile.java
+++ b/src/main/java/spoon/compiler/SpoonFile.java
@@ -54,8 +54,7 @@ public interface SpoonFile extends SpoonResource {
 	 */
 	default char[] getContentChars(Environment env) {
 		byte[] bytes;
-		try {
-			InputStream contentStream = getContent();
+		try (InputStream contentStream = getContent()) {
 			ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
 			IOUtils.copy(contentStream, outputStream);
 			bytes = outputStream.toByteArray();
--- a/src/main/java/spoon/compiler/SpoonFile.java
+++ b/src/main/java/spoon/compiler/SpoonFile.java
@@ -16,7 +16,11 @@
  */
 package spoon.compiler;
 
+import spoon.SpoonException;
+
+import java.io.IOException;
 import java.io.InputStream;
+import java.nio.charset.Charset;
 
 /**
  * This interface represents files that can be used as resources for the Spoon
@@ -41,4 +45,25 @@ public interface SpoonFile extends SpoonResource {
 	 * @return
 	 */
 	boolean isActualFile();
+
+	/**
+	 * Gets the file content as a char array, considering encoding or encoding
+	 * provider.
+	 */
+	default char[] getContentChars(Environment env) {
+		byte[] bytes;
+		try {
+			InputStream contentStream = getContent();
+			bytes = new byte[contentStream.available()];
+			contentStream.read(bytes);
+		} catch (IOException e) {
+			throw new SpoonException(e);
+		}
+		if (env.getEncodingProvider() == null) {
+			return new String(bytes, env.getEncoding()).toCharArray();
+		} else {
+			Charset encoding = env.getEncodingProvider().detectEncoding(this, bytes);
+			return new String(bytes, encoding).toCharArray();
+		}
+	}
 }
--- a/src/main/java/spoon/support/compiler/VirtualFile.java
+++ b/src/main/java/spoon/support/compiler/VirtualFile.java
@@ -23,7 +23,6 @@ import spoon.compiler.SpoonFolder;
 import java.io.ByteArrayInputStream;
 import java.io.File;
 import java.io.InputStream;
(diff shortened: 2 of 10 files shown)


src/test/java/spoon/test/compilationunit/TestCompilationUnit.java
```diff
@@ -267,4 +271,40 @@ public void visitCtCompilationUnit(CtCompilationUnit compilationUnit) {
...
+				.findFirst()
+				.get();
+
+		CtType<?> cp1251Type = model.getAllTypes()
+				.stream()
+				.filter(t -> "Cp1251".equals(t.getSimpleName()))
+				.findFirst()
+				.get();
+
+		assertEquals("\"Привет мир\"", utf8Type.getField("s1").getAssignment().toString());
```
monperrus: I love this expected value :-)

pvojtechovsky:
I would like to have available file name+path in encoding detection callback.
Because for example in Eclipse IDE the encoding of source files can be defined individually using property file with properties like:
```java
encoding//src/main/java/spoon/support/visitor/replace/ReplacementVisitor.java=ISO-8859-1
```
So I can imagine that many clients might be able to select encoding depending on the file name or file path. Of course bytes of file might be good input too!

WDYT?

But before we design the final API for encoding detection, please note this bug in Spoon: 
```java
class VirtualFile implements SpoonFile {
...
	@Override
	public InputStream getContent() {
		return new ByteArrayInputStream(content.getBytes());
	}
...
}
```
here we convert String value to bytes using system dependent encoding ... and later we convert bytes back to chars using another encoding.
It would be nicer to be able to directly return String or char[] from the `SpoonFile` or at least from `VirtualFile`.

... but how to mix it together with encoding detector?? ;-)

@surli @monperrus  Do you agree to refactor the related file API here?

pvojtechovsky:
```java
interface SpoonFile {
  default char[] getContentChars(Environment env) {
    ... move loading code from `FileCompilerConfig#initializeCompiler` here...
  }
}
class VirtualFile implements SpoonFile {
...
	@Override
	public char[] getContentChars(Environment env) {
		return content.toCharArray();
	}
...
}
```
WDYT?

pvojtechovsky:
The `File` should be replaced by `SpoonFile`, because there is no `File` in case of accessing of jar content.

And I personally would prefer new interface `FileEncodingProvider` or `FileEncodingProvider` or ... instead of BiFunction.

Egor18:
So, as far as I understand you want me to add `char[] getContentChars(Environment env)` method to the `SpoonFile` interface and then implement it in `VirtualFile`, `ZipFile` and `FileSystemFile`. Then we could use `getContentChars` instead of `getContent` in `initializeCompiler` like this:

```
for (SpoonFile f : getFiles(compiler)) {

    if (compiler.filesToBeIgnored.contains(f.getPath())) {
        continue;
    }

    String fName = f.isActualFile() ? f.getPath() : f.getName();
    Environment env = jdtCompiler.getEnvironment();
    cuList.add(new CompilationUnit(f.getContentChars(env), fName, f.getDetectedEncoding(env)));
}
```
Am I right?

As you can see `CompilationUnit` constructor takes encoding as an argument too, so I have to add something like `Charset getDetectedEncoding(env)` to the `SpoonFile` interface as well. So detectedEncoding would memorize the encoding, which was detected during the last callbak call. 

WDYT?

pvojtechovsky:
> and then implement it in VirtualFile, ZipFile and FileSystemFile

if you add "default" implementation in `SpoonFile`, then it is enough to implement special code into `VirtualFile` only.

> am I right

yes

> CompilationUnit constructor takes encoding as an argument too

I believe that this encoding is ignored (The compilation unit has no access to origin byte[] so encoding is useless.), so we can simply provide default encoding from environment.

Egor18:
@pvojtechovsky, I refactored the API as you suggested.
Note: I had to make a small change in `FileSystemFolderTest.java`.

src/main/java/spoon/compiler/SpoonFile.java
```diff
@@ -41,4 +45,25 @@
...
+
+	/**
+	 * Gets the file content as a char array, considering encoding or encoding
+	 * provider.
+	 */
+	default char[] getContentChars(Environment env) {
+		byte[] bytes;
+		try {
+			InputStream contentStream = getContent();
+			bytes = new byte[contentStream.available()];
```
pvojtechovsky: There is no way how to get size of the input stream.
Use `ByteArrayOutputStream`, then IOUtils.copy to copy input to output and then `ByteArrayOutputStream#toByteArray`

src/main/java/spoon/support/compiler/VirtualFile.java
```diff
@@ -44,6 +46,17 @@ public InputStream getContent() {
...
+	@Override
+	public char[] getContentChars(Environment env) {
+		byte[] bytes = content.getBytes();
+		if (env.getEncodingProvider() == null) {
+			return new String(bytes, env.getEncoding()).toCharArray();
+		} else {
+			Charset encoding = env.getEncodingProvider().detectEncoding(this, bytes);
+			return new String(bytes, encoding).toCharArray();
+		}
+	}
```
pvojtechovsky: `content.getBytes()` was an old bug in Spoon, E.g. if content contains special characters and then you use getBytes, which encodes bytes using system (potentially wrong) encoding, then information is lost.
The correct implementation is simply
```java
char[] getContentChars(Environment env) {
   return content.toCharArray();
}
```
because it makes no sense to ask for `byte array encoding`, if we already have `char array` stored in the `String content`.

src/main/java/spoon/compiler/SpoonFile.java
```diff
@@ -54,8 +56,9 @@
 		byte[] bytes;
 		try {
 			InputStream contentStream = getContent();
```
pvojtechovsky: The contentStream must be closed. The best is:
```java
try (InputStream contentStream = getContent()) {
```

Egor18: Ooops, my bad, thanks.

pvojtechovsky:
Looks got to me. Thank You Egor!
@monperrus will you merge?

src/test/java/spoon/test/compilationunit/TestCompilationUnit.java
monperrus: If you don't read Cyrillic, it means "Hello World" :)

PR #2789, opened 2018-11-27
fix(FieldAccessTest): adapt FieldAccessTest to oracle jdk 11/jdt changes

nharrand:
Related to adding a travis build with jdk 11 See #2782 

`CtFieldWrite.getTarget()` on `array.length` used to return a `CtVariableWrite` for some reasons... But with jdk 11 now returns  a `CtVariableRead`...

`FieldAccessTest#testFieldAccessOnUnknownType()` has been changed to handle both... by expecting a `CtVariableAccess`.

This fixes tests for jdk 11 as can be seen here ( https://travis-ci.org/INRIA/spoon/jobs/460289160 ).
Wether this is satisfaying or not, is up to debate. WDYT?

### DIFF
--- a/src/test/java/spoon/test/fieldaccesses/FieldAccessTest.java
+++ b/src/test/java/spoon/test/fieldaccesses/FieldAccessTest.java
@@ -32,6 +32,7 @@ import spoon.reflect.code.CtTypeAccess;
 import spoon.reflect.code.CtUnaryOperator;
 import spoon.reflect.code.CtVariableWrite;
 import spoon.reflect.code.UnaryOperatorKind;
+import spoon.reflect.code.CtVariableAccess;
 import spoon.reflect.declaration.CtClass;
 import spoon.reflect.declaration.CtField;
 import spoon.reflect.declaration.CtMethod;
@@ -388,7 +389,7 @@ public class FieldAccessTest {
 			@Override
 			public <T> void visitCtFieldWrite(CtFieldWrite<T> fieldWrite) {
 				visited++;
-				assertEquals("array", ((CtVariableWrite) fieldWrite.getTarget()).getVariable().getSimpleName());
+				assertEquals("array", ((CtVariableAccess) fieldWrite.getTarget()).getVariable().getSimpleName());
 				assertEquals("length", fieldWrite.getVariable().getSimpleName());
 			}
 		}


surli:
So apparently this test is checking the following file: https://github.com/INRIA/spoon/blob/master/src/test/resources/noclasspath/FieldAccessRes.java

But for me the following code looks really wrong:
```
A [] array = new A[10];
array.length = 5;
```

since I don't think we are allowed to change the length of an array... IMO it worth it to change the test to do a more classical check, such as:
```
A a = new A();
a.length = 5;
```

nharrand:
For some reason `a.length` fieldWrite.getTarget() still returns a `CtVariableReadImpl` and not `CtVariableWriteImpl`...

surli:
> For some reason a.length fieldWrite.getTarget() still returns a CtVariableReadImpl and not CtVariableWriteImpl...

That's actually interesting. Could you try by upgrading jdt to 3.15.0: this is the last version on Maven and it has been released for Java 11, maybe it will fix it.

nharrand:
With:
```java
A a = new A();
a.length = 5;
```

we have:
```
CtFieldWriteImpl "a.length
└─ target CtVariableReadImpl "a"
└─ variable CtFieldReferenceImpl "l"
```

nharrand:
Migrating to jdt 3.15.0 does solve the issue. But `CompilationUnit.setModule(CompilationUnit module)` does not exist anymore, so it breaks `JDTBatchCompiler#getCompilationUnits()`. Any ideas?

src/main/java/spoon/support/compiler/jdt/JDTBatchCompiler.java
```diff
@@ -108,7 +108,7 @@ public JDTBatchCompiler(JDTBasedSpoonCompiler jdtCompiler) {
 					} else {
 						for (Map.Entry<String, CompilationUnit> entry : pathToModCU.entrySet()) {
 							if (fileName.startsWith(entry.getKey())) { // associate CUs to module by common prefix
-								compilationUnit.setModule(entry.getValue());
+								//compilationUnit.setModule(entry.getValue());
```
nharrand: This is the broken api call.

monperrus:
> Any ideas? @monperrus I see that you wrote the line in question.

No, sorry :thought_balloon:

pvojtechovsky:
@nharrand I pushed the potential solution for 3.15.0 compatibility problem

nharrand:
This should work, thanks to @pvojtechovsky !
Ready for review.

src/main/java/spoon/support/compiler/jdt/JDTBatchCompiler.java
```diff
@@ -94,21 +94,21 @@ public JDTBatchCompiler(JDTBasedSpoonCompiler jdtCompiler) {
...
-							pathToModCU.put(String.valueOf(modulePath), compilationUnit);
 
 							lastSlash = CharOperation.lastIndexOf(File.separatorChar, modulePath);
 							if (lastSlash == -1) {
 								lastSlash = 0;
 							} else {
 								lastSlash += 1;
 							}
-
+							//TODO the module name parsed by JDK compiler is in `this.modNames`
```
surli: Why keeping this as a todo? If I understand well it simplifies the work for getting the name, no?

pvojtechovsky: but we need to found correct record. It is possible, but it needs few minutes more development 
effort ;-)

surli: Oh I thought you already identified it properly. So ok leave it like that for now.

PR #2795, opened 2018-11-29
fix: check style doesn't check left curly on lambda

pvojtechovsky:
I installed latest checkstyle plugin and I get this checkstyle error
image

I guess this PR is correct solution for that. The `LAMBDA` was removed from the list of tokens.

WDYT?

### DIFF
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -53,7 +53,9 @@
 
     <!-- Checks for blocks. You know, those {}'s         -->
     <!-- See http://checkstyle.sf.net/config_blocks.html -->
-    <module name="LeftCurly" />
+    <module name="LeftCurly" >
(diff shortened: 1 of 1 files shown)


surli:
There's apparently an error in your config:

> [ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:2.17:checkstyle (default) on project spoon-core: An error has occurred in Checkstyle report generation. Failed during checkstyle configuration: cannot initialize module TreeWalker - Token "OBJBLOCK" was not found in Acceptable tokens list in check com.puppycrawl.tools.checkstyle.checks.blocks.LeftCurlyCheck -> [Help 1]

or maybe we need to update the checkstyle plugin in our pom

surli:
Apparently the last version is 3.0.0 (https://mvnrepository.com/artifact/org.apache.maven.plugins/maven-checkstyle-plugin/3.0.0). Would you mind to update the version in the pom to check if it works?

PR #2802, opened 2018-11-30
feat: add support for javadoc inline tags

monperrus:
eg `{@link Exception}`


monperrus:
To me it's finished too.

The next step will be to update the name in `{@link}` when in fully-qualified mode.

PR #2803, opened 2018-11-30
feature: AllTypeMembersFunction mode=[All, SKIP_PRIVATE]

pvojtechovsky:
AllTypeMembersFunction can now return
A) all type members
B) only these type members which are visible in input type. It means inherited private type members are not returned


monperrus:
LGTM.

I'm not sure about the name `INERNALLY_ACCESSIBLE` (typo?)

According to the doc `Returns only type members, which are accessible from the input type. It means that inherited private type members are skipped`, 

what about `SKIP_PRIVATE_TYPE`?

pvojtechovsky:
yes, there is type. I will fix it

pvojtechovsky:
> SKIP_PRIVATE_TYPE

it is not about types. but about type members.

I renamed to `SKIP_PRIVATE`.

WDYT?

PR #2805, opened 2018-11-30
feature: new property CtTypeReference#implicitParent

pvojtechovsky:
During refactoring of import handling #2683, I often needed to know whether type reference 
A) shows only simple name and it's parent is implicit
B) shows fully qualified name (it's parent is not implicit)
Therefore I found useful these two new API methods.

WDYT?

### DIFF
--- a/src/main/java/spoon/reflect/reference/CtTypeReference.java
+++ b/src/main/java/spoon/reflect/reference/CtTypeReference.java
@@ -213,5 +213,17 @@ public interface CtTypeReference<T> extends CtReference, CtActualTypeContainer, 
 	@DerivedProperty
 	CtTypeParameter getTypeParameterDeclaration();
 
+	/**
+	 * @param parentIsImplicit false then fully qualified name is printed.
+	 * 		true then type simple name is printed.
+	 */
+	@DerivedProperty
+	CtTypeReference<T> setImplicitParent(boolean parentIsImplicit);
 
+	/**
+	 * @return false then fully qualified name is printed.
+	 * 		true then type simple name is printed.
+	 */
+	@DerivedProperty
+	boolean isImplicitParent();
 }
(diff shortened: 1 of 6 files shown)


monperrus:
LGTM.

I would propose to rename to `isParentImplicit` which would be a perfect English question.

pvojtechovsky:
I am ok with any rename. The main question is whether the word `parent` should be there at all...
Note that implicit value doesn't come from parent, but from package or declaring type ...
So may be names like
* isFullyQualified - is near, but not 100% correct
* isQualified
* isJustSimpleName
* ....

I really do not know the good name here...

pvojtechovsky:
may be `implicitScope`, `implicitQualifier`, ... ?

monperrus:
What's the intent there?

1. that the typeref itself contains enough type information to be fully qualified? Then `canBeFullyQualified` 
1. that the fully-qualified name is derived from the parent? Then `isFullyQualifiedFromParent` 
1. that the fully-qualified name is derived from something else that is not `this`? Then `isFullyQualifiedDerived`
1. Something else?

pvojtechovsky:
> What is the intent?

to define whether scope of type reference has to be printed or not. E.g. 
```java
CtTypeReference ref = factory.createTypeReference("java.lang.String");
ref.setImplicitParent(false); //prints: "java.lang.String"
ref.setImplicitParent(true); //prints: "String"
ref.setImplicit(true); //prints: ""
```
In the case above it is about implicitness of package reference. But there are other case like type reference: `some.package.OuterType.NestedType.LeafType`, where setImplicitParent(true), causes that it will print `LeafType` only.

pvojtechovsky:
This PR needs to be finished before most of other PRs. Any idea about good name for that API method?

monperrus:
According to your last explanation, the current name is OK.

I'm OK to merge. Ready?

pvojtechovsky:
Yes, it is ready for merge

PR #2809, opened 2018-12-01
WiP fix: JDTCommentBuilder handles compilationUnit, imports, pckgDecl

pvojtechovsky:


pvojtechovsky:
This is also hard to extract from #2683, because it would need more changes, which would be then rolled back by #2683...
I am closing this PR and the code will stay as part of #2683

PR #2812, opened 2018-12-01
WiP fix: Spoon model CtTypeReference implicit and implicitParent are correct

pvojtechovsky:


pvojtechovsky:
Extracting of this code from #2683 is quite complicated, because it influences many test cases, which would have to be changed in wrong way to pass this PR then would have to be changed again in correct way because of changes in #2683.
So I am closing this PR and I would prefer to keep these changes as part of #2683

PR #2820, opened 2018-12-04
fix: Refactoring#copyType()

pvojtechovsky:
When the model is more complicated then the origin code failed, because the checking of the model consistency in the middle of renaming cannot work.
So this PR first changes names and then checks that everything is consistent


pvojtechovsky:
I do not understand why it passed before,  but as you can see today evening on failing #2683 http://spoon3r.lille.inria.fr/github-webhook//traces/tmp0hpdnorq , the problem is reproducible with code of #2683.

This PR is ready for review.

Note: after #2820, #2819 and #2813 is merged, I can rebase #2683 and we can start to discuss that PR. I do not see any simple way now, how to make it even smaller.

monperrus:
I love the meta-touch of refactoring of refactoring.

PR #2825, opened 2018-12-08
fix(position): position of type modifier after comment

pvojtechovsky:
```java
/* word public confused position builder */
public class A {}
```


monperrus:
Thanks Pavel

PR #2827, opened 2018-12-08
feat: add module for control-flow analysis

monperrus:
original code by @marcelinorc

### DIFF
--- a/spoon-control-flow/pom.xml
+++ b/spoon-control-flow/pom.xml
@@ -91,12 +91,6 @@
         </dependency>
 
         <dependency>
-            <groupId>log4j</groupId>
-            <artifactId>log4j</artifactId>
-            <version>1.2.17</version>
-        </dependency>
-
-        <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
             <version>4.11</version>
--- a/spoon-control-flow/pom.xml
+++ b/spoon-control-flow/pom.xml
@@ -94,6 +94,7 @@
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
             <version>4.11</version>
+            <scope>test</scope>
         </dependency>
     </dependencies>
 
--- a/spoon-control-flow/pom.xml
+++ b/spoon-control-flow/pom.xml
@@ -33,6 +33,25 @@
                     <target>1.8</target>
                 </configuration>
             </plugin>
+
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-checkstyle-plugin</artifactId>
+                <version>3.0.0</version>
+                <configuration>
+                    <failsOnError>true</failsOnError>
+                    <configLocation>../checkstyle.xml</configLocation>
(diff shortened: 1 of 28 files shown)


pvojtechovsky:
Cool! 🥇

pvojtechovsky:
Looks like a nice Christmas present :-)

surli:
@monperrus thanks for adding this one! Still I have a question regarding the license: the files you added do not have the license header. Does it mean they are in another license? What about the packaging of this module? What does it mean for the releases of Spoon? I really think we should be consistent on that subject: we already have a not-well-known license so let's avoid more ambiguities :)

nharrand:
@surli I think in their [original repo](https://github.com/marcelinorc/spooncontrolflow) they had no header, and no licenses at all. As far as I understand, that means that we can do anything we want with them. 

So I agree we should put as header the license we want for spoon sub modules.

surli:
> As far as I understand, that means that we can do anything we want with them.

We should have at least the agreement from @marcelinorc here and to put the original repository in the header then.

nharrand:
Ok. I will write him an eamil then.

monperrus:
> By default, the source code has the same copyright than books.
Interesting. Do you have an authoritative URL about this?

nharrand:
Not necessarily authoritative: https://choosealicense.com/no-permission/

surli:
https://en.wikipedia.org/wiki/License-free_software

> he software owner has not made explicit the terms of the license which makes the software fully copyright protected according to the Berne convention.

monperrus:
https://en.wikipedia.org/wiki/Berne_Convention

nharrand:
Marcelino has kindly accepted to put a MIT license on his project [as can be seen here](https://github.com/marcelinorc/spooncontrolflow/blob/master/README.md) so I think we can proceed with merging once we have updated the headers of these files.

pvojtechovsky:
Thank you @marcelinorc !

nharrand:
I think this should be ready for review.

spoon-control-flow/pom.xml
```diff
@@ -0,0 +1,106 @@
...
+            <version>0.9.1</version>
+        </dependency>
+
+        <dependency>
+            <groupId>log4j</groupId>
+            <artifactId>log4j</artifactId>
+            <version>1.2.17</version>
+        </dependency>
+
+        <dependency>
```
surli: junit, only for test no? And I thought @monperrus did a PR for bringing junit in the parent pom?

surli:
Is it possible to add a Readme.md?

spoon-control-flow/pom.xml
nharrand: The PR in question is not merge-able yet #2828 if I am not mistaken.

nharrand: But as soon as it is possible, I agree that we should move junit in parent.

nharrand:
Yes definitely!

monperrus:
Thanks a lot @nharrand for the last mile

monperrus:
And kudos to @marcelinorc for the foundational work.

PR #2828, opened 2018-12-08
chore: JUnit dependency is in parent pom

monperrus:


nharrand:
Don't you need to redeploy parent pom to make this change works?

monperrus:
Fixed! Ready for merge!

PR #2829, opened 2018-12-08
chore: clean old Travis scripts

monperrus:
cleaning under the carpet of #2782


surli:
Just to be sure: do we agree that spoon-decompiler is built/tested in the other jobs @nharrand? 
Regarding the job with another seed might be great to keep it in some Jenkins job.

nharrand:
> Just to be sure: do we agree that spoon-decompiler is built/tested in the other jobs @nharrand?
`spoon-decompiler` is indeed currently built/tested in `chore/travis/travis-verify.sh`

Issue #2858, opened 2019-01-26
StackOverflowError in CtAnnotation#getValueOfObject


seintur:
CtAnnotation#getValueOfObject fails with a StackOverflowError when the value of the annotation parameter is an operation that references a constant that is not in the model built by Spoon.

The sample project https://github.com/seintur/spoon-annotation-getValueAsObject reproduces the bug.

monperrus:
Hi Lionel,
Could you create a pull-request with a failing test case?
This would be super useful for fast fixing :-)
Thanks.

monperrus:
closed by #2866

PR #2866, opened 2019-02-01
fix: refactor code->runtime methods so that CtAnnotation#getValueOfObject does not overflow

seintur:
Failing test case for CtAnnotationImpl#getValueAsObject for https://github.com/INRIA/spoon/issues/2858

Don't have a fix yet.


monperrus:
Thanks Lionel. I'm looking at it.

monperrus:
the reason is that the partial evaluator does not know how to evaluate `\"a+\"+Integer.SIZE`

What's the best behavior to your opinion?

- null
- Object()
- exception?

seintur:
I guess an exception that would mention that the evaluation is unsupported would be best.
Thanks.

monperrus:
@oscarlvp there is a gift for you here :-)

src/test/java/spoon/test/annotation/AnnotationTest.java
```diff
@@ -1547,4 +1547,17 @@ public void testAnnotationArray() throws Exception {
 		assertTrue(shadowAnnotationOne.getValue("role") instanceof CtLiteral); // should be CtLiteral
 		assertEquals(annotationOne.getValue("role"), shadowAnnotationOne.getValue("role")); // should pass
 	}
+
+	@Test(expected = SpoonException.class)
+	public void testGetValueAsObject() {
+		// contract: when one cannt evaluate an expression, throw an exception
```
seintur: cannt -> cannot

monperrus: Thanks, but already merged. Could you do a live pull-request with web-based editing?

PR #2882, opened 2019-02-17
feat: check if a CtType is an array type

monperrus:
I'm not satisfied by this answer. Do you see another working solution?

@pvojtechovsky would you have a look at the bug in isSubTypeOf?

### DIFF
--- a/src/main/java/spoon/reflect/declaration/CtTypeInformation.java
+++ b/src/main/java/spoon/reflect/declaration/CtTypeInformation.java
@@ -186,4 +186,10 @@ public interface CtTypeInformation {
 	 */
 	@DerivedProperty
 	CtTypeReference<?> getTypeErasure();
+
+
+	/**
+	 * @return true if this represents an array e.g. Object[] or int[]
+	 */
+	boolean isArray();
 }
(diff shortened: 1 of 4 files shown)


pvojtechovsky:
> bug in isSubTypeOf

I will try to give it some time next week.

pvojtechovsky:
I fixed the problem CtArrayTypeReference isSubtypeOf java.lang.Array

There remains: the call of getTypeDeclaration on CtArrayTypeReference should return null or throw an exception.
The pretending that
```java
class int[] implements Serializable, Cloneble {}
```
is something real, is a bad idea from my point of view
WDYT?

monperrus:
Thanks for the fix!

> There remains: the call of getTypeDeclaration on CtArrayTypeReference should return null or throw an exception.

null would be dangerous NPEs for client code, I would not go this way. An exception is an option but may require specific handlers in client code. What about returning a shadow class of `java.lang.reflect.Array`?

pvojtechovsky:
> What about returning a shadow class of java.lang.reflect.Array or java.lang.Object?

It is an interesting idea. I would probably expect more java.lang.reflect.Array then Object.

monperrus:
yes, java.lang.reflect.Array is good and consistent with the rest.

Would you have time to give it a try?

pvojtechovsky:
I have not enough time to finish the pull requests I already started. So I have currently no time to do that one too. But this change should be quite simple, so any Spoon newcomer might do it too.

monperrus:
I'm done here

Issue #2872, opened 2019-02-08
setImplicit should be recursive


oscarlvp:
The `setImplicit` method of the `CtTypeReference` implementations should be recursive.
If not, it may affect code rendering, for example, if we have a class `A` and we want to generate a call to the generic method `m` of `A` we get

`A.<java.util.ArrayList<[]>>m()` instead of `A.<java.util.ArrayList<java.lang.Object[]>>m()`

even when we set `setImplicit(false)` for the instance of `CtTypeReference` representing `<java.util.ArrayList<java.lang.Object[]>>`

The solution is to make `setImplicit` recursive, that is, it should also affect the component type ia array references and all type arguments in generic type references.

monperrus:
Interesting point... but it may have side-effects.

could you give it a try in a pull-request?

pvojtechovsky:
It must not be always recursive. Here is the reason
```java
//code with implicit block
if (x) return;
//code with explicit block
if (x) {
 return;
}
```

oscarlvp:
I meant recursive for type references.
In the initial example, the top type reference `java.util.ArrayList<java.lang.Object[]>` was set as not implicit, however, this did not reach the component type of the array, and the pretty printer produced `java.util.ArrayList<[]>`, since the reference for the component type of the array was implicit from the beginning.

pvojtechovsky:
may be an utility method `setImplicitRecursive` would be helpful. I think it is not safe to apply recursivity automatically. But if it is limited to type references, then it is MAY BE correct. Hard to say...

surli:
I have the feeling when I read the thread that the current setter can only be problematic on a type reference when calling it with `false`, because of the pretty printer behaviour.
So maybe we should just introduce a method `removeImpliciteness()` or something with a better name which would call recursively `setImplicit(false)`.

monperrus:
@oscarlvp would you make a PR with a recursive behavior to see what breaks or not?

oscarlvp:
Sure.

PR #2897, opened 2019-03-04
WIP: setImplicit applied to arguments and component types for type references

oscarlvp:
Changes concerning #2872


monperrus:
LGTM. Could you add a test case to specify this new behavior?  Thanks!

monperrus:
ping @oscarlvp

pvojtechovsky:
I would say it is not always correct to mark all children non implicit. For example in code like:
```java
List<>
//internally there is where Object typeRef is implicit
List<Object>
```
Now if you call setImplicit(true) on that List, then all is implicit - OK
The you call setImplicit(false) on that List, then all is NOT implicit, 
```java
List<Object>
```
but what if client wants to again
```java
List<>
```
I agree with @surli that calling of setImplicit(false) with recursive algorithm is risky and some clients might expect different behavior - sometime recursive and sometime not recursive.
It seems to me like the call of setImplicit(true) can be recursive always.
WDYT?

monperrus:
closing the stalled PR, thanks Oscar for the suggestion!

PR #2903, opened 2019-03-08
Remove deprecated `mvn eclipse:eclipse` in README

mickaelistria:
`mvn eclipse:eclipse` has been deprecated for many years now. Importing a Maven project in Eclipse IDE is usually better done by the File > Import > Maven > Existing Maven project... wizard.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,14 +78,6 @@ mvn test
 
 The download information is at <http://spoon.gforge.inria.fr/>.
 
-### Eclipse IDE Setup
-
-In order to generate the Eclipse project files required for importing run the following commands from the root spoon directory (requires Maven):
-```
-mvn eclipse:clean
-mvn eclipse:eclipse
-```
-
 ### Ecosystem
 
 See <http://spoon.gforge.inria.fr/ecosystem.html>


surli:
Do you know if https://github.com/INRIA/spoon/blob/master/spoon-pom/pom.xml#L240 is related to it? If so it should be removed too I  guess.

mickaelistria:
Nope it's not related.
This block mentions the Eclipse IDE that the "copy-dependencies" step should be ignored. There can be several reasons for that: 1. The copy-depenendencies goal isn't "incremental" so it cannot be performed during Eclipse IDE build, or 2. The copy-dependencies goal isn't working well in the Eclipse IDE and is better kept out, or 3. The copy-dependencies goal is useless when inside the IDE and is better being skipped to save time, or maybe others. Depending on the reason, there are several possible actions. In an ideal work, the copy-depedencies goal can works fine under Eclipse IDE (ie conforms to recommendation in https://www.eclipse.org/m2e/documentation/m2e-making-maven-plugins-compat.html), in which case you can remove this block and you'd get good outcome in the Eclipse IDE. For other cases, it could be worth trying to improve Maven itself to have the mojo enabling https://www.eclipse.org/m2e/documentation/m2e-making-maven-plugins-compat.html (it's not that hard), or to sort out which issue causes this plugin to be disabled for m2e and report (and even fix) them where most appropriate.

monperrus:
Thanks @mickaelistria and welcome to Spoon!

PR #2906, opened 2019-03-08
feature(CtElement): add a method to get direct children of an element

### LINKED ISSUE #2902: question: how to get descendants with the same parent by Breadth-First-Search
## WANT
 get descendants with the same parent

## Implement
get something like `CtElement#descendantIterator()` to filter descendants with parent. Since `CtElement#descendantIterator()` seems to be DFS, it is not effective when the tree is too deep. **Has any approach fulfilling my need existed?**

nharrand:
fix #2902

### DIFF
--- a/doc/filter.md
+++ b/doc/filter.md
@@ -21,6 +21,12 @@ In addition, there exists a generic getter based on the role played by an elemen
 methods = ctClass.getValueByRole(CtRole.METHOD);
 ```
 
+While not recommended, it is also possible to get all direct children of an element
+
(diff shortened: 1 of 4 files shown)


monperrus:
add a test?

monperrus:
and some documentation in "AST traversal"?

Thanks!

nharrand:
Shall we rename this method `getDirectChildren()` to avoid confusion?

monperrus:
LGTM

PR #2907, opened 2019-03-08
fix(ContextBuilder): catch PositionBuilder exception and log them instead of crashing

nharrand:
fix#2888

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/ContextBuilder.java
+++ b/src/main/java/spoon/support/compiler/jdt/ContextBuilder.java
@@ -24,12 +24,14 @@ import org.eclipse.jdt.internal.compiler.ast.TypeReference;
 import org.eclipse.jdt.internal.compiler.lookup.FieldBinding;
 import org.eclipse.jdt.internal.compiler.lookup.ReferenceBinding;
 
+import spoon.Launcher;
 import spoon.compiler.Environment;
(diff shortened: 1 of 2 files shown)


monperrus:
do we want to accept incorrect code?

the other solution would be to crash with an exception with a clear message.

WDYT?

nharrand:
I agree that we do not necessarily want to accept silently incorrect code. 
But I don't think the place to catch it is in the PositionBuilder anyway. I would argue that the position builder should be best effort, and if it contains a bug it should not break the model build.

monperrus:
I agree.

nharrand:
Ok, after some thougths, I think that yes we do want to accept incorrect code inthis case.

In JDTBatchCompiler:168:

```java
if (jdtCompiler.getEnvironment().getNoClasspath()) {
	// in no classpath, we should proceed on error,
	// as we will encounter some
	errorHandlingPolicy = new IErrorHandlingPolicy() {
		@Override
		public boolean proceedOnErrors() {
			return true;
		}
	[...]
	};
}
```

In no classpath mode, we specifically say that the compiler should proceed on errors. otherwise it would fail on the first import outside of the project. I am not sure if there is a way to cherry pick errors that we want to tolerates and errors that we don't.

monperrus:
Thanks

PR #2912, opened 2019-03-09
community: update the list of contributors

monperrus:


nharrand:
I don't get how the update makes names disappear. Shouldn't it just add new ones? (with possible reordering)

surli:
yep it looks quite wrong to remove @martinezmatias or @barais from the contributors...

monperrus:
Fixed for @martinezmatias 

But for @barais, no way to find a single commit from him in spoon-core, it seems that his reputation is so wide that he gets credits beyond his real commits :-)

(seriously, Olivier's key contribution was for SpoonEmf, which I would mention in the ecosystem paper) 

@barais?

PR #2930, opened 2019-04-04
test: clean duplicate test to avoid confusion

monperrus:
This test is a clone of a test is spoon-core: https://github.com/INRIA/spoon/blob/master/src/test/java/spoon/test/architecture/SpoonArchitectureEnforcerTest.java


monperrus:
`spoon-decompiler/src/test/java/spoon/test/architecture/SpoonArchitectureEnforcerTest.java` is a clone of `src/test/java/spoon/test/architecture/SpoonArchitectureEnforcerTest.java`

This PR deletes the former (and not the one from #2929)

surli:
OK I haven't checked the module part. Right then.

nharrand:
It is indeed a clone, but id does not run the test on the same module. I might did something wrong here, but the point is to enforce package declaration in spoon-bytecode in the same way it is enforce in spoon.
Correct me if I am wrong but Line 29 does not apply to `./src/main/java` but to `spoon-bytecode/src/main/java` as test should not be executed from the same directory.

surli:
It makes sense indeed in that case.
@nharrand would you add a comment on the class of spoon-decompiler/SpoonArchitectureEnforcer test so we don't have any doubt that this it exists on purpose there too.
Now what we could do is using Spoon inside our Maven build with some custom enforcer. So Spoon enforcing Spoon rules ;)

monperrus:
I would also rename it "DecompilerArchitectureTest"

nharrand:
Renamed and commented!

PR #2932, opened 2019-04-05
Doc: fix links in doc page "Examples of Spoon Usages"

gluckzhang:
* In the title, `usages` should be `Usages` according to other pages
* Bound example points to the wrong page
* Some other examples' links are not to the root folder, which seems a bit confusing


nharrand:
Thanks @gluckzhang .

Issue #2927, opened 2019-04-03
Problem modifying file code with static import.


LuanPereiraLima:
When I modify the file (I only modify the parts that have TRY CATCH). Everything that is modified with static (example: import static org.fusesource.leveldbjni.JniDBFactory.asString;), are removed from the code, so when I am going to compile the code again I have not found variable errors, which are exactly these static imports .

I have here the two files for comparison.
I will show only the imports.

ORIGINAL CODE: 

```
package org.apache.hadoop.mapred;

import static org.fusesource.leveldbjni.JniDBFactory.asString;
import static org.fusesource.leveldbjni.JniDBFactory.bytes;
import static org.jboss.netty.buffer.ChannelBuffers.wrappedBuffer;
import static org.jboss.netty.handler.codec.http.HttpHeaders.Names.CONTENT_TYPE;
import static org.jboss.netty.handler.codec.http.HttpMethod.GET;
import static org.jboss.netty.handler.codec.http.HttpResponseStatus.BAD_REQUEST;
import static org.jboss.netty.handler.codec.http.HttpResponseStatus.FORBIDDEN;
import static org.jboss.netty.handler.codec.http.HttpResponseStatus.INTERNAL_SERVER_ERROR;
import static org.jboss.netty.handler.codec.http.HttpResponseStatus.METHOD_NOT_ALLOWED;
import static org.jboss.netty.handler.codec.http.HttpResponseStatus.NOT_FOUND;
import static org.jboss.netty.handler.codec.http.HttpResponseStatus.OK;
import static org.jboss.netty.handler.codec.http.HttpResponseStatus.UNAUTHORIZED;
import static org.jboss.netty.handler.codec.http.HttpVersion.HTTP_1_1;

import java.io.File;
import java.io.FileNotFoundException;
import java.io.IOException;
import java.io.RandomAccessFile;
import java.net.InetSocketAddress;
import java.net.URL;
import java.nio.ByteBuffer;
import java.nio.channels.ClosedChannelException;
import java.util.ArrayList;
import java.util.Collections;
import java.util.HashMap;
import java.util.List;
import java.util.Map;
import java.util.concurrent.ConcurrentHashMap;
import java.util.concurrent.ExecutionException;
import java.util.concurrent.ThreadFactory;
import java.util.concurrent.TimeUnit;
import java.util.concurrent.atomic.AtomicInteger;
import java.util.regex.Pattern;

import javax.crypto.SecretKey;

[...]

Issue #2933, opened 2019-04-05
feat: CtImportKind.Unresolved


nharrand:
I feel like this sort of problem have been raised before, so pardon me if this is just silly.

Currently when in noClassPath mode and with AutoImport, we remove all imports that we cannot resolved (`JDTImportBuilder#build()`) . Would it be possible to just create Unresolved imports (containing the original string) that could be printed later?

I know that either setting a classpath, or not using autoimport or using the SniperPrettyPrinter more or less solves this kind of issue, but I feel it still could be useful.

But maybe it creates more problem down the line, that I do not see right now.

A simpler option could be at least to log the fact that some import are simply ignored, to inform the user of other solutions.

WDYT?

monperrus:
> Would it be possible to just create Unresolved imports (containing the original string) that could
> be printed later?
In noClassPatch, that's an excellent idea.

monperrus:
done in #2936

PR #2936, opened 2019-04-08
feat(CtUnresolvedImport): keep all imports in noClassPath mode

nharrand:
Related to #2927 .


nharrand:
Ready for reviews!

spoon-pom/pom.xml
```diff
@@ -274,12 +274,14 @@
                     <configuration>
                         <header>LICENSE-short.txt</header>
                         <includes>
-                            <include>src/main/java/**</include>
```
monperrus: that results in deactivating license check (basedir is not src)

src/main/java/spoon/experimental/CtUnresolvedImport.java
```diff
@@ -0,0 +1,81 @@
+/**
```
monperrus: should be the new dual license

src/main/java/spoon/reflect/declaration/CtImportKind.java
```diff
@@ -10,5 +10,6 @@
 	ALL_TYPES,
 	ALL_STATIC_MEMBERS,
 	FIELD,
-	METHOD
+	METHOD,
+	UNRESOLVED
```
monperrus: some documentation of the meaning of the new enum value?

src/main/java/spoon/reflect/factory/CoreFactory.java
```diff
@@ -529,6 +529,11 @@ BodyHolderSourcePosition createBodyHolderSourcePosition(
 	 */
 	CtImport createImport();
 
+	/**
+	 * Creates an unresolved import.
+	 */
+	CtImport createUnresolvedImport();
```
monperrus: what does this mean for the end user?

src/test/java/spoon/test/jdtimportbuilder/ImportBuilderTest.java
```diff
@@ -103,7 +105,8 @@ public void testInternalImportWhenNoClasspath() {
 
 		CtClass classA = spoon.getFactory().Class().get("it.feio.android.omninotes.models.Attachment");
 		CompilationUnit unitA = spoon.getFactory().CompilationUnit().getMap().get(classA.getPosition().getFile().getPath());
-		assertTrue(unitA.getImports().isEmpty());
+		assertTrue(unitA.getImports().stream().filter(i -> !(i instanceof CtUnresolvedImport)).collect(Collectors.toList()).isEmpty());
+		assertEquals(1, unitA.getImports().size());
```
monperrus: assert in the import value?

monperrus:
cool stuff, thanks! see my comments.

spoon-pom/pom.xml
nharrand: Okay! Now I understand why I struggled so much... :-( My local copy was out of sync (I had basedir=src).

nharrand:
Related to #2933

nharrand:
The current implementation should work BUT it might lead to overlapping import when pretty-printing.
An example is `src/test/java/spoon/test/prettyprinter/testclasses/Rule.java`.
When imports are computed, 
```
import java.util.Arrays;
import java.util.Collections;
import java.util.Comparator;
import java.util.HashSet;
import java.util.List;
import java.util.Set;
```
are added (as these types are indeed used) but the import overlap with the pre existing CtUnresolvedImport `import java.util.*;`...

nharrand:
OK, so now (with this PR), in noClassPath mode:
 1. Unresolved import in the original sources will remain when the model is pretty printed.
 2. But that means that import computation will never remove them even if they are unused.

This mean, that if an `import my.package.*` is unresolved we don't get the nice clean up effect that we use to have.

monperrus:
Equals is not tested https://coveralls.io/builds/23377168/source?filename=src%2Fmain%2Fjava%2Fspoon%2Fexperimental%2FCtUnresolvedImport.java#L73

This is dangerous, could you add one assertTrue and one assertFalse?

monperrus:
Great feature for the noClasspath mode. Thanks!

PR #2939, opened 2019-04-09
chore: switch to junit5

monperrus:


surli:
@monperrus do you plan to convert more tests than ProcessingTest or was that just a test to check that the new jupiter assert works as well as the old ones?

monperrus:
This was the only test to convert (required otherwise it does not compile).

ready for merge.

PR #2940, opened 2019-04-10
chore: exclude GUI from coverage

monperrus:


surli:
What's the rationale behind that change?

monperrus:
increase the coverage to reach 90% while not doing Swing GUI testing in Travis.

surli:
Makes sense. Would you add it in the comment on the exception then?

monperrus:
done

PR #2943, opened 2019-04-10
fix: test and fix untested method createNewClass()

monperrus:


src/main/java/spoon/reflect/factory/CodeFactory.java
```diff
@@ -202,21 +204,21 @@ public CodeFactory(Factory factory) {
...
-	 *
-	 * @param type the decelerating type of the constructor.
-	 * @param anonymousClass Anonymous class in the new class.
-	 * @param parameters the arguments of the constructor call.
-	 * @param <T> the actual type of the decelerating type of the constructor if available/
-	 * @return the new class.
+	 * see {@link #createNewClass(CtClass, CtExpression[])}
 	 */
-	public <T> CtNewClass<T> createNewClass(CtTypeReference<T> type, CtClass<?> anonymousClass, CtExpression<?>...parameters) {
+	public <T> CtNewClass<T> createNewClass(CtType<T> superClass, CtExpression<?>...parameters) {
```
surli: I'm surprised that revapi is not complaining there?

monperrus:
gakoci/revapi is dead since mid january :-)

monperrus:
ready to merge

src/main/java/spoon/reflect/factory/CodeFactory.java
surli: I'm not a huge fan on this signature change: it's really a potential breaking change for users. Why don't you keep the existing one and add a new one?

monperrus: This method 1) was useless and incorrect 1) is not used anywhere (consequence of 1))

So I'd say there is no need to keep this technical debt.

Issue #2946, opened 2019-04-16
reverting on incorrect .java files


Charmik:
Hello! I am trying to parse .java files, but I have some files which were decompiled with synthetic methods. example:

```
@Retention(RetentionPolicy.RUNTIME)
@Target({ ElementType.ANNOTATION_TYPE })
public @interface ListenerClass {
    ListenerMethod[] method() default {};
    String remover() default "";

    public enum NONE {
        private static final /* synthetic */ NONE[] $VALUES;
        static {
            $VALUES = new NONE[0];
        }
        public static NONE valueOf(final String s) {
            return Enum.valueOf(NONE.class, s);
        }
        public static NONE[] values() {
            return NONE.$VALUES.clone();
        }
    }
}

```
it has synthetic enum methods, so it doesn't compile.
I try to parse it by: 
Launcher launcher = new Launcher();
launcher.addInputResource(source.toAbsolutePath().toString());
CtModel model = launcher.buildModel(); 

it gives me:
java.lang.IllegalArgumentException: The method remover() should be a spoon.reflect.declaration.CtAnnotationMethod

it breaks in CtAnnotationTypeImpl.addMethod because method has a type "CtMethodImpl", not a "CtAnnotationMethod".

but if you swap 2 lines: 
String remover() default "";
ListenerMethod[] method() default {};

it works. Can you suggest what I can do with not really good .java files? Can I just skip places which doesn't compile? I didn't find any options for it. I think it works on jdt-eclipse. Thanks!

surli:
Hi @Charmik 

I've the feeling that there is two different issues here:
  - the synthetic enum
  - the two methods that are not in the right order

Could you confirm that? By checking if it fails if you remove the synthetic enum? 

> Can you suggest what I can do with not really good .java files? Can I just skip places which doesn't compile?

I don't think we can skip compilation error, now you can build the model without the erroring files: by default, Spoon is in "no classpath mode", which means it will work even if not all sources files are given as input.

Charmik:
> Hi @Charmik
> 
> I've the feeling that there is two different issues here:
> 
> * the synthetic enum
> * the two methods that are not in the right order
> 
> Could you confirm that? By checking if it fails if you remove the synthetic enum?
> 
> > Can you suggest what I can do with not really good .java files? Can I just skip places which doesn't compile?
> 
[...]

PR #2947, opened 2019-04-18
fix: handle special cases in CtAnnotationType

Charmik:
I added tests from discussion:
https://github.com/INRIA/spoon/issues/2946

1 test passes and the second one fails with exception. I can fix tests and commit them after your fix.
I am not sure that I put them in the right place, by the way.


monperrus:
The bug comes from JDT :-(, it does not call `visit(AnnotationMethodDeclaration, ClassScope` but `visit(MethodDeclaration, ClassScope)`

Just pushed a workaround here.

WDYT?

Charmik:
> The bug comes from JDT :-(, it does not call `visit(AnnotationMethodDeclaration, ClassScope` but `visit(MethodDeclaration, ClassScope)`
> 
> Just pushed a workaround here.
> 
> WDYT?

seems okay for me!
can we merge it to master?
did you create bug in JDT repository? I think better to do it;)
your build seems failed, by the way.

Charmik:
> The bug comes from JDT :-(, it does not call `visit(AnnotationMethodDeclaration, ClassScope` but `visit(MethodDeclaration, ClassScope)`
> 
> Just pushed a workaround here.
> 
> WDYT?

by the way your condition 
`&& (type instanceof CtAnnotationType)`
doesn't make sense, because at the start there is `if` 
`!(child instanceof CtTypeParameter)` and `return`

so this condition is always true, so you can remove it.

monperrus:
The previous implementation and fix made too many assumptions on the type of methods.

Here is a better fix for which all test cases pass.

@nharrand WDYT? Merge?

PR #2949, opened 2019-04-20
fix: permissions of files in doc folder

alshopov:
Some of files should be executable (*.py, *.sh)
Rest should not, incl. *.js that are meant for browser


monperrus:
thanks, doc pull-requests are very welcome.

PR #2950, opened 2019-04-21
governance: adopt dual licensing MIT-CECILL-C

### LINKED ISSUE #2898: License: change license to double-license Cecill-C/MIT
It's been years that we're aware that the Cecill-C license hinders the development of Spoon (see #321, #1861, #1930, #2713).

And finally, @ftomassetti hits again in [his great post](https://tomassetti.me/analyze-generate-and-transform-java-code/) (and he's right to do so, thanks for it :-).

I've consulted Inria, and it seems we have a solution, which is acceptable and backward compatible: double-licensing.

__We would change the license from Cecill-C to a dual-license Cecill-C/MIT__ 

We're commenting on this issue before the end of this month in order to reach a consensus.

monperrus:
fix #2898


tdurieux:
Legally all the people that contributed to spoon have to accept this licence change.
Technically this PR is violating copyright laws (at least the European one).

I really don't think it will be a problem for this project but for for your other projects it can be.

nharrand:
How do you propose to proceed?

PR #2952, opened 2019-04-29
fix(JarLauncher): fix #2951, Really delete temporary decompiled source directory

### LINKED ISSUE #2951: JarLauncher always use the same tmp directory
The JarLauncher in the spoon decompiler always uses the same tmp directory, no matter if the jar changes. This result in having classes inside the model that were not actually inside the jar, just because the decompiled source code is still there from the previous Jar.

The line that gives the problem is line 92 in JarLauncher.java:
`decompiledSrcPath = System.getProperty("java.io.tmpdir") + System.getProperty("file.separator") + "spoon-tmp";`

I suggest to create a new random directory every time and delete it at the end, or just delete this one every time.
Thanks

nharrand:


monperrus:
do we need a test case to specify this behavior?

nharrand:
You mean to check that the silly developer knows how to delete a directory? Well before the issue I would have said no...

monperrus:
:-)

nharrand:
But all jokes aside, we could have a test that check that we can perform two analysis in a row.

nharrand:
By the way, there might be something wrong with travis scripts for jdk-10 and jdk-11, downloading seems to fail...

nharrand:
[This issue](https://github.com/sormuras/bach/issues/51) seems to be related.

monperrus:
Thnkas @nharrand

Issue #2955, opened 2019-05-03
Check if CtPackage contains a package-info.java


msteinbeck:
Is there a method in CtPackage which allows me to check whether a package contains a `package-info.java` file?

nharrand:
Even if it is not super clean, you can do the following:

```java
//Assuming that p is a CtPackage
((CtElementImpl) p).getPosition().contains("package-info.java");
```

When there is no `package-info.java` file, `(CtElementImpl) p).getPosition()` returns `Unknown file`, when there is one, it returns a position in `package-info.java`.

surli:
Actually you can check it from the compilation units:
```
CompilationUnit cu = element.getPosition().getCompilationUnit();
if (cu.getUnitType() == CtCompilationUnit.UNIT_TYPE.PACKAGE_DECLARATION) {
  ... 
}
```

nharrand:
I think `cu` is `null` when the `package-info.java` does not exists. (But I agree it is better.)

nharrand:
May be we could add a something like `boolean CtPackage#hasPackageInfo()` to the API.

msteinbeck:
Similar to `package-info.java`, I need to know whether a CtPackage contains a `module-info.java`. How about adding an `isEmpty` method to CtPackage?

nharrand:
Shouldn't `module-info.java` be associated with `CtModule` rather than `CtPackage`?
(Nitpicking) I don't really like the name `isEmpty` as I find it missleading, it could mean that the package does not contain any `CtType`. But regardless of the name, I agree that we should add the behavior.

msteinbeck:
> I don't really like the name isEmpty as I find it missleading, it could mean that the package does not contain any CtType

That's exactly what I need. I want to know whether a package does not contain a type, package-info.java, and module-info.java (and other info files provided by future SDKs). I can't just check if the path of a package contains any *-info.java file because they might have been excluded from build, thus being not part of a model.

msteinbeck:
Due to being a `CtNamedElement`, checking for any `CtModule` (`module-info.java`) should be feasible with `CtPackage#getElements(...)`.

nharrand:
Ok I am sold on the `isEmpty` method then.
But, I am a bit fuzzy on modules. In my understanding, a module declares packages, but packages do not contain modules. It just happen that in most build system, package tree are reproduced in terms of directory tree. Am I missing something?

nharrand:
[...]

PR #2957, opened 2019-05-03
feat(CtPackage): add two utility methods #isEmpty() and #hasPackageInfo()

nharrand:
Related to #2955

### DIFF
--- a/src/main/java/spoon/reflect/declaration/CtPackage.java
+++ b/src/main/java/spoon/reflect/declaration/CtPackage.java
@@ -149,4 +149,14 @@ public interface CtPackage extends CtNamedElement, CtShadowable {
 	 * See JLS §7.4.2. Unnamed Packages.
 	 */
 	boolean isUnnamedPackage();
+
+	/**
+	* @return true if the package contains a package-info.java file
+	*/
+	boolean hasPackageInfo();
+
+	/**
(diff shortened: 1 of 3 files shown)


src/main/java/spoon/reflect/declaration/CtPackage.java
```diff
@@ -149,4 +149,14 @@
...
 	 */
 	boolean isUnnamedPackage();
+
+	/**
+	* @return true if the package contains a package-info.java file
+	*/
+	boolean hasPackageInfo();
+
+	/**
+	 * @return true if the package contains a package-info.java and no types
```
monperrus: hard to guess the behavior from the method name. what about removing the test about package-info?

nharrand: Yes. I agree.

nharrand:
@monperrus do you merge?

monperrus:
yes, I look at and merge PRs which do not have WIP in the title :-)

is that one ready?

nharrand:
> yes, I look at and merge PRs which do not have WIP in the title :-)

Indeed that make sense :-D

> is that one ready?

Yes.

monperrus:
Thanks a lot!

Issue #2873, opened 2019-02-09
add support for type inference


monperrus:
Spoon relies on the declared static of variables and methods.

For some transformations, it would be very useful to have the actual type of a variable / expression.

For this, we could add a method `getInferredType()` in the metamodel.

Egor18:
Do you mean that you want to add data-flow analysis for type inference?
Consider the following example:

```
List list = new ArrayList();
if (condition)
{
	list = new LinkedList();
}
//getInferredType() - ?
```
How are you going to specify the type of `list` element here?

monperrus:
Yes.

To my knowledge, nobody is working on this at the moment.

Are you interested in contributing on this?

Egor18:
Implementing data-flow analysis is a big task. I think the best approach is to use some SMT solver as a backend for this.

Getting possible types statically is pretty much the same thing as getting possible values.

But even with an SMT solver, it could not be 100% accurate - we need to handle loops, virtual function calls and so on.

So I was curious how do you see data-flow analyzer in Spoon.

monperrus:
What other tools do you know that do this for Java?

Egor18:
I was quite curious, so I made a small data-flow analyzer with Spoon and Z3 SMT solver from Microsoft: https://github.com/Egor18/jdataflow
(Check out the README.md for a more detailed explanation)

Here are just several examples of what it is able to detect right now:
```java
void f(boolean c) {
    int x = 5;
    int y = c ? 3 : 2;
    if (x < y) {} // <= always false
}
```
```java
void g(int x, int y) {
    if (2 * x + 3 * y == 12) {
        if (5 * x - 2 * y == 11) {
            if (x == 3) {} // <= always true
            if (y == 2) {} // <= always true
        }
    }
}
```
```java
void h(Something x) {
    if (x == null) {
        x.f(); // <= null dereference
    }
}
```
```java
void m() {
    C a = new C();
    a.x = 10;
    C b = a;
    b.x = 20;
    if (a.x == 20) {} // <= always true
}
```
At the moment it calculates values only, but it should be relatively easy to add types too.

monperrus:
Hi Egor,

Cool to see progress here. Your README is remarkably clear, the test runner is clever and elegant, and the tests are quite comprehensive.

Overall, this is impressive.

Would you be interested in having this as a submodule in the Spoon main repo?

[...]

PR #2970, opened 2019-05-11
feature: Add support for dataflow analysis in new module spoon-dataflow

Egor18:
This PR is to add experimental spoon-dataflow module (see https://github.com/INRIA/spoon/issues/2873).
Lambdas and anonymous classes are not fully supported at the moment (see tests marked with FIXME).

P.S. The main logic is inside DataFlowScanner.java


monperrus:
Excellent! Welcome to spoon-dataflow!

Initial questions:
- could we download `com.microsoft.z3.jar` from on a repo somewhere instead of pushing binary code to Git?
- it would be great to activate CI for `spoon-dataflow`, I guess it means adding a few lines to `chore/travis/travis-verify.sh`. Could you have a look?

Egor18:
Z3 solver itself is written in C++ and the jar file basically contains only JNI wrappers. This jar file is a part of a zip release from here: https://github.com/Z3Prover/z3/releases
Unfortunately, it seems like the developers do not provide any maven repo. I see several options here:
1) We can build Z3 solver with java option ourselves (requires CMake and C++ compiler)
2) We can create some repo and put it in there
3) We can automatically download and install Z3 from the link above via gradle build script

Egor18:
I'll check out CI script.

monperrus:
I would go for solution 2 for now. Just created https://github.com/SpoonLabs/spoon-dependencies and added you as contributor.

Egor18:
Regarding the CI script, we have to have z3 on the system to run tests, so I need to add something like this:
```
cd spoon-dataflow
wget https://github.com/Z3Prover/z3/releases/download/z3-4.8.4/z3-4.8.4.d6df51951f4c-x64-ubuntu-14.04.zip
unzip z3-4.8.4.d6df51951f4c-x64-ubuntu-14.04.zip
export LD_LIBRARY_PATH=./z3-4.8.4.d6df51951f4c-x64-ubuntu-14.04/bin
./gradlew build
```
Is it ok for you?

monperrus:
yes, that's a good way of doing it.

Egor18:
spoon.MavenLauncherTest stuck, so have to reopen this PR to trigger Travis to rerun.

monperrus:
LGTM.

I propose to remove:

* binary file `spoon-dataflow/gradle/wrapper/gradle-wrapper.jar`
* unnecessary files `gradlew.bat` and `gradlew`

WDYT

monperrus:
you can also push an empty commit to trigger a new build.

Egor18:
> LGTM.
> 
> I propose to remove:
> 
> * binary file `spoon-dataflow/gradle/wrapper/gradle-wrapper.jar`
> * unnecessary files `gradlew.bat` and `gradlew`
> 
> WDYT

I believe that having gradle wrapper committed in VCS is a common practice.
In fact almost every gradle-based java project on GitHub has wrapper in the repository.
For example:
https://github.com/ReactiveX/RxJava.git
https://github.com/spring-projects/spring-framework.git
https://github.com/elastic/elasticsearch
https://github.com/jMonkeyEngine/jmonkeyengine

The whole idea of a wrapper is to be able to build project even without having gradle (with the corresponding version) installed on the machine.
Moreover, it is only 50KB.

monperrus:
> I believe that having gradle wrapper committed in VCS is a common practice.

In Spoon, we have a tradition of daring to do things beyond the common practice :-)

> The whole idea of a wrapper is to be able to build project even without having gradle (of the corresponding version) installed on the machine.

I see. Is it required for Travis?

Egor18:
I think it is not required for Travis (gradle should be preinstalled since `language: java` is in travis.yml).

By the way, even the official gradle docs say that wrapper should be in VSC:
"The scripts generated by this task are intended to be committed to your version control system. This task also generates a small gradle-wrapper.jar bootstrap JAR file and properties file which **should also be committed to your VCS**. The scripts delegates to this JAR."
https://docs.gradle.org/current/dsl/org.gradle.api.tasks.wrapper.Wrapper.html

Gradle docs also say that it is recommended to always execute a build with the Wrapper to ensure reliable, controlled and standardized execution of the build.

If we still don't want to have it in VCS, I can add an additional step to generate wrapper:
`gradle wrapper --gradle-version X.X`
then we can run build as usual:
`./gradlew build`

WDYT?

monperrus:
> |gradle wrapper --gradle-version X.X|
> then we can run build as usual:
> |./gradlew build|
|sounds perfect!|

Egor18:
Tried to remove the wrapper, but unfortunately, gradle in Travis is too old (4.0) and it seems that it does not know about Java 9 versioning yet: `Could not determine java version from '9.0.1'`.
That's an example why it is better to use a wrapper, so I suggest to keep it.

monperrus:
One option is to download the wrapper from somewhere else (as any other build dependencies that are also downloaded). I would go this way but I understand  Egor's preference for the "standard" solution.

If anybody is strongly against adding a binary JAR file in the repo, raise your voice now.

monperrus:
Thanks a lot, this is an important milestone for Spoon.

There is a lack of a good and active symbolic engine for Java, and spoon-dataflow may be the foundation for this. That's great.

PR #2972, opened 2019-05-12
fix NPE with duplicate inner class

Charmik:
This code is decompilation of the bytecode file, but as I understand parser should never throw exception when it tries to build AST. I would prefer if it would just skip the second class, if it's possible. Or maybe you prefer another way for it? Thanks.

stacktrace:
	at spoon.support.compiler.jdt.JDTTreeBuilderHelper.createType(JDTTreeBuilderHelper.java:623)
	at spoon.support.compiler.jdt.JDTTreeBuilder.visit(JDTTreeBuilder.java:1663)
	at org.eclipse.jdt.internal.compiler.ast.TypeDeclaration.traverse(TypeDeclaration.java:1485)

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilder.java
+++ b/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilder.java
@@ -1067,7 +1067,9 @@ public class JDTTreeBuilder extends ASTVisitor {
 		CtConstructor<Object> c = factory.Core().createConstructor();
 		// if the source start of the class is equals to the source start of the constructor
 		// it means that the constructor is implicit.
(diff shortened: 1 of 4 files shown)


monperrus:
I reproduced the NPE, thanks for the failing test case, it's very valuable.

Looking at it.

monperrus:
The fix was easy: protect against null.

JDT is very inconsistent, and in some cases, some fields are null. Since we have no control over this, JDTTreeBuilder is full of defensive checks.

Charmik:
> The fix was easy: protect against null.
> 
> JDT is very inconsistent, and in some cases, some fields are null. Since we have no control over this, JDTTreeBuilder is full of defensive checks.

nice! thanks! 
can I ask why you don't create PR in JDT? because it takes too much time?:)

nharrand:

PR #2984, opened 2019-05-23
fix: attach annotation to CtCatchVariable

### LINKED ISSUE #2934: Bug: Annotation on CtCatchVariable are attached to CtCatch instead
Example:

```java
try {
       testStr.length();
} catch (@MyCatchAnnotation(something =  "annotation string") NullPointerException e) {
       e.printStackTrace();
}
```

Here `CtCatch` is annotated and not `CtCatchVariable`.

Egor18:
Fix #2934.
As far as I understand, `CtCatch` gets annotated and not `CtCatchVariable` because JDT does not actually have specific nodes representing catch and represents the whole catch as `Argument` of the `Try` node, so
Spoon creates `CtCatch` from `Argument` (and at that moment it gets processed and annotated) and only after that it creates `CtVariable` itself.
Since annotations cannot be attached to `CtCatch` itself, we can simply transfer them to `CtCatchVariable`.


monperrus:
LGTM.

@nharrand you had a look at this with @gluckzhang, WDYT?

nharrand:
Oops, I messed up the diff. Sorry about that. Anyway, yes LGTM too.
I'll merge if CI is happy!

gluckzhang:
Agree with this fix!

PR #2992, opened 2019-05-26
doc: remove stackoverflow

monperrus:
Nobody in the community monitors and answers on stackoverflow, see unanswered questions on https://stackoverflow.com/tags/inria-spoon


nharrand:
I agree with the observation. Plus I like the idea that the repo is the main place where everything happen.
Shall I merge? Does anyone else have an issue with this?

PR #2993, opened 2019-05-26
Add configurable log path

### LINKED ISSUE #2672: feat: Allow to configure the path where to put the log files
Currently the path of Spoon log files is hard-coded to be put in `/tmp/spoon-log.log` (see: https://github.com/INRIA/spoon/blob/f8a96bdb648f598ce86dbf3bc5d4cd287f59531a/src/main/resources/log4j.properties#L10). 

We should allow user to configure that path.

Egor18:
fix #2672
I'd rather not to have the initialization in Launcher, but I don't see another way to do that.

### DIFF
--- a/src/main/java/spoon/Launcher.java
+++ b/src/main/java/spoon/Launcher.java
@@ -130,6 +130,11 @@ public class Launcher implements SpoonAPI {
 	}
 
 	static {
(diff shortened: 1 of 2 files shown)


monperrus:
Cool, thanks!

PR #2994, opened 2019-05-26
Fix isGenerics for top-level types

### LINKED ISSUE #2989: isGenerics always returns false for CtType
`isGenerics()` method always returns false when called on a top-level type (even if it has generic parameters):
https://github.com/INRIA/spoon/blob/151176f4fae4d472c5bad16fc07c384ec85ca207/src/main/java/spoon/support/reflect/declaration/CtTypeImpl.java#L536-L539
I think it should be something like this instead:
```
@Override
public boolean isGenerics() {
    for (CtTypeParameter ref : formalCtTypeParameters) {
        if (ref.isGenerics()) {
            return true;
        }
    }
    return false;
}

```
Here is an example:
`class A <T> { /* ... */ }`
```
CtType<?> a = model.getAllTypes().stream().filter(t -> t.getSimpleName().equals("A")).findFirst().get();
a.isGenerics() // => false (I think it should be true)
```

WDYT?

Egor18:
fix #2989

### DIFF
--- a/src/main/java/spoon/support/reflect/declaration/CtTypeImpl.java
+++ b/src/main/java/spoon/support/reflect/declaration/CtTypeImpl.java
@@ -535,6 +535,11 @@ public abstract class CtTypeImpl<T> extends CtNamedElementImpl implements CtType
 
 	@Override
 	public boolean isGenerics() {
+		for (CtTypeParameter ref : formalCtTypeParameters) {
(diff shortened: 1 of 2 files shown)


monperrus:
oops, a stupid merge conflict.

monperrus:
Thanks a lot.

PR #2997, opened 2019-05-27
fix(pom.xml): upgrade jackson-databind to 2.9.9

nharrand:


nharrand:
Oops, this is a duplicate of #2991
I guess I should read my notifications...

Issue #2985, opened 2019-05-23
Bug lambda with diamond operator and several interfaces 


lodart:
Hi,

There seems to be an error with a particular use of lambda expressions : 
```java
public interface TestLambda{
    default <T extends Ben & Jerry> Optional<T> bar () {
        return (Optional<T>) foo().stream().filter(elt -> elt.benMethod(Ben.class)).findFirst();
    }

    default <T extends Ben & Jerry> List<T> foo () {
        return Collections.emptyList();
    }
}
```
`benMethod` being a method declared in `Ben` interface.
It becomes :
```java
public interface TestLambda {
    default <T extends Ben & Jerry> Optional<T> bar() {
        return ((Optional<T>) (foo().stream().filter((Ben & Jerry elt) -> elt.benMethod(Ben.class)).findFirst()));
    }

    default <T extends Ben & Jerry> List<T> foo() {
        return Collections.emptyList();
    }
}
```
(Notice the `(Ben & Jerry elt) -> ` that doesn't compile)

Thank you for your support :)

nharrand:
Hi @lodart ,
Thanks a lot for the bug report.
Would you make a pull request with a failing test out of it?

PR #2999, opened 2019-05-29
Added a failing test case for Multi-inherited generic lambda

lodart:
Here is a failing test case for bug #2985 .
I tried to respect original code as much as possible, but feel free to tell me if I should change something.


nharrand:
Thanks for the PR.
If I understand this correctly, `elt` has actually the correct type. The problem is that it is incorrectly prettyprinted. Right?

lodart:
Exactly :)

nharrand:
Thanks a lot for the bug reproduction!
I am closing this, as it is merged in #3006 .

PR #3006, opened 2019-06-03
Fix(DefaultJavaPrettyPrinter): handle implicit type reference to intersection type

nharrand:
Integrate #2999 and fix#2985


monperrus:
thanks a lot

PR #3007, opened 2019-06-05
fix(JDTTreeBuilderHelper): executable reference return type

### LINKED ISSUE #2966: Wrong return type for method references
The CtExecutableReference interface specifies that `getType()` should return the method return type for a reference to a method, however the current implementation of the transformation from the JDT model to the spoon model is incorrect (at least, it appears so to me): [at line 585 in JDTTreeBuilderHelper.java](https://github.com/INRIA/spoon/blob/2208265abc5a13a32f83ed6e3aa4196e6e9ba342/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java#L585) the type of the newly-created ExecutableReference is set to the type declaring the method, rather than the method return type itself. 

This bug breaks, for example, the `Refactoring.changeMethodName(CtMethod)` method, since the ExecutableReference does not match with the one returned from the method itself, and overall makes matching method references to their corresponding method pretty awkward, as it has to be done manually.

nharrand:
Executable reference return type should match actual executable.
Reproduce, test and fix #2966


monperrus:
Thanks a lot!

PR #3014, opened 2019-06-08
Add literal base

### LINKED ISSUE #2396: feature: add support for verbatim number literals
Hi, can I ask how do I disable spoon's auto conversion of the numbers?

Example, when I write "short a = 0x10;", spoon will print out "short a = 16;" instead.

This affects the readability of the generated code as I want to view some numbers as hexadecimal at times.

Thank you.

Egor18:
fix #2396

This PR adds base (2, 8, 10 or 16) to the numeric CtLiterals. Pretty printer will also output literals according to their base.
Note: I'd like to set the return value of `getBase()` to `Optional<LiteralBase>` instead of nullable `LiteralBase`, but the API test does not like it since it wants to see the actual getter. I can fix it if you want because I think that `Optional` makes more sense here since only numeric literals can have a base and having null for other literals is a bit unintuitive.


src/main/java/spoon/reflect/visitor/LiteralHelper.java
```diff
@@ -16,17 +17,57 @@
...
+
+	private static String getBasedString(Float value, LiteralBase base) {
+		if (base == LiteralBase.HEXADECIMAL) {
+			return Float.toHexString(value) + "F";
+		}
+		return Float.toString(value) + "F";
+	}
+
+	private static String getBasedString(Double value, LiteralBase base) {
+		if (base == LiteralBase.HEXADECIMAL) {
```
nharrand: Shouldn't we have:
```java
		if (base == LiteralBase.HEXADECIMAL) {
			return Double.toHexString(value) + "D";
		}
		return Double.toString(value) + "D";
```

Egor18: Spoon was not adding "D" suffix before this commit for doubles, so I kept this behavior. That makes sense since it is double by default.

nharrand: Ok that makes sense indeed.

monperrus:
Well done, esp. since you had to understand all the internals when adding a new role :-)

LGTM, ok to merge.

PR #3019, opened 2019-06-13
Possible fix for auto imports (issue #2996)

### LINKED ISSUE #2996: auto import
Hello 
i'm using auto import in my configuration to save the result of my transformation in a file  
but when it come to this instruction 
View minusButton = child.findViewById(R.id.contact_remove);

it is printed like this 
View minusButton = child.findViewById(contact_remove);
and i have an error : cannot find symbol variable contact_remove

how can i solve this ? 
with setAutoImport(true)

Egor18:
fix #2996
As far as I understand from debugging, the problem is that when the binding of some unknown method in chain is null, it is not set to static by default (like for `ProblemMethodBinding`), which affects how `ImportsScanner` processes it. 
So my workaround is to set static = true here as well.


monperrus:
Thanks a lot for the fix Egor.

PR #3021, opened 2019-06-16
fix: bug when a field is accessed in a parenthesized expression

seintur:
When a field is accessed in a parenthesized expression, the CtFieldAccess contains the parentheses and not just the field name.

This PR contains only a failing test case.


monperrus:
Thanks a lot @seintur a failing test case is always very valuable. Issue in #3023.

Egor18:
Fixed by #3024.

Issue #2724, opened 2018-10-27
chore: add spoon-test to check that spoon works on any repo


monperrus:
I propose to add a script ``spoon-test` as follows:

    spoon-test https://github.com/google/guava

It would:

* check that the model can be built without any exception (contract inspired from build.sh)
* check that the pretty-printed code can be compiled with any exception (contract inspired from build.sh)
* check that the pretty-printed code passes all the test (contract inspired from build.sh)
* check all contracts on the model (see #2716)

This is basically a rework or extension of `build.sh`. We may consider a clean Python or Groovy version instead of this Shell monster.

surli:
> We may consider a clean Python or Groovy version instead of this Shell monster.

+1

Issue #2330, opened 2018-08-02
Spoon generate duplicate brackets in invocations.


nemo9cby:
After applying processing, there will be duplicate brackets inside invocations if they contain "+".

For example, if the original line is like:
`logger.info("Value declared in if:" + c);`

After parsing in Spoon, it would become:
`logger.info(("Value declared in if:" + c));`

Another example:
before:
`LOG.warn("AuthenticationToken ignored: " + ex.getMessage());`
after:
`LOG.warn(("AuthenticationToken ignored: " + (ex.getMessage())));`

The processor looks like:
```
public class CusProcessor extends AbstractProcessor<CtStatement> {
    public void process(CtStatement element) {
        if (element instanceof CtInvocationImpl) {
            CtInvocationImpl invoke = (CtInvocationImpl) element;
            System.out.println(invoke);
        }
    }
}
```

Launcher code:
```
        Launcher launcher = new Launcher();
        launcher.addInputResource(filePath);
        launcher.getEnvironment().setNoClasspath(true); 
        launcher.getEnvironment().setAutoImports(true);
        launcher.getEnvironment().setOutputType(OutputType.NO_OUTPUT);
        launcher.getEnvironment().setCommentEnabled(false);
        launcher.addProcessor(new CusProcessor());
        launcher.run();
```
My guess is that it is because of the internal compiler, is there a way to fix this?

surli:
Hi @nemo9cby thanks for reporting this one!
It was already reported by @pvojtechovsky and it should have been fixed with #2113. So you might want to try the latest snapshot of Spoon to check if it solve your problem: it will be integrated in the next release.

nemo9cby:
Hi @surli , thanks for the reply!
I am currently using 7.0.0 and it shows in the change log that the bug #2113 has been fixed in this snapshot. So by saying "next release" do you mean that 7.0.0 is not able to deal with this issue and we need to build the snapshot using the latest commit?

nemo9cby:
Actually I just checked version 7.0.0, it already contains the code changes of #2113. This issue still exists.

surli:
My bad then: I though #2113 was not yet integrated in a Spoon release. So yeah it's a new bug.

Egor18:
Resolved by #3024

PR #3024, opened 2019-06-18
fix: Remove parentheses around targeted expr in binary or unary op

Egor18:
fix #3021
I think there is no need to wrap any `CtTargetedExpression` in parentheses when it is used in binary / unary operations (precedence is higher anyway).

### DIFF
--- a/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
+++ b/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
@@ -360,7 +360,7 @@ public class DefaultJavaPrettyPrinter implements CtVisitor, PrettyPrinter {
 		}
 		try {
 			if ((e.getParent() instanceof CtBinaryOperator) || (e.getParent() instanceof CtUnaryOperator)) {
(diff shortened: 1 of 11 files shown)


tdurieux:
I think it can be required when the element is casted.

Egor18:
The case with casts is still handled:
https://github.com/INRIA/spoon/blob/fb04264db07ac7a91a014a84c58aa848aab0e673/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java#L357-L360

I added a test just in case.

monperrus:
LGTM.

we may miss a special case here, but if there is no test, we don't know. We will catch it with jenkins and upcoming #2724

Egor18:
Well, I looked through all of the possible CtTargetedExpressions: 
`CtAnnotationFieldAccess<T>, CtArrayAccess<T,E>, CtArrayRead<T>, CtArrayWrite<T>, CtConstructorCall<T>, CtExecutableReferenceExpression<T,E>, CtFieldAccess<T>, CtFieldRead<T>, CtFieldWrite<T>, CtInvocation<T>, CtNewClass<T>, CtSuperAccess<T>, CtThisAccess<T>`
and all possible binary and unary operators, and I don't see any counterexamples.

BTW, the similar fix is possible for #2330.

nharrand:
LGTM too.

PR #3028, opened 2019-06-20
fix: template matching ignores casts

### LINKED ISSUE #3026: Matcher not working in case of cast
Hi!
I have a matcher that should match all the method invocations of type `this.subFlow(...)`. In the construction of the code I have this:
```
public class MatcherContainer extends FlowLogic {

    public TemplateParameter<FlowLogic<?>> _subFlow_;

    private void subFlowMatcher() throws FlowException {
        subFlow(_subFlow_.S());
    }

    @Override
    public Object call() throws FlowException {
        return null;
    }
}
```

And from having this file in a model I compute:
```
CtMethod<?> method = (CtMethod<?>) model.getElements(
                    new NamedElementFilter(CtMethod.class, "subFlowMatcher")).get(0);
CtElement templateRoot = method.getBody().getStatement(0);
TemplateMatcher myMacther = new TemplateMatcher(templateRoot);
```

Now, if inside `CtStatement statement` I have this line:
`            SignedTransaction fullyStx = (SignedTransaction) this.subFlow(new CollectSignaturesFlow(partStx, Collections.singleton(counterpartySession)));`
Than doing
`statement.getElements(myMatcher);` 
gives an empty result.
While having the same line *without a cast* 
`SignedTransaction fullyStx = this.subFlow(new CollectSignaturesFlow(partStx, Collections.singleton(counterpartySession)));`
Correctly matches the method invocation.

Is this a bug or am I doing something wrong in the matcher?
If necessary I can create a repository to reproduce the issue.
Let me know
Thanks 
Luca

Egor18:
Fix #3026.
I'm not so familiar with template matching in Spoon yet, but it looks like a bug to me.
I think matcher `f(_x_.S());` should match `f(new X())` as well as `(Y) f(new X())`, isn't it?


monperrus:
Thanks a lot Egor for the bug reproduction.

Pavel @pvojtechovsky, we hope you're doing well. Any chance to have a look at this bug?

Egor18:
I added a possible fix.
This is probably a bit arguable solution because we skip casts during template matching now, but on the other hand we already do that with comments and some other stuff.

monperrus:
Thanks Egor, I like this solution. We just need to document it. Could you add the explanation in the corresponding Markdown file?

monperrus:
added documentation, will merge after CI.

monperrus:
thanks @Egor18

PR #3030, opened 2019-06-22
test: Failing test case when filtering with anonymous classes.

seintur:
Filtering does not visit anonymous classes. This PR contains a failing test case that illustrates the issue with a method run() defined in an anonymous class that extends Runnable in a class C. AFAIK the filter should return one method (run) but currently returns 0.


Egor18:
See https://github.com/INRIA/spoon/issues/3031#issuecomment-504706227

seintur:
ok my mistake, sorry about that. Thanks.

PR #3035, opened 2019-06-24
fix: fix Problems with javadoc parameters @exception and others

### LINKED ISSUE #3033: Javadoc Tags @serialField and @serialData
I found following problem: The real javadoc tag is @serialField and @serialData, but spoon expected @serial_Field and @serial_Data. 

You can read it here:
https://docs.oracle.com/javase/7/docs/technotes/tools/windows/javadoc.html#serialField
https://docs.oracle.com/javase/7/docs/technotes/tools/windows/javadoc.html#serialData

### LINKED ISSUE #3034: javadoc tag exception
I found the following problem. If I have this two lines in my javadoc:

`@throws IllegalArgumentException if ....`
`@exception FileNotFoundException if ....`

If I call the getParam() method from the @throws tag, I get "IllegalArgumentException". That is good, but if I call the getParam() method from the @exception tag, I get null.

The getContent() from @throws is "if ...." and from @exception "FileNotFoundException if ...." 
it looks like the parameter would not be separated properly from the content

Egor18:
fix #3033
fix #3034


src/main/java/spoon/javadoc/internal/JavadocBlockTag.java
```diff
@@ -50,7 +50,7 @@
 		private String keyword;
 
 		boolean hasName() {
-			return this == PARAM || this == THROWS;
+			return this == PARAM || this == THROWS || this == EXCEPTION;
```
monperrus: we have code duplication here. What about refactoring to have a single implementation?

Egor18: Yes, I've just been writing about it. It would be useful for #3033 issue too.

monperrus:
Thanks a lot for the refactoring. In Spoon, we are super careful with backward compatibility (unless it's in an "internal" package), what do you think of moving code of new `spoon.javadoc.JavadocTagType` to previous `spoon.reflect.code.CtJavaDocTag.TagType`?

Egor18:
Sure, you're right.

monperrus:
Cool, thanks!

monperrus:
Thanks a lot Egor.

PR #3042, opened 2019-06-29
fix: ClassCastException in getTypeParameterDeclaration

### LINKED ISSUE #3040: Class cast bug
Hi!
I get a class cast exception, looking at the code the problem seems to be in `CtTypeReferenceImpl` line 811

```
// case 2: this is an actual type argument of a method/constructor reference
if (parent instanceof CtExecutableReference) {
	CtExecutable<?> exec = ((CtExecutableReference<?>) parent).getExecutableDeclaration();
	if (exec instanceof CtMethod || exec instanceof CtConstructor) {
		return findTypeParamDeclarationByPosition((CtFormalTypeDeclarer) exec, ((CtTypeReference) parent).getActualTypeArguments().indexOf(this));
	}
}
```
Here we're checking if the parent is a `CtExecutableReference`, but then casting it to a `CtTypeReference`.

Thanks
Luca

Egor18:
fix #3040


monperrus:
LGTM, thanks a lot @Egor18

PR #3047, opened 2019-07-15
doc: fix discoverability of templates

monperrus:
Just spent 5 minutes to find the template documentation on the website. It means that we have discoverability problem. Fixed here.


monperrus:
@nharrand merge?

PR #3048, opened 2019-07-15
Test: add test for untested class JavaIdentifiers

monperrus:
(one day we'll have 0 class with 0% coverage)


monperrus:
@nharrand merge?

PR #3056, opened 2019-07-23
fix: Fix issue #3052

### LINKED ISSUE #3052: bug: Unknown source position when using lambda in inner class
Hello,

I have a quite weird behavior in my processor when analyzing this class: https://github.com/woffs/tazapp-android/blob/master/tazapp/src/main/java/de/thecode/android/tazreader/reader/usertoc/UserTocAdapter.java#L180-L191

I tracked down the bug to this minimal code sample: 
```java
public class Test {
    public static class Failing {
        Any listener;
    }

    private static class ExtendedClass extends Failing {
        public void test() { 
           // The difference in behaviour happens if I comment this line out
           Runnable r = v -> listener.someAction();
        }
    }
}
```

Here, the class `Failing` will have a position returning `(uknown file)` in my processor if the line with the lambda stays uncommented.
My processor will work fine if I comment it out though.

You will find as follow the configuration of my spoon launcher:
```java

        launcher.addInputResource(appPath);
        launcher.buildModel();

        AbstractProcessor<CtClass> classProcessor = new ClassProcessor();
        AbstractProcessor<CtInterface> interfaceProcessor = new InterfaceProcessor();
        launcher.addProcessor(classProcessor);
        launcher.addProcessor(interfaceProcessor);
        launcher.process();
```

And I will also attach the two states from my debugger.
Do you have an idea of the issue? Anything I can do about it?

Thanks in advance.

2019-07-18-233325_565x504_scrot
2019-07-18-233420_1045x419_scrot

Egor18:
Fix #3052.
Here is why this bug is happening:
Eclipse jdt tells us that the binding of the `listener` variable inside the lambda is null in noclasspath mode, 
so we run `createVariableAccessNoClasspath()`, it calls `getVariableDeclaration()`, which uses `classFactory.create()` for inner classes improperly. This leads to the creation of a new top-level class `Failing` (with unknown position), so references get messed up and we get this new class with no position instead of the original `Failing`.

More specifically the problem was that the description of the `classFactory.create()` said that it can handle both `.` and `$`, but its implementation handles only `$`, so I added `getNormalQualifiedName()` and fixed context builder. By the way, it's not possible to handle them both in `classFactory.create()`, since `.` is used as a package separator and `$` as an inner type separator and we need to distinguish between the two of them.

### DIFF
--- a/src/main/java/spoon/reflect/factory/ClassFactory.java
+++ b/src/main/java/spoon/reflect/factory/ClassFactory.java
@@ -59,8 +59,7 @@ public class ClassFactory extends TypeFactory {
 	 * @param <T>
 	 *            type of created class
 	 * @param qualifiedName
-	 *            full name of class to create. Name can contain . or $ for
-	 *            inner types
+	 *            full name of class to create. Name can contain $ for inner types
 	 */
 	public <T> CtClass<T> create(String qualifiedName) {
 		if (hasInnerType(qualifiedName) > 0) {
--- a/src/main/java/spoon/reflect/factory/InterfaceFactory.java
+++ b/src/main/java/spoon/reflect/factory/InterfaceFactory.java
@@ -45,7 +45,13 @@ public class InterfaceFactory extends TypeFactory {
 	}
 
 	/**
-	 * Creates an interface.
+	 * Creates an interface from its qualified name.
+	 *
+	 * @param <T>
+	 * 		type of created interface
+	 *
+	 * @param qualifiedName
+	 * 		full name of interface to create. Name can contain $ for inner types
 	 */
 	@SuppressWarnings("unchecked")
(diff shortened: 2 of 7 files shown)


monperrus:
That's great! thanks a lot @Egor18 

@aveuiller would you make a try and confirm that it fixes your problem before we merge to master?

aveuiller:
I tried it out and the fix is working. Thanks for the quick reaction !

Do you know when you will to release a patch version with this fix?

monperrus:
> Do you know when you will to release a patch version with this fix?

When #3057 is merged.

monperrus:
Should be released now :-)

aveuiller:
I confirm, and it's working perfectly. Thanks!

PR #3060, opened 2019-07-25
fix: compilation errors are made visible

monperrus:
Needed after an horror debug session due to silenced messages.


monperrus:
ready to merge

Issue #3065, opened 2019-07-29
bug: spoon.SpoonException: Declaring type of field result isn't defined


trexsatya:
I am trying to process Java classes from ElasticSearch. There is a certain class which if added to Launcher.
Here's my code to process the file.

```java

public static void analyze(String path) throws IOException, InterruptedException {
        Launcher launcher = new Launcher();

        Thread.setDefaultUncaughtExceptionHandler((x,y) -> {
            y.printStackTrace();
        });

        launcher.addInputResource("/home/satyendra/IdeaProjects/elasticsearch/modules/elasticsearch/src/main/java/org/elasticsearch/http/Utf8HttpResponse.java");

        launcher.getEnvironment().setAutoImports(true); // optional

        launcher.getEnvironment().setNoClasspath(true); // optional

        launcher.run();
        final Factory factory = launcher.getFactory();
        final ProcessingManager processingManager = new QueueProcessingManager(factory);
        final ClassProcessor processor = new ClassProcessor();
        processingManager.addProcessor(processor);
        processingManager.process(factory.Class().getAll());

        System.out.println(factory.Class().getAll());

        CtType s = new TypeFactory().get(String.class);
//        s.getFields().forEach(System.out::println);
    }
```
The class that fails is [UtfHttpResponse](https://gist.github.com/trexsatya/6d9f15dc2a25c4cd9f1caa116b58ffd1)

I did a little digging and find that the problem is due to multi-line static block.

surli:
I just tried locally and I indeed got the following stacktrace:
```
Exception in thread "main" spoon.SpoonException: Declaring type of field result isn't defined
	at spoon.support.reflect.reference.CtFieldReferenceImpl.getActualField(CtFieldReferenceImpl.java:57)
	at spoon.support.reflect.reference.CtFieldReferenceImpl.getModifiers(CtFieldReferenceImpl.java:238)
	at spoon.reflect.visitor.DefaultJavaPrettyPrinter.isInitializeStaticFinalField(DefaultJavaPrettyPrinter.java:850)
	at spoon.reflect.visitor.DefaultJavaPrettyPrinter.printCtFieldAccess(DefaultJavaPrettyPrinter.java:797)
	at spoon.reflect.visitor.DefaultJavaPrettyPrinter.visitCtFieldWrite(DefaultJavaPrettyPrinter.java:756)
	at spoon.support.reflect.code.CtFieldWriteImpl.accept(CtFieldWriteImpl.java:16)
[...]

PR #3066, opened 2019-07-29
review fix: don't check modifiers when CtVariable#getDeclaringType() returns null #3065

surli:
* Patch proposal
  * Unit test for bug reproduction


monperrus:
LGTM. Thanks a lot for the patch @surli

monperrus:
Thanks Simon

Issue #3068, opened 2019-07-30
Parsing a one-line class may produce incorrect line number


arnobl:
Spoon 7.5.0. openjdk 11.

Analysed code: `public class A { public Object b() { return a < b; }}` in a virtual file.

As results, all the line positions are incorrect (-1).
No issue with two-line code:
`public class A { public Object b() { return a <\n b; }}`

Test example:

```java
public class Test {
  public static void main(final String[] args) {
    final Launcher launcher = new Launcher();

    launcher.addInputResource(
      new VirtualFile("public class A { public Object b() { return a < b; }}", "chunk.java"));

    launcher.buildModel();
    final CtModel model = launcher.getModel();
    
    model.getElements(new TypeFilter<>(CtElement.class))
      .stream()
      .filter(elt -> !(elt.getPosition() instanceof NoSourcePosition))
      .forEach(elt -> System.out.println(elt.getClass().getSimpleName() + ": " +
        elt.getPosition().getLine() + " " + elt.getPosition().getEndLine()));
  }
```

monperrus:
Thanks for the bug report. Could you push a pull request with a failing test case?

PR #3069, opened 2019-07-31
fix(core): Parsing a one-line class may produce incorrect line number #3068

arnobl:
Test added to spot the issue #3068 (parsing a one-line class may produce incorrect line number).


arnobl:
Added a fix to the bug. Quite simple: if no line break, the code returned -1 instead of 1.

monperrus:
Thanks a lot @arnobl !

PR #3075, opened 2019-08-10
bug: Type of string concatenation is null

Egor18:
Hi!
I found out that Spoon returns null type for 2 or more concatenations of string literals.
For example:
`String s = "one" + "two" + "three";`
Spoon says that the type of "two" + "three" is null, but it's actually String.
I provided the test that demonstrates that.

Moreover, I'm sure that the problem is here:
https://github.com/INRIA/spoon/blob/f2ecd5def520d0e8fb5f7c43a8889affd1347aa2/src/main/java/spoon/support/compiler/jdt/ParentExiter.java#L421-L432
We create a new `CtBinaryOperator` but do not set its type.

So I tried to fix that by adding setType call:
`op.setType((CtTypeReference) operator.getFactory().Type().STRING);`

This fix works and type appears, but `AstParentConsistencyChecker` does not like it, because `setType` sets parent of the reference to `op`, so I'm getting:
```
java.lang.IllegalStateException: Element: java.lang.String
Signature: spoon.support.reflect.reference.CtTypeReferenceImpl@943a4c32
Class: class spoon.support.reflect.reference.CtTypeReferenceImpl
position: (unknown file)
 is set as child of
Element: (("two" + "three") + "four")
Signature: spoon.support.reflect.code.CtBinaryOperatorImpl@1
Class: class spoon.support.reflect.code.CtBinaryOperatorImpl
position: (/home/egor/spoon/src/test/java/spoon/test/type/testclasses/Foo.java:7)
however it is visited as a child of
Element: ("two" + "three")
Signature: spoon.support.reflect.code.CtBinaryOperatorImpl@1
Class: class spoon.support.reflect.code.CtBinaryOperatorImpl
position: (/home/egor/spoon/src/test/java/spoon/test/type/testclasses/Foo.java:7)
```

Do you have any ideas how to set the parent properly here?
Thanks.

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/ParentExiter.java
+++ b/src/main/java/spoon/support/compiler/jdt/ParentExiter.java
@@ -423,7 +423,7 @@ public class ParentExiter extends CtInheritanceScanner {
 				op.setKind(BinaryOperatorKind.PLUS);
 				op.setLeftHandOperand(operator.getRightHandOperand());
 				op.setRightHandOperand((CtExpression<?>) child);
-				//op.setType((CtTypeReference) operator.getFactory().Type().STRING);
+				op.setType((CtTypeReference) operator.getFactory().Type().STRING.clone());
 				operator.setRightHandOperand(op);
 				int[] lineSeparatorPositions = jdtTreeBuilder.getContextBuilder().getCompilationUnitLineSeparatorPositions();
 				SourcePosition leftPosition = op.getLeftHandOperand().getPosition();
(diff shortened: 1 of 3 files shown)


monperrus:
Thanks for the fix!

Maybe the short `op.setType((CtTypeReference) operator.getFactory().Type().STRING.clone())` (call to clone)

or
```
ref =
(CtTypeReference) operator.getFactory().Type().STRING.clone()
ref.setParent(op)
op.setType(ref)
```

Egor18:
Yes, it works, but Travis is stuck. I'll try to trigger it again.

Egor18:
I think it's ready :)

monperrus:
Thanks a lot!

PR #3081, opened 2019-08-22
fix(pom.xml): upgrade jackson-databind version to 2.9.9.3

nharrand:
It seems that the security notification from github has been dismissed, but as upgrading cost nothing...


monperrus:
thanks!

Issue #2938, opened 2019-04-08
Bug SniperJavaPrettyPrinter and loops with no block


edouard-gillet:
Hi,

When using SniperJavaPrettyPrinter over loops that are not using braces, this issue occurs (for while, foreach, or for loops):

`spoon.SpoonException: The element of class class spoon.support.reflect.code.CtWhileImpl does not have CtRole.STATEMENT
	at spoon.reflect.meta.impl.RoleHandlerHelper.getRoleHandler(RoleHandlerHelper.java:66)
	at spoon.support.sniper.internal.ElementSourceFragment.getRoleHandler(ElementSourceFragment.java:248)
	at spoon.support.sniper.internal.ElementSourceFragment.addChild(ElementSourceFragment.java:188)
	at spoon.support.sniper.internal.ElementSourceFragment.access$000(ElementSourceFragment.java:55)
	at spoon.support.sniper.internal.ElementSourceFragment$1.enter(ElementSourceFragment.java:152)
	at spoon.reflect.visitor.CtScanner.visitCtInvocation(CtScanner.java:505)
	at spoon.support.reflect.code.CtInvocationImpl.accept(CtInvocationImpl.java:55)
	at spoon.reflect.visitor.EarlyTerminatingScanner.doScan(EarlyTerminatingScanner.java:154)
	at spoon.reflect.visitor.EarlyTerminatingScanner.scan(EarlyTerminatingScanner.java:130)
	at spoon.reflect.visitor.CtScanner.scan(CtScanner.java:176)
	at spoon.reflect.visitor.EarlyTerminatingScanner.scan(EarlyTerminatingScanner.java:115)
	at spoon.reflect.visitor.EarlyTerminatingScanner.scan(EarlyTerminatingScanner.java:92)
	at spoon.reflect.visitor.CtScanner.visitCtBlock(CtScanner.java:305)
	at spoon.support.reflect.code.CtBlockImpl.accept(CtBlockImpl.java:67)
	at spoon.reflect.visitor.EarlyTerminatingScanner.doScan(EarlyTerminatingScanner.java:154)
	at spoon.reflect.visitor.EarlyTerminatingScanner.scan(EarlyTerminatingScanner.java:130)
	at spoon.reflect.visitor.CtScanner.scan(CtScanner.java:176)
	at spoon.reflect.visitor.EarlyTerminatingScanner.scan(EarlyTerminatingScanner.java:115)
[...]

PR #3084, opened 2019-08-30
fix:(ParentExiter.java): set position for implicit loop body block

nharrand:
add tests provided by @edouard-gillet in #2938
and fix#2938


monperrus:
really nice, thanks a lot @edouard-gillet for the bug report and @nharrand for the fix!

PR #3085, opened 2019-08-31
review2 fix: order of processing LexicalScope elements of Type

pvojtechovsky:
The type members names must be visited before Type name itself.

Will be tested by GenericsWithAmbiguousStaticField after #2683 is merged.


pvojtechovsky:
it is waitig for #3086, then rebase and from my point of view ready for review and merge too.

nharrand:
Thanks!

PR #3090, opened 2019-09-01
WIP: fix: improve the creation of qualified type references 

tdurieux:
I did those changes but I am not sure if it makes real sense. 

It should compute correctly the access path to inner public classes in a protected class.
The previous implementation only supported one level. This implementation should support more than one.


tdurieux:
The failing test cases highlight some different behaviors in Spoon.
Some times the declaring type contains the access type sometimes the declaring class.
Not sure what we want. 

But the problem with `getAccessType()` is that we don't when to print the fully qualified name or not as it was in the original source code.

monperrus:
>  Some times the declaring type contains the access type sometimes the declaring class.

That's a bug, thanks for finding it. We want the declaring class.

tdurieux:
I think fixing this will cause problems to print some elements.

monperrus:

Thanks!

PR #3096, opened 2019-09-11
WIP fix: does not print fully-qualified of "java.lang"

monperrus:


src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
```diff
@@ -1731,7 +1731,10 @@ public void visitCtWildcardReference(CtWildcardReference wildcardReference) {
 	}
 
 	private boolean printQualified(CtTypeReference<?> ref) {
-		return ignoreImplicit || !ref.isImplicitParent();
+		// in Java, package "java.lang" has a special status
+		// it's the only one that does not need to fully-qualifed
```
seintur: qualified

monperrus: thanks

monperrus:
123 failures:

* I don't want to change 123 assertions
* If we do so, we will also break client code

So I will simply try to stick to the previous behavior.

monperrus:
not the good solution

PR #3103, opened 2019-09-15
fix: indeed call the preprocessors in toString/autoimports

monperrus:
indeed call the preprocessors in toString/autoimports, not in only when printing to file

we're stabilizing even more, towards fixing the client builds

### DIFF
--- a/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
+++ b/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
@@ -316,14 +316,11 @@ public class DefaultJavaPrettyPrinter implements CtVisitor, PrettyPrinter {
 	}
 
 	@Override
-	public DefaultJavaPrettyPrinter prettyprint(CtElement e) {
-		CtType<?> parent = e.getParent(CtType.class);
-		if (parent != null) {
-			// call the validators
-			calculate(e.getFactory().createCompilationUnit(), Arrays.asList(new CtType<?>[]{parent}));
-		}
+	public String prettyprint(CtElement e) {
 		reset();
-		return scan(e);
+		applyPreProcessors(e);
+		scan(e);
+		return this.getResult();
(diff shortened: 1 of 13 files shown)


src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
```diff
@@ -2050,14 +2054,14 @@ private PrinterHelper getPrinterHelper() {
 	/**
 	 * @param preprocessors list of {@link CompilationUnitValidator}, which have to be used to validate and fix model before it's printing
 	 */
-	public void setPreprocessors(List<Processor<CtCompilationUnit>> preprocessors) {
-		this.preprocessors = preprocessors;
+	public void setPreprocessors(List<Processor<CtElement>> preprocessors) {
+		this.preprocessors.addAll(preprocessors);
 	}
```
nharrand: `setPreprocessors` add preprocessors, isn't that a bit missleading? Can we rename the method addPreprocessors? (Even if it changes the API?

src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
```diff
@@ -2050,14 +2054,14 @@ private PrinterHelper getPrinterHelper() {
 	/**
 	 * @param preprocessors list of {@link CompilationUnitValidator}, which have to be used to validate and fix model before it's printing
 	 */
-	public void setPreprocessors(List<Processor<CtCompilationUnit>> preprocessors) {
-		this.preprocessors = preprocessors;
+	public void setPreprocessors(List<Processor<CtElement>> preprocessors) {
```
nharrand:
```suggestion
	public void setPreprocessors(List<Processor<CtElement>> preprocessors) {
		this.preprocessors.clear();
```

monperrus:
thanks for the suggestion, merged.

PR #3105, opened 2019-09-16
feat(SpoonifierVisitor): a visitor that generate the spoon code to recreate a spoon model

nharrand:
…red to recreate a spoon model

PR #3107, opened 2019-09-17
feat(visualisation): panel added to show Spoon objects' properties

arnobl:
**Work in progress.**

A feature for visualising the Spoon objects' attributes, example:

[screenshot](https://raw.githubusercontent.com/arnobl/spoon/spoon-visualisation-attrs/spoon-visualisation/doc/props.png)

The current version of this new feature adds a hyperlink for each tree node to display a window that contains the properties of the concerned Spoon object.

I have two questions:

1. do you prefer a new window or a third panel in the main view? With a new window you may lost the context, while a third panel may make the main window heavy
2. To identify properties, I currently gather methods that are getters (start with 'get', 'is'; no param; not static; defined in Spoon interfaces). Spoon has several annotations for identifying properties: `PropertyGetter`, `DerivedProperty`). Which solution do you prefer?


nharrand:
Hi @arnobl ,
1) I think I would prefer a third panel but both are fine IMHO.
2) I think the annotation would be cleaner, but in theory we should not have a mismatch between the two method, right?

Also, I think it could be useful to be able to access the path of an element (`CtElement#getPath()`) (as a property? when hoovering over the element with the mouse?)

And what would you think about having way to get the spoon code necessary recreate a similar AST (i.e. Spoonify code #3105 )? May be by right clicking on the element (in a sub menu)?

arnobl:
> I think the annotation would be cleaner, but in theory we should not have a mismatch between the two method, right?

I spotted Spoon getters with parameters in `CtClass`:
```java
@DerivedProperty
@PropertyGetter(role = CONSTRUCTOR)
CtConstructor<T> getConstructor(CtTypeReference<?>... parameterTypes);
```
and getters not annotated in `CtTypeInformation` :
```java
boolean isPrimitive();
boolean isAnonymous();
```

Thanks to this new panel, I discover `CtElement#getPath()`! Indeed, very useful! I will create a PR for that in a near future.

Regarding Spoonify, I was thinking about such a tool for the exact reason you mention! Poke me when it will be released.

nharrand:
> ```java
> @DerivedProperty
> @PropertyGetter(role = CONSTRUCTOR)
> CtConstructor<T> getConstructor(CtTypeReference<?>... parameterTypes);
> ```
This could be bypassed by just spotting the role and using `getValueByrRole`, but then you need to handle containers (i.e. lists, sets and maps). (If you want to include this sort of properties.)

> ```java
> boolean isPrimitive();
> boolean isAnonymous();
> ```
Yes, may be we should do something about this.

> Thanks to this new panel, I discover `CtElement#getPath()`! Indeed, very useful! I will create a PR for that in a near future.

Great!

> Regarding Spoonify, I was thinking about such a tool for the exact reason you mention! Poke me when it will be released.

It is almost ready. But it does not generate super human readable code though...

arnobl:
PR updated.
No more windows. A new panel on the right:
[screenshot](https://raw.githubusercontent.com/arnobl/spoon/spoon-visualisation-attrs/spoon-visualisation/doc/props.png)

Much more usable IMO.
I also added the path property at the top of the panel.
Documentation updated.

Ready for review before merge for me.

monperrus:
Nice! LGTM.

nharrand:
Awsome!

PR #3108, opened 2019-09-18
fix: Incorrect model for empty for loop

Egor18:
Hi! I found out that Spoon generates incorrect model for the following code:
```
for (;;)
    x = 5;
```

But the model is:
`for (; x = 5; );`

This PR provides corresponding test and possible fix.

### DIFF
--- a/src/test/java/spoon/test/loop/LoopTest.java
+++ b/src/test/java/spoon/test/loop/LoopTest.java
@@ -17,6 +17,8 @@
 package spoon.test.loop;
 
 import org.junit.Test;
+import spoon.Launcher;
+import spoon.reflect.CtModel;
 import spoon.reflect.code.CtBlock;
 import spoon.reflect.code.CtFor;
 import spoon.reflect.code.CtForEach;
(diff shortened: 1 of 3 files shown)


src/test/java/spoon/test/loop/LoopTest.java
```diff
@@ -61,4 +65,16 @@ public void testForeachShouldHaveAlwaysABlockInItsBody() throws Exception {
...
+	@Test
+	public void testEmptyForLoopExpression() {
+		Launcher launcher = new Launcher();
+		launcher.addInputResource("./src/test/java/spoon/test/loop/testclasses/EmptyLoops.java");
+		CtModel model = launcher.buildModel();
+		CtFor ctFor = model.getElements(new TypeFilter<>(CtFor.class)).get(0);
+		assertTrue(ctFor.getForInit().isEmpty());
+		assertNull(ctFor.getExpression());
+		assertTrue(ctFor.getForUpdate().isEmpty());
+		assertNotNull(ctFor.getBody());
```
monperrus: assertEquals on the body instead?

Egor18: Done

monperrus:
Thanks a lot!

PR #3112, opened 2019-09-20
chore: restore revapi

monperrus:
Revapi was executed by an Inria server, now dead, it is now executed by Travis.


src/main/java/spoon/reflect/visitor/ImportCleaner.java
```diff
@@ -118,7 +118,7 @@ protected void handleTypeReference(CtTypeReference<?> reference, Context context
 		}
 	}
```
nharrand: If these inner classes become public, shouldn't them be documented?

monperrus: done!

nharrand:
Ready for merge?

monperrus:
yes, it works!

PR #3123, opened 2019-09-24
feat(ProcyonDecompiler): add support for procyon

nharrand:
This PR adds support for procyon. It also deprecate fernflower (as it was based on a repackaging of fernflower that was not updated since 2015).


nharrand:
It is not really a work in progress but it depends on #3122 being merged.

PR #3127, opened 2019-09-25
fix: fix more regressions after auto-import refactoring

monperrus:
fix last regressions in NpeFix

TLDR: complex transformations yield original and unexpected models :-)


monperrus:
ready for merge!

PR #3131, opened 2019-09-27
fix regression after auto-import refactoring

monperrus:
this should fix the regression in https://github.com/Spirals-Team/casper


nharrand:
LGTM.

PR #3132, opened 2019-09-28
fix: fix regression (default DJPP is fully qualified)

monperrus:
Plus: `CtElement#toString` is much more regular

(contributes to fixing gumtree-spoon)


monperrus:
ready to merge

PR #3135, opened 2019-10-05
WiP: fix: lambda parameter position

pvojtechovsky:
@HarisAdzemovic ,@monperrus
Here is the failing test which shows that position of lambda parameter is not correct. That problem causes malfunction of sniper pretty printer reported here:
https://github.com/kth-tcs/sonarqube-repair/pull/22

Try to fix it and let me know then if there is still problem with sniper mode.


nharrand:
Closed per  #3137

PR #3144, opened 2019-10-07
bug: SniperPrettyPrinter fails when calling toString on CtElement

HarisAdzemovic:
Failing test case which showcases how SniperPrettyPrinter crashes with SpoonException: Missing SourceFragmentContext when calling toString on a CtElement.

To see that it is the SniperPrettyPrinter which fails, lines 181-184 in TestSniperPrinter.java
can be commented to deactivate SniperPretty. The test then passes.

This is **not** reproducible in version 7.5 so it's a pretty recent bug.

### DIFF
--- a/src/test/java/spoon/test/prettyprinter/TestSniperPrinter.java
+++ b/src/test/java/spoon/test/prettyprinter/TestSniperPrinter.java
@@ -168,6 +168,24 @@ public class TestSniperPrinter {
 		});
 	}
 
+	@Test
+	public void testBinaryOperatorElement() throws Exception {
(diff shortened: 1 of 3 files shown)


HarisAdzemovic:
@nharrand correctly pointed out that the test passes if using toStringDebug rather than toString which should be another clue in figuring it out.

monperrus:
fixed by #3147?

nharrand:
Yes #3147 includes this test.

HarisAdzemovic:
Yes it does! Can I retrigger the Travis build or are only maintainers able to?

PR #3149, opened 2019-10-09
fix: fix regression after import refactoring

monperrus:
fixes build of AssertJ https://ci.inria.fr/sos/job/AssertJ/


monperrus:
ready for review

PR #3152, opened 2019-10-15
doc: add API documentation to public methods

ZhectorSM:
I modified the 2 methods which appeared on the test case, not 3 like the title says


monperrus:
Thanks a lot for your contribution!

Issue #3155, opened 2019-10-20
Usage of createVariableRead instead of createVariableWrite


maakbay:
in spoon/reflect/factory/CodeFactory.java:

`public <A, T extends A> CtAssignment<A, T> createVariableAssignment(CtVariableReference<A> variable, boolean isStatic, CtExpression<T> expression)`

This method uses `createVariableRead(variable, isStatic)` instead of `createVariableWrite(variable, isStatic)`. It produces bugs on code analysis and manipulations. 

Please fix it.

monperrus:
Thanks for the report. Would you make a pull request? That would be great!

maakbay:
Okay, i will try to implement my solution into Factory. Is it right way to do?

monperrus:
closed per #3157

monperrus:
Thanks a lot for the bug report and the fix!

PR #3157, opened 2019-10-22
fix(CodeFactory): createVariableAssignment now creates a CtVariableWrite

maakbay:
Related to https://github.com/INRIA/spoon/issues/3155 .
createVariableWrite methods were creating CtVariableAccesses as CtVariableRead instead of CtVariableWrite. Replaced it usages with createVariableWrite and createFieldWrite. And added new methods which they are symmetrical implementations of write-access generator methods.

### DIFF
--- a/src/main/java/spoon/reflect/factory/Factory.java
+++ b/src/main/java/spoon/reflect/factory/Factory.java
@@ -292,6 +292,11 @@ public interface Factory {
 	<T> CtVariableAccess<T> createVariableRead(CtVariableReference<T> variable, boolean isStatic);
 
 	/**
+	 *  @see CodeFactory#createVariableWrite(CtVariableReference,boolean)
+	 */
+	<T> CtVariableAccess<T> createVariableWrite(CtVariableReference<T> variable, boolean isStatic);
+
+	/**
 	 *  @see CodeFactory#createCtField(String,CtTypeReference,String,ModifierKind[])
 	 */
 	<T> CtField<T> createCtField(String name, CtTypeReference<T> type, String exp, ModifierKind... visibilities);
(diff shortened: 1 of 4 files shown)


src/main/java/spoon/reflect/factory/Factory.java
```diff
@@ -321,6 +326,11 @@
 	 */
 	List<CtExpression<?>> createVariableReads(List<? extends CtVariable<?>> variables);
 
+	/**
+	 *  @see CodeFactory#createVariableWrites(List)
+	 */
+	List<CtExpression<?>> createVariableWrites(List<? extends CtVariable<?>> variables);
+
```
nharrand: For what usages do we need to create a list of `CtVariableWrite`?
(I assume the list of `CtVariableRead` is useful when passing them as parameters to a method invocation.)

maakbay: Right. I just added it to keep methods in right polarity.
I can remove it if you think that there is no need to implement it.

nharrand:
Would you add a test like:

```java
	@Test
	public void testCreateVariableAssignement() {
		Factory factory = createFactory();
		CtType t = factory.Class().create("my.MyClass");
		CtField f = factory.createCtField("f", factory.Type().booleanPrimitiveType(), "false", ModifierKind.PRIVATE);
		t.addField(f);
		CtAssignment va = factory.Code().createVariableAssignment(f.getReference(),false,factory.createLiteral(true));

		//Variable assignment's assignee is a CtVariableWrite that point toward the right variable.
		assertTrue(va.getAssigned() instanceof CtVariableWrite);
		assertEquals(f.getReference(), ((CtVariableWrite) va.getAssigned()).getVariable());
	}
```

in `src/test/java/spoon/test/factory/CodeFactoryTest.java` ?

(I don't have the write rights on your branch.)

nharrand:
Otherwise, LGTM! Thanks a lot for the contribution.

maakbay:
I think i don't have enough understanding on styles of tests of Spoon. So it is better to you modify the codes. I just invited you @nharrand as a collaborator on my fork.

nharrand:
Thanks!

Issue #2732, opened 2018-10-30
bugs in generic contracts on corner cases


monperrus:
See carpet debugging at https://github.com/INRIA/spoon/blob/master/src/test/java/spoon/test/main/MainTest.java#L84

* [ ] bug 1
* [ ] bug 2
* [ ] bug 3
* [ ] bug 4
* [ ] bug 5
* [ ] bug 6
* [ ] problem from #3049

monperrus:
See also #3049

ZhectorSM:
Hi, 
I want to help with this issue, but do you know if its already fixed?
The test case is successful.
![issue2732](https://user-images.githubusercontent.com/40966500/66957154-8a342580-f01a-11e9-8947-02293859c8c6.png)

Thank you

monperrus:
You would have to first comment out one of the filters from line 85 to 103.

ZhectorSM:
Hi,
I'm working on bug 1 of this issue.
What I found is when the method checkShadow in the class ContractVerifier does not have any break instruction when the element is not a shadow, thus the program continues the execution triggering the bug.

Test case values at the moment of the error:

reference	              CtTypeReferenceImpl
isShadow	              false	
typeDeclaration	      null	

The solution I propose is just to exit the method when isShadow = false, unless there is a specific behaviour that should follow.
With this, the test case passes without error.

Current code:

```java
@Override
public void scan(CtElement element) {
	if (element != null && CtShadowable.class.isAssignableFrom(element.getClass())) {
		assertFalse(((CtShadowable) element).isShadow());                                           
	}
	super.scan(element);
}
```


Proposed solution:

```java
@Override
public void scan(CtElement element) {
	if (element != null && CtShadowable.class.isAssignableFrom(element.getClass())) {
		assertFalse(((CtShadowable) element).isShadow());                                        
                         return;
	}
	super.scan(element);
}
```


Thanks.

monperrus:
Thanks a lot for the detailed analysis. @nharrand WDYT?

@ZhectorSM would you create a WIP pull-request?

ZhectorSM:
@monperrus yeah, it's done!

monperrus:
updated with todos at the top comment

PR #3159, opened 2019-10-25
WIP fix: bugs in generic contracts on corner cases (bug 1)

ZhectorSM:
Fixed bug 1: method checkShadow in the class ContractVerifier
Related to #2732

### DIFF
--- a/src/main/java/spoon/ContractVerifier.java
+++ b/src/main/java/spoon/ContractVerifier.java
@@ -202,7 +202,8 @@ public class ContractVerifier {
 			@Override
 			public void scan(CtElement element) {
 				if (element != null && CtShadowable.class.isAssignableFrom(element.getClass())) {
-                                        assertFalse(((CtShadowable) element).isShadow());                                        
+					assertFalse(((CtShadowable) element).isShadow());
+					return;
                                     return;
 				}
 				super.scan(element);
--- a/src/main/java/spoon/ContractVerifier.java
+++ b/src/main/java/spoon/ContractVerifier.java
@@ -204,7 +204,6 @@ public class ContractVerifier {
 				if (element != null && CtShadowable.class.isAssignableFrom(element.getClass())) {
 					assertFalse(((CtShadowable) element).isShadow());
 					return;
-                                    return;
 				}
 				super.scan(element);
 			}
--- a/src/main/java/spoon/ContractVerifier.java
+++ b/src/main/java/spoon/ContractVerifier.java
@@ -202,7 +202,8 @@ public class ContractVerifier {
 			@Override
 			public void scan(CtElement element) {
 				if (element != null && CtShadowable.class.isAssignableFrom(element.getClass())) {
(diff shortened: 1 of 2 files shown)


ZhectorSM:
There is an error with the spaces, I tried to fix it but it didn't work.
I don't know why.

src/main/java/spoon/ContractVerifier.java
```diff
@@ -202,7 +202,8 @@ public void checkShadow() {
 			@Override
 			public void scan(CtElement element) {
 				if (element != null && CtShadowable.class.isAssignableFrom(element.getClass())) {
-					assertFalse(((CtShadowable) element).isShadow());
+                                        assertFalse(((CtShadowable) element).isShadow());
```
nharrand:
```suggestion
					assertFalse(((CtShadowable) element).isShadow());
					return;
```

src/main/java/spoon/ContractVerifier.java
```diff
@@ -202,7 +202,8 @@ public void checkShadow() {
 			@Override
 			public void scan(CtElement element) {
 				if (element != null && CtShadowable.class.isAssignableFrom(element.getClass())) {
-					assertFalse(((CtShadowable) element).isShadow());
+                                        assertFalse(((CtShadowable) element).isShadow());                                        
+                                    return;
```
nharrand:
```suggestion
```

nharrand:
It seems to be just a small issues with tabulation versus spaces (and line should not ennd with white characters).

nharrand:
By the way, if you want to check if there are styling issues locally, you can run:

```sh
mvn checkstyle:checkstyle
```

ZhectorSM:
Perfect!
Thank you for your help.

src/main/java/spoon/ContractVerifier.java
```diff
@@ -203,6 +203,7 @@ public void checkShadow() {
 			public void scan(CtElement element) {
 				if (element != null && CtShadowable.class.isAssignableFrom(element.getClass())) {
 					assertFalse(((CtShadowable) element).isShadow());
+					return;
```
nharrand: If you do add this `return` here, you prevent the scanner (here `checkShadow`) to visitit descendent of a `CtElement` if it is a class that implement `CtShadowable`. As we run contract versifier on a root package (an instance of `CtPackage` which does implement `CtShadowable`), this patch would stop the visit immediatly, i.e. it stop verifying the contract. While it does make the test pass, it basically remove its body.

I think the root of the bug (at least in the case of `fieldaccesses/testclasses/Tacos`) is that `typeDeclaration` line 218 is null for a reference pointing to `Foo.SubInner`. 
(And AFAIU it should not be as we have the sources for `spoon.test.fieldaccesses.testclasses.internal.Bar.Inner.SubInner`, and `Foo` extends `Bar.Inner`, so  `Foo.SubInner` should exist and be correctly resolved.)

PR #3165, opened 2019-10-31
fix(equality): Use .equals() when comparing Strings and Boxed types.

HarisAdzemovic:
This fixes 3 Sonarqube violations of rule S4973:
https://rules.sonarsource.com/java/RSPEC-4973


monperrus:
Thanks!

To avoid potential NPEs, it would be better to invert receiver/argument and write `Boolean.FALSE.equals()`.

monperrus:
really nice, thanks Haris.

PR #3182, opened 2019-11-28
refactor: avoid throw in finally

monperrus:
Fixes https://sonarqube.ow2.org/project/issues?id=fr.inria.gforge.spoon%3Aspoon-core&resolved=false&severities=CRITICAL&types=BUG


tdurieux:
Are you aware that you are changing the behavior of the code?

monperrus:
WDYT?

nharrand:
Wouldn't that PR introduce a behavior where when `code.run();` throws an exception, the stack `sourceFragmentContextStack` is no longer cleared?
So if someone attempt to print a model, fails, and try to correct it, it will no longer be printable?

tdurieux:
At least this PR changes that behavior. I m not sure what does core.run().

monperrus:
Agree, maybe it's not worth breaking something to fix a Sonarqube rule.

PR #3184, opened 2019-12-01
refactor: remove @Deprecated methods

monperrus:
Remove @Deprecated methods and types.

Cool factor: I used Spoon in sniper mode for doing this, see method `removeDeprecatedMethods`. It worked, kudos to the sniper mode!

`.cvsignore` is removed because it crashes the sniper pretty-printer.


monperrus:
Not so many resources on automated removal of deprecated methods, a serious one is [Drupal 9: Automated Deprecated Code Removal - A Proof of Concept](https://pronovix.com/blog/drupal-9-automated-deprecated-code-removal-proof-concept)

monperrus:
ready to merge

src/main/java/spoon/reflect/cu/CompilationUnit.java
```diff
@@ -12,37 +12,4 @@
  * public type declaration and other secondary types declarations (not public).
  */
 public interface CompilationUnit extends CtCompilationUnit {
```
surli: I guess the whole interface should be put as deprecated now? or is it still used internally?

src/test/java/spoon/test/imports/ImportScannerTest.java
```diff
@@ -231,94 +228,6 @@ private boolean isTypePresentInStaticImports(String type, Collection<String> sta
 		return imports;
 	}
 
-	@Test
```
surli: I guess the whole class should be renamed since `ImportScanner` does not exist anymore

surli: Actually I'm not really comfortable with removing those tests: shouldn't they be refactored?

src/test/java/spoon/test/refactoring/RefactoringTest.java
```diff
@@ -131,4 +132,12 @@ public void testTransformedInstanceofAfterATransformation() {
 		assertEquals("o", instanceofInvocation.getLeftHandOperand().toString());
 		assertEquals("spoon.test.refactoring.testclasses.AClassX", instanceofInvocation.getRightHandOperand().toString());
 	}
-}
+
+	@Test
+	public void testremoveDeprecatedMethods() {
```
surli: Would be better to put an ignore on the test, instead of commenting its content, wdyt?

monperrus:
There are testing the implementation, not the functionality, so they can be removed.

monperrus:
I'm not in favor of Ignored tests, this usually results in maintenance and technical debt.

src/main/java/spoon/reflect/cu/CompilationUnit.java
monperrus: good idea, done.

src/test/java/spoon/test/imports/ImportScannerTest.java
monperrus: you're right the rest if `ImportScannerTest` has been moved to `ImoprtTest`

monperrus:
pushed changes

monperrus:
fixed ci

surli:
> I'm not in favor of Ignored tests, this usually results in maintenance and technical debt.

Well, I'm not a huge fan of commented code, which I find harder to find for maintenance: at least with ignores you're sure that the test is still compiling and you know it's there with the test result.
Now it definitely isn't a blocker for this PR.

Re the refactoring of tests, I didn't check those in the details so I trust you on this one, but I guess we should add a coverage threshold at a point :)

monperrus:
Thanks!

PR #3189, opened 2019-12-09
review: fix: issue #2237 reject invalid names

### LINKED ISSUE #2237: bug: should reject invalid names according to the Java specification

MartinWitt:
fix #2237

this should fix issue #2237. But a lot of Code uses CtReferenceImpl::setSimpleName for creating illegal references. This code must be changed for this small issue. Like CodeFactory::createClassAccess creates a fieldReference with simpleName "class", which is a illegal identifier.

### DIFF
--- a/src/main/java/spoon/support/reflect/reference/CtReferenceImpl.java
+++ b/src/main/java/spoon/support/reflect/reference/CtReferenceImpl.java
@@ -112,7 +112,7 @@ public abstract class CtReferenceImpl extends CtElementImpl implements CtReferen
 	}
 	private boolean checkAllParts(String[] simplenameParts) {
 		for (String simpleName:simplenameParts) {
-			//because arrays use e.g. int[] and @Number is used for instances of an object e.g. foo@1 
+			//because arrays use e.g. int[] and @Number is used for instances of an object e.g. foo@1
 			simpleName = simpleName.replaceAll("\\[\\]|@", "");
 			if (isKeyword(simpleName) || checkIdentifierChars(simpleName)) {
 				return true;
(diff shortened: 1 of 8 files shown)


monperrus:
nice!

monperrus:
Hi @MartinWitt, how does it go with the failing test cases?

MartinWitt:
Hi @monperrus , after reviewing the failed test cases, i looks like jdt uses e.g. $Number$Name for anonymous classes. This is not a correct identifier for the JLS ( but we should not change jdt :) )
In this solution spoon catches a lot of wrong identifiers but not all, which should be still better than allowing all.
wdyt
I cant find the reason why 3 tests still fail.

monperrus:
what about slightly changing the JDT name before called setSimpleName?

MartinWitt:
I wouldn´t change the jdt name. I the current solution all correct identifier and empty, starting with number and follow the format $identifier@$Number identfier are allowed.
- Starting with number is syntax for local classes.  
- $identifier@$Number is used for reference objects instances.
Now spoon should catch almost all wrong identifier and having some wrong passing is fine for me.

monperrus:
>  Now spoon should catch almost all wrong identifier and having some wrong passing is fine for me.

I agree.

Thanks a lot for the test updates, LGTM.

monperrus:
Thanks a lot, that's very nice!

PR #3193, opened 2019-12-14
fix: adds support to the VisitorPartialEvaluator for the NEG UnaryOperator

stone-wall:
Adds support to the VisitorPartialEvaluator for the NEG UnaryOperator. Adds test for the operator.

Before this when a NEG UnaryOperator was encountered by the VisitorPartialEvaluator it would throw the unsupported operator Runtime Exception.


monperrus:
Thanks! Cool that your put your hands in VisitorPartialEvaluator!

PR #3202, opened 2020-01-09
refactor: migrate to org.apache.logging.log4j:log4j-core:2.13.0

nharrand:
`log4j:log4j` is stuck to `1.2.7`, which contains a security vulnerability. 
The new artifact for log4j is now `org.apache.logging.log4j:log4j-core`.
The api has slightly changed for version `2.x`, see [this page](https://logging.apache.org/log4j/2.x/manual/migration.html) for migration information.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -106,7 +106,7 @@
     <dependency>
       <groupId>com.fasterxml.jackson.core</groupId>
       <artifactId>jackson-databind</artifactId>
-      <version>2.9.10</version>
+      <version>2.9.10.1</version>
     </dependency>
     <dependency>
       <!-- support for compressed serialized ASTs -->
--- a/pom.xml
+++ b/pom.xml
@@ -78,9 +78,9 @@
       <version>2.1</version>
     </dependency>
     <dependency>
-      <groupId>log4j</groupId>
-      <artifactId>log4j</artifactId>
-      <version>1.2.17</version>
+      <groupId>org.apache.logging.log4j</groupId>
+      <artifactId>log4j-core</artifactId>
+      <version>2.13.0</version>
     </dependency>
     <dependency>
       <groupId>commons-io</groupId>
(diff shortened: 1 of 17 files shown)


monperrus:
Thanks a lot!

monperrus:
FYI, this change breaks some usages, starting with using Spoon in Groovy: https://travis-ci.org/SpoonLabs/spoon-ci-external/builds/635334772

It seems that it's because of dependency shadowing, but I'm not sure.

monperrus:
Also breaks the Nopol build: https://ci.inria.fr/sos/job/nopol/719/display/redirect

log4j-core seems not usable as is at runtime.

nharrand:
For Nopol:

It seems that these lines use classes from the old dependency, while not declaring it. (L59 could simply be replacde by `factory.getEnvironment().setLevel("OFF");`

https://github.com/SpoonLabs/nopol/blob/2bcc7dc685a1e66189363aa2a6be082868440106/nopol/src/main/java/fr/inria/lille/commons/spoon/SpoonedFile.java#L5
https://github.com/SpoonLabs/nopol/blob/2bcc7dc685a1e66189363aa2a6be082868440106/nopol/src/main/java/fr/inria/lille/commons/spoon/SpoonedFile.java#L59

nharrand:
FOr Groovy, you might be right. I am not sure of what's happening. But maybe just updating log4j was not the smartest move. Maybe we should move to slf4j and avoid logger compatibility issues. WDYT?

monperrus:
I confirm that Nopol is fixed: https://ci.inria.fr/sos/job/nopol/722/

monperrus:
> Maybe we should move to slf4j and avoid logger compatibility issues.

that's a great idea.

monperrus:
really want to get Groovy working again

#3208 is an option

PR #3208, opened 2020-01-19
fix logging issue

monperrus:
we only depen on log4j-api and slf4j provdes the implementation

### DIFF
--- a/src/main/java/spoon/support/StandardEnvironment.java
+++ b/src/main/java/spoon/support/StandardEnvironment.java
@@ -8,7 +8,6 @@ package spoon.support;
 
 import org.apache.logging.log4j.Level;
 import org.apache.logging.log4j.Logger;
-import org.apache.logging.log4j.core.config.Configurator;
 import spoon.Launcher;
 import spoon.OutputType;
 import spoon.SpoonException;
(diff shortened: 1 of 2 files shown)


monperrus:
Removed call to Configurator because it's not in the public API, see https://stackoverflow.com/a/23434603

monperrus:
ready for review

src/main/java/spoon/support/StandardEnvironment.java
```diff
@@ -510,7 +502,7 @@ public void setNoClasspath(boolean option) {
 	@Override
 	public boolean getNoClasspath() {
 		if (this.noclasspath == null) {
-			logger.warn("Spoon is currently use with the default noClasspath option set as true. Read the documentation for more information: http://spoon.gforge.inria.fr/launcher.html#about-the-classpath");
+			print("Spoon is currently use with the default noClasspath option set as true. Read the documentation for more information: http://spoon.gforge.inria.fr/launcher.html#about-the-classpath", level.WARN);
```
seintur: use -> used

monperrus: thanks. Done.

monperrus:
@nharrand @tdurieux merge?

nharrand:
So if I understand this code correctly, we don't migrate per say to slf4j, we keep using log4j's api and use a bridge implementation to route the call to slf4j.

I am ok with this. I'll merge.

monperrus:
> we don't migrate per say to slf4j, we keep using log4j's api and use a bridge

yes, for now.

PR #3209, opened 2020-01-19
fix: fix pretty-printing of special characters in literals

### LINKED ISSUE #3203: LiteralHelper replaces printable Unicode characters with escape sequence
My code contains a lot of text containing German Umlaute. When I pretty print this text, all the Umlaute are replaced by unicode escape sequences. That makes the text unreadable.
Since my code is UTF-8 no escaping is necessary. But I can't turn off this 'feature'.
I would prefer a way to tell the pretty printer, which characters I want escaped and which not.
Also, for some reason the single quote character (') is escaped, too.

Egor18:
Fix #3203
Fix #3203


monperrus:
Thanks a lot Egor, LGTM

monperrus:
Many thanks Egor!

Issue #3219, opened 2020-01-31
Bug: default method modifier is removed when using sniper


woutersmeenk:
When using sniper and modifying this method:

```java
public interface DefaultMethod {
  default public void test() {
  }
}
```

It removes the default modifier

monperrus:
closed per #3220

PR #3220, opened 2020-01-31
Add test for sniper removing default method.

woutersmeenk:
See #3219


monperrus:
I'm on it,making progress but it's tough.

monperrus:
fixed in #3239 :-)

Issue #3224, opened 2020-01-31
Bug: Crash when modifying code with double type bound and using sniper


woutersmeenk:
When using sniper to modify this code:

```java
import java.io.*;

public class DoubleBound<T extends Writer & Readable>
{
}
```

I get this stacktrace:
spoon.SpoonException: The element of class class spoon.support.reflect.declaration.CtTypeParameterImpl does not have CtRole.BOUND
	at spoon.reflect.meta.impl.RoleHandlerHelper.getRoleHandler(RoleHandlerHelper.java:55)
	at spoon.support.sniper.internal.ElementSourceFragment.getRoleHandler(ElementSourceFragment.java:251)
	at spoon.support.sniper.internal.ElementSourceFragment.addChild(ElementSourceFragment.java:191)
	at spoon.support.sniper.internal.ElementSourceFragment.access$0(ElementSourceFragment.java:187)
	at spoon.support.sniper.internal.ElementSourceFragment$1.enter(ElementSourceFragment.java:155)
	at spoon.reflect.visitor.CtScanner.visitCtTypeReference(CtScanner.java:762)

monperrus:
closed per #3225

PR #3225, opened 2020-01-31
fix: fragments can be computed for IntersectionTypeReference and used in sniper

woutersmeenk:
See #3224


moraispgsi:
Also having this issue. Looking forward for this fix.

monperrus:
Bug fixed. The problem was that IntersectionTypeReferences had no position.

nharrand:
LGTM!

monperrus:
Thanks Nicolas!

PR #3235, opened 2020-02-09
Static method import fails test case.

moraispgsi:
This is a very simple problem, similar to a previous problem with static imports. The problem occurs when importing a static method and invoking it. Note that the test does not fail but it displays the error message.


monperrus:
conflict, not failing, closing

Issue #3175, opened 2019-11-17
ParentNotInitializedException


moraispgsi:
```
spoon.reflect.declaration.ParentNotInitializedException: parent not initialized for OnCheckedChangeListener(class spoon.support.reflect.reference.CtTypeReferenceImpl)
	at spoon.support.reflect.declaration.CtElementImpl.getParent(CtElementImpl.java:362)
	at spoon.support.sniper.SniperJavaPrettyPrinter.hasImplicitAncestor(SniperJavaPrettyPrinter.java:191)
	at spoon.support.sniper.SniperJavaPrettyPrinter.printElement(SniperJavaPrettyPrinter.java:202)
	at spoon.support.reflect.declaration.CtElementImpl.toString(CtElementImpl.java:295)
	at java.lang.String.valueOf(String.java:2994)
	at java.lang.StringBuilder.append(StringBuilder.java:131)
	at spoon.support.compiler.jdt.JDTTreeBuilderHelper.handleImplicit(JDTTreeBuilderHelper.java:519)
	at spoon.support.compiler.jdt.JDTTreeBuilderHelper.handleImplicit(JDTTreeBuilderHelper.java:484)
	at spoon.support.compiler.jdt.ReferenceBuilder.getTypeReference(ReferenceBuilder.java:508)
	at spoon.support.compiler.jdt.ReferenceBuilder.buildTypeReference(ReferenceBuilder.java:130)
	at spoon.support.compiler.jdt.ReferenceBuilder.buildTypeReference(ReferenceBuilder.java:124)
	at spoon.support.compiler.jdt.ReferenceBuilder.buildTypeReference(ReferenceBuilder.java:142)
	at spoon.support.compiler.jdt.JDTTreeBuilder.visit(JDTTreeBuilder.java:1511)
	at org.eclipse.jdt.internal.compiler.ast.QualifiedTypeReference.traverse(QualifiedTypeReference.java:212)
	at org.eclipse.jdt.internal.compiler.ast.QualifiedAllocationExpression.traverse(QualifiedAllocationExpression.java:682)
	at org.eclipse.jdt.internal.compiler.ast.MessageSend.traverse(MessageSend.java:1107)
	at org.eclipse.jdt.internal.compiler.ast.MethodDeclaration.traverse(MethodDeclaration.java:365)
	at org.eclipse.jdt.internal.compiler.ast.TypeDeclaration.traverse(TypeDeclaration.java:1415)
[...]

PR #3236, opened 2020-02-09
Test case for the parent not initialized problem.

moraispgsi:
Replicating Issue https://github.com/INRIA/spoon/issues/3175 with a test case.


monperrus:
Thanks for the valuable failing test case, fixed in #3247.

moraispgsi:
I am glad it helped :)

PR #3237, opened 2020-02-09
Test case with full source code from https://github.com/ccrama/Slide

moraispgsi:
Errors found here:
- PARENT NOT INITIALIZED - https://github.com/INRIA/spoon/issues/3175
- java.lang.NullPointerException
- ERROR spoon.Launcher - The element of class class spoon.support.reflect.declaration.CtTypeParameterImpl does not have CtRole.BOUND - https://github.com/INRIA/spoon/issues/3224
- [main] ERROR spoon.Launcher - The SourcePosition of elements are not consistent
- ERROR spoon.Launcher - Cannot compare this: [20931, 20942] with other: ["20930", "20937"]

The files with errors have a leading comment with the error description. I suppose Spoon should be able to analyze all these files.

See src/test/java/spoon/processing/ProcessingTest.java for test.


monperrus:
Thanks for the failing test cases, we will look at them, but no ETA for now.

moraispgsi:
I mean there are about 90 files failing due to those 5 errors in about 350 files in just this one codebase. 63 errors are due to Parent not Initialized, that would be the priority in this case. Anyway, is there any workaround that I can take?  Can those files be skipped? Can I still set the classpath dependencies without adding those 90 files? Can I setup Spoon so that when a file gives an exception it just skips the file and omits the error?

monperrus:
> Can I still set the classpath dependencies without adding those 90 files?


Yes, in noclasspath mode.

moraispgsi:
fix-3236 seems to have fixed PARENT NOT INITIALIZED leaving  4 remaining issues:
- java.lang.NullPointerException
- ERROR spoon.Launcher - The element of class class spoon.support.reflect.declaration.CtTypeParameterImpl does not have CtRole.BOUND - #3224
- [main] ERROR spoon.Launcher - The SourcePosition of elements are not consistent
- ERROR spoon.Launcher - Cannot compare this: [20931, 20942] with other: ["20930", "20937"]

monperrus:
we don't need to keep this one open, our process with isolated tests supersedes it.

PR #3242, opened 2020-02-10
ref: remove deprecated command-line option noclasspath

monperrus:


src/main/java/spoon/Launcher.java
```diff
@@ -476,12 +464,6 @@ protected void processArguments() {
 		environment.useTabulations(jsapActualArgs.getBoolean("tabs"));
 		environment.setCopyResources(!jsapActualArgs.getBoolean("no-copy-resources"));
 
-		if (jsapActualArgs.getBoolean("enable-comments")) {
```
surli: Not sure why you included this change in that PR: should be part of another PR to remove the "enable-comments" parameter no?

Issue #3186, opened 2019-12-02
Fail build in case an import towards an external *.internal.* package is made


surli:
We should never import something from the internal package of an external tool. We did it for example in the case of #3185 where we use `org.eclipse.jdt.internal.compiler.lookup.TypeConstants`. We should have a test for that.

monperrus:
closed per #3244

PR #3244, opened 2020-02-11
chore(checkstyle): add check for internal imports

MartinWitt:
Fix for #3186. Spoon has a lot of jdt.internal imports so we cant reject them in checkstyle.
But the regex `^(?!.*(spoon|jdt)).*internal.*$` should catch all other unwanted internal imports.
The update of the checkstyle version from 3.0.0 to 3.1.0 is necessary, because 3.0.0 does not contain regex property.

### DIFF
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -28,6 +28,10 @@
     <module name="AvoidStarImport"/>
     <module name="IllegalImport"/> <!-- defaults to sun.* packages -->
     <module name="RedundantImport"/>
+    <module name="IllegalImport">
+      <property name="regexp" value="true" />
+      <property name="illegalClasses" value="^(?!.*(spoon|jdt)).*internal.*$" />
+    </module>
     <module name="UnusedImports">
       <property name="processJavadoc" value="true"/>
     </module>
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -31,6 +31,10 @@
     <module name="IllegalImport">
       <property name="regexp" value="true" />
       <property name="illegalClasses" value="^(?!.*(spoon|jdt)).*internal.*$" />
+      <!--The regex looks for imports containing internal and not containing spoon or jdt. Some examples are:
+      import spoon.pattern.internal.node.ForEachNode; // allowed because internal package but from spoon.
+      import org.eclipse.jdt.internal.compiler.batch.CompilationUnit; //allowed because internal package but from jdt.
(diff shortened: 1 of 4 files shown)


monperrus:
thanks for the PR, some CI problems to be fixed.

MartinWitt:
the CI was fixed in the last commit 3ea8bed

checkstyle.xml
```diff
@@ -28,6 +28,10 @@
     <module name="AvoidStarImport"/>
     <module name="IllegalImport"/> <!-- defaults to sun.* packages -->
     <module name="RedundantImport"/>
+    <module name="IllegalImport">
+      <property name="regexp" value="true" />
+      <property name="illegalClasses" value="^(?!.*(spoon|jdt)).*internal.*$" />
```
surli: Could you provide some examples in a comment of what exactly you want to catch? I'm not sure to understand it in the context of the PR so I guess I won't understand it at all in few weeks :) 

The regexp is to define which imports are illegal? Or in which package to look for illegal classes? So for example, does it look in https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/compiler/jdt/JDTImportBuilder.java? Thanks!

MartinWitt: This change adds a check look at all imports and stops user from importing internal packages expect spoon and jdt ones.
The regex looks for imports containing `internal` and **not** containing  spoon or jdt.

Some examples are:
`import java.util; //fine
import spoon.pattern.internal.node.ForEachNode; // allowed because internal package but from spoon.
import org.eclipse.jdt.internal.compiler.batch.CompilationUnit; //allowed because internal package but from jdt.
import org.apache.commons.jexl2.internal.introspection; //rejected because import of internal package`
Internal packages from jdt are allowed because it's needed in some spoon classes.
Should i add a comment in the pom for clarification?

surli: Thanks for the explanation :) 

> The regex looks for imports containing internal and not containing spoon or jdt.

nice I had to check back the Pattern documentation to find that `?!` was for a: 
> (?!X) | X, via zero-width negative lookahead

didn't know about that :)

> Should i add a comment in the pom for clarification?

That would be really cool. At least one line such as the quote above and one example of allowed/rejected. Thanks

monperrus:
Thanks a lot!

monperrus:
Tip: for automatically the corresponding issue at merge time, write `Fix #3186` and not `Fix for #3186`

Issue #3245, opened 2020-02-11
getExecutableDeclaration returning null on a known class


santos-samuel:
Hello,
I have a doubt about Spoon.

I am using Spoon to construct a CallGraph of some classes belonging to a project folder.
I am iterating over the classes/methods and following a DFS approach.

To start, I am giving a folder as input like shown:
```
Launcher launcher = new Launcher();
launcher.addInputResource(".../edition-ldod");

```

This folder has a lot of classes, which includes:
```
(...)/edition-ldod/src/main/java/pt/ist/socialsoftware/edition/ldod/controller/UserController.java
(...)/edition-ldod/src/main/java/pt/ist/socialsoftware/edition/ldod/validator/ChangePasswordValidator.java
```

By giving the whole folder as input to the launcher, I believe both of these classes will belong to the model (`launcher.buildModel();`).

So, I want to construct a callgraph of a method `changePassword` of the `UserController` class:
```
@RequestMapping(method = RequestMethod.POST, value = "/changePassword")
public String changePassword(@Valid ChangePasswordForm form, BindingResult formBinding) {
	(...)

	ChangePasswordValidator validator = new ChangePasswordValidator(this.passwordEncoder);
	validator.validate(form, formBinding);

	(...)
}
```

To do that, I retrieve all the CtAbstractInvocation Elements of this method.
`method.getElements(new TypeFilter<>(CtAbstractInvocation.class))`

At some point I have a `CtAbstractInvocation` variable (`calleeLocation`) that represents the line `validator.validate(form, formBinding);` (which is in runtime a `CtInvocationImpl`).

Then, to add the method that is accessed in this line to a list I try to do:
`calleeLocation.getExecutable().getExecutableDeclaration()`
but this returns "null" and I can't understand why.

I believe the `getExecutableDeclaration()` should return the method that is being called and that is known because is present in the list of classes provided to the model.
(For instance: `launcher.getFactory().Class().getAll()` shows the `ChangePasswordValidator` class.)

Method being called below:
```
package pt.ist.socialsoftware.edition.ldod.validator;

(...)

public class ChangePasswordValidator implements Validator {

	(...)

	@Override
	public void validate(Object target, Errors errors) {
		(...)
	}

	(...)
}
```

Can someone explain me why this is happening?

Thanks in advance!

monperrus:
> calleeLocation.getExecutable().getExecutableDeclaration()
> but this returns "null" and I can't understand why.


[...]

PR #3246, opened 2020-02-13
test: reproduce behavior of getExecutableDeclaration discussed in #3245

santos-samuel:
Replicating issue #3245 with a test case.


src/test/java/spoon/getExecutableDeclarationBugTest/testClasses/ChangePasswordValidator.java
```diff
@@ -0,0 +1,43 @@
+package pt.ist.socialsoftware.edition.ldod.validator;
```
stone-wall: This package name is incorrect, it should be declared as `package spoon.getExecutableDeclarationBugTest.testClasses;`

src/test/java/spoon/getExecutableDeclarationBugTest/testClasses/ChangePasswordValidator.java
```diff
@@ -0,0 +1,43 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.security.crypto.password.PasswordEncoder;
```
stone-wall: The spoon project doesn't currently have any dependency on the Spring Framework so these imports/class dependencies will always cause the test to fail.

src/test/java/spoon/getExecutableDeclarationBugTest/testClasses/ChangePasswordValidator.java
```diff
@@ -0,0 +1,43 @@
...
+import org.springframework.validation.Errors;
+import org.springframework.validation.Validator;
+
+import pt.ist.socialsoftware.edition.ldod.domain.LdoD;
+import pt.ist.socialsoftware.edition.ldod.domain.LdoDUser;
+import pt.ist.socialsoftware.edition.ldod.forms.ChangePasswordForm;
```
stone-wall: Same as the above. Try to re-create the test case without any extra dependencies on external projects or files.

src/test/java/spoon/getExecutableDeclarationBugTest/testClasses/UserController.java
```diff
@@ -0,0 +1,50 @@
+package pt.ist.socialsoftware.edition.ldod.controller;
```
stone-wall: Should be 
`spoon.getExecutableDeclarationBugTest.testClasses;`

src/test/java/spoon/getExecutableDeclarationBugTest/testClasses/UserController.java
```diff
@@ -0,0 +1,50 @@
...
+import org.springframework.security.crypto.password.PasswordEncoder;
+import org.springframework.stereotype.Controller;
+import org.springframework.validation.BindingResult;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+
+import pt.ist.socialsoftware.edition.ldod.domain.LdoD;
+import pt.ist.socialsoftware.edition.ldod.domain.LdoDUser;
+import pt.ist.socialsoftware.edition.ldod.forms.ChangePasswordForm;
+import pt.ist.socialsoftware.edition.ldod.validator.ChangePasswordValidator;
```
stone-wall: Same problem as in ChangePasswordValidator.java

santos-samuel:
Is there any folder which is ignored by the compiler where i can put these files as external files and not as actual project files?
Because removing those dependencies implies changing most part of the method which will cause de bug to dissapear.

In other words, is it possible not to treat the `spoon.getExecutableDeclarationBugTest.testClasses` as a package?

monperrus:
> Is there any folder which is ignored by the compiler

yes, `src/test/resources`

monperrus:
cloned locally, had a look.

`getExecutableDeclaration` returns null because the argument types do not match:

* in `UserController` you call `validate(pt.ist.socialsoftware.edition.ldod.forms.ChangePasswordForm, org.springframework.validation.BindingResult)`
* in `ChangePasswordValidator`, validate receives as argument `validate(Object, org.springframework.validation.Errors)` 

org.springframework.validation.BindingResult != org.springframework.validation.Errors

santos-samuel:
Oh i see... So it isn't a bug... Spoon only doesn't know that BindingResult is hierarchically related with Errors.
Ok, that closes this issue. Thank you for your availability!

monperrus:
thanks for the super useful failing test case!

PR #3249, opened 2020-02-15
ref: toString() in sniper return toStringDebug()

monperrus:
otherwise a debugger session is a nightmare (the debugger always calls toString :-)


monperrus:
CI green

PR #3250, opened 2020-02-16
Source Position Inconsistent Test

moraispgsi:
Having a lot of problems regarding "The SourcePosition of elements are not consistent" error. This is an isolated test. It fails in almost every file of https://github.com/AntennaPod/AntennaPod


monperrus:
Thanks. Isolated failures are the best! ... really better than #3237 (full project) and #3235 (no failing assertion)

moraispgsi:
I can transform the full project into an individual failing test if you would like. I thought it would be easier to have all the files and uncomment them one by one to see what is resolved as they are marked with the error. I will be separating them if that make it easier and faster to fix. There are some errors that can't be catched for some reason they just fail silently with the printed stacktrace, in that case I can't put an assert, what can be done with those cases?

monperrus:
conflict?

moraispgsi:
Yes, you merged the A.java file in the compilation4 folder to master. This PR also has a different file in the same folder, that is all.

monperrus:
could you fix it?

moraispgsi:
Yes, after work maybe.

moraispgsi:
Conflict resolved @monperrus

monperrus:
checked out locally, merged with master, had to resolve conflicts.

The test passes.

WDYT?

Issue #3133, opened 2019-09-30
SourcePosition from unexpected compilation unit


moraispgsi:
Hi there,

This is my first time posting an issue in this repository. I have been using Spoon for some time and it has been an enjoyable activity. I am currently using it in my master's thesis. I have reached a problem that I cannot solve by myself, so I require your assistance. A little bit of context... I am in the middle of the process o creating a gradle plugin to refactor android applications, powered by Spoon of course. In this particular experiment I used [Slide](https://github.com/ccrama/Slide) an open-source android project. I gathered the dependencies automatically and give them to the spoon launcher classpath. When I try to run the launcher I get different errors depending on the PrettyPrinter that I use (I want to use SniperJavaPrettyPrinter). 

Stacktrace with SniperJavaPrettyPrinter:

```
spoon.SpoonException: SourcePosition from unexpected compilation unit: (/repositories/Slide/app/src/main/java/me/ccrama/redditslide/UserSubscriptions.java:288) expected is: (/repositories/Slide/app/src/main/java/me/ccrama/redditslide/Activities/MainActivity.java:180)
	at spoon.support.sniper.internal.ElementSourceFragment.addChild(ElementSourceFragment.java:225)
	at spoon.support.sniper.internal.ElementSourceFragment.access$000(ElementSourceFragment.java:44)
	at spoon.support.sniper.internal.ElementSourceFragment$1.enter(ElementSourceFragment.java:145)
	at spoon.reflect.visitor.CtScanner.visitCtTypeReference(CtScanner.java:762)
	at spoon.support.reflect.reference.CtTypeReferenceImpl.accept(CtTypeReferenceImpl.java:73)
	at spoon.reflect.visitor.EarlyTerminatingScanner.doScan(EarlyTerminatingScanner.java:143)
	at spoon.reflect.visitor.EarlyTerminatingScanner.scan(EarlyTerminatingScanner.java:119)
	at spoon.reflect.visitor.CtScanner.scan(CtScanner.java:165)
	at spoon.reflect.visitor.EarlyTerminatingScanner.scan(EarlyTerminatingScanner.java:104)
	at spoon.reflect.visitor.CtScanner.visitCtFieldReference(CtScanner.java:447)
	at spoon.support.reflect.reference.CtFieldReferenceImpl.accept(CtFieldReferenceImpl.java:50)
	at spoon.reflect.visitor.EarlyTerminatingScanner.doScan(EarlyTerminatingScanner.java:143)
[...]

PR #3256, opened 2020-02-17
fix: fragment construction should not consider type in local variable references

monperrus:
fix #3254

is the same fix as #3133

depends on #3249 that must be merged first


monperrus:
CI done, ready to merge

Issue #1563, opened 2017-09-28
feature: create a fluent api for Spoon Launcher


surli:
Some of use are considering that the actual Launcher class for Spoon is tedious to use. So we plan to create a new Launcher class to replace it on the long term, which would be easier and more explicit.
I invite you to express here your use cases and how you would want to use such a class, giving some example, to help understand your needs and how we can build such a new class. 

As a first example, I proposed this kind of solution: 
```java
CtModel model = new LauncherBuilder().useAutoimport()
         .useNoclasspath()
         .addInputSource("./myDirectory/")
         .addInputSource("./anotherOne")
         .setOutputDir("./target")
         .build();
```
# Launcher usecases
(please update the list here - so the concept is at one place)
UC1) build AST model from defined set of sources and appropriate config parameters
UC2) print sources from current AST model using appropriate config parameters
UC3) check that AST is consistent
UC4) print sources from current AST model and compile it so class loader can use it
UC5) incremental update of already built AST model by compiling of files modified since last build
UC6) incremental print of sources of model elements modified since last print

So @tdurieux @pvojtechovsky @monperrus @msteinbeck what are your proposal?

pvojtechovsky:
I have added basic use cases as I see them from my point of view. Feel free to add another use cases or change my definitions to have it more understandable.

DaveJarvis:
We have a number of different projects. Each project has a separate directory. It would be convenient if we could provide a list of projects and generate spoon-transformed output in a directory relative to each project's source. This could resemble:

```
// Package containing all processors to run.
final Package processors = Main.class.getPackage();

// Paths to project source files, relative to this Java project.
final List<Project> projects = Arrays.asList(
        new Project( "../../ProjectA", "src" ),
        new Project( "../../ProjectB", "src" )
);

// A LauncherBuilder builds Launcher instances; a CtModelBuilder builds CtModels.
final Launcher launcher = new LauncherBuilder()
        .withProjects(projects)
        .withProcessors(processors)
        .withSourceOutputDirectory("spoon")
        .build();

launcher.run();
```

This would run each processor found in the Main program's package on every source file in all given projects and write to a spoon directory relative to the projects' source directory. Such as:

* **Input.** `../../ProjectA/src/com/company/stardom/Awesome.java`
* **Output.** - `../../ProjectA/spoon/com/company/stardom/Awesome.java`.

To build CtModels, I'd recommend a slight variation to avoid confusion with how builders typically behave (following the principle of least astonishment):

[...]

PR #3258, opened 2020-02-18
feature: add fluent API for Launcher

MartinWitt:
Showcase some code for discussion #1563 . There are still a lot of calls to the environment missing. It's only to showcase the idea.


monperrus:
ohoh, welcome FluentLauncher!

src/main/java/spoon/FluentLauncher.java
```diff
@@ -0,0 +1,160 @@
...
+import spoon.reflect.visitor.Filter;
+import spoon.reflect.visitor.PrettyPrinter;
+import spoon.support.JavaOutputProcessor;
+
+/**
+ * TODO: doc
+ */
+public class FluentLauncher {
+
+	Launcher launcher;
```
DaveJarvis: Can this be `private`?

src/main/java/spoon/FluentLauncher.java
```diff
@@ -0,0 +1,160 @@
...
+	public Factory getFactory() {
+		return launcher.getFactory();
+	}
+
+	// needed?
+	public Environment createEnvironment() {
+		return launcher.createEnvironment();
+	}
+
+	public CtModel buildModel() {
```
DaveJarvis: Is there a need for `buildModel` and `buildCtModel`?

src/main/java/spoon/FluentLauncher.java
```diff
@@ -0,0 +1,160 @@
...
+		launcher.run();
+		return launcher.getModel();
+	}
+
+	public FluentLauncher outputDirectory(String path) {
+		launcher.setSourceOutputDirectory(path);
+		return this;
+	}
+
+	public void setSourceOutputDirectory(File outputDirectory) {
```
DaveJarvis: Not part of the fluent interface? This could be confusing for some as it doesn't follow the fluent interface idiom.

src/main/java/spoon/FluentLauncher.java
```diff
@@ -0,0 +1,160 @@
...
+		launcher.setOutputFilter(qualifedNames);
+		return this;
+	}
+
+	public FluentLauncher binaryOutputDirectory(String path) {
+		launcher.setBinaryOutputDirectory(path);
+		return this;
+	}
+
+	public void binaryOutputDirectory(File outputDirectory) {
```
DaveJarvis: Does this need to follow the fluent interface idiom?

src/main/java/spoon/FluentLauncher.java
```diff
@@ -0,0 +1,160 @@
...
+		return launcher.createCompiler();
+	}
+
+	// needed?
+	public SpoonModelBuilder createCompiler(List<SpoonResource> inputSources) {
+		return launcher.createCompiler(inputSources);
+	}
+
+	// needed?
+	public Factory createFactory() {
```
DaveJarvis: How does a `Factory` differ from the fluent interface? Could mixing metaphors be confusing?

src/main/java/spoon/FluentLauncher.java
```diff
@@ -0,0 +1,160 @@
...
+	public Environment getEnvironment() {
+		return launcher.getEnvironment();
+	}
+
+	public String toString() {
+		return launcher.toString();
+	}
+
+	// needed???
+	public SpoonModelBuilder compiler(Factory factory) {
```
DaveJarvis: How does a `SpoonModelBuilder` differ from the `FluentLauncher`? Conceptually, to me, they seem like they are trying to accomplish the same task: using a builder pattern to configure Spoon. Exposing both may lead to some confusion.

MartinWitt:
thank you for the review, but the code here is just a draft and far from even worth reading. There are still settings missing, unneeded methods and maybe it's possible to wrap all 3 launcher in the same class. Finding a better name is open too.

surli:
Thanks for starting this @MartinWitt!
Quite frankly, I would use KISS a lot for this: the main purpose of proposing to create a fluent API was to ease using Spoon. 

So I'd just start with some easy well-defined use cases and iterate over them in the launcher. So basically remove everything with the "needed" question, and everything's out of the scope of the usecases expressed in #1563. It should drastically reduce the size of the class. 

The idea is to have a first launcher to fit 90% of usecases, for other option users will still have the old non-fluent API and we can improve it later to add ways to access those options.

src/main/java/spoon/FluentLauncher.java
MartinWitt: yes needes to be changed.

src/main/java/spoon/FluentLauncher.java
MartinWitt: the createFactory method was part of the spoonAPI interface, which i tried to implement first. But the interface has setter and getter, which are not good for the fluent idom

src/main/java/spoon/FluentLauncher.java
MartinWitt: The SpoonModelBuilder interace doesn't expose a fluent api.

src/main/java/spoon/FluentLauncher.java
MartinWitt: yes 100%.

MartinWitt:
as @surli suggested, i made the api slim.
Supported operations:

- setting input resources(String, Iterable<String>)
- adding processors(processor, Iterable<Processor>)
- setting output directory(String, File)
- enabling autoImports
- disableConsistencyChecks
- setting compilanceLevel
- setting sourceClassPath
- activating noClasspath mode
- setting encoding

Are there any imported options someone is missing?

src/main/java/spoon/FluentLauncher.java
MartinWitt: i can't find `buildCtModel` anywhere. SpoonAPI has the method `buildModel` in the interface.
Could you show where the term buildCtModel is used?

surli:
> Incremental build and maven launcher could be supported, with creating a second constructor having a launcher type object as parameter. public FluentLauncher(Launcher launcher) , but would make the class more complex.

I'm +1 to add a support for MavenLauncher since it's already an easy usecase for Spoon and it would show how to extend the capabilities of the FluentLauncher with another existing launcher.

src/main/java/spoon/FluentLauncher.java
DaveJarvis: I was wondering if there's a `Model` and a `CtModel` that can be built. If there's no `Model` class to build, then the question is moot.

MartinWitt:
There were 2 options for adding the mavenlauncher.
Either copy all constructors with arguments(pretty bad) or creating a new constructor ` public FluentLauncher(Launcher launcher)`.  The second option allows using new Launchers in the future and using the IncrementalLauncher too. 
The testcases are only for showing how to use and should be removed afterwards. Is there any need to write real testcases for a wrapper class?

monperrus:
Looks great, we just need a test case, the examples you had are great, could you add them back?

surli:
That's really cool! +1 for a small unit test, could be done with a mock for the wrapped launcher.

MartinWitt:
Maybe #3259 is not compatible with the spoon-control-flow code? The CI failure 

> `[ERROR] /home/travis/build/INRIA/spoon/spoon-control-flow/src/main/java/fr/inria/controlflow/ControlFlowBuilder.java:[122,8] fr.inria.controlflow.ControlFlowBuilder is not abstract and does not override abstract method <T,S>visitCtSwitchExpression(spoon.reflect.code.CtSwitchExpression<T,S>) in spoon.reflect.visitor.CtVisitor`
seems unrelated to this feature.

surli:
(just for info in case you don't know: you shouldn't need a commit to re-run the CI: triggering Travis from the UI is enough since it performs the build by merging changes with master each time)

monperrus:
Thanks a lot @MartinWitt

PR #3260, opened 2020-02-19
fix: not finding sources in non-standard directories

scootafew:
In Maven POM, the standard directory layout for sources and test sources can be overridden from the standard `src/main/java` and `src/test/java` as follows:

```xml
<build>
    <sourceDirectory>source/code</sourceDirectory>
    <testSourceDirectory>test</testSourceDirectory>
</build>
```

The description for the field `sourceDirectory` states:
>This element specifies a directory containing the source of the project. The generated build system will compile the source in this directory when the project is built. **The path given is relative to the project descriptor.**

Previously, when using MavenLauncher, if this value is defined it will be used instead of the standard directory structure when searching for sources. The issue is that this relative path is not prefixed with the absolute path to the project directory. This means that it assumes the current working directory, which is not necessarily the path to the Maven project being analysed.

PR contains previously failing test case with suggested fix.

### DIFF
--- a/src/main/java/spoon/support/compiler/SpoonPom.java
+++ b/src/main/java/spoon/support/compiler/SpoonPom.java
@@ -121,9 +121,10 @@ public class SpoonPom implements SpoonResource {
 			sourcePath = build.getSourceDirectory();
 		}
 		if (sourcePath == null) {
-			sourcePath = Paths.get(directory.getAbsolutePath(), "src", "main", "java").toString();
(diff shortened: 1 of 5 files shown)


monperrus:
Thanks for the PR and the detailed explanation, @nharrand WDYT?

nharrand:
Yes indeed. Makes sense. Thanks for the fix!

scootafew:
No problem! Thanks for the quick review 😄

PR #3261, opened 2020-02-21
chore: Update CFR and spoon core versions in spoon-decompiler

stone-wall:
Wasn't sure if spoon-decompiler is still updated, but I do use it occasionally and have had better success with the newer version of CFR. Also switched spoon dependency version to use 8.1.0


nharrand:
Glad to hear that someone is using it! ;-). But yes this update is welcome!

Issue #3262, opened 2020-02-21
Cannot access child module POM information from SpoonPom


scootafew:
I've been using the `MavenLauncher` class to analyse multi (maven) module projects, but only the top level `SpoonPom` is exposed. The field `modules` on `SpoonPom` is package-private and I would like to suggest adding a getter to allow retrieval of child module POM information (artifactId, version etc.).

Is there any reason why this would be undesirable?

nharrand:
> Is there any reason why this would be undesirable?
No, it is welcome.
Although, we still don't support multi module. In a perfect world, we would support accessing the sources from all modules.

PR #3269, opened 2020-02-22
refactor: adding getter for modules to SpoonPom

scootafew:
As per #3262 I'd like to propose adding a getter to `SpoonPom` so that information about the (maven) modules declared in that POM can be retrieved. Currently only the top level `SpoonPom` is accessible from `MavenLauncher`.


nharrand:
LGTM! Thanks for the PR.

PR #3270, opened 2020-02-23
fix(MavenLauncher): allowing rebuilding of classpath

scootafew:
The class `MavenLauncher` has constructors with parameter `forceRefresh`, with the JavaDoc comment:

> @param forceRefresh force the regeneration of classpath

However the value was actually never passed through to the `SpoonPom` method responsible for generating the classpath, as it was hardcoded to false. This is fixed and a method `rebuildClasspath` is added to `MavenLauncher` to enable rebuilding of the classpath without instantiating a new launcher (or directly calling the `SpoonPom` method). Tests for both use cases also added.

### DIFF
--- a/src/main/java/spoon/MavenLauncher.java
+++ b/src/main/java/spoon/MavenLauncher.java
@@ -125,7 +125,7 @@ public class MavenLauncher extends Launcher {
 		}
 
 		if (classpath == null) {
-			classpath = model.buildClassPath(mvnHome, sourceType, LOGGER, false);
+			classpath = model.buildClassPath(mvnHome, sourceType, LOGGER, forceRefresh);
 		}
 
 		// dependencies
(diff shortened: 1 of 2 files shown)


src/main/java/spoon/MavenLauncher.java
```diff
@@ -134,4 +134,12 @@ private void init(String mavenProject, String[] classpath) {
 		// compliance level
 		this.getEnvironment().setComplianceLevel(model.getSourceVersion());
 	}
+
+	/**
+	 * Triggers regeneration of the classpath
```
monperrus:
```suggestion
	 * Triggers regeneration of the classpath that is used for building the model, based on pom.xml
```

monperrus:
Thanks for the PR, LGTM + see suggestion.

nharrand:
LGTM! Thanks!

PR #3259, opened 2020-02-19
Feature: Java 12 Switch Expressions

Egor18:
Fix #3011
This PR is to support Java 12 Switch Expressions preview feature and new syntax.
More info:
https://openjdk.java.net/jeps/325
https://openjdk.java.net/jeps/354
https://docs.oracle.com/en/java/javase/13/language/switch-expressions.html

Roadmap:
1. Update eclipse JDT [DONE]
2. Support new arrow syntax (`case 1 -> x = 1;`) in switch [DONE]
3. Support multiple expressions (`case 1, 2, 3 -> x = 1;`) in case [DONE]
4. Support switch expression (`int x = switch(i) {...};`) [DONE]

Moreover, Java 13 introduced `yield` keyword to replace break in switch. We should probably add this one later as well.

Note: Right now Switch Expressions are a preview feature, but I believe it will be in standard Java 14.

PR #3271, opened 2020-02-24
fix(spoon-control-flow): add switchExpression to visitor

MartinWitt:
currently the ControlFlowBuilder doesn't override the method `	public <T, S> void visitCtSwitchExpression(CtSwitchExpression<T, S> switchExpression)`. With this fix the CI should work again, as soon as the latest snapshot with the merged switchexpressions #3259 is published on mavencentral. As soon as the new version is published this patch should compile. Updated version of spoon dependency, because CI is using the newest anyway.
Edit: Seems like the CI already has the newest version.

### DIFF
--- a/spoon-control-flow/pom.xml
+++ b/spoon-control-flow/pom.xml
@@ -81,7 +81,7 @@
         <dependency>
             <groupId>fr.inria.gforge.spoon</groupId>
             <artifactId>spoon-core</artifactId>
-            <version>7.3.0-SNAPSHOT</version>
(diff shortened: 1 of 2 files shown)


nharrand:
Thanks for the PR. Could you add a test that covers this?

MartinWitt:
it's only a stub add the not overriden method. An empty method cant be tested or?

monperrus:
That's the right fix to make spoon-control-flow compile, thanks!

PR #3277, opened 2020-02-28
fix: comments should not crash the full process

monperrus:
fix https://github.com/sourcegraph/lsif-java/issues/17


nharrand:
Makes sense.

Issue #1111, opened 2017-01-12
Parallel Version Of EarlyTerminatingScanner


retuxx:
Is it possible to add a parallel version of the `EarlyTerminatingScanner`? This scanner could try to distribute the traversal of the root node to several threads (as many as the underlying machine provides).

What do you think?

monperrus:
I really like this idea.

I've read somewhere that the code  infrastructure at Google does
parallel analysis and transformation, even across multiple machines
using map-reduce.

tdurieux:
An easy solution is to parallelize the scan of the element collections/map.

PR #3280, opened 2020-02-28
feature: add support for parallel processor

### LINKED ISSUE #2659: add support for parallel analysis
for pure read-only analysis we can run the processors on multiple cores (one AST node per core)

MartinWitt:
Fix #2659

The new class allows parallel processing for stateless processors. Subtyping processor allows easy usage in launcher and keeps it simple for the user.
Currently the are still some problems and needed changes:

- [x] writing doc ( and testing)
- [X] easier usage => added constructor with Consumer as argument
- [X] allow usage for non stateless processors => some problems where you need a shared state you could easily share the state between processors by collections. Some problems are never parallel doable. so some work for the user will be left.

wdyt about it?


monperrus:
I think it's really cool!

MartinWitt:
i think the feature is finished if no one sees any problems?

monperrus:
Great progress.

In Spoon, each time we add a feature, we document it in the `doc` folder. Could you document the parallel processor in `processor.md`?

In Spoon, we try to add a line `// contract: the tested contract in natural language` in each test case. Could you do so?

Thanks!

src/main/java/spoon/processing/AbstractParallelProcessor.java
```diff
@@ -0,0 +1,137 @@
...
+	/**
+	 * Creates a new AbstractParallelProcessor from given consumer. The processors
+	 * must follow the guidelines given in the class description.
+	 *
+	 * @param processFunction    Represents an operation that accepts a single
+	 *                           element E and returns no result.
+	 * @param numberOfProcessors number of concurrent running processors.
+	 * @throws IllegalArgumentException if numberOfProcessors is less than 1.
+	 */
+	public AbstractParallelProcessor(Consumer<E> processFunction, int numberOfProcessors) {
```
monperrus: neat API!

MartinWitt: thanks

MartinWitt:
For both failing test cases see #3289 (?) and wait with merges till it is resolved?

monperrus:
Thanks a lot @MartinWitt

monperrus:
For the record, #1111 is related to this.

Issue #3281, opened 2020-03-04
Annotation with named element does not get annotation type package marked implicit


slarse:
Hello,

I ran into an issue where the _package of the annotation type_ of an annotation does not get marked implicit, even though it is. It only happens when named elements are used (e.g. writing `value = "unchecked"` instead of just `"unchecked"`). It does not appear to affect `Launcher.parseClass`, I've only experienced the issue when building an entire model.

I've inlined a minimal test case that reproduces the issue. I had no idea of where to put it for a PR (and I don't have a fix for the problem), so inlining the test case was the best I could come up with.

```java
@Test
public void test_AnnotationWithNamedElement_HasImplicitAnnotationTypePackage() throws Exception {
	Launcher launcher = new Launcher();
	launcher.addInputResource(
            new VirtualFile(
                "class Cls { @SuppressWarnings(value=\"unchecked\") void meth() {} }"));

	CtModel model = launcher.buildModel();
	List<CtElement> elems = model.getElements(e -> e instanceof CtAnnotation);

	assertEquals(1, elems.size());
	CtAnnotation<?> annotation = (CtAnnotation<?>) elems.get(0);
	assertTrue(annotation.getAnnotationType().getPackage().isImplicit());
}
```
It passes fine if you remove the `value=` from the annotation.

slarse:
Now that I think about it, I think this may not be limited to annotations as I've encountered fully qualified names pretty-printed in places that I did not expect them (using `ignoreImplicit=false` in the printer). I unfortunately don't have the output saved, but I will most assuredly encounter the same problems again if they actually exist as I'm operating on a fixed data set.

monperrus:
Thanks for the bug report, and the test case is immensely valuable.

See https://github.com/INRIA/spoon/pull/3285

PR #3283, opened 2020-03-04
test: add annotation for WIP test cases

### LINKED ISSUE #3001: feat: UnresolvedBug annotation
The most useful sort of bug report is actually a pull request with a failing test case. We do have quite a lote of those, and that's great.
But, today we cannot merge them as they would break CI. 
What we could do (thanks @oscarlvp ) is having an annotation for test case `@UnresolvedBug`. We would not run these test in CI, but a regular `mvn test` would run them.

WDYT?

MartinWitt:
Fix #3001

added one marker class and annotation for two features.
`@Category(UnresolvedBug.class) `excludes test from CI but not from maven test.
`@GithubIssue(String)` allows linking your issue number with a (failing) testcase.

Added check for every GitHub annotation, that an issue with the number exists and every testcase with an Category Annotation has an Github annotation.
Checking if the issue is open would be possible, but maybe you want to link a closed issue?

For showcasing the new feature i add the testcase from #3281.
If someone likes it, it could be included in the CONTRIBUTING guide.

The new feature needed a json parser. Gson was pretty easy to use, but could be changed if someone wants to include a different one.
Lastly both new files need an exclusion from the SpoonArchitectureEnforcerTest::testGoodTestClassNames because having test in the classnames would harm the readability. 

wdyt?

### DIFF
--- a/src/test/java/spoon/test/GitHubIssue.java
+++ b/src/test/java/spoon/test/GitHubIssue.java
@@ -3,16 +3,17 @@
  *
  * Spoon is available either under the terms of the MIT License (see LICENSE-MIT.txt) of the Cecill-C License (see LICENSE-CECILL-C.txt). You as the user are entitled to choose the terms under which to adopt Spoon.
  */
-package spoon;
+package spoon.test;
 
 import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
-//TODO: check all annotations if issue is open automatically
+@SuppressWarnings("CheckStyle")
+
 @Retention(RetentionPolicy.RUNTIME)
 @Target(ElementType.METHOD)
-public @interface GithubIssue {
+public @interface GitHubIssue {
 	String issueNumber();
 }
--- a/src/test/java/spoon/test/GitHubIssue.java
+++ b/src/test/java/spoon/test/GitHubIssue.java
@@ -10,7 +10,6 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
-@SuppressWarnings("CheckStyle")
 
 @Retention(RetentionPolicy.RUNTIME)
 @Target(ElementType.METHOD)
--- a/src/test/java/spoon/test/GitHubIssue.java
+++ b/src/test/java/spoon/test/GitHubIssue.java
@@ -14,5 +14,5 @@ import java.lang.annotation.Target;
 @Retention(RetentionPolicy.RUNTIME)
 @Target(ElementType.METHOD)
 public @interface GitHubIssue {
-	String issueNumber();
+	int issueNumber();
 }
--- a/src/test/java/spoon/test/UnresolvedBugTest.java
+++ b/src/test/java/spoon/test/UnresolvedBugTest.java
@@ -45,7 +45,7 @@ public class UnresolvedBugTest {
 			// because readAllBytes is jdk9 only
 			String data = new BufferedReader(new InputStreamReader(url.openStream())).lines().collect(Collectors.joining());
 			JsonObject issue = new Gson().fromJson(data, JsonObject.class);
-			assertTrue(issue.get("number").getAsString().equals(issueNumber));
+			assertTrue(issue.get("number").getAsInt()==issueNumber);
 		}
 	}
 
--- a/src/test/java/spoon/test/UnresolvedBugTest.java
+++ b/src/test/java/spoon/test/UnresolvedBugTest.java
@@ -40,7 +40,7 @@ public class UnresolvedBugTest {
 				.filter(v -> v.hasAnnotation(Test.class) && v.hasAnnotation(GitHubIssue.class))
 				.collect(Collectors.toList());
 		for (CtMethod<?> ctMethod : testMethods) {
-			String issueNumber = ctMethod.getAnnotation(GitHubIssue.class).issueNumber();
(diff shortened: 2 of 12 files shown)


monperrus:
Super cool!

What about simplifying and having a single annotation:

```
@Category(UnresolvedBug.class)
@GitHubIssue(issueNumber = "3281")
```

becomes 

```
@UnresolvedBug(issueNumber = "3281")
```
?

MartinWitt:
Would be better, but for ignoring tests in the CI the [junit category](https://junit.org/junit4/javadoc/4.12/org/junit/experimental/categories/Category.html) seems the clean way in junit 4.12.
Not using this annotation could be possible, but would make different testing in CI and local very hacky (see [https://stackoverflow.com/questions/37603075/create-annotation-for-junit-category](url). Upgrading to junit 5 would make it possible [see](https://junit.org/junit5/docs/current/user-guide/#writing-tests-meta-annotations)

monperrus:
We already use junit5 (see pom.xml). One solution may be that we require the annotation to be put on
Junit5 tests only.

MartinWitt:
I sadly wasn't able to filter and run tests via junit5 tags. Maybe someone is more experienced in junit5 and maven surefire.

monperrus:
I see.

Does it work with the category only?

Plus `UnresolvedBug` is both a marker and a test class containing tests, which is confusing. Could we split them in `UnresolvedBug` and `UnresolvedBugTest`?

As a marker, could `UnresolvedBug` be an interface?

MartinWitt:
I wasn't able to get it running using junit5 tags.
Splitting the class in marker interface and test class is a good point.
Added one sentence in the contributing guide for failing test cases.

src/test/java/spoon/test/GitHubIssue.java
```diff
@@ -0,0 +1,18 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.METHOD)
+public @interface GitHubIssue {
+	String issueNumber();
```
monperrus: what about returning an int instead of a String?

src/test/java/spoon/test/architecture/SpoonArchitectureEnforcerTest.java
```diff
@@ -232,6 +232,9 @@ public void metamodelPackageRule() {
 
 	@Test
 	public void testGoodTestClassNames() {
+		//some testclasses are util only and dont have to follow the naming convention for better readability
+		Set<String> testUtilClasses = new HashSet<>(Arrays.asList("spoon.test.GitHubIssue", "spoon.test.UnresolvedBug"));
```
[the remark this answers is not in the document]
MartinWitt: 100% correct. Changed it

src/test/java/spoon/test/GitHubIssue.java
MartinWitt: Sounds good to me. We can safely assume github issue numbers fit in int.

src/test/java/spoon/test/UnresolvedBugTest.java
```diff
@@ -0,0 +1,74 @@
...
+		testMethods = testMethods.stream()
+				.filter(v -> v.hasAnnotation(Test.class) && v.hasAnnotation(GitHubIssue.class))
+				.collect(Collectors.toList());
+		for (CtMethod<?> ctMethod : testMethods) {
+			int issueNumber = ctMethod.getAnnotation(GitHubIssue.class).issueNumber();
+			URL url = new URL(githubURL + issueNumber);
+			// because readAllBytes is jdk9 only
+			String data = new BufferedReader(new InputStreamReader(url.openStream())).lines().collect(Collectors.joining());
+			JsonObject issue = new Gson().fromJson(data, JsonObject.class);
+			assertTrue(issue.get("number").getAsInt()==issueNumber);
```
monperrus: +check that the issue is still open? (looks straighforward to do)

MartinWitt: added

monperrus:
Thanks a lot @MartinWitt for this cool testing feature!

monperrus:
Master is broken: https://travis-ci.org/INRIA/spoon/jobs/660725438

A flaky test? WDYT?

MartinWitt:
is there some way to get the full error message? Maybe using the github API in a ci wasn't a good idea. Possible fix would be remove the test, because a flaky test is really bad.

monperrus:
Upper in the log:

```
[INFO] Running spoon.test.UnresolvedBugTest
[ERROR] Tests run: 2, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 13.449 s <<< FAILURE! - in spoon.test.UnresolvedBugTest
[ERROR] checkGithubIssueAnnotations(spoon.test.UnresolvedBugTest)  Time elapsed: 0.383 s  <<< ERROR!
java.io.IOException: Server returned HTTP response code: 403 for URL: https://api.github.com/repos/INRIA/spoon/issues/3281
	at spoon.test.UnresolvedBugTest.checkGithubIssueAnnotations(UnresolvedBugTest.java:47)
```

MartinWitt:
403 Error means either rate limit or missing user agent. Looking at the other CI tests i would guess rate limit exceeded [see](https://developer.github.com/v3/). For unregistered users 60 API calls per hour per IP. Guessing that we share the CI IP with other users, it's a bad test. 
So either have a oauth token in the CI ( is this possible?), check before the test the rate limit or remove the test?

monperrus:
On master, there is an environment variable GITHUB_AUTH_TOKEN that we can use in the test case.

PR #3286, opened 2020-03-05
feature(SpoonPom): expose result of maven classpath build

scootafew:
When the classpath is built in `SpoonPom` it would be useful to know if the Maven invocation was successful. Invocation was successful if exit code was 0, or failed otherwise. Unfortunately I can't find a way to relay information about the reason for failure, so for now a simple boolean flag would allow action to be taken if the classpath was not built successfully.


monperrus:
Thanks, could you add a test case?

scootafew:
Sure, test added!

scootafew:
Actually, my reason for wanting this was due to the fact I appeared to be seeing behavior where the classpath temp file was generated even if Maven invocation failed. I can now see this was an error on my part so I'm not sure this PR adds much value, as the same result can be achieved by simply checking for the existence of the temp file. As such I'm going to go ahead and close this.

monperrus:
ack, thanks for the proposal.

PR #3292, opened 2020-03-06
fix documentation

peroksid90:


peroksid90:
Probably the same for

`src/main/java/spoon/reflect/code/CtArrayWrite.java:20: * If you process this element, keep in mind that you won't process array[0]++.
src/main/java/spoon/reflect/code/CtArrayRead.java:19: * If you process this element, keep in mind that you will process array[0]++ too.
src/main/java/spoon/reflect/code/CtVariableWrite.java:23: * If you process this element, keep in mind that you won't process var++.
src/main/java/spoon/reflect/code/CtVariableAccess.java:19: * If you process this element, keep in mind that you will process var++ too.`

monperrus:
Thanks for your contribution.

I think this line was and is still confusing, I propose to simply remove it.

nharrand:
Thanks!

PR #3303, opened 2020-03-12
refactor: Test can be annotated with @Ignore("UnresolveBug")

nharrand:
Instead of having a specific annotation `@UnresolvedBug`, we just use `@Ignore` with a specific value `"UnresolvedBug"`.

This way, by default maven does not run these tests. But we have a test `UnresolvedBugTest#checkThatUnresolvedBugTestFail()` that do run them and check that they are failing. Previous test that checked that all of these UnresolvedBug testcase do have a corresponding `@GitHubIssue` tag remains, but `checkGithubIssueAnnotations` is now ignored because it was flacky in CI.

Here is an example: 
```java
@Test
@GitHubIssue(issueNumber = 3281)
@Ignore("UnresolvedBug")
public void test_AnnotationWithNamedElement_HasImplicitAnnotationTypePackage() throws Exception {
	launcher.addInputResource(new VirtualFile("class Cls { @SuppressWarnings(value=\"unchecked\") void meth() {} }"));
	List<CtElement> elems = model.getElements(e -> e instanceof CtAnnotation);
	assertEquals(1, elems.size());
	CtAnnotation<?> annotation = (CtAnnotation<?>) elems.get(0);
	assertTrue(annotation.getAnnotationType().getPackage().isImplicit());
}
```


src/test/java/spoon/test/UnresolvedBugTest.java
```diff
@@ -12,31 +12,52 @@
...
-	private List<CtMethod<?>> testMethods = findTestMethods();
+	private static List<CtMethod<?>> testMethods = findTestMethods();
 	private final String githubURL = "https://api.github.com/repos/INRIA/spoon/issues/";
 
+
+	@BeforeClass
+	public static void setup() {
+		testMethods = testMethods.stream()
+				.filter(v -> v.hasAnnotation(Test.class) &&  v.hasAnnotation(Ignore.class)
+						&& v.getAnnotation(Ignore.class).value().equals("UnresolvedBug"))
```
MartinWitt: how about https://docs.oracle.com/javase/7/docs/api/java/lang/String.html#equalsIgnoreCase(java.lang.String) instead of `equals`?

nharrand: Sure! Why not!

MartinWitt:
I'm unsure about the magic string in `@Ignore( "UnresolvedBug")`, but this change looks good. Could you change the CI scripts back (removing the new -Dexcludegroups flag?

nharrand:
> I'm unsure about the magic string in @Ignore( "UnresolvedBug")

I did not find a way to make maven ignore the `@UnresolvedBug` by default. And I don't like the idea of haveing `mvn test` fail if not run with a flag, I think it is a bit annoying for newcommers. Hence this proposition for a compromise.

> Could you change the CI scripts back (removing the new -Dexcludegroups flag?

Yes most defenitly!

MartinWitt:
Having an enum/class with constants could be possible, if we decide we need some more types like, "requiresNetwork", "win/unix only", "requires jdkX" etc. But for the moment this seem like a good solution. Great work.

nharrand:
Oh I see. Yes, if we have several categories, I agree that a String is not necessarily the smartest choice.

monperrus:
LGTM, really cool refinement of the @Ignore concept.

I will merge it.

monperrus:
Thanks @nharrand

PR #3307, opened 2020-03-16
 WIP: feature: forbid raw types

MartinWitt:
Currently spoon still uses a lot raw types like `List list`. Raw types should not be used, see [effective java](https://www.informit.com/articles/article.aspx?p=2861454). With the following test we enforce the usage of at least `List<?> list`. The test only checks the src/main code, because for testing raw types are sometimes needed. If someone else likes it I would fix all raw types in the code (around 600) and include it to this PR.

wdyt?


monperrus:
It's a good idea.

monperrus:

Thanks!

PR #3314, opened 2020-03-25
feature: add support for java 14 switch/yield

MartinWitt:
# Java 14 Switches # 
[Jep361](https://openjdk.java.net/jeps/361) defines a new switchExpression.
```java
switch (day) {
    case MONDAY, FRIDAY, SUNDAY -> System.out.println(6);
    case TUESDAY                -> System.out.println(7);
    case THURSDAY, SATURDAY     -> System.out.println(8);
    case WEDNESDAY              -> System.out.println(9);
}
```
@Egor18 implemented them already without yield( awesome work btw).
This pr adds support for java14 switches with yield and with this change spoon supports all java14 features that are non preview(records and textblocks are only part of jdt beta branch and still preview)

## Whats part of this PR ##

- Remove break [expression] and java 12 preview switches
- add support for yieldStatements in switch
- upgrade jdt version 3.20
- fully refactor switchCaseTest to junit5

## Consequences ##

There are 2 consequence expect more code.
#### spoon loses the ability to parse java 12 preview code with break [expression] ####
Currently spoon supports them, but i would instantly drop support for a feature that was never part java (only preview) and wont come back. `Yield` replaces this. 
Supported preview features should be seen as an experimental feature and i hope no one is relaying on them.
####junit5 in testclasses####
I cant guarantee that every testrunner in everyones IDE fully supports junit5.
I would be nice if you could test the refactored SwitchCaseTest.
Reward is a much clearer design for the testclasses.

### upgrading jdt to 3.20 produced 2 problems ###
Luckily there were only 2 visible problems after upgrading it.
#### `break.isImplicit` flag was removed #### 
wasn't a big problem, because there are no longer needed.

#### PackageBinding.java was changed to a abstract class ####
Is a bigger problem, but current there is a fix.
For (potential) fix see src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java#363
Maybe someone with more knowledge about this class could review it.  

## interesting more academic fact ##
Is allowed code
```java
int a = switch("a){
default -> {yield 3;}
};
```
but the following is not allowed.
```java
int a = switch("a){
default -> yield 3;
};
```
This behavior seems intended, because looking at java grammar we find:
SwitchRule:
SwitchLabel -> Expression ;
SwitchLabel -> Block
SwitchLabel -> ThrowStatement
and yield is not an expression but a StatementWithoutTrailingSubstatement from my understanding.
## tl;dr; ##

- spoon supports java 14 switches with yield
- If you have some knowledge about some parts of spoon feel free to review the new code with new tests


src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
```diff
@@ -358,8 +360,14 @@ CtReference getDeclaringReferenceFromImports(char[] expectedName) {
 						if (packageBinding == null || packageBinding instanceof ProblemPackageBinding) {
 							// Big crisis here. We are already in noclasspath mode but JDT doesn't support always
 							// creation of a package in this mode. So, if we are in this brace, we make the job of JDT...
-							packageBinding = new PackageBinding(chars, null, environment, environment.module);
-						}
+							packageBinding = new PackageBinding(chars, null, environment, environment.module) {
+								@Override
+								public PlainPackageBinding getIncarnation(ModuleBinding arg0) {
+									//https://github.com/eclipse/eclipse.jdt.core/blob/master/org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/lookup/PlainPackageBinding.java#L43
```
MartinWitt: #3314 just as a link

monperrus:
That's a serious PR, thanks a lot! Will look at it in detail.

src/main/java/spoon/reflect/visitor/CtAbstractVisitor.java
```diff
@@ -523,4 +524,9 @@ public void visitCtPackageDeclaration(CtPackageDeclaration packageDeclaration) {
 	public void visitCtTypeMemberWildcardImportReference(CtTypeMemberWildcardImportReference wildcardReference) {
 
 	}
+
+	@Override
+	public void visitCtYieldStatement(CtYieldStatement statement) {
+		String foo = "";
```
monperrus: to be removed

src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
monperrus: I don't understand, what do you mean?

src/main/java/spoon/reflect/visitor/CtAbstractVisitor.java
MartinWitt: done

src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
MartinWitt: the class `PackageBinding ` was changed to an abstract class. To fix the problem, that you cant create an instance of it directly, i created an anonymous class. But the class needs to override the `public PlainPackageBinding getIncarnation(ModuleBinding arg0)` Method. I linked the original method from `PlainPackageBindung` to explain why this method always returns null.

monperrus:
Thanks. Could you add this explanation in the source code directly?

monperrus:
Thanks a lot @MartinWitt it's awesome that we now support Java 14!

Issue #3329, opened 2020-04-16
Static field access sometimes get target as CtThisAccess instead of CtTypeAccess (noclasspath)


slarse:
Hello!

I've encountered some bugs with statically accessed variables. This was the most easily reproducible one, I'll try to reproduce the other ones later.

Anyway, my problem is that sometimes when a static field is accessed from a class for which the source is not available, the target of the field read is a `CtThisAccess` instead of a `CtTypeAccess`. This results in some pretty gnarly problems for me. I found this old issue discussing something similar: #346 

Here's an example:

```java
import spoon.Launcher;

public class Main {
    public static void main(String[] args) {
        Launcher.logger.error("Hello :)");
    }
}
```

Here, I expect the target of the `logger` field read to be a type access to the `Launcher` type. But it's a `CtThisAccess`. So if I just parse this and immediately print it, this is what I get.

```java
import spoon.Launcher;

public class Main {
    public static void main(String[] args) {
        logger.error("Hello :)");
    }
}
```

A this access in a static context doesn't make sense to begin with, and what makes even less sense is that the target of the `CtThisAccess` is the `Launcher` type. So it's a strange cross-type this access in a static context. The problem persists in a non-static methods as well.

If I add a `Launcher` class in the same file, as shown below, then it correctly resolves a type access instead of a this access.

```java
public class Main {

    private static class Launcher {
        static Logger logger;
    }

    public static void main(String[] args) {
        Launcher.logger.error("Hello :)");
    }
}
```

I'll have a PR with this test case up in a few minutes.

Egor18:
After some research, I figured out that despite being in noclasspath mode spoon still loads `spoon.Launcher` class from classloader:
https://github.com/INRIA/spoon/blob/1de9318fcb97fa99ba18c3fd40025163a55df89b/src/main/java/spoon/support/compiler/jdt/JDTImportBuilder.java#L150-L154

The thing is there is no `logger` field in `spoon.Launcher`, there is `LOGGER`, so this drives it a bit crazy.
You can try renaming it to `LOGGER` or replacing `import spoon.Launcher` with `import abc.Launcher` or something to see it yourself.
Anyway, I provided a PR with a possible fix for this corner case.

monperrus:
Thanks a lot to @slarse for the detailed bug report and to @Egor18 for the super fast analysis & fix!

slarse:
Great, thanks @Egor18 @monperrus

PR #3330, opened 2020-04-16
WIP: fix: Add failing test case for static field read

slarse:
This is a test case to showcase the problem in #3329


monperrus:
thanks a lot for the failing test case, that was super helpful.

closed per #3331

PR #3331, opened 2020-04-16
fix: fix missing CtTypeAccess reported in #3329

### LINKED ISSUE #3329: Static field access sometimes get target as CtThisAccess instead of CtTypeAccess (noclasspath)

Egor18:
Fix #3329

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java
+++ b/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java
@@ -659,7 +659,11 @@ public class JDTTreeBuilderHelper {
 		} else if (ref.isStatic()) {
 			target = createTypeAccess(qualifiedNameReference, ref);
 		} else if (!ref.isStatic() && !ref.getDeclaringType().isAnonymous()) {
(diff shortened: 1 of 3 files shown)


monperrus:
LGTM

monperrus:
Thanks a lot!

PR #3335, opened 2020-04-18
fix: fix nested class field read problem

### LINKED ISSUE #3334: Nested non-static class this access to enclosing type's non-static field turns into qualified static field access (noclasspath)
Hello!

Unfortunately, the patch in #3331 appears to have caused the reverse issue of what it attempted to solve when it comes to nested classes. Essentially, a nested non-static class making an unqualified access to **a field of an enclosing type's non-static field** results in a qualified _static_ field access, which is incorrect. I know the description is a bit of a headache, here's an example.

```java
public class Outer {
    SomeClass cls = new SomeClass();

    private class Inner {
        public testMethod() {
            int a = cls.var;
        }
    }
}
```

The access to `cls.var` resolves to the static access `Outer.cls.var`, which is clearly incorrect.

The problem here is that the inner class is built, in its entirety, before the `cls` field is built. This results in the `Outer` class not actually having the the `cls` field at the time the field read at `cls.var` is resolved, causing it to fall into the then-branch of the if-statement below.

https://github.com/INRIA/spoon/blob/e060bd250f46d3287a5f87f6b4ddbdbb26b44fa3/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java#L662-L666

This results in the field read becoming `Outer.cls.var` instead of `cls.var` with an implicit `this`.

I believe a fix for this can be attained by checking what #3331 checks in the JDT tree instead of in the Spoon tree.

I think I've managed to fix the problem, just testing some stuff before I publish a PR.

slarse:
Fix #3334

Hi!

This is a suggested fix for #3334. To the best of my knowledge, this does the same thing that @Egor18  implemented with #3331, but using the JDT tree instead.

Please see #3334 for a detailed description of the problem. I'm also running some tests locally with [Spork](https://github.com/kth/spork) to see if this solves the problems I encountered, will check back once those have completed.

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java
+++ b/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java
@@ -659,7 +659,7 @@ public class JDTTreeBuilderHelper {
 		} else if (ref.isStatic()) {
 			target = createTypeAccess(qualifiedNameReference, ref);
 		} else if (!ref.isStatic() && !ref.getDeclaringType().isAnonymous()) {
-			if (ref.getDeclaringType().getDeclaredField(ref.getSimpleName()) == null) {
+			if (!JDTTreeBuilderQuery.isResolvedField(qualifiedNameReference)) {
 				target = createTypeAccessNoClasspath(qualifiedNameReference);
 			} else {
 				target = jdtTreeBuilder.getFactory().Code().createThisAccess(jdtTreeBuilder.getReferencesBuilder().<Object>getTypeReference(qualifiedNameReference.actualReceiverType), true);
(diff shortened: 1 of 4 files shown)


slarse:
This fix caused further issues for me. Will keep debugging.

slarse:
After extensive testing, I've found that the further issues I encountered were caused by Spoon incorrectly marking some package references as implicit (that are in fact explicit), and vice versa. I'm currently trying to figure out whether the two problems are related or not.

monperrus:
Thanks a lot, merging this one which is atomic, CI green and LGTM.

#3331 introduced a major regression see https://ci.inria.fr/sos/

Let's how this better fix improve the situation.

monperrus:
@slarse your fix was effective, all the broken full-project transfos are back to green: https://ci.inria.fr/sos/

thanks again.

slarse:
@monperrus Awesome, glad I could help!

PR #3338, opened 2020-04-19
fix: correct package assumption when there is a star import

### LINKED ISSUE #3337: Type access to class not on classpath causes Spoon to assume it belongs to compilation unit's package (noclasspath)
Hello!

Another hour, another bug. I've begun to understand enough of how Spoon works to debug these things now :)

The problem I have is that if I have a type access to a type that's not on the classpath (e.g. to access static variable or method), then Spoon does two things wrong:

1. It assumes that the type belongs to the compilation unit's package. This assumption is invalid if there is any `*` import from a package that's not on the classpath. It was introduced to fix #1293.
2. It makes the reference explicit, causing it to show in pretty-printing.

In conjunction, 1 and 2 can cause compile errors as a class that was imported with a `*` import may suddenly be qualified with the wrong package. The place where it all happens is here:

https://github.com/INRIA/spoon/blob/e7d42138db21a59beff566af22d72aca4141ddbc/src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java#L1053-L1064

The second point also affects all cases where the JDT does not provide info about something belonging to `java.lang` (see #1307), i.e. that the `java.lang` package statement suddenly becomes explicit. This alone isn't very serious as the code still compiles fine, it's just a bit annoying.

Here's a simple example of where there are issues:

```java
package bug;

import some.pkg.*;
import some.other.pkg.*;

public class StaticMemberAccess {
    int someValue = SomeClass.generateValue();
}
```

Spoon will attach an explicit package reference to `bug` to `SomeClass`, but it could be in any of `bug`, `some.pkg` or `some.other.pkg`, there is no way to tell. So, one needs to check if there is any `*` import from a package that's not resolved before assuming that the compilation unit's package should be used, and I also think that the package should be marked implicit both in the case of the missing `java.lang` thing, and when the package is guessed.

I'm not sure what to do when the package cannot be determined however, is it allowed to set the package as `null`?

I'm working on a PR for this.

slarse:
Fix #3337 

Hello!

This is an attempt to fix Spoon incorrectly guessing the package of a type in noclasspath mode. See #3337 for details.

I need some help with this PR, see the comments.

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
+++ b/src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
@@ -1079,7 +1079,7 @@ public class ReferenceBuilder {
 	}
 
 	private static boolean containsStarImport(ImportReference[] imports) {
-	    return imports != null && Arrays.stream(imports).anyMatch(imp -> imp.toString().endsWith("*"));
+		return imports != null && Arrays.stream(imports).anyMatch(imp -> imp.toString().endsWith("*"));
 	}
 
 	/**
--- a/src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
+++ b/src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
@@ -1061,8 +1061,8 @@ public class ReferenceBuilder {
 				ref.setPackage(javaLangPackageReference);
 			} catch (NoClassDefFoundError | ClassNotFoundException e) {
 				ContextBuilder ctx = jdtTreeBuilder.getContextBuilder();
-				boolean isNoClasspath = jdtTreeBuilder.getFactory().getEnvironment().getNoClasspath();
-				if (isNoClasspath && containsStarImport(ctx.compilationunitdeclaration.imports)) {
+				assert jdtTreeBuilder.getFactory().getEnvironment().getNoClasspath();
+				if (containsStarImport(ctx.compilationunitdeclaration.imports)) {
 					// If there is an unresolved star import in noclasspath,
 					// we can't tell which package the type belongs to (#3337)
 					CtPackageReference pkgRef = jdtTreeBuilder.getFactory().Core().createPackageReference();
--- a/src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
+++ b/src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
@@ -82,6 +82,7 @@ import spoon.reflect.reference.CtVariableReference;
 import spoon.reflect.reference.CtWildcardReference;
 import spoon.support.reflect.CtExtendedModifier;
 
+
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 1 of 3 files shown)


src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
```diff
@@ -1059,14 +1060,28 @@ void setPackageOrDeclaringType(CtTypeReference<?> ref, CtReference declaring) {
 				javaLangPackageReference.setSimpleName("java.lang");
 				ref.setPackage(javaLangPackageReference);
 			} catch (NoClassDefFoundError | ClassNotFoundException e) {
-				// in that case we consider the package should be the same as the current one. Fix #1293
-				ref.setPackage(jdtTreeBuilder.getContextBuilder().compilationUnitSpoon.getDeclaredPackage().getReference());
+				ContextBuilder ctx = jdtTreeBuilder.getContextBuilder();
+				boolean isNoClasspath = jdtTreeBuilder.getFactory().getEnvironment().getNoClasspath();
+				if (isNoClasspath && containsStarImport(ctx.compilationunitdeclaration.imports)) {
```
slarse, line 1065: I'm not sure we need to check if it's noclasspath mode, we shouldn't get here in classpath mode? Perhaps `assert isNoClasspath` instead?

src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
```diff
@@ -1059,14 +1060,28 @@ void setPackageOrDeclaringType(CtTypeReference<?> ref, CtReference declaring) {
...
+					ref.setPackage(ctx.compilationUnitSpoon.getDeclaredPackage().getReference());
+				}
 			}
 		} else {
 			throw new AssertionError("unexpected declaring type: " + declaring.getClass() + " of " + declaring);
 		}
 	}
 
+	private static boolean containsStarImport(ImportReference[] imports) {
+	    return imports != null && Arrays.stream(imports).anyMatch(imp -> imp.toString().endsWith("*"));
```
slarse, line 1082: This will return true if there's _any_ star import, which is too restrictive. For example, if there's a `java.util.*`, then this will return true and we won't assign the CU's package to the type. But if the star import is actually available, which should be the case for stdlib packages, depending on JDK version, then it should be fine to guess assign the CU's package to the type.

Is there any way to check which imported packages are actually available at this stage?

src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
monperrus: > Perhaps assert isNoClasspath instead?

good idea.

src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
monperrus: > Is there any way to check which imported packages are actually available at this stage?

Do you mean which subpackages under `.**`are available in the classpath?

slarse: No, I was rather thinking of just checking if the containing package is available. For example, if any type from `java.util` is available on the classpath, then we conclude that all types from `java.util.*` are available on the classpath. If that is the case, we assume that had `SomeClass` actually belonged to `java.util`, it would too have been on the classpath too, and since it is not we assign it to the compilation unit's package.

But as I typed the above paragraph I realized a few things. First, third party packages are often split across projects leading to packages possibly being partially available, so the assumption "all types from the `x.y.*` import are available if any type from package `x.y` is available" is not necessarily true (think e.g. `org.apache`). Second, differing versions of the standard library may cause issues. Third, accounting for what I just pointed out makes this incredibly complex.

I think the simple and probably sufficient solution is to simply not guess the package if the type is not on the classpath, and there is any `*` import.

slarse:
@monperrus I'm not sure what the cleanest way to resolve this is. The current state of the PR is to just create a new, empty package reference whenever there is a star import. But that coincides with how unnamed packages are implemented, which is not what this is. it's rather an _unknown_ package.

While it works to have unknown packages simply be an unnamed package, it seems like there should be a distinction between an unnamed package and an unknown package, but I'm not sure of the best way to do that.

Then again, as this is an isolated problem in noclasspath mode when there also happens to exist star imports in the file, simply having the unknown package default to an unnamed package is acceptable if it's properly documented.

monperrus:
To me, it's acceptable this way. I generally prefer not to over-engineer for corner-cases.

If you unWIP, and nobody is against it, I'll merge.

slarse:
Done.

slarse:
Not sure why the build failed, should I try to merge master into this branch?

https://travis-ci.org/github/INRIA/spoon/jobs/683288854

monperrus:
The CI error is not related.

So we merge, thanks a lot for your contribution.

monperrus:
FYI, CI fix at #3356

Issue #3339, opened 2020-04-21
Don't display warning when some processor properties are optional


vmassol:
Right now, I get:

```
warning: No value found for property 'componentsTxtPath' in processor ...  
```

And my processor has:

```
    @Property
    private String componentsTxtPath;
...
    private Path getComponentsTxtPath(SourcePosition position)
    {
        String path = this.componentsTxtPath;
        if (path == null) {
            path = "target/classes/META-INF/components.txt";
        }
        return Paths.get(computeMavenModulePath(position), path);
    }
```

As you can see I handle the case when it's null in the processor and thus I'd like this property to be optional and not have the warning displayed.

Thanks!

monperrus:
Thanks for the bug report! I fully agree with you, that's not a useful warning.

It seems to be an easy PR to fix that, would you give it a try?

Thanks!

MartinWitt:
looks fixed and can be closed?

monperrus:
Thanks @MartinWitt

PR #3342, opened 2020-04-22
Feat: add notNullable property for processor properties

MartinWitt:
See #3339 for discussion why. I would add a simple boolean field to the property annotation because:
- Change the null to a nullObject like in SourcePosition would be backwards breaking
- Doesn't pollute the namespace with a new method.

Maybe a open issue issue is where to document it? 

Edit: need to fix git sry
Edit2: git fixed

### DIFF
--- a/src/main/java/spoon/processing/Property.java
+++ b/src/main/java/spoon/processing/Property.java
@@ -24,7 +24,7 @@ public @interface Property {
 	 */
 	String value() default "";
 	/**
-	 * If a value is nullable, spoon will expect that the user handles null values and will not print errors.
+	 * If a value is notNullable, spoon will throw a {@link spoon.SpoonException} for null values.
 	 */
-	boolean nullable() default false;
+	boolean notNullable() default false;
 }
--- a/src/main/java/spoon/processing/Property.java
+++ b/src/main/java/spoon/processing/Property.java
@@ -23,4 +23,8 @@ public @interface Property {
 	 * An optional text that describes the property.
 	 */
 	String value() default "";
+	/**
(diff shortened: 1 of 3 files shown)


monperrus:

The idea is interesting, but it does not make the log more useful.

I would propose to invert the annotation nullable -> notNullable.

If notNullable==true and properti is null we throw a SpoonException.

WDYT?

monperrus:
ping @vmassol

vmassol:
I think it's not the role of the spoon framework to handle that and that's the jib of the Processor to tell what it needs and to report when it doesn't get what it needs to work.

MartinWitt:
@monperrus but wouldn't your change break the old behavior? If a user updates and gets your version, he could get exceptions for code working before. 
Is this "issue" so important, that you would throw away backwards compatibility?

@vmassol Your message reads like you would remove the log message completely, would the proposed change still fit your needs?

vmassol:
> @vmassol Your message reads like you would remove the log message completely,

Yes I'd remove it completely and let the processor do the validation.

>  would the proposed change still fit your needs?

Sure.

monperrus:
> @monperrus but wouldn't your change break the old behavior?

No, it's fully backward compatible if properties are still nullable by default.

MartinWitt:
Changed nullable -> notNullable and added the SpoonException. Should we still write into the log, even if there is a exception thrown?

monperrus:
Great.

> Should we still write into the log, even if there is a exception thrown?

no

could you add the `//contract ` line in each test?

with that, we'll go towards merge

thanks!

MartinWitt:
github seems to have some problems, but the CI looks green.

PR #3350, opened 2020-04-30
chore: spoon-dataflow depends on latest spoon-core

monperrus:


MartinWitt:
you could replace the version with a "+" in gradle e.g. 
`implementation group: 'fr.inria.gforge.spoon', name: 'spoon-core', version: '+'`. Then would gradle search the newest version automatically.

monperrus:
Thanks @MartinWitt, did that. There is another problem to fix but I have no time to look at it for now.

PR #3351, opened 2020-04-30
test(UnresolvedBugTest): do not override the content of testMethods

nharrand:
Each unit test was working separatly, but running the whole class skipped two of them...


monperrus:
Thanks!

PR #3352, opened 2020-04-30
WIP fix(ElementSourceFragment): fix fragment handling of duplicated annotations

nharrand:
…ations

### DIFF
--- a/src/main/java/spoon/support/sniper/internal/ElementSourceFragment.java
+++ b/src/main/java/spoon/support/sniper/internal/ElementSourceFragment.java
@@ -245,6 +245,8 @@ public class ElementSourceFragment implements SourceFragment {
 							 * it can happen for annotations of type TYPE_USE and FIELD
 							 * In such case the annotation belongs to 2 elements
 							 * And one of them cannot have matching source position - OK
+							 *
+							 * See https://github.com/INRIA/spoon/issues/3321
 							 */
 							return null;
 						}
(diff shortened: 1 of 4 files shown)


nharrand:
I am not sure I am happy with this fix. 

Concretly the problem we have it that the following method declaration is represented in spoon with a duplicated `@NotNull` annotation.

```java
public JavaxImportTestSource setObject(@NotNull SomeObjectDto someObjectDto)
```

The annotation is put on both the tyep reference `SomeObjectDto` and the parameter `SomeObjectDto someObjectDto`. (The Annotation element is cloned.) This lead to an issue as the type reference has a child element which position is not included in its own position. Hence the SnipperPrettyPrinter failing.

This fix hide the error and makes the SnipperPrettyPrinter not crash.

WDYT? Any opinion? Is it better not to fix until we have a better idea?

monperrus:
Thanks for the analysis.

This is the definition of a workaround.

Pragmatically, I'm OK to merge it as long as long as we have a well explained open issue about it and a link to the issue in the source code.

Could you open the issue and document it in code?

monperrus:
ping @nharrand

nharrand:
The thing I'm not completely clear about is how does this case relates to Java language specification. Should the annotation be indeed cloned and added to both the parameter and the type reference, or is it a workaround of spoon?
Because in the first case, we have (to my understanding) in the position of the type reference (it should include the annotation's position). In the second case, we need a workaround for the workaroud, which can be what is implemented in this PR AFAIU.

nharrand:
See #3359

Issue #3358, opened 2020-05-07
Annotations source position handling on paramters and parameter type annotation.


nharrand:
It seems that currently annotation on type usages lead to spoon cloning annotations to put them on both the typed element and the type reference when the annotation's target is ambiguous. (Which is probably legitimate.)

This leads indirectly to position problem and sniper printer problems as documented in  #3321 . Even if this can be worked around ( #3352 ), or may be fixed by changing the start position of the type reference.

Consider the following case:

```java
//A type use annotation
@Target({ElementType.TYPE_USE})
@Retention(RetentionPolicy.RUNTIME)
public @interface MyAnno {
}

//A parameter annotation
@Target({ElementType.PARAMETER})
@Retention(RetentionPolicy.RUNTIME)
public @interface OtherAnno {
}

//An annotation the can target both parameter and type usage
@Target({ElementType.PARAMETER, ElementType.TYPE_USE})
@Retention(RetentionPolicy.RUNTIME)
public @interface Ambiguous {
}

public class AnnoUser {
	public void m1(@MyAnno String p) {} // Annotation is on the type reference to String
	public void m2(@OtherAnno String p) {} // Annotation is on parameter p
	public void m3(@Ambiguous String p) {} // Annotation is on both
}
```

Corresponding bytecode
```
  public void m(java.lang.String);
    descriptor: (Ljava/lang/String;)V
    flags: ACC_PUBLIC
    Code:
      stack=0, locals=2, args_size=2
         0: return
      LineNumberTable:
        line 4: 0
      LocalVariableTable:
        Start  Length  Slot  Name   Signature
            0       1     0  this   Lse/kth/access/testannoagain/AnnoUser;
            0       1     1     p   Ljava/lang/String;
    RuntimeVisibleTypeAnnotations:
      0: #16(): METHOD_FORMAL_PARAMETER, param_index=0

  public void m2(java.lang.String);
    descriptor: (Ljava/lang/String;)V
    flags: ACC_PUBLIC
    Code:
      stack=0, locals=2, args_size=2
         0: return
      LineNumberTable:
        line 5: 0
      LocalVariableTable:
        Start  Length  Slot  Name   Signature
            0       1     0  this   Lse/kth/access/testannoagain/AnnoUser;
            0       1     1     p   Ljava/lang/String;
    RuntimeVisibleParameterAnnotations:
      parameter 0:
        0: #19()

  public void m3(java.lang.String);
    descriptor: (Ljava/lang/String;)V
[...]

PR #3359, opened 2020-05-07
fix(PositionBuilder): Position of annotations on type reference are included in parent position

### LINKED ISSUE #3321: wrong prettyprint of javax annotation in sniper
When using the SniperJavaPrettyPrinter, It throws the error below when processing javax annotations

`
otherFragment ((/path/to/file/Class.java:21)) OTHER_IS_BEFORE of (/path/to/file/Class.java:21)
spoon.SpoonException: otherFragment ((/path/to/file/Class.java:21)) OTHER_IS_BEFORE of (/path/to/file/Class.java:21)
`

The specific line it complains about has this code
```java
public TestSource(@NotBlank String firstName, String lastName) {
        this.firstName = firstName;
        this.lastName = lastName;
    }
```
I am guessing its the javax annotation thats causing it. So its probably related to #3320 

This same code works fine using the DefaultJavaPrettyPrinter class

nharrand:
fix #3321

Reproduces #3358


nharrand:
This a cleaner workaround for #3359 . It fix the position of annotated CtTypeReferences.
Yet it still does not handle the case `m6` where a mix of annotation affect alternatively the type reference and the variable in an unfortunate order.

But the approach here cannot handle this case. To do it perfectly we need a post processing pass on all annotations and their parent to update parents positions.

monperrus:
nice, thanks @nharrand ! FYI CI still red.

monperrus:
Thanks a lot @nharrand

PR #3362, opened 2020-05-07
fix: Fix dropping diamonds on constructor calls

### LINKED ISSUE #3360: bug: <> dropped from constructor calls to generic types (noclasspath)
Hi!

I've found another noclasspath bug. It's not too critical, but it affects my use case a bit horribly. In any case, whenever there is a constructor call to a generic type `A`, and the actual type arguments are omitted, then Spoon doesn't actually pick up the type arguments **iff `A` is not on the classpath**. Here's an example:

```java
java.util.ArrayList<Integer> list = new ArrayList<>();
A<Integer> a = new A<>();
```
In the snippet above, the type access in `new ArrayList<>();` will have an implicit `Integer` type argument. However, the type access in `new A<>();` will not. Just parsing and pretty-printing the above will yield the following.

```java
java.util.ArrayList<Integer> list = new ArrayList<>();
A<Integer> a = new A();
```
Note how the type access in `new A()` suddenly is raw.

This is only a problem under these specific conditions:

1. The actual type arguments are omitted in the source code.
2. The generic type that's specialized is not available on the classpath OR the constructor call is contained in a context where the type arguments cannot be inferred, such as inside of a call to an unresolved method.

In summary, this only affects noclasspath mode when using non-stdlib generic types, or using non-stdlib methods that have specialized generic types as arguments. The problem is that the JDT node doesn't actually have the type information about the type arguments that should be inferred from the context.

I'll have a PR up with a test case later today, but I've only been able to partially solve the issue.

slarse:
Fix #3360 

This is an attempt to fix the problem discussed in #3360 

What I do here is essentially to try to recover the type arguments from the `AllocationExpression`'s expected type. To the best of my knowledge, empty diamonds are only allowed on constructor calls, so I only have an assert that the stack's top element is an `AllocationExpression`. It shouldn't be possible for it to be anything else.

If the constructor call occurs inside of an unresolved method, or another constructor call, then the `AllocationExpression` doesn't have an expected type (because it can't be resolved). In that case, I add the new `OMITTED_TYPE_ARG_TYPE` (name could use some work) to the type argument list so at least Spoon can detect that the constructor is in fact parameterized, and will then print `<>`.

This appears to work as intended, but I'll be running some more tests locally with my merge tool to see if other issues crop up.

### DIFF
--- a/src/main/java/spoon/reflect/factory/TypeFactory.java
+++ b/src/main/java/spoon/reflect/factory/TypeFactory.java
@@ -93,6 +93,7 @@ public class TypeFactory extends SubFactory {
 	public final CtTypeReference<Set> SET = createReference(Set.class);
 	public final CtTypeReference<Map> MAP = createReference(Map.class);
 	public final CtTypeReference<Enum> ENUM = createReference(Enum.class);
+	public final CtTypeReference<?> OMITTED_TYPE_ARG_TYPE = createReference(CtTypeReference.OMITTED_TYPE_ARG_NAME);
 
 	private final Map<Class<?>, CtType<?>> shadowCache = new ConcurrentHashMap<>();
 
--- a/src/main/java/spoon/reflect/reference/CtTypeReference.java
+++ b/src/main/java/spoon/reflect/reference/CtTypeReference.java
@@ -35,6 +35,11 @@ public interface CtTypeReference<T> extends CtReference, CtActualTypeContainer, 
 	String NULL_TYPE_NAME = "<nulltype>";
 
 	/**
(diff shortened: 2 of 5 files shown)


slarse:
I'm pretty happy with this now, it appears to work for my purposes. None of the cases where I lost `<>` seem to happen anymore.

I'm unsure about the use of the `OMITTED_TYPE_ARG_TYPE`, I'm not certain that's the way things should be done in Spoon, but purely functionally I think this works well.

monperrus:
LGTM, will merge.

monperrus:
Thanks @slarse well done

slarse:
Happy to contribute!

PR #3364, opened 2020-05-11
fix: Ensure unqualified member access remains unqualified

### LINKED ISSUE #3363: bug: Implicit type access/package reference made explicit (noclasspath)
Hello!

Another day, another noclasspath bug :). For once, this bug is pretty simple. Sometimes when an unqualified type access is in the code to a type that is not on the classpath, Spoon will make the type/package explicit. The case where this messed things up for me was with static imports.

```java
package pkg;

import static pkg.constants.Constants.SOMETHING;

public class Main {
        System.out.println(SOMETHING);
    }
}
```
Here, `SOMETHING` is clearly entirely unqualified, yet it will get parsed into `Constants.SOMETHING`, where `Constants` will be explicit. This causes compile errors for me as the class `Constants` is not itself imported.

Remove that static import, and it instead becomes a qualified `pkg.SOMETHING`, which is arguably incorrect as it doesn't really make sense for accessing a constant, but if `SOMETHING` was a type or something like that instead then it would be correct. Perhaps we want to fix that, too, but now I'm mainly out to fix the implicitness.

The problem lies here:

https://github.com/INRIA/spoon/blob/37b100acf42cff9dec91c7e5c6ecfee0454a36d1/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java#L635-L644

As this is a `SingleNameReference`, any qualification attached to it should be made implicit as far as I can gather. Yet, anything that can be scrounged up from the imports is attached without abandon:

https://github.com/INRIA/spoon/blob/37b100acf42cff9dec91c7e5c6ecfee0454a36d1/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java#L642

I have a PR in the works, the fix seems to be as straightforward as simply making any reference found from the imports implicit.

slarse:
Fix #3363 

This is a fix for the aforementioned issue. The approach I've taken here is outlined in #3363. If I'm not mistaken, this fix is simple and does not break anything.

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java
+++ b/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java
@@ -639,7 +639,12 @@ public class JDTTreeBuilderHelper {
 		} else {
 			typeReference.setSimpleName(CharOperation.charToString(singleNameReference.binding.readableName()));
 		}
-		jdtTreeBuilder.getReferencesBuilder().setPackageOrDeclaringType(typeReference, jdtTreeBuilder.getReferencesBuilder().getDeclaringReferenceFromImports(singleNameReference.token));
+		CtReference packageOrDeclaringType = jdtTreeBuilder.getReferencesBuilder().getDeclaringReferenceFromImports(singleNameReference.token);
+		if (packageOrDeclaringType != null) {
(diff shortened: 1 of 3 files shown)


slarse:
First CI will fail because of indentaion. I don't know why but IntelliJ just keeps randomly switching between tabs and spaces.

monperrus:
Thanks a lot @slarse

PR #3371, opened 2020-05-16
fix: Unqualified static method calls should be implicit

### LINKED ISSUE #3370: bug: Unqualified type access becomes fully qualified in some circumstances (noclasspath)
This bug specifically affects the method `JDTTreeBuilderHelper#createTypeAccessNoclasspath`

https://github.com/INRIA/spoon/blob/3a8d04df7cb0319896ae98b08986a7006b208fa6/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java#L635-L649

My previous fix (PR #3364 for issue #3363) was only partially successful. Apparently, it can happen that the `packageOrDeclaringType` variable is `null` (duh), and then it must be made implicit afterward This occurs when the JDT compiler can't resolve a package or declaring type. This occurs for example in the code below, where Spoon will guess that the type belongs to the CU's package.

```java
package pkg;

public class Cls {
        SomeClass.someMethod(args);
    }
}
```
The type access to `SomeClass` will become explicitly qualified as `pkg.SomeClass`.

The fix is simple, I'm working on a PR.

slarse:
Fix #3370 

This is a suggestion for solving #3370. It is a unified solution for both this and #3363 , as the fix I implement for the latter in #3364 was apparently only a partial solution to the problem.

I decided to add the `setImplicitPackageOrDeclaringType` to complement `setPackageOrDeclaringType` as the latter seems to be used in places where it's not obvious that the package/declaring type should be implicit. But in this particular case, we know we want it to be implicit as the `createTypeAccessNoClasspath(SingleNameReference)` is called with a `SingleNameReference` (i.e. not qualified).

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java
+++ b/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java
@@ -640,11 +640,8 @@ public class JDTTreeBuilderHelper {
 			typeReference.setSimpleName(CharOperation.charToString(singleNameReference.binding.readableName()));
 		}
 		CtReference packageOrDeclaringType = jdtTreeBuilder.getReferencesBuilder().getDeclaringReferenceFromImports(singleNameReference.token);
-		if (packageOrDeclaringType != null) {
-			// must be implicit as a SingleNameReference is not qualified, see #3363
-			packageOrDeclaringType.setImplicit(true);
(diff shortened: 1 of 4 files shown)


slarse:
Ugh, sorry, promise to start running checkstyle before I open a PR :)

monperrus:
Thanks a lot!

PR #3373, opened 2020-05-16
fix: Qualified name in catch clause resolved incorrectly (noclasspath)

### LINKED ISSUE #3372: bug: Qualified type references in catch clauses resolved incorrectly (noclasspath)
I have a problem in noclasspath mode where a qualified type reference in catch clauses are resolved incorrectly. Take the following code snippet.

```java
public class CatchQualifiedReference {
        try {
        } catch (some.neat.pkg.CustomException e) {
        }
    }
}
```
In noclasspath mode, `some.neat.pkg.CustomException` becomes a type references with the qualified name `some`. Which is pretty wild. The problem appears to be that `TypeBinding.readableName()` actually returns just `some` here:

https://github.com/INRIA/spoon/blob/3a8d04df7cb0319896ae98b08986a7006b208fa6/src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java#L961

So this could be a problem with the JDT library rather than with Spoon. I'm not certain. I'll put up a PR with a test case, but I haven't had the time to look into a solution.

slarse:
This is a test case to reproduce #3372 

I haven't yet had the time to look into solutions.

update fixes #3372

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java
+++ b/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java
@@ -127,7 +127,12 @@ public class JDTTreeBuilderHelper {
 			//do not set type of variable yet. It will be initialized later by visit of multiple types. Each call then ADDs one type
 			return result;
 		} else {
-			CtTypeReference ctTypeReference = jdtTreeBuilder.getReferencesBuilder().<Throwable>getTypeReference(typeReference.resolvedType);
+			CtTypeReference ctTypeReference;
+			if (typeReference.resolvedType instanceof ProblemReferenceBinding) {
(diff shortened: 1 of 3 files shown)


slarse:
It appears as if using the closest match (if available from `TypeBinding.closestMatch()`) at least lets you resolve the correct readable name. I'll try some stuff with that.

MartinWitt:
Hi,
if you mark your test with @Ignore("UnresolvedBug") the CI will check that the test fails and be green afterwards.

slarse:
Hi!

Based on the contributing guidelines, I thought that was supposed to be used if you want to submit just a failing test case? I intend to try to solve this problem, I just haven't had the time to look into it yet.

monperrus:
just pushed a fix, let's wait for CI

slarse:
@monperrus Looks like it solved the problem? CI failure seems unrelated.

monperrus:
unwip-ed, @nharrand ping

Issue #3381, opened 2020-05-25
[BUG] toString crashes due to comment handling


santos-samuel:
Hello,
I just came across an error.

When I add certain classes as input resources to the launcher, Spoon, when building the model, will only partially parse the class and throw an error that is shown on some fields saying:
`Method threw 'spoon.SpoonException' exception. Cannot evaluate spoon.support.reflect.declaration.CtClassImpl.toString()`

And when I call the toString method it says:
`spoon.SpoonException: Unexpected next line after last line`

~~It's not that I care about the toString() method, but this error is not allowing me, for instance, to get the Executable Declaration of a method based on its executable.~~ (actually this error was coming from another place)

I created a PR with two failing test cases so you can inspect it:
#3382 

Thanks in advance!

EDIT1:
After some debugging I found that the error is on class `JDTCommentBuilder.java` on line 608.
I believe the problem is that you are not considering line comments inside block comments, for example:
```
            /*
            // Set the SecurityContextHolder authentication to null
            SecurityContextHolder.getContext().setAuthentication(null);

            // Call the commence method of authenticationEntryPoint
            authenticationEntryPoint.commence(request,response,new BadCredentialsException("IP not allowed"));
            */
```

When the code sees the first '/' on the first line of the comment it will immediatly do `isLastLine = true;`. And then the second line `SecurityContextHolder.getContex (...)` will throw the error.

santos-samuel:
If the comments are not needed I believe this fixes eventual problems:
`launcher.getEnvironment().setCommentEnabled(false);`

monperrus:
Thanks for the bug report.

What happens if you call toStringDebug instead of toString?

santos-samuel:
> Thanks for the bug report. What happens if you call toStringDebug instead of toString?

`toStringDebug()` seems to be working fine. It does not throw the Exception

santos-samuel:
Since the PR was merged already i'm closing this issue

PR #3382, opened 2020-05-25
review fix: fix crashing condition in cleanComment

santos-samuel:
PR with failing test cases.
Related with issue #3381


monperrus:
Thanks a lot for the failing test case.

Bug reproduced and fixed. Cleaned the test case. Pushed on your branch directly.

The cause was a mix of block and line comment:
```
/* block
// line comment in block comment
*/
```

Ready for merge.

monperrus:
@nharrand ready for merge!

nharrand:
Nice one! Thanks for both the bug report and the fix!

Issue #3365, opened 2020-05-14
HashcodeVisitor and EqualsChecker do not take the declaring type into account


dnl50:
The current [`HashcodeVisitor`](https://github.com/INRIA/spoon/blob/3a8d04df7cb0319896ae98b08986a7006b208fa6/src/main/java/spoon/support/visitor/HashcodeVisitor.java) implementation does not take the [Declaring Type](https://github.com/INRIA/spoon/blob/3a8d04df7cb0319896ae98b08986a7006b208fa6/src/main/java/spoon/reflect/declaration/CtTypeMember.java#L24) of a [`CtTypeMember`](https://github.com/INRIA/spoon/blob/3a8d04df7cb0319896ae98b08986a7006b208fa6/src/main/java/spoon/reflect/declaration/CtTypeMember.java) into account. This leads to unexpected behaviour when using `HashSet`s containing `CtElement`s. Overriding the `scanCtTypeMember` in `HashcodeVisitor` and `EqualsChecker` would solve that problem.

## Example

Two `CtFieldImpl` instances representing two Java fields with **the same name** but **declared in different classes** are equal and generate the same hash code:

```java
public class Person {

    private String name;

}

public class City {

    private String name;

}

CtClass<?> personCtClass = ...;
CtClass<?> cityClass = ...;

CtField<?> personNameField = personCtClass .getField("name");
CtField<?> cityNameField = cityClass.getField("name");

personNameField.equals(cityNameField) // true
personNameField.hashCode() == cityNameField .hashCode() // true

```

MartinWitt:
After having a first look #3273 seems related. But you are correct the behavior is unexpected. Would you mind writting a testcase for the bug or even better try to fix it?

dnl50:
I will try my best to implement a fix and submit a PR in the coming days

dnl50:
After adapting the bahaviour of `HashcodeVisitor` and `EqualsChecker` to include the declaring type's qualified name some tests are failing which seem to be related to cloned elements. The parent element isn't set on a cloned element, so cloned `CtClass`es end up in the default package for example (see `CloneTest#testCloneListener`). Adding the behaviour of cloning the parent object understandably leads to a `StackOverflowException` being thrown. Looking at the abstract `CtScanner` from which the `CloneVisitor` is magically generated it seems like only the element itself and all its child elements should be cloned when cloning an element. So would it be sufficient to just set the parent reference after cloning with something like this?

```java
[...]

PR #3385, opened 2020-05-31
WIP: Take qualified name of declaring type into account in equals and hashCode

dnl50:
Fix for issue mentioned in #3365

### DIFF
--- a/src/main/java/spoon/support/visitor/HashcodeVisitor.java
+++ b/src/main/java/spoon/support/visitor/HashcodeVisitor.java
@@ -7,9 +7,7 @@
  */
 package spoon.support.visitor;
 
-import spoon.reflect.declaration.CtElement;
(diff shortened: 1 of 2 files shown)


monperrus:
Thanks a lot for the patch proposal @dnl50 . In Spoon, we always add a test case with a patch, showing the fixed behavior. Could you add such a test case? Thanks!

dnl50:
Sure, but there is also a underlying problem with cloning not setting the declaring type of the cloned element, thats why some test cases fail. #3365 contains a more detailed explanation of the problem. I will try to look into it in the coming days.

monperrus:
OK, a failing test case is always super valuable to fully understand the problem.

monperrus:
Revive or close?

monperrus:

Thanks!

Issue #3386, opened 2020-06-01
bug(sniper): fields and local variables with joint declaration cannot be snipered


fermadeiral:
Hi all,

I have the following code:

```java
        int a, c = 0;

	void foo(int a) {
		a = a;
	}
```

I want to remove the statement `a = a;`, which is fine. However, after the sniper printer prints the code, I get: `java.lang.IllegalStateException: Registering symbol: 'c' twice in the same scope`. This happens because the printed code is:

```java
        int a, c = 0;

	int c = 0;

	void foo(int a);
```

See, everything is fine with my transformation, but the variable `c` is printed twice. I created a failing test case to expose this issue, I hope it helps for fixing the bug:

```java
@Test
	public void testPrintOneLineMultipleVariableDeclaration() {
		testSniper(OneLineMultipleVariableDeclaration.class.getName(), type -> {
			CtMethod<?> m = type.getMethodsByName("foo").get(0);
			m.setBody(null);
		}, (type, printed) -> {
			assertIsPrintedWithExpectedChanges(type, printed, "\\Qvoid foo(int a) {\n\t\ta = a;\n\t}", "void foo(int a);");
		});
	}
```
This is the content of the `OneLineMultipleVariableDeclaration` class:

```java
package spoon.test.prettyprinter.testclasses;

public class OneLineMultipleVariableDeclaration {

	int a, c = 0;

	void foo(int a) {
		a = a;
	}
}
```

monperrus:
For the record, making progress with #3390, #3391, #3392

fermadeiral:
Nice!

fermadeiral:
Thanks @monperrus and @nharrand!

fermadeiral:
Hi,

The workaround doesn't work for the context I presented here. In my case, I have `int a, c = 0;`, but in your test you changed it to `int a, c;`.

The error changed for me now:

```
spoon.SpoonException: spoon.SpoonException: Unexpected source text: 0;

	at spoon.Launcher.prettyprint(Launcher.java:774)
	at spoon.test.prettyprinter.TestSniperPrinter.testSniper(TestSniperPrinter.java:281)
	at spoon.test.prettyprinter.TestSniperPrinter.testPrintOneLineMultipleVariableDeclaration(TestSniperPrinter.java:113)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[...]

PR #3387, opened 2020-06-01
REVIEW fix(SniperPrettyPrinter): fix pretty-printing of joint declarations (#3386)

nharrand:
Reproduces #3386

### DIFF
--- a/.github/workflows/lsif.yml
+++ b/.github/workflows/lsif.yml
@@ -19,3 +19,4 @@ jobs:
           endpoint: https://sourcegraph.com
           github_token: ${{ secrets.GITHUB_TOKEN }}
           file: dump.lsif
+          ignore_upload_failure: true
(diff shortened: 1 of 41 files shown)


monperrus:
Ohh that will be tough

the ElementSourceFragment is `int a;|99, 108|int a, c;|`, it covers two elements while being attached to only `int a` (`elementFragment.getElement() == <int a>` only)

monperrus:
The ElementSourceFragment of `int c` is merged into that of `int a` because it is a strict subset.

monperrus:
Here is a first fix so that @fermadeiral can proceed.

It's a partial fix, fields with joint declarations won't be snipered, but the improvement is that the resulting code will compile. 

It's mergeble but it does not close #3386.

fermadeiral:
Thanks @monperrus!

PR #3391, opened 2020-06-03
Feat: add utility class SameFilter

monperrus:


src/main/java/spoon/reflect/visitor/filter/SameFilter.java
```diff
@@ -0,0 +1,25 @@
...
+ * Copyright (C) 2006-2019 INRIA and contributors
+ *
+ * Spoon is available either under the terms of the MIT License (see LICENSE-MIT.txt) of the Cecill-C License (see LICENSE-CECILL-C.txt). You as the user are entitled to choose the terms under which to adopt Spoon.
+ */
+package spoon.reflect.visitor.filter;
+
+import spoon.reflect.declaration.CtElement;
+import spoon.reflect.visitor.Filter;
+
+/** Finds the element given in parameter, useful for checking if an element is in an ancestor */
```
MartinWitt, line 13: maybe add in the doc, that you check with ==. The term "same" is sadly not well known and sometimes used instead of equals.

monperrus: Fixed, thanks!

PR #3394, opened 2020-06-05
chore: allow LSIF upload action to pass even if upload fails

rvantonder:
Hi @monperrus I noticed that GH was showing a red ❌  for the build due to a setting for that uploading LSIF data in the GH action (sorry about that!). 

It's unfortunate that we can't tell GH to ignore the status of failed actions (and so give a nice ✅  instead of X even if this action fails). There is already an upstream fix for the action, but this PR will also stop any other spurious causes that may show a red ❌  for the build if it is LSIF upload related.

### DIFF
--- a/.github/workflows/lsif.yml
+++ b/.github/workflows/lsif.yml
@@ -19,3 +19,4 @@ jobs:
           endpoint: https://sourcegraph.com
           github_token: ${{ secrets.GITHUB_TOKEN }}
           file: dump.lsif
+          ignore_upload_failure: true


monperrus:
Thanks a lot Rijnard.

monperrus:
HI @rvantonder 

LSIF is breaking the CI status on master: https://github.com/INRIA/spoon/commits/

Any idea to fix it?

rvantonder:
@monperrus this should fix it: https://github.com/INRIA/spoon/pull/3614. Alternatively, we can disable/remove the action until we're able to release more stable versions--definitely don't want this to cause any trouble for your CI.

rvantonder:
It's possible to rerun the current stuck jobs by going to the action tab and hitting rerun for the job (I don't have permissions on this repo). That should work--if it doesn't, let's remove https://github.com/INRIA/spoon/blob/master/.github/workflows/lsif.yml which is enough to stop the action from running.

monperrus:
re-ran all jobs, everything is green, thanks :)

PR #3407, opened 2020-06-10
test: change test resource for Dependabot

monperrus:
so that dependabot is not confused anymore (#3406)


monperrus:
of course this breaks a test, closing and following up on #3409

Issue #3097, opened 2019-09-11
sniper: unable to rename class


tdurieux:
Using this code:

```java
final CtClass<?> ctClass = f.Class().get("spoon.test.prettyprinter.testclasses.Validation");
ctClass.setSimpleName("Test");
```

The model is correctly changed, but the sniper mode does not succeed to generate the correct output.

monperrus:
Do you mean 

* only rename the class 
* or rename the class and all references (full-fledged refactoring)

tdurieux:
A full refactoring could be nice but here the problem is the the sniper mode does not change the name of the class. In the output the old name is still uses.

ITWOI:
Hi, I have a similar problem. But it leads to NPE.

I use Spoon 8.1.0-beta-9. 

Here is the code, note that if I remove `launcher.getEnvironment().setPrettyPrinterCreator(() -> new SniperJavaPrettyPrinter(launcher.getEnvironment()));`, then every thing works fine.

```
public class ClassRename extends AbstractProcessor<CtClass> {
    int renameIndex;
    public ClassRename(int renameIndex) {
        this.renameIndex = renameIndex;
    }
    @Override
    public void process(CtClass arg0) {
        String newMethodName = arg0.getSimpleName() + renameIndex;
        arg0.setSimpleName(newMethodName);
    }
}
```

```
   void entry(List<StmtTrace> ele, int index) {
        final Launcher launcher = retLauncher();
        launcher.addProcessor(new ClassRename(index));
        launcher.run();
    }
    Launcher retLauncher(){
        final Launcher launcher = new Launcher();
        launcher.setArgs(arg);
        launcher.getEnvironment().setPrettyPrinterCreator(() -> new SniperJavaPrettyPrinter(launcher.getEnvironment()));
        return launcher;
    }
```

```
Exception in thread "main" spoon.SpoonException: java.lang.NullPointerException
	at spoon.Launcher.prettyprint(Launcher.java:789)
	at spoon.Launcher.run(Launcher.java:725)
	at MutManager.callMut(MutManager.java:307)
	at MutManager.doMutate(MutManager.java:297)
	at MutManager.executeMutation(MutManager.java:184)
	at MutManager.runManager(MutManager.java:227)
	at Main.main(Main.java:71)
Caused by: java.lang.NullPointerException
	at spoon.support.reflect.declaration.CtCompilationUnitImpl.getMainType(CtCompilationUnitImpl.java:148)
[...]

PR #3411, opened 2020-06-12
feat: add method to add new CompilationUnits to be pretty-printed

monperrus:
this new method comes from a refactoring, so it is already called and tested.

preliminary work for #3097


monperrus:
@nharrand ready

PR #3425, opened 2020-06-17
refactor: remove unnecessary DirectPrinterHelper

monperrus:
MutableTokenWriter is enough, we don't need more customization


monperrus:
Status is broken but CI is green https://travis-ci.org/github/INRIA/spoon/builds/699339752

ping @nharrand

Issue #3420, opened 2020-06-16
JDTImportBuilder not using input class loader


loetifuss:
Hi folks,

I'm trying to collect all imports of a set of source files with Spoon by using the following in a processor (simplified)
```java	
//CtType<?> element;
CompilationUnit compilationUnit = element.getFactory().CompilationUnit().getOrCreate(element);
compilationUnit
	.getImports()
	.stream()
	.forEach(i -> { System.out.println("Found import " + i.getShortRepresentation()); });
```
The binary classpath of the sources is set to the launcher through a custom class loader
```java
// URLClassloader containing all dependencies as JARs
launcher.getEnvironment().setInputClassLoader(urlClassLoader);
```

I noticed that this only works if all imports are visible to the class loader that is used to load "JDTImportBuilder" because of line 151
```java
// this should really use the input class loader in 'Environment.getInputClassLoader()', if set
Class zeClass = this.getClass().getClassLoader().loadClass(className);
```
see https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/compiler/jdt/JDTImportBuilder.java#L151

So, "JDTImportBuilder" should honour a custom input class loader if specified. There are many cases where classloaders may differ, e.g. a Gradle task performing source analysis using Gradle compile classpath.

monperrus:
Thanks a lot for the bug report. Would you propose a failing test case in a pull-request? That would be super helpful!

loetifuss:
@monperrus see PR with proposed fix and test case to reproduce the issue https://github.com/INRIA/spoon/pull/3454

monperrus:
closed per #3454

PR #3454, opened 2020-06-20
fix: use input classloader for import resolution

loetifuss:
This PR addresses the problem observed in https://github.com/INRIA/spoon/issues/3420
If an input classloader is specified in Environment it should be used to resolve imports. 
Otherwise only imported classes on the same class loader as JDTImportBuilder are found.

To reproduce the problem see tests in spoon.support.compiler.jdt.JDTImportBuilderTest


monperrus:
Thanks a lot @loetifuss We love PRs with test + fix :-)

PR #3494, opened 2020-07-15
feat: add CtStatement#comment to comment code

### LINKED ISSUE #2013: add method `CtStatement#comment()`
to comment code in the pretty-printed code, this is a traceable version of (`#delete`)

LakshyAAAgrawal:
Add "comment" method to comment out a statement. 

In case of a CtBlock, if the parent is a method(the block defines a method body) then the statements within the block are commented out, otherwise, the statement itself is commented out. In case of multiline statements, a block comment is created, otherwise an inline comment is created.

Currently working on testcases for all the child classes of CtStatement.

Fixes #2013

### DIFF
--- a/src/main/java/spoon/reflect/code/CtBlock.java
+++ b/src/main/java/spoon/reflect/code/CtBlock.java
@@ -27,4 +27,12 @@ public interface CtBlock<R> extends CtStatement, CtStatementList, TemplateParame
 
 	@Override
 	CtBlock<R> clone();
+
+	/**
+	 * If the block is a method body, then all the statements are
+	 * replaced by corresponding comments within the block. If not
+	 * then the block itself is replaced by a Block comment.
+	 */
+	@Override
+	public void comment();
 }
--- a/src/main/java/spoon/reflect/code/CtBlock.java
+++ b/src/main/java/spoon/reflect/code/CtBlock.java
@@ -34,5 +34,5 @@ public interface CtBlock<R> extends CtStatement, CtStatementList, TemplateParame
 	 * then the block itself is replaced by a Block comment.
 	 */
 	@Override
-	public void comment();
+	void comment();
 }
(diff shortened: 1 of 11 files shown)


monperrus:
Thanks for the proposal.

There is a lot of code duplication here.

Here is a tentative solution, by putting adding a single method in CtElementImpl#comment

```java
if (this instanceof CtStatement && getParent() instanceof CtBlock) {
 // do it
} else {
  throw new UnsupportedOperationException()
}
```

This would make this feature more understandable and maintainable.

WDYT?

monperrus:
Thanks Benjamin.

MartinWitt:
Couldn't the instanceOf be replaced with a visitor?

monperrus:
that's also an option. but we like to KISS: the simpler the better.

LakshyAAAgrawal:
Thank you very much for the kind suggestions. 

I will make the following changes:
1. Remove CtStatementImpl#comment
2. Add CtElementImpl#comment as described above
3. Keep CtBlock#comment - The implementation for this instance is different from above, since in case of the block being a method body, the inner statements are commented out instead of commenting out the block itself
4. Keep CtStatament#comment - I believe that #comment must be a part of the interface CtStatement but not CtElement.

Kindly let me know about your views on these and I will go ahead with the implementation.

monperrus:
yes, and remove changes in CtClassImpl, CtUnaryOperatorImpl, CtInvocationImpl, CtConstructorCallImpl, etc.

thanks!

monperrus:
Thanks a lot!

In Spoon, we have one comment line per test that states the test intention, the contract that is being tested. By convention, it starts with `// contract:` eg

    // contract: a single statement can be commented out

Could you add this contract line in each test?

Thanks!

LakshyAAAgrawal:
I have added the contract statements.

monperrus:

monperrus:
Thanks a lot @LakshyAAAgrawal for your contribution!

Issue #3514, opened 2020-07-31
Analyzing project creates a NullPointerException


layornos:
Hi @ll,

when I try analyzing the project [1] I am getting the following exception
```
[main] INFO spoon.Launcher - Running in NOCLASSPATH mode (doc: http://spoon.gforge.inria.fr/launcher.html).
[main] DEBUG spoon.support.compiler.jdt.JDTTreeBuilder - Could not find declaration for variable StereotypeAPI at (./Palladio-Analyzer-SimuLizar/bundles/org.palladiosimulator.simulizar.action.repository/src/org/palladiosimulator/simulizar/action/repository/black/ProfilesLibrary.java:76).
[main] DEBUG spoon.support.compiler.jdt.JDTTreeBuilder - Could not find declaration for variable LOGGER at (./Palladio-Analyzer-SimuLizar/bundles/org.palladiosimulator.simulizar.action/src-man/org/palladiosimulator/simulizar/action/interpreter/TransientEffectInterpreter.java:371).
[main] DEBUG spoon.support.compiler.jdt.JDTTreeBuilder - Could not find declaration for variable state at (./Palladio-Analyzer-SimuLizar/bundles/org.palladiosimulator.simulizar.action/src-man/org/palladiosimulator/simulizar/action/interpreter/TransientEffectInterpreter.java:373).
[main] DEBUG spoon.support.compiler.jdt.JDTTreeBuilder - Could not find declaration for variable LOGGER at (./Palladio-Analyzer-SimuLizar/bundles/org.palladiosimulator.simulizar.action/src-man/org/palladiosimulator/simulizar/action/interpreter/TransientEffectInterpreter.java:396).
[main] DEBUG spoon.support.compiler.jdt.JDTTreeBuilder - Could not find declaration for variable MetricDescriptionUtility at (./Palladio-Analyzer-SimuLizar/bundles/org.palladiosimulator.simulizar.slidingwindow/src/org/palladiosimulator/simulizar/slidingwindow/probeframework/SlidingWindowProbeFrameWorkListenerDecorator.java:100).
[main] DEBUG spoon.support.compiler.jdt.JDTTreeBuilder - Could not find declaration for variable MetricDescriptionUtility at (./Palladio-Analyzer-SimuLizar/bundles/org.palladiosimulator.simulizar.slidingwindow/src/org/palladiosimulator/simulizar/slidingwindow/runtimemeasurement/SlidingWindowRuntimeMeasurementsRecorder.java:88).
[...]

PR #3519, opened 2020-08-03
fix: fix bug in CtLineElementComparator

MartinWitt:
## Core Problem
As we saw in issue #3514 analyzing Palladio-SimuLizar produces jdk erros while sorting.
The reason for this is because the CtLineComparator does not follow the compareTo contract. This patch fixes the comparator. 

## Some Math Stuff
A comparator must follow the compareTo contract in java. The compareTo contract defines two formulas:

(1)  *antisymmetric* sgn(x.compareTo(y)) == -sgn(y.compareTo(x))
(2)  *transitive* (x.compareTo(y)>0 && y.compareTo(z)>0) => x.compareTo(z)>0

Assuming we have 3 points X,Y and Z and all points have invalid positions.

`compareTo(X,Y)` == -1 ==> X < Y  
`compareTo(Y,Z)` == -1 ==> Y < Z  
`compareTo(X,Z)` == -1 ==> X < Z  
Looks good till here but now:  
`compareTo(Z,X)` == -1 => Z<X
here we violate the *transitive* requirement. Because Z is now less than X.
If we do the same after the change:
`compareTo(X,Y)` == 0 ==> X=Y  
`compareTo(Y,Z)` == 0 ==> Y=Z  
`compareTo(X,Z)` == 0 ==> X=Z  
`compareTo(Z,X)` == 0 ==> Z=X  
The change still follows the antisymmetric contract, because sign(0)=0== -sign(0).

## Consequences
Some testcases relay on the old sorting, because order was different. Some refactoring for the testcases is needed.  
There is a risk, that some users have problems with new correct ordering.


MartinWitt:
A comparator must follow the compareTo contract in java. The compareTo contract defines two formulas:

(1)  *antisymmetric* sgn(x.compareTo(y)) == -sgn(y.compareTo(x))
(2)  *transitive* (x.compareTo(y)>0 && y.compareTo(z)>0) => x.compareTo(z)>0

Assuming we have 3 points X,Y and Z and all points have invalid positions.

`compareTo(X,Y)` == -1 ==> X < Y  
`compareTo(Y,Z)` == -1 ==> Y < Z  
`compareTo(X,Z)` == -1 ==> X < Z  
Looks good till here but now:  
`compareTo(Z,X)` == -1 => Z<X
here we violate the *transitive* requirement. Because Z is now less than X.
If we do the same after the change:
`compareTo(X,Y)` == 0 ==> X=Y  
`compareTo(Y,Z)` == 0 ==> Y=Z  
`compareTo(X,Z)` == 0 ==> X=Z  
`compareTo(Z,X)` == 0 ==> Z=X  
The change still follows the antisymmetric contract, because sign(0)=0== -sign(0).

monperrus:
Thanks for the fix!

2 failing tests are on generated code, we just have to regenerate the code.

So only `CtClassTest.getConstructor` is really failing, which suggests that we won't break many things.

Overall, the fix is reasonable to me.

MartinWitt:
@monperrus merge/review ping

src/test/java/spoon/test/ctClass/CtClassTest.java
```diff
@@ -95,18 +95,10 @@ public void getConstructor() throws Exception {
 		cons.addParameter(cons.getFactory().createParameter().setType(cons.getFactory().Type().OBJECT));
 		// now that we have changed the signature we can call getConstructors safely
 		assertEquals(4, foo.getConstructors().size());
-		// we cloned the first constructor, so it has the same position, and comes before the 2nd and 3rd constructor
-		assertSame(cons, foo.getTypeMembers().get(1));
```
monperrus, line 99: it's better to change the expected value than to remove assertions.

MartinWitt: now the test should still showcase adding a parameter does not change the constructor order and is still green.

monperrus:
Thanks a lot Martin!

Issue #3520, opened 2020-08-05
Bug when trying to get the reference of a lambda cast with an intersection


quentinLeDilavrec:
Hi,

There seems to be an error with a particular use of lambda expressions :
`Runnable f = (Serializable & Runnable) () -> system.out.println()`

If I try to get the reference of the lambda expression I get the Spoon exception :

>  The lambda can be based on interface, which has one method. But java.io.Serialize has no one

That is thrown in `CtLambdaImpl.getOverriddenMethod()`.

In fact, `Serialize` do not declare any method but `Runnable` does.

Thanks

monperrus:
Thanks a lot for the bug report. Would you be able to make a pull-request with a failing test case? That would be super useful.

PR #3521, opened 2020-08-08
fix: fix handling of intersection types for lambda parameters

quentinLeDilavrec:
#3520

### DIFF
--- a/src/main/java/spoon/support/reflect/code/CtLambdaImpl.java
+++ b/src/main/java/spoon/support/reflect/code/CtLambdaImpl.java
@@ -7,11 +7,14 @@ package spoon.support.reflect.code;
 
 import spoon.SpoonException;
 import spoon.reflect.annotations.MetamodelPropertyField;
+import spoon.reflect.code.CtAssignment;
 import spoon.reflect.code.CtBlock;
 import spoon.reflect.code.CtBodyHolder;
 import spoon.reflect.code.CtExpression;
 import spoon.reflect.code.CtLambda;
+import spoon.reflect.code.CtLocalVariable;
 import spoon.reflect.code.CtStatement;
+import spoon.reflect.declaration.CtElement;
 import spoon.reflect.declaration.CtMethod;
 import spoon.reflect.declaration.CtNamedElement;
 import spoon.reflect.declaration.CtParameter;
@@ -19,6 +22,7 @@ import spoon.reflect.declaration.CtType;
 import spoon.reflect.declaration.CtExecutable;
 import spoon.reflect.declaration.ModifierKind;
 import spoon.reflect.reference.CtExecutableReference;
+import spoon.reflect.reference.CtIntersectionTypeReference;
 import spoon.reflect.reference.CtTypeReference;
 import spoon.reflect.visitor.CtVisitor;
 import spoon.support.UnsettableProperty;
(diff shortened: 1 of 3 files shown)


src/main/java/spoon/support/reflect/code/CtLambdaImpl.java
```diff
@@ -101,7 +105,32 @@ public String getSimpleName() {
 			//it can be null in noclasspath mode, so we do not know which method is called, by lambda
 			return null;
 		}
-		CtType<T> lambdaType = lambdaTypeRef.getTypeDeclaration();
+		CtMethod<?> lambdaExecutableMethod = null;
+		if (!(lambdaTypeRef instanceof CtIntersectionTypeReference)) {
+			return getOverriddenAux(lambdaTypeRef);
+		} else {
+			CtElement parent = lambdaTypeRef.getParent();
```
monperrus, line 112: refactor the else block in `getOverriddenMethodForIntersectionType`?

src/main/java/spoon/support/reflect/code/CtLambdaImpl.java
```diff
@@ -101,7 +105,32 @@ public String getSimpleName() {
...
+				parentTypeReference = ((CtAssignment) parent).getAssigned().getType();
+			}
+			for (CtTypeReference<?> ctTypeReference : ((CtIntersectionTypeReference<?>) lambdaTypeRef).getBounds()) {
+				CtMethod<?> tmp = getOverriddenAux(ctTypeReference);
+				if (tmp != null && (lambdaExecutableMethod == null || ctTypeReference.equals(parentTypeReference))) {
+					lambdaExecutableMethod = tmp;
+				}
+			}
+		}
+		if (lambdaExecutableMethod == null) {
```
monperrus, line 126: do we need to move this check here? could we keep it at the previous place?

src/test/java/spoon/test/lambda/LambdaTest.java
```diff
@@ -458,6 +458,16 @@ public void testLambdaWithGenericExtendingMultipleInterfaces() {
 		assertIsWellPrinted("( elt) -> elt.test()", lambda);
 	}
 
+	@Test
+	public void testCastLambdaWithIntersection() {
+		final CtLambda<?> lambda1 = intersection.getElements(new TypeFilter<CtLambda<?>>(CtLambda.class)).get(1);
```
monperrus, line 463: could you add a one line comment stating the test intention in natural language:

    // contract: intersection types on lambda parameters are supported

src/main/java/spoon/support/reflect/code/CtLambdaImpl.java
quentinLeDilavrec: It follows the standard compiler behavior.
That is, at least one interface should have one method,
thus, some of these interfaces might not even declare any methods.

But as a side note, I just match the type that will be taken by the lambda in its parent,
but I did not implement a real type intersection, thus I am less strict than the compiler.
A great improvement would be to implement this type intersection in spoon.
But it is too complicated for me :/ and I would not be useful for my use case.

monperrus: OK for me, thanks.

monperrus:
Thanks. We have a few checkstyle problems.

nharrand:
LGTM! @monperrus do you want to merge?

monperrus:
Yes, thanks a lot @quentinLeDilavrec and @nharrand for the polishing commit!

Issue #3530, opened 2020-08-13
bug: MavenLauncher doesn't work on macOS systems


ChrisSquare:
Running the MavenLauncher sample as shown in the [Usage page](http://spoon.gforge.inria.fr/launcher.html) fails on OSX systems:
```
Exception in thread "main" java.io.IOException: Cannot run program "mvn": error=2, No such file or directory
```
This is thrown by `Process p = Runtime.getRuntime().exec(cmd);` in `SpoonPom.guessMavenHome()`, the code currently works on Unix type of OS but not OSX.

monperrus:
Thanks for the bug report. The fix might be easy, would you like to give it a try in a PR? That would be awesome.

ChrisSquare:
@monperrus Absolutely

monperrus:
closed per #3533

PR #3533, opened 2020-08-16
fix: Use shell command to detect Maven home version on Mac systems without failure

ChrisSquare:
This is a fix for bug #3530.
The fallback command that retrieves Maven home using a `mvn -version` command only works on Unix-based systems.
For it to work on MacOS systems it requires a shell execution that will search for Maven executable in the PATH environment variable: `sh -c "mvn -version"`.


monperrus:
LGTM, thanks a lot.

nharrand:
Thanks for the PR!

PR #3528, opened 2020-08-13
clean: clean old, never used class

monperrus:

PR #3536, opened 2020-08-18
chore: remove very old, unused, untested classes

monperrus:
#3528 was half-baked, now we really clean.


monperrus:
@nharrand ready for merge

Issue #3547, opened 2020-08-25
mvn test fails with "Test (wrong name: maven-launcher/system-dependency/target/classes/Test)"


Strum355:
Maven version: 3.6.3
Java versions: jdk8, 11 and 14
Followed https://github.com/INRIA/spoon#compiling 

```
[INFO] 
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[ERROR] Test (wrong name: maven-launcher/system-dependency/target/classes/Test)
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 0, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  01:12 min
[INFO] Finished at: 2020-08-25T11:34:15Z
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:3.0.0-M5:test (default-test) on project spoon-core: There are test failures.
[ERROR] 
[ERROR] Please refer to /home/mvn/target/surefire-reports for the individual test results.
[ERROR] Please refer to dump files (if any exist) [date].dump, [date]-jvmRun[N].dump and [date].dumpstream.
[ERROR] There was an error in the forked process
[ERROR] Test (wrong name: maven-launcher/system-dependency/target/classes/Test)
[ERROR] org.apache.maven.surefire.booter.SurefireBooterForkException: There was an error in the forked process
[ERROR] Test (wrong name: maven-launcher/system-dependency/target/classes/Test)
[ERROR]         at org.apache.maven.plugin.surefire.booterclient.ForkStarter.fork(ForkStarter.java:733)
[ERROR]         at org.apache.maven.plugin.surefire.booterclient.ForkStarter.run(ForkStarter.java:305)
[ERROR]         at org.apache.maven.plugin.surefire.booterclient.ForkStarter.run(ForkStarter.java:265)
[ERROR]         at org.apache.maven.plugin.surefire.AbstractSurefireMojo.executeProvider(AbstractSurefireMojo.java:1314)
[ERROR]         at org.apache.maven.plugin.surefire.AbstractSurefireMojo.executeAfterPreconditionsChecked(AbstractSurefireMojo.java:1159)
[...]

PR #3549, opened 2020-08-25
fix: fetchin module from the JDT compiler

Strum355:
Hi, I'm a new code intel SWE at Sourcegraph, and I'll be owning the https://github.com/sourcegraph/lsif-java project for the foreseeable time :) 

We came across an issue with Java 9 modules (tracking [here](https://github.com/sourcegraph/lsif-java/issues/8#issuecomment-666879764)), and I discovered a relevant `TODO` https://github.com/INRIA/spoon/blob/d7da4c6684/src/main/java/spoon/support/compiler/jdt/JDTBatchCompiler.java#L94

The following screenshot gives some insight into the state leading up to the exception, where it seems to extract the wrong module name from the file folder

<details>
<summary>VSCode Java Debugger state</summary>

image
</details>

I havent tested this PR in multi-module scenarios, this was the simpler path instead of sorting `this.filenames` and `this.modNames` and determining the most recent module name (from the sorted list), given non module-info.java and package-info.java entries in this.filenames have `null` set for their index in `this.modNames`.
If this doesn't work as expected in multi-module scenarios (which I will follow up on after this PR is made), this PR can be amended or a new one made for it.

Thanks :)

Note: I'm relying on CI tests to run as [I'm unable to run them locally](https://github.com/INRIA/spoon/issues/3547) as of yet

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/JDTBatchCompiler.java
+++ b/src/main/java/spoon/support/compiler/jdt/JDTBatchCompiler.java
@@ -91,8 +91,7 @@ public class JDTBatchCompiler extends org.eclipse.jdt.internal.compiler.batch.Ma
 							} else {
 								lastSlash += 1;
 							}
-							//TODO the module name parsed by JDK compiler is in `this.modNames`
-							compilationUnit.module = CharOperation.subarray(modulePath, lastSlash, modulePath.length);
+							compilationUnit.module = this.module.name();
 							pathToModName.put(String.valueOf(modulePath), compilationUnit.module);
 						}
 					} else {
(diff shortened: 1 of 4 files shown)


monperrus:
Could you move the change to ` .gitignore ` in a separate chore PR? Thanks!

monperrus:
@Strum355 still draft here? Would you be able to add the corresponding test case?

monperrus:
Thanks a lot for the test case!

In Spoon, we try to always have a contract in natural language in each test, see the other `// contract:` lines in test files.

Could you add this? Then we merge!

Strum355:
Added some contracts, let me know if you'd like any amendments to them :)

monperrus:
Thanks a lot @Strum355

PR #3552, opened 2020-08-27
fix: fix PositionBuilder if modifier is followed by generic with out whitespace e.g public static<T>

Strum355:
The following line would result in `o2` being of a value that would cause `modifierName` to include the generic type param, causing the looking into `explicitModifiersByName` to return null. https://github.com/google/gson/blob/master/gson/src/test/java/com/google/gson/ParameterizedTypeFixtures.java#L129

If a space would be between `static` and `<T>`, the modifier position is extracted properly.

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/PositionBuilder.java
+++ b/src/main/java/spoon/support/compiler/jdt/PositionBuilder.java
@@ -7,6 +7,7 @@
  */
 package spoon.support.compiler.jdt;
 
+import org.apache.commons.lang3.ArrayUtils;
 import org.eclipse.jdt.internal.compiler.ast.ASTNode;
 import org.eclipse.jdt.internal.compiler.ast.AbstractMethodDeclaration;
 import org.eclipse.jdt.internal.compiler.ast.AbstractVariableDeclaration;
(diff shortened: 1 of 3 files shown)


monperrus:
Thanks a lot for the bug fix.

What is the cause of the assertion changes? What if you add new method `n` at the end of the test file instead?

Strum355:
Ah those were because of the added import. Its no longer references in the test class so I will remove it and revert the previous assertions

monperrus:
cool, thanks.

Strum355:
Done that :+1: only getEndLine() and getSourceEnd() assertions had to be changed :)

monperrus:
Thanks for the new commits.

> getEndLine() and getSourceEnd() assertions had to be changed :)

because of a change of behavior or a change of test data? in the latter case, we would try to avoid this.

Thanks!

Strum355:
The latter case because a new method was added that caused the issue that this PR addresses. As a new method was added to the test input class, the source length (in chars and lines) was increased. The change in assertion numbers is simply to reflect this addition

monperrus:
> As a new method was added to the test input class, the source length (in chars and lines) was increased.

FYI, we avoid spurious assertion changes, so in this case, one option is to put the code in a new test input class.

monperrus:
Thanks a lot for your contribution @Strum355

PR #3554, opened 2020-08-28
test: fix assertions for newer jdks and refactor old test code.

MartinWitt:
As i wrote in issue #3547 we have some testcase, which are jdk dependent. Goal of this patch is to fix them all and as always refactor the testcase to reach a higher quality standard. Looking into the older testcases, there is still alot of refactoring work needed unrelated to bugs. I try to create as minimal diffs as possible while still refactoring the code to a newer style.

- [x] spoon.test.model.TypeTest > testGetAllExecutables 
- [x] spoon.test.reference.ExecutableReferenceTest > testSpecifyGetAllExecutablesMethod 
- [x] spoon.test.model.TypeTest > testAllTypeMembersFunctionMode 
- [x] spoon.test.factory.TypeFactoryTest > reflectionAPI

### DIFF
--- a/src/test/java/spoon/test/factory/TypeFactoryTest.java
+++ b/src/test/java/spoon/test/factory/TypeFactoryTest.java
@@ -1,18 +1,16 @@
 /**
- * Copyright (C) 2006-2018 INRIA and contributors
- * Spoon - http://spoon.gforge.inria.fr/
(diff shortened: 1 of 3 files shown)


monperrus:
Thanks a lot @MartinWitt

PR #3556, opened 2020-08-30
chore: clean dead class CtListener

monperrus:
CtListener is not used, not tested, not documented, poorly named. Its superinterface is important though.

Hence, this cleaning PR.


monperrus:
@nharrand ready

PR #3560, opened 2020-09-01
chore: added vscode files to .gitignore

Strum355:
As requested in https://github.com/INRIA/spoon/pull/3549#issuecomment-682325394

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,9 @@ maven-eclipse.xml
 !.idea/codeStyleSettings.xml
 *.iml
 
+# VSCode
+.vscode
+
 # Spoon
 spooned-classes/
 spooned/


monperrus:
Thanks. It seems to me that `*.gradle` would remove `build.gradle`. WDYT?

Strum355:
I did this under the assumption that spoon would be staying with Maven and not using Gradle or both (having to maintain both Maven and Gradle files). If something else is under consideration, I'd be happy to adjust for it

monperrus:
FYI, we already have some gradle in subprojects, see https://github.com/INRIA/spoon/blob/master/spoon-dataflow/build.gradle

Strum355:
Is it desirable to not gitignore gradle files at the root of the repo which exists as a maven project? If so, I can amend the gitignore to reflect that.
As it currently exists, these gitignore changes only affect from the root of the repo, so the files in spoon-dataflow/ would be unaffected.

monperrus:
Thanks!

PR #3562, opened 2020-09-02
fix: fix Maven source and test source directory resolving

Strum355:
By climbing the pom.xml hierarchy to find the nearest ancestor that declares the source directory rather than immediately falling back to src/(main|test)/java if not declared in this immediate pom.

Also added interpolating (project|pom|'').basedir which is not uncommonly found in projects.

### DIFF
--- a/src/test/java/spoon/support/compiler/SpoonPomTest.java
+++ b/src/test/java/spoon/support/compiler/SpoonPomTest.java
@@ -6,12 +6,9 @@ import spoon.MavenLauncher;
 import spoon.support.StandardEnvironment;
 
 import java.io.IOException;
-<<<<<<< HEAD
 import java.util.List;
 import java.util.regex.Pattern;
-=======
 import java.nio.file.Paths;
->>>>>>> 7c8b18599... Added test for getting maven source directory from ancestors
 
 import static org.junit.Assert.*;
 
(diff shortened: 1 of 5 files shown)


monperrus:
Cool! Would you add the corresponding test case? Thanks!

Strum355:
Not sure why the test is failing, it works locally

monperrus:
Thanks a lot for the test case + the fix. I guess the CI environment is slightly different or a file is missing?

Strum355:
@monperrus fixed the test case, the folder was missing because there were no files :+1:

src/test/java/spoon/support/compiler/SpoonPomTest.java
```diff
@@ -51,4 +52,20 @@ public void checkProfilesModules(String path, String[] expected, Pattern profile
...
+		);
+	}
+
+	public void checkSourceDirectory(String path, String expected) throws IOException, XmlPullParserException {
+		SpoonPom pomModel = new SpoonPom(path, null, MavenLauncher.SOURCE_TYPE.APP_SOURCE, new StandardEnvironment());
+
+		SpoonPom childModel = pomModel.getModules().get(0);
+		//contract: source directory is derived from parent pom.xml if not declared in the current
+		// (childModel) SpoonPom
+		assertEquals(expected, childModel.getSourceDirectories().get(0).getAbsolutePath(), expected);
```
monperrus, line 69: expected should be the second argument, not the last
you can remove the first one, the message, not relevant here

monperrus (commented):
Thanks, we're almost there, see comment below

src/test/java/spoon/support/compiler/SpoonPomTest.java
Strum355: Fixed according to the hover definition :+1: 
image

monperrus:
Thanks a lot @Strum355

Issue #3564, opened 2020-09-03
Not allowed javaletter or symbol found


darius-sas:
Hi,
I am using spoon for my PhD research and I am encountering the following exception while analysing several projects.

```
Caused by: spoon.SpoonException: Not allowed javaletter or keyword in identifier found. See JLS for correct identifier. Identifier: ?
	at spoon.support.reflect.reference.CtReferenceImpl.checkIdentiferForJLSCorrectness(CtReferenceImpl.java:108)
	at spoon.support.reflect.reference.CtReferenceImpl.setSimpleName(CtReferenceImpl.java:52)
	at spoon.reflect.factory.TypeFactory.createIntersectionTypeReferenceWithBounds(TypeFactory.java:703)
	at spoon.support.compiler.jdt.ReferenceBuilder.getTypeReference(ReferenceBuilder.java:983)
	at spoon.support.compiler.jdt.ReferenceBuilder.getTypeReference(ReferenceBuilder.java:745)
	at spoon.support.compiler.jdt.ReferenceBuilder.getTypeReference(ReferenceBuilder.java:790)
	at spoon.support.compiler.jdt.ReferenceBuilder.getExecutableReference(ReferenceBuilder.java:402)
	at spoon.support.compiler.jdt.ReferenceBuilder.getExecutableReference(ReferenceBuilder.java:473)
	at spoon.support.compiler.jdt.JDTTreeBuilder.visit(JDTTreeBuilder.java:1398)
	at org.eclipse.jdt.internal.compiler.ast.MessageSend.traverse(MessageSend.java:1109)
	at org.eclipse.jdt.internal.compiler.ast.MessageSend.traverse(MessageSend.java:1119)
	at spoon.support.compiler.jdt.JDTTreeBuilder.visit(JDTTreeBuilder.java:1057)
	at org.eclipse.jdt.internal.compiler.ast.CastExpression.traverse(CastExpression.java:705)
	at org.eclipse.jdt.internal.compiler.ast.ReturnStatement.traverse(ReturnStatement.java:379)
	at org.eclipse.jdt.internal.compiler.ast.MethodDeclaration.traverse(MethodDeclaration.java:365)
	at org.eclipse.jdt.internal.compiler.ast.TypeDeclaration.traverse(TypeDeclaration.java:1479)
	at org.eclipse.jdt.internal.compiler.ast.CompilationUnitDeclaration.traverse(CompilationUnitDeclaration.java:826)
[...]

PR #3568, opened 2020-09-05
fix: intersection types with wildcard "?" are allowed as simplename

MartinWitt:
Sometimes intersectiontypes set simplename for a typereference to "?". We should allow this. Sadly we are cutting more and more corners in this check, because we have 1 check for all `setSimplename `methods. Now spoon would accept `int a? =3;`, which isn't problematic for analyzing code, but for transformations. More problematic is that we have at least 2 implementations for the identifier check, see [here](https://github.com/INRIA/spoon/blob/1de9318fcb97fa99ba18c3fd40025163a55df89b/src/main/java/spoon/reflect/visitor/JavaIdentifiers.java). But maybe we should continue step by step fixing the issue and after that find a solution for the problem described above.


MartinWitt:
Created a testcase for the problem described in #3564 having only 1 testclass.
Fix for the bug and a testcase without byte buddy code is coming.

monperrus:
Thanks a lot @MartinWitt!

monperrus:
> More problematic is that we have at least 2 implementations for the identifier 

A refactoring PR to unify this is more than welcome :-)

Issue #3570, opened 2020-09-06
Spoon throws StackOverflowError


darius-sas:
Hello, I wanted to report another common issue I am encountering during my research. For instance, while parsing [https://github.com/stanfordnlp/CoreNLP/commit/b95cd18ce7e865bcb7cc3f3ade2c2e08295c6d6f](https://github.com/stanfordnlp/CoreNLP/commit/b95cd18ce7e865bcb7cc3f3ade2c2e08295c6d6f) using Spoon as a library, I get this error:

```
16:56:41 [main] DEBUG Could not find declaration for variable Language at (/home/fenn/data/atd-estimation/repos/CoreNLP/src/edu/stanford/nlp/pipeline/CoreNLPProtos.java:161).
16:56:41 [main] DEBUG Could not find declaration for variable Sentiment at (/home/fenn/data/atd-estimation/repos/CoreNLP/src/edu/stanford/nlp/pipeline/CoreNLPProtos.java:288).
16:56:41 [main] DEBUG Could not find declaration for variable NaturalLogicRelation at (/home/fenn/data/atd-estimation/repos/CoreNLP/src/edu/stanford/nlp/pipeline/CoreNLPProtos.java:433).
16:56:48 [main] DEBUG Freed memory of version: 1-b95cd18ce7e865bcb7cc3f3ade2c2e08295c6d6f
Exception in thread "main" java.lang.StackOverflowError
	at spoon.reflect.visitor.CtScanner.visitCtBinaryOperator(CtScanner.java:287)
	at spoon.support.reflect.code.CtBinaryOperatorImpl.accept(CtBinaryOperatorImpl.java:34)
	at spoon.reflect.visitor.CtScanner.scan(CtScanner.java:177)
	at spoon.reflect.visitor.AstParentConsistencyChecker.scan(AstParentConsistencyChecker.java:33)
	at spoon.reflect.visitor.CtScanner.scan(CtScanner.java:169)
	at spoon.reflect.visitor.CtScanner.visitCtBinaryOperator(CtScanner.java:290)
	at spoon.support.reflect.code.CtBinaryOperatorImpl.accept(CtBinaryOperatorImpl.java:34)
	at spoon.reflect.visitor.CtScanner.scan(CtScanner.java:177)
	at spoon.reflect.visitor.AstParentConsistencyChecker.scan(AstParentConsistencyChecker.java:33)
	at spoon.reflect.visitor.CtScanner.scan(CtScanner.java:169)
[...]

PR #3573, opened 2020-09-07
WIP test case for StackOverflowError for absurdly-long strings Issue #3570

darius-sas:
Pull request for issue #3570.

It seems that the error is caused by the absurdly-long string created in `src/test/resources/issue3570/DefaultTeXHyphenData.java` (copied from CoreNLP project).
If the class is removed from the project the analysis runs as expected.
I created a test case to replicate the issue: `src/test/java/spoon/Issue3570Test.java`.

Note that the exception might be raised while the debugging statement I added is called (therefore the stack trace includes OutputStreamWriter and other similar classes). If you remove the statement (`src/main/java/spoon/reflect/visitor/EarlyTerminatingScanner.java:145`) though, that doesn't happen and the error is the same as described in issue 3570.


monperrus:
Thanks a lot for the failing test case, that's very useful. Would you give it a try to find a fix?

monperrus:
@darius-sas what happens if you increase the stack size of the JVM running the tests, eg `java -Xss512M` or `java -Xss4096M`?

darius-sas:
The test passes and the class is parsed properly.

Sorry for not replying to you, but I am busy with another project ATM. If you haven't found a fix yet, I'll try give it a look next week.

monperrus:
OK, so this is a performance problem not a functional bug: we take too many calls per depth in the binary addition operation tree.

With the default stack size, what happens if you disable the checks with `launcher.getEnvironment().disableConsistencyChecks()`

darius-sas:
Consistency checks were already disabled.
```Java
            Environment environment = launcher.getEnvironment();
            environment.setEncoding(StandardCharsets.UTF_8);
            environment.disableConsistencyChecks();
            environment.setComplianceLevel(9);
            environment.setIgnoreDuplicateDeclarations(true);
            environment.setCommentEnabled(false);
            environment.setNoClasspath(true);
            environment.setShouldCompile(false);
            environment.debugMessage("ALL");
```

monperrus:

Thanks!

PR #3580, opened 2020-09-09
test: add test for the template engine

### LINKED ISSUE #2987: Template class gets added to the import as well as its source code gets added to the output source folder
Dear All,

I have two projects and therefore two source code folders
* gen - where the spoon-related code resides
* myapp - where the sources of a main application lives and where the generated code is being added

When generating a new class with some class template that extends ExtensionTemplate, lets say BaseLogicTemplate, an UNUSED import statement `import spoon.template.ExtensionTemplate;` gets added to the generated class.

I have another template, CustomLogicTemplate that extends BaseLogicTemplate, and when generating a class with this CustomLogicTemplate, another an UNUSED import statement `import spoon.template.BaseLogicTemplate;` gets added to the generated class. 

In addition, source code of both templates `BaseLogicTemplate` and `CustomLogicTemplate` are copied from `gen` project source code to `myapp` source code.

Here is how a Launcher gets created:

        launcher.addInputResource(mainProjectFolder + "/src/main/java");
        launcher.addTemplateResource(new FileSystemFolder("./src/main/java/gen/core/templates"));
        launcher.setSourceOutputDirectory(mainProjectFolder + "/src/main/java");
        launcher.getEnvironment().setCommentEnabled(true);

The code is generated with `launcher.prettyprint()` method.

Expected behaviour:
* No UNUSED import statement gets added to the generated class.
* Source code of templates should not be copied to the sourceOutputDirectory.

Can you please point me out where and how I can fix it?

I use the following version of Spoon in my gradle build file:
    compile 'fr.inria.gforge.spoon:spoon-core:7.4.0'

Thank you,
Dmitry

xsebek:
This tests the example from #2987 

The issue is that a Launcher with `addTemplateResource` applies a template on that template and even adds it to output.

Fix #2987


xsebek:
Hey @monperrus, does this work as a test case?

Also do you know why the template evaluates the `CtExpression`?
I might be using it wrong as it doesn't happen with other code elements 😕

monperrus:
Yes, it's a good failing test case :-)

> Also do you know why the template evaluates the `CtExpression`?

What do you mean? Could you elaborate?

xsebek:
In the example class, the line is `assert 42 == 6 * 7` and the template (when checking for itself) should change it into this:
```java
// This is what I expect (SimpleIfAsserted.java)
if (42 == 6 * 7) {
    assert true;
} else {
    assert false;
}
```
but somewhere this is exposed and result is `assert true` in the spooned output.

See that this does not happen when a variable or a function is part of the expression:
```java
int x = 7;
assert 6 * 7 == 6 * x;
// changes to this:
if ((6 * 7) == (6 * x)) {
    assert true;
} else {
    assert false;
}
```

I mention this because it happens when the template is applied to itself 🤦 
```java
if (_asserted_.S())
    assert true;
else
    assert false;
// changes to this:
if (_asserted_.S()) {
    {
        assert true;  // this must have been if (true) that simplified
    }
} else {
    {
        assert false;
    }
}
```

xsebek:
This will block the `assertThat("SimpleAssert").isEqualTo("SimpleIfAsserted")` so I should probably change it to the second example above.

Is there a way to avoid this behaviour @monperrus? If not, then this might be a good thing to document/make an Issue for 😄

monperrus:
> In the example class, the line is `assert 42 == 6 * 7` and the template (when checking for itself) should change it into this:
> ```java
> // This is what I expect (SimpleIfAsserted.java)
> if (42 == 6 * 7) {
>     assert true;
> } else {
>     assert false;
> }
> ```
>  but somewhere this is exposed and result is `assert true` in the spooned output.

Agree. So it seems your test case highlight two bugs: one in the template code, and the other about writing the template to the output.

I'm looking at it.

monperrus:
@pvojtechovsky this is a bug in the matcher code behind the template engine. I don't know this code, you're the world expert, would you have some time to look at it? Thanks!

monperrus:
@pvojtechovsky @xsebek good news! There is no bug in the template engine.

@xsebek, the template engine is clever and does symbolic execution, because of `pb.setAutoSimplifySubstitutions(true);` which triggers calling `code.partiallyEvaluate()`

See #3586 

So the only bug is the rewriting bug

monperrus:
With #3586 and #3588 ready and merged, that should be fixed.

xsebek:
Thanks a lot, @monperrus! Those PRs already do together pass this test.
Sorry for putting you through so much work and if I can help let me know 😅

xsebek:
Thanks again @monperrus, your fixes work great 🎉 Feel free to merge/close this one 🙂

monperrus:
Thanks a lot @xsebek for the bug report and the test!

:+1::+1: for the rebase :)

PR #3586, opened 2020-09-11
refactor: template engine does not partially evaluate by default

monperrus:
in order to comply with the principle of least surprise


monperrus:
ping @nharrand ready to merge

PR #3587, opened 2020-09-11
refactor: add nullobject implementation for SpoonProgress

MartinWitt:
Currently we have a lot of null checks, before using `SpoonProgress`. How about setting the default implementation for all methods in `SpoonProgress` interface to a NOP and remove the null checks? Could be a breaking change if someone relays on calling `(getEnvironment().getSpoonProgress() == null)`. Comparable to this change is this usage of the `NoSourcePosition` class.
wdyt? And shall I add something like a `boolean isMock` method for the default implementation, allowing a client to check if a real `SpoonProgress` implementation is used?


monperrus:
I really like this PR, it makes the code much more readable

monperrus:
Thanks a lot @MartinWitt

PR #3588, opened 2020-09-11
REVIEW: introduce a separate factory for templates

monperrus:
plus some various performance optimizations found while debugging

is part of the solution for #3580


monperrus:
@nharrand ready

PR #3595, opened 2020-09-14
fix: fix non terminating PositionBuilder

Strum355:
PositionBuilder would get caught in the `<` in the annotation string, resulting in `o2` == `o1` and no advancement being made.
This fix ignores the opening angle bracket if it comes at the start of the search space

### DIFF
--- a/src/test/java/spoon/test/position/PositionTest.java
+++ b/src/test/java/spoon/test/position/PositionTest.java
@@ -90,6 +90,7 @@ import spoon.test.position.testclasses.FooLabel;
 import spoon.test.position.testclasses.FooLambda;
 import spoon.test.position.testclasses.FooMethod;
 import spoon.test.position.testclasses.FooStatement;
+import spoon.test.position.testclasses.AnnotationWithAngleBracket;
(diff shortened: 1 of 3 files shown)


src/test/java/spoon/test/position/PositionTest.java
```diff
@@ -424,6 +426,13 @@ public void testPositionGeneric() throws Exception {
 		assertEquals("protected static", contentAtPosition(classContent, position3.getModifierSourceStart(), position3.getModifierSourceEnd()));
 	}
 
+	@Test(timeout=10000)
```
MartinWitt, line 429: is there a reason for the timeout?

Strum355: The test specifically tests to make sure PositionBuilder terminates in a reasonable time. The existing bug caused non-termination, therefore if it doesn't terminate in a reasonable time, we assume it to be non-terminating

monperrus:
LGTM, all tests are green, will merge.

monperrus:
Thanks a lot @Strum355

PR #3607, opened 2020-09-18
fix: poetry has a place in software

monperrus:
"It was from the artists and poets that the pertinent answers came"

In Spoon, we do believe that poetry has a place in software.

This class counters the dark side of the force which won at https://github.com/rust-lang/rust/issues/13871


nharrand:
Haha! Will definitely merge!

monperrus:
and thus poetry resurrects on Maven Central @nddrylliog

PR #3614, opened 2020-09-24
chore: pin LSIF workflow to good commit

rvantonder:
`lsif-java@master` introduces some breaking commits for the action, so for using it with Spoon we can pin to a known good commit until we version stable releases of the tool/action.

Addresses https://github.com/INRIA/spoon/pull/3394#issuecomment-698150215.

### DIFF
--- a/.github/workflows/lsif.yml
+++ b/.github/workflows/lsif.yml
@@ -10,7 +10,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Generate LSIF data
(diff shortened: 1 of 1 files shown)


rvantonder:
Actually this should fix it, no need to merge the chang here: https://github.com/sourcegraph/lsif-java-action/commit/11dd91b4cba0f2a143914b14162d27d202f6726e

monperrus:
Thanks a lot!

PR #3624, opened 2020-10-01
WIP: Feature: allow lazy reference creation in TypeFactory creation

MartinWitt:
While developing a visitor for checking correctness of identifiers, i came across a problem in `TypeFactory`. Currently everytime we create a new `TypeFactory` we create TypeReferences, but mostly dont use them. This change removes the reference creation to later time. It is implemented with a supplier doing lazy initializing and caching.
This change should increase the performance and shouldn't cost much memory. 
Before i fix all test errors, is there any reason why we have public fields and not methods hiding the implementation? For some types we even have them. I would like to make all fields private and use methods.
Wdyt?

### DIFF
--- a/src/main/java/spoon/metamodel/MetamodelProperty.java
+++ b/src/main/java/spoon/metamodel/MetamodelProperty.java
@@ -203,7 +203,7 @@ public class MetamodelProperty {
 		if (valueType instanceof CtTypeParameterReference) {
 			valueType = ((CtTypeParameterReference) valueType).getBoundingType();
 			if (valueType == null) {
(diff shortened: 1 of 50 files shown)


monperrus:
Hi @MartinWitt 

This change is not backward compatible and may break much client code. In Spoon, we are giving very high importance to backward compatibility.

If the performance improvement is big after concrete measurement, we may still go this way, but we would have to deprecate first before removing.

xsebek:
I actually didn't notice the methods and used the fields to set types in expression, are the fields supposed to be internal?

monperrus:

Thanks!

PR #3628, opened 2020-10-03
refactor: remove unused imports

MartinWitt:
As title says removed unused imports in non-test classes and one used field. Only used vscode language server so there could be some more unused imports. Why we even allow unused fields is strange? Maybe write a architectural check?


monperrus:
LGTM. If you do want to it 100% clean it would be two PRs: one for unused imports, and one for unused fields (conceptually, it is a very different refactoring).

And then, in the PR related to unused fields, you could add this new cool architectural test :)

MartinWitt:
@monperrus can you rerun the CI and merge if all is green? Splitted the field part from this pr

monperrus:
Thanks a lot!

PR #3629, opened 2020-10-03
test: move test class to the correct place.

MartinWitt:
- Removed sysout from testcase to make the log more readable.
- Moved the testcase to correct place and deleted old file.

Ps: Sry for both small PRs #3628 and #3629 just wanted to finish my Hacktoberfest and i believe both could have been one. But these style problems needed to be fixed someday


monperrus:
Thanks!

PR #3630, opened 2020-10-03
chore: bump to jdt 3.23

MartinWitt:
Some testcase needed a small fix before updating the jdt version. Mostly removing the preview flag for java13.


monperrus:
Thanks a lot @MartinWitt

PR #3631, opened 2020-10-03
refactor: remove unused fields + corresponding architectural test

MartinWitt:
As discussed in #3628, here is a test case checking that every private field has a read. We dont need to check for writes, because a fieldRead without a fieldWrite before will never happen. Will writing this test case i found 8 unused fields. 
- One was a constructor parameter and i added a comment
- One is for later use and i commented it out
- 6 were simply unused.

### DIFF
--- a/src/main/java/spoon/metamodel/MMMethodKind.java
+++ b/src/main/java/spoon/metamodel/MMMethodKind.java
@@ -77,13 +77,12 @@ public enum MMMethodKind {
 	private final Predicate<CtMethod<?>> detector;
 	private final int level;
 	private final boolean multi;
-	private final int valueParameterIndex;
 
 	MMMethodKind(int valueParameterIndex, boolean multi, int level, Predicate<CtMethod<?>> detector) {
 		this.multi = multi;
 		this.level = level;
 		this.detector = detector;
-		this.valueParameterIndex = valueParameterIndex;
+		// valueParameterindex is never used
 	}
 
 	/**
--- a/src/main/java/spoon/metamodel/MMMethodKind.java
+++ b/src/main/java/spoon/metamodel/MMMethodKind.java
@@ -22,16 +22,18 @@ public enum MMMethodKind {
 	 * Getter.
 	 * T get()
 	 */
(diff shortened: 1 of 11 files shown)


src/main/java/spoon/metamodel/MMMethodKind.java
```diff
@@ -77,13 +77,12 @@
...
 	private final int level;
 	private final boolean multi;
-	private final int valueParameterIndex;
 
 	MMMethodKind(int valueParameterIndex, boolean multi, int level, Predicate<CtMethod<?>> detector) {
 		this.multi = multi;
 		this.level = level;
 		this.detector = detector;
-		this.valueParameterIndex = valueParameterIndex;
+		// valueParameterindex is never used
```
monperrus, line 85: Instead of commenting the fields and co, it would be better to simply remove them

src/test/java/spoon/test/architecture/SpoonArchitectureEnforcerTest.java
```diff
@@ -459,4 +465,23 @@ private static String computeDifference(Set<?> set1, Set<?> set2) {
 		}
 		return StringUtils.join(results, "\n");
 	}
+	private void checkFields(CtModel model) {
```
monperrus, line 468: out of curiosity how long does this new check take?

src/main/java/spoon/metamodel/MMMethodKind.java
MartinWitt: I was unsure about it, because the field is marked for future use.

src/test/java/spoon/test/architecture/SpoonArchitectureEnforcerTest.java
MartinWitt: some testing shows around 60 seconds but I have an idea to speed it up.

src/main/java/spoon/metamodel/MMMethodKind.java
monperrus: let's remove it completely

MartinWitt: You are correct, it's an Enum => we can simply remove the parameter because enums cant be subclassed and the value was private.

monperrus:
Thanks a lot @MartinWitt

Issue #3183, opened 2019-12-01
bug: sniper mode does not work when one removes a type


monperrus:
When one removes a top-level type in Sniper mode and that the type is the only one in the file, the file should be removed when pretty-printing.

Use case: automated removal of deprecated classes, see `removeDeprecatedMethods`.

monperrus:
To do it really nicely, we can do a `rm` or event `git rm` from Spoon if the source folder is the same as the target folder.

PR #3632, opened 2020-10-04
WIP: feature: add flag for skipping deleted types in printing

MartinWitt:
As described in #3183 this is the first iteration for the feature. There are some design decisions and printer specific questions. Do we add printer modes like this with flags in environment or is there a better place? Adding a method to an interface could be a breaking change, but we could add a default implementation to the method if we want to avoid this.

I would avoid using shell comands like `git rm,` but we could implement (and document) deleting types if sourceFolder = outputFolder. Assuming we implement this, do we want a flag to enable/disable this behavior, because removing source code as a result of a bad configuration is pretty bad experience for a user. We need to keep in mind not every project is part of a git repository and the code is not always restorable.   

### Open Work
There is still some work missing. These are more Printer specific topics.
#### Documenting
How and where should we document a new printer flag? Is there some manuell where I need to add it?
#### Naming
Some good idea for naming the flag? 
=> resolved `omitDeletedTypes`
#### Setter
Do we need to allow deactivating the flag after activating it?
#### Default mode
Maybe a topic we should discuss after heavy testing, but shall the mode be on as default?

### DIFF
--- a/src/main/java/spoon/compiler/Environment.java
+++ b/src/main/java/spoon/compiler/Environment.java
@@ -450,6 +450,7 @@ public interface Environment {
 		/** force everything to be fully-qualified */
 		FULLYQUALIFIED
 	}
+
 	/**
 	 * Sets the printer to skip printing empty Types.
 	 * A type is empty if it has no inner classes and no members.
@@ -459,7 +460,7 @@ public interface Environment {
 
 	/**
 	 * Returns if empty types are printed.
-   * A type is empty if it has no inner classes and no members.
+	 * A type is empty if it has no inner classes and no members.
 	 * Empty interfaces are not skipped, because they could be marker interfaces.
 	 * @return true if empty types are not printed, false otherwise
 	 */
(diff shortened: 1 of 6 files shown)


monperrus:
Thanks @MartinWitt 

I would have two concerns:
* I'm not sure it's the responsibility of `JavaOutputProcessor` to do code analysis. Its responsibility is to print everything that is in the model.
* Per #3183, we focus on deleted types, a deleted type is more conceptual, it's not a type that `hasOnlyDefaultConstructor(type)`

WDYT?

MartinWitt:
Yeah i thought about my approach later and you are correct. The printer should only check a flag. I changed the behavior to a flag in `CtType`. The `isRemoved` flag is set to true if the user calls `CtType.delete()`. But resetting it again to false, is more difficult. 
If we make 2 assumptions this could be possible: 
1. If a `CtType` has no parent it isn't part of the AST and can be omitted. 
2. If a deleted `CtType` gets a new parent we assume it's part of the AST again.

Assumption 1 seems correct, but about assumption 2 I'm unsure. Do you have some idea about identifying the adding a deleted type to AST?

monperrus:
Do you mean that we would have two alternatives for deletion, one is the flag, the other is automated inference?

If yes, I think the flag is enough.

monperrus:

Thanks!

PR #3633, opened 2020-10-04
clean: remove very old classes

monperrus:
remove very old outdated classes that are not necessary, not executed, and probably not executable anymore


monperrus:
@nharrand ready

PR #3640, opened 2020-10-07
fix(CtLineElementComparator): one can add two values to an annotation

### LINKED ISSUE #3639: Possible regression on adding values to CtAnnotation
I use Spoon to generate Java source files running Java 8. Using spoon-core 8.3.0-SNAPSHOT and 8.3.0-beta-10 I see a regression where adding a second value to a CtAnnotation makes the first added value not appear in the annotation in the created source.

To reproduce I have a simple annotation

    package tryspoon;

    public @interface Summary {
        String title();
        String date();
    }

and the following main class

    package tryspoon;

    import spoon.Launcher;
    import spoon.SpoonAPI;
    import spoon.reflect.declaration.CtAnnotation;

    public class Spooner {

            CtTypeReference<Summary> typeRef = spoon.getFactory().createCtTypeReference(Summary.class);
            CtAnnotation<Summary> annotation = spoon.getFactory().createAnnotation(typeRef);

            annotation.addValue("title", "First summary");
            annotation.addValue("date", "2020-10-05");

            // This outputs @tryspoon.Summary(date = "2020-10-05")
            System.out.println(annotation);

            // This outputs {date="2020-10-05", title=null}
           System.out.println(annotation.getAllValues());

           // This *does* output First summary
           System.out.println(annotation.getValue("title").toString());
           }
    }

The value of the first added key-value pair is seemingly being nulled when adding the second pair. In my actual code, this results in a created Java source file where the first annotation element is missing. I note that actually getting the value
works, that is `annotation.getValue("title").toString()` does return "First summary".

Originally I thought I saw the same behaviour using spoon-core 8.2.0 but I cannot reproduce with 8.2.0 now.

nharrand:
fix #3639 .

I seems that the entry set of the values contained an CtAnnotationImpl are ordered based on their source position. The issue is that so far `CtLineElementComparator#compare()` returns 0 if both element have `NoSourcePosition`. But this is the case when both elements are constructed programmatically. Hence the entry set arbitrarily retains only one of them.

My proposed solution is to make `CtLineElementComparator#compare()` returns 0 only if both elements have the same exact position or if both elements have `NoSourcePosition` and are equals. Otherwise we return -1 (i.e. order them aribtrarily). This way our entry set is pseudo order but does not delete unequals elements.

### DIFF
--- a/src/main/java/spoon/support/comparator/CtLineElementComparator.java
+++ b/src/main/java/spoon/support/comparator/CtLineElementComparator.java
@@ -19,14 +19,14 @@ import spoon.reflect.declaration.CtElement;
 public class CtLineElementComparator implements Comparator<CtElement>, Serializable {
 
 	/**
-	 * Returns 0 if o1 has the same position as o2, or both positions are invalid.
-	 * Returns -1 if o1 is before o2 in the file, or o1 has no valid position.
+	 * Returns 0 if o1 has the same position as o2, or both positions are invalid and o1.equals(o2).
+	 * Returns -1 if o1 is before o2 in the file, or o1 has no valid position or both positions are invalid !o1.equals(o2).
 	 * Returns 1 if o2 is after o1 in the file, or o2 has no valid position.
 	 */
 	@Override
 	public int compare(CtElement o1, CtElement o2) {
 		if (!o1.getPosition().isValidPosition() && !o2.getPosition().isValidPosition()) {
-			return 0;
+			return o1.equals(o2) ? 0 : -1;
 		}
 		if (!o1.getPosition().isValidPosition()) {
 			return -1;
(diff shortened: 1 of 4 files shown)


MartinWitt:
I tried to write down the math why this is wrong in #3519. If we return -1, we have a sleeping but and wait till the sort algorithm founds this.

src/main/java/spoon/support/comparator/CtLineElementComparator.java
```diff
@@ -19,14 +19,14 @@
...
-	 * Returns -1 if o1 is before o2 in the file, or o1 has no valid position.
+	 * Returns 0 if o1 has the same position as o2, or both positions are invalid and o1.equals(o2).
+	 * Returns -1 if o1 is before o2 in the file, or o1 has no valid position or both positions are invalid !o1.equals(o2).
 	 * Returns 1 if o2 is after o1 in the file, or o2 has no valid position.
 	 */
 	@Override
 	public int compare(CtElement o1, CtElement o2) {
 		if (!o1.getPosition().isValidPosition() && !o2.getPosition().isValidPosition()) {
-			return 0;
+			return o1.equals(o2) ? 0 : -1;
```
nharrand, line 29: If you want to respect the antisymetric property, this line could be replaced with
return o1.equals(o2) ? 0 : (o1.hashCode() < o2.hashCode() : -1 : 1);

MartinWitt: Nice fix. At first glance I see no problem with transitive and antisymmetric property. LGTM.

monperrus:
Thanks @nharrand

PR #3645, opened 2020-10-11
refactor: remove useless class SpoonReferenceBinding

monperrus:


nharrand:
@monperrus Do you know what `SpoonReferenceBinding` was used for originally?

monperrus:
No, that's old legacy.

nharrand:
Fair enough. If someone complains, we can still revert the change! ;-)

Issue #3638, opened 2020-10-06
add support for reusable architectural tests


monperrus:
Spoon is great for writing architectural tests, see http://spoon.gforge.inria.fr/architecture_test.html

Some tests we write could actually be useful for lots of Java projects, such as the recent one that checks for unused fields. 

We could package the reusable architectural tests in some way so that we can share them with the world.

monperrus:
> Architectural checks is the most exciting topic, because the other 3 are "solved" to some degree. Do you have some ideas for checks spoons is missing?

All tests that I thought of are already implemented :-) For identifying new ones, I would recommend to ask practitioners. I can introduce you to relevant persons if you'd like to proceed this way.

MartinWitt:
It would be a pleasure to proceed this way and gain more knowledge in this field.

MartinWitt:
After reading @vmassol mail, I would describe the current state as following:
Current state:

- Checkstyle doesn't provide easy check creation for complex tests.
- Sonarqube is too complex for projects even at xwiki size.
- Checks must run and fail the CI like unit tests.

So there is some need for a better solution. 
### Xwiki Checks
After reading the the xwiki tests, I tried to formulate them as architectural rule. The rules follow a common pattern.
Lets first look at the xwiki architectural checks:

#### Requirement/Rule:
Verify that `@Unstable` annotation don't stay too long (and that a `@since` annotation is present too)

 => Each element that has a `@Unstable` Annotation also has a `@since` annotation.
 => Each `@since` annotation isn't too old. 

#### Requirement/Rule: 
Verify that Script Services are not located in the internal package.

 => Each class(?) that is script service, doesn't have a package name containing `internal`. 

#### Requirement/Rule:
Verify that `@since` javadoc tags have the correct format (and that they don't contain several versions per tag). 

 => Each `@since` annotation has only 1 value, and the value is conform to format.

#### Requirement/Rule:
Verify none of the listed methods are called in our Java code

 => For every method invocation, the method name is never .../ is not part of a list.

#### Requirement/Rule:
Verify that JUnit4 and JUnit5 APIs are not both used at the same time in a test

 => For every method, holds there is no method/field from junit 4 AND junit 5.

#### Requirement/Rule:
Verify that JUnit tests don't output content to stdout or stderr.
 
  => For every method, holds there is no reference to system.out/system.err

### Generalization 

We see a base rule:

#### Base rule:
∀ astElement FilterRule(a) => condition(a)

#### FilterRule

[...]

PR #3654, opened 2020-10-16
WIP: Add an architecture check framework

MartinWitt:
Currently it's all draft and the API should be more in the focus, naming methods etc. are all wip.
For discussion see #3638. The is some configuration still wip like parent-pom, license, checkstyle but I solve this in a later stage.
The current stage is defining an API for a starter user.  As easy as possible and only as many configuration as needed.


MartinWitt:
#### New Feature
- Simplified the invocation of checks a lot. The following is all a new user needs. 
```java
	@Test
	public void allChecksMustRun() {
		SpoonArchitecturalCheckerImpl.createChecker().runChecks();
	}
``` 
- `src/main/java`and `src/test/java` are default paths now for srcModel and testModel
- Add options for multiple models.The user registers models  with `SpoonArchitecturalCheckerImpl.insertInputPath(String,String) or insertInputPath(String,CtModel)`.
- Add metadata to `@Architecture` String[]. The strings describe the models needed. Default is srcModel and testModel.
It's now possible to write tests for a specific submodel e.g a package only.
 
#### Other progress
- Add spoon-pom as maven parent

MartinWitt:
All checks in spoon expect `testInterfacesAreCtScannable` are now implemented in the framework as proof of concept. Now it's time to look at them and discus the api

MartinWitt:
Currently testcase check that no class uses junit4 and 5. Im unsure if we need more helper methods for checking names e.g. support regex.
```java
  @Architecture(modelNames = "testModel")
  public void junit4Junit5Test(CtModel testModel) {
    // get all Classes as precondition
    Precondition<CtClass<?>> pre = Precondition.of(DefaultElementFilter.CLASSES.getFilter());
    // check that no class has a junit4 and junit5 type usage.
    // a constraint will call the ErrorReporter if all conditions are true. Each condition is connected by logical AND by default.
    // You write the constraint as forall Elements X must hold and the runner checks that there is no where not X is true. 
    Constraint<CtClass<?>>  con = Constraint.of(new ErrorReporter(),
      // check for any junit 5 usage
      (element) -> element.getReferencedTypes().stream().anyMatch(clazz -> clazz.getQualifiedName().startsWith("org.junit.jupiter")),
      // check for any junit 4 usage
      (element) -> element.getReferencedTypes().stream().anyMatch(clazz -> clazz.getQualifiedName().startsWith("org.junit")));
      // run the check
      ArchitectureTest.of(pre, con).runCheck(testModel);
  }
  private class ErrorReporter implements IError<CtClass<?>> {

    @Override
    public void printError(CtClass<?> element) {
      System.out.println(String.format("There's a mix of JUnit4 and JUnit5 APIs at [%s]", element.getPosition()));
    }
    
  }
```

monperrus:
Waoow, big piece of work!

In your examples, I understand that there are two APIs:

* one for the user would use spoon-architecture-checks, and in the PR there is a single example: `new SpoonArchitecturalCheckerImpl.Builder().useDefaultPath().addReportPrinter(new ShellPrinter()).build().runChecks();`

* one for the user would specify new architecture tests, with interface `IConstraint` being the focal point.

Is that correct?

MartinWitt:
there are 2 APIs for running the checks:
- `new SpoonArchitecturalCheckerImpl.Builder().useDefaultPath().addReportPrinter(new ShellPrinter()).build().runChecks();`
- `new SpoonArchitecturalCheckerImpl.createChecker().runChecks()`
The second is the easy way to start without a lot of options and default settings.
The first one allows more options like specific report printer, more complex path options. And we can allow more settings in the first option if there is some need.

A testcase is always a `ArchitectureTest(Precondition, IConstraint)`. Precondition is only for filtering of the elements.
IConstraint is the focal point, where a user formulates his architectural rules.

MartinWitt:
An explanation how to write the first architecture rule could look like the following. 

### Your first Architecture-Rule

Lets start with the rule. The rule is "Every testmethod has the a name starting with 'test'".
 A rule is a method with the annotation `@Architecture`.  Both arguments are models from your source code. We look at them later. The method looks like this:
```java
@Architecture
public void testMethodNamesStartWithTest(CtModel srcModel, CtModel testModel) {
}
```
Lets look back at the rule. We want all methods that have the `@Test` and check if the methodname starts with `test`.
The precondition is all methods annotated with `@Test` and the constraint is the methodname starts with `test`.
A architecture rule consists of two parts:
- A preconditions specifying the elements, that must hold a property.
- A constraint describing a condition all elements must hold e.g. methodname starts with test.

#### Precondition

We already have predefined filters for faster start. We want all filter that gives us all methods.
```java
Precondition<CtMethod<?>> pre =	Precondition.of(
              DefaultElementFilter.METHODS.getFilter(),
              Visibility.PUBLIC,
              AnnotationHelper.hasAnnotationMatcher(Test.class));
```
You can ignore the CtMethod generic for now. We create a precondition with the static method `of` having 2 arguments. The first an element selector, here `DefaultElementFilter.METHODS.getFilter()` and the second argument is varargs of Predicates. These predicates help filtering the elements before checking the condition. We have 2 Predicates here, the first checks if the method is public. A junit testcase is always public. The second method checks if the method has the `@Test` annotation.
After we created the precondition lets look at the constraint.

#### Constraint
The constraint is created with the static `Constraint.of(IError, Predicate<?>...) `method. The `IError` is for reporting errors and for simplicity for now a simple print of the element. Our rule was that every testmethods name starts with test. In the precondition we filtered all test methods, so now we only need to check if the name starts with "test". For naming checks there is a class with predefined default methods. Naming.startsWith checks if a name starts with the given string. 
```java
		Constraint<CtNamedElement> con = Constraint.of(System.out::println, Naming.startsWith("test"));
```
#### ArchitectureTest
Now we have a the precondition and constraint and we only need to create the ArchitectureRule.
The rule is create by the static method `ArchitectureTest.of(Precondition<T>, Constraint<T>)`. The `runCheck(testModel);` tells the runner which model should be checked, here the testModel.
Thats simply `ArchitectureTest.of(pre, con).runCheck(testModel);`

#### Runner
Now you may ask yourself how can I run this test? We have a runner integrated. The following line integrates your architecture rule into junit and allows easy usage.
```java	
@Test
	public void allChecksMustRun() {
		SpoonArchitecturalCheckerImpl.createChecker().runChecks();
       }
```

#### Wrapping it up

The full example rule is the following:
````java
	@Architecture
	public void methodNameStartsWithTest(CtModel srcModel, CtModel testModel) {
		Precondition<CtMethod<?>> pre =	Precondition.of(DefaultElementFilter.METHODS.getFilter(), Visibility.PUBLIC, AnnotationHelper.hasAnnotationMatcher(Test.class));
		Constraint<CtNamedElement> con = Constraint.of(System.out::println, Naming.startsWith("test"));
		ArchitectureTest.of(pre, con).runCheck(testModel);
	}
```

monperrus:
This is great and elaborate text. Would you put it directly the a README.md to be added in the PR?

monperrus:
What does the user have to do so that `@Architecture` methods are recognized and run?

MartinWitt:
> This is great and elaborate text. Would you put it directly the a README.md to be added in the PR?
I would start creating a doc folder and start collecting documentation there, because I believe a user needs way more documentation. But it could be added to the Readme.md later.
> What does the user have to do so that @Architecture methods are recognized and **run?**
`SpoonArchitecturalCheckerImpl.createChecker().runChecks();` for a fast start with default options.

monperrus:
> I would start creating a doc folder and start collecting documentation there, because I believe a user needs way more documentation

perfect

monperrus:
> `SpoonArchitecturalCheckerImpl.createChecker().runChecks()`

Do you mean that this scans `src/test/java` to find new architectural tests?

MartinWitt:
The runner searches in the testmodel for architectural tests, see [AbstractSpoonArchitecturalChecker.runChecks()](https://github.com/INRIA/spoon/pull/3654/files#diff-5254fe2c137597a806dba3889b03ad8921128406e2052cb1b236edc2348b06d8R21) . By default the testmodel has the path `/src/test/java` but a user can provide a different path for the testmodel.

monperrus:
That's clear.

Are you OK if I make a pass on the documentation in the Markdown file and push to your branch?

MartinWitt:
Sure go ahead.

MartinWitt:
As we still have the luxury to change bigger things, today i thought about generic meta models. Sometimes a user has non java constraints for his project, like every maven plugin has a version specified. There exists some awesome tools but not always. for all users needs. 
### What is changed?
With a few generic changes a ArchitectureTest is now a `ArchitectureTest<T, M>` where T is the model element type e.g. CtMethod and M is the model type e.g. CtModel. The testcase consists of a `IPrecondition` which is a `Function<M,Collection<T>>` and a `Checkable<T> `. 
### What does this change?
For a user nothing, but it allows the extension for different meta models e.g. a meta model for maven config, etc. It's not important now, but if we dont think about possible extensions now we can never allow them without breaking changes. 
It's important to note, i plan not to add a maven meta model before any stable release and it could never happen.

monperrus (commented):
Great. A few comments:

* rename SpoonArchitecturalCheckerImpl to SpoonArchitecturalChecker?
* reuse gitignore and checkstyle of parents
* add the module to the CI pipeline
* merge the two MD files together
* add a link in http://spoon.gforge.inria.fr/architecture_test.html

Thanks!

MartinWitt:
To finish this pr up i believe there are 2 points open. License and near 100% test coverage.
- For license I would keep it simple and use the dual license spoon is using if no one has any suggestion.
- For tests I hope I find time to create them this week.
The Spoon integration/ replacing the architecture checks would I split in a other PR.

MartinWitt:
I reached the state, where the API is final and the test coverage is high. The last step is creating examples in `/test/examples/`but the core source code is ready for review.

MartinWitt:
After my exmas break I´m back to finish this PR. The current plan is to create a stable version 0.01 and grow to version 1.0
After version 0.01 I would like to switch from 1 big PR to smaller PRs for high level helper classes. These classes provide high level helpers and it`s easier for me to design them 1 by 1. A helper class could look like this.
```java
class Switches {
boolean hasDefaultCase(CtSwitchcase switch)
boolean hasDefaultCaseLastCtSwitchcase switch)
int getDefaultCasePosition(CtSwitchcase switch)
CtCase getDefaultCase(CtSwitchcase switch)
boolean hasArrowLabels
boolean hasDotLabels
boolean hasMultiCases
....
}
```
These methods should allow easier usage for users and reduce the knowledge about spoon meta model. An user shouldn't need to know that default cases in spoon have null as case expressions. There is a really old issue #2101 which was concerned about code quality. Fixing some messy old code in spoon should be a good process allowing to finding good rules for things we dont want in our code.

monperrus:

Thanks!

PR #3652, opened 2020-10-15
chore: remove uncovered, unnecessary code

monperrus:
internal code for Javadoc parsing

PR #3660, opened 2020-10-22
Review: refactor(javadoc): remove old pair class and replace it with commons-lang3 class

MartinWitt:
Title says all. Was discussed in #3652
a = `getLeft()` and b = `getRight()`
Edit: There is an import reording moving an import to it's correct place. I can remove it, if you want super minimal diffs.


nharrand:
Thanks for the PR! It looks good to me.

> Edit: There is an import reording moving an import to it's correct place. I can remove it, if you want super minimal diffs.

No that's fine as it is.

PR #3661, opened 2020-10-22
refactor(ArchitectureEnforcer): replace treeSet with hashset

MartinWitt:
A few lines above the code states: 
// contract: Spoon's code never uses TreeSet constructor, because they implicitly depend on Comparable (no static check, only dynamic checks)
So lets fulfil the contract. Simply TreeSet -> HashSet. Shouldn't change anything.


monperrus:
Thanks, LGTM.

monperrus:
Thanks!

PR #3664, opened 2020-10-23
feature: add support for pretty-printing Java 15 text blocks

### LINKED ISSUE #3605: add support for Java 15
See feature list at https://openjdk.java.net/projects/jdk/15/

LakshyAAAgrawal:
Implement CtTextBlock class and add factories 
Closes #3605


LakshyAAAgrawal:
Todo:
* Modify JDTBasedSpoonCompiler to create TextBlock when parsed literal is TextBlock.
* Modify existing tests to check creation CtTextBlock
* Write additional tests

LakshyAAAgrawal:
The ASTVisitor abstract class in org.eclipse.jdt.internal.compiler does provide TextBlock class as part of the package org.eclipse.jdt.internal.compiler.ast; However, in the ASTVisitor class, which is the base class for JDTTreeBuilder, the visit and endvisit methods are only defined for StringLiteral class in JDT AST, but not for TextBlock. I tried to retrieve the token source for it too from the StringLiteral, however, it just gives the string content, and not the source token. If the source token was provided, we could simply detect 3 consecutive (") in the beginning and confirm it being a TextBlock.

MartinWitt:
>  If the source token was provided, we could simply detect 3 consecutive (") in the beginning and confirm it being a TextBlock.
JLS 3.10.6 stats the grammar for textblocks `" " " {TextBlockWhiteSpace} LineTerminator {TextBlockCharacter} " " "`` and TextBlockWhiteSpace == Whitespace. So the regex `"""\s*\n.*"""` should do the trick. Maybe it helps you

LakshyAAAgrawal:
> > If the source token was provided, we could simply detect 3 consecutive (") in the beginning and confirm it being a TextBlock.
> 
> JLS 3.10.6 stats the grammar for textblocks `" " " {TextBlockWhiteSpace} LineTerminator {TextBlockCharacter} " " "` and TextBlockWhiteSpace == Whitespace. So the regex `"""\s*\n.*"""` should do the trick. Maybe it helps you

Thank you very much for the reply. Actually my question had been that I am not able to get the token source from a call to getSource in the first place. For example, if the textblock was:
```
String s = """
        Test String""";
```
Then calling ```stringLiteral.source``` returns the string - "Test String" instead of "\"\"\"\n        Test String\"\"\"".

But the issue has been resolved now in the latest commit. The stringLiteral that was being passed to the visitor was actually an instanceof TextBlock from jdt ast, and I checked it using instanceof.

src/main/java/spoon/reflect/factory/CodeFactory.java
```diff
@@ -269,6 +270,19 @@ public CodeFactory(Factory factory) {
...
+	/**
+	 * Creates a TextBlock with the given string value.
+	 * @param value
+	 * 		the string value of the literal
+	 * @return a new literal
+	 */
+	public CtTextBlock createTextBlock(String value) {
+		CtTextBlock textblock = factory.Core().createTextBlock();
+		textblock.setValue(value);
+		// TODO set textblock.setType
```
monperrus, line 282: fix or remove?

monperrus (commented):
Really cool PR! LGTM but a minor comment to fix.

src/main/java/spoon/reflect/factory/CodeFactory.java
LakshyAAAgrawal: Thank you very much for pointing out. I have made some changes. Kindly give them a look.

LakshyAAAgrawal:
I believe that with this PR, Java 15 support in Spoon is complete. Relevant changes in the webpage/documentation can be made either now, or upon the next release.

monperrus:
Thanks @LakshyAAAgrawal impressive work.

LakshyAAAgrawal:
Thanks a lot for merging @monperrus

PR #3668, opened 2020-10-26
refactor: add support for malformed files through inheritance

darius-sas:
A common issue that I get while analysing big projects is that the analysis fails just because a few files contain issues, or special characters that are not allowed by the Java compiler: like for example archetype resource files:

```Java
#set( $symbol_pound = '#' )
#set( $symbol_dollar = '$' )
#set( $symbol_escape = '\' )

import ${package}.path.to.Clazz;
```

This very simple modification ensures that Spoon provides a best-effort attempt at parsing a certain project without failing just because of one malformed file. In case an error occurs, a message is printed on the console, so the user is notified.

I am not very practical of Spoon's internals, so there might be better solutions to this. If that's the case, pointers on how to implement a better one are welcome :)

### DIFF
--- a/src/main/java/spoon/Launcher.java
+++ b/src/main/java/spoon/Launcher.java
@@ -596,7 +596,7 @@ public class Launcher implements SpoonAPI {
 	 * 		the factory this compiler works on
 	 */
 	public SpoonModelBuilder createCompiler(Factory factory) {
-		SpoonModelBuilder comp = new JDTBasedSpoonCompiler(factory);
+		SpoonModelBuilder comp = getCompilerInstance(factory);
 		Environment env = getEnvironment();
 		// building
 		comp.setBinaryOutputDirectory(jsapActualArgs.getFile("destination"));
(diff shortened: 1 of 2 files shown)


monperrus:
Not sure we want to silence so many errors.

What we can do instead is wrapping this call in a method that can be overridden, such that you create your own SilencingJDTCompiler (see extension point in Launcher).

darius-sas:
That sounds like a good option. I tried to look on how to provide to a Launcher your own `SpoonModelBuilder` but it seems to always [instantiate](https://github.com/INRIA/spoon/blob/4ccbf0a1e7cc092d2b75aac767d3797eb10c0399/src/main/java/spoon/Launcher.java#L599) a `JDTBasedSpoonCompiler`. It's very likely ofc that I am wrong :P 

Another one would be to add an option (e.g. "silenceParseErrors", or something like that) that silences the error only if the user explicitly asks for it:

```Java
try {
    consumer.accept(unit);
} catch (Exception e) {
   if(!silenceParseErrors){
       throw e;
   }
   // log the error and continue parsing files.
}
```

Thanks for the help.

monperrus:
> Another one would be to add an option (e.g. "silenceParseErrors", or something like that) that silences the error only if the user explicitly asks for it:

In Spoon, for rare use cases, we avoid binary switches, because they are are hard to test and maintain.

darius-sas:
Ok thanks for the info. I'll try to implement the solution you mentioned.

darius-sas:
I made the modifications we discussed and also added an overridable method in Launcher that creates the compiler object which should allow users to instatiate their own compiler that overrides the visiting the of CUDs.

monperrus:
Thanks a lot, this refactoring is good, will merge.

monperrus:
Thanks a lot @darius-sas

darius-sas:
Thank you!

PR #3670, opened 2020-10-29
feature: add CtModifiable#isTransient and related methods

### LINKED ISSUE #3651: Feature: CtModifiable#isTransient
Looking at CtModifiable class, there a lot of convenience method, but `isTransient` is missing? Is there any reason for it? 
A implementation could be the following. 
```java
public boolean isTransient() {
return this.hasModifier(ModifierKind.TRANSIENT);
}
```

http://spoon.gforge.inria.fr/mvnsites/spoon-core/apidocs/spoon/reflect/declaration/CtModifiable.html

Edit: I looks like some more Modifiers from `ModifierKind` are missing. Feel free to implement them also.
I created this as good first issue, because it's relative easy and no high priority change. Some1 can have an easy start with this issue.

Artamm:
fix #3651 
Implemented modifiers from ModifierKind (`Transient`, `Volatile`, `Synchronized`, `Native`, `Strictfp`) for  CtModifiable class.

### DIFF
--- a/src/main/java/spoon/reflect/declaration/CtModifiable.java
+++ b/src/main/java/spoon/reflect/declaration/CtModifiable.java
@@ -117,4 +117,43 @@ public interface CtModifiable extends CtElement {
 	 */
 	@DerivedProperty
 	boolean isAbstract();
+
+	/**
+	 * Returns true if it contains a transient modifier (see {@link #hasModifier(ModifierKind)})
+	 */
+
+	@DerivedProperty
+	boolean isTransient();
+
+
+	/**
+	 * Returns true if it contains a volatile modifier (see {@link #hasModifier(ModifierKind)})
+	 */
+
+	@DerivedProperty
+	boolean isVolatile();
+
+	/**
+	 * Returns true if it contains a synchronized modifier (see {@link #hasModifier(ModifierKind)})
+	 */
+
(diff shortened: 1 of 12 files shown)


MartinWitt:
LGTM, nice work!

Artamm:
> LGTM, nice work!

Thank you!

monperrus:
Hi @Artamm 
Thanks a lot for your contribution.

In Spoon we aim at strongly testing all functionalities. 

Could you add test cases to specify those new methods?

Thanks!

Artamm:
@monperrus , thank you for your response. 
I added some test cases in `ModifiersTest`. Maybe I should add another test case/ edit what I did?

monperrus:
Thanks a lot @Artamm , that's good testing. Do you confirm that we test all those new methods with both "true" and "false"?

Artamm:
@monperrus, I think yes, these tests are enough to cover all cases for new methods.
Well, I had in mind to add fields in `ConcreteClass` for additional checks too, but this class extends `AbstractClass` - so it'll just duplicate the code, IMHO.
If it's needed to add something - let me know, I will add.

monperrus:
Big thanks for your contribution @Artamm and thanks a lot @MartinWitt for the shepherding

PR #3676, opened 2020-11-03
fix: Fix NullPointerException in SniperJavaPrettyPrinter.printTypes

### LINKED ISSUE #3675: bug: Calling printTypes on SniperJavaPrettyPrinter always causes a NullPointerException
Hi,

The `printTypes` method is broken for `SniperJavaPrettyPrinter`, it always causes a `NullPointerException`. It's caused by the implementation of `printTypes` in `DefaultJavaPrettyPrinter` passing `null` for the CU to `calculate`

https://github.com/INRIA/spoon/blob/1cc3c746cc79674525be4516834543fbc66a13e6/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java#L2006-L2010

and the `SniperJavaPrettyPrinter` assuming that the CU is not null

https://github.com/INRIA/spoon/blob/1cc3c746cc79674525be4516834543fbc66a13e6/src/main/java/spoon/support/sniper/SniperJavaPrettyPrinter.java#L112-L116

I don't think it's possible for the sniper printer to do its job without the CU. Working on a PR.

slarse:
Fix #3675 

This PR fixes the problem explained in #3675 by overriding `printTypes` in `SniperJavaPrettyPrinter`, and trying to resolve a compilation unit from the types passed in. If no compilation unit can be found, or the types have mismatching compilation units, it raises an exception. Two new tests verify the behavior within a reasonable amount of certainty.

### DIFF
--- a/src/main/java/spoon/support/sniper/SniperJavaPrettyPrinter.java
+++ b/src/main/java/spoon/support/sniper/SniperJavaPrettyPrinter.java
@@ -12,7 +12,6 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.Deque;
 import java.util.List;
-import java.util.Optional;
 
 import spoon.OutputType;
 import spoon.SpoonException;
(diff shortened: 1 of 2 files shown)


slarse:
This is ready for review.

monperrus:
Thanks! LGTM, will merge after the typical one-day latency for everybody to see it.

monperrus:
Thanks a lot @slarse

Issue #3691, opened 2020-11-12
Discussion(CI): Which jdks for testing?


MartinWitt:
As we have a new JDK every half year and [java 16](https://openjdk.java.net/projects/jdk/16/) isn't that far away. We could starting a discussion what JDK versions are interesting for CI usage. Currently we use 8, 10 and 11. We could use all JDKs for testing, but the CI takes already really long for testing.
### travis limits:
A job gets canceled 
- When a job produces no log output for 10 minutes.
- When a job on a public repository takes longer than 50 minutes.
- When a job on a private repository takes longer than 120 minutes.

### JDKs
- 8 minimal version we support, important to test. 
- 9 breaking changes because module system
- 10 only `var` as interesting change, but shouldn't break anything.
- 11,12,13,14 nothing really interesting
- 15 latest/bleeding edge is always good to test

We could create a job for each JDK, or choose 3 (?) interesting for default CI and run with all JDKs only when some condition is met(if travis supports this, probably exploration needed). The condition could be : 
- A PR is marked as review
- A merge/commit is done to master
- A new version/snapshot gets released
### Other OS containers
Another discussion is, if we want to use windows containers in travis for more testing. Java on Windows should produce the same results as on unix, but sometimes tests are platform dependent without a good reason. For MacOS the situation is more complex, because travis changed the costs and MacOS seems not free anymore. Supporting MacOSin the CI, would either need free credits from travis or a macos docker container.
### Discussion 
Which JDK versions do you think are the best/most interesting for using in the CI?
Shall we extend the CI with windows runners?
Shall we split the CI in different complexity depending on the PR state?
Or do you want performance increases?

monperrus:
> Which JDK versions do you think are the best/most interesting for using in the CI?

It would be great to replace Java 10 by Java 15

> Shall we extend the CI with windows runners?

We had that at one point with AppVeyor, but it was too unstable. We also had a Windows slave on the Spoon Jenkins but it did not work well. Any effort on this would be welcome.

MartinWitt:
> It would be great to replace Java 10 by Java 15

Should be done with the new PR #3692 

> We had that at one point with AppVeyor, but it was too unstable. We also had a Windows slave on the Spoon Jenkins but it did not work well. Any effort on this would be welcome.

As there is a lot of bug backlog and I want to finish the architecture framework. I would add this to the end of backlog, but if somebody wants to research about travis go ahead. Maybe an issue helps finding someone with travis knowledge-

[...]

PR #3689, opened 2020-11-12
REVIEW chore: default JDK is LTS Java 11

monperrus:

PR #3692, opened 2020-11-12
Review: chore(jdk15): add jdk15 ci task

MartinWitt:
A comment in #3691 mentioned replacing jdk10 with 15. This PRs tries to add a new CI task and is for exploration what is breaking.


MartinWitt:
2 changes were needed because object has 11 instead of 12 methods in newer JDKs. I thought i made this change for newer JDKs already(?) but now it should work. I believe merging #3689 and then rerunning the CI is the way to go, because I don't want to break the CI. I removed java10 as commented by @monperrus.

nharrand:
LGTM! Thanks for the PR!

PR #3694, opened 2020-11-13
refactor: improve error message when sniper printer is set after building model

### LINKED ISSUE #3693: bug: Crash in SniperJavaPrettyPrinter.calculate if the printer is set after building the model
Hi,

As the title says, `SniperJavaPrettyPrinter.calculate` crashes if the printer is set with `Environment.setPrettyPrinterCreator` _after_ one builds the model. The reason for this is that when building the model, this happens: https://github.com/INRIA/spoon/blob/55bdd962eeb132ad93700082c2cce028396da428/src/main/java/spoon/support/compiler/jdt/JDTBasedSpoonCompiler.java#L128-L131

And in the call to calculate, the sniper tries to access that change collector (with `getChangeCollector()`: https://github.com/INRIA/spoon/blob/55bdd962eeb132ad93700082c2cce028396da428/src/main/java/spoon/support/sniper/SniperJavaPrettyPrinter.java#L138

I have a use case where I need to collect some information from the model to feed to a printer preprocessor, which one typically sets in the supplier passed to `setPrettyPrinterCreator`. That's how I encountered this problem. It's easy enough to work around (e.g. by setting a "dummy" sniper printer before building the model), but the crash isn't easy to understand if you don't know too much about Spoon.

Any ideas on a solution? The easy solution would of course be to always attach the change collector, but I don't know what kinds of overhead that would impose on non-sniper use cases.

slarse:
Fix #3693 

The sniper printer must be set in the environment before building the model in order for the change collector (which is otherwise unnecessary overhead) to be attached. The error message you currently get if you set the sniper after building the model is uninformative to the uninitiated, and just says:

> ChangeCollector was not attached to the environment

This PR augments that error message to say:

> ChangeCollector was not attached to the environment. This typically means that the Sniper printer was set after building the model. It must be set before building the model.

I think this should be less confusing for non-expert users that run into this error (I was confused at first).


slarse:
I'm not sure what tag (i.e. `feat:` `doc:` etc) to use in the PR title, so feel free to choose one that's appropriate!

monperrus:
Thanks! Decided to go for a `refactor:` commit prefix: same behavior, but better for the developer.

Issue #3680, opened 2020-11-08
ParentNotInitializedException when using Java 14 switch expressions with cases which have multiple values


davidog23:
spoon-core version: 8.3.0
Code to reproduce the bug:
```java
import spoon.Launcher;
import spoon.reflect.declaration.CtClass;
import spoon.reflect.declaration.CtElement;

public class A {
    public static void main(String[] args) {
        CtClass l = Launcher.parseClass("class A { { int x = switch(1) { case 1, 3 -> 0; default -> 1}; } }");
        for (CtElement a : l.asIterable()) {
            a.getParent();
        }
    }
}
```
```
Exception in thread "main" spoon.reflect.declaration.ParentNotInitializedException: parent not initialized for class spoon.support.reflect.code.CtLiteralImpl (unknown file)
	at spoon.support.reflect.declaration.CtElementImpl.getParent(CtElementImpl.java:366)
	at A.main(A.java:9)
```
The CtLiteralImpl which represents the number 3 throws a ParentNotInitializedException when getting its parent. I guess that is not intended behaviour.

MartinWitt:
Fixed by #3689

PR #3695, opened 2020-11-13
Review: Fix(CtCase): add fix for missing parents

MartinWitt:
Potential fix for #3680. I'm unsure why the problem lived for so long, but setting the parent in `addCaseExpression` should fix it. Is there any other place the parent should have been set?


nharrand:
It seems good to me. if we discover that we forget to set the parent in other cases we will fix it then!
Thanks for the PR!

PR #3696, opened 2020-11-15
fix: fix incorrect parent for type reference of a CtParameter

### LINKED ISSUE #3671: parameter.getReference() would set parent of parameter type, and Spoon cannot get declaration of generic method by reference.
I'm working with Spoon 8.3.0, and I want to analyze some Java8 code.

Here is a snippet to reproduce the issue

```
        spoon.addInputResource("/home/usr/TestGeneric.java");
        spoon.buildModel();

        Map<CtExecutableReference<?>, List<CtParameterReference<?>>> exe2Parameter =
                new HashMap<>();
        for (CtExecutable<?> cm : spoon.getModel().getRootPackage()
               .getElements(new TypeFilter<>(CtExecutable.class))) {

            List<CtParameterReference<?>> pList = new ArrayList<>();
            for(CtParameter<?> parameter : cm.getParameters()) {
                //getReference() change the parameter.getType().getParent()
                pList.add(parameter.getReference());
            }
            exe2Parameter.put(cm.getReference(), pList);
        }
        for (CtExecutableReference<?> exeRef : exe2Parameter.keySet()) {
            //Getting declaration of generic method would throw exception.
            System.out.println(exeRef.getExecutableDeclaration());
        }
```
TestGeneric.java
```
public class TestGeneric {
    public static class Cell<T> {
        public T val;
        public T getVal() { return val; }
        public void setVal(T val) { this.val = val;}
    }

    public static class Day {
        public int day;
        public Day() { }
        public Day(int val) {this.day = val;}
    }

    public void bar() {
        Cell<Day> c = new Cell<>();
        Day e = new Day();
        c.setVal(e);
    }
}
```

```
Exception in thread "main" spoon.SpoonException: Cannot resolve type erasure of the type parameter reference, which is not able to found it's declaration.
	at spoon.support.reflect.reference.CtTypeParameterReferenceImpl.getTypeErasure(CtTypeParameterReferenceImpl.java:180)
	at spoon.support.reflect.declaration.CtTypeImpl.isSameParameter(CtTypeImpl.java:752)
	at spoon.support.reflect.declaration.CtTypeImpl.hasSameParameters(CtTypeImpl.java:744)
	at spoon.support.reflect.declaration.CtTypeImpl.getMethod(CtTypeImpl.java:718)
	at spoon.support.reflect.reference.CtExecutableReferenceImpl.getCtExecutable(CtExecutableReferenceImpl.java:121)
	at spoon.support.reflect.reference.CtExecutableReferenceImpl.getExecutableDeclaration(CtExecutableReferenceImpl.java:113)
	at GetInvocation2.main(GetInvocation2.java:30)

```

Is this a bug, or I just need to avoid getting reference of method parameter before getting method declaration?

andrewbwogi:
close #3671

### DIFF
--- a/src/main/java/spoon/reflect/factory/ExecutableFactory.java
+++ b/src/main/java/spoon/reflect/factory/ExecutableFactory.java
@@ -79,6 +79,8 @@ public class ExecutableFactory extends SubFactory {
 	public <T> CtParameterReference<T> createParameterReference(CtParameter<T> parameter) {
 		CtParameterReference<T> ref = factory.Core().createParameterReference();
 		ref.setSimpleName(parameter.getSimpleName());
+
+		// true when parameter comes from a lambda expression
 		if (parameter.getType() == null) {
 			ref.setType(parameter.getType());
 		} else {
--- a/src/main/java/spoon/reflect/factory/ExecutableFactory.java
+++ b/src/main/java/spoon/reflect/factory/ExecutableFactory.java
@@ -79,7 +79,11 @@ public class ExecutableFactory extends SubFactory {
 	public <T> CtParameterReference<T> createParameterReference(CtParameter<T> parameter) {
 		CtParameterReference<T> ref = factory.Core().createParameterReference();
 		ref.setSimpleName(parameter.getSimpleName());
-		ref.setType(parameter.getType().clone());
+		if(parameter.getType() == null) {
+			ref.setType(parameter.getType());
+		} else {
+			ref.setType(parameter.getType().clone());
+		}
 		return ref;
 	}
 
--- a/src/main/java/spoon/reflect/factory/ExecutableFactory.java
+++ b/src/main/java/spoon/reflect/factory/ExecutableFactory.java
@@ -79,7 +79,7 @@ public class ExecutableFactory extends SubFactory {
 	public <T> CtParameterReference<T> createParameterReference(CtParameter<T> parameter) {
 		CtParameterReference<T> ref = factory.Core().createParameterReference();
 		ref.setSimpleName(parameter.getSimpleName());
-		if(parameter.getType() == null) {
+		if (parameter.getType() == null) {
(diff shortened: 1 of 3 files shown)


monperrus:
Thanks @andrewbwogi for the work on this fix. Could you have a look at the CI failures?

andrewbwogi:
This solution works but it hides an implicit difference between ordinary parameters and parameters in lambda expressions. Is it better to create a CtLambdaParameter class?

src/main/java/spoon/reflect/factory/ExecutableFactory.java
```diff
@@ -79,7 +79,11 @@ public CtAnonymousExecutable createAnonymous(CtClass<?> target, CtBlock<Void> bo
 	public <T> CtParameterReference<T> createParameterReference(CtParameter<T> parameter) {
 		CtParameterReference<T> ref = factory.Core().createParameterReference();
 		ref.setSimpleName(parameter.getSimpleName());
-		ref.setType(parameter.getType());
+		if (parameter.getType() == null) {
+			ref.setType(parameter.getType());
+		} else {
+			ref.setType(parameter.getType().clone());
+		}
```
andrewbwogi, line 86: This solution works but it hides an implicit difference between ordinary parameters and parameters in lambda expressions. Is it better to create a CtLambdaParameter class?

monperrus:
>  Is it better to create a CtLambdaParameter class?

It would probably be overkill.

Do you mean that we don't have a fully backward compatible fix for now?

andrewbwogi:
The fix only expands information in the AST so it should be backwards compatible.

monperrus:
Great, thanks @andrewbwogi !

In Spoon, we always have a one line contract in natural language for each test:

    // contract: after adding an element in a list, the size is incremented

Could you add the contract?

Then we merge!

monperrus:
Thanks @andrewbwogi

PR #3701, opened 2020-11-17
fix: Fix sniper printing of comments when changing modifiers

### LINKED ISSUE #3697: bug: Sniper printer drops whitespace after comment if next element is type member with modified modifier list
I've encountered a problem where the sniper printer will not insert newlines correctly on modified or new elements. I've encountered this when inserting import statements, as well as when modifying fields. This is a big problem for the fields, because inline comments can cause a modified field to become commented out. For example, if we have this class:

```java
public class PrinterNewlines {
    // if a newline is not inserted on this line, the field below will be commented out
    public static int NON_FINAL_FIELD = 42;
}
```

And insert a modifier on the field (say `final`), this is the output:

```java
public class PrinterNewlines {
    // if a newline is not inserted on this line, the field below will be commented outpublic static final int NON_FINAL_FIELD = 42;
}
```

I'm working on figuring out why this happens.

slarse:
Fix #3697 

~This is heavily WIP, I don't have a fix yet. The problems are thoroughly described in #3697~

This PR introduces a fix for the problem of disappearing whitespace between comments and type members with modified modifier lists. For example, given the following commented type member:

```java
// this is a type member
private static int number = 2;
```

Adding a `final` modifier to the modifier list would result in the following print-out:

```java
// this is a type memberprivate static int number = 2;
```

The source fragments will look something like so:

```
"// this is a type member"
["\n\t", "private", "static"]
...
```
Where the comment is a normal `ElementSourceFragment` and the list is a `CollectionSourceFragment`. The reason the whitespace isn't printed is that the it becomes part of the `CollectionSourceFragment`. The problem with this is that the modifier list is only entered when the first modifier is encountered, in this case `private`, and then the preceeding whitespace is printed. But as the `CollectionSourceFragment` hasn't been entered, the only preceeding whitespace is the comment (which is treated as whitespace).

The fix this PR does is to make the whitespace its own separate source fragment, like so.

```
"// this is a type member"
"\n\t"
["private", "static"]
```

This fix is only applied when the whitespace preceding a `CollectionSourceFragment` is the suffix whitespace of a comment. This is the only case where I've found it to be necessary.

Ps. This explanation may seem overly long, but I'm documenting this mostly for myself if this causes a bug tomorrow, and I've forgotten all about it.

### DIFF
--- a/src/main/java/spoon/support/sniper/internal/AbstractSourceFragmentPrinter.java
+++ b/src/main/java/spoon/support/sniper/internal/AbstractSourceFragmentPrinter.java
@@ -197,7 +197,7 @@ abstract class AbstractSourceFragmentPrinter implements SourceFragmentPrinter {
 
 						if (childAtIdxIsModifiedCollectionSourceFragment(i + 1)) {
 							// This is typically when the next child is a modified modifier list.
-						    // We must forcibly print a linefeed in this case, as the linefeed
+							// We must forcibly print a linefeed in this case, as the linefeed
 							// is part of the collection fragment and will thus not be printed with
 							// the comment, nor with the collection fragment itself due to corner
 							// case of entering the collection fragment "too late".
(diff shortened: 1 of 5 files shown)


slarse:
I've been working on this for half a workday now. I'm not sure I can fix this reasonably, there are simply too many corner cases with the way the sniper printer currently works.

slarse:
My new attempts at fixing this seem to have worked. It's really two entirely different problems. I'll write up an explanation tomorrow.

Ps. Not ready to merge, need to tidy up.

slarse:
This is ready for review, see the initial comment for an explanation.

slarse:
There is an issue related to this where removing the final element in a modifier will cause any preceding comment to disappear, but it's a separate implementation issue and so I'm treating it separately (see #3732).

monperrus:
LGTM. Will merge. Thanks!

monperrus:
The sniper mode is ever better! Thanks @slarse

PR #3702, opened 2020-11-17
fix: Fix Sniper printer failing to put added import statement on separate line

### LINKED ISSUE #3698: bug: Sniper fails to add newline when adding new import statement
For example, with this file:

```java
import java.util.Set;

public class Cls {}
```
adding an import statement will result in a pretty-print that looks something like this:

```java
import java.util.ArrayList;java.util.Set;

public class Cls {}
```

slarse:
Fix #3698 

This is ready for review, see comments further down :)

### DIFF
--- a/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
+++ b/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
@@ -1084,14 +1084,18 @@ public class DefaultJavaPrettyPrinter implements CtVisitor, PrettyPrinter {
 				//note: the package-info.java may contain type declarations too
 			break;
 		case TYPE_DECLARATION:
-				scan(compilationUnit.getPackageDeclaration());
-				for (CtImport imprt : getImports(compilationUnit)) {
-					scan(imprt);
-					printer.writeln();
-				}
-				for (CtType<?> t : compilationUnit.getDeclaredTypes()) {
-					scan(t);
-				}
+			scan(compilationUnit.getPackageDeclaration());
+			if (!compilationUnit.getDeclaredPackage().isUnnamedPackage()) {
+				printer.writeln();
+			}
+
+			for (CtImport imprt : getImports(compilationUnit)) {
+				scan(imprt);
+				printer.writeln();
+			}
+			for (CtType<?> t : compilationUnit.getDeclaredTypes()) {
+				scan(t);
+			}
 			break;
 		default:
 				throw new SpoonException("Unexpected compilation unit type: " + compilationUnit.getUnitType());
@@ -1111,7 +1115,7 @@ public class DefaultJavaPrettyPrinter implements CtVisitor, PrettyPrinter {
 		CtPackageReference ctPackage = packageDeclaration.getReference();
 		elementPrinterHelper.writeComment(ctPackage, CommentOffset.BEFORE);
 		if (!ctPackage.isUnnamedPackage()) {
-			elementPrinterHelper.writePackageLine(ctPackage.getQualifiedName());
+			elementPrinterHelper.writePackageStatement(ctPackage.getQualifiedName());
 		}
 	}
 
--- a/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
+++ b/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
@@ -1085,7 +1085,9 @@ public class DefaultJavaPrettyPrinter implements CtVisitor, PrettyPrinter {
 			break;
 		case TYPE_DECLARATION:
 			scan(compilationUnit.getPackageDeclaration());
-			if (!compilationUnit.getDeclaredPackage().isUnnamedPackage()) {
+
+			CtPackage pkg = compilationUnit.getDeclaredPackage();
+			if (pkg != null && !pkg.isUnnamedPackage()) {
 				printer.writeln();
 			}
 
--- a/src/main/java/spoon/reflect/visitor/ElementPrinterHelper.java
+++ b/src/main/java/spoon/reflect/visitor/ElementPrinterHelper.java
@@ -348,9 +348,20 @@ public class ElementPrinterHelper {
 		}
 	}
 
+	/**
+	 * Write a package statement and a newline.
+	 */
 	public void writePackageLine(String packageQualifiedName) {
+		writePackageStatement(packageQualifiedName);
+		printer.writeln();
+	}
+
+	/**
+	 * Write a package statement.
+	 */
+	public void writePackageStatement(String packageQualifiedName) {
 		printer.writeKeyword("package").writeSpace();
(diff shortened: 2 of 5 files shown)


src/test/java/spoon/test/prettyprinter/TestSniperPrinter.java
```diff
@@ -398,16 +418,7 @@ private void testSniper(String testClass, Consumer<CtType<?>> transformation, Bi
...
-			SniperJavaPrettyPrinter printer = new SniperJavaPrettyPrinter(launcher.getEnvironment());
-			printer.setPreprocessors(Collections.unmodifiableList(Arrays.<Processor<CtElement>>asList(
-					//remove unused imports first. Do not add new imports at time when conflicts are not resolved
-					new ImportCleaner().setCanAddImports(false),
-					//solve conflicts, the current imports are relevant too
-					new ImportConflictDetector(),
-					//compute final imports
-					new ImportCleaner()
-			)));
-			return printer;
```
slarse, line 410: Removed as it interfered with my test cases (removed unused imports), and no other test cases actually rely on this functionality. Very unclear why it's here.

monperrus:
LGTM, unwip?

slarse:
@monperrus I thought this was not good enough a fix, and I was right: it doesn't fix the problem if you add an import statement to a project that has a package statement. The problem is an off-by-one error that seemingly only comes into play when adding stuff (any stuff) at the very top of the file.

Back to the drawing board, I need to actually fix the off-by-one error :)

slarse:
New test case [shows the newline fail here](https://travis-ci.com/github/INRIA/spoon/jobs/442522194#L1255-L1273)

slarse:
@monperrus I've _finally_ figured this out. The missing newline for an import following a package statement, and the missing newline between import statements, are really two separate problems. Anyway, here's the rundown.

### Two import statements ending up on the same line
This problem is occurs when adding a new import statement, which is printed first in the file, and being immediately followed by a pre-existing import statement. As the pre-existing import statement has index 0 in the fragment child list, any newline events created by the default printer will never be printed.

The fix here is to always print newline events when printing the first pre-existing import statement. I _think_ this might affect other child lists than import statements, but we can't unconditionally print newline events at the start of every child list as the default printer sometimes generates unwanted events (e.g. in if/else statements) that we definitely don't want. I think we will simply have to keep adding exceptions to this particular case as they pop up.

### A newly added import statement ending up on the same line as the package declaration
This occurs when adding a new import statement that is printed after the package declaration. Previously, the `visitCtPackageDeclaration` printed its own newline. This is a big no-no for compatibility with the sniper printer, as if it decides that an element is unaltered, it will print its original source code _without_ the newline. Therefore, in general, no single `CtElement`'s visit method must include a newline for the visited element; it must be a completely standalone printer event. Of course, composite elements may still insert newlines between their constituent elements, but it should not terminate with a newline, or stuff like this happens. I think we will see this problem crop up again.

The fix was simply to not let `visitCtPackageDeclaration()` print a terminating newline, and instead print it in `visitCtCompilationUnit()`. Something similar may be the problem in #3697.

Also see some additional comments in the code.

**Let me know if you want me to split this PR in two**, it's technically two independent fixes with similar symptoms.

src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
```diff
@@ -1084,14 +1084,20 @@ public void visitCtCompilationUnit(CtCompilationUnit compilationUnit) {
...
-				}
-				for (CtType<?> t : compilationUnit.getDeclaredTypes()) {
-					scan(t);
-				}
+			scan(compilationUnit.getPackageDeclaration());
+
+			CtPackage pkg = compilationUnit.getDeclaredPackage();
+			if (pkg != null && !pkg.isUnnamedPackage()) {
+				printer.writeln();
+			}
```
slarse, line 1092: This is the only difference here, the rest is just indentation. In combination with the fix in `visitCtPackageDeclaration`, this fixes newlines after package declarations in the sniper printer.

src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
```diff
@@ -1111,7 +1117,7 @@ public void visitCtPackageDeclaration(CtPackageDeclaration packageDeclaration) {
 		CtPackageReference ctPackage = packageDeclaration.getReference();
 		elementPrinterHelper.writeComment(ctPackage, CommentOffset.BEFORE);
 		if (!ctPackage.isUnnamedPackage()) {
-			elementPrinterHelper.writePackageLine(ctPackage.getQualifiedName());
+			elementPrinterHelper.writePackageStatement(ctPackage.getQualifiedName());
```
slarse, line 1120: This is the "fix in visitCtPackageDeclaration" I mention.

src/main/java/spoon/support/sniper/internal/AbstractSourceFragmentPrinter.java
```diff
@@ -102,6 +102,11 @@ public int update(PrinterEvent event) {
 			//so skip printing of this comment
 			//comment will be printed at place where it belongs to - together with spaces
 			return -1;
+		} else if (event.getRole() == CtRole.DECLARED_IMPORT && fragmentIndex == 0) {
+			// this is the first pre-existing import statement, and so we must print all newline
+			// events unconditionally to avoid placing it on the same line as a newly added import
+			// statement. See PR #3702 for details
+			printStandardSpaces();
```
slarse, line 109: This is the special handling of the first import statement. We may need to add more roles here as more problems crop up. Using only the condition `fragmentIndex == 0` causes unwanted newlines in if/else statements, and perhaps elsewhere as well.

src/main/java/spoon/reflect/visitor/ElementPrinterHelper.java
```diff
@@ -348,9 +348,20 @@ public void writeImports(Collection<CtImport> imports) {
...
+		printer.writeln();
+	}
+
+	/**
+	 * Write a package statement.
+	 */
+	public void writePackageStatement(String packageQualifiedName) {
 		printer.writeKeyword("package").writeSpace();
-		writeQualifiedName(packageQualifiedName).writeSeparator(";").writeln();
+		writeQualifiedName(packageQualifiedName).writeSeparator(";");
```
slarse, line 364: Here, I've added the method `writePackageStatement`, as I did not want to break the previous API by modifying `writePackageLine`.

monperrus:
Excellent work and explanations, thanks a lot @slarse

slarse:
You're welcome :). I learned a ton about the sniper printer doing this, so I think I'm ready to get back to tackling #3697 now.

Ps. it looks like merging my PR broke the build on master, but I'm pretty sure that wasn't my fault.

monperrus:
> it looks like merging my PR broke the build on master, but I'm pretty sure that wasn't my fault.

yes, it's not your fault, it's a flaky check due to Github API Rate 
Limiting, we may remove this low value check.

PR #3704, opened 2020-11-17
test: replace jre matching check with JUnit annotation

### LINKED ISSUE #3682: Refactor(Test): Remove jre matching from test cases and replace with junit annotation
### Problem
At spoon we have some test cases which have some weird problems at jdk8 and are run on jdk9+ only.
Example: [testRemoveDeprecatedMethods](
https://github.com/INRIA/spoon/blob/f6a2849bb9d3de9a992055fbbb67a0a27e7dc874/src/test/java/spoon/test/refactoring/RefactoringTest.java#L152)
Junit got recently and annotation for it. We could replace the self written code with junit.
see https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/condition/DisabledForJreRange.html
### Tasks
Tasks are:
- Find all testcases using the self written jre matching
- Replace with annotation
- Remove/refactor unused functions afterwards

As this is a starter issue, feel free to give it a try, ask questions or ping me for help if you need some.

### Why refactor this ?

As always self written code is nice, but requires maintenance. If a stable framework provides the same features and we already use it, we can replace our code with their out of the box solution. This reduces the codebase from spoon and improves the maintainability.

Artamm:
(Close #3682)
Some methods in test cases are replaced with annotations. In `VariableTest` class  the annotation is also added on tests that use methods "since jdk_version".


monperrus:
Cool!

Artamm:
I'm facing a problem in `TestModule` class. In this class are used `@Test`  from junit-vintage, which ignore junit5 annotations (including `@DisabledForJreRange`). However, if I change import, the test case[ `testSimpleModuleCanBeBuilt` ](https://github.com/INRIA/spoon/blob/0834d917094513c3003ba0e76fcdbf1d436e43da/src/test/java/spoon/test/module/TestModule.java#L289) fails because instead of 2 modules it gets only 1 (`simple_module_with_code` is not included).
On the one hand,  disabling jre is not important here and can be ignored, but maybe there is a way to fix this problem?

monperrus:
thanks for the update.

>  the test case testSimpleModuleCanBeBuilt fails because instead of 2 modules it gets only 1 (simple_module_with_code is not included).

Seems we can simply change this assertion.

MartinWitt:
Hi,
can't we replace the junit4 `@Test` with junit5 `@Test`? Changing the test seems wrong because there should be 2 modules.
I`m surprised how well the annotation works :D

Artamm:
> LGTM and CI seems green.
Hello. I think CI is green because now in `TestModule` is used junit4 `@Test` annotation, so in the test case there are 2 modules and checks pass (but  `@DisabledForJreRange` is ignored).  If replace with junit5 `@test`, the test case will have 1 module and check will fail, but the annotation will work. 
 If it's okay, I think that changing the assertion can work (like @monperrus said) .

MartinWitt:
> Or removing annotation, since in launcher is set jre version anyway.

If the test is green afterwards perfect solution, but changing the assertion is fine for me as well.

> If replace with junit5 @test, the test case will have 1 module and check will fail, but the annotation will work.

This seems strange but debugging junit5 is not a nice task. I wouldn't prefer not to look into it.

monperrus:
Thanks a lot @Artamm LGTM

@MartinWitt OK to merge?

MartinWitt:
Yes! Works perfect, nice work @Artamm !

monperrus:
Very nice contribution @Artamm !

Issue #3618, opened 2020-09-27
prettyprint() adds static variables as imports after generating code


AmineOzil:
the prettyprint() method always adds static variables as imports on classes where there are calls to these variables and deletes their class name  on it specific calls on code. (see illustration) 
![image](https://user-images.githubusercontent.com/26030917/94367605-21857200-00e0-11eb-9be1-1c5a3c2d8eda.png)
SVGAbstractTranscoder.KEY_HEIGHT & SVGAbstractTranscoder.KEY_WIDTH are static variables

monperrus:
Thanks for the bug report. Would you be able to reproduce the bug in a failing test case?

Issue #3214, opened 2020-01-25
static import not written out by Spoon


shoeoffhead:
I just did a little test reading in and writing out commons-io-2.6-sources.jar with Spoon.

IOUtils defines: public static final int EOF = -1;

This is statically imported by org.apache.commons.io.output.ByteArrayOutputStream and org.apache.commons.io.EndianUtils

Spoon just adds this static import to ByteArrayOutputStream but NOT EndianUtils

monperrus:
Thanks a lot for the bug report.

Could you share a link to the source file you're analyzing?

shoeoffhead:
Just the latest commons-io version: 2.6
https://mvnrepository.com/artifact/commons-io/commons-io/2.6

You can also download it manually: https://repo1.maven.org/maven2/commons-io/commons-io/2.6
The file is called: commons-io-2.6-sources.jar  

I also use the latest Spoon.

monperrus:
Would you be able to isolate the bug in a failing test case in a pull request? That would be super
useful. Thanks!

shoeoffhead:
I did now several other tests read/writing sources from big libraries. I always have issues with static imports not written out.

I will prepare a minimal project for you showing the bug. I am not very familiar with Github ;)

monperrus:
excellent, looking forward to it!

shoeoffhead:
1. extract the attached zip file
2. execute on the cli: gradle run
3. the files are written to folder 'output'. Add 'output' as a source folder to your IDE to see the described error in my first post.

[commons-collections.zip](https://github.com/INRIA/spoon/files/4129883/commons-collections.zip)

shoeoffhead:
Did you have time looking into this?

monperrus:
Not yet. If you could reproduce the bug in a failing test case in a pull request, that would be super useful. Thanks!

shoeoffhead:
Sorry. I am not really familiar with Git. The zip project is super simple though. You can reproduce the problem in seconds.

Issue #3600, opened 2020-09-16
Move CI to travis-ci.com


monperrus:
Travis CI will close travis-ci.org, we need to move to travis-ci.com

> Hi there!
> 
> As you’ll no doubt have noticed, Travis CI has, for some time, been split between travis-ci.com and travis-ci.org. This was initially planned to enable open-source and commercial/private projects to be distinctly separate, but back in May 2018, we announced that open source builders would have the opportunity to build and test their projects under travis-ci.com, meaning your public and private repositories could be managed under one domain with a consistent interface and build environment.
> 
> 
> In an effort to ensure that all of our users - whether you build open-source, public or private repositories - receive regular feature updates, security patches and UX/UI enhancements, we are announcing that travis-ci.org will be officially closed down completely on December 31st, 2020, allowing us to focus all our efforts on bringing new features and fixes to travis-ci.com and all of our awesome users like yourself on the travis-ci.com domain. 
> 
> What this means for you
> 
> The main thing to expect is that your projects will be built and tested with the same speed and robustness as you’ve come to expect from Travis CI - but also with great new features such as support for AWS Graviton2 which is only available on travis-ci.com. Open-source projects will be free as they always have, with lighting fast builds provided by partners like Packet.
> 
> 
> We encourage you to migrate your existing repositories that are currently on travis-ci.org over to travis-ci.com as soon as possible, enabling you to identify any additional changes required to your code or configuration well in advance of the December 31st deadline. We’ve documented everything you need to know in order to migrate your repositories, and of course, you can reach out on the community forum for any additional help.
> 
> 
> If you have any questions or concerns, please feel free to contact our support team who will be happy to talk through any migration queries you may have!
>  
> 
> Happy building on travis-ci.com!

MartinWitt:
[https://docs.travis-ci.com/user/migrate/open-source-repository-migration](https://docs.travis-ci.com/user/migrate/open-source-repository-migration ) Looks like guide for moving. Someone with write permissions need to follow these steps and move the repo.

monperrus:
We now use Travis CI.com https://travis-ci.com/INRIA/spoon/

monperrus:
Travis CI is down, message is "Builds have been temporarily disabled for private and public repositories due to a negative credit balance. Please go to the Plan page to replenish your credit balance."

Seems like the move was useless

monperrus:
Sent the following email to support

> Hi Travis,
> 
[...]

PR #3716, opened 2020-12-02
Fix: fix pretty-printing of imports for static methods

danglotb:
Hello, 

I'm opening this pull-request that contains a test to reproduce the bug reported in #3618 and #3214.

Thank you

### DIFF
--- a/src/main/java/spoon/reflect/visitor/ImportCleaner.java
+++ b/src/main/java/spoon/reflect/visitor/ImportCleaner.java
@@ -183,8 +183,8 @@ public class ImportCleaner extends ImportAnalyzer<ImportCleaner.Context> {
 				//java.lang is always imported implicitly. Ignore it
 				return;
 			}
-			if (Objects.equals(packageQName, packageRef.getQualifiedName()) &&
-					!( ref instanceof CtExecutableReference<?> && ((CtExecutableReference<?>) ref).isStatic())) {
+			if (Objects.equals(packageQName, packageRef.getQualifiedName())
+					&& !(ref instanceof CtExecutableReference<?> && ((CtExecutableReference<?>) ref).isStatic())) {
 				//it is reference to a type of the same package. Do not add it
 				return;
 			}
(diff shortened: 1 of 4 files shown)


monperrus:
Thanks @danglotb ! Are you trying to fix it?

danglotb:
Hi @monperrus, I'll investigate how to. Does the test seem correct? Do you think I can use it as a base to fix the issue?

danglotb:
Hi, @monperrus. Here my attempt of fixing these issues.

src/main/java/spoon/reflect/visitor/ImportCleaner.java
```diff
@@ -85,7 +85,7 @@ protected void handleTargetedExpression(CtTargetedExpression<?, ?> targetedExpre
 			return;
 		}
 
-		if (target != null && target.isImplicit()) {
+		if (target.isImplicit()) {
```
monperrus, line 88: What about keeping `target != null` to avoid crashing with NPE on `target.isImplicit`?

monperrus (commented):
Thanks a lot for the fix. LGTM, see mino comments.

Plus we have a no CI run, see https://github.com/INRIA/spoon/issues/3600

src/main/java/spoon/reflect/visitor/ImportCleaner.java
danglotb: This check is redundant with the check just above([here](https://github.com/INRIA/spoon/blob/65e15161aacd2915b38e6b3ab8cd1d495a2c15be/src/main/java/spoon/reflect/visitor/ImportCleaner.java#L79)):
```java
if (target == null) {
   ...
   return;
}
```
If `target == null`, the condition `target.isImplicit()` will not be hit, isn't?

monperrus:
OK for me. Unwip? Ready for merge?

monperrus:
Thanks a lot @danglotb

PR #3717, opened 2020-12-03
review: feat: Make sniper printer infer indentation style of compilation unit

### LINKED ISSUE #3715: bug: Sniper printer always prints indentation for newly added elements with tabs
This is a big problem for [Sorald](https://github.com/spoonlabs/sorald), as we can't repair projects that don't use spaces for indentation. I've been looking through the sniper printer, and it doesn't appear as if there are any attempts in it to use anything but tabs to print indentation for new elements with the same style of indentation as there is in the source file.

What the sniper should do here is to infer the indentation style from the file. That's easy enough, just look at the indentation of any type members of the top-level type, and pick the best fit of 2/4 spaces or 1 tabs. That should be sufficient for the vast majority of projects and files.

Any input on this before I get to work on solving it? Have I missed something, perhaps?

slarse:
Fix #3715 

This PR introduces indentation style inference to the sniper printer. It's simple in principle: look at the whitespace preceding the top-level type members and make a best guess as to what the indentation style is out of 1, 2 or 4 spaces/tabs. The original set of source code fragments are used to detect indentation.

If for any given compilation unit no top-level elements can be found, indentation detection defaults to tabs.

We will start using this ASAP in Sorald, preferably with the new beta release this Saturday if we can get it merged by then. Therefore, we should be able to detect any obvious issues this might introduce relatively quickly.

Note: This will crash and burn if you try to sniper print a compilation unit that does not have an origin source (e.g. if you add a CU through the Spoon API). That appears to be the case before this addition as well. Just thought I'd make note of it.

### DIFF
--- a/src/main/java/spoon/support/sniper/SniperJavaPrettyPrinter.java
+++ b/src/main/java/spoon/support/sniper/SniperJavaPrettyPrinter.java
@@ -169,7 +169,7 @@ public class SniperJavaPrettyPrinter extends DefaultJavaPrettyPrinter implements
 	}
 
 	private static Pair<Integer, Boolean> detectIndentation(List<ElementSourceFragment> typeFragments) {
-		List<String> spacePrecedingTypeMember = new ArrayList<>();
+		List<String> wsPrecedingTypeMembers = new ArrayList<>();
 
 		for (ElementSourceFragment typeSource : typeFragments) {
 			assert typeSource.getRoleInParent() == CtRole.DECLARED_TYPE;
@@ -181,13 +181,17 @@ public class SniperJavaPrettyPrinter extends DefaultJavaPrettyPrinter implements
 					TokenSourceFragment cur = (TokenSourceFragment) children.get(i);
 					ElementSourceFragment next = (ElementSourceFragment) children.get(i + 1);
 					if (cur.getType() == TokenType.SPACE && next.getRoleInParent() == CtRole.TYPE_MEMBER) {
-						spacePrecedingTypeMember.add(cur.getSourceCode().replace("\n", ""));
+						wsPrecedingTypeMembers.add(cur.getSourceCode().replace("\n", ""));
 					}
 				}
 			}
 		}
 
-		double avgIndent = spacePrecedingTypeMember.stream()
+		return guessIndentationStyle(wsPrecedingTypeMembers);
+	}
+
+	private static Pair<Integer, Boolean> guessIndentationStyle(List<String> wsPrecedingTypeMembers) {
+		double avgIndent = wsPrecedingTypeMembers.stream()
 				.map(String::length)
 				.map(Double::valueOf)
 				.reduce((acc, next) -> (acc + next) / 2).orElse(4d);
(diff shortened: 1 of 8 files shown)


src/main/java/spoon/support/sniper/internal/MutableTokenWriter.java
```diff
@@ -22,8 +22,36 @@
...
+		 */
+		@Override
+		protected void autoWriteTabs() {
+			int setTabulationSize = env.getTabulationSize();
+			env.useTabulations(originSourceUsesTabulations);
+			env.setTabulationSize(originSourceTabulationSize);
+			super.autoWriteTabs();
+			env.setTabulationSize(setTabulationSize);
+			env.useTabulations(true);
+		}
```
slarse, line 54: This class is a hack that I'm not very happy with. I couldn't figure out another way to override the printing of tabs. Suggestions for other ways to solve the problem are welcome.

src/main/java/spoon/support/sniper/internal/MutableTokenWriter.java
```diff
@@ -22,8 +22,38 @@
...
+		protected void autoWriteTabs() {
+			int setTabulationSize = env.getTabulationSize();
+			env.useTabulations(originSourceUsesTabulations);
+			env.setTabulationSize(originSourceTabulationSize);
+
+			super.autoWriteTabs();
+
+			env.setTabulationSize(setTabulationSize);
+			env.useTabulations(true);
+		}
```
slarse, line 56: This is the actual thing that overrides the printing of tabs with whatever indentation style is present in the source file. I'm not completely satisfied with this solution as it seems a bit hacky, but I couldn't figure out a better way to do it.

src/main/java/spoon/support/sniper/SniperJavaPrettyPrinter.java
```diff
@@ -132,6 +134,12 @@ public void calculate(CtCompilationUnit compilationUnit, List<CtType<?>> types)
 
 		//use line separator of origin source file
 		setLineSeparator(detectLineSeparator(compilationUnit.getOriginalSourceCode()));
+
+		// use indentation style of origin source file for new elements
+		Pair<Integer, Boolean> indentationInfo = IndentationDetector.detectIndentation(compilationUnit);
+		mutableTokenWriter.setOriginSourceTabulationSize(indentationInfo.getLeft());
+		mutableTokenWriter.setOriginSourceUsesTabulations(indentationInfo.getRight());
```
slarse, line 141: The token writer is updated with indentation style for each CU to be printed.

slarse:
Ping @nharrand

src/main/java/spoon/support/sniper/internal/IndentationDetector.java
```diff
@@ -0,0 +1,94 @@
...
+ * Utility class for detecting the indentation style used in a compilation unit.
+ */
+public class IndentationDetector {
+
+	private IndentationDetector() {
+	}
+
+	/**
+	 * Detect the indentation style of the given compilation unit as 1, 2 or 4 spaces or tabs.
+	 *
```
nharrand, line 28: Could you add a comment about the strategy that you choosed? (Relying on type members).

nharrand:
LGTM! If you can just add a comment on the description of `detectIndentation(CtCompilationUnit cu)`, I'll merge!

slarse:
@nharrand Done! Now just waiting for CI.

nharrand:
Thanks for the PR @slarse !

PR #3722, opened 2020-12-07
review: fix(sniper): Fix rounding error in indentation detection on single type member

slarse:
This fixes a rather silly rounding error in the indentation detection: if there's a single type member, `usesTabs = 1 / 2 = 0`. Therefore, a single type member will always be regarded as indented with tabs, regardless of actual indentation. It's extra silly because it affected the particular use case that I implemented the indentation detection for in the first place. Now explicitly uses doubles instead for the comparison.

Tiny rounding errors don't really matter here, in the general case there will either be 0 tabs or all tabs. It was just the massive rounding error of integer division of `1 / 2` that was problematic.

### DIFF
--- a/src/main/java/spoon/support/sniper/internal/IndentationDetector.java
+++ b/src/main/java/spoon/support/sniper/internal/IndentationDetector.java
@@ -79,9 +79,9 @@ public class IndentationDetector {
 			indentationSize = 1;
 		}
 
-		boolean usesTabs = wsPrecedingTypeMembers.stream()
+		boolean usesTabs = (double) wsPrecedingTypeMembers.stream()
 				.filter(s -> s.contains("\t"))
-				.count() >= wsPrecedingTypeMembers.size() / 2;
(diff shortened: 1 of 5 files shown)


slarse:
@monperrus Any idea of why CI is not triggering?

monperrus:
Because `Builds have been temporarily disabled for public repositories due to a negative credit balance. Please go to the Plan page to replenish your credit balance or alter your Consume paid credits for OSS setting. ` [source](https://travis-ci.com/github/INRIA/spoon) and #3600 for the cheapest solution to the problem.

slarse:
oops, my resources are missing

slarse:
This is ready for review

slarse:
@monperrus is c2bc881fa78ac16e7f6a22c10bb824704d20fe17 supposed to be there or was it just to trigger CI?

nharrand:
LGTM!

PR #3723, opened 2020-12-08
chore: setup GitHub Actions workflow for continuous integration

slarse:
#3600 

Adds GitHub workflows for tests. The following is adapted from the Travis jobs:

* `mvn test` with JDK 1.8, 11 and 15
* Coverage uploading to coveralls
* The extra checks in `travis-extras.sh`, except for the external CI stuff

The following is _not_ adapted from the Travis jobs:

* Running of the external CI jobs in https://github.com/SpoonLabs/spoon-ci-external, as they depend on Travis
    - This is simply skipped in `travis-extra.sh` as the environment variables don't match for the final if-check
* The `chore/travis/travis-master.sh` script, as it depends on Travis
    - This script isn't executed at all, it appears to post data to `spoon-ci-external`.

In regards to #3691, it would probably be simple to add a Windows build to the workflow as well.

I have not removed the Travis stuff, I'm assuming that we'll use GH Actions as a backup for when Travis becomes unavailable.


slarse:
Ping @monperrus 

Took me a while to get coverage uploading to work, but now all of the most important tests and checks are running on GH Actions.

slarse:
I realized something: the new `gha-extras.sh` is essentially a copy of `travis-extras.sh`, with the final part of it removed. I'm gonna try running with `travis-extras.sh` instead, to reduce duplication. It should work as-is.

slarse:
Sorry, unping Martin, I had missed some things.

slarse:
@monperrus Okay, now it's done!

monperrus:
Thanks a lot LGTM.

monperrus:
Thanks a lot @slarse now we have CI again!

slarse:
Great!

PR #3730, opened 2020-12-15
fix: Avoid import duplication on unresolved imports

### LINKED ISSUE #3729: bug: ImportCleaner duplicates unresolved imports
If there's an unresolved import in a compilation unit, the ImportCleaner duplicates that import. Minimal example:

```java
// this import will be unresolved as the package does not exist
import non.existing.pkg.SomeClass;

public class UnresolvedImport {
        SomeClass instance = new SomeClass();
    }
}
```

`import non.existing.pkg.SomeClass` will become a `CtUnresolvedImport`, as the package doesn't exist. When computing imports, the ImportCleaner creates regular `CtImportImpl`, which always compare unequal to `CtUnresolvedImport`. I'm working on a fix.

slarse:
Fix #3729 

This is ready for review. As explained in #3729, the problem here is that unresolved imports and "regular" imports never compare equal, even if they are textually identical (which is what actually matters here), causing `ImportCleaner` to duplicate them. With this PR, imports computed by the `ImportCleaner` are only added to the CU if they are unequal to any existing imports, both in terms of object equality and textual equality.

### DIFF
--- a/src/main/java/spoon/reflect/visitor/ImportCleaner.java
+++ b/src/main/java/spoon/reflect/visitor/ImportCleaner.java
@@ -38,6 +38,7 @@ import spoon.support.util.ModelList;
 import spoon.support.visitor.ClassTypingContext;
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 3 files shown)


monperrus:
Thanks a lot @slarse

PR #3737, opened 2020-12-21
fix: fix NPE in ReferenceBuilder#tryRecoverTypeArguments

dya-tel:
`ReferenceBuilder#tryRecoverTypeArguments` doesn't check whether `expectedType.typeArguments()` is null and crashes with an NPE. Could be a bug in JDT, but the check nevertheless should be there - [ParameterizedTypeBinding](https://github.com/eclipse/eclipse.jdt.core/blob/master/org.eclipse.jdt.core.tests.model/workspace/Compiler/src/org/eclipse/jdt/internal/compiler/lookup/ParameterizedTypeBinding.java) has a lot of null-checks for this field.


monperrus:
Thanks a lot for your contribution @dya-tel

PR #3738, opened 2020-12-22
fix: equality evaluation of elements in different inner classes

### LINKED ISSUE #3719: CtExecutableReference equal problem
Hi, I'm working with Spoon 8.3.0, and I want to analyze some Java8 code. 
When I tried to figure out whether some invocation is invoking method declared in source code, I found some problem about comparing CtExecutableReference. Here is a snippet to reproduce the issue.

```
        spoon.addInputResource("/home/usr/TestGeneric.java");
        spoon.buildModel();

        // all executable references of methods declared in source code
        HashSet<CtExecutableReference<?>> exeRefSet = new HashSet<>();
        for (CtMethod<?> meth :
                spoon.getModel().getRootPackage()
                .getElements(new TypeFilter<>(CtMethod.class))) {
            // adding method's executable reference 
            exeRefSet.add(meth.getReference());
        }

        for (CtMethod<?> meth :
                spoon.getModel().getRootPackage()
                .getElements(new TypeFilter<>(CtMethod.class))) {

            // all invocations in meth
            for (CtInvocation<?> inv :
                    meth.getElements(new TypeFilter<>(CtInvocation.class))) {
                // whether inv is invoking a method declared in src 
                if (! exeRefSet.contains(inv.getExecutable()))
                    System.out.println("invoking a method not in src:" + inv.getExecutable());
                else
                    System.out.println("invoking a method in src:" + inv.getExecutable());
            }
        }
```
TestGeneric.java
```
public class TestGeneric {
    public static class Cell<T> {
        public T val;
        public T getVal() { return val; }
        public void setVal(T val) { this.val = val;}
    }

    public static class Day {
        public int day;
        public Day() { }
        public Day(int val) {this.day = val;}
    }

    public void bar() {
        Cell<Day> c = new Cell<>();
        Day e = new Day();
        c.setVal(e);
    }

    public void bar2() {
        Cell<Day> c = new Cell<>();
        Day d = c.getVal();
    }

}
```

Run snippet code, I will get output like this.
```
invoking a method in src:setVal(java.lang.Object)
invoking a method not in src:getVal()
``` 

Why the executable reference getVal cannot be identified equal to the getVal reference stored in hashset, but method setVal(java.lang.Object) can?

andrewbwogi:
Closes #3719

### DIFF
--- a/src/main/java/spoon/support/visitor/equals/EqualsVisitor.java
+++ b/src/main/java/spoon/support/visitor/equals/EqualsVisitor.java
@@ -126,15 +126,13 @@ public class EqualsVisitor extends CtBiScannerDefault {
 		} catch (ParentNotInitializedException e) {
 			return false;
 		}
-		if (innerClassParent1 == null || innerClassParent2 == null) {
+		if (innerClassParent1 == null || innerClassParent2 == null
+				|| (innerClassParent1.isTopLevel() && innerClassParent2.isTopLevel())) {
 			return false;
 		}
 		if ((!innerClassParent1.isTopLevel() && innerClassParent2.isTopLevel())
 				|| (innerClassParent1.isTopLevel() && !innerClassParent2.isTopLevel())) {
 			return true;
-		}
-		if (innerClassParent1.isTopLevel() || innerClassParent2.isTopLevel()) {
-			return false;
 		}
 		int lastDot1 = innerClassParent1.getQualifiedName().lastIndexOf('.');
 		int lastDot2 = innerClassParent2.getQualifiedName().lastIndexOf('.');
--- a/src/main/java/spoon/support/visitor/equals/EqualsVisitor.java
+++ b/src/main/java/spoon/support/visitor/equals/EqualsVisitor.java
@@ -8,9 +8,13 @@
 package spoon.support.visitor.equals;
 
 
+import spoon.reflect.declaration.CtClass;
 import spoon.reflect.declaration.CtElement;
+import spoon.reflect.declaration.ParentNotInitializedException;
 import spoon.reflect.path.CtRole;
+import spoon.reflect.reference.CtReference;
 import spoon.reflect.visitor.CtBiScannerDefault;
+import spoon.reflect.visitor.filter.TypeFilter;
 
 import java.util.Collection;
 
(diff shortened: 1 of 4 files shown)


src/main/java/spoon/support/visitor/equals/EqualsVisitor.java
```diff
@@ -107,6 +114,35 @@ public void biScan(CtRole role, CtElement element, CtElement other) {
...
+		}
+		if ((!innerClassParent1.isTopLevel() && innerClassParent2.isTopLevel())
+				|| (innerClassParent1.isTopLevel() && !innerClassParent2.isTopLevel())) {
+			return true;
+		}
+		int lastDot1 = innerClassParent1.getQualifiedName().lastIndexOf('.');
+		int lastDot2 = innerClassParent2.getQualifiedName().lastIndexOf('.');
+
+		// remove package name
+		String name1 = innerClassParent1.getQualifiedName().substring(lastDot1 + 1);
```
MartinWitt, line 141: maybe add a comment for the magic +1 addition?

src/main/java/spoon/support/visitor/equals/EqualsVisitor.java
```diff
@@ -96,9 +100,12 @@ public void biScan(CtRole role, CtElement element, CtElement other) {
...
 			return;
 		}
-
 		try {
 			lastRole = role;
+			if (haveDifferentDeclaringInnerClasses(element, other)) {
+				fail(role, element, other);
+				return;
+			}
 			super.biScan(element, other);
```
dya-tel, line 109: Traversing the whole hierarchy up on every element seems a bit redundant and could probably affect performance quite a bit.

andrewbwogi: Thanks, I moved the new check out from the recursive call

src/main/java/spoon/support/visitor/equals/EqualsVisitor.java
andrewbwogi: I expanded the comment to make the +1 more clear:
```java
// remove package name (everything prior to and including the last dot)
```

monperrus:
Hi @andrewbwogi Thanks for the PR.

I'm not sure about the behavior. In this case, it's normal that the methods are considered equals. The equality contract is never some information from the parent (or from any ancestor).

WDYT?

andrewbwogi:
You are right @monperrus because equality here is interpreted in relation to syntactic structure and not to the meaning of the syntax. That is why context or parent information does not enter into account.

monperrus:
Good, we're on the same wavelength.

PR #3739, opened 2020-12-24
chore(jdt): bump jdt version and fix testcase accordingly

MartinWitt:
fixes #3731 
The fix was already as a todo there. JDT allows only for the latest java version preview features. Updating to version 15 and disabling preview features works for now. As soon as we implement java 15 preview features, we need to reenable them. 
As we discussed somewhere, some testhelpers could be super useful. 
```java
class JDTVersions {
/**
* Returns the latest jdt supported java version.
*/
String getLatestSupportedJavaVersion()

/**
* Checks if a jdt version supports preview features
*/
boolean String isPreviewSupported(String/int version)
}
```
This could remove a lot of hard coded versions from the test code. In many cases a latest version is what we want, but we simply write the latest current version. The supported jdt versions has some kind of api see [https://github.com/eclipse/eclipse.jdt.core/blob/master/org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java#L3144](https://github.com/eclipse/eclipse.jdt.core/blob/master/org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java#L3144) 
###  Location
I have no clue where testhelpers are placed inside the test folder. Any hints?
Currently SpoonTestHelpers is in spoon/test/ but maybe a clear folder inside /test for testhelpers is way more obvious to use?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -43,7 +43,7 @@
     <dependency>
       <groupId>org.eclipse.jdt</groupId>
       <artifactId>org.eclipse.jdt.core</artifactId>
-      <version>3.23.0</version>
+      <version>3.24.0</version>
       <exclusions>
         <exclusion>
           <groupId>org.eclipse.platform</groupId>
--- a/src/test/java/spoon/test/comment/CommentTest.java
+++ b/src/test/java/spoon/test/comment/CommentTest.java
@@ -804,9 +804,8 @@ public class CommentTest {
 		launcher.getEnvironment().setNoClasspath(true);
 		launcher.getEnvironment().setCommentEnabled(true);
 
-		// TODO: Compliance level should be set at 15 and previews disabled
-		launcher.getEnvironment().setComplianceLevel(14);
-		launcher.getEnvironment().setPreviewFeaturesEnabled(true);
+		launcher.getEnvironment().setComplianceLevel(15);
+		// launcher.getEnvironment().setPreviewFeaturesEnabled(true);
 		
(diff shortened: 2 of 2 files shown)


slarse:
Would it be possible to wait with merging this until after the next beta release this coming weekend? Sorald needs a critical fix to the Sniper printer that went in this week (#3744), but we are currently bound to using JDT 3.23.0 until our primary dependency is updated to use 3.24.0, which is probably not going to happen until next month's release.

monperrus:
Done! Thanks @MartinWitt for the test case adjustment.

slarse:
Oh bummer, now we can't use the next beta build in Sorald :(

monperrus:
Sorry, I misunderstood your post, I read it "would it be possible to merge this"? Do we have a counter measure?

slarse:
It might work anyway, as we exclude the JDT dependency from Spoon due to a signature mismatch. I'll try out the snapshot once it goes live and get back to you :)

slarse:
@monperrus Tested with the snapshot (and knock on wood) it all appears to work fine! As there were no changes in Spoon's production code to accommodate for the new JDT version, I don't expect there to be any issues.

This is something that very likely will happen again, so it's good to know that our current workarounds are effective at dealing with a slight JDT version mismatch between Spoon and Sonar (obviously, API mismatches will lead to disaster).

monperrus:
Good news!

Issue #3734, opened 2020-12-17
bug: ParentNotInitializedException in AbstractTypingContext#adaptType()


dya-tel:
[This line](https://github.com/INRIA/spoon/blob/7abce70c8b5742f5e2f1bc59151871dd1d2050c6/src/main/java/spoon/support/visitor/AbstractTypingContext.java#L49) throws an exception when the `result` is a CtTypeReference without a parent, but with type parameters. And "standard" types from a `TypeFactory` don't have parents.

It's also kinda connected to #3733:
1. Type reference for `java.lang.Object` gets corrupted with type parameters
2. `getBound()` returns the corrupted type reference for `java.lang.Object` [here](https://github.com/INRIA/spoon/blob/7abce70c8b5742f5e2f1bc59151871dd1d2050c6/src/main/java/spoon/support/visitor/MethodTypingContext.java#L244)
3. The corrupted type reference destroys `adaptType()` because it doesn't have a parent as a "standard" type in a `TypeFactory`

I haven't managed to make a minimal reproducible sample and can't share the main code, sorry.

dya-tel:
Actually, looking through the CtElementImpl code I've found a bunch of problems.

The first one, which is related to the issue, is that if an element has `parent == null` then getParent throws an exception because the element's parent wasn't initialized. This seems alright for the most part, but doesn't make sense for elements which do not have a parent - like the "static" type references in the TypeFactory. Those should either be assigned a parent or override the method so it doesn't throw an exception.

Then things get more interesting about the two getParent overloads.

Those overloads throw an exception if current element's parent is null (to match the main no-parameter overload; exception is thrown by the main overload), or return null when no parents match the predicate. But the thing is that one of the overloads doesn't behave like this:
https://github.com/INRIA/spoon/blob/3435facc2b7089119a0ac278a2e008fde7d7c6b7/src/main/java/spoon/support/reflect/declaration/CtElementImpl.java#L384-L392
```java
// yikes
CtTypeReference<?> object = factory.Type().OBJECT;
object.getParent(CtClass.java); // null
object.getParent(new TypeFilter<>(CtClass.java)); // exception
```
The solution: fix the "Class" overload so both overloads handle null-parent the same way as the main method.

The javadoc states that the "Filter" overload can return the same object (which doesn't make sense for me and probably for a lot of other people, it's a get*Parent* method), but it doesn't state the same for the "Class" overload. That's just confusing:
[...]

PR #3741, opened 2020-12-29
fix: ParentNotInitializedException in AbstractTypingContext

dya-tel:
Semi-fix for #3734. Doesn't fix the cause of the issue, but at least fixes invalid element cloning logic. Though I'm not 100% sure whether no-parent elements should be handled or not. Anyway, in case they are supposed to be rejected there should be a separate check.


monperrus:
Thanks for your contribution @dya-tel

PR #3746, opened 2021-01-07
chore: Add a windows build for Java 1.8

### LINKED ISSUE #3691: Discussion(CI): Which jdks for testing?

slarse:
fix #3691 

Trying to add a Windows build.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -38,7 +38,7 @@ jobs:
         run: mv chore/travis/logback.xml src/test/resources/
       - name: Build and test
         env:
-          MAVEN_OPTS: -Djava.src.version=${{ matrix.java }}
+          MAVEN_OPTS: -Djava.src.version ${{ matrix.java }}
         run: mvn test
 
   coverage:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -22,6 +22,8 @@ jobs:
       matrix:
         java: [1.8, 11, 15]
         os: [ubuntu-latest, windows-latest]
+        env:
+          MAVEN_OPTS: -Djava.src.version=${{ matrix.java }}
         exclude:
           - os: windows-latest
             java: 11
@@ -37,7 +39,7 @@ jobs:
       - name: Use silent log config
         run: mv chore/travis/logback.xml src/test/resources/
       - name: Build and test
-        run: mvn -Djava.src.version=${{ matrix.java }} test
+        run: mvn test
 
   coverage:
     runs-on: ubuntu-latest
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -17,10 +17,11 @@ on:
 
 jobs:
   build:
-    runs-on: [ubuntu-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
         java: [1.8, 11, 15]
+        os: [ubuntu-latest, windows-latest]
         exclude:
           - os: windows-latest
             java: 11
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -22,8 +22,6 @@ jobs:
       matrix:
         java: [1.8, 11, 15]
         os: [ubuntu-latest, windows-latest]
-        env:
-          MAVEN_OPTS: -Djava.src.version=${{ matrix.java }}
         exclude:
           - os: windows-latest
             java: 11
@@ -39,6 +37,8 @@ jobs:
       - name: Use silent log config
         run: mv chore/travis/logback.xml src/test/resources/
       - name: Build and test
+        env:
+          MAVEN_OPTS: -Djava.src.version=${{ matrix.java }}
         run: mvn test
 
   coverage:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -30,7 +30,7 @@ jobs:
           - os: windows-latest
             java: 15
 
-    name: Tests with Java ${{ matrix.java }}
+    name: Tests with Java ${{ matrix.java }} on ${{ matrix.os }}
     steps:
       - uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f # v2.3.4
       - uses: actions/setup-java@d202f5dbf7256730fb690ec59f6381650114feb2 # v1.4.3
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -17,10 +17,16 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: [ubuntu-latest, windows-latest]
     strategy:
       matrix:
         java: [1.8, 11, 15]
+        exclude:
+          - os: windows-latest
+            java: 11
+          - os: windows-latest
+            java: 15
+
     name: Tests with Java ${{ matrix.java }}
     steps:
       - uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f # v2.3.4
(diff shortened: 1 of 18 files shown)


slarse:
@MartinWitt Initial windows results:

```
Tests run: 1854, Failures: 53, Errors: 4, Skipped: 18
```

That's not too bad, really.

MartinWitt:
I´m surprised I thought we get way more errors on windows. A solution for the failing tests could be `@DisabledOnOs(WINDOWS)` first and then look step by step into each tests if it could be ported to windows. Fixing 53 testcases can take some time. Is there a timer how much the windows build slows the CI down?
`

slarse:
The last Windows job took 12m36s, while the corresponding Linux job took 11m7s. Total build time was 12m50s for all jobs, so it seems like the Windows job only slowed it all down by a minute or so. Build times will of course vary quite a bit from build to build, and the total time will vary by the amount of available workers, but just given that data, running one Windows job is only slightly slower than running one Linux job.

Overall, I'd say that's a perfectly acceptable performance hit. Note that I'm only running the Windows build for one version of Java, I think that's sufficient.

slarse:
And yes, I think disabling the failing tests is the most feasible next step.

slarse:
@MartinWitt Unfortunately, the `DisabledOnOs` annotation doesn't work for Spoon's test suite as it uses the JUnit4 test annotation `org.junit.Test`. Swapping that out for the JUnit5 annotation `org.junit.jupiter.api.Test` lets the `DisabledOnOs` annotation function, but then there are some incompatibilities with JUnit4 specific features. I'd have to migrate the affected test classes (17 of them in total) to JUnit5 in order to use the annotation.

There's no equivalent JUnit4 annotation, but the following works:

```java
@Test
public void someTest() {
    org.junit.Assume.assumeFalse(System.getProperty("os.name").toLowerCase().contains("windows"));
    // run test
}
```

It's not as explicit, but at least we don't have to contend with having both JUnit4 and JUnit5 test cases in the test suite. What do you think?

MartinWitt:
Look at the testcases we should first stop using \n for newlines and use `System.lineSeparator()`. Error Prone[0] helps a lot with platform depended problems like these. But this is an other topic, we can discuss in a other issue if needed.

For the problem we have 2 solution possibilities:
- Convert 17 tests to junit5
- Create testhelper
 
Converting even 17 test cases from junit 4 to 5 is a big task and if we start converting test cases from junit 4 to 5 we have to ask ourselves why we only convert some. But removing junit 4 is a really big task and isn't the goal of this PR. Mixing junit4 and 5 would even reduce the sometimes poor test quality. Therefore we should step away from it.

Adding a assumption for the tests cases and create a issue for a rewrite in windows friendly style is a much simpler solution oriented approach. A single PR can't fix the whole world. A step by step improvement is fine for me. But I would create a test helper like the following wrapping the assumption and add a TODO/FIXME for each testcase.
```
public static void assumeNotWindows() {
    org.junit.Assume.assumeFalse(System.getProperty("os.name").toLowerCase().contains("windows"));
}
```

[0] https://github.com/google/error-prone

slarse:
> Looking at the testcases we should first stop using \n for newlines and use `System.lineSeparator()`. Error Prone[0] helps a lot with platform depended problems like these. But this is an other topic, we can discuss in a other issue if needed.

That depends on the use case. For example, when testing the sniper printer, it is entirely correct to use a literal `\n` in an assertion if that's what's in the source file. The sniper should use the line separator found in the source file and not the default line sep for the system.

> For the problem we have 2 solution possibilities:
> 
>     * Convert 17 tests to junit5
> 
>     * Create testhelper
> 
> 
> Converting even 17 test cases from junit 4 to 5 is a big task and if we start converting test cases from junit 4 to 5 we have to ask ourselves why we only convert some.

I miscounted, there are 16 affected test classes (but on the order of 100 test cases in them).

>  But removing junit 4 is a really big task and isn't the goal of this PR. Mixing junit4 and 5 would even reduce the sometimes poor test quality. Therefore we should step away from it.

I agree that it's not the goal of this PR, but it's not necessarily that big a task. Most of it can be automated, e.g. IntelliJ IDEA has automatic conversion (it's not bulletproof, but it mostly works). I've done it quite a few times, and it's typically straightforward if you know both frameworks reasonably well. There's actually already some JUnit5 used throughout the tests, I found uses in these classes:

```
src/test/java/spoon/generating/CorrectIdentifierTest.java
src/test/java/spoon/test/factory/TypeFactoryTest.java
src/test/java/spoon/test/model/SwitchCaseTest.java
src/test/java/spoon/test/model/TypeTest.java
src/test/java/spoon/test/module/TestModule.java
src/test/java/spoon/test/position/PositionTest.java
src/test/java/spoon/test/processing/ProcessingTest.java
src/test/java/spoon/test/processing/processors/ParallelProcessorTest.java
src/test/java/spoon/test/refactoring/RefactoringTest.java
src/test/java/spoon/test/reference/ExecutableReferenceTest.java
src/test/java/spoon/test/variable/VariableTest.java
```

So completely migrating to JUnit5 would probably be worthwile, imo. It has many nice features that JUnit4 lacks, such as better support for parameterized tests, awesome temporary directory handling, a certain annotation that I wanted to use here, etc.

> Adding a assumption for the tests cases and create a issue for a rewrite in windows friendly style is a much simpler solution oriented approach. A single PR can't fix the whole world. A step by step improvement is fine for me. But I would create a test helper like the following wrapping the assumption and add a TODO/FIXME for each testcase.
> 
> ```
> public static void assumeNotWindows() {
>     org.junit.Assume.assumeFalse(System.getProperty("os.name").toLowerCase().contains("windows"));
> }
> ```

I agree that this is the best solution for the time being. I'll return to this next week, have a bunch of other open-source stuff to tend to over the weekend.

> Edit: I believe creating a follow up PR fixing all testcases for windows takes a few hours. I could provide the PR if no one wants to /can do it

I don't have a Windows dev environment, so I'm not a great candidate to do it at least.

slarse:
Woop, it worked.

```
Tests run: 1854, Failures: 0, Errors: 0, Skipped: 75
```

slarse:
I'd say this is ready for review. I opened #3752 regarding fixing of Windows-related test failures/errors.

monperrus:
LGTM, thanks Simon!

@MartinWitt WDYT?

MartinWitt:
It`s a really good solution and a big step for gaining the platform independent property back. +1 for merge

slarse:
@MartinWitt The test suite code should be portable to any Windows environment. The build script in `tests.yml` would of course have to be adapted to Travis', and we'd have to work around the fact that [Java does not appear to be an officially supported language in Travis' Windows environment](https://docs.travis-ci.com/user/reference/windows/).

monperrus:
Thanks a lot @slarse

PR #3747, opened 2021-01-08
fix: correctly sniper-print type member comments after removal of all modifiers

### LINKED ISSUE #3732: bug: Sniper printer fails to print comment preceding type member if all modifiers in a modifier list are removed
Hi!

This is related to #3697, but in terms of implementation it's an entirely separate issue. Anyway, when you have a commented type member like so:

```java
// this is a comment
private static int a = 2;
```

And you remove `private` and `static`, the comment also disappears.

```java
int a = 2;
```

This happens because of how the field is split into source fragments, the modifier list becoming one collection of fragments. When the comment is to be printed (which happens when the next concrete element is printed, in this case `int`, as comments are treated like whitespace), the non-printed modifier list "blocks" discovery of the comment when searching for preceding whitespace/comments to print.

I'm working on a fix.

slarse:
Fix #3732 

This PR fixes the bug described in #3732 by keeping track of the last printed element, and allowing the backwards search for previous whitespace to bypass modifier lists that have not been printed. To illustrate the effect of the fix, here's the initial test case file:

```java
public class TypeMemberComments {
    // field comment
    public static int NON_FINAL_FIELD = 42;

    // method comment
    public void nonStaticMethod() {

    }

    // nested type comment
    private class NonStaticInnerClass {

    }
}
```

The test case removes all of the modifiers from all of the type members. Before the fix, the sniper-printed result was the following:

```java
public class TypeMemberComments {
     int NON_FINAL_FIELD = 42;

     void nonStaticMethod() {

    }

     class NonStaticInnerClass {

    }
}
```

Note that the comments have disappeared and that the indentation of the type members is a bit off. After the fix, the result is like so:

```java
public class TypeMemberComments {
    // field comment
    int NON_FINAL_FIELD = 42;

    // method comment
    void nonStaticMethod() {

    }

    // nested type comment
    class NonStaticInnerClass {

    }
}
```

As it should!

### DIFF
--- a/src/main/java/spoon/support/sniper/internal/AbstractSourceFragmentPrinter.java
+++ b/src/main/java/spoon/support/sniper/internal/AbstractSourceFragmentPrinter.java
@@ -10,6 +10,7 @@ package spoon.support.sniper.internal;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
+import java.util.function.Predicate;
 
 import spoon.SpoonException;
 import spoon.reflect.code.CtComment;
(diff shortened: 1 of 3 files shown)


monperrus:
Thanks!

Issue #3752, opened 2021-01-11
Fix test failures and errors on Windows


slarse:
With #3746 we're adding a new build for Java 1.8 on Windows. There are 53 test failures and 4 errors that are specific to Windows that should be fixed. All of these start with the following line:

```java
assumeNotWindows(); // FIXME Make test case pass on Windows
```

monperrus:
@slarse the Windows CI build fails on master: https://github.com/INRIA/spoon/actions/runs/484260233

Would you have a look? Thanks!

slarse:
The failing test was on a branch that didn't have the Windows build, and was merged very close to it. Many of the sniper printer tests don't work well on Windows, it probably has something to do with newlines not being correctly printed with LF line endings (just a hunch).

I'm disabling that test for the time being, and then when someone gets around to fixing Windows test failures they can deal with it. It's very likely that most, perhaps all, of the sniper test failures are caused by the same issue.

slarse:
Just an FYI, I don't intend to be the one to fix these test errors. Anyone with a Windows dev environment would be a better candidate.

@MartinWitt Did you feel like doing it?

MartinWitt:
On windows we have 2 problems, paths and carriage returns. The path problem is fixed by using the java api and not comparing strings. For the carriage return problem exist multiple solutions. Loo
```java
	@Test
	public void testTextBlockCreation(){
		// contract: Test creation of TextBlock and prettyprinting
		Factory factory = getSpoonFactory();
		CtClass<?> c = Launcher.parseClass("class Test{public String m1(){String s = \"\";}}");
		CtBlock<?> body = c.getMethod("m1").getBody();
		CtReturn ret = factory.createReturn();
		ret.setValueByRole(CtRole.EXPRESSION, factory.createTextBlock("Hello, \"World\"!\nTesting\n\tTabs"));
		body.insertEnd(ret);
		assertEquals(
				"class Test {\n"
				+ "    public String m1() {\n"
				+ "        String s = \"\";\n"
				+ "        return \"\"\"\n"
				+ "        Hello, \"World\"!\n"
				+ "        Testing\n"
				+ "        	Tabs\"\"\";\n"
				+ "    }\n"
				+ "}",
				c.toString()
		);
	}
```
If we take a look at this testcase, we have 3 solutions for the common problem with carriage returns.
### Remove "\r"
Adding a `.replaceAll("\r", "")` for the `toString` fixes the testcase, but not the core problem
### Write own assertions
[...]

PR #3753, opened 2021-01-14
test Disable failing test on Windows build

slarse:
#3752 

A branch that was developed in parallel with the Windows build being added had a test that does not pass on Windows. This PR disables that test for Windows. It should be fixed along with the other failing sniper tests, they all likely suffer from the same or similar issues.


monperrus:
Thanks Simon

PR #3755, opened 2021-01-16
refactor: use slf4j API instead of log4j-api

### LINKED ISSUE #3686: refactor: move from Log4j API to SLF4J API
For sake of future maintenance, we need to depend on the SLF4J API (instead of the Log4J API) 

See discussion in #3678

andrewbwogi:
closes #3686


MartinWitt:
Nice work! 2 Points from me:
- How about replacing the Logger initialization with `private static final Logger LOGGER = Logger.getLogger(MethodHandles.lookup().lookupClass());`. If we have a convention for using this, no one starts to use the `Logger.getLogger(String)` with non class names. 
- Replace the star imports with there correct imports?

monperrus:
Really nice, thanks a lot @andrewbwogi LGTM.

Why does CI fail?

MartinWitt:
3 Checkstyle errors

monperrus:
Great, thanks, will merge tomorrow.

monperrus:
Thanks a lot @andrewbwogi, that's very valuable.

seintur:
Hi all,

Since this PR has been merged, it seems that several projects that use Spoon fail with an java.lang.NoClassDefFoundError: org/slf4j/event/Level

For those with access to the Inria CI, see for instance Casper (the same for Juliac although since the log is much larger, the failure is less obvious to spot). Certainly an issue with slf4j then. Any idea or pointer to the solution?

MartinWitt:
By merge there were problems in our CI see https://github.com/INRIA/spoon/runs/1726213633. At first glance it looks like a race condition in ParallelProcessor on windows(argh)

monperrus:
We can probably add a slf4j-something dependency to fix this.

MartinWitt:
Can some1 try to build a failing product with the following steps?
- add https://mvnrepository.com/artifact/org.slf4j/slf4j-simple to spoon
- rebuild spoon
- rebuild $project with the local built spoon?

slarse:
Possibly related failures (fits the time frame of when this was merged):

https://ci.inria.fr/sos/job/juliac/278/console

https://ci.inria.fr/sos/job/Jsoup/1567/console

https://ci.inria.fr/sos/job/Okio/1590/console

monperrus:
`org/slf4j/event/Level` comes from slf4j-api-1.7.30.jar (verified), which is included in this PR.

How come it would it not be found by clients?

That may be due to those jobs being executed with JDK8.

slarse:
> The weirdness is that this not a ClassNotFoundException but a NoClassDefFoundError. AFAIK this means that the class was present at compile-time but no longer at load/run-time.

That's correct. It's not all that strange, really, just means that for some reason the required class is not present on the classpath at runtime.

Now every other build is failing with this error message when running Spoon's Maven plugin:

> Failed to execute goal fr.inria.gforge.spoon:spoon-maven-plugin:3.1:generate (default) on project <PROJECT>: Execution default of goal fr.inria.gforge.spoon:spoon-maven-plugin:3.1:generate failed: A required class was missing while executing fr.inria.gforge.spoon:spoon-maven-plugin:3.1:generate: org/slf4j/event/Level

Going to try to start replicating some errors locally and see if I can figure it out.

slarse:
Can't replicate on my system. I'll try spinning up a few Ubuntu VMs.

PR #3761, opened 2021-01-19
WIP (Feature/printer): allow changing line separator

MartinWitt:
As discussed here #3752, we fix a lot of testcase by allowing to set line separator. This PRs adds a flag to `Environment` to allow easier usage, as the needed method is not exposed see [link](https://github.com/INRIA/spoon/issues/3752#issuecomment-762169344). The enum `PlatformLineSeparator` should allow easier usage with the 3 values, UNIX, WINDOWS, System_Default.  

Is there any need to support different not normal line separators? Could there be any printer/sniper related problems?
Edit: is there even a need for using `System.lineSeparator()` as default. I couldn't find any windows software that needs `\r\n`

### DIFF
--- a/src/main/java/spoon/compiler/Environment.java
+++ b/src/main/java/spoon/compiler/Environment.java
@@ -440,15 +440,15 @@ public interface Environment {
 	void setIgnoreDuplicateDeclarations(boolean ignoreDuplicateDeclarations);
 
 	/**
-	 * Sets the line ending to the given value. See {@link spoon.compiler.Environment.LineEnding} for possible values.
-	 * @param lineEnding  lineEnding to use
+	 * Sets the line separator to the given value. See {@link spoon.compiler.Environment.PlatformLineSeparators} for possible values.
+	 * @param lineSeparator  lineEnding to use
 	 */
-	void setLineEndings(LineEnding lineEnding);
+	void setLineSeparators(PlatformLineSeparators lineSeparator);
 
 	/**
-	 * @return  the used line endings as a String.
+	 * @return  the used line separator as a String.
 	 */
-	String getLineEnding();
+	String getLineSeparators();
 	/** Drives how the model is pretty-printed to disk, or when {@link CtElement#prettyprint()} is called */
 	enum PRETTY_PRINTING_MODE {
 		/** direct in {@link spoon.reflect.visitor.DefaultJavaPrettyPrinter}, no preprocessors are applied to the model before pretty-printing }. */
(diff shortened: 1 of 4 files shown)


monperrus:
Thanks for the first draft @MartinWitt !

* in order to avoid unnecessary bloating in the environment class, it would be better to do this automatically in `PrinterHelper` with platform inference
* this could be merged/unified with the mechanism of `lastCharWasCR` in `PrinterHelper`

WDYT?

slarse:
> it would be better to do this automatically in PrinterHelper with platform inference

This is the way it currently works, and is the entire problem as the output from the printer becomes platform dependent. The fact that there's no easy way to set the printer's line separator is both a PITA in tests that need to pass both Windows and Unix-likes, and a usability problem as a dev using Windows for a project that uses LF line endings will generate code with the wrong code style. The latter would even be a problem for Spoon, as generated code is committed to the repo, and there's no auto-formatter to fix the line endings.

But I agree that bloating the environment with options is a problem. Perhaps we should consider adding a `PrinterOptions` object that collects all the printer options, that can then be set on the environment. Then, we can have however many printer options we want without bloating the environment. For example:

```java
Environment env = ...;
env.setPrinterOptions(PrinterOptions.defaults()
    .setUseAutoImports(true)
    .setLineSeparator(PlatformLineSeparator.UNIX)
    .setAnotherOptionWeDontWantInTheEnvironment("yes")
);
```

We then deprecate all of the setters/getters on the environment that set printer-related options, and have them simply set the corresponding options on the `PrinterOptions` object until they are removed completely.

Just a thought to both allow for @MartinWitt's PR to work out while at the same time debloating the already somewhat bloated environment.

MartinWitt:
So the long term goal is to split the current environment in multiple smaller classes? 
> But I agree that bloating the environment with options is a problem. Perhaps we should consider adding a PrinterOptions object that collects all the printer options, that can then be set on the environment. Then, we can have however many printer options we want without bloating the environment itself. For example:But I agree that bloating the environment with options is a problem. Perhaps we should consider adding a PrinterOptions object that collects all the printer options, that can then be set on the environment. Then, we can have however many printer options we want without bloating the environment itself. For example:

As I´m a big fan of fluent interfaces, i suggest that `PrinterOptions` has the builder pattern you showed and normal getter/setters. Makes adding a new flag a bit more work, but client code is more readable.   

Do you prefer to do this in 1 or 2 PRs? We could split this in two. The first creates the new PrinterOptions, deprecates methods , changes all code on our side, and the second adds the new feature?

slarse:
> So the long term goal is to split the current environment in multiple smaller classes?

At least when we have groups of tightly related settings, such as those for the printer, I think it makes sense. It would also help end-users figure out which settings affect what parts of Spoon's behavior.

> Do you prefer to do this in 1 or 2 PRs?

If we go ahead with it, implementing a `PrinterOptions` wrapper should definitely be its own PR. It's not complicated but it'll probably lead to a pretty big diff.

@monperrus wdyt?

monperrus:
> So the long term goal is to split the current environment in multiple smaller classes?

I would not put it this way, because in Spoon, we aim at maximum backward compatibility. But for new options, with very specific scope for advanced users, yes we put them elsewhere.

monperrus:

Thanks!

PR #3771, opened 2021-01-27
fix: Search for packages in all available Java modules

### LINKED ISSUE #3769: bug: Getting the type declarations from type references doesn't work for types in named modules
I'm having some trouble with modules in Spoon. The most prominent issues are related to links between type references and declarations. For example, for any type in the unnamed module, the following round-trip works fine:

```java
CtType<?> someType = [... any type in the source input ...];
someType.getReference().getTypeDeclaration() == someType;
```

However, if `someType` is declared in a module, then `someType.getReference().getTypeDeclaration()` returns `null`, which is very strange as the type in question is in the source input. A side effect of this is that any compilation unit declared in a module returns a list of null values if you call `CtCompilationUnit.getDeclaredTypes()`.

I'm not super familiar with how modules work, but I have a nagging suspicion that this has to do with the wrong classloader being used in `CtTypeReferenceImpl.findClass()`.

The tests for modules in `TestModule` are also insufficient, either being disabled, asserting incorrect things (such as checking that there's only the unnamed module for a project with a named module), or both. I suspect modules in Spoon need some work overall. I'm getting right on that, as this is a blocker for spoonlabs/sorald#339, and we must be able to support projects with modules.

**If anyone has any insight on classloaders and modules**, feel free to share!

slarse:
Fix #3769 

The issue describes the symptoms and the root cause of them. The tl;dr is that `PackageFactory.get(String)` only searches for packages in the unnamed module, causing type resolution in named modules to not work correctly. This PR fixes that by searching for packages in all modules.

As noted in #3769, there's probably a whole lot of code that assumes that there's only the unnamed module, and therefore use `CtModel.getRootPackage()` (which returns the root package of the unnamed module) as the definitive root package, which as of Java 9 is not appropriate.

### DIFF
--- a/src/main/java/spoon/reflect/factory/PackageFactory.java
+++ b/src/main/java/spoon/reflect/factory/PackageFactory.java
@@ -162,9 +162,15 @@ public class PackageFactory extends SubFactory {
 		return factory.getModel().getAllModules().stream()
 				.map(module -> getPackageFromModule(qualifiedName, module))
 				.filter(Objects::nonNull)
-				.findFirst().orElse(null);
+				.findFirst()
+				.orElse(null);
 	}
 
+	/**
+	 * @param qualifiedName Qualified name of a package.
+	 * @param module A module in which to search for the package.
+	 * @return The package if found in this module, otherwise null.
+	 */
 	private static CtPackage getPackageFromModule(String qualifiedName, CtModule module) {
 		StringTokenizer token = new StringTokenizer(qualifiedName, CtPackage.PACKAGE_SEPARATOR);
 		CtPackage current = module.getRootPackage();
--- a/src/main/java/spoon/reflect/factory/PackageFactory.java
+++ b/src/main/java/spoon/reflect/factory/PackageFactory.java
@@ -12,7 +12,6 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Objects;
-import java.util.Optional;
 import java.util.StringTokenizer;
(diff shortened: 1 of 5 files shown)


slarse:
This is very odd. If I now run all tests locally with IntelliJ IDEAs test runner, I get a test failure on `LogTest.NonParameterizedTest.testMavenLauncherLogs()`.

```
Expected :Running in FULLCLASSPATH mode. Source folders and dependencies are inferred from the pom.xml file (doc: http://spoon.gforge.inria.fr/launcher.html).
Actual   :Running in FULLCLASSPATH mode. Classpath is manually set (doc: http://spoon.gforge.inria.fr/launcher.html).
```

If I run that test individually, it does not fail. If I run all tests with Maven, there are no failures either. Something odd is going on with that test.

slarse:
Ready for review.

monperrus:
This is a serious bug! I'm not that surprised given the relatively low usage of modules in the field.

The fix LGTM, thanks @slarse !

slarse:
> This is a serious bug! I'm not that surprised given the relatively low usage of modules in the field.

Indeed! I'm hoping this solves the most severe issues we're facing in Sorald. But I have a nagging feeling this isn't the last module-related bug we'll see. It's just such a significant metamodel shift from the pre-Java 8 times.

slarse:
@monperrus Think this could be merged today? It's a blocker for handling Java 9+ projects with Sorald.

monperrus:
Very important fix wrt modules, thanks a lot @slarse

PR #3772, opened 2021-01-27
doc: Correct doc about getTypeDeclaration returning null

slarse:
`CtTypeReference.getDeclaration()` says that `CtTypeReference.getTypeDeclaration()` is great because it _never returns null_. That's actually not the case: `getTypeDeclaration()` does return null if the referenced type is neither in the input source nor on the classpath. It even says so in the doc for `getTypeDeclaration()`. For completeness, here's a way to make that happen:

```java
CtTypeReference<?> ref = launcher.getFactory().Type().createReference("SomeType");
assert ref.getTypeDeclaration() == null;
```

This PR simply changes "never" to "less often". The reason I'm changing this is because I ran into some (to me) inexplicable `NullPointerException`s due to a method I thought would never return null :)

### DIFF
--- a/src/main/java/spoon/reflect/reference/CtTypeReference.java
+++ b/src/main/java/spoon/reflect/reference/CtTypeReference.java
@@ -65,7 +65,7 @@ public interface CtTypeReference<T> extends CtReference, CtActualTypeContainer, 
 	 * reference or <code>null</code> if the type declaration is not in the
 	 * analyzed source files,
 	 *
-	 * {@link #getTypeDeclaration()} is a newer and better alternative that never returns null.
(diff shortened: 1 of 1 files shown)


monperrus:
An alternative would be to add a post-condition to the method and throw an exception. But that would likely break some clients, so I'm in favor of this contract weakening.

slarse:
It's not actually a contract weakening, as the documentation for `getTypeDeclaration()` correctly states that it can return `null`. The problem here is that `getDeclaration()` (which is the javadoc I'm editing) misrepresents the contract for `getTypeDeclaration()` in saying that the latter never returns `null`. So I'd say this PR is a documentation bugfix :)

monperrus:
Documentation is essential for success. @slarse your attention to details is greatly appreciated.

Issue #3775, opened 2021-01-29
chore: CI failing because of logging dependencies missing at runtime


slarse:
We currently have a [ton of failures over on Jenkins](https://ci.inria.fr/sos/) due to some sl4j stuff not being available at runtime in the maven plugin. I have not been able to replicate it locally; it all works fine for me.

See also the discussion in #3755 

I'll be trying to figure this out, but if anyone has any ideas that'd be great :)

monperrus:
Indeed the same as the problem reported by @seintur

monperrus:
One fix maybe to change the scope of logback to compile. But that goes against my current understanding of slf4j which is meant to work only with the API.

slarse:
What's very odd is that I can't reproduce it. When I run the CI build script locally, everything works out fine. That makes me think that there isn't anything overtly wrong with the Maven config. I will try reproducing with a few different VMs this Monday.

seintur:
ok don't have a solution yet, but I've done a couple of investigations on project Juliac. In this case, I'm using exec-maven-plugin for invoking Spoon. exec-maven-plugin:3.0.0 depends on slf4j-api 1.7.5, an older version than 1.7.30 that is used by Spoon. Whatever I'm trying to do, exec-maven-plugin constructs an URLClassLoader that contains his version of slf4j-api and not the one that comes with Spoon. And that's the source of the problem because the Event, and certainly others, classes that are used by Spoon and that come from 1.7.30, are absent from 1.7.5. Still have to figure out a way to tell exec-maven-plugin to construct a classpath with the newer version of slf4j-api...

seintur:
follow up from my previous message. So it seems that adding slf4j-api:1.7.30 in the plugin dependencies (not the project) does the trick. The good news is that exec-maven-plugin is moving from 1.7.5 to 1.7.30 in their upcoming 3.0.1 release that is still in snapshot. So whenever 3.0.1 will be released, the trick will certainly no longer be needed.

slarse:
Great! As Juliac was suffering from the same error as pretty much all the other builds (`java.lang.NoClassDefFoundError: org/slf4j/event/Level`), this should be applicable to the other projects as well. I'm still confused as to why none of these errors occur on my own machine.

[...]

PR #3796, opened 2021-02-15
fix: Use self-defined logging level instead of org.slf4j.event.Level

slarse:
#3775 

This PR introduces a self-defined `LogLevel` enum that can be used to get or set the logging level. The reason we need this is that `org.slf4j.event.Level` makes Spoon backwards-incompatible with libraries that uses `slf4j-api < 1.7.15`. Which is not good. I've tested this locally and it fixes the test errors in Casper, as mentioned in #3775.

**This is a breaking change**, as the return type of `Environment.getLevel()` is changed. #I purposefully changed the name of the type from `Level` to `LogLevel` to make this more evident. #3755 was in fact also breaking, as the `org.slf4j.event.Level` is not the same as the previous `Level` type from `log4j`. With this change, however, we won't have breakage again, as we now define the type ourselves.

I have two more thoughts on this.

1. Do we want to implement this breaking change now, or should we wait for Spoon 9? If we don't want to do it now, we must also revert #3755.
2. As we're breaking the API anyway, we may also want to consider renaming `Environment.get/setLevel()` to `Environment.get/setLogLevel()`. These names are more descriptive ("level" is obvious in a logging framework, but not in a metaprogramming library), and also make the breaking change more obvious.

### DIFF
--- a/src/main/java/spoon/Launcher.java
+++ b/src/main/java/spoon/Launcher.java
@@ -35,7 +35,7 @@ import spoon.reflect.visitor.PrettyPrinter;
 import spoon.reflect.visitor.filter.AbstractFilter;
 import spoon.support.DefaultCoreFactory;
 import spoon.support.JavaOutputProcessor;
-import spoon.support.LogLevel;
+import spoon.support.Level;
 import spoon.support.StandardEnvironment;
 import spoon.support.compiler.FileSystemFile;
 import spoon.support.compiler.FileSystemFolder;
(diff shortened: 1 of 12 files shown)


slarse:
@monperrus per our discussion, changed `LogLevel` to `Level`.

monperrus:
Perfect, LGTM, will merge.

monperrus:
Very important follow-up of #3775, thanks very much.

PR #3798, opened 2021-02-17
fix: Prevent ImportCleaner from importing inherited static methods

### LINKED ISSUE #3797: bug: ImportCleaner imports inherited static methods
There's a bug in the ImportCleaner class, causing it to import static methods even though they are inherited. Minimal example:

```java
// base class with static method
public class Base {
    public static int getMeaning() {
        return 42;
    }
}
```

```java
// derived class using inherited static method
public class Derived extends Base {
        System.out.println(getMeaning());
    }
}
```

Running this through the import cleaner will cause it to add `import static Base.getMeaning` to the compilation unit of `Derived`. I'm working on a fix.

slarse:
Fix #3797 

This PR prevents the ImportCleaner from importing static methods that are inherited from a parent class. Such imports cause compile errors if the static method is defined in a class in the default package. At best, they do absolutely nothing.

### DIFF
--- a/src/main/java/spoon/reflect/visitor/ImportCleaner.java
+++ b/src/main/java/spoon/reflect/visitor/ImportCleaner.java
@@ -171,6 +171,7 @@ public class ImportCleaner extends ImportAnalyzer<ImportCleaner.Context> {
 				// we would like to add an import, but we don't know to where
 				return;
 			}
+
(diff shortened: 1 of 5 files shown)


monperrus:
LGTM, thanks!

slarse:
Tested the snapshot on the previously problematic repair case, and it worked! I'll have a new PR for Verificatum on Monday.

PR #3800, opened 2021-02-18
fix: Fix unqualified type reference in catch block becoming qualified

### LINKED ISSUE #3799: bug: Package of unqualified type name in catch not marked as implicit
Given a try/catch like so:

```java
try {
} catch (RuntimeException e) {
}
```
The package of the exception type is not marked as implicit. A side-effect of this is that the type name is printed fully qualified during pretty-printing, like so:

```java
try {
} catch (java.lang.RuntimeException e) {
}
```

This is related to #3268, #3372, #3373

slarse:
Fix #3799 

Fixes the problem by building the type reference with a higher-level method in the JDTTreeHelper, where this particular problem is already addressed. I had to adapt the signature of `JDTTreeBuilderHelper.createCatchVariable()` to also accept a scope for this, but I think this is more correct than it was previously.


monperrus:
LGTM, the fix is surprisingly simple and regular. Will merge.

monperrus:
Thanks @slarse

PR #3801, opened 2021-02-22
doc: fix command line to launch visualization

treaz:
with the list of the current modules I get:
java.lang.module.FindException: Module annotations not found
By updating to --add-modules=ALL-MODULE-PATH, the error is gone and the UI start correctly


monperrus:
@treaz thanks a lot, we love documentation PRs, will merge

Issue #3811, opened 2021-02-26
Sniper printer code formatting problems


woutersmeenk:
I am testing Spoon on our project and there were a number of formatting changes (some with functional impact). I decided too check what happened when rewriting all elements in all files of our projects.

I used the following processor: 
``` java
new AbstractProcessor<CtElement>() {
	public void process(CtElement element) {
		// Do a no-op change, this will force the sniper printer to update the source
		SourcePosition pos = element.getPosition();
		element.setPosition(SourcePosition.NOPOSITION);
		element.setPosition(pos);
	};
}
```

I checked the most important problems and made testcases for them.

The kind of problems:
- Brackets being removed/inserted
- Exception class names being fully qualified
- Comments in conditional statements being incorrectly handled
- Semicolon at end of enums being added
- Comments in enum arguments (causes crash)
- Comment formatting being changed

I suspect there are more issues, but these are the most frequent ones.

slarse:
Hi @woutersmeenk,

Thank you for the bug report and test cases!

What you're presenting here is a worst-case scenario for the sniper printer. In general, the more elements that have been modified, the worse it will perform. When the sniper detects that an element has been modified, it will print it with the default printer, which has its own notion of how to print elements (explaining the addition/removal of braces, semicolons etc). As you've modified every element that has a source position, almost all elements are printed with the default printer. There are only a few things that the sniper can do here, mostly being related to preserving whitespace, but sometimes that backfires.

The one big problem I see in your provided test cases are the inline comments being moved such that they comment out code that shouldn't be commented out. That's bad, and I'll see what I can do about it. Formatting changes are however unavoidable when elements have been modified, as the default printer must take over.

> * Exception class names being fully qualified

This is a bug that's most likely not in the printer. It's probably a failure of Spoon to mark the reference as simply qualified.

> * Comments in enum arguments (causes crash)

I've seen the error before, it's caused by the sniper incorrectly resolving the source code bounds of an element when considering its child elements.

In most cases where types have been edited to a moderate extent, the current sniper printer works very well. Due to its design, it's not well suited to print types that have been extensively modified, especially not when they have been _completely_ modified, as in this case. We're working on a redesign of the sniper (see #3794) that should make it more reliable, but it's still on a conceptual level. It's however likely that we'll focus our efforts on the redesign rather than fixing all bugs in the existing printer. Your test cases present nice benchmarks for implementing the new design.

[...]

PR #3812, opened 2021-02-26
test: Add test for sniper printer issues (#3811)

woutersmeenk:
Tests for issue #3811

### DIFF
--- a/src/test/java/spoon/test/prettyprinter/SniperNoChangeDiffTest.java
+++ b/src/test/java/spoon/test/prettyprinter/SniperNoChangeDiffTest.java
@@ -1,0 +1,81 @@
+package spoon.test.prettyprinter;
+
+import static org.junit.Assert.assertTrue;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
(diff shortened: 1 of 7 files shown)


monperrus:
Added `@Disabled("UnresolvedBug")` to proceed with merge.

slarse:
@monperrus The test is implemented with JUnit4, so a JUnit5 annotation has no effect.

As we're moving away from Junit4 (see #3919), it would be best if we don't merge new JUnit4 test classes, especially parameterized ones (they are vastly different in 4 and 5). This would be best incorporated as a single parameterized test case in the (JUnit5) `TestSniperPrinter` test class.

@woutersmeenk Sorry about the long silence here, we had a tad bit too many stale PRs open and lost track of this one. Do you by any chance have time to incorporate this as a test case in the aforementioned test class? It's understandable if you don't feel like traveling half a year back in time, and I could take over and perform the conversion if you'd like.

woutersmeenk:
@slarse I refactored the test to use JUnit 5 Parameterized tests and added it to `TestSniperPrinter`

slarse (approved):
Thanks @woutersmeenk, looks good to me!

Issue #3806, opened 2021-02-24
bug/chore: ParallelProcessor tests are flaky


slarse:
> By merge there were problems in our CI see https://github.com/INRIA/spoon/runs/1726213633. At first glance it looks like a race condition in ParallelProcessor on windows(argh)

_Originally posted by @MartinWitt in https://github.com/INRIA/spoon/issues/3755#issuecomment-763401113_

This now just happened again, but in a different test: https://github.com/INRIA/spoon/runs/1969383888#step:5:12462

Perhaps we should disable this test on Windows until someone can dedicate some time to the Windows compatibility issues.

slarse:
To be clear, we've had flaky failures in `compareWithSingleThreaded1` and `compareWithSingleThreaded3`. I'd suggest disabling all `compareWithSingleThreadedX` tests until someone can deal with this.

slarse:
Actually, `compareWithSingleThreaded2` probably does not need to be disabled as it runs on a single thread.

slarse:
Okay, now there was a failure on Ubuntu: https://github.com/INRIA/spoon/runs/1969811221?check_suite_focus=true#step:5:15517

So these tests are just flaky.

slarse:
I'll try to have a look at what's causing the test failures tomorrow. I've no idea of what the parallel processing architecture looks like, so I may or may not be able to debug this with the little time I can spare to it.

slarse:
Step one here is to improve the assertions in the tests to let us actually see what the "wrong" output is, the next time it happens.

MartinWitt:
I hoped my documentation about what I was testing was enough. See https://github.com/INRIA/spoon/blob/eac3535a6ffaa30db1beb829f7363a6f49f50e79/src/test/java/spoon/test/processing/processors/ParallelProcessorTest.java#L80
More or less I test If the parallelprocessor visits the some number of elements as a normal processor.

slarse:
@MartinWitt you misunderstand me, I didn't mean I don't understand _what's_ being tested, I meant that the assertion just says "nope". I want to see what the numbers are :)

slarse:
I.e. this is the failure info:

```
java.lang.AssertionError
	at spoon.test.processing.processors.ParallelProcessorTest.compareWithSingleThreaded1(ParallelProcessorTest.java:87)
```

[...]

PR #3813, opened 2021-02-26
test: improve the assertions in ParallelProcessorTest

slarse:
#3806 

This PR just adds assertions that produce more informative output than the current ones for ParallelProcessorTest. For example, this is the current output from an `assertTrue` in `compareWithSingleThreaded3`:

```
Error:  spoon.test.processing.processors.ParallelProcessorTest.compareWithSingleThreaded3  Time elapsed: 0.171 s  <<< FAILURE!
java.lang.AssertionError
	at spoon.test.processing.processors.ParallelProcessorTest.compareWithSingleThreaded3(ParallelProcessorTest.java:166)
```

Whereas with this PR, we get something like this instead (note that I faked the numbers here to get an error!):

```
[ERROR] spoon.test.processing.processors.ParallelProcessorTest.compareWithSingleThreaded3  Time elapsed: 0.056 s  <<< FAILURE!
java.lang.AssertionError: 

Expected: <180>
     but: was <181>
        at spoon.test.processing.processors.ParallelProcessorTest.compareWithSingleThreaded3(ParallelProcessorTest.java:164)
```

Oh, and "expected" is the single-threaded result.

### DIFF
--- a/src/test/java/spoon/test/processing/processors/ParallelProcessorTest.java
+++ b/src/test/java/spoon/test/processing/processors/ParallelProcessorTest.java
@@ -7,6 +7,8 @@
  */
 package spoon.test.processing.processors;
 
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.hamcrest.MatcherAssert.assertThat;
 import static org.junit.Assert.assertTrue;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 
@@ -15,6 +17,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReferenceArray;
+import java.util.stream.IntStream;
 
 import org.junit.Rule;
 import org.junit.Test;
@@ -77,14 +80,9 @@ public class ParallelProcessorTest {
 			}
 		}).noClasspath(true).outputDirectory(folderFactory.newFolder()).buildModel();
 
-		// after processing both |singleThreadCounter| == sum(|atomicCounter|) must be
-		// true.
-		// for checking this subtract each array value from the
-		// singleThreadCounter and check for == 0
-		for (int j = 0; j < atomicCounter.length(); j++) {
-			singleThreadCounter.set(singleThreadCounter.get() - atomicCounter.get(j));
-		}
-		assertTrue(singleThreadCounter.get() == 0);
+		// after processing both |singleThreadCounter| == sum(|atomicCounter|) must be true.
+		int parallelCount = IntStream.range(0, atomicCounter.length()).map(atomicCounter::get).sum();
+		assertThat(parallelCount, equalTo(singleThreadCounter.get()));
 	}
 
 	@Test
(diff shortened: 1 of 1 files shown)


slarse:
And of course it doesn't fail now :). I'll just trigger CI a few more times to see if we can get some useful data.

slarse:
@MartinWitt care to review this PR to verify that the assertions match your intentions?

src/test/java/spoon/test/processing/processors/ParallelProcessorTest.java
```diff
@@ -160,12 +158,12 @@ public void process(CtElement element) {
...
 		}).noClasspath(true).outputDirectory(folderFactory.newFolder()).buildModel();
-		for (int j = 0; j < atomicCounter.length(); j++) {
-			singleThreadCounter.set(singleThreadCounter.get() - atomicCounter.get(j));
-		}
-		assertTrue(singleThreadCounter.get() == 0);
+
+
+		int parallelCount = IntStream.range(0, atomicCounter.length()).map(atomicCounter::get).sum();
+		assertThat(parallelCount, equalTo(singleThreadCounter.get()));
 		// because only 3 are used
```
MartinWitt, line 165: maybe add a "the 4. processor must be unused" here?

src/test/java/spoon/test/processing/processors/ParallelProcessorTest.java
```diff
@@ -160,12 +158,12 @@ public void process(CtElement element) {
 				singleThreadCounter.incrementAndGet();
 			}
 		}).noClasspath(true).outputDirectory(folderFactory.newFolder()).buildModel();
-		for (int j = 0; j < atomicCounter.length(); j++) {
-			singleThreadCounter.set(singleThreadCounter.get() - atomicCounter.get(j));
-		}
-		assertTrue(singleThreadCounter.get() == 0);
+
+
```
MartinWitt, line 162: Add the following as comment so noone needs to read other tests to understand this?
// after processing both |singleThreadCounter| == sum(|atomicCounter|) must be true.

MartinWitt (approved):
2 small additions otherwise way better assertions then mine.

src/test/java/spoon/test/processing/processors/ParallelProcessorTest.java
slarse: I prefer to improve the code rather than put inline comments. I think it should be very clear from the contract of the test what the assertions are checking now, see the updated code. Wdyt?

src/test/java/spoon/test/processing/processors/ParallelProcessorTest.java
slarse: Same as above.

src/test/java/spoon/test/processing/processors/ParallelProcessorTest.java
MartinWitt: looks good, the comment is no longer needed.

monperrus:
Thanks a lot @slarse for the valuable infrastructure work and @MartinWitt for the review.

PR #3815, opened 2021-02-28
perf: move code outside loop in FileCompilerConfig.initializeCompiler

rmkellogg:
Moved invariant variable outside of loop.  Will hopefully give performance boost on projects with lots of source files.


monperrus:
Thanks a lot for your contribution @rmkellogg

PR #3818, opened 2021-03-03
fix: custom URLClassLoaders are now actually used

### LINKED ISSUE #3789: Discussion : Prevent spoon from accessing the JVM class path
Is there any way, i can prevent spoon parser from accessing the JVM classpath and provide necessary jars (rt.jar etc) in launcher.setSouceClassPath() ?

andrewbwogi:
closes #3789


monperrus:
Thanks a lot @andrewbwogi, could you add a test case reproducing the bug?

monperrus:
Thanks @andrewbwogi  for the test case, we love test cases in Spoon.

Any idea why src/test/resources/classloader-test/spoon/Launcher.class is an empty file? And why it works this way?

andrewbwogi:
This Launcher has the same qualified name as the ordinary Spoon Launcher. In this test, if the default classloader is used, the ordinary Launcher will be loaded, and if the custom classloader is used, the empty Launcher will be loaded. It is empty so that the two Launchers can be distinguished.

slarse:
Just to clarify the confusion here, there seems to be a bug in the "Files changed" tab here on GitHub, which claims that `Launcher.class` is an empty file. If you actually view the file, it's 196 bytes, and I checked it just to be sure, it contains the following:

```
public class spoon.Launcher {
  public spoon.Launcher();
    Code:
       0: aload_0
       1: invokespecial #1                  // Method java/lang/Object."<init>":()V
       4: return
}
```

monperrus:
very nice fix @andrewbwogi

Issue #3809, opened 2021-02-24
extraneous parentheses at pretty-printing


monperrus:
We all know that Spoon is generous at putting parentheses, in particular for binary operators and cast expressions.

One way to solve this is to use the implicit information from CTElement to signal parenthesis to be written:

CtBinaryOperator.isImplicit() == true => no parentheses required
CtBinaryOperator.isImplicit() == false => parentheses required

See also the discussion about sniper at https://github.com/INRIA/spoon/issues/3794

slarse:
Interesting idea. It might prove a bit troublesome, though, as whether or not parentheses need to be printed can change when the AST is modified. For example, consider expression:

```java
1 + 2
```

So, if we write binary operators `BinOp(<OPERATOR>, <LEFT_HAND_OPERAND>, <RIGHT_HAND_OPERAND>, <IMPLICIT>) we have an expression like so:

```
BinOp(+, 1, 2, true)
```

However, now let's assume we modify this expression by making it the right-hand operand of a multiplication with `8`, like so:

```
Binop(*, 8, BinOp(+, 1, 2, true), true)
```

If we actually honor the implicitness, then this would be written out as `8 * 1 + 2`, which is actually not what the AST represents. It _must_ be `8 * (1 + 2)`. So, in making arbitrary modifications to expressions, we must also adjust the implicitness of operators. This might be difficult.

slarse:
Just off the top of my head, I think it might be easier to analyze expressions when they are to be printed to determine a minimal amount of required parentheses. It should be pretty straightforward given a table of operator precedences.

slarse:
I'm giving this a spin today.

slarse:
So, this really isn't very difficult (I think). We don't need to mess with any implicitness of elements, as given an element we can determine if it needs to be parenthesized based only on its parent. If I'm not missing something, we must parenthesize an expression `e` with an operator in it iff:

1. The parent `p` is a unary operator.
2. The parent `p` is a binary operator with higher precedence than `e`
3. The parent `p` is a binary operator, `e` appears as the X-hand-side operand of `p`, _and_  `e`'s operator is Y-associative

[...]

Issue #3794, opened 2021-02-12
Discussion: Redesigning the sniper printer


slarse:
Hi!

Over the past few months, I've been working out kinks and bugs in the sniper printer quite frequently. It's a very cool piece of software that does some impressive things, but I have noted several problems with the design that, in my opinion, cannot be overcome with the current architecture. I'd therefore like to discuss a complete redesign.

For context, I'll first provide a brief description of what the sniper printer does and how it does it, then I'll move on to the problems, and finally my proposed solution.

### What is the purpose of the sniper printer?

The sniper printer provides a means to parse source code into a Spoon AST, perform some transformations, and then print it back to file with minimal formatting changes to non-transformed parts. In the literature, this has been referred to as _high-fidelity code transformation_.

For applications where the output of a Spoon transformation is stored in the

slarse:
I've thought about maintaining an LL-AST in the background some more, and come to the following insights.

Deletions of elements should be very straightforward: simply delete the corresponding subtree in the LL-AST.

Insertions should be fairly straightforward. On an insertion, pretty-print from the root of the inserted subtree. It should be relatively straightforward to map the print actions to the correct nodes in the tree by using the "enter" and "exit" methods of the pretty-printer. I'm sure there will be some complications, but I don't think anything insurmountable should appear here.

Modifications to existing nodes (e.g. renaming a method or inserting a modifier on a field) is where I think the bulk of the work has to be made. I'm hoping to make some clever use of `CtRole` here, but I'm relatively certain that there will be difficulties here.

jrfaller:
Hi @slarse!

I'm coming here since it a better place to discuss than our long email thread.

First thanks for the analysis of prior art, it is clearly very handy to have a structure with everything in it such as the LL-AST.

Maybe a dumb question for starters but I am asking it anyway: why would it be so bad to have that directly in Spoon main's representation? They could be hidden to the user by default, but still present? Or there are various technical reasons making this impossible?

I ask that because from experience, I find it hard to keep consistency between complex structures, something that will be required to do with your proposal of background LL-AST, so if we could avoid it it would be great! If we cannot, I strongly suggest that one of the two structures is automatically derived from the other (presumably the AST from the LL-AST).

Then for the transformations, I am not sure any operation is easy in fact because it depends on how we can ventilate the whitespaces into the LL-AST.

[...]

PR #3823, opened 2021-03-04
wip: feat: Print minimal brackets for nested operators

slarse:
Related to #3809

### DIFF
--- a/spoon-pom/pom.xml
+++ b/spoon-pom/pom.xml
@@ -54,7 +54,13 @@
  		<version>5.7.1</version>
  		<scope>test</scope>
 	</dependency>
-  
+    <dependency>
+        <groupId>org.junit.jupiter</groupId>
+        <artifactId>junit-jupiter-params</artifactId>
+        <version>5.7.1</version>
+        <scope>test</scope>
+    </dependency>
+
   <dependency>
     <!-- must come after junit5 deps, otherwise it overrides the junit version -->
     <!-- must be here in the parent POM, because the parent always comes last, see https://stackoverflow.com/questions/28999057/order-of-inherited-dependencies-in-maven-3 -->
--- a/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
+++ b/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
@@ -2164,6 +2164,7 @@ public class DefaultJavaPrettyPrinter implements CtVisitor, PrettyPrinter {
 
 		int parentPrecedence = OperatorHelper.getBinaryOperatorPrecedence(parent.getKind());
 		int precedence = OperatorHelper.getBinaryOperatorPrecedence(operator.getKind());
-		return precedence < parentPrecedence || associativity != positionInParent;
+		return precedence < parentPrecedence
+				|| (precedence == parentPrecedence && associativity != positionInParent);
 	}
 }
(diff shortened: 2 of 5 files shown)


spoon-pom/pom.xml
```diff
@@ -54,7 +54,13 @@
  		<version>5.7.1</version>
  		<scope>test</scope>
 	</dependency>
-  
+    <dependency>
+        <groupId>org.junit.jupiter</groupId>
+        <artifactId>junit-jupiter-params</artifactId>
+        <version>5.7.1</version>
+        <scope>test</scope>
+    </dependency>
```
slarse, line 62: This dependency is necessary for using the `@ParameterizedTest` annotation.

jrfaller:
Hi @slarse! It seems neat! I don't see in the diff when this method is triggered but I believe this will only be applied to transformed nodes, to keep the unmodified code with its original parentheses?

slarse:
> Hi @slarse! It seems neat! I don't see in the diff when this method is triggered but I believe this will only be applied to transformed nodes, to keep the unmodified code with its original parentheses?

In terms of sniper printing, yes. This patch applies to the default printer (DJPP = DefaultJavaPrettyPrinter), which has its own printing style, and is what we must use when we can't sniper print for some reason (modified node, new node, etc), or when we just don't want to sniper print.

In terms of the current state of affairs, it will also improve the sniper printing itself in some cases, as there will be far fewer parentheses to try to match to actual parentheses that exist in the source code. Parentheses should now never appear out of thin air, as has previously been a problem both in default printing and sniper printing. It will be less beneficial for sniper printing if we manage to implement our ideas in #3794, but it'll still be useful when we need to print new or changed expressions.

jrfaller:
Thanks for the explanation! Therefore LGTM!

slarse:
@monperrus This is ready for your consideration. Here are a a couple of thoughts I have:

* The word "brackets" is used internally in the printers. I think "parentheses" is more precise, as brackets more often refer to square brackets `[]`, while I've never heard parentheses refer to anything but `()`. Thoughts?
* Currently, the parenthesis optimization is enabled by default for the sniper printer, but not for the default printer. There is no way to use the public API to enable it, as I didn't want to introduce more clutter to the DJPP API. Keep this way, or enable for DJPP, or make it possible to enable through DJPP API?

monperrus:
Thanks a lot @slarse 

> The word "brackets" is used internally in the printers. I think "parentheses" is more precise, as brackets more often refer to square brackets [], while I've never heard parentheses refer to anything but (). Thoughts?

Both are valid AFAIK, but it's good to be consistent. All comment and name changes outside the public API are welcome!

> Currently, the parenthesis optimization is enabled by default for the sniper printer, but not for the default printer.

This is good this way.

> I'm not sure "optimization" is the best word to use here, it's a bit vague. Minimization? Redundancy elimination?

I like "Minimization" (optimization is either minimization or maximization) 

Overall, excellent PR, and we love parametrized tests! Will merge.

slarse:
> Thanks a lot @slarse
> 
> Both are valid AFAIK, but it's good to be consistent. All comment and name changes outside the public API are welcome!

Yes, both are valid: "bracket" is British and "parenthesis" is American English. The problem I find in Spoon is that the word "bracket" is used for everything (square bracket, round bracket, curly bracket are all just denoted bracket). This is a bit confusing to me when reading the code. This is all internal, though, I couldn't find a single mention in the public API.

However, I think that this PR is not the one to change widespread terminology use in Spoon. I'll use the term "round bracket".

> > Currently, the parenthesis optimization is enabled by default for the sniper printer, but not for the default printer.
> 
> This is good this way.

Right, so just to be clear here, it also _can't_ be enabled for the default printer via the public API as the setter is protected. That's intentional, I think it's good we keep this internal until it's been field tested a bit in the Sniper.

> I like "Minimization" (optimization is either minimization or maximization)

That is true. I'll migrate over to use "minimization".

slarse:
On the note of perfectly reproducing formatting under AST transformations, db058e1and 73338f7 show that IntelliJ isn't perfect either ...

monperrus:
Thanks a lot @slarse . This is a very important contribution for the sniper mode.

PR #3824, opened 2021-03-04
doc: Improve documentation for the DJPP get/setPreprocessors methods

slarse:
The docs for DJPP get/setPreprocessors currently contain a broken link to some (probably very) old type that is no longer available, and aren't particularly clear about how DJPP actually uses the preprocessors. This PR removes the broken link and clarifies how the preprocessors are used.


monperrus:
thanks @slarse , we :heart: doc PRs

slarse:
I really like writing nice docs, so I'll try to start sprinkling a few of these in every now and then, at least on the parts of the code base I'm working on :)

PR #3830, opened 2021-03-10
Remove useless classloader

seintur:
classloader is never used independently from inputClassLoader. The StandardEnvironment only deals with inputClassloader and its class path elements. classloader is then useless and can be removed.


seintur:
Sorry my mistake

Issue #3831, opened 2021-03-11
Dependency should have a test scope


cesarsotovalero:
I noticed that the dependency `org.tukaani:xz:1.8`, located in `spoon-core`, has a (default) `compile` scope. However, it is only used in method `testLZMACompressionType` of class  `spoon.test.serializable.ModelStreamerTest`. Therefore, this dependency should have a `test` scope. By changing the scope of this dependency, the size of the `spoon-core` bundled artifact is reduced by [106 KB](https://mvnrepository.com/artifact/org.tukaani/xz/1.8), which is good for the Spoon clients because they will pull fewer transitive dependencies when using the Spoon library.

PS: I analyzed the Spoon dependencies with [DepClean](https://github.com/castor-software/depclean).

PR #3832, opened 2021-03-11
fix: dependency rg.tukaani:xz should have a test scope

cesarsotovalero:
Fix https://github.com/INRIA/spoon/issues/3831


monperrus:
Indeed, XZ support is not implemented, see https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/SerializationModelStreamer.java#L65

LGTM, will merge, thanks a lot @cesarsotovalero

monperrus:
thanks for your contribution @cesarsotovalero you're really a dependency king :)

Issue #3833, opened 2021-03-11
Integrate DepClean in the Spoon CI pipeline


cesarsotovalero:
The purpose of the PR is to suggest integrating [DepClean](https://github.com/castor-software/depclean) in the Spoon CI pipeline. Having a dedicated DepClean job is useful to keep track of the status of the Spoon dependencies. 

The DepClean plugin allows having a guard to ensure that unused dependencies are not added to the Spoon bundled artifact. For example, the build may fail if a PR adds at least one unused direct dependency to the Spoon dependency tree.

See also https://github.com/INRIA/spoon/pull/3167 and https://github.com/INRIA/spoon/pull/3832

PR #3834, opened 2021-03-11
chore: add dependency checking with DepClean in the CI pipeline

cesarsotovalero:
Fix https://github.com/INRIA/spoon/issues/3833

This PR integrates [DepClean](https://github.com/castor-software/depclean) (v2.0.0) in the Spoon CI. 
DepClean plugin is added as a dedicated Maven profile in `spoon-pom`
The plugin is currently configured to fail in the presence of unused direct dependencies (although more strict guards can be also configured, eg. to fail if there are unused transitive dependencies).

**NOTE:** The DepClean dependency analysis is fast: the execution time of this job on Travis CI is < 3min


monperrus:
nice! what about having this in travis-extra.sh?

cesarsotovalero:
Hi @monperrus, I just added a depclean guard in `travis-extra.sh`. The build of `spoon-decompiler` project will fail if there is any unused direct dependency in this project.

chore/travis/travis-extra.sh
```diff
@@ -33,6 +33,7 @@ git diff
 
 mvn -q test
 mvn -q checkstyle:checkstyle license:check
+mvn -q se.kth.castor:depclean-maven-plugin:2.0.0:depclean -DfailIfUnusedDirect=true
```
slarse, line 36: Isn't `-DfailIfUnusedDirect=true` now duplicated? It's already in the pom.xml?

cesarsotovalero: No, it's not. Let me explain.

In `spoon-pom/pom.xml` depclean is added as a Maven profile configured for `spoon-core`, which allows the job to be triggered separately in Travis. 

In `travis-extra.sh`, depclean is executed directly from the command line because this script makes several additional checks in several Spoon extra projects. Using depclean directly gives more flexibility to developers who can configure depclean with [optional parameters](https://github.com/castor-software/depclean#optional-parameters) for each project. For example, to ignore some dependencies that they know are used dynamically or that simply don't want to analyze.

slarse: Right, I get that, but why not use the depclean profile here as well? I.e.

```suggestion
mvn -q se.kth.castor:depclean-maven-plugin:2.0.0:depclean -Pdepclean
```

I guess I just don't understand why we'd want different configurations.

On a related note, I think what @monperrus meant was to put the depclean job as part of `travis-extra` instead of having it's own dedicated workflow. With this configuration, depclean will run twice, which doesn't seem useful?

cesarsotovalero: > With this configuration, depclean will run twice, which doesn't seem useful?

It depends on the decision of the maintainer. I saw a `travis-coverage.sh` so I added depclean job in the Travis matrix.
But of course, having depclean only in `travis-extra` is also an option.

@monperrus Is this what you meant?

monperrus: yes, I meant adding some lines in `travis-extra.sh` instead of adding a new file/job.

BTW: now that we're out of Travis, we could rename `travis*.sh` to `ci-*.sh`

slarse: @monperrus On that note, we should probably remove all Travis stuff.

slarse: See #3836

.github/workflows/tests.yml
```diff
@@ -76,3 +76,18 @@ jobs:
...
+    name: Dependency usage analysis
+    steps:
+      - uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f # v2.3.4
+      - uses: actions/setup-java@d202f5dbf7256730fb690ec59f6381650114feb2 # v1.4.3
+        with:
+          java-version: 11
+      - name: Use silent log config
+        run: mv chore/travis/logback.xml src/test/resources/
+      - name: Build and test
+        run: mvn package -DskipTests -Pdepclean
```
slarse, line 93: This still causes duplication. It's rather confusing, but `travis-extra.sh` is actually executed in the GitHub Actions `extra` job, so just having `mvn package -DskipTests -Pdepclean` in there is sufficient to run it with our current CI setup.

Sorry about the confusing setup, it may or may not be my fault :). I'm gonna fix it ... #3836

cesarsotovalero: Thank @slarse. This is indeed confusing. 
To make it more clear, I've renamed the references to the travis scripts, from `travis-*` to `ci-*`.
And of course, I removed the redundant depclean GH actions job (now added to `ci-extra.sh`)

monperrus:
> it may or may not be my fault :). I'm gonna fix it 

Thanks @slarse No need to use the word 'fault', Spoon is a blameless community, see https://sre.google/sre-book/postmortem-culture/

slarse:
> Thanks @slarse No need to use the word 'fault', Spoon is a blameless community, see https://sre.google/sre-book/postmortem-culture/

Poor choice of word, I just wanted to clarify that the confusion there to begin with, and not introduced by the PR.

monperrus:
Thanks for the update.

Two questions:

- my understanding of Maven is that we don't need to modify the pom if we use a plugin from the command line as in `mvn -q se.kth.castor:depclean-maven-plugin:2.0.0:depclean -Pdepclean`. Can we do this here?

- here we would have the version number `2.0.0` at two different places, in the pom and in the command line. That's double maintenance and a risk of mismatch. Could we simply call `mvn -q se.kth.castor:depclean-maven-plugin:depclean`?

cesarsotovalero:
@monperrus To answer your questions:

> my understanding of Maven is that we don't need to modify the pom if we use a plugin from the command line as in mvn -q se.kth.castor:depclean-maven-plugin:2.0.0:depclean -Pdepclean. Can we do this here?

You're definitely right, we don't need to modify the `pom` when executing a Maven plugin from the command line. 

However in the case of the line `mvn -q se.kth.castor:depclean-maven-plugin:2.0.0:depclean -Pdepclean`, what it does is executing a [Maven profile](https://maven.apache.org/guides/introduction/introduction-to-profiles.html) with id `depclean` (through the parameter `-Pdepclean`). I added this profile in `spoon-pom/pom.xml` in this PR. 

There are two alternatives:

1. Not using this profile at all, and just call depclean directly from the command line in `ci-extra`, i.e. `mvn -q se.kth.castor:depclean-maven-plugin:2.0.0:depclean -DfailIfUnusedDirect=true`. The drawback of this approach is that the version of the plugin needs to be specified directly in the command (this is not necessary when using a profile).

2. Add the depclean plugin directly in the `pom` of Spoon projects', as it is done with checkstyle in https://github.com/INRIA/spoon/blob/master/pom.xml#L251-L273, https://github.com/INRIA/spoon/blob/master/spoon-smpl/pom.xml#L33-L50, and https://github.com/INRIA/spoon/blob/master/spoon-decompiler/pom.xml#L83-L100. The drawback of this approach is that there is a risk of having plugin duplicates, as with checkstyle. 

> here we would have the version number 2.0.0 at two different places, in the pom and in the command line. That's double maintenance and a risk of mismatch. Could we simply call mvn -q se.kth.castor:depclean-maven-plugin:depclean?

This is correct. We can simply call `mvn -q se.kth.castor:depclean-maven-plugin:depclean -Pdepclean` if we use depclean as a Maven profile. 

I just found that we can actually do better thanks to the [Maven plugin prefix resolution](https://maven.apache.org/guides/introduction/introduction-to-plugin-prefix-mapping.html), so the command gets simplified to `mvn -q depclean:depclean -Pdepclean`.

### Proposal

I think the best approach is not using a global profile and just adding depclean in the `pom` of the project (alternative 2. above). This way, we can just call `mvn -q depclean:depclean` in the CI, as it is done with checkstyle, see https://github.com/INRIA/spoon/blob/master/chore/travis/travis-extra.sh#L35.

WDYT?

monperrus:
Thanks for the analysis.

> I think the best approach is not using a global profile and just adding depclean in the `pom` of the project (alternative 2. above). This way, we can just call `mvn -q depclean:depclean` in the CI

Yes, that's a good solution. Plus I think that @renovate-bot would only update versions in the pom, and not versions in arbitrary shell scripts.

slarse:
I also like alternative 2. In addition to what @monperrus, it also makes it much less of a hassle to run depclean locally due to the much simpler command, if one needs to do some dependency debugging.

I also strongly believe that clearly declaring project dependencies in the pom-file leads to less risk of inconsistencies down the road.

cesarsotovalero:
I added depclean as a plugin in spoon-core, spoon-decompiler, spoon-smpl, and spoon-visualization; it runs on the CI. DepClean is configured to fail if there is at least one unused direct dependency in those projects. Notice that in spoon-visualization project some dependencies are excluded from the depclean analysis (https://github.com/INRIA/spoon/pull/3834/commits/56a0029b5fe63edaac4ff5f0747402c296bd8104), I let a comment there.

**Side note:** I found the current Spoon CI very verbose (mostly due to tests logs) and confusing (mostly due to legacy travis stuff). I hope this will be improved soon.

monperrus:
Thanks a lot! LGTM.

slarse:
> **Side note:** I found the current Spoon CI very verbose (mostly due to tests logs) and confusing (mostly due to legacy travis stuff). I hope this will be improved soon.

Noted, see #3839

monperrus:
Thanks a lot @cesarsotovalero, looking forward to our first broken build because of a bloated dependency!

PR #3838, opened 2021-03-12
review: chore: Clean up output in GitHub Actions workflows

### LINKED ISSUE #3839: CI output is very verbose
It's absolutely full of "Download progress". I'm working on reducing overall verbosity and making it easier to find only the test output (which is typically what one wants to look at).

slarse:
Fix #3839 

This PR significantly cleans up the CI output of the primary build matrix by building the project and fetching dependencies in a separate step, before running the tests. Fetching _all_ test dependencies requires a little hack (start running tests, then abort after 20 seconds), because no matter what you do Maven will not fetch certain dependencies needed for test execution before-hand. Even when using something like `dependency:go-offline`.

If you have a look at the [test step of the current build (NOTE: You may need to refresh after clicking this link)](https://github.com/INRIA/spoon/pull/3838/checks?check_run_id=2111450564#step:7:1), the output actually starts with tests, and not with thousands of lines of "download progress", as was previously the case.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -27,6 +27,8 @@ jobs:
             java: 11
           - os: windows-latest
             java: 15
+    env:
+      MAVEN_OPTS: -Djava.src.version=${{ matrix.java }} -Dhttp.keepAlive=false -Dmaven.wagon.http.pool=false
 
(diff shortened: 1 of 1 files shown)


nharrand:
Thanks!

PR #3842, opened 2021-03-15
review: chore: Add daily cache of Maven dependencies

### LINKED ISSUE #3840: Cache Maven dependencies to speed up build process
We spend almost 4 minutes just downloading dependencies in each build (about 1/3 of total build time). That's made worse by us disabling connection pooling to avoid transfer timeouts.

Caching the .m2 directory could significantly speed up the builds.

slarse:
Fix #3840

This PR adds a cache for Maven dependencies in the primary build matrix. The cache is dependent on the current date, and so will never be used for more than 24 hours. This makes the risk of stale dependencies causing trouble relatively small, and at worst the problem will go away the next day.

The speedup is significant, e.g. for `Tests with Java 1.8 on ubuntu-latest`:

* Before cache: [12m31s](https://github.com/INRIA/spoon/runs/2111403925)
* After cache: [9m9s](https://github.com/INRIA/spoon/pull/3842/checks?check_run_id=2111720116)

All-in-all, we see roughly a 3-minute reduction across all jobs individually, as well as the total wall time for CI (not counting CPU time, but actual time a developer has to wait), which is down from ~14 to ~11 minutes.

Unfortunately, I couldn't find a way to enable a "global" cache for all jobs, so the caching steps must be duplicated for each job that needs them. There's currently no mechanism in GitHub Actions to reuse steps, either. I think the speedup is well worth that complexity, and as we use caches that are at most 24 hours old, I think there's minimal risk of any noteworthy complications with caching.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -43,9 +43,7 @@ jobs:
         uses: actions/cache@26968a09c0ea4f3e233fdddbafd1166051a095f6 # v2.1.4
         with:
           path: ~/.m2/repository
-          key: |
-            # date is used in the key ==> cache hits are never more than 24 hours old
-            ${{ runner.os }}-${{ steps.get-date.outputs.date }}-${{ hashFiles('**/pom.xml') }}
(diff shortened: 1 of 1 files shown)


nharrand:
LGTM. Let's try this.

Issue #1876, opened 2018-02-22
good first issue: add documentation for 3 undocumented public methods


monperrus:
In Spoon, we want that all public methods are properly documented.

We have a test case to check this: https://github.com/INRIA/spoon/blob/51c4b2211922e71914f74ce4341a6232c079f174/src/test/java/spoon/test/architecture/SpoonArchitectureEnforcerTest.java#L120

This test case temporarily skips small methods.

To get this issue done:

- List the undocumented public methods:
  - lower the threshold `size() > 33` in the test case
  - run the test case, this will list you undocumented public methods
- Add appropriate Javadoc documentation `/** .... */`
- Create a pull request with your changes

ZhectorSM:
Hi guys, I'm new in the open-source world, so
I'll try to help you with some issues starting with this one.

monperrus:
Excellent, welcome @ZhectorSM!!

ZhectorSM:
doc: add API documentation to public methods
PR #3152

slarse:
Also see #3923

TheMarvelFan:
Is this issue still open?

MartinWitt:
Yes, feel free to start working on it

TheMarvelFan:
@MartinWitt @slarse @monperrus @ZhectorSM Can someone please review PR #5490 
It would mean a lot to me if someone can provide feedback on this.

TheMarvelFan:
Hi has anyone looked at PR #5490 yet?

AzazelSensei:
Opened https://github.com/INRIA/spoon/pull/6881 with Javadoc on scanFields and the threshold moved to 31.

PR #3854, opened 2021-03-22
doc: Add API documentation to 3 undocumented public methods

Rohitesh-Kumar-Jain:
Hi,
Added documentation for 3 undocumented public methods and lowered the threshold for documentation to 32
issue: #1876

I hope the documentations makes some sense :P

Best,
Rohitesh

### DIFF
--- a/src/main/java/spoon/pattern/PatternBuilder.java
+++ b/src/main/java/spoon/pattern/PatternBuilder.java
@@ -303,6 +303,11 @@ public class PatternBuilder {
 		return false;
 	}
 
+	/**
+	 * Builds a Pattern and returns it
+	 * @throws SpoonException if the patter has been built already
+	 * @return the built patter
+	 */
 	public Pattern build() {
 		if (built) {
 			throw new SpoonException("The Pattern may be built only once");
--- a/src/main/java/spoon/pattern/PatternBuilder.java
+++ b/src/main/java/spoon/pattern/PatternBuilder.java
@@ -305,8 +305,8 @@ public class PatternBuilder {
 
 	/**
 	 * Builds a Pattern and returns it
-	 * @throws SpoonException if the patter has been built already
-	 * @return the built patter
+	 * @throws SpoonException if the pattern has been built already
+	 * @return the built pattern
 	 */
 	public Pattern build() {
 		if (built) {
--- a/src/main/java/spoon/reflect/visitor/PrinterHelper.java
+++ b/src/main/java/spoon/reflect/visitor/PrinterHelper.java
@@ -230,6 +230,11 @@ public class PrinterHelper {
 		return this;
 	}
 
+	/**
+	 * writes as many newlines as needed for the current line to match the end line of the passed element
+	 * @param e is a CtElement whose line number will be preserved by adding newlines
+	 * @return PrinterHelper
+	 */
 	public PrinterHelper adjustEndPosition(CtElement e) {
 		if (env != null && env.isPreserveLineNumbers() && e.getPosition().isValidPosition()) {
 			// let's add lines if required
--- a/src/main/java/spoon/reflect/visitor/PrinterHelper.java
+++ b/src/main/java/spoon/reflect/visitor/PrinterHelper.java
@@ -232,7 +232,7 @@ public class PrinterHelper {
 
 	/**
 	 * writes as many newlines as needed for the current line to match the end line of the passed element
-	 * @param e is a CtElement whose line number will be preserved by adding newlines
+	 * @param e element whose line number will be preserved by adding newlines
 	 * @return PrinterHelper
 	 */
 	public PrinterHelper adjustEndPosition(CtElement e) {
--- a/src/main/java/spoon/support/visitor/java/reflect/RtMethod.java
+++ b/src/main/java/spoon/support/visitor/java/reflect/RtMethod.java
@@ -138,8 +138,8 @@ public class RtMethod {
 	}
 
 	/**
-	 * @param method is the Method which is to be wrapped in a RtMethod
-	 * @return an object of RtMethod after calling the parameterized constructor of the RtMethod class
+	 * @param method is the {@link Method} which is to be wrapped in a RtMethod
+	 * @return a wrapper around the passed method
 	 */
 	public static RtMethod create(Method method) {
 		return new RtMethod(method.getDeclaringClass(), method, method.getName(), method.getReturnType(),
--- a/src/main/java/spoon/support/visitor/java/reflect/RtMethod.java
+++ b/src/main/java/spoon/support/visitor/java/reflect/RtMethod.java
@@ -137,6 +137,10 @@ public class RtMethod {
 		return getDeclaringClass().getName().hashCode() ^ getName().hashCode();
 	}
 
+	/**
+	 * @param method is the Method which is to be wrapped in a RtMethod
+	 * @return an object of RtMethod after calling the parameterized constructor of the RtMethod class
+	 */
 	public static RtMethod create(Method method) {
 		return new RtMethod(method.getDeclaringClass(), method, method.getName(), method.getReturnType(),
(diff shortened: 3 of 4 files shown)


src/main/java/spoon/support/visitor/java/reflect/RtMethod.java
```diff
@@ -137,6 +137,11 @@ public int hashCode() {
 		return getDeclaringClass().getName().hashCode() ^ getName().hashCode();
 	}
 
+	/**
+	 * Creates a new {@link RtMethod}
+	 * @param method is passed as Method
+	 * @return {@link RtMethod}
+	 */
```
slarse, line 144: This Javadoc does not really say anything, it mostly just duplicates the method header.

What this method does is to wrap a `Method` in an `RtMethod`. Note that it's entirely fine to have a Javadoc _without_ a general description. That is to say, like so:

```suggestion
	/**
	 * @param method Say something about the parameter
	 * @return Say something about the return value
	 */
```

Could you try to improve this Javadoc to say something meaningful?

slarse (changes requested):
Hi @Rohitesh-Kumar-Jain , thanks a lot for this first contribution! We love docs contributions in Spoon, so it's very much appreciated.

I have added some comments on your PR for potential improvements. Could you please look them over and see what you can do?

Rohitesh-Kumar-Jain:
Hi @slarse,

I am so glad that you reviewed my PR and gave feedback, I will address all the comments in the next commit. 

After resolving issues with this PR, I would love to work on other issues as well.

src/main/java/spoon/reflect/visitor/PrinterHelper.java
```diff
@@ -231,10 +231,9 @@ public PrinterHelper adjustStartPosition(CtElement e) {
 	}
 
 	/**
-	 * If the environment in which spoon is executed tries to preserve the original line number and the the passed
-	 * indexes of CtElement e are known then it will generate new lines
-	 * @param e is a {@link CtElement} on which new lines are going to be generated
-	 * @return {@link PrinterHelper}
+	 * writes as many newlines as needed for the current line to match the end line of the passed element
+	 * @param e is a CtElement whose line number will be preserved by adding newlines
+	 * @return PrinterHelper
```
slarse, line 236: I think you misinterpreted my feedback about referring to types: just writing e.g. `CtElement` is less informative than `{@link CtElment}`, and it _still_ refers explicitly to the type. Look for example at [this standard library Javadoc](https://docs.oracle.com/javase/7/docs/api/java/util/ArrayList.html#add(int,%20E)). The Javadoc describes the semantic meaning of the values, not their types. Note also that the type of each parameter is very evident in the Javadoc as they appear in the signature, so unless it actually helps understand something, I would omit the explicit types.

 What I meant was something like this:

```java
// no; refers explicitly to the type, which is already specified in the method header
* @param e is a {@link CtElement} whose line number will be preserved by adding newlines

// no; also refers explicitly to the type, but there will be no clickable link in the Javadoc HTML
* @param e is a CtElement whose line number will be preserved by adding newlines

// yes; does not refer explicitly to the type, we just say "element", which is more general
* @param e element whose line number will be preserved by adding newlines
```

To be clear, if you need to refer to the type, use `@link` tags. If you don't need to refer to the type, describe the value in natural language without the type name.

src/main/java/spoon/support/visitor/java/reflect/RtMethod.java
```diff
@@ -138,9 +138,8 @@ public int hashCode() {
 	}
 
 	/**
-	 * Creates a new {@link RtMethod}
-	 * @param method is passed as Method
-	 * @return {@link RtMethod}
+	 * @param method is the Method which is to be wrapped in a RtMethod
+	 * @return an object of RtMethod after calling the parameterized constructor of the RtMethod class
```
slarse, line 142: The `@param` is good in terms of information, but the comment about types still applies. If you want to refer to types (which isn't unreasonable here; this method just wraps an object of one type in an object of another type), then use `@link`.

The `@return` is much too specific. Only write what's actually important to a caller of the method. Does it matter _how_ the wrapper is created? What if we change `RtMethod` to a fluent API instead, should that matter to this method? I would suggest something like `@return a wrapper around the passed method` instead, because that describes the semantics of the value, and not the details of its type or how it came to be.

slarse (changes requested):
Hi @Rohitesh-Kumar-Jain,

Thank you for your revision, now the docs are really shaping up. I've added some final comments, I think with one more revision this will be ready to merge.

As I think you misunderstood one of my previous comments: don't hesitate to ask about anything you think is unclear!

src/main/java/spoon/pattern/PatternBuilder.java
```diff
@@ -304,8 +304,9 @@ boolean isInModel(CtElement element) {
 	}
 
 	/**
-	 * Builds a Pattern, if the pattern isn't built yet
-	 * @return {@link Pattern} which is built after executing this method
+	 * Builds a Pattern and returns it
+	 * @throws SpoonException if the patter has been built already
+	 * @return the built patter
```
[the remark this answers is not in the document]
Rohitesh-Kumar-Jain: Thanks for pointing this out :)

One thing I am curious about is why one of my commits failed 1 out of 7 tests (locally I passed all the test cases), when I committed the same changes again, it passed all 7.

src/main/java/spoon/reflect/visitor/PrinterHelper.java
Rohitesh-Kumar-Jain: Thanks for sharing the example and clearing things up, for now on I will just ask if something isn't clear to me.

src/main/java/spoon/pattern/PatternBuilder.java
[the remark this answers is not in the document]
slarse: That was a connection timeout, it had nothing to do with your work: https://github.com/INRIA/spoon/runs/2186431693#step:8:3661

We have some problems with an old Maven repository that we used to use, the last traces of it are being removed (#3860).

src/main/java/spoon/support/visitor/java/reflect/RtMethod.java
Rohitesh-Kumar-Jain: Thanks for the suggestions, I will incorporate the necessary changes in the next commit.

src/main/java/spoon/reflect/visitor/PrinterHelper.java
slarse: > for now on I will just ask if something isn't clear to me.

Very good!

src/main/java/spoon/pattern/PatternBuilder.java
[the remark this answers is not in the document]
Rohitesh-Kumar-Jain: Thanks for letting me know the cause.

slarse:
@Rohitesh-Kumar-Jain Could you make a final pass over your PR and check that you've fixed everything you want to fix, and then ping me when you're happy with it?

Rohitesh-Kumar-Jain:
> @Rohitesh-Kumar-Jain Could you make a final pass over your PR and check that you've fixed everything you want to fix, and then ping me when you're happy with it?

Hi @slarse this commit seems fine to me, excited to see what lies ahead of this PR.

slarse:
@Rohitesh-Kumar-Jain I'm satisfied as well.

@monperrus Ping for final approval.

slarse (approved):

monperrus:
@Rohitesh-Kumar-Jain congratulations and thank you very much for your first contribution. Welcome to the Spoon community!

Thanks @slarse for the shepherding.

Rohitesh-Kumar-Jain:
Hi @monperrus,
I am glad to make the first contribution, I will look up for some other issues to work upon.
If you want to suggest a particular issue, I will be happy to work on it as well.

PR #3856, opened 2021-03-23
fix: Fix sniper printer failing to prepend whitespace to top-added CtModifiable

### LINKED ISSUE #3855: bug: Sniper printer fails to prepend whitespace to CtModifiable added at start of non-empty collection
If you add a type member at the top of a class, the sniper printer prints it on the same line as the header, just after the opening curly bracket. This happens only if the class is non-empty, with an empty class the newline and indentation is properly prepended to the type member.

This applies to all kinds of type members. Working on a fix.

slarse:
Fix #3855 

This fixes a bug causing the sniper printer to not prepend whitespace to type members and local variables added at the start of a non-empty collection. For example, given this class:

```java
class Cls {
    private int someInt = 2;
}
```

adding a new field `private int otherInt = 3` at the top would be printed like so:

```java
class Cls {private int otherInt = 3;
    private int someInt = 3;
```

The same thing applies to local variables in non-empty statement lists.

The bug is caused by the sniper printer looking the source fragment of any `CtModifiable` or `ModifierKind` only by role, which causes _any_ source fragment belonging to a `CtModifiable` to match any other `CtModifiable`. This itself is required to avoid other corner cases, it appears. Thus, the sniper always thinks it finds a match for a newly created `CtModifiable` when the collection it's added in is non-empty, and then does not print the "synthetic" whitespace.

The fix is simply to require any `CtModifiable` to already exist (have a valid source position) to be matched by role alone, see 5a03df6.

**Note:** The test cases fail on Windows due to containing literal newlines. All test cases in the sniper printer tests that contain literal newlines fail on Windows. That's an issue that isn't related to this particular patch, so the new test cases currently ignore Windows. I need to debug the newline problem separately. Most likely, somewhere CRLF line endings are inserted instead of LF line endings, even though the sniper printer should respect the line endings of the original source file.

### DIFF
--- a/src/main/java/spoon/support/sniper/internal/AbstractSourceFragmentPrinter.java
+++ b/src/main/java/spoon/support/sniper/internal/AbstractSourceFragmentPrinter.java
@@ -277,8 +277,9 @@ abstract class AbstractSourceFragmentPrinter implements SourceFragmentPrinter {
 	protected int findIFragmentIndexCorrespondingToEvent(PrinterEvent event) {
 		CtRole role = event.getRole();
 		if (role != null) {
-			if (event.getElement() instanceof CtModifiable || role == CtRole.MODIFIER) {
-				// using only roles for handling modifiers correctly
+			if ((event.getElement() instanceof CtModifiable && event.getElement().getPosition().isValidPosition())
+					|| role == CtRole.MODIFIER) {
+				// using only roles for handling modifiers and preexisting modifiables correctly
 				return findIndexOfNextChildTokenOfRole(childFragmentIdx + 1, role);
 			}
 			return findIndexOfNextChildTokenOfElement(event.getElement());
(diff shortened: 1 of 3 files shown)


monperrus:
LGTM will merge.

Note that it's a typical [Nopol](https://hal.archives-ouvertes.fr/hal-01285008/document) and [Dynamoth](https://hal.archives-ouvertes.fr/hal-01279233/document) patch, so in theory we should be able to synthesize the patch only based on the failing test cases.

slarse:
> Note that it's a typical Nopol and Dynamoth patch, so in theory we should be able to synthesize the patch only based on the failing test cases.

I could try that, but I'd still need a Windows environment to do it (otherwise, no failures :). I do have a Windows system, I just don't have it set up for any kind of development. I'll get around to it eventually!

monperrus:
Thanks @slarse!

Issue #3759, opened 2021-01-18
update website publication job


monperrus:
Until today our website http://spoon.gforge.inria.fr was hosted on the [INRIA GForge](https://gforge.inria.fr/). Yes, this service will be discontinued, see https://intranet.inria.fr/Actualite/Report-a-l-ete-2021-et-plan-d-action-Arret-de-la-forge

As alternative solution, @seintur and myself have worked to move it to Github.io.

Now, the HTML pages are hosted on https://github.com/SpoonLabs/SpoonLabs.github.io/

Yet, we need to update the cron task for publishing the website, which is a Jenkins job:

* Job: https://ci.inria.fr/sos/job/Website%20Deployer/
* Source :https://github.com/SpoonLabs/spoon-ci-config/blob/master/jenkins/jobs/Website%20Deployer.xml

monperrus:
The job must handle both the Jekyll site in `docs` and the Javadoc, see #3762

slarse:
I'm doing some infrastructure work at the moment so I'll pick up this one, too.

slarse:
It took me a while to familiarize myself with Jenkins, but now I've got it all mostly figured out. I've adapted the current deployment script to generate the new site, but I haven't written the deployment part yet.

@monperrus Should we deploy automatically to SpoonLabs/spoonlabs.github.io, or should we open pull requests whenever the website changes?

Also, I can't find that the Javadocs for previous (stable) versions are hosted anywhere, are they? Should we to set it up such that we host the docs for, let's say, every stable release from Spoon 8 and up? Plus snapshots?

monperrus:
that's great.

> @monperrus Should we deploy automatically to SpoonLabs/spoonlabs.github.io

yes!

slarse:
@monperrus I've got everything working now in a simple GitHub Action workflow (simpler than Jenkins as we have the credentials at hand). Here's an [automatic update example on my own fork](https://github.com/slarse/SpoonLabs.github.io/commit/810b4c12c16b4bf7ea0c6a6e1e14848b59c23854). Note that the diff is quite large due to the switch in markdown processor (see #3859).

As the move from Travis to GHA was kind of painful, I've written the deployment script to be relatively agnostic of where it is executed. In case we need to move again. For the final revision, I'll configure the action **to update the website once a day**.

Before I can make the final preparations for this to work on the actual Spoon website, I need #3859, #3862 and #3863 to be merged.

monperrus:
[...]

PR #3859, opened 2021-03-24
chore: Use the kramdown Markdown processor for Spoon's website

slarse:
#3759 

The `redcarpet` processor isn't supported by Jekyll 4.0+, see https://github.com/jekyll/jekyll/issues/7838#issuecomment-537254511.

### DIFF
--- a/doc/_config.yml
+++ b/doc/_config.yml
@@ -34,11 +34,8 @@ exclude:
 
 # same for all
 sidebar_accordion: true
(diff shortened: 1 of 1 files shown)


slarse:
CI failure is due to maven.inria.fr being down again, and we apparently still have some links to that. I'll look into removing all of them tomorrow.

monperrus:
good renovation!

PR #3863, opened 2021-03-25
chore: Add website CNAME

slarse:
#3759 

The CNAME record is required for the custom URL to work, and currently it has been added manually over in spoonlabs.github.io.

We could also generate the CNAME file in the deployment script, but it seems simpler to just have it in the repo.


monperrus:
Thanks @slarse

Issue #3216, opened 2020-01-27
Don't expose maven.inria.fr in Spoon parent-pom


surli:
The Spoon snapshot repository is exposed in the parent-pom which might cause malfunction if the nexus server is down (which is the case right now) even so Spoon does not need this repository to works: it's only use for revapi comparison. It would be better to use a profile for revapi and to add the repo only in that profile.

surli:
Or maybe we should just pushing the snapshot on maven.inria.fr an use central for that too?...

surli:
FTR the server restarted, but looks like I informed them about it, so not something really well monitored or used it seems :)

monperrus:
simple PR welcome :-)

surli:
Well I'm a bit afraid to break the CI infra, I've not followed well the work done for continuous deployment... I guess best would be to push snapshots on central if it's possible, not sure what needs to be changed though.

monperrus:
What about pushing the snapshots to https://repository.ow2.org/nexus/ instead of maven.inria.fr?

surli:
Could be better, but I still think we should avoid to release Spoon with a reference to an external repository, especially if it's not needed for using it.

surli:
Note we got the same problem than last year just now: apparently maven.inria.fr is down and it's blocking XWiki build since Spoon declared the repository.

slarse:
@surli We don't use maven.inria.fr anymore, we switched to OW2 nexus a few months back. So I'm guessing this is for an older version of Spoon? But I agree that it would be better for the snapshot repository to be optional. It's a bit of an annoyance even when working as expected, as Maven searches that repository for everything else as well.

I suppose I'm the one who's most down and dirty with the CI infrastructure at the moment, so I'll see what I can do to remove the snapshot repository from the pom file. It shouldn't be much of a problem, most of the CI builds inject the snapshot repository into other projects' pom files anyway. Perhaps only the snapshot deployer job actually needs the repository declared.

surli:
> So I'm guessing this is for an older version of Spoon?

Indeed we're still few version behind I need to upgrade but last time I tried we got issues. 

>  I'll see what I can do to remove the snapshot repository from the pom file

Well back when I created this issue, the snapshot repo was used for revapi, if it's still the case then you can probably just declare it for a specific profile and use the profile in the CI, that's what we have at xwiki for example: https://github.com/xwiki/xwiki-commons/blob/master/pom.xml#L2846

slarse:
[...]

PR #3864, opened 2021-03-26
review: chore: Remove snapshot repository from parent pom

slarse:
#3216 

See my latest comments in #3216. I don't see any reason to declare the snapshot repository for anything but distribution management. In Jenkins jobs, we inject the snapshot repository into pom files of other projects.

If we agree to merge this, I'll (after merge) go over all Jenkins jobs and make sure that they correctly pull the latest snapshot (I already know of several that pull from the old snapshot repository).

EDIT: Apparently, `spoon-smpl` depends on a SNAPSHOT version of `spoon-control-flow`, so it does need the snapshot repository enabled by default. But that's the concern of `spoon-smpl`, and not _all_ of the Spoon Maven modules, and so we define it only in `spoon-smpl`.

### DIFF
--- a/spoon-pom/pom.xml
+++ b/spoon-pom/pom.xml
@@ -25,14 +25,6 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     </properties>
 
-    <repositories>
-        <repository>
-            <id>ow2.org-snapshot</id>
-            <name>Maven Repository for Spoon Snapshots</name>
-            <url>https://repository.ow2.org/nexus/content/repositories/snapshots/</url>
-        </repository>
-    </repositories>
-
     <dependencies>
 
         <!-- we all depend on Junit with scope=test -->
--- a/spoon-smpl/pom.xml
+++ b/spoon-smpl/pom.xml
@@ -19,6 +19,18 @@
         <version>1.0</version>
         <relativePath>../spoon-pom</relativePath>
     </parent>
+
+    <repositories>
+        <repository>
+            <!-- Required for spoon-control-flow SNAPSHOT -->
+            <id>ow2.org-snapshot</id>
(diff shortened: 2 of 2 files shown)


monperrus:
My understanding is that we do this, we won't push snapshots to OW2 anymore during CD. We need to have it in a profile for everything. WDYT?

slarse:
> My understanding is that we do this, we won't push snapshots to OW2 anymore during CD. We need to have it in a profile for everything. WDYT?

I don't believe this is the case. We have the snapshot repository defined in the `distributionManagement` section, which should be all that's needed to _push_ snapshots: https://github.com/INRIA/spoon/blob/668a18b62d2a0f651e22f329fb181dfe1dc27918/spoon-pom/pom.xml#L92-L107

From the Maven reference (https://maven.apache.org/pom.html#repository):

> Whereas the **repositories element specifies in the POM the location and manner in which Maven may download** remote artifacts for use by the current project, **distributionManagement specifies where (and how) this project will get to a remote repository when it is deployed**. The repository elements will be used for snapshot distribution if the snapshotRepository is not defined.

Unless I'm completely misinterpreting the docs, there's no reason to define a repository under the `repository` element unless we intend to download artifacts from that repository.

monperrus:
Thanks for the additional info. LGTM.

monperrus:
Thanks @slarse

slarse:
Now I need to vet all of the Jenkins jobs to verify that an up-to-date Spoon snapshot is actually being pulled :)

monperrus:
excellent idea!

slarse:
Oh, and more relevant, it appears as if the snapshot deployment went fine: https://ci.inria.fr/sos/job/Spoon-Snapshot-Deployer/3403/console

monperrus:
cool!

PR #3873, opened 2021-04-09
chore: Add script for SNAPSHOT version injection in CI

slarse:
This PR adds a Python script that updates the Spoon version in a pom.xml file to the latest snapshot version. It does so by 1) injecting the snapshot repository into the pom file (if it's not already there), and 2) running `mvn versions:use-latest-versions` with snapshots enabled to update the version of Spoon, and 3) running `mvn dependency:purge-local-repository` to avoid using an old snapshot of the same version.

The script works with Python 3.5+ and requires the `defusedxml` package. We will use this for the smoke tests in the Jenkins pipeline where a project uses Spoon (i.e. Spoon is already a dependency), and we want to run its test suite with the latest snapshot.

### DIFF
--- a/doc/jenkins/inject_spoon_snapshot.py
+++ b/doc/jenkins/inject_spoon_snapshot.py
@@ -24,6 +24,7 @@ MAVEN_NAMESPACE = "http://maven.apache.org/POM/4.0.0"
 NAMESPACES = {"": MAVEN_NAMESPACE}
 
 MAVEN_VERSIONS_COMMAND = "mvn -B -U versions:use-latest-versions -DallowSnapshots -Dincludes=fr.inria.gforge.spoon".split()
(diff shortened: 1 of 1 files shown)


slarse:
Here's an example execution of the script: https://ci.inria.fr/sos/job/jperturb/1460/console

monperrus:
Thank @slarse

slarse:
Great! I've updated the following builds to use the snapshot injector script:

* AssertFixer
* Astor
* Casper
* cocospoon
* coming - later
* dspot
* gumtree-spoon-ast-diff
* jPerturb
* metamutator
* nopol
* npefix
* sorald
* spoon-control-flow
* spoon-decompiler
* spork
* juliac

The good news is that we don't appear to have any regressions (knock on wood, a few builds are still running). Which is pretty nice considering all of these builds were running with a months old 8.3.0-SNAPSHOT.

I will now start considering how to properly document the Jenkins builds. Right now, it's all somewhat arcane knowledge that I had to dig up by just looking through the configurations.

monperrus:
> The good news is that we don't appear to have any regression

yeah!

PR #3881, opened 2021-04-16
chore: Disable Git's autocrlf feature to fix Windows build failures

slarse:
Many of the build failures on Windows are due to `core.autocrlf` begin active in Git, which on that platform automatically converts the line endings in all files to CRLF on checkout and back to LF on check-in. This causes all tests with hard-coded LF line endings to fail.

Why would we want hard-coded LF line endings in tests? Well, for example, in the sniper printer tests, we want to make sure that the sniper printer uses the line endings contained in the source file, regardless of platform. Thus, if there are LF line endings in the input, there should be LF line endings in the output, which simply cannot be tested if Git automatically converts these line endings.

This PR fixes 40 test failures on Windows, leaving 18 failures and 3 errors. Most of the remaining failures seem to be related to incorrect use of hard-coded line endings.


slarse:
Apparently, this wasn't the entire solution, but there are significantly fewer test failures now than before. Notably, none of the sniper printer tests fail now!

slarse:
Suddenly Travis checks have started running again. @monperrus did they get back to you about the free project thing? GHA is working very well for us, so I don't see much value in putting in the effort of going back to Travis at the moment.

monperrus:
> @monperrus did they get back to you about the free project thing?

Yes, quoting their answer coming 4 months after our request (sig):

> Please accept our sincere apologies for any inconvenience caused due to delays in processing your open-source credit request while we made improvements to the platform.  Open source has always been and always will be at the core of what Travis CI stands for. After reviewing your open source qualifications, I am pleased to notify you that 25k credits were added to your account allowing you to resume running OSS builds.

monperrus:
> GHA is working very well for us

Agree.

slarse:
GHA currently having trouble, we'll have to just wait a bit: https://www.githubstatus.com/

monperrus:
neat!

slarse:
Now ready for review again. Fixes most of the Windows build errors.

The remaining failures are also mostly problems with line endings, but are probably due to incorrect usage. For example, if we pretty-print (non-sniper) an element, we should expect CRLF line endings on Windows. Fixing those is outside the scope of this PR.

monperrus:
Thanks a lot @slarse

Issue #3836, opened 2021-03-11
Remove/migrate all remaining Travis stuff


slarse:
As we're unlikely to ever return to Travis CI, references to Travis are confusing, e.g. the directory containing the CI scripts being called `chore/travis` and the `.travis.yml` config file being present. We should remove anything that is no longer necessary, and rename everything else appropriately.

In the process of removing the Travis CI traces, we should also ensure that we've migrated all functionality that we want over to GitHub Actions.

Known other problems:

* The "external CI" scripts are not being pulled as they depend on Travis environment variables: https://github.com/INRIA/spoon/blob/4880da00459c28f920d66cb0f5630c8427fb57c3/chore/travis/ci-extra.sh#L101-L105

KishkinJ10:
hello sir , i am a beginner and i am participating in the GSoC 2021 with Project Spoon , sir can you please brief me what are we going to do in this project ?? and how can i write timeline section in my proposal?

slarse:
@KishkinJ10 sorry, I missed this comment. Hope you found your way to the GSoC issue #3828.

Closing this as #3834 removed the final traces of Travis.

PR #3883, opened 2021-04-19
chore: Remove all traces of Travis CI

slarse:
#3836 

This PR removes all traces of Travis CI.

Note "removes", and not "migrates". Two pieces of functionality is lost here, both executing external scripts. This adds significant complexity to the CI pipeline, without any tangible benefit.

First, the final portion `chore/ci-extra.sh`, which is supposed to execute an external CI script on pull requests, is removed. It executed a script that [there never appears to have contained anything](https://github.com/SpoonLabs/spoon-ci-external/blob/master/spoon-pull-request.sh). Therefore, it's hard to justify keeping that.

Second, this PR also removes the `chore/travis/ci-master.sh` script which contained the following line:

```
 curl -s -X POST -H "Content-Type: application/json" -H "Accept: application/json" -H "Travis-API-Version: 3" -H "Authorization: token $TRAVIS_TOKEN" https://api.travis-ci.com/repo/SpoonLabs%2Fspoon-ci-external/requests
```

I believe this was used to trigger the upload of surefire reports to [spoon-ci-data](https://github.com/spoonlabs/spoon-ci-data). I'm also uncertain of the value of this functionality. It hasn't been running for 5 months, and no one seems to have missed having it. In any case, documenting its removal here so we can reimplement it if desired.

If we want these two removed pieces of functionality to be reimplemented, then #3836 is still relevant. If we don't (which I argue for, we have enough complexity with the CI as it is), then this PR concludes #3836.


slarse:
OOPS, hold on a second, missed the badge in the README.

chore/ci-extra.sh
```diff
@@ -93,13 +93,3 @@ git diff
...
 mvn -q depclean:depclean
-
-##################################################################
-## Trigerring extra tasks that we don't want to commit to master
-## (For experimental CI features, short lived tasks, etc)
-
-if [[ "$TRAVIS_REPO_SLUG" == "INRIA/spoon" ]] && [[ "$TRAVIS_PULL_REQUEST" != "false" ]]
-then
-  echo "downloading extra CI PR script from SpoonLabs/spoon-ci-external"
-  curl https://raw.githubusercontent.com/SpoonLabs/spoon-ci-external/master/spoon-pull-request.sh | bash
```
monperrus, line 104: we may keep that, it's cool for experimenting. How to update the if condition in GHA?

monperrus (commented):
Thanks @slarse LGTM, see comment.

chore/ci-extra.sh
slarse: Re-added, you can see here that the "downloading extra CI PR..." is now being executed: https://github.com/INRIA/spoon/runs/2387878660#step:8:3814

monperrus:
Thanks a lot @slarse

monperrus:
Good bye @travis-ci! It was great to work with you, you helped Spoon very much over several years.

Issue #3884, opened 2021-04-20
Add missing addXAt(int, CtElement<?>)-like methods for ordered collections to public API


slarse:
In the diff minimization project currently headed by @algomaster99, we have a need to make insertions at specific indices in list-like collections. Many Spoon metamodel elements support this through the `addX(int, CtELement)` methods (such as `CtStatementList.addStatement(int, CtStatement)`), but `CtInvocation` curiously only has a `addArgument(CtExpression<?>)`.

I propose adding `CtAbstractInvocation.addArgument(int, CtExpression<?>)` to the public API, because it a) makes for a more consistent experience in using Spoon, and b) this functionality is already implemented in all implementations of `CtAbstractInvocation`, so the extra work is minimal.

Currently, the only precise workaround that we have is this ugly beast that draws on implementation details of Spoon, and is thus undesirable to me:

```java
List<CtExpression> arguments = invocation.getArguments();

if (arguments.isEmpty()) {
    // empty model lists may be `EmptyClearableList` instances, which we can't add to
    invocation.addArgument(argument);
} else {
    // non-empty model lists are always `ModelList` instances, which update the metamodel on mutation
    arguments.add(index, argument);
}
```

### Progress

- [x] `CtAbstractInvocation.addArgumentAt(int, CtExpression)`
- [x] `CtFormalTypeDeclarer.addFormalCtTypeParameterAt(int, CtFormalTypeParameter)` (#3894) - @slarse 
- [x] `CtExecutable.addParameterAt(int, CtParameter)` (#3895) - @algomaster99

monperrus:
Great idea.

slarse:
Actually, by extension, this should apply to _all_ ordered collections in the Spoon metamodel, for which it makes sense to expose an `addX(int, CtElement)` without needlessly exposing internals. This should at the very least apply to all collections that are ordered by design in the Java Language Specification, and it would make sense to apply this treatment everywhere there is a `setXs(List)` method (which implies the underlying collection is a list).

After a very quick browse, I "found" several more missing methods:

* `CtExecutable.addParameter(int, CtParameter)`
* `CtExecutable.addThrownType(int, CtTypeReference)`
* `CtFormalTypeDeclarer.addFormalCtTypeParameter(int, CtTypeParameter)`

[...]

PR #3885, opened 2021-04-21
feat: Add CtAbstractInvocation.addArgumentAt method

slarse:
#3884

This PR adds the method `CtAbstractInvocation.addArgumentAt(int, CtExpression<?>)` to the public API, as discussed in #3884.

This PR will determine whether we go with the `addX` naming convention, or the `addXAt` naming convention. Both are already present in the project, but we should adopt one as the predominant style. I favor `addXAt(int,CtElement)` because it makes it easier to see that the method is distinct, but `addX(int,CtElement)` is also valid as that is a common overload in the Java standard library collections.


monperrus:
useful new API method! thanks @slarse

PR #3887, opened 2021-04-23
chore: Bump license-maven-plugin to version 4.0

slarse:
This PR does the same thing as #3868 in bumping the version of `license-maven-plugin`, but also fixes the errors. There are two fixes, apart from bumping the version:

1. Wrap the configuration in the `<licenseSets><licenseSet>` tags, which is mandatory for version 4.0
	- We might want to look at the new features for handling multiple licenses, that's what this is for! I think that's for a different PR, though.
2. Adopt the new default format (slashstar, header starts with `/*` as opposed to Javadoc `/**`).
    - I prefer this format, as linters such as the one in IntelliJ IDEA complain about "dangling Javadoc comment" when the Javadoc style is used for a header.

### DIFF
--- a/spoon-control-flow/pom.xml
+++ b/spoon-control-flow/pom.xml
@@ -59,11 +59,15 @@
                     <groupId>com.mycila</groupId>
                     <artifactId>license-maven-plugin</artifactId>
                     <configuration>
-                        <header>LICENSE.txt</header>
-                        <basedir>src</basedir>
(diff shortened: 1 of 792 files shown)


monperrus:
LGTM, will merge!

> We might want to look at the new features for handling multiple licenses, that's what this is for! I think that's for a different PR, though.

Agree.

Thanks!

PR #3896, opened 2021-04-29
feat: Add CtFormalTypeDeclarer.addFormalCtTypeParameterAt

slarse:
#3884 

This PR adds `CtFormalTypeDeclarer.addFormalCtTypeParameterAt(int, CtTypeParameter)`.

~I'm not happy with the current implementation, as the code for adding a type parameter in a specific position is duplicated across all implementations of the method. I'm trying out some ideas to reduce this code duplication.~

~This seems to be a problem for most of the `CtFormalTypeDeclarer` methods. Since Java doesn't have any kind of support for mixin classes, extracting the functionality is tricky.~

We'll deal with the code duplication separately.

~This is ready for review!~ ~Actually no it isn't saw a typo ...~ Ok, now it's ready for review!


src/test/java/spoon/test/method/MethodTest.java
```diff
@@ -29,11 +30,16 @@
 import spoon.test.method.testclasses.Tacos;
 
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
```
algomaster99, line 34: This import statement is unnecessary.

slarse: indeed it is, good catch

monperrus:
Thanks a lot @slarse

PR #3897, opened 2021-04-30
fix: fix snippet compilation bugs

### LINKED ISSUE #3898: Problems with compiling code snippets
There are two problems.

1. It is not possible to compile snippets added to a class in an unnamed package.

2. If a snippet only contains a comment it can produce the wrong code when it is compiled.

There is a failing test case for each problem in #3897. I'll try to fix this.

andrewbwogi:
closes #3898


slarse:
@andrewbwogi FYI, to trigger the CI you can just close and re-open the PR. If you require a new commit, you can push an empty commit like so:

```bash
$ git commit --allow-empty -m 'Empty commit to trigger CI'
$ git push
```

andrewbwogi:
@slarse, good tip!

monperrus:
Thanks @andrewbwogi LGTM.

Snippets were originally made only for quick pretty-printing of anything. On the fly snippet compilation was added afterwards and is indeed hard to get right in all situations.

monperrus:
Thanks @andrewbwogi !

Issue #3901, opened 2021-05-02
Prevent logging?


gchaperon:
Is it currently possible to prevent logging completely?

I am looking at the source and it seems to me that it is not possible, since only valid names defined in [`spoon.support.Level`](https://github.com/INRIA/spoon/blob/fecec810886819c09ffc4955faeb4f50f64e07a3/src/main/java/spoon/support/Level.java#L14) are allowd here
https://github.com/INRIA/spoon/blob/fecec810886819c09ffc4955faeb4f50f64e07a3/src/main/java/spoon/support/StandardEnvironment.java#L190

Please let me know if I have overlooked something and it is actually possible.

If it is not possible, how would you feel about adding `OFF(0)` to `Level`. Would that allow to turn logging completely off?

Best,
Gabriel.

slarse:
Hi @gchaperon,

I see no reason not to offer that. I'd say it's rather unexpected to find that you can set logging levels, but none of them allow you to disable logging entirely.

Care to do a PR? :)

monperrus:
In theory, the logging can be entirely disabled via slf4j (by configuring logback in logback.xml)

slarse:
> In theory, the logging can be entirely disabled via slf4j (by configuring logback in logback.xml)

Ah, yes, that is true. But as we have an API for setting the logging level, wouldn't it also make sense to add an `OFF` level? It wouldn't be much effort at all, just adding the level to `Level` and then patching up the switch statement that deals with logging in `StandardEnvironment`.

monperrus:
> But as we have an API for setting the logging level, wouldn't it also make sense to add an OFF level?

yes it would!

slarse:
> yes it would!

Let's put it on the TODO list then :)

@gchaperon Would you like to try your hand at a PR for this feature?

gchaperon:
hi, ya sure, i can try
looking more in detail I have a question.
i see that [this](https://github.com/INRIA/spoon/blob/fecec810886819c09ffc4955faeb4f50f64e07a3/src/main/java/spoon/support/StandardEnvironment.java#L276) method is only used internally, so the `setLevel` config on `StandardEnvironment` only applies to the logging produced by that class. I think this is somewhat confusing for the user, shouldn't be all logging managed by slf4j?

PR #3904, opened 2021-05-03
WIP: feature: Add OFF log level to StandardEnvironment

gchaperon:
As discussed in  #3901, this PR adds a new logging level `OFF(0)` and modifies the printing logic in `StandardEnvironment` so that it is turned off if this new level is set.

I tagged this as a feature following the tag of the issue.
The contributing guidelines state that feature PR should have tests and corresponding documentation. I though this PR is maybe to small for that, but please let me know if you would like tests added and what should they cover.

### DIFF
--- a/src/main/java/spoon/support/StandardEnvironment.java
+++ b/src/main/java/spoon/support/StandardEnvironment.java
@@ -274,7 +274,7 @@ public class StandardEnvironment implements Serializable, Environment {
 	}
 
 	private void print(String message, Level messageLevel) {
-		if (messageLevel.toInt() <= this.level.toInt()) {
+		if (this.level.compareTo(Level.OFF) > 0 && messageLevel.toInt() <= this.level.toInt()) {
 			switch (messageLevel) {
 				case ERROR: logger.error(message);
 					break;
--- a/src/main/java/spoon/support/Level.java
+++ b/src/main/java/spoon/support/Level.java
@@ -11,7 +11,7 @@ package spoon.support;
  * Enum for representing logging levels.
  */
 public enum Level {
-	ERROR(100), WARN(200), INFO(300), DEBUG(400), TRACE(500);
+	OFF(0), ERROR(100), WARN(200), INFO(300), DEBUG(400), TRACE(500);
 
 	private final int levelValue;
 


src/main/java/spoon/support/StandardEnvironment.java
```diff
@@ -274,7 +274,7 @@ public void report(Processor<?> processor, Level level, String message) {
 	}
 
 	private void print(String message, Level messageLevel) {
-		if (messageLevel.toInt() <= this.level.toInt()) {
+		if (this.level.compareTo(Level.OFF) > 0 && messageLevel.toInt() <= this.level.toInt()) {
```
slarse, line 277: Perhaps this is more clear?

```suggestion
		if (this.level != Level.OFF && messageLevel.toInt() <= this.level.toInt()) {
```

I think it's a bit easier to read as it doesn't require one to go check the declaration order of the `Level` values. This is probably a matter of preference, though, if you prefer your solution feel free to keep it.

slarse (changes requested):
Thanks for the PR @gchaperon!

No behavioral change is too small for a test case :). See `spoon.test.logging.LogTest`. It shouldn't be too hard to verify that no message gets logged for any logging level if the level is set to `OFF`. You can draw inspiration from the `testAllLevelsForLogs` test case (see how it captures the logging output).

src/main/java/spoon/support/StandardEnvironment.java
slarse: Note that there's a tiny behavioral difference: `level != Level.OFF` will cause the if-block not to be entered even if the passed level is `OFF`, whereas your original solution will cause it to be entered (but do nothing as no case in the switch matches)

gchaperon: sure, that's better, I didn't know enums could compare like that (that's how little java I know 😅)

gchaperon:
hi!

I changed the line following your advice and added a test.
I have no idea _whatsoever_ as to where to put the test. I saw there was a class for nonparameterized tests, so I added it there. 
Also, since `print` is private I used (what I think is) the closest proxy for it, which is `report`. This method receives a `Processor` but does not use it, so I used whatever `Processor` implementation was the easiest to instantiate.

Let me know what you think.

src/test/java/spoon/test/logging/LogTest.java
```diff
@@ -87,5 +91,24 @@ public void testMavenLauncherLogs() {
 			mavenLauncher = new MavenLauncher("./pom.xml", MavenLauncher.SOURCE_TYPE.APP_SOURCE, new String[]{"./"});
 			assertEquals("Running in FULLCLASSPATH mode. Classpath is manually set (doc: http://spoon.gforge.inria.fr/launcher.html).",logger.getLoggingEvents().get(0).getMessage());
 		}
+
+		@Test
+		public void testLoggingOff() {
+			final TestLogger logger = TestLoggerFactory.getTestLogger(Launcher.class);
```
slarse, line 97: Could you add a contract as a comment at the top of the test method? I.e. a comment stating what the test should do along, like so (ofc replacing my nonsense text):

```suggestion
		@Test
		public void testLoggingOff() {
			// contract: The logger should do X given Y and perhaps Z
			final TestLogger logger = TestLoggerFactory.getTestLogger(Launcher.class);
```

slarse (changes requested):
I think this looks very good! The placement of the test is correct. The old JUnit4 parameterization is awkward because it parameterizes the entire class, which is why a nested class is needed for the non-parameterized tests.

JUnit5 parameterization is far less confusing (very easy to apply to just a single test).

Anyway, I just had a single comment in that we typically try to put a contract at the top of the test, stating what the test should do. See for example `testMavenLauncherLogs` (and my comment).

gchaperon:
there, I just added a contract

src/test/java/spoon/test/logging/LogTest.java
```diff
@@ -22,6 +22,9 @@
 import org.junit.runners.Parameterized;
 import spoon.Launcher;
 import spoon.MavenLauncher;
+import spoon.processing.Processor;
+import spoon.reflect.declaration.CtElement;
```
slarse, line 26: These two imports seem redundant.

```suggestion
```

slarse (changes requested):
Just one final thing, there are a couple of redundant imports in `LogTest`, probably added automatically by your IDE when you were trying stuff out. I missed them before, sorry!

After that I'm satisfied and we ping an integrator :)

gchaperon:
That is exactly what happened, thanks for the notice.

monperrus:
Thanks a lot @gchaperon for your first contribution in Spoon.

PR #3905, opened 2021-05-04
fix: Fix infinite loop on type parameter in qualified name

### LINKED ISSUE #3903: Infinite loop in buildModel
Hi!
I am trying to process a file but the library gets stuck when trying to build the model.
The file that I am trying to process is [this one](https://pastebin.com/dmkqHTbN)

The code I am using is below:
```java
import spoon.FluentLauncher;
import spoon.reflect.CtModel;

public class TestBuildModelStuck {
		String path = "/path/to/myfile.java";
		CtModel model = new FluentLauncher().inputResource(path).buildModel();

	}
}
```

Using the debugger I managed to see that buildModel is stuck in an infinite loop here:
https://github.com/INRIA/spoon/blob/fecec810886819c09ffc4955faeb4f50f64e07a3/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java#L513
`off` never goes below zero

The part of my file it gets stuck on is `T.Entry<K, V>`, found in line 508 of my file.

I am failing to see why this is happening. Is that syntax not allowed or invalid?

slarse:
Fix #3903 

This PR fixes the infinite loop described in #3903 by assuming that any qualified name that contains a type parameter reference is fully qualified. I'm pretty sure that assumption is valid, as one can only reference a type parameter from within the class or method that defines it.

In any case, it's better than an infinite loop :)

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java
+++ b/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java
@@ -64,6 +64,7 @@ import spoon.reflect.reference.CtModuleReference;
 import spoon.reflect.reference.CtPackageReference;
 import spoon.reflect.reference.CtParameterReference;
 import spoon.reflect.reference.CtReference;
+import spoon.reflect.reference.CtTypeParameterReference;
 import spoon.reflect.reference.CtTypeReference;
 import spoon.reflect.reference.CtVariableReference;
 import spoon.support.reflect.CtExtendedModifier;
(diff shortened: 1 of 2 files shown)


monperrus:
Nice fix! This is heavily tested, so I will merge with confidence.

monperrus:
Thanks @slarse for the bug and the fix!

slarse:
Thanks to me for the fix, thanks to @gchaperon for reporting the bug along with an easily reproducible scenario!

PR #3907, opened 2021-05-04
chore: Add CI check for method/constructor Javadoc quality

### LINKED ISSUE #3906: Enforce Javadoc comment quality
Poking around in the metamodel interfaces, I've noticed that many of them lack proper documentation of parameters and return types.

I think we should enforce that public methods and constructors are properly documented. Checkstyle can do this with the `JavadocMethod` module. The problem is that it raises ~2600 errors when activated on the public scope alone, so obviously this needs to be introduced gradually. I'm thinking we can add a script that computes a penalty (the amount of errors), and if the penalty increases in relation to the master branch, we break the build. This doesn't entirely guard against poorly documented methods as one could e.g. remove an error from one javadoc and introduce a new one somewhere else, but it's pretty good and pretty easy to implement.

Good docs are important to me, so I'm gonna give this a shot. Decreasing the penalty can then be a great first issue, much like #3854.

slarse:
Fix #3906 

This PR adds a script `chore/ci-checkstyle-javadoc.sh` which runs Checkstyle on the project, looking for Javadoc style problems on public methods/constructors. It counts the amount of Checkstyle errors on the current branch, and on the master branch, and fails the build if the current branch has more errors.

The quality check can be seen here: https://github.com/INRIA/spoon/pull/3907/checks#step:9:10

Here is the checkstyle rule used [JavadocMethod](https://checkstyle.org/config_javadoc.html#JavadocMethod). We're using default settings with the exception that we only check public methods/constructors (default is to check all visibilities).


monperrus:
Great infrastructure contribution! Thanks @slarse

PR #3935, opened 2021-05-20
chore: Add dependabot.yml

### LINKED ISSUE #3933: Setup GitHub-native Dependabot
Since a while back, Dependabot has been integrated into GitHub and can be activated by adding a `dependabot.yml` file to the `.github` directory. As far as I can tell, Spoon still uses the Dependabot Preview GitHub App, which will be [removed as of August 23d 2021](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/upgrading-from-dependabotcom-to-github-native-dependabot)

I suggest we upgrade to GitHub-native Dependabot, and in doing so we also **configure it to upgrade GitHub Actions**. The very neat thing is that nowadays, Dependabot is clever enough to analyze commit shas, so we can have the safety of pinned actions but still get automatic updates. I just added this in Spork, see for example KTH/Spork#190.

There is one slight problem, namely that Dependabot does not specify the actual pinned version in a comment (if you have a look at https://github.com/KTH/spork/pull/190/files, you'll see that it changes the commit sha but leaves the version comment in place). If we trust Dependabot to keep the versions up-to-date, I don't think that's too much of a problem, and it's easy to just by manual inspection find the tag that corresponds to a particular commit sha (of course assuming no one has tampered with the tag, which is why we use commit shas in the first place :).

slarse:
Fix #3933 

This PR adds a configuration for [GitHub-native Dependabot](https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/), which replaces the Dependabot Preview application.

Dependabot is supposed to support [Multi-module Maven projects](https://dependabot.com/blog/multimodule-maven-files/) by pointing it at the parent pom. So, if I'm not mistaken, we should be able to get updates for spoon-core and all subprojects by pointing dependabot to the `/spoon-pom/pom.xml` file.

@monperrus any reason not to try that?

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+version: 2
+updates:
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+
+  - package-ecosystem: "maven"
+    directory: "/"
+    schedule:
+      interval: "daily"


monperrus:
In Spoon, we use @renovate-bot, made in Stockholm with :heart: by our friend @rarkins

In theory, everything done by Dependabot is also done by Renovate.

@slarse, do you see a feature in Dependabot that we miss?

slarse:
Ah, my mistake, it's almost as if I should have  known that :D

But yes, renovate cannot update workflow files unless [provided with a personal access token](https://docs.renovatebot.com/modules/manager/github-actions/). That's a hassle we probably don't want. It's also unclear whether they support updating actions pinned to commit shas, rather than version tags.

How about using Dependabot only for updating our GHA workflow files?

rarkins:
That documentation is out of date. GitHub lifted the restriction against apps updating actions.

slarse:
@rarkins Ah, good! Does renovate support updating actions that are pinned by commit sha?

rarkins:
Yes

Issue #3919, opened 2021-05-11
Migrate test suite to JUnit5


slarse:
We currently have a little bit of JUnit5 (such as [VariableTest](https://github.com/INRIA/spoon/blob/639dccff96b2818b64f2d9af61339445aab77d18/src/test/java/spoon/test/variable/VariableTest.java)), but mostly JUnit4 in the test suite. I think it's worth our while to start migrating entirely to JUnit5. Off the top of my head, here are some benefits that come with JUnit5:

* Much better support for [parameterized tests](https://www.baeldung.com/parameterized-tests-junit-5)
* Better support for [temporary directories](https://www.baeldung.com/junit-5-temporary-directory)
* Better support for [conditional test execution](https://junit.org/junit5/docs/5.5.0/api/org/junit/jupiter/api/condition/package-summary.html)
* ~The awesome [`assertThrows` assertion](https://www.baeldung.com/junit-assert-exception)~
    - Edit: Actually, this was backported to JUnit 4.13, so it's a moot point

This is also very easy to do incrementally, one test class at a time, making it a **very good first issue**.

### Why not keep both?

While we can keep both JUnit4 and JUnit5, and one can even write test cases of both flavors in the same test class, it's just messy to do so due to all of the assertions and annotations having the same simple names. You can also be pretty confident that if a contributor comes into a test class and sees _only_ JUnit4 tests, that contributor will write JUnit4 tests as well. The longer we wait, the more work it will be to leave JUnit4 behind!

### How do I know if a test class is using JUnit4?

The easiest way to tell is to look for imports from the JUnit4 API, the most telling one being `import org.junit.Test`. At the moment of writing, there are 189 such imports in the test suite.

```bash
$ grep -r 'import org.junit.Test' src/test/java | wc -l
189
```

### How do I know if a test class is using JUnit5?

The easiest way to tell is to look for imports from the JUnit5 API, the most telling one being `import org.junit.jupiter.api.Test`. At the moment of writing, there are 3 such imports in the test suite.

```bash
$ grep -r 'import org.junit.jupiter.api.Test' src/test/java | wc -l
3
```

### How can I convert from JUnit4 to JUnit5?

See #3921 for an example.

[...]

PR #3939, opened 2021-05-21
test: Migrate LogTest to JUnit5

slarse:
#3919 

The `LogTest` has long been a pet-peeve of mine, because it _always fails_ when I run all tests in IntelliJ IDEA. Running it alone is fine, and running it with Maven is also fine. I think it's something with the JUnit4 parameterized test runner (which never made it past the experimental stage) that doesn't work well with my particular version of IDEA.

This PR migrates LogTest to JUnit5 and its much better parameterized test support. And finally I can run all tests and not have this one fail.

**Ignore whitespace changes in diff:** This PR has a _lot_ of indentation changes. To view the diff without whitespace changes, append `?w=1` to the GitHub URL (or click this link: https://github.com/INRIA/spoon/pull/3939/files?w=1)

### DIFF
--- a/src/test/java/spoon/test/logging/LogTest.java
+++ b/src/test/java/spoon/test/logging/LogTest.java
@@ -16,10 +16,10 @@
  */
 package spoon.test.logging;
 
-import org.junit.Test;
-import org.junit.experimental.runners.Enclosed;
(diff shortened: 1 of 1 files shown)


monperrus:
LGTM, will merge!

PR #3941, opened 2021-05-21
fix: Ensure that for init/update are registered in listeners

### LINKED ISSUE #3940: Sniper: Deletion in for-loop header is not detected by `ChangeCollector` (and thus not printed correctly)
So, when deleting an element in a for-loop header, the deletion isn't registered in the `ChangeCollector` because no listener callback is called, and therefore the loop isn't printed correctly. For example, given this for loop:

```java
for (int i = 0; i < 10; i++) {
}
```

deleting the update `i++` correctly removes it from the model, but the sniper still prints the exact same thing as the input.

slarse:
Fix #3940 

The `setForInit` and `setForUpdate` does not call change listeners if the value set was the empty list or null. Note that deleting the last element in a e.g. a for init is implemented such that `setForInit` is called with an empty list, and so the last deletion is not registered. This causes problems for the sniper printer, as a patch that only deletes the final element in a for update or init does not register any changes, and so the sniper just prints the original source code.

This PR simply moves the `onDeletAll()` call to the change listeners to the top of the `setForInit` and `setForUpdate` methods, such that they are always called. This is valid, as setting an empty list or `null` is equivalent to clearing the list.

### DIFF
--- a/src/main/java/spoon/support/reflect/code/CtForImpl.java
+++ b/src/main/java/spoon/support/reflect/code/CtForImpl.java
@@ -76,11 +76,11 @@ public class CtForImpl extends CtLoopImpl implements CtFor {
 
 	@Override
 	public <T extends CtFor> T setForInit(List<CtStatement> statements) {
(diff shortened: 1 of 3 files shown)


monperrus:
Thanks @slarse

PR #3945, opened 2021-05-23
review: fix additional snippet bugs

andrewbwogi:


monperrus:
Hi @andrewbwogi 

Moving the exception in another package is a breaking change. Why would we need to do that?

Thanks,

andrewbwogi:
@monperrus, `SnippetCompilationError` just went along with the package change of `SnippetCompilationHelper`, I've placed it back again. Is moving `SnippetCompilationHelper` also a breaking change that is not worth the trouble? It was moved to use package-private methods in `PositionBuilder`.

monperrus:
> Is moving `SnippetCompilationHelper` also a breaking change that is not worth the trouble?

Yes it would be breaking if there is at least one public method in the class.

> It was moved to use package-private methods in `PositionBuilder`.

One solution is to make them public instead.

andrewbwogi:
I somehow did not see that this branch was already merged (#3897). I'm fine with it so I'm closing this one.

PR #3947, opened 2021-05-24
feat: add shortcut API to get or create the compilation unit

### LINKED ISSUE #3942: Add `CtElement.getCompilationUnit` method to the public API
Currently, we have to retrieve the `CtCompilationUnit` of an element, `CtType` in my particular case, using the factory methods defined.
```java
element.getFactory().CompilationUnit().getOrCreate((CtType<?>) element);
```

I propose we add a public API to directly get the `CtCompilationUnit`. It can be a part of the `CtElement` as each element belongs to some compilation unit.

Reference: https://github.com/SpoonLabs/diffmin/pull/65#discussion_r636910498

algomaster99:
Fixes #3942 

The changes add `getOrCreateCompilationUnit()` API which would serve as a shortcut to get the compilation unit. This API needs to be tested with the following elements.

- [ ] CtType
   - [x] CtClass
   - [ ] CtInterface 
- [ ] CtModule
- [ ] CtPackage

### DIFF
--- a/src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
+++ b/src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
@@ -9,6 +9,7 @@ package spoon.support.reflect.declaration;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import spoon.SpoonException;
 import spoon.reflect.ModelElementContainerDefaultCapacities;
 import spoon.reflect.annotations.MetamodelPropertyField;
 import spoon.reflect.code.CtBlock;
@@ -18,10 +19,13 @@ import spoon.reflect.cu.SourcePosition;
 import spoon.reflect.declaration.CtAnnotation;
 import spoon.reflect.declaration.CtCompilationUnit;
 import spoon.reflect.declaration.CtElement;
+import spoon.reflect.declaration.CtModule;
 import spoon.reflect.declaration.CtNamedElement;
+import spoon.reflect.declaration.CtPackage;
 import spoon.reflect.declaration.CtShadowable;
 import spoon.reflect.declaration.CtType;
 import spoon.reflect.declaration.ParentNotInitializedException;
+import spoon.reflect.factory.CompilationUnitFactory;
 import spoon.reflect.factory.Factory;
 import spoon.reflect.factory.FactoryImpl;
 import spoon.reflect.meta.RoleHandler;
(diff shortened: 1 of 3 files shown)


src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
```diff
@@ -186,6 +190,24 @@ public String getDocComment() {
...
+		} else if (this instanceof CtPackage) {
+			return cuFactory.getOrCreate((CtPackage) this);
+		} else if (this instanceof CtModule) {
+			return cuFactory.getOrCreate((CtModule) this);
+		} else {
+			CtType<?> enclosingType = getParent(CtType.class);
+			if (enclosingType != null) {
+				return cuFactory.getOrCreate(enclosingType);
+			} else {
+				throw new SpoonException(this.getShortRepresentation() + " does not belong to a compilation unit");
```
algomaster99, line 206: Does using `this.getShortRepresentation()` better the error message in some way? If not, I am willing to replace it with "The element".

src/test/java/spoon/test/ctType/CtTypeTest.java
```diff
@@ -267,4 +268,21 @@ public void testTypeDeclarationToReferenceRoundTripInNamedModule() {
...
+
+	@Test
+	public void testGetOrCreateCompilationUnit() {
+		// contract: the compilation unit should be returned using the shortcut method
+		final Launcher launcher = new Launcher();
+		final Factory factory = launcher.getFactory();
+
+		// create the elements required
+		CtCompilationUnit expectedCu = factory.createCompilationUnit();
+		CtClass<?> klass = factory.createClass("CompilationUnit");
```
algomaster99, line 280: Not sure which is the best way here. I could either create a test file and add it to the launcher or just use factory methods. I think both will do a similar task but the latter was shorter to write.

algomaster99: I tested for a `CtClass` instance in `CtType`. Is this alright? Or should I move this test case to `CtClassTest`? Moroever, do we also need to test for `CtInterface` because they are also `CtType` implementations?

slarse:
Given the discussion in #3942, and the realization that the current implementation of compilation units in Spoon is _very_ tied into source positions, I'm not entirely convinced we should add this method right now, as its behavior might change if we manage to untangle compilation units from source positions.

algomaster99:
@slarse oh, okay! Sounds good. We can close this for now and reopen it later when we are able remove the CU's dependency on position in factory methods.

Had to ask you one more thing which I came across while working on this PR. Should we include `CtModel.getPackage(String qualfiedName)`? Currently, one can either get the root package or the collection of all packages. I was debugging something in the testcase in this PR before so I felt I could have used such an API. Let me know if it is beneficial in some other scenarios too.

slarse:
> @slarse oh, okay! Sounds good. We can close this for now and reopen it later when we are able remove the CU's dependency on position in factory methods.

I'm not certain that's possible, or how to do it. I will discuss it with @monperrus this week and get back to you.

> Had to ask you one more thing which I came across while working on this PR. Should we include `CtModel.getPackage(String qualfiedName)`? Currently, one can either get the root package or the collection of all packages. I was debugging something in the testcase in this PR before so I felt I could have used such an API. Let me know if it is beneficial in some other scenarios too.

No, because that makes working with modules confusing. `CtModel.getRootPackage()` is a common source of errors (both inside of Spoon itself and in clients), because its name makes it sound like it returns a unique root package for a given model, but it in fact returns the root package of the _unnamed module_. In multi-module projects, there is no one unique root package, each module has its own. Thus, it's easy to make programmatic errors with the assumption that each Java project has a unique root package, which is not the case since modules made their entrance.

`CtModule.getPackage(String qualifiedName)` might make sense.

PR #3959, opened 2021-05-26
test: Migrate ProcessingTest to Juit5

Rohitesh-Kumar-Jain:
#3919

This PR just migrates ProcessingTest to Junit5, pure refactoring.

Also, I have replaced the string message in assertArrayEquals with a supplier (I believe this is part of refactoring right?)

### DIFF
--- a/src/test/java/spoon/processing/ProcessingTest.java
+++ b/src/test/java/spoon/processing/ProcessingTest.java
@@ -153,7 +153,7 @@ public class ProcessingTest {
 		l.run();
 
 		// If template is applied to itself then there will be modified spoon/...Template.java on output
-		assertArrayEquals(new String[]{"SimpleAssert.java"}, outputPath.toFile().list(), () -> "Template source found in output");
+		assertArrayEquals(new String[]{"SimpleAssert.java"}, outputPath.toFile().list(),"Template source found in output");
 		// Check that the template worked as intended
 		assertThat(outputPath.toString() + "/SimpleAssert.java")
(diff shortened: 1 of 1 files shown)


Rohitesh-Kumar-Jain:
One of the major breaking changes for the Assertions is that the order of the parameters changed with the output message parameter moved as the last parameter. 

And now as we need to change that order, we can replace the String messages with Supplier.

slarse:
@Rohitesh-Kumar-Jain I don't see a good reason to replace the messages with suppliers. Can you justify that change?

Rohitesh-Kumar-Jain:
> @Rohitesh-Kumar-Jain I don't see a good reason to replace the messages with suppliers. Can you justify that change?

I read in a guide that in case we want to enhance our test methods during the migration, we can replace the String messages with Supplier.

Screenshot 2021-05-26 at 2 44 21 PM

[Here's](https://www.baeldung.com/junit-5-migration) the link to the resource, I downloaded it from there only.

slarse:
I have a very hard time seeing the point of that, it just seems like more code for the same result to me. I don't see how it's an enhancement, and the article you linked does not say.

In general, for these migrations, do what you need to do to make it work, and nothing more :)

Rohitesh-Kumar-Jain:
I will replace the supplier with the comment in the next commit

slarse (approved):

slarse:
@monperrus ping

monperrus:
thanks @Rohitesh-Kumar-Jain

Issue #3965, opened 2021-05-28
Refactor `ReferenceBuilder.getTypeReference(TypeBinding, boolean)`


slarse:
I'm currently trying to fix a bug in `ReferenceBuilder.getTypeReference(TypeBinding, boolean)` related to #3951, and it's just too complex a method. It's 250 lines long and incredibly complicated. I need to refactor the method before I can fix the bug, or I'll just make it even worse.

slarse:
I'm going to do this in several PRs because the method is just that complex. It's a key component of Spoon, too, so I think it deserves some attention :)

slarse:
I'm satisfied with the refactoring of this method now. The next phase is to break `ReferenceBuilder` into smaller classes.

PR #3964, opened 2021-05-28
fix: Fix crash on handling parameterized ProblemReferenceBinding

slarse:
Fix #3951 

The problem is described in the issue. It's very rare for JDT to parse a parameterized type reference to a ProblemReferenceBinding (even when the actual referenced type can't be resolved). In the particular case of error code 15 (using a type that isn't closable on a try-with-resources), it does however resolve the reference to a ProblemReferenceBinding with type parameters in the name, which caused the crash.

This PR does _not_ attempt to parse the type arguments, it only fixes the crash and lets the pre-existing logic for creating type references from problem bindings do its thing. In other words, a type reference is created with the correct simple name, and that's about it. A warning that the type arguments are ignored is also issued.

I found a way to actually resolve the correct type binding in the particular corner case represented in the test case I added, which I may add in a later PR as that coincidentally triggers the long-standing bug reported in #3708.

PR #3966, opened 2021-05-28
refactor(ReferenceBuilder): Extract parameterized type reference creation into helpers

slarse:
#3965 

First PR in refactoring `ReferenceBuilder.getTypeReference()`.

This PR extracts the creation of parameterized type references into a set of three smaller helper methods. This will make it easier for me to later handle #3964


monperrus:
have you read "Clean code"? if not, you'll like it @slarse :)

slarse:
@monperrus I read most of it during my first year in CS (so, that's almost 6 years ago now). It's safe to say it influenced my programming style quite a bit. I've planned to revisit it for quite some time but... time flies.

Issue #3923, opened 2021-05-12
Improve quality of Spoon's API documentation


slarse:
A lot of the documentation of the public API is missing documentation for things like parameters and return types (think `@param` and `@return` tags). In my opinion, the most glaring omissions of these details are in the metamodel docs, found in the subpackages of [`spoon.reflect`](https://github.com/INRIA/spoon/tree/master/src/main/java/spoon/reflect).

For example, at the time of writing this, [CtImport](https://github.com/INRIA/spoon/blob/009ad17627741c67d1a04e1f363730e6dfa17ee3/src/main/java/spoon/reflect/declaration/CtImport.java) is entirely missing tags in its Javadocs.

I will expand on this issue with more instructions soon.

bhargavi-kotha:
@slarse Is this still open to start with contribution? And i'm willing to work on GSoC 2022 Implement Javadoc quality project idea.
can u please guide me on how to start with if not this PR to go on. I have experience on java based microservices and spring boot.I have implemented unit test cases with sonar coverage 85% in many of the projects.

MartinWitt:
Hi @bhargavi-kotha,

Yes, this issue is still open and there should be countless Javadoc issues left. You are pretty fast because I have still no confirmation that we are accepted as GSoC project. In the next days(when I/we get the confirmation) we create a new GSoC post on things you can do to get into the project spoon. For now, you can have a look at the old post from @monperrus https://github.com/INRIA/spoon/issues/3828 from 2021.

slarse:
What @MartinWitt said :)

bhargavi-kotha:
Thanks @MartinWitt and @slarse . I would go through any documentation and explore the project. would start off with any issue, until we get confirmation.

monperrus:
Javadoc warning's today

```
[WARNING] Javadoc Warnings
[WARNING] /home/martin/martin-no-backup/spoon/spoon-core/src/main/java/spoon/compiler/Environment.java:579: warning - invalid usage of tag >
[WARNING] /home/martin/martin-no-backup/spoon/spoon-core/src/main/java/spoon/processing/AbstractParallelProcessor.java:34: warning - Tag @link: reference not found: Executors#newFixedThreadPool()
[WARNING] /home/martin/martin-no-backup/spoon/spoon-core/src/main/java/spoon/refactoring/Refactoring.java:234: warning - @param argument "input" is not a parameter name.
[...]

PR #3968, opened 2021-05-28
doc: Improve Javadoc quality in Environment interface

slarse:
#3923 

The `Environment` interface is arguably one of the most important ones in the entire Spoon library, and it's fairly underdocumented. This PR improves the documentation by adding all missing `@param` and `@return` tags, fixing errors and adding missing explanations of what things actually do.

Javadoc method quality score increases like so:

```
JAVADOC QUALITY SCORE (lower is better)
    Compare: 2641
    Current: 2584
```

There are still improvements to be made here, especially in explaining what the options you can set actually do, but I felt like this was enough for a single PR.

### DIFF
--- a/src/main/java/spoon/compiler/Environment.java
+++ b/src/main/java/spoon/compiler/Environment.java
@@ -36,29 +36,40 @@ public interface Environment {
 
 	/**
 	 * Gets the Java version compliance level.
+	 *
+	 * @return the compliance level
 	 */
 	int getComplianceLevel();
 
 	/**
 	 * Sets the Java version compliance level.
+	 *
+	 * @param level the compliance level
 	 */
 	void setComplianceLevel(int level);
 
 	/**
 	 * Returns true if preview language features are enabled.
+	 *
+	 * @return true iff preview features are enabled
 	 */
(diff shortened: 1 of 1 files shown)


src/main/java/spoon/compiler/Environment.java
```diff
@@ -142,60 +170,83 @@
...
 	 * Sets the size of the tabulations in the generated source code.
+	 *
+	 * @param size tabulation size to set
 	 */
 	void setTabulationSize(int size);
 
 	/**
 	 * Tells if Spoon uses tabulations in the source code.
+	 *
+	 * @return true iff Spoon uses tabsulationswhen pretty-printing
```
andrewbwogi, line 218: Hi @slarse, you have a typo here: `tabsulationswhen`

andrewbwogi, line 218:
```suggestion
	 * @return true iff Spoon uses tabulations when pretty-printing
```

slarse: Thanks!

monperrus:
We :heart: documentation PR. Thanks Simon.

PR #3969, opened 2021-05-29
test: migrate CtTypeReferenceTest to Junit5

Rohitesh-Kumar-Jain:
#3919

This PR migrates FilterTest to Junit5.

Hi @slarse,

I tried to replace```Rule``` with ```ExtendWith```, but wasn't able to replace it successfully.

I would be glad if someone may help me perform this replacement.

I am unable to understand  [MockitoExtension](https://www.javadoc.io/doc/org.mockito/mockito-junit-jupiter/2.23.4/org/mockito/junit/jupiter/MockitoExtension.html) isn't getting resolved despite adding dependencies. (references : [baeldung](https://www.baeldung.com/mockito-junit-5-extension) & [stackoverflow](https://stackoverflow.com/questions/40961057/how-to-use-mockito-with-junit5)).

### DIFF
--- a/spoon-pom/pom.xml
+++ b/spoon-pom/pom.xml
@@ -59,7 +59,7 @@
             <scope>test</scope>
         </dependency>
 
-        <dependency>
+  <dependency>
     <!-- must come after junit5 deps, otherwise it overrides the junit version -->
     <!-- must be here in the parent POM, because the parent always comes last, see https://stackoverflow.com/questions/28999057/order-of-inherited-dependencies-in-maven-3 -->
     <groupId>com.github.stefanbirkner</groupId>
--- a/spoon-pom/pom.xml
+++ b/spoon-pom/pom.xml
@@ -52,8 +52,14 @@
         <version>5.7.2</version>
         <scope>test</scope>
     </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-junit-jupiter</artifactId>
+            <version>3.10.0</version>
+            <scope>test</scope>
+        </dependency>
 
-  <dependency>
+        <dependency>
     <!-- must come after junit5 deps, otherwise it overrides the junit version -->
     <!-- must be here in the parent POM, because the parent always comes last, see https://stackoverflow.com/questions/28999057/order-of-inherited-dependencies-in-maven-3 -->
     <groupId>com.github.stefanbirkner</groupId>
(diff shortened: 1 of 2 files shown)


slarse:
Hi @Rohitesh-Kumar-Jain,

There's no problem resolving the Mockito extension. The error message here says that there are unnecessary stubbings (i.e. some methods are being stubbed but aren't used in the test case): https://github.com/INRIA/spoon/pull/3969/checks?check_run_id=2701008263#step:10:452

There are two stubbings that are uneccessary for one test (but necessary for the other), the error message says exactly which lines correspond to the unnecessary stubbings. Since these stubbings already existed as-is, I would make them lenient (see the [baeldung article you linked](https://www.baeldung.com/mockito-junit-5-extension#building-the-test-class)) to avoid having to redesign the test case.

When you get a failure in GitHub Actions, click the "Details" button and it will take you to the error. You may have to scroll up and down to actually find the source of the error. Note that a _test failure_ can't occur if there are failures to resolve dependencies, as in that case the project won't even compile (there'll be a _compile error_).

Rohitesh-Kumar-Jain:
@slarse thanks for helping me making this PR become green : )

spoon-pom/pom.xml
```diff
@@ -52,8 +52,14 @@
...
     </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-junit-jupiter</artifactId>
+            <version>3.10.0</version>
+            <scope>test</scope>
+        </dependency>
 
-  <dependency>
+        <dependency>
```
slarse, line 62: Unnecessary whitespace change, not related to your patch. The indentation in this file is pretty off but that's for a separate PR.

slarse (changes requested):
Just revert an unrelated whitespace change and then we're good!

In general, try to avoid changing stuff outside of the scope of your PR. It's a good thing to look over your changes one extra time when you consider yourself done :)

Rohitesh-Kumar-Jain:
> In general, try to avoid changing stuff outside of the scope of your PR. It's a good thing to look over your changes one extra time when you consider yourself done :)

Thanks, I will keep this in my mind for future PRs

slarse (approved):
@Rohitesh-Kumar-Jain LGTM, nice work! I suggest you try to get started with writing some tests now, perhaps in one of the test files you've converted. It seems to me like you have no problems with the general flow of a pull request anymore.

@monperrus ping for merge

monperrus:
:rocket: @Rohitesh-Kumar-Jain and thanks :1st_place_medal: @slarse for the great onboarding

PR #3971, opened 2021-05-31
fix(noclasspath): Handle type resolution of anonymous class with unresolved constructor

### LINKED ISSUE #3913: getTypeReference(TypeReference) returns unexpected result when a type has multiple parameters
# Abstract
`getTypeReference(TypeReference)` in `spoon.support.compiler.jdt.ReferenceBuilder` returns unexpected result when a type has multiple parameters. It can cause exception when building a model.

# Cause of the problem and mitigation

https://github.com/INRIA/spoon/blob/c3ed4be2a825588c8253418caa4905336c4e985f/src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java#L660

When a type has multiple parameters, the result of `res.toStringDebug()` contains a whitespace after commas between the parameters (e.g., `Foo<?, ?>`). In contrast, `nameParameterized` does not contain any whitespaces (e.g., `Foo<?,?>`).  To my understanding, `replace(", ?", ",?")` was inserted to fill the gap. However, it's worked only when the parameters are wildcard. Therefore, it should be `replace(", ", ",")` rather than `replace(", ?", ",?")`.

# How to reproduce

I couldn't create a minimum code to reproduce the problem. I found this problem when I try to build a model from following code with no class path mode.
https://github.com/google/guava/blob/559e1ac1359e8d5dc0941aa72691698185d65407/android/guava/src/com/google/common/collect/Maps.java

slarse:
Fix #3913 

This PR fixes a problem in resolving generic types in noclasspath mode. The symptoms are specifically evident when creating an anonymous subclass of a generic type while using a constructor that does not exist. It can probably occur elsewhere as well, but this is where it was found to be easily reproducible.

See #3913 for further details.

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
+++ b/src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
@@ -659,7 +659,7 @@ public class ReferenceBuilder {
 			CtPackageReference packageReference = index >= 0 ? packageFactory.getOrCreate(concatSubArray(namesParameterized, index)).getReference() : packageFactory.topLevel();
 			inner.setPackage(packageReference);
 		}
-		if (!res.toStringDebug().replace(", ?", ",?").endsWith(nameParameterized)) {
+		if (!res.toStringDebug().replace(", ", ",").endsWith(nameParameterized)) {
 			// verify that we did not match a class that have the same name in a different package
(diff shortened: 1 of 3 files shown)


monperrus:
Thanks @slarse for the nice one-liner.

Issue #3967, opened 2021-05-28
track progress of @Rohitesh-Kumar-Jain - GSoC 2021 student


Rohitesh-Kumar-Jain:
Hi all, 

This issue is for keeping track of my deliverables and work in the Google Summer of Code, also this issue can be used for communication between mentors/org members and the student.

Org members feel free to drop suggestions and feedback regarding the GSoC program : )

## Timeline

Here's a link to the [official GSoC timeline](https://summerofcode.withgoogle.com/how-it-works/#timeline).

|  Phase  | Work Summary | Related PRs |
| ------------- | -------------------- | -----------------------|
| Community Bonding  <br/> (May 17 – June 7)  |  Migrated CtTypeReferenceTest, AstCheckerTest, FilterTest, ProcessingTest, CtGenerationTest, CorrectIdentifierTest to Junit 5 | #3969, #3960, <br/> #3959, #3958, #3956, #3946
| Week 1 <br/> (June 7 – 13)  | Improved tests for NameFilter & SpoonTagger class, migrated  all the  tests in template directory to Junit5, had a discussion about  the location & extraction of tests | #3986, #3982, #3975, #3978 
| Week 2 <br/> (June 14 – 20) | This week I transitioned from improving existing tests to writing new tests, and added tests for abstractTemplate and Substitution class |  #3990, #3995
| Week 3 <br/> (June 21 – 27) | Added tests for `insertAllNestedTypes`, `insertGeneratedNestedTypes`, `insertAllConstrutors`, `insertConsturctor`, `SpoonTreeBuilder.enter`, `SpoonTreeBuilder.exit`, `SpoonTreeBuilder.createNode`. Migrated all `spoon.refactoring`, `spoon.support` and `spoon.testing` tests to Junit5  | #3998, #4005, #4006, #4007, #4009
| Week 4 <br/> (June 28 – July 4) | Added test for `Substitution.insertAllSuperInterfaces`, `substituteMethodBody`, `substituteStatement`, `substituteFieldDefaultExpression` | #4012, #4016, #4018
| Week 5  <br/> (July 5 – 11) | Added tests for `asssertNotNull`, `assertExists`, `assertIsSame` | #4034
| Week 6 <br/> (July 12 – 18) | added test for `createPackageFile`, `createModuleFile`, `CtCommentImpl.equals`, migrated few test classes to Junit5 | #4043, #4045, #4053|
[...]

Issue #1818, opened 2018-01-12
Improve tests in Spoon


surli:
I just had a report from @oscarlvp who is working on software testing that we have a lot of methods not properly tested in Spoon.
For a lot of them we have a coverage, but removing the body of the method does not change at all the test results. For other methods, deciding an arbitrary return value (like `return true`) does not change either the test results.

So we should improve our test basis. The complete list of badly tested methods (and why) is available there: https://github.com/STAMP-project/descartes-experiments/blob/master/spoon/index.md

zielint0:
@surli 
> The complete list of badly tested methods (and why) is available there: 

404

> we have a lot of methods not properly tested in Spoon.

Please elaborate what is *proper* testing (in case of Spoon).
I would appreciate some examples.
Mercii in advance.

oscarlvp:
The list of issues has been placed [here](https://github.com/STAMP-project/descartes-experiments/blob/master/actionable-hints/spoon/index.md)
For more information about the way these issues were found please take a look at [descartes](https://github.com/STAMP-project/pitest-descartes).
A sample of those issues was inspected before by @surli. The result can be seen [here](https://github.com/STAMP-project/descartes-experiments/blob/master/actionable-hints/spoon/sample.md). Those that have marks are the ones that would need a closer look.
I hope this helps.

zielint0:
¡Gracias hermano!

KishkinJ10:
can i do this in my GSoC coding period please??

Rohitesh-Kumar-Jain:
Hi @oscarlvp, @surli & @monperrus 

> The list of issues has been placed [here](https://github.com/STAMP-project/descartes-experiments/blob/master/actionable-hints/spoon/index.md)

I feel analysis of methods present in spoon.compiler, spoon.experimental, spoon.javadoc.internal, spoon.legacy, spoon.metamodel and spoon.pattern directories are missing from this list. 

As this list is outdated, I have tried to run Descartes on few subsets of tests, [here's the pit report](https://github.com/Rohitesh-Kumar-Jain/Descartes_Pit_Reports) generated after running Descartes (I am yet to run Descartes on all the methods and tests).

I would be glad if someone may go through the pit reports and let me know whether I am running the Descartes in the desired fashion.

Rohitesh-Kumar-Jain:
Hi @monperrus, 

[...]

PR #3975, opened 2021-06-04
test: improve tests for NameFilter class.

Rohitesh-Kumar-Jain:
#1818

This PR aims to improve the code coverage of the class [NameFilter](https://github.com/INRIA/spoon/blob/2fff4c50266f79bc06b961e3b01ca6c6e0fcc1ab/src/main/java/spoon/legacy/NameFilter.java).

Link to #3967

### DIFF
--- a/src/test/java/spoon/test/filters/FilterTest.java
+++ b/src/test/java/spoon/test/filters/FilterTest.java
@@ -111,7 +111,7 @@ public class FilterTest {
 
 	@Test
 	public void testNameFilter() throws Exception {
-		// contract: the constructor of the class NameFilter creates and initialises an object.
+		// contract: NameFilter finds the expected amount of elements when filtering
 
 		CtClass<?> foo = factory.Package().get("spoon.test.filters.testclasses").getType("Foo");
 		assertEquals("Foo", foo.getSimpleName());
--- a/src/test/java/spoon/test/filters/FilterTest.java
+++ b/src/test/java/spoon/test/filters/FilterTest.java
@@ -136,16 +136,16 @@ public class FilterTest {
 		assertTrue(matches);
 	}
 
-	@Test()
+	@Test
 	public void testNameFilterThrowsExceptionForNull() {
 		// contract: NameFilter constructor should throw IllegalArgumentException when passed null
 
 		assertThrows(IllegalArgumentException.class, () -> new NameFilter<>(null));
 	}
 
-	@Test()
+	@Test
 	public void testNameFilterGetType() {
-		// contract: getType method returns the object of class NameFilter
+		// contract: getType method returns the datatype that the NameFilter class returns
 
 		NameFilter<CtNamedElement> nameFilter = new NameFilter<>("i");
 		assertEquals(CtNamedElement.class, nameFilter.getType());
--- a/src/test/java/spoon/test/filters/FilterTest.java
+++ b/src/test/java/spoon/test/filters/FilterTest.java
@@ -118,6 +118,12 @@ public class FilterTest {
 		assertEquals(1, elements.size());
 	}
 
+	@Test()
+	public void testNameFilterThrowsException() {
+		CtClass<?> foo = factory.Package().get("spoon.test.filters.testclasses").getType("Foo");
+		assertThrows(IllegalArgumentException.class, () -> foo.getElements(new NameFilter<>(null)));
+	}
+
 	@Test
 	public void testFilters() {
 		CtClass<?> foo = factory.Package().get("spoon.test.filters.testclasses").getType("Foo");
--- a/src/test/java/spoon/test/filters/FilterTest.java
+++ b/src/test/java/spoon/test/filters/FilterTest.java
@@ -111,20 +111,45 @@ public class FilterTest {
 
 	@Test
 	public void testNameFilter() throws Exception {
-		// contract: legacy NameFilter is tested and works
+		// contract: the constructor of the class NameFilter creates and initialises an object.
+
 		CtClass<?> foo = factory.Package().get("spoon.test.filters.testclasses").getType("Foo");
 		assertEquals("Foo", foo.getSimpleName());
 		List<CtNamedElement> elements = foo.getElements(new NameFilter<>("i"));
-		// contract: tests NameFilter constructor for cases other than null
 		assertEquals(1, elements.size());
-		// contract: tests NameFilter constructor for null value case
-		assertThrows(IllegalArgumentException.class, () -> foo.getElements(new NameFilter<>(null)));
-		// contract: tests the getType method of the NameFilter class
-		assertEquals("CtNamedElement",new NameFilter<>("i").getType().getSimpleName());
-		// contract: tests the matches method of the NameFilter class
-		assertTrue(new NameFilter<>("i").matches(elements.get(0)));
 	}
 
+	@Test
+	void testNameFilterMatchesReturnsTrueForMatchingName() {
+		// contract: NameFilter.matches should return true for an element with a matching name
+
+		// arrange
+		String name = "someNiceName";
+		NameFilter<CtNamedElement> nameFilter = new NameFilter<>(name);
+		CtNamedElement elemWithMatchingName = new Launcher().getFactory().createLocalVariable();
+		elemWithMatchingName.setSimpleName(name);
+
+		// act
+		boolean matches = nameFilter.matches(elemWithMatchingName);
+
+		// assert
+		assertTrue(matches);
+	}
+
+	@Test()
+	public void testNameFilterThrowsExceptionForNull() {
+		// contract: NameFilter constructor should throw IllegalArgumentException when passed null
+
+		assertThrows(IllegalArgumentException.class, () -> new NameFilter<>(null));
+	}
+
+	@Test()
+	public void testNameFilterGetType() {
+		// contract: getType method returns the object of class NameFilter
+
+		NameFilter<CtNamedElement> nameFilter = new NameFilter<>("i");
+		assertEquals(CtNamedElement.class, nameFilter.getType());
+	}
 
 	@Test
 	public void testFilters() {
--- a/src/test/java/spoon/test/filters/FilterTest.java
+++ b/src/test/java/spoon/test/filters/FilterTest.java
@@ -120,11 +120,6 @@ public class FilterTest {
 		assertThrows(IllegalArgumentException.class, () -> foo.getElements(new NameFilter<>(null)));
 	}
 
-	@Test()
-	public void testNameFilterThrowsException() {
-		CtClass<?> foo = factory.Package().get("spoon.test.filters.testclasses").getType("Foo");
-		assertThrows(IllegalArgumentException.class, () -> foo.getElements(new NameFilter<>(null)));
-	}
 
 	@Test
 	public void testFilters() {
--- a/src/test/java/spoon/test/filters/FilterTest.java
+++ b/src/test/java/spoon/test/filters/FilterTest.java
@@ -113,9 +113,11 @@ public class FilterTest {
 	public void testNameFilter() throws Exception {
 		// contract: legacy NameFilter is tested and works
 		CtClass<?> foo = factory.Package().get("spoon.test.filters.testclasses").getType("Foo");
-		assertEquals("Foo", foo.getSimpleName());
 		List<CtNamedElement> elements = foo.getElements(new NameFilter<>("i"));
 		assertEquals(1, elements.size());
+		assertEquals("CtNamedElement",new NameFilter<>("i").getType().getSimpleName());
+		assertTrue(new NameFilter<>("i").matches(elements.get(0)));
+		assertThrows(IllegalArgumentException.class, () -> foo.getElements(new NameFilter<>(null)));
 	}
 
 	@Test()
--- a/src/test/java/spoon/test/filters/FilterTest.java
+++ b/src/test/java/spoon/test/filters/FilterTest.java
@@ -113,11 +113,16 @@ public class FilterTest {
 	public void testNameFilter() throws Exception {
 		// contract: legacy NameFilter is tested and works
 		CtClass<?> foo = factory.Package().get("spoon.test.filters.testclasses").getType("Foo");
+		assertEquals("Foo", foo.getSimpleName());
 		List<CtNamedElement> elements = foo.getElements(new NameFilter<>("i"));
+		// contract: tests NameFilter constructor for cases other than null
 		assertEquals(1, elements.size());
-		assertEquals("CtNamedElement",new NameFilter<>("i").getType().getSimpleName());
-		assertTrue(new NameFilter<>("i").matches(elements.get(0)));
+		// contract: tests NameFilter constructor for null value case
 		assertThrows(IllegalArgumentException.class, () -> foo.getElements(new NameFilter<>(null)));
+		// contract: tests the getType method of the NameFilter class
+		assertEquals("CtNamedElement",new NameFilter<>("i").getType().getSimpleName());
+		// contract: tests the matches method of the NameFilter class
+		assertTrue(new NameFilter<>("i").matches(elements.get(0)));
 	}
 
 
--- a/src/test/java/spoon/test/filters/FilterTest.java
+++ b/src/test/java/spoon/test/filters/FilterTest.java
@@ -145,7 +145,7 @@ public class FilterTest {
 
 	@Test
 	public void testNameFilterGetType() {
-		// contract: getType method returns the datatype that the NameFilter class returns
+		// contract: NameFilter.getType() should return the CtNamedElement class
 
 		NameFilter<CtNamedElement> nameFilter = new NameFilter<>("i");
 		assertEquals(CtNamedElement.class, nameFilter.getType());


Rohitesh-Kumar-Jain:
I have added one additional test to improve the line coverage of the tests for NameFilter class.

I would be glad if @slarse or maybe someone else may tell me what should be the scope for a single PR and a commit for improving tests.

slarse:
> what should be the scope for a single PR and a commit for improving tests.

There's no easy answer to that, it depends. Arguably, a PR can't really be too small, so there's really no minimum scope. Somewhat arbitrarily we can say that a maximum scope would be adding tests relating to a single public method. That is to say, don't target multiple pieces of functionality with a single PR.

Rohitesh-Kumar-Jain:
Hi,

One of the difficulties I am facing is that I am not being able to generate the desired Descartes results on a smaller scope.

What I wish to say is, that I was able to [generate Descartes results as a whole](https://github.com/Rohitesh-Kumar-Jain/Descartes_on_Spoon), here's the result for NameFilter Class after the generation.

Screen Shot 2021-06-04 at 12 21 26-fullpage

But as running Descartes on the whole Spoon takes around 20 + hours to run, I tried running individually but it isn't generating the desired results.

The plugin I used was 
```
       <plugin>
            <groupId>org.pitest</groupId>
            <artifactId>pitest-maven</artifactId>
            <version>1.6.4</version>
            <configuration>
                <mutationEngine>descartes</mutationEngine>
                <targetClasses>
                    <param>spoon.legacy.NameFilter</param>
                </targetClasses>
                <targetTests>
                    <param>spoon.test.filters.FilterTest</param>
                </targetTests>
            </configuration>
            <dependencies>
                <dependency>
                    <groupId>eu.stamp-project</groupId>
                    <artifactId>descartes</artifactId>
                    <version>1.3.1</version>
                </dependency>
            </dependencies>
        </plugin>
```
But running that is generating this :

Screen Shot 2021-06-04 at 12 26 28-fullpage

So how do I measure the test parameters effectively ?

In the meeting @slarse suggested it is not necessary to run Descartes all the time, suggestions were to remove some part of the code and see tests are working, also IntelliJ IDEA comes with an inbuilt code coverage parameter, would anyone like to suggest something?

mentioning @nharrand and @monperrus

monperrus:
> In the meeting @slarse suggested it is not necessary to run Descartes all the tim

Fully agree. You can run Descartes every other day, in order to:

1. verify that the new merged tests indeed improve the Descartes metrics
2. select your next test target.

slarse:
@Rohitesh-Kumar-Jain See my comment in #3967. For measuring code coverage, using Descartes is incredibly inefficient.

> You can run Descartes every other day, in order to:

Considering that it takes @Rohitesh-Kumar-Jain 20 hours to run Descartes on Spoon, I wouldn't recommend running it more than once a week.

Rohitesh-Kumar-Jain:
> Considering that it takes @Rohitesh-Kumar-Jain 20 hours to run Descartes on Spoon, I wouldn't recommend running it more than once a week.

Yes it keeps running for 20 hours in the background, my effort is to just run this, but still, I have to wait almost a day to see the new report. Although Descartes can be run with subsets however in this case the reports aren't as expected after targeting the exact class and the test.

slarse:
> Yes it keeps running for 20 hours in the background, my effort is to just run this, but still, I have to wait almost a day to see the new report. Although Descartes can be run with subsets however in this case the reports aren't as expected after targeting the exact class and the test.

That's most likely because there are other tests that exercise the code you target. A large part of Spoon's test code is not unit tests, but rather run the entire thing from parsing a model to printing a result, and so there can be many tests that touch e.g. `NameFilter` without them being called anything relating to `Filter`.

Rohitesh-Kumar-Jain:
> 
> That's most likely because there are other tests that exercise the code you target. A large part of Spoon's test code is not unit tests, but rather run the entire thing from parsing a model to printing a result, and so there can be many tests that touch e.g. `NameFilter` without them being called anything relating to `Filter`.

Thanks for letting me know this

Rohitesh-Kumar-Jain:
Hi, 

I feel this line of code ```assertEquals("Foo", foo.getSimpleName());``` is redundant in [testNameFilter](https://github.com/INRIA/spoon/blob/master/src/test/java/spoon/test/filters/FilterTest.java) method, as this assertion has nothing to do with NameFilter class, also this assertion has been called for about 11 more times in other 6 files, so I believe removing this line of won't reduce test quality and is redundant ?

looping in @nharrand

Rohitesh-Kumar-Jain:
Hi team, 

Now there are two options available : 
1) I have extracted the NameFilterTest form FilterTest in #3978, so after the extraction is merged I can close this PR and raise a new one for improving the test in the NameFilterTest.
2) If we don't wish to extract, we can close #3978, and simply just focus on this PR.

slarse:
Hi @Rohitesh-Kumar-Jain,

As one point of feedback, avoid working on things in parallel that conflict with each other. If either this or #3978 is merged, there will be a merge conflict between them.

With that said, this PR needs a little bit of work. When it comes to tests, it's important that _each test tests one thing_, one concept. That doesn't necessarily mean that each test needs to have precisely one assert, but it does mean that you shouldn't test happy-path behavior (everything works as it should) and crashes in the same test. Having a lot of assertions in the same test is also indicative that it's testing too many concepts.

As a rule of thumb, you should only call the method under test _once_ in the test itself. This doesn't always hold (sometimes you specifically want to test what happens if you call it twice, for example), but I find that most often this rule is good.

Furthermore, I strongly recommend that you follow the arrange/act/assert pattern (AAA). You can look it up in your preferred search engine.

So, in terms of this PR, break out your new assertions into separate test cases and write them according to all of the rules we've discussed, including a contract for the test clearly specifying what it's supposed to test. Note that the original test that you altered here is not a good test, as it does not say what it's actually supposed to verify ("NameFilter is tested and works" is extremely vague).

Rohitesh-Kumar-Jain:
> As one point of feedback, avoid working on things in parallel that conflict with each other. If either this or #3978 is merged, there will be a merge conflict between them.

I will keep this in mind for future PRs.

> Furthermore, I strongly recommend that you follow the arrange/act/assert pattern (AAA). You can look it up in your preferred search engine.

I will surely have a look at this, thanks for making me aware of the AAA pattern.

monperrus:
This PR looks good in the sense that it add new assertions. 

We don't need to remove `assertEquals("Foo", foo.getSimpleName());`, unless it is incorrect, which is not the case here.

slarse:
> This PR looks good in the sense that it add new assertions.

IMO this is not an inherently good thing, arbitrarily adding new assertions to an existing test is not sustainable. It then becomes unclear what that test is supposed to verify. The original test doesn't have a clear goal (the contract essentially just says "test that name filter is correct"), which doesn't make things better.

"God tests" that verify everything about some method or class are a recipe for maintainability degradation.

As for the removal of the guard assertion, I agree that it's unrelated to remove. I don't think it should have been there to begin with as it is essentially just clutter, but that's a separate issue.

monperrus:
> arbitrarily adding new assertions to an existing test is not sustainable

yes, it's an engineering tradeoff between adding assertions to an existing test (if the test intention is the same), and creating a new test.

slarse:
> > arbitrarily adding new assertions to an existing test is not sustainable
> 
> yes, it's an engineering tradeoff between adding assertions to an existing test (if the test intention is the same), and creating a new test.

If the test intention is the same, I'm all for adding a "missing" assertion, in no way do I mean that there should be only one assertion per test. In the test case in this PR, I don't think the test intention is clear, and so wouldn't want to add anything to it.

monperrus:
>  In the test case in this PR, I don't think the test intention is clear, and so wouldn't want to add anything to it.

OK for me. @Rohitesh-Kumar-Jain is NameFilter in the Descartes report?

Rohitesh-Kumar-Jain:
Sorry for the late reply

> OK for me. @Rohitesh-Kumar-Jain is NameFilter in the Descartes report?

yes, [it is here](https://github.com/Rohitesh-Kumar-Jain/Descartes_on_Spoon/blob/master/202104291236/spoon.legacy/NameFilter.java.html)

Screen Shot 2021-06-08 at 11 54 34-fullpage

monperrus:
OK, then let's finish this PR :) it's in our plan.

Could you document the test intention of the added assertions with a comment line `// contract:` before each of them? Thanks!

src/test/java/spoon/test/filters/FilterTest.java
```diff
@@ -115,9 +115,17 @@ public void testNameFilter() throws Exception {
 		CtClass<?> foo = factory.Package().get("spoon.test.filters.testclasses").getType("Foo");
 		assertEquals("Foo", foo.getSimpleName());
 		List<CtNamedElement> elements = foo.getElements(new NameFilter<>("i"));
+		// contract: tests NameFilter constructor for cases other than null
 		assertEquals(1, elements.size());
+		// contract: tests NameFilter constructor for null value case
```
slarse, line 120: The contract should not say what the test does, but what is expected to happen. For example, this one ought to say `contract: NameFilter constructor should throw IllegalArgumentException when passed null`

The same problem exists with all of the contracts you have written here, they just say what the test does, while they should detail the expected outcome of some action.

src/test/java/spoon/test/filters/FilterTest.java
```diff
@@ -115,9 +115,17 @@ public void testNameFilter() throws Exception {
...
+		assertThrows(IllegalArgumentException.class, () -> foo.getElements(new NameFilter<>(null)));
```
slarse, line 121: Calling `foo.getElements` is redundant as `new NameFilter<>(null)` already throws.

```suggestion
		assertThrows(IllegalArgumentException.class, () -> new NameFilter<>(null));
```

src/test/java/spoon/test/filters/FilterTest.java
```diff
@@ -115,9 +115,17 @@ public void testNameFilter() throws Exception {
...
+		// contract: tests the getType method of the NameFilter class
+		assertEquals("CtNamedElement",new NameFilter<>("i").getType().getSimpleName());
```
slarse, line 123: Avoid unnecessary chains. Here, what you're checking is the return value of `getType`, so check that!

```suggestion
		assertEquals(CtNamedElement.class,new NameFilter<>("i").getType());
```

src/test/java/spoon/test/filters/FilterTest.java
```diff
@@ -115,9 +115,17 @@ public void testNameFilter() throws Exception {
...
+		// contract: tests the matches method of the NameFilter class
+		assertTrue(new NameFilter<>("i").matches(elements.get(0)));
```
slarse, line 125: This is a self-fulfilling test. The element in `elements` are all elements for which `matches` returned `true`, and so this must be `true`.

slarse (changes requested):
Hi @Rohitesh-Kumar-Jain,

This is a good start. Please see my comments.

I also have some general comments.

Also note that each test should have _one_ contract. The assertions you've added in this test each have their own contract, and so should be separate tests. Yes, there are tests in the test suite which have multiple contracts, but that's not great. One test should test one concept <==> one test should have one contract.

In other words: break out your added assertions into separate tests.

src/test/java/spoon/test/filters/FilterTest.java
Rohitesh-Kumar-Jain: Got it 👍🏼

src/test/java/spoon/test/filters/FilterTest.java
Rohitesh-Kumar-Jain: Yes, it is actually redundant, thanks for pointing it out I welcome it.

src/test/java/spoon/test/filters/FilterTest.java
Rohitesh-Kumar-Jain: will change this

src/test/java/spoon/test/filters/FilterTest.java
Rohitesh-Kumar-Jain: @slarse I don't get this one, this test is supposed to test the matches method of the class ```NameFilter```, which it is testing. I have tried applying mutation manually like "replacing the method body with false" which the test is killing, so where's the problem?

Rohitesh-Kumar-Jain:
Hi @slarse, thanks for reviewing PRs : )

> In other words: break out your added assertions into separate tests.

Understood 👍🏼

src/test/java/spoon/test/filters/FilterTest.java
slarse: It's not incorrect, it's just redundant. Spoon's filtering mechanism works by gathering all elements for which a filter returns `true`. Thus, by definition, all elements in the filtered list will match the filter that was used. If you look at it in isolation, this is what you have:

```java
List<CtNamedElement> elements = foo.getElements(new NameFilter<>("i"));
// invariant: for every element e in elements, `new NameFilter<>("i") == true` by the definition of Spoon's filters
assertTrue(new NameFilter<>("i").matches(elements.get(0))); // redundant assertion
```

As long as `getElements(Filter)` is working correctly, it's given that every element in `elements` is one such that the used filter's matches method returned true. It is _not_ the responsibility of a test for `NameFilter` to verify that the overall filtering mechanism works as it should, that would fall on a separate test.

Finally, the mutant you are mentioning is killed by the very first assertion (i.e. `assertEquals(1, elements.size())`), because if `matches` always returns `false`, then no elements are collected.

If you in isolation wanted to test the `matches` method of `NameFilter`, you would do something like so instead:

```java
	@Test
	void testNameFilterMatchesReturnsTrueForMatchingName() {
		// contract: NameFilter.matches should return true for an element with a matching name

		// arrange
		String name = "someNiceName";
		NameFilter<CtNamedElement> nameFilter = new NameFilter<>(name);
		CtNamedElement elemWithMatchingName = new Launcher().getFactory().createLocalVariable();
		elemWithMatchingName.setSimpleName(name);
		
		// act
		boolean matches = nameFilter.matches(elemWithMatchingName);
		
		// assert
		assertTrue(matches);
	}
```

Tests should be as precisely controlled as possible, and rely as little as possible on the component they are testing for setting up the preconditions.

Rohitesh-Kumar-Jain: So I am supposed to follow this structure from now on. I have seen 10 more tests in the codebase following this contract, arrange, act, assert principle I will try to understand those as well.

Actually in most of the tests in spoon contract is missing, AAA is not being followed. But we should follow the best practices while writing the new tests and improving the existing tests. Thanks for letting me know, I appreciate it.

slarse: > I have seen 10 more tests in the codebase following this contract, arrange, act, assert principle I will try to understand those as well.

You will find that there are hundreds of tests following this principle, but most of them do not contain comments saying which phase is where. I put them there for emphasis, but for most tests they are redundant as it's obvious which part is what.

If you find another test structure that's more appropriate for some given test, feel free to use that. In general, I find that AAA works well and makes for clean and easy to read tests.

> Actually in most of the tests in spoon contract is missing, AAA is not being followed.

Spoon is old and has a lot of legacy from times when things were done differently. You will find a lot of things that aren't great. This is the case for almost any code base you'll ever work on. To paraphrase Uncle Bob, all we can do is to try to leave it in a little bit less of a mess than it was when we got there.

src/test/java/spoon/test/filters/FilterTest.java
```diff
@@ -111,20 +111,45 @@ public void setup() throws Exception {
...
+	@Test()
+	public void testNameFilterThrowsExceptionForNull() {
+		// contract: NameFilter constructor should throw IllegalArgumentException when passed null
+
+		assertThrows(IllegalArgumentException.class, () -> new NameFilter<>(null));
+	}
+
+	@Test()
+	public void testNameFilterGetType() {
+		// contract: getType method returns the object of class NameFilter
```
slarse, line 148: This contract is not correct, the `NameFilter.getType()` is not tested to return a `NameFilter` object.

src/test/java/spoon/test/filters/FilterTest.java
```diff
@@ -111,20 +111,45 @@ public void setup() throws Exception {
...
+		elemWithMatchingName.setSimpleName(name);
+
+		// act
+		boolean matches = nameFilter.matches(elemWithMatchingName);
+
+		// assert
+		assertTrue(matches);
+	}
+
+	@Test()
```
slarse, line 139:
```suggestion
	@Test
```

src/test/java/spoon/test/filters/FilterTest.java
```diff
@@ -111,20 +111,45 @@ public void setup() throws Exception {
...
+		NameFilter<CtNamedElement> nameFilter = new NameFilter<>(name);
+		CtNamedElement elemWithMatchingName = new Launcher().getFactory().createLocalVariable();
+		elemWithMatchingName.setSimpleName(name);
+
+		// act
+		boolean matches = nameFilter.matches(elemWithMatchingName);
+
+		// assert
+		assertTrue(matches);
+	}
```
slarse, line 137: You're free to keep this test if you want, but it doesn't improve line or mutation coverage since the method is exercised in the same way by the original `testNameFilter` test. So, it's a good and isolated test, but it doesn't actually improve any testing metrics.

It's not necessarily a bad thing to have overlapping tests, but for your future PRs you should focus on tests that do improve the metrics, especially the mutation metrics.

slarse (changes requested):
Just two things left to fix, final contract seems incorrect and there's a rogue parentheses on one of the `@Test` annotations.

src/test/java/spoon/test/filters/FilterTest.java
Rohitesh-Kumar-Jain: Let's keep this test as a reference for whoever will be working on FilterTest in future.

Yes, I will be next targeting tests with mutation metrics. 

I also have a doubt regarding mutation coverage, some files have 100% mutation strength but 40-50% mutation coverage, that does that implies?

src/test/java/spoon/test/filters/FilterTest.java
Rohitesh-Kumar-Jain: should the contract be like this : getType method returns the type of class which is begin used ?

src/test/java/spoon/test/filters/FilterTest.java
slarse: > I also have a doubt regarding mutation coverage, some files have 100% mutation strength but 40-50% mutation coverage, that does that implies?

This doesn't seem to be documented at all, but I'm pretty sure it's like this:

* Mutation coverage = all mutants / killed mutants
* Test strength = mutants in covered code / killed mutants

In other words, test strength shows how good the current tests are at killing mutants, while mutation coverage also includes mutants that exist in code which isn't even covered by the current tests.

Rohitesh-Kumar-Jain: got it

src/test/java/spoon/test/filters/FilterTest.java
slarse: There's no one answer to what a contract should say, but it cannot be incorrect. What the `getType` method actually is supossed to return is the `CtNamedElement` _class_, which is not an object of class `NameFilter`, and thus the comment here is explicitly incorrect.

The contract comment should state in human-readable terms what the test is supposed to verify. One potential contract here would be: `// contract: NameFilter.getType() should return the CtNamedElement class`

A less specific one would be: `// contract: NameFilter.getType() should return the correct class`, where the test itself then specifies that correct class.

Note that we don't usually test getters directly, because they are trivial pieces of code, and so the test and its contract here necessarily become a bit contrived.

src/test/java/spoon/test/filters/FilterTest.java
```diff
@@ -111,13 +111,46 @@ public void setup() throws Exception {
 
 	@Test
 	public void testNameFilter() throws Exception {
-		// contract: legacy NameFilter is tested and works
+		// contract: the constructor of the class NameFilter creates and initialises an object.
```
slarse, line 114: This contract is also not correct, this test is not for `NameFilter`'s constructor. The test verifies that `NameFilter` finds the expected amount of elements when filtering with it (it's a very weak test).

The original comment was technically correct, it was just pointless. But you don't have to change this contract when adding your other test methods, as you don't touch this test method. I'm guessing this is just left over from when you modified this method.

Either fix this comment such that it's correct, or revert to the original contract.

slarse (changes requested):

src/test/java/spoon/test/filters/FilterTest.java
```diff
@@ -111,13 +111,46 @@ public void setup() throws Exception {
...
+	@Test
+	public void testNameFilterThrowsExceptionForNull() {
+		// contract: NameFilter constructor should throw IllegalArgumentException when passed null
+
+		assertThrows(IllegalArgumentException.class, () -> new NameFilter<>(null));
+	}
+
+	@Test
+	public void testNameFilterGetType() {
+		// contract: getType method returns the datatype that the NameFilter class returns
```
slarse, line 148: I'm truly sorry for going on and on here, but this contract still doesn't make sense to me. It says that a method returns the datatype that a class returns. But a class does not return anything, and so it does not make sense.

It's very important to be able to succinctly specify what a test is supposed to verify, so we need to keep iterating on this until you've nailed it. Feel free to just comment in this comment with suggestions before committing to something.

src/test/java/spoon/test/filters/FilterTest.java
```diff
@@ -111,13 +111,46 @@ public void setup() throws Exception {
 
 	@Test
 	public void testNameFilter() throws Exception {
-		// contract: legacy NameFilter is tested and works
+		// contract: NameFilter finds the expected amount of elements when filtering
```
slarse, line 114: Very good!

src/test/java/spoon/test/filters/FilterTest.java
Rohitesh-Kumar-Jain: > I'm truly sorry for going on and on here
no no, it's my fault that iterations are still going on, had I raised a perfect PR, no iterations would have occurred. I will try to come with a  succinct contract.

Rohitesh-Kumar-Jain: I would like to go with the one suggested by you only ```// contract: NameFilter.getType() should return the CtNamedElement class```

Rohitesh-Kumar-Jain:
I am sorry for taking so many iterations, I am going to spend tomorrow reading lots of existing tests of Spoon so that my next PR may require lesser iterations : )

slarse (approved):
@Rohitesh-Kumar-Jain 

> I am sorry for taking so many iterations, I am going to spend tomorrow reading lots of existing tests of Spoon so that my next PR may require lesser iterations : )

Don't be sorry, it's perfectly normal and expected for the first contributions to a project to be take some time :)

@monperrus good for merge

monperrus:
Many thanks @Rohitesh-Kumar-Jain and @slarse

src/test/java/spoon/test/filters/FilterTest.java
Rohitesh-Kumar-Jain: > Note that we don't usually test getters directly, because they are trivial pieces of code

I have one question, so I can skip adding tests for getter like ```getFactory()``` in this [class](https://github.com/INRIA/spoon/blob/2fff4c50266f79bc06b961e3b01ca6c6e0fcc1ab/src/main/java/spoon/template/AbstractTemplate.java) ?

slarse: > I have one question, so I can skip adding tests for getter like getFactory() in this class ?

Directly testing that would be fairly far down the priority list, yes. It would for example be much more important to test `spoon.template.Substitution.getFactory`, as that method is more complex. Overall, the `Substitution` class looks fairly poorly tested when looking at the [Coveralls report](https://coveralls.io/builds/40529606/source?filename=src%2Fmain%2Fjava%2Fspoon%2Ftemplate%2FSubstitution.java), so perhaps that'd be a good target for you!

PR #3976, opened 2021-06-04
refactor(ReferenceBuilder): Remove dead code

slarse:
#3965 

While trying to wrap my head around the ReferenceBuilder, I found some unused code. This PR removes it.


monperrus:
Thanks @slarse

PR #3981, opened 2021-06-07
refactor(ReferenceBuilder): Extract small reference-building subroutines into helpers

slarse:
#3965 

This PR extracts all of the "sufficiently small subroutines" in `getTypeReference` into methods of their own.

I can create one PR for each little helper, but as they edit code very close to each other I'd need to do them sequentially (wait for one to get merged before opening the next), so I figured I'd submit like this to begin with.

@monperrus  Let me know if you want me to break it up into smaller PRs (perhaps split into 3 PRs with 3 extractions each?), or if this size is OK.


monperrus:
Thanks @slarse.

This refactoring has the good granularity. Hard to review manually but I fully trust the test suite. Will merge.

PR #3991, opened 2021-06-14
refactor(ReferenceBuilder): Refactor getting type refeference from TypeVariableBinding

### LINKED ISSUE #3965: Refactor `ReferenceBuilder.getTypeReference(TypeBinding, boolean)`

slarse:
#3965 

Extracts getting a type reference from a TypeVariableBinding into a separate method. This is pretty incredibly complicated, so I aim to also simplify the process.

This is the final significant PR in refactoring `ReferenceBuilder.getTypeReference`, but I intend to do one final PR after this to slightly refactor the top-level method. Then I'm happy.

UPDATE: I've now finished this refactoring. It was _hard_, it took me several hours to sort out which part was doing what. In the end, I managed to significantly flatten the structure, and even remove the `bounds` field entirely (which I initially had a really hard time understanding the purpose of). Note that I didn't put it in a local variable instead, the value just isn't needed anymore.

On a side note, I'm really proud of the fact that I extracted the building of a type reference from a type variable binding into six methods, yet only increased the line count by 3 lines :)

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
+++ b/src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
@@ -894,7 +894,7 @@ public class ReferenceBuilder {
 			return getTypeReference(((CaptureBinding) binding).wildcard, resolveGeneric);
 		} else if (binding instanceof CaptureBinding) {
 			CtWildcardReference wildcard = this.jdtTreeBuilder.getFactory().Core().createWildcardReference();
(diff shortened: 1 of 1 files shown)


monperrus:
Very nice! Removing state is always welcome :) Thanks @slarse

PR #3995, opened 2021-06-16
review: test: add test for insertAllFields method

Rohitesh-Kumar-Jain:
#1818

Hi, I have added a new test for `insertAllFilelds` method of the `Substituion` class, this test improves code coverage, mutation converge and test strength all from 0% to 100%.

Screenshot 2021-06-16 at 6 22 03 PM

I have manually tested that this test is killing that mutation.

Under the hood, it is testing more code, 

Screenshot 2021-06-16 at 6 28 03 PM

so I will run Descartes again when the iterations for this PR will end and this gets merged. Until then I believe I should work on some other poorly tested classes suggested by @slarse?

Link to #3967


src/test/java/spoon/test/template/TemplateTest.java
```diff
@@ -900,6 +901,28 @@ public void testSubstituteInnerClass() {
 		assertEquals("x.Result$Inner",fr.getDeclaringType().getQualifiedName());
 	}
 
+	@Test
+	public void testSubstitutionInsertAllFields() {
+		// contract: insertAllFields method of the Substitution class inserts all the fields from a given template by substituting all the template parameters by their values.
```
slarse, line 906: This contract does not adequately describe the test. What _actually_ happens in this test is that a single field, `String testString = "goodName";` is inserted into the created class. There is no substitution going on, there is only a single as-is field insertion.

A more apt description would be something like "Substitution.insertAllFields inserts the only field from a single-field template into the target class"

Additionally, the test actually only verifies that the field has the correct name, so it doesn't verify the field as a whole.

src/test/java/spoon/test/template/testclasses/InsertAllFieldsTemplate.java
```diff
@@ -0,0 +1,21 @@
...
+    @Local
+    public InsertAllFieldsTemplate(String parameter) {
+        _parameter_ = parameter;
+    }
+
+    @Override
+    public void statement() {
+        System.out.println(_parameter_);
+    }
+}
```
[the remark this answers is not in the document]
slarse: Finally, the template does not align with the test. From what I can surmise of the contract and test, what you're testing is that the single field `testString` is inserted. To test just that, you should create a minimal template for testing just that.

```java
public class InsertAllFieldsTemplate extends StatementTemplate {
    String testString = "goodName";

    @Override
    public void statement() { }
}
```

slarse (changes requested):
Hi @Rohitesh-Kumar-Jain,

It's great that your testing the `Substitution` class, it looks woefully undertested.

The PR needs some work, however. I've added quite a lot of comments in the code itself. In addition, I think this also calls for creating a new test class, `SubstitutionTest`, because we will need to add more tests for that class in particular.

Ping me when you're done :)

src/test/java/spoon/test/template/TemplateTest.java
Rohitesh-Kumar-Jain: Yes, I have changed the contract, I thoughtt I was supposed to write what the method is doing in the contract, than what is test is supposed to do.

> test actually only verifies that the field has the correct name, so it doesn't verify the field as a whole.

Yes, I have added another assertion for checking the assignment of the field.

src/test/java/spoon/test/template/TemplateTest.java
```diff
@@ -900,6 +901,28 @@ public void testSubstituteInnerClass() {
...
+		CtType<?> targetType = factory.Class().create("someClass");
+		StatementTemplate template = new InsertAllFieldsTemplate("someParameter");
+
+		Substitution.insertAllFields(targetType, template);
+
+		try {
+			assertEquals("testString", targetType.getFields().get(0).getSimpleName());
+		} catch (IndexOutOfBoundsException e) {
+			fail();
+		}
```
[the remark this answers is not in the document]
Rohitesh-Kumar-Jain: I have removed that from the code, will keep this in mind in the future.

src/test/java/spoon/test/template/testclasses/InsertAllFieldsTemplate.java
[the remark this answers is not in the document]
Rohitesh-Kumar-Jain: Good catch 😅

Rohitesh-Kumar-Jain:
>  I think this also calls for creating a new test class, SubstitutionTest

Yes, I have created a new test class for SubstitutionTest, and the PR is once again ready to review @slarse.

src/test/java/spoon/test/template/SubstitutionTest.java
```diff
@@ -0,0 +1,60 @@
...
+        assertEquals(expectedFieldNames, actualFieldNames);
+
+        List<String> expectedFieldAssignments = Collections.singletonList("\"goodName\"");
+        List<String> actualFieldAssignments = new ArrayList<>();
+        for (CtField<?> ctField : targetType.getFields()) {
+            CtExpression<?> assignment = ctField.getAssignment();
+            actualFieldAssignments.add(String.valueOf(assignment));
+        }
+        // verifies that the field has correct assignment
+        assertEquals(expectedFieldAssignments, actualFieldAssignments);
```
slarse, line 58: I might have tripped you up here: I think that just checking that the name is correct is actually sufficiently strong for this test. So I wasn't sufficiently clear in my comment. But I think this can be a good learning experience on how to use Spoon, so I'll give you additional feedback here.

If you want to check that the field in its entirety is correct (which is indeed a stronger test oracle), then you're still missing the type here. And potential modifiers. There's a _lot_ to a field other than just the name and assignment.

As we're only looking for a single field here, a good solution is to construct the expected field, like so:

```suggestion
        Factory factory = spoon.getFactory();

        CtField<String> expectedField = factory.createField();
        expectedField.setSimpleName("testString");
        expectedField.setAssignment(factory.createLiteral("goodName"));
        expectedField.setType(factory.Type().stringType());

        CtType<?> targetType = factory.Class().create("someClass");
        StatementTemplate template = new SingleFieldTemplate();

        Substitution.insertAllFields(targetType, template);

        assertEquals(Collections.singletonList(expectedField), targetType.getFields());
```

src/test/java/spoon/test/template/SubstitutionTest.java
```diff
@@ -0,0 +1,60 @@
...
+        CtType<?> targetType = factory.Class().create("someClass");
+        StatementTemplate template = new SingleFieldTemplate();
+
+        Substitution.insertAllFields(targetType, template);
+
+        List<String> expectedFieldNames = Collections.singletonList("testString");
+        List<String> actualFieldNames = targetType.getFields().stream()
+                .map(CtField::getSimpleName)
+                .collect(Collectors.toList());
+        // verifies that the field has correct name
```
slarse, line 48: This is the kind of comment that I would recommend you to avoid: it just says _what_ the code does. The assertion is entirely clear on its own and should not have a comment explaining what it asserts.

src/test/java/spoon/test/template/SubstitutionTest.java
```diff
@@ -0,0 +1,60 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class SubstitutionTest {
+
+    public class SingleFieldTemplate extends StatementTemplate {
+        String testString = "goodName";
+
+        @Override
+        public void statement() { }
+    }
```
slarse, line 27: This is a personal preference of mine, and others may think otherwise, but I would put this _below_ the test. I like to read files top-down, so if a method A uses a method or class B which is defined in the same file, then A is defined before B.

Again, I think this is a matter of taste, and you should form your own opinion. If you agree, then move the class down. If you disagree, feel free to keep it where it is.

slarse (changes requested):
Hi @Rohitesh-Kumar-Jain,

So, this is good, but I think it can be even better. I also think that my initial review was a bit unclear and blabbery, and had it been better this would have been ready to merge now. See the review comments for more details.

An additional note: I think this is a test that could benefit from arrange, act and assert comments, because it's not entirely clear at first glance which phase is where, especially when it comes to arrange and act, as the arrangement is so large.

src/test/java/spoon/test/template/TemplateTest.java
slarse: The contract should state any relevant preconditions for the test, method/class/etc is being tested, and what the expected out come is. How one gets from the preconditions to the expected outcome is an implementation detail a test in general is not concerned with :)

Sometimes it may be necessary to test an implementation detail, especially when testing corner cases or reproducing bugs. Then it may be relevant to explain such things in the contract. So, as with everything, this is not a rule set in stone, but rather a general guideline.

Assume that pretty much all of my feedback is either a general guideline, or just my opinion, or both.

src/test/java/spoon/test/template/SubstitutionTest.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class SubstitutionTest {
+
+    public class SingleFieldTemplate extends StatementTemplate {
```
slarse, line 22: Sorry, one thing I missed. This is test input that is only relevant to a single test here, so it would be prudent to make it `private` and `static` (private makes it inaccessible to other test classes, static detaches it from the state of the enclosing class).

```suggestion
    private static class SingleFieldTemplate extends StatementTemplate {
```

src/test/java/spoon/test/template/SubstitutionTest.java
Rohitesh-Kumar-Jain: I will place it downwards with no issues, actually, the first programming language that I learnt was C++, and in that one should declare the function at the top of the file calling the function, so probably I have this habit from there.

Rohitesh-Kumar-Jain:
@slarse feel free to suggest changes for 3rd iteration if the PR needs improvement. I understand the purpose of GSoC is not to churn a lot of code, but to merge quality code in production and for students to learn new things and change their old habits : )

slarse (approved):
> @slarse feel free to suggest changes for 3rd iteration if the PR needs improvement. I understand the purpose of GSoC is not to churn a lot of code, but to merge quality code in production and for students to learn new things and change their old habits : )

This test is now of much higher quality than most other tests in Spoon (including tests I've written myself). I can't speak for the purpose of GSoC as I'm not officially involved, but the code quality here is very good. You're learning, and my reviews will grow shorter and shorter as you do.

src/test/java/spoon/test/template/SubstitutionTest.java
slarse: Totally understand, and as I said it's a matter of taste. Some prefer declaring functions before they are used, while others like me prefer to declare them just after they are used. There's no right or wrong, but in Java I find that the latter is more common.

PR #3998, opened 2021-06-21
review: test: add test for SubstituionInsertAllNestedTypes

Rohitesh-Kumar-Jain:
#1818

Hi, I have added a new test for `insertAllNestedFilelds` method of the `Substitution` class, this test improves code coverage, mutation converge and test strength all from 0% to 100%.

Screenshot 2021-06-21 at 5 12 48 PM

The exception added in this test, is able to kill the mutation but, probably isn't the strongest assertion, I wan't able to make it stronger.

### DIFF
--- a/src/test/java/spoon/test/template/SubstitutionTest.java
+++ b/src/test/java/spoon/test/template/SubstitutionTest.java
@@ -53,7 +53,7 @@ public class SubstitutionTest {
     }
 
     @Test
-    public void testSubstitutionInsertAllNestedTypes() {
+    public void testInsertAllNestedTypes() {
         // contract: Substitution.insertAllNestedTypes inserts the only nested class from a singly nested template into the target class
 
         // arrange
--- a/src/test/java/spoon/test/template/SubstitutionTest.java
+++ b/src/test/java/spoon/test/template/SubstitutionTest.java
@@ -5,14 +5,10 @@ import spoon.Launcher;
 import spoon.reflect.declaration.CtField;
 import spoon.reflect.declaration.CtType;
 import spoon.reflect.factory.Factory;
-import spoon.reflect.reference.CtTypeReference;
-import spoon.reflect.visitor.CtVisitor;
 import spoon.support.compiler.FileSystemFile;
-import spoon.support.reflect.declaration.CtTypeImpl;
 import spoon.template.StatementTemplate;
 import spoon.template.Substitution;
 
-import java.lang.annotation.Annotation;
 import java.util.Collections;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
@@ -71,6 +67,7 @@ public class SubstitutionTest {
 
         // assert
         assertEquals(1, targetType.getNestedTypes().size());
+        assertEquals("nestedClass", targetType.getNestedType("nestedClass").getSimpleName());
     }
 
     private static class SinglyNestedTemplate extends StatementTemplate {
--- a/src/test/java/spoon/test/template/SubstitutionTest.java
+++ b/src/test/java/spoon/test/template/SubstitutionTest.java
@@ -75,16 +75,7 @@ public class SubstitutionTest {
 
     private static class SinglyNestedTemplate extends StatementTemplate {
 
-        public class nestedClass<T extends Annotation> extends CtTypeImpl<T> {
-
-            @Override
-            public boolean isSubtypeOf(CtTypeReference<?> type) {
-                return false;
-            }
-
-            @Override
-            public void accept(CtVisitor visitor) {
-            }
+        class nestedClass {
         }
 
         @Override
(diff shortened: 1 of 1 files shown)


Rohitesh-Kumar-Jain:
Link to #3967

src/test/java/spoon/test/template/SubstitutionTest.java
```diff
@@ -47,4 +51,44 @@ public void testSubstitutionInsertAllFields() {
         @Override
         public void statement() { }
     }
+
+    @Test
+    public void testSubstitutionInsertAllNestedTypes() {
```
slarse, line 56: It's already in the `SubstitutionTest` class, so no need to name all tests something with `Substitution`.

```suggestion
    public void testInsertAllNestedTypes() {
```

src/test/java/spoon/test/template/SubstitutionTest.java
```diff
@@ -47,4 +51,44 @@ public void testSubstitutionInsertAllFields() {
...
+        Factory factory = spoon.getFactory();
+
+        CtType<?> targetType = factory.Class().create("goodClass");
+        StatementTemplate template = new SinglyNestedTemplate();
+
+        // act
+        Substitution.insertAllNestedTypes(targetType, template);
+
+        // assert
+        assertEquals(1, targetType.getNestedTypes().size());
```
slarse, line 73: This assertion is a bit too weak. At least check that the inserted type has the expected name.

src/test/java/spoon/test/template/SubstitutionTest.java
```diff
@@ -47,4 +51,44 @@ public void testSubstitutionInsertAllFields() {
...
+
+            @Override
+            public boolean isSubtypeOf(CtTypeReference<?> type) {
+                return false;
+            }
+
+            @Override
+            public void accept(CtVisitor visitor) {
+            }
+        }
```
slarse, line 88: As with your first attempt with the last template, there's a lot going on here that doesn't seem related to the test. To pass the test, the minimal required nested type is this:

```suggestion
        class nestedClass {
        }
```

In general, make tests as minimal as possible, and don't include things that aren't used. Is there any reason for including the other things? If there is, then you should attempt to use them in assertions such that it's obvious they're there for a reason.

slarse (changes requested):
Looks good, I just have some minor comments on this one.

src/test/java/spoon/test/template/SubstitutionTest.java
Rohitesh-Kumar-Jain: yeah actually originally I was aiming for stronger assertion than size, but wasn't able to think of it 😅

src/test/java/spoon/test/template/SubstitutionTest.java
Rohitesh-Kumar-Jain: Yes sure, I will push a commit to rectify that

Rohitesh-Kumar-Jain:
@slarse I would be glad if you may have again have a look at the PR : )

src/test/java/spoon/test/template/SubstitutionTest.java
```diff
@@ -47,4 +47,36 @@ public void testSubstitutionInsertAllFields() {
...
+
+        CtType<?> targetType = factory.Class().create("goodClass");
+        StatementTemplate template = new SinglyNestedTemplate();
+
+        // act
+        Substitution.insertAllNestedTypes(targetType, template);
+
+        // assert
+        assertEquals(1, targetType.getNestedTypes().size());
+        assertEquals("nestedClass", targetType.getNestedType("nestedClass").getSimpleName());
```
slarse, line 70: The new assertion strengthens the test case in the sense that it verifies that there's a nested type with name `nestedClass`, and that's what was requested. So I think this is fine now.

However, I just wanted to note that the layout of the assertion makes it look like a test for `CtType.getNestedType(String)`. To be clear, it looks like the assertion verifies that getting a nested type by a given name returns a type with that name. It would be sufficient here to just verify that the returned type is non-null, due to the functionality of `getNestedType`.

```suggestion
        assertNotNull(targetType.getNestedType("nestedClass"));
```

Personally, I would probably have written both assertions like this instead, so as not to have to query `targetType` multiple times.

```java
        List<CtType<?>> nestedTypes = targetType.getNestedTypes();
        assertThat(nestedTypes.size(), equalTo(1));
        assertThat(nestedTypes.get(0), equalTo("nestedClass"));
```

slarse (approved):
Looks good to me! Just had a minor note on the new assertion's layout. But I think this is fine to merge, if you want to fix anything up at a later time you can just come back here.

src/test/java/spoon/test/template/SubstitutionTest.java
Rohitesh-Kumar-Jain: Thanks for the suggestions, so I can add this in the PR that I will raise for another new test?

PR #4004, opened 2021-06-24
fix(spoon-control-flow): fix disconnected graphs for switch statements

### LINKED ISSUE #3787: Disconnected control-flow graph when building a CFG for a method with a switch case statement
Hi, i'm using spoon-control-flow graph to build a cfg for a method which contains a switch case statement and the control flow graph i'm getting is disconnected (There is an edge missing).

Here is the function i'm building CFG for: 

```
public void fun()
    {
        int a = 10;
        int b=0;
        switch(a)
        {
            case 1:
            b=10;
            case 2:
            b=20;
        }
        System.out.println("Not Connected");
    }
```

Code for building the CFG: 

```
val builder = ControlFlowBuilder()
val cfg = builder.build(ctMethod) // ctMethod corresponds to the above method
cfg.simplify()
cfg.toGraphVisText()
```

The generate graph: 
test

There is an edge missing between statement: 70 and statement: 72

andrewbwogi:
closes #3787

### DIFF
--- a/spoon-control-flow/src/main/java/fr/inria/controlflow/ControlFlowBuilder.java
+++ b/spoon-control-flow/src/main/java/fr/inria/controlflow/ControlFlowBuilder.java
@@ -768,7 +768,6 @@ public class ControlFlowBuilder implements CtVisitor {
 		breakingBad.push(convergenceNode);
 
 		lastNode = switchNode;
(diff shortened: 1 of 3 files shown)


monperrus:
Thanks @andrewbwogi

PR #4006, opened 2021-06-26
review: test: add tests for the SpoonTreeBuilder class

Rohitesh-Kumar-Jain:
#1818

Hi, I have created a new class `SpoonTreeBuilderTest` for the `SpoonTreeBuilder` class, as per Descartes's report the testing for this class was zero in all parameters hence tests were to be written from scratch.

Report before these tests : 

Screen Shot 2021-06-26 at 23 01 07-fullpage

These tests improve the line coverage from 0 to 60% and improve mutation coverage and strength to 60% as well.

There are testing methods `enter`, `exit`, `create` node.

Link to #3967

### DIFF
--- a/src/test/java/spoon/support/gui/SpoonTreeBuilderTest.java
+++ b/src/test/java/spoon/support/gui/SpoonTreeBuilderTest.java
@@ -1,0 +1,53 @@
+package spoon.support.gui;
+
+import org.junit.jupiter.api.BeforeEach;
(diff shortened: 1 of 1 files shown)


Rohitesh-Kumar-Jain:
@nharrand, @monperrus I would be glad to know if you may suggest how we can make this test better and stronger.

nharrand:
LGTM.

PR #4008, opened 2021-06-27
Review: bug(*): fix method resolution for inner interfaces

MartinWitt:
Currently, spoon fails to resolve methods in a strange edge case, with inner interfaces. Using `CtType#getAllExecutable` on the following type produced a `spoon.SpoonException: Cannot create MethodTypingContext for method declared in different ClassTypingContext`

```java
public class BarBaz implements FooBar.Crashy {

  @Override
  public String foo() {
    return "bar!";
  }

}
```
In java, all nested interfaces are implicitly static. This fixes the method `CtType#getAllExecutables` and handles now inner interfaces correctly as implicitly static.  I'm unsure about the location of the test case, and it's contract. Any suggestions?

This bug and the minimal example was discovered/created  by @I-Al-Istannen

### DIFF
--- a/src/test/java/spoon/test/method/MethodTest.java
+++ b/src/test/java/spoon/test/method/MethodTest.java
@@ -221,15 +221,4 @@ public class MethodTest {
 				() -> method.addFormalCtTypeParameterAt(1, typeParam));
 	}
 
-	@Test
-	public void testGetAllExecutablesOnTypeImplementingNestedInterface() {
-		// contract: implicit static nested interfaces are correct handled in getAllExecutables and dont throw an error.
-		Launcher launcher = new Launcher();
-		launcher.addInputResource("src/test/resources/extendsStaticInnerType");
-		CtModel model = launcher.buildModel();
-		CtType<?> type = model.getAllTypes().stream().filter(v -> v.getSimpleName().contains("BarBaz")).findAny().get();
-		assertDoesNotThrow(() -> type.getAllExecutables());
-		// on jdk8 there are 14 types on newer 13 because a method in Object.java was removed
-		assertTrue(13 == type.getAllExecutables().size() || 14 == type.getAllExecutables().size());
-	}
 }
--- a/src/test/java/spoon/test/method/MethodTest.java
+++ b/src/test/java/spoon/test/method/MethodTest.java
@@ -222,7 +222,7 @@ public class MethodTest {
 	}
 
 	@Test
-	public void testGetAllMethodsInnerClassExtended() {
+	public void testGetAllExecutablesOnTypeImplementingNestedInterface() {
 		// contract: implicit static nested interfaces are correct handled in getAllExecutables and dont throw an error.
 		Launcher launcher = new Launcher();
 		launcher.addInputResource("src/test/resources/extendsStaticInnerType");
--- a/src/test/java/spoon/test/method/MethodTest.java
+++ b/src/test/java/spoon/test/method/MethodTest.java
@@ -18,6 +18,7 @@ package spoon.test.method;
 
 import org.junit.Test;
 import spoon.Launcher;
+import spoon.reflect.CtModel;
 import spoon.reflect.declaration.CtClass;
 import spoon.reflect.declaration.CtMethod;
 import spoon.reflect.declaration.CtParameter;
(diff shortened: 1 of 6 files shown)


src/test/java/spoon/test/method/MethodTest.java
```diff
@@ -218,4 +220,16 @@ public void test_addFormalCtTypeParameterAt_throwsOutOfBoundsException_whenPosit
 		assertThrows(IndexOutOfBoundsException.class,
 				() -> method.addFormalCtTypeParameterAt(1, typeParam));
 	}
+
+	@Test
+	public void testGetAllMethodsInnerClassExtended() {
```
slarse, line 225:
```suggestion
	public void testGetAllExecutablesOnTypeImplementingNestedInterface() {
```
?

src/test/java/spoon/test/method/MethodTest.java
```diff
@@ -218,4 +220,16 @@ public void test_addFormalCtTypeParameterAt_throwsOutOfBoundsException_whenPosit
...
+		// contract: implicit static nested interfaces are correct handled in getAllExecutables and dont throw an error.
+		Launcher launcher = new Launcher();
+		launcher.addInputResource("src/test/resources/extendsStaticInnerType");
+		CtModel model = launcher.buildModel();
+		CtType<?> type = model.getAllTypes().stream().filter(v -> v.getSimpleName().contains("BarBaz")).findAny().get();
+		assertDoesNotThrow(() -> type.getAllExecutables());
+		// on jdk8 there are 14 types on newer 13 because a method in Object.java was removed
+		assertTrue(13 == type.getAllExecutables().size() || 14 == type.getAllExecutables().size());
```
slarse, line 233: I have two comments on these assertions.

First of all, `assertDoesNotThrow` is redundant as, if `type.getAllExecutables()` throws an exception, then the `assertTrue` will surely fail. So, if the test is meant to verify only that `getAllExecutables()` does not throw, then keep only the `assertDoesNotThrow` assertion. If the test is meant to verify the amount of returned executables (which is a stronger oracle than that it does not throw), then keep only the `assertTrue`.

Second, I strongly advice against using `assertTrue` for anything but actual boolean values, because when the test fails you only get "false", and not what the values actually were. If the test should verify the amount of returned executables, then I would refactor it like so:

```suggestion
		int expectedNumExecutablesInJDK8 = 13;
		int expectedNumExecutablesPostJDK8 = 14;

		int numExecutables = type.getAllExecutables().size();

		assertThat(numExecutables, anyOf(
				equalTo(expectedNumExecutablesInJDK8),
				equalTo(expectedNumExecutablesPostJDK8))
		);
```

src/test/java/spoon/test/method/MethodTest.java
```diff
@@ -218,4 +220,16 @@ public void test_addFormalCtTypeParameterAt_throwsOutOfBoundsException_whenPosit
 		assertThrows(IndexOutOfBoundsException.class,
 				() -> method.addFormalCtTypeParameterAt(1, typeParam));
 	}
+
+	@Test
+	public void testGetAllMethodsInnerClassExtended() {
+		// contract: implicit static nested interfaces are correct handled in getAllExecutables and dont throw an error.
```
slarse, line 226: Depending on whether you want the test to verify that there's no crash, or that the correct amount of executables are returned, I would update the contract to state that.

Like I mentioned, invoking a method in a test implicitly tests that it does not crash, and so if you test anything in addition to the fact that the method does not crash (e.g. that it returns the expected amount of executables), the contract does not need to state that it tests that the method doesn't crash. That was a messy sentence but I hope you understand what I mean.

slarse (changes requested):
Hi @MartinWitt!

The fix LGTM, but I think the test is a bit unclear in purpose. See what you think of my comments.

As this is a test of `CtType.getAllExecutables()`, I would place the test in the `CtTypeTest` test class.

src/test/java/spoon/test/method/MethodTest.java
MartinWitt: It was a crash while parsing the JDK. I wanted to include a real assertion to guarantee that it's not only not crashing, but working correctly. But you are correct, "Does not crash" is not a useful assertion if we can check if the result is correct.

src/test/java/spoon/test/method/MethodTest.java
MartinWitt: Good suggestion and way better than my abuse of `assertTrue`

slarse (approved):
Looks good!

PR #4010, opened 2021-06-27
Review: bug: NPE while querying sneaky throws

MartinWitt:
This PR fixes the `CtType#getAllExecutable` method for so called sneaky throws constructs.


monperrus:
Thanks @MartinWitt !

PR #4012, opened 2021-06-28
reveiw: test: add test for Substituion.insertAllSuperInterfaces

Rohitesh-Kumar-Jain:
#1818

Hi, I have added a new test for `insertAllSuperInterfaces` method of the `Substitution` class, following are the features of this test:

1) The test kills this mutation

Screenshot 2021-06-28 at 2 20 29 PM

2) It is killing this mutation as well

Screenshot 2021-06-28 at 2 21 23 PM

Link to #3967

### DIFF
--- a/src/test/java/spoon/test/template/SubstitutionTest.java
+++ b/src/test/java/spoon/test/template/SubstitutionTest.java
@@ -25,11 +25,7 @@ public class SubstitutionTest {
         // contract: Substitution.insertAllFields inserts the only field from a single-field template into the target class
 
         // arrange
-        Launcher spoon = new Launcher();
-        spoon.addTemplateResource(new FileSystemFile("./src/test/java/spoon/test/template/SubstitutionTest.java"));
-
-        spoon.buildModel();
-        Factory factory = spoon.getFactory();
+        Factory factory = createFactoryWithTemplates();
 
         CtField<String> expectedField = factory.createField();
         expectedField.setSimpleName("testString");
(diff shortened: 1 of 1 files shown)


src/test/java/spoon/test/template/SubstitutionTest.java
```diff
@@ -114,4 +116,36 @@ public SingleConstructorTemplate() { }
...
+    @Test
+    public void testInsertAllSuperInterfaces() {
+        // contract: Substitution.insertAllSuperInterfaces inserts the only superInterface from a single interface implementing template into the target class
+
+        // arrange
+        Launcher spoon = new Launcher();
+        spoon.addTemplateResource(new FileSystemFile("./src/test/java/spoon/test/template/SubstitutionTest.java"));
+
+        spoon.buildModel();
+        Factory factory = spoon.getFactory();
```
slarse, line 129: I'm starting to see a pattern here: these four lines occur in every single test case of this class, but only the factory is ever used. I suggest extracting this into a helper method like so:

```java
    private static Factory createFactoryWithTemplates() {
        spoon.addTemplateResource(new FileSystemFile("./src/test/java/spoon/test/template/SubstitutionTest.java"));
        spoon.buildModel();
        return spoon.getFactory();
    }
```

And then use that in the arrange part of the tests. In the interest of not piling up PRs that depend on each other, I think it's fine to do that refactoring in this PR. What do you think?

slarse (changes requested):
Test LGTM, but there's some duplication in the tests that we might want to address. Have a look at my comment.

src/test/java/spoon/test/template/SubstitutionTest.java
Rohitesh-Kumar-Jain: Yeah, this makes sense.

I have one question, where should I place this new method? below all the tests I guess?

slarse: Yeah, I would place it at the bottom.

Rohitesh-Kumar-Jain: @slarse I think it's fixed now

slarse: Yep, looks good!

slarse (approved):
Looks good, will merge!

PR #4015, opened 2021-06-28
feat: Add API method CtAbstractSwitch.addCaseAt

algomaster99:
Reference: https://github.com/INRIA/spoon/issues/3884#issuecomment-869403561.

### DIFF
--- a/src/main/java/spoon/reflect/code/CtAbstractSwitch.java
+++ b/src/main/java/spoon/reflect/code/CtAbstractSwitch.java
@@ -62,6 +62,11 @@ public interface CtAbstractSwitch<S> extends CtElement {
 
 	/**
 	 * Adds a case at the specified position.
+	 *
+	 * @param <T>      type of the switch - {@link CtSwitch} or {@link CtSwitchExpression}
+	 * @param position index at which the case needs to be inserted
+	 * @param c        case which has to be inserted in the switch block
+	 * @return         switch block in which the case is inserted
 	 */
 	@PropertySetter(role = CASE)
 	<T extends CtAbstractSwitch<S>> T addCaseAt(int position, CtCase<? super S> c);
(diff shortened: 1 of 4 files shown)


algomaster99:
I had a few questions, a little unrelated to this PR, about two APIs in `CtCase` - [`setCaseExpression`](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/reflect/code/CtCase.java#L45) and [`addCaseExpression`](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/reflect/code/CtCase.java#L66). What is the difference between their purposes? As far as I can see, the former modifies `this.caseExpressions` while the latter modifies `this.caseExpression`. Do we need to have the attribute `this.caseExpression`? I understand before Java 12 only one expression was allowed per case so maybe that's why `this.caseExpression` exists but now we can implement `setCaseExpression` to clear the list and just add one expression to the list of case expressions. Similary, `getCaseExpression` can just return the first element of the list.

I was using the `setCaseExpression` before and was trying to pretty-print `CtCase` after that but I did not see any expression in there because [the visitor](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/reflect/visitor/CtScanner.java#L330-L337) doesn't even look for `this.caseExpression`. The above suggestion can help resolve this.

monperrus:
LGTM, will merge ;)

algomaster99:
@monperrus squash and merge it because some of the commits in between don't pass the CI tests so we don't want to break the master at any point. Moreover, the commit messages don't follow the spoon conventions too but the PR title does.

slarse:
@algomaster99 We always squash :)

slarse:
> I understand before Java 12 only one expression was allowed per case so maybe that's why this.caseExpression exists but now we can implement setCaseExpression to clear the list and just add one expression to the list of case expressions.

Agree, open a separate issue/PR about that.

monperrus:
Thanks a lot @algomaster99

PR #4016, opened 2021-06-29
test: unite all tests for Substitution in one place

Rohitesh-Kumar-Jain:
As I created a new class `SubstitutionTest` and merged 4 new tests in it, I thought I should extract two existing tests so all tests can be together which were Specifically written for the `Substitution` class.

Now all tests related to this class are at the same place

link to #3967


monperrus:
Thanks @Rohitesh-Kumar-Jain

src/test/java/spoon/test/template/TemplateTest.java
```diff
@@ -816,23 +808,6 @@ public void testTemplateInvocationSubstitution() {
...
-		Factory factory = spoon.getFactory();
-
-		CtClass<?> testSimpleTpl = factory.Class().create("TestSimpleTpl");
-		//whitespace seems wrong here
-		new SimpleTemplate("HelloWorld").apply(testSimpleTpl);
-
-		Set<CtMethod<?>> listMethods = testSimpleTpl.getMethods();
-		assertEquals(0, testSimpleTpl.getMethodsByName("apply").size());
-		assertEquals(1, listMethods.size());
-	}
```
slarse, line 834: This is not a direct test of `Substitution.java`, strictly speaking it should not have been moved. This is a higher-level test on a whole template.

Rohitesh-Kumar-Jain: @slarse  I thought it is was probably meant for testing Substitution.insertAll(), but was testing Substitution.insertAllMethods() only. That was the use case of the test I got after going through it.

slarse: The method under test here is `StatementTemplate.apply()`. While that method in turn uses `Substitution`, it also uses a bunch of other classes and methods, so saying that this is a test of the `Substitution` class in isolation is a stretch of the imagination.

This is an end-to-end test of using a template, and so is perfectly suitable to be located in `TemplateTest`.

Rohitesh-Kumar-Jain: I'm sorry. I hurried to push the PR, I see that this extraction was wrong and unnecessary. I will be more sincere next time, it won't happen again, and I will fix this soon in the other PR which is already open for this class.

slarse: You don't need to be sorry, I just noted that moving this seemed a bit off and made a note of it. A PR in Spoon is (at least) a two-person job, and both you and @monperrus missed this detail in this case. And that's OK, we can fix it later.

> it won't happen again

It might happen again, and that's fine. We all make mistakes from time to time, especially when the "correct" solution isn't clear-cut (which with testing it rarely is).

Issue #3653, opened 2020-10-16
add support for Java 16


monperrus:
What’s coming in Java 16?
https://www.infoworld.com/article/3569150/jdk-16-whats-coming-in-java-16.html#tk.rss_java

* [x] Records ([JEP 395](https://openjdk.java.net/jeps/395))
* [x] instanceof pattern matching ([JEP 394](https://openjdk.java.net/jeps/394))

MartinWitt:
https://openjdk.java.net/projects/jdk/16/ as the official link, if somebody wants to read all jeps. From my understanding in December after rampdown phase 1 the feature list is more or less final and we only have to wait for jdt. I would expect records and instanceOf to reach the release state, and a preview for pattern matching (for more about pattern matching you can look at old languages like haskell)

monperrus:
The official Java 16 spec: https://docs.oracle.com/javase/specs/jls/se16/jls16.pdf

Rohitesh-Kumar-Jain:
Going through the Java 16 specs, one of the possible changes, I identified is [JEP 392: Packaging Tool](https://openjdk.java.net/jeps/392). 

jpackage tool is moved from jdk.incubator.jpackage to jdk.jpackage

slarse:
Having a quick look at the release notes, it seems like there are only 2 new stable features and 1 new preview feature that require extensions in Spoon. IMO preview features are not worth supporting, though.

### Stable features

* Records ([JEP 395](https://openjdk.java.net/jeps/395))
    - We need a new `CtRecord` interface to model it
    - Note: This is pure syntactical sugar, it just generates a class with some stuff (e.g. getters for fields, an equals method, a toString, and some other stuff). Pretty much built-in Lombok.
* instanceof pattern matching ([JEP 394](https://openjdk.java.net/jeps/394))
    - Allows an optional variable to cast to the checked type (if check is successful)
    - E.g. `obj instanceof String s` instantiates `s` to `(String) obj` iff `obj` is in fact a `String`
    - May require a new `CtInstanceOf` interface that subtypes `CtBinaryOperator` to make sense

### Preview features

* Sealed types (preview 2, [JEP 397](https://openjdk.java.net/jeps/397))
    - Classes and interfaces can restrict how they can be subtyped with the `sealed` modifier
    - IMO not worth supporting until stable

jakobbraun:
Can you give an estimation when java 16 support will be available? It would be really great to have this.

slarse:
[...]

PR #4020, opened 2021-06-30
review: feature: add support for pattern matching in Java 16 (JEP 394)

SirYwell:
This is a design to support the new pattern matching features that were finalized in Java 16. It addresses #3653 partly. 
As I'm not that familiar with the whole spoon project yet, there might be misunderstandings regarding some concepts from my side. Please let me know if anything is off.

**Design Notes**:
- As there are more pattern matching features planned ([JEP 405: Record Patterns & Array Patterns (Preview)](https://openjdk.java.net/jeps/405) and [JEP 406: Pattern Matching for switch (Preview)](https://openjdk.java.net/jeps/406)) it might make sense to have a common interface for different pattern matching variants. I had a `CtPattern` in the beginning but it ended up empty, so I decided to remove it again. If it can provide common functionality in future, it can be added back I guess.
- In the current [JLS](https://docs.oracle.com/javase/specs/jls/se16/html/jls-14.html#jls-Pattern), a `LocalVariableDeclaration` *is* a `TypePattern` (from my understanding). In spoon, that would make a `CtLocalVariable` to an `CtExpression`, as `CtBinaryOperator` wants a `CtExpression` as right hand operand. I decided to make `CtLocalVariable` a property of `CtTypePattern` instead to avoid issues with that.
- This design is (meant to be) backwards compatible. Other approaches like specialising `CtBinaryOperator` would likely cause issues in that respect (at least everything I came up with).
- JDT only [added a `LocalDeclaration` field to their `InstanceOfExpression`](https://github.com/eclipse/eclipse.jdt.core/blob/master/org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/ast/InstanceOfExpression.java#L41). This seems to be less future-proof to me as the design proposed in this PR.

**Remaining questions/issues from my side**:
- The generated code partly changed license headers. I don't know how to deal with that (and why it is like that in the first place).
- There is one remaining TODO in `CtTypePatternImpl`, as I wasn't sure if that's how the `onObjectUpdate` method is used correctly.
- Generating and replacing the code in `Metamodel` produced a large diff (seems like `EMODIFIER` was missing in multiple places, while other roles were just moved a few lines up or down). I fixed/adapted the indendation differences, but I'm not sure how to deal with the rest of it.
- Edit (2021-07-02): I needed to bump the spoon version used in the other modules to a `SNAPSHOT` version, Is it supposed to be like that?

Thanks to @MartinWitt and @I-Al-Istannen for helping with the design decisions and issues I came accross.


slarse:
Hi @SirYwell , thanks for starting to work on this!

Just as an FYI, this is a false positive that you can ignore: https://github.com/INRIA/spoon/pull/4020/checks?check_run_id=2963075760#step:3:848

Don't know why that is being flagged as a dead store, it's not even a remotely complicated case.

SirYwell:
Thanks for letting me know, I already wondered how I should deal with such issues in the auto generated code.

SirYwell:
The Java 8 test on windows timed out, I could re-run it by pushing a new commit again but at this point I'll just wait for feedback.

slarse:
Getting some very strange errors in the CI, looks like backend problems on GitHub. Trying to rerun a second time :)

SirYwell:
As the only remaining CI issue is the one you mentioned earlier already, I'm happy to receive any feedback on the design and my questions. I hope I didn't forget anything.

slarse:
Hi @SirYwell,

I'll do my very best to get back to this during the week, I'm just a little bit preoccupied with a hard deadline or two. Rest assured that I _will_ get to this issue and we're very happy you took the time to add this feature.

@monperrus @nharrand If either of you find the time, please do have an initial look at the design here. I will get to this as soon as I'm able (and I do want to chip in because this is kind of a "big thing" :)

monperrus:
Just had a look. First of all, congrats for mastering the complexity of adding a metamodel type! (with all the code generators involved).

There are indeed a few design alternatives, let me think about it.

slarse:
I've also had a first look now. Like @monperrus said, very impressive that you managed to extend the metamodel, that's really not an easy thing to do.

Before going into any kind of detail here, I have one major concern with the design: a pattern, and by extension a type pattern, is _not_ an expression. That is quite evident from looking at the [syntax of Java 16](https://docs.oracle.com/javase/specs/jls/se16/html/jls-19.html). Patterns are only ever allowed in `instanceof` expressions. The problem is that, given this design, I can put a pattern where it does not belong (anywhere I can put an expression of arbitrary type). This significantly weakens the safety of using Spoon. Ideally, we want an AST that type checks to be structurally sensible.

`CtExpression<Void>` also doesn't really make sense to me. I know it's used for `CtTypeAccess`, but it's a hack IMO. An expression, by definition, must yield a value. A void expression would then be an expression that does not yield a value, which I can't really reconcile.

I think the primary shortcoming in Spoon that makes this JEP difficult to support is that operator specializations are represented by enums. If they were not, we could simply extend `instanceof` like they did in JDT. Overall, enum usage in Spoon has been a massive pain point, modifiers in particular, and now we see it for operator kinds as well.

@SirYwell I think you've done a really good job in concern to the constraints you've placed, especially in terms of being backwards compatible. Now we'll iterate on this until we're all satisfied :)

@monperrus Thoughts?

SirYwell:
> Before going into any kind of detail here, I have one major concern with the design: a pattern, and by extension a type pattern, is not an expression.

Yes, that is my main pain point too. I also thought about having something like an adapter that wraps a Pattern into an Expression to keep it compatible. This way, there's more difference, but specialising the BinaryOperator to accept a Pattern doesn't sound better to me.

> Patterns are only ever allowed in `instanceof` expressions.

While this is true for now, looking at [JEP 406](https://openjdk.java.net/jeps/406) indicates that Patterns will be allowed as cases too at some point. This is another reason to make a `CtPattern` a `CtExpression`, as `CtCase` only accepts `CtExpression`s at the moment. Keeping that in mind will help in future I think.

A general issue is that another case was added to the `InstanceofExpression`. While it was only `RelationalExpression instanceof ReferenceType` before, we now have a `RelationalExpression instanceof Pattern` too. Representing that in Java without a common super type is basically impossible.

Another approach I didn't try would be to create a common super type (like `CtExpressionOrPattern`) for CtExpression and CtPattern and accept this type in a specialisation of the binary operator. This would require some more type magic under the hood (like `CtBinaryOperator` still giving you a `CtExpression`) and probably cause more work in future, and as I didn't try it, there might be even more caveats I'm not aware yet.

slarse:
> A general issue is that another case was added to the InstanceofExpression. While it was only RelationalExpression instanceof ReferenceType before, we now have a RelationalExpression instanceof Pattern too. Representing that in Java without a common super type is basically impossible.

This isn't a problem, we'd just have a potentially empty field for the pattern. Syntax and Spoon's type hierarchy doesn't necessarily need to be the same (in a lot of cases they're not), it's just sometimes a neat guide.

> Another approach I didn't try would be to create a common super type (like CtExpressionOrPattern) for CtExpression and CtPattern and accept this type in a specialisation of the binary operator. This would require some more type magic under the hood (like CtBinaryOperator still giving you a CtExpression) and probably cause more work in future, and as I didn't try it, there might be even more caveats I'm not aware yet.

This isn't really doable in Java, as it does not support union types. Necessarily, an `AorB` type would need to either need to implement both types, or have getters for the precise type it defines (like `AorB.getA()` would return an `A` if it's actually an `A`). It's a mess. The former is a mess, the latter is breaking.

Anyway, @monperrus and I discussed this in some detail yesterday. I opened #4035 with a summary of the discussion. In light of patterns to be used as switch cases, and especially the guarded patterns (where the pattern appears in an operator), my conclusion is that the only way to prepare reasonably for that is to make patterns expressions. @monperrus I kind of shifted my stance when I was tallying up the suggestions :)

While we then don't have any type safety for patterns, we can add fail-fast mechanisms at runtime. For example, when setting the parent of a pattern, we could check that the parent is one of the allowed types (currently, only binary operators with type `instanceof`).

slarse:
I do suggest that we add a `CtPattern` type as a common supertype of patterns, as type patterns, record patterns and array patterns are fundamentally different in content (see JEP 405 linked in the opening post).

* Type pattern: Just a local variable declaration
* Record pattern: A destructuring declaration (e.g. `Point(int x, int y)`) that captures the fields of a record. Definitely not convinient to model with a local variable.
* Array pattern: Also destructuring declaration (e.g. `String[] { String s1, String s2, String s3 }`), again I don't see an easy way to model it with a local variable declaration.

SirYwell:
Hi, first of all, thanks for the in-depth analysis of the different approaches.

> I do suggest that we add a `CtPattern` type as a common supertype of patterns

Yes, that's what I had before, but it remained empty so I removed it again (it could be introduced at any point I guess). I'm happy to add it back, but I'm wondering if it makes sense to make it extend `CtExpression` or if that should be added to the specific patterns. With the latter, it may be more clear that patterns aren't expressions, although the practical difference for users will be basically non-existent. WDYT?

I'm against modelling different patterns in one interface, because that could become messy. The Java developers are looking into a lot more pattern matching possibilities (constants, deconstructing patterns for normal classes, [static patterns](https://cr.openjdk.java.net/~briangoetz/amber/pattern-match.html#static-patterns), ...). While this is far from being part of the language, I'd expect such features to come at some point.

slarse:
> Yes, that's what I had before, but it remained empty so I removed it again

Yeah, and chances are it will remain empty, but I still think we need that common supertype.

> I'm happy to add it back, but I'm wondering if it makes sense to make it extend CtExpression

I'm not certain. I don't think it would hurt to be initially conservative and just have `CtPattern` be a `CtElement`, and then have e.g. `CtTypePattern` extend both that and `CtExpression<Void>`. It's possible that the only use of `CtPattern` will be to check if something is a pattern or not (i.e. `e instanceof CtPattern`), and to logically group the different patterns together in the metamodel.

SirYwell:
Hi, I came across an issue when implementing the `CtPattern` interface. The `CtGenerationTest` fails with following error:
```
spoon.SpoonException: spoon.support.reflect.code.CtPatternImpl doesn't have declaration in the source path for scan
CtPattern(spoon.reflect.code.CtPattern)
        at spoon.processing.CtGenerationTest.testGenerateCloneVisitor(CtGenerationTest.java:146)
```
Obviously, there is no `CtPatternImpl`, as I can't extend multiple classes in `CtTypePatternImpl` (and it would be empty too). The fix I'm aware of would be to make `CtPattern` extend `CtExpression`, but that's what we wanted to prevent. Is there another way to not have to implement that interface directly that makes the test happy?

src/test/java/spoon/test/pattern/TypePatternTest.java
```diff
@@ -0,0 +1,51 @@
...
+	private static CtModel createModelFromString(String code) {
+		Launcher launcher = new Launcher();
+		launcher.getEnvironment().setComplianceLevel(16);
+		launcher.getEnvironment().setNoClasspath(true);
+		launcher.addInputResource(new VirtualFile(code));
+		return launcher.buildModel();
+	}
+
+	@Test
+	public void testWithInstanceOf() {
```
slarse, line 33: Add an inline contract comment at the top of this test saying what it's supposed to test, it's not clear to me from the title alone.

```java
// contract: When X and Y then Z happens
```

slarse (changes requested):
Hi @SirYwell,

Sorry about this being a bit slow, all integrators are either away or (like me) busy with other things at the moment. So the final steps of this PR will likely take a little longer than one would like.

There's one final thing I'm thinking about, and that's if we should restrict where type patterns can be inserted. For example, we could override `setParent()` in `CtTypePatternImpl`, and throw an exception if it's added to anything but an `instanceof` binary operator (unfortunately, we can't check that it's added to the RHS as the parent is set before the value is set in the RHS). We could also disallow changing the kind of a `CtBinaryOperator` if its current kind is `instanceof` and the RHS contains a `CtPattern`. @monperrus thoughts about such mechanisms?

src/test/java/spoon/test/pattern/TypePatternTest.java
SirYwell: The test basically just checks if the model is built correctly (not sure if the BinaryOperatorKind even needs to be checked).

In general, it feels like one test isn't enough, I could probably add a test to check the pretty printed output or something like that...

*If* we're going to restrict accepted parent types, there would be more to test.

SirYwell:
> If you're wondering why this is progressing a bit sluggishly on our side, it's because all integrators are either away or (like me) busy with other things at the moment. 

That's totally fine, still better than rushing it and regretting decisions later.

> There's one final thing I'm thinking about, and that's if we should restrict where type patterns can be inserted. [...]

Rather a question: From what I've seen, spoon barely validates the model directly. As an example, it is possible to run
```java
var operator = launcher.getFactory().createBinaryOperator();
var one = launcher.getFactory().createLiteral(1);
var two = launcher.getFactory().createLiteral(2);
operator.setLeftHandOperand(one);
operator.setRightHandOperand(two);
operator.setKind(BinaryOperatorKind.INSTANCEOF);
```
without exception, although this is clearly invalid java code. The situation with patterns might be little bit different, but it would definitely add work in future (e.g., `setParent` needs to allow cases once implemented, but probably only if the compliance level is set accordingly(?)). Would such validation really fit into spoon directly?

slarse:
> Rather a question: From what I've seen, spoon barely validates the model directly.

That is correct. I've however found no trace that this is a conscious design decision. The only viable reason I could see for this is that omitting such validation might yield some performance benefits. The more probable reason is that it's simply hard to implement such checks and keep them consistent.

We can't do much about what's already there, because we'd be breaking existing behavior by putting in new checks. We do have a choice when putting in new stuff, however. However, if we want to implement consistency checks, it may well be better to offer that as a separate "strict" mode where we attach a listener to the AST that does the validation completely independently. The performance overhead will then probably be tangible, but it'll be opt-in and can be implemented incrementally for all elements.

slarse:
@SirYwell Vacations are in full force right now, so very likely we'll have to get back to this PR around mid August. Just know that we haven't forgotten about it!

slarse:
Ping @monperrus, let's try to squeeze in a small discussion of this PR during our meeting tomorrow.

monperrus:
>> From what I've seen, spoon barely validates the model directly.

> That is correct. I've however found no trace that this is a conscious design decision

Let's say that Spoon validates the model on a best-effort basis, when we, contributors, are able to implement and push valuable static or dynamic checks.

slarse:
A reasonable dynamic check here would be for `CtTypePatternImpl` to override `setParent` and check that the parent it's set in is in fact a binary operator with kind `instanceof`. I.e. something like:

```java
@Override
public <E extends CtElement> E setParent(E parent) {
	if (!isInstanceOfBinop(parent)) {
		throw new SpoonException("type pattern can only be used in an instanceof binary operator");
	}
	return super.setParent(parent);
}
```

That won't stop anyone from putting it as the left hand operand, or later changing the operator to some other kind, but I think checking those things are far more effort than they're worth.

@SirYwell how about implementing a check like this, and adding a test for it, and then I think we can wrap this up?

src/test/java/spoon/test/pattern/TypePatternTest.java
slarse: That's all good, but there should still be a contract at the top of the test stating what it's supposed to verify :)

src/main/java/spoon/support/reflect/code/CtTypePatternImpl.java
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	@Override
+	public <C extends CtTypePattern<?>> C setVariable(CtLocalVariable<?> variable) {
+		if (variable != null) {
+			variable.setParent(this);
+		}
+		getFactory().getEnvironment().getModelChangeListener()
+				// TODO is this the proper role? New role needed?
+				.onObjectUpdate(this, CtRole.VARIABLE, variable, this.variable);
```
slarse, line 34: This looks like the correct role to me, @monperrus?

monperrus: LGTM

SirYwell:
> how about implementing a check like this, and adding a test for it, and then I think we can wrap this up?

Done.

I encountered an issue when writing the test though, I needed to add the `Executable` cast, as it tries to cast the `CtTypePattern` to `CtBinaryOperator` otherwise. This is caused by the signature of the `setParent` method: `<E extends CtElement> E setParent(E parent)`,  the return type is expected to be the same as the type of the parent, which seems to be odd... Another way would be to downcast the parent to a common super type, but I'm somehow confused about that.

slarse:
> This is caused by the signature of the `setParent` method: `<E extends CtElement> E setParent(E parent)`, the return type is expected to be the same as the type of the parent, which seems to be odd...

I've noted this as well, I've just been too lazy to do anything about it as it hasn't caused any problems before. It's clearly an error in `CtElement`. See #4098.

> Edit: The SetParentTest#testContract doesn't like the check as it passes a CtCpmditionalImpl. How to deal with that?

I'll have a look, will get back to you after I've seen to #4098.

slarse:
If I'm not mistaken, this is currently blocked by #4099

slarse:
@SirYwell Merge with master to get the update in #4099 

As for the `SetParentTest`, I propose the following changes to it.

```diff
diff --git a/src/test/java/spoon/test/parent/SetParentTest.java b/src/test/java/spoon/test/parent/SetParentTest.java
index 7d9e73eb..fb734572 100644
--- a/src/test/java/spoon/test/parent/SetParentTest.java
+++ b/src/test/java/spoon/test/parent/SetParentTest.java
@@ -20,6 +20,9 @@ import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
 import spoon.reflect.CtModelImpl;
+import spoon.reflect.code.BinaryOperatorKind;
+import spoon.reflect.code.CtBinaryOperator;
+import spoon.reflect.code.CtTypePattern;
 import spoon.reflect.declaration.CtElement;
 import spoon.reflect.declaration.CtMethod;
 import spoon.reflect.declaration.CtType;
@@ -35,7 +38,6 @@ import static org.junit.Assert.assertNotSame;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
-import static spoon.test.parent.ContractOnSettersParametrizedTest.createCompatibleObject;
 import static spoon.test.parent.ContractOnSettersParametrizedTest.createReceiverList;
 import static spoon.testing.utils.ModelUtils.createFactory;
 
@@ -58,12 +60,8 @@ public class SetParentTest<T extends CtVisitable> {
 		Object o = factory.Core().create((Class<? extends CtElement>) toTest.getActualClass());
 		CtMethod<?> setter = factory.Type().get(CtElement.class).getMethodsByName("setParent").get(0);
 
-		Object argument = createCompatibleObject(setter.getParameters().get(0).getType());
+		CtElement argument = createCompatibleParent(toTest);
 
-		if (!(argument instanceof CtElement)) {
-			// is a primitive type or a list
-			throw new AssertionError("impossible, setParent always takes an element");
-		}
 		// we create a fresh object
 		CtElement receiver = ((CtElement) o).clone();
 
@@ -91,7 +89,7 @@ public class SetParentTest<T extends CtVisitable> {
 		}
 
 		Method actualMethod = setter.getReference().getActualMethod();
-		CtElement argumentClone = ((CtElement) argument).clone();
+		CtElement argumentClone = argument.clone();
 		actualMethod.invoke(receiver, new Object[]{argument});
 
 		// contract: the parent has not been changed by a call to setParent on an elemnt
@@ -100,4 +98,15 @@ public class SetParentTest<T extends CtVisitable> {
 
 	}
 
+	private static CtElement createCompatibleParent(CtType<?> e) {
+	    return CtTypePattern.class.getSimpleName().equals(e.getSimpleName())
+				? createInstanceOfBinaryOperator()
+				: e.getFactory().createAssignment();
+	}
+
+	private static CtBinaryOperator<?> createInstanceOfBinaryOperator() {
+		CtBinaryOperator<?> op = factory.createBinaryOperator();
+		op.setKind(BinaryOperatorKind.INSTANCEOF);
+		return op;
+	}
 }
```

SirYwell:
I think we're ready :)

src/main/java/spoon/reflect/code/CtTypePattern.java
```diff
@@ -0,0 +1,69 @@
...
+ */
+public interface CtTypePattern<T> extends CtPattern, CtExpression<Void> {
+
+	/**
+	 * Returns the local variable declared by this type pattern.
+	 *
+	 * @return the local variable.
+	 */
+	@PropertyGetter(role = VARIABLE)
+	CtLocalVariable<?> getVariable();
```
slarse, line 43:
```suggestion
	CtLocalVariable<T> getVariable();
```

src/main/java/spoon/reflect/code/CtTypePattern.java
```diff
@@ -0,0 +1,69 @@
...
+
+	/**
+	 * Sets the local variable for this type pattern.
+	 *
+	 * @param variable the variable to set for this type pattern.
+	 * @param <C> the type of this pattern.
+	 * @return this pattern.
+	 */
+	@PropertySetter(role = VARIABLE)
+	<C extends CtTypePattern<?>> C setVariable(CtLocalVariable<?> variable);
```
slarse, line 53:
```suggestion
	CtTypePattern<T> C setVariable(CtLocalVariable<T> variable);
```

slarse:
> I think we're ready :)

I'm terribly sorry, but I have some more concerns :p.

Functionally, I'm perfectly satisfied with this implementation. There are just two things in the API of `CtTypePattern` that bother me a bit.

### Type variable `T` is not used in a meaningful way
It's only ever used in the return type of `clone`, meaning that it's more or less pointless. As it's supposed to be the type of the local variable, it should be supplied to all uses of `CtLocalVariable` (see comments).

This will also require some adjusted use/implementation of the affected methods.

### The return type of `setVariable` is problematic
The header of `setVariable` is as follows:

```java
	<C extends CtTypePattern<?>> C setVariable(CtLocalVariable<?> variable);
```

I completely understand _why_ you wrote the return type like you did, as it's a prevalent pattern throughout Spoon. This pattern is deeply problematic and IMO the single largest design flaw in Spoon: methods that return `this` but **let the caller decide the return type**, which fundamentally sidesteps the type system and always requires an unchecked cast in the implementation. It's similar to why we had a completely nonsensical return type for `setParent` for years, which was just recently fixed in #4099.

To demonstrate why this is so problematic, I'll give an example with the near identical `CtExpression.setTypeCasts` method, which is defined like so:

```java
public <C extends CtExpression<T>> C addTypeCast(CtTypeReference<?> type) {
```

Let's write some nonsens.

```java
private static void nonsense() {
	CtLiteral<Double> expr = factory.createLiteral(1.2);
	CtBinaryOperator<Double> definitelyNotABinaryOperator =
			expr.addTypeCast(factory.Type().integerPrimitiveType());
	System.out.println(definitelyNotABinaryOperator);
}
```

This type checks just fine. But it of course doesn't make sense (a literal doesn't suddenly become a binary operator by adding a type cast to it), and it blows up at runtime with a class cast exception. Of course, this only becomes a problem if we have subtypes of `CtTypePattern`, but the pattern is also only useful if we have subtypes of `CtTypePattern`, so it's a moot point.

I strongly believe that it's about time we stop using this pattern. I wrote at length about the badness of using patterns like this in #1846. The _correct_ way to do this is to have each extending interface override the method and explicitly specify itself in the return type (methods are covariant in the return type, so this is perfectly fine).

@SirYwell @monperrus thoughts on this?

slarse (commented):
> I think we're ready :)

I'm terribly sorry, but I have some more concerns :p.

Functionally, I'm perfectly satisfied with this implementation. There are just two things in the API of `CtTypePattern` that bother me a bit.

### Type variable `T` is not used in a meaningful way
It's only ever used in the return type of `clone`, meaning that it's more or less pointless. As it's supposed to be the type of the local variable, it should be supplied to all uses of `CtLocalVariable` (see comments).

This will also require some adjusted use/implementation of the affected methods.

### The return type of `setVariable` is problematic
The header of `setVariable` is as follows:

```java
	<C extends CtTypePattern<?>> C setVariable(CtLocalVariable<?> variable);
```

I completely understand _why_ you wrote the return type like you did, as it's a prevalent pattern throughout Spoon. This pattern is deeply problematic and IMO the single largest design flaw in Spoon: methods that return `this` but **let the caller decide the return type**, which fundamentally sidesteps the type system and always requires an unchecked cast in the implementation. It's similar to why we had a completely nonsensical return type for `setParent` for years, which was just recently fixed in #4099.

To demonstrate why this is so problematic, I'll give an example with the near identical `CtExpression.setTypeCasts` method, which is defined like so:

```java
public <C extends CtExpression<T>> C addTypeCast(CtTypeReference<?> type) {
```

Let's write some nonsens.

```java
private static void nonsense() {
	CtLiteral<Double> expr = factory.createLiteral(1.2);
	CtBinaryOperator<Double> definitelyNotABinaryOperator =
			expr.addTypeCast(factory.Type().integerPrimitiveType());
	System.out.println(definitelyNotABinaryOperator);
}
```

This type checks just fine. But it of course doesn't make sense (a literal doesn't suddenly become a binary operator by adding a type cast to it), and it blows up at runtime with a class cast exception. Of course, this only becomes a problem if we have subtypes of `CtTypePattern`, but the pattern is also only useful if we have subtypes of `CtTypePattern`, so it's a moot point.

I strongly believe that it's about time we stop using this pattern. I wrote at length about the badness of using patterns like this in #1846. The _correct_ way to do this is to have each extending interface override the method and explicitly specify itself in the return type (methods are covariant in the return type, so this is perfectly fine).

@SirYwell @monperrus thoughts on this?

SirYwell:
> Type variable `T` is not used in a meaningful way

This comes from a typing issue, I had it in the other places first too. Here is an example of the limitation I encountered:
```java
CtTypePattern<?> pattern = ...;
CtLocalVariable<?> localVariable = ...;
pattern.setVariable(localVariable); // compilation error, the expected `?` capture is a different one
```
the same applies for every other non-compatible types, as expected, but
```java
CtTypePattern<Object> pattern = launcher.getFactory().Core().createTypePattern();
pattern.setVariable(launcher.getFactory().createLocalVariable()); // compiles due to the fact that the type is inferred
```
works. Looking at other classes, it seems to be somewhat inconsistent (e.g. `CtClass#addConstructor` vs `CtEnum#addEnumValue`.

I can change it back using `T` if you want.

> The return type of `setVariable` is problematic

Yes, I only did that for consistency reasons.
> The *correct* way to do this is to have each extending interface override the method and explicitly specify itself in the return type

Another approach would be to use recursive generics, but using that properly would be fully incompatible with existing code and does not solve the problem in every case.

I'll do it whatever way you decide on, as I don't really have a strong opinion on consistency vs type safety here.

slarse:
> > Type variable `T` is not used in a meaningful way
> 
> This comes from a typing issue, I had it in the other places first too. Here is an example of the limitation I encountered:
> 
> ```java
> CtTypePattern<?> pattern = ...;
> CtLocalVariable<?> localVariable = ...;
> pattern.setVariable(localVariable); // compilation error, the expected `?` capture is a different one
> ```

Well, this is the type system doing its job correctly, wildcards defined in different places are not necessarily the same type. The standard pattern to work with when you need to specify the same wildcard in more than one place is to [capture it](https://docs.oracle.com/javase/tutorial/java/generics/capture.html). In this case, you could do something like this:

```java
private static <T> CtTypePattern<T> createTypePattern(CtLocalVariable<T> localVariable) {
	CtTypePattern<T> typePattern = localVariable.getFactory().Core().createTypePattern();
	typePattern.setVariable(localVariable);
	return typePattern;
}

// use like so
CtTypePattern<?> pattern = createTypePattern((CtLocalVariable<?>) child);
```

One could argue that the average Java dev won't know about wildcard capture, and that may be true. But if the type parameter is not meaningfully used, it's simply a hindrance and may as well be removed. **Also, we can help developers along by placing a helper method like this directly in the factory (which is in fact commonly done for other factory methods).**

> ```java
> CtTypePattern<Object> pattern = launcher.getFactory().Core().createTypePattern();
> pattern.setVariable(launcher.getFactory().createLocalVariable()); // compiles due to the fact that the type is inferred
> ```

Sure, that also works I suppose :)

> I can change it back using `T` if you want.

I think it's either that, or simply remove the type parameter. Again, we could add the method I demonstrated above directly into the factory.

> > The return type of `setVariable` is problematic
> 
> Yes, I only did that for consistency reasons.

Yeah, I figured as much.

> > The _correct_ way to do this is to have each extending interface override the method and explicitly specify itself in the return type
> 
> Another approach would be to use recursive generics, but using that properly would be fully incompatible with existing code and does not solve the problem in every case.

Out of curiosity, could you give an example? I don't see the solution. We don't want to break the world obviously, but I'm just curious.

> I'll do it whatever way you decide on, as I don't really have a strong opinion on consistency vs type safety here.

Let's wait for @monperrus to weigh in. But I don't think we should let past mistakes shape the future of the library. Every so often, questions crop up about what the type parameters actually mean in the different types, and it's super awkward to answer "well in this particular instance, nothing tangible in practice" :)

I-Al-Istannen:
@slarse Probably this:
```java
class Scratch {

    new ConcreteBuilder()
        .withBar("Hello")
        .withInteger(20) // we can mix and match the abstraction level here
        .withFoo("hey");
  }
}

abstract class ParentBuilder<SELF> {

  private String foo;

  abstract SELF getSelf();

  public SELF withFoo(String foo) {
    this.foo = foo;
    return getSelf();
  }
}

abstract class StepOneBuilder<SELF> extends ParentBuilder<SELF> {

  private String bar;

  public SELF withBar(String bar) {
    this.bar = bar;
    return getSelf();
  }
}

class ConcreteBuilder extends StepOneBuilder<ConcreteBuilder> {

  private int integer;

  public ConcreteBuilder withInteger(int integer) {
    this.integer = integer;
    return this;
  }
  
  @Override
  ConcreteBuilder getSelf() {
    return this;
  }
}
```

The benefit is that you can actually mix-and-match the levels. If you just overwrite the return type, you'd need to overwrite *every single method* of the superclass otherwise you will no longer be able to call methods from the more specialized one:
```java
ctType
  .setXY()        // from CtType, returns CtType, everything is fine
  .setInherited() // inherited from CtElement, returns CtElement
  .setXY()        // no longer possible as the information that it was a "CtType" was lost!
```

### Problems
As far as I know:
You need to terminate the recursive generic parameter somewhere. Once you terminated it you can no longer inherit from it, so you could only have one concrete type in your inheritance chain.

slarse:
@I-Al-Istannen That is super cool, I've never seen that before. The limitation you mention makes sense, but that would also make it entirely incompatible with Spoon, as other libraries are dependent on subtyping the existing AST types.

> If you just overwrite the return type, you'd need to overwrite every single method of the superclass otherwise you will no longer be able to call methods from the more specialized one:

Unless I'm misunderstanding you, this is not the case, an override is always compatible with the super type (in terms of typing).

<details>
    <summary>This is getting pretty off topic, so putting this in a spoiler :)</summary>
In most programming languages, functions/methods are covariant in the return type and contravariant in the parameter type. However, as Java also has overloading, methods are invariant in the parameter type.

There's nothing stopping a subtype that overrides a return type from calling other methods from the supertype, here's an example where the subtype `Derived` overrides `addTwo`, and defines `addFour` which in turn calls `addThree` from the supertype.

```java
class Base {
    public Number addTwo(Number lhs, Number rhs) {
        return lhs.intValue() + rhs.intValue();
    }

    public Number addThree(Number lhs, Number mid, Number rhs) {
        return addTwo(addTwo(lhs, mid), rhs);
    }
}

class Derived extends Base {
    @Override
    public Integer addTwo(Number lhs, Number rhs) {
        return lhs.intValue() + rhs.intValue();
    }

    public Integer addFour(Number first, Number second, Number third, Number fourth) {
        return addTwo(addThree(first, second, third), fourth);
    }
}

class Main {
        Base base = new Base();
        System.out.println(base.addThree(1, 2, 3));

        Derived derived = new Derived();
        System.out.println(derived.addFour(1, 2, 3, 4));
    }
}
```
</details>

I-Al-Istannen:
> Unless I'm misunderstanding you, this is not the case, an override is always compatible with the super type (in terms of typing, behavior may of course be incompatible).

That comment was for 
```java
ctType
  .setXY()        // from CtType, returns CtType, everything is fine
  .setInherited() // inherited from CtElement, returns CtElement
  .setXY()        // no longer possible as the information that it was a "CtType" was lost!
```
This will not compile as `setInherited` is inherited and not overwritten and its declaration returns a `CtElement` - so you lose the information, that `ctType` was a `CtType` and therefore you can no longer call methods declared in `CtType` on it (`setXY` here).

You can fix this by overwriting `setInherited` in `CtType` to return a `CtType` (and just delegate to super) - but then you would need to re-declare *every single method*. With the SELF recursive generic you get those things for free, as java can resolve the correct return type for you automatically.

> This is getting pretty off topic, so putting this in a spoiler :)

Yea, sorry for that ^^

slarse:
> That comment was for
> [...]

Ah, of course it was. Apparently I sometimes forget that sentences within a paragraph usually are related :)

I agree with the rest if what you said. If we could start all over again, it would be a contender!

monperrus:
regarding type safety, I'm in favor of dropping the type parameter. Time has shown that this design, while elegant on paper, is not useful, and even detrimental for clients in practice.

monperrus:
Regarding setVariable, `CtTypePattern setVariable(CtLocalVariable<?> variable);` is simpler. @slarse (and @leventov before) made a very clear case that the legacy solution with a return-only type parameter is bad, so let's not do it for new API methods.

SirYwell:
> Regarding setVariable, `CtTypePattern setVariable(CtLocalVariable<?> variable);` is simpler.

Should the `T` type on the class even kept then? It's currently only used with the clone method as @slarse already pointed out. (or were you referring to *that* type parameter in your previous comment?)

slarse:
> > Regarding setVariable, `CtTypePattern setVariable(CtLocalVariable<?> variable);` is simpler.
> 
> Should the `T` type on the class even kept then? It's currently only used with the clone method as @slarse already pointed out. (or were you referring to _that_ type parameter in your previous comment?)

We agreed to remove `T` entirely. The parameterization of classes has mostly led to confusion, and very little gain (it's pretty rare _not_ to use a wildcard ...), so moving forward we're going to steer away from it. In most use cases, the concrete type simply can't be known.

We've got some pretty grand ideas for the latter in regards to the entirety of the current API, I'll open a separate issue about that later :)

slarse:
@SirYwell Looks like you might have forgotten to remove the type argument in the subprojects? See the [extra checks](https://github.com/INRIA/spoon/pull/4020/checks?check_run_id=3376237409) workflow run.

slarse (approved):
This LGTM to me now, I have no further comments. Thanks for the excellent work and effort, @SirYwell!

I will leave final approval and merge to @monperrus.

monperrus:
Kudos @SirYwell for the heavy-duty work. Big thanks @slarse for mentoring \& reviewing.

SirYwell:
Thanks. That was very insightful in many ways for me.

PR #4023, opened 2021-07-01
refactor: simplify implementation of `getCaseExpression` and `setCaseExpression`

### LINKED ISSUE #4019: `this.caseExpression` attribute in `CtCase` is no longer needed
Reference: https://github.com/INRIA/spoon/pull/4015#issuecomment-869972744

Before Java 12, only one case expression was allowed per case, so spoon maintained an attribute [`this.caseExpression`](https://github.com/INRIA/spoon/blob/3e10fd5ac22b5e7e75773bd9323870f47372f3d2/src/main/java/spoon/support/reflect/code/CtCaseImpl.java#L31) which would store the expression corresponding to the case. Java 12 introduced a feature to add multiple expressions to a single case, so [`this.caseExpressions`](https://github.com/INRIA/spoon/blob/3e10fd5ac22b5e7e75773bd9323870f47372f3d2/src/main/java/spoon/support/reflect/code/CtCaseImpl.java#L34) was introduced, which would contain a list of `CtExpression`s. The second attribute, `this.caseExpressions`, made `this.caseExpression` obsolete.

I propose that we remove the attribute and modify its getters and setters to, instead, return the list's first element and clear the list and insert a single element, respectively. This would reduce redundancy. Another benefit is that some visitors like [`CtScanner`](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/reflect/visitor/CtScanner.java#L330-L337) do not even access `this.caseExpression`. So if someone would have set the case expression using [`this.setCaseExpression`](https://github.com/INRIA/spoon/blob/3e10fd5ac22b5e7e75773bd9323870f47372f3d2/src/main/java/spoon/support/reflect/code/CtCaseImpl.java#L58), they won't even see it in their pretty-printed string.

algomaster99:
Fixes #4019 

These changes removed the `this.caseExpression` attribute as it is obsolete since the arrival of `this.caseExpressions`. The latter is enough to handle one or more case expressions for a single case.

Before merging, the following tasks need to be done.
- [x] Fix the failing test case - `APITest.testSetterInNodes`.
- [x] Write tests for the new implementation.


algomaster99:
The test fails [here in the CI](https://github.com/INRIA/spoon/pull/4023/checks?check_run_id=2965457943#step:10:468). It expects one `if` condition in `setCaseExpression` but it gets none as asserted [here](https://github.com/INRIA/spoon/blob/master/src/test/java/spoon/test/api/APITest.java#L436). I think the modified implementation is unable to match the [template](https://github.com/INRIA/spoon/blob/3e10fd5ac22b5e7e75773bd9323870f47372f3d2/src/test/java/spoon/test/api/APITest.java#L395-L408) and that's why no `if` condition is found in `setCaseExpression` even though [there is one](https://github.com/INRIA/spoon/pull/4023/files#diff-b0869fb13450494876525960e4e513f88a64f1e565ddc20e9acc047f82805f09R59-R61). I need some help fixing this because the other setters in `CtCaseImpl` are written in a similar format but they pass the test. I am not sure what is going wrong for this single setter.

src/main/java/spoon/support/reflect/code/CtCaseImpl.java
```diff
@@ -56,11 +56,13 @@ public void accept(CtVisitor visitor) {
 
 	@Override
 	public <T extends CtCase<E>> T setCaseExpression(CtExpression<E> caseExpression) {
-		if (caseExpression != null) {
-			caseExpression.setParent(this);
+		if (caseExpression == null) {
+			this.caseExpressions = emptyList();
+			return (T) this;
```
slarse, line 61: The architecture test that fails is that each setter should ensure that the set child has the correct parent. It's not sophisticated enough to figure out that the parent is set later. Something like this should do as a workaround:

```suggestion
		if (caseExpression != null) {
			// setting the parent here is redundant as it's later set in addCaseExpression, but we need to set it
                        // here to appease the architecture tests that check for parent consistency
			caseExpression.setParent(this);

Comment (slarse) [reply src/main/java/spoon/support/reflect/code/CtCaseImpl.java:61]: Perhaps @monperrus has a less dirty workaround?

Comment (algomaster99) [reply src/main/java/spoon/support/reflect/code/CtCaseImpl.java:61]: > setting the parent here is redundant as it's later set in addCaseExpression,

That's exactly why I didn't write that statement here.

> but we need to set it here to appease the architecture tests that check for

Why doesn't [this setter](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/reflect/code/CtCaseImpl.java#L73) fail in that case? This also sets the parent later in `addExpression`.

Comment (slarse) [reply src/main/java/spoon/support/reflect/code/CtCaseImpl.java:61]: The test only looks for setters where the argument is a subtype of `CtElement`. The one you linked takes a list.

File: src/test/java/spoon/test/model/SwitchCaseTest.java [hunk 2]
```diff
@@ -365,4 +367,52 @@ public void test_addCaseAt_throwsIndexOutOfBoundsException_whenPositionIsOutOfBo
...
+		Factory factory = new Launcher().getFactory();
+
+		CtCase<Integer> ctCase = factory.createCase();
+		CtExpression<Integer> first = factory.createCodeSnippetExpression("1");
+		CtExpression<Integer> second = factory.createCodeSnippetExpression("2");
+		CtExpression<Integer> third = factory.createCodeSnippetExpression("3");
+
+		ctCase.setCaseExpressions(Arrays.asList(first, second, third));
+
+		assertEquals(first, ctCase.getCaseExpression());
```
algomaster99, line 383: This statement would be more coherent once https://github.com/INRIA/spoon/pull/4024 is merged. As of now, substituting `second` or `third` instead of `first` would still pass the test.

algomaster99:
@monperrus @slarse I was also wondering if writing `contract` as a comment is necessary here as I have already written it using the `DisplayName` annotation. I could write it for the sake of consistency so let me know.

algomaster99:
Ping @monperrus @slarse for reviews.

src/test/java/spoon/test/model/SwitchCaseTest.java
```diff
@@ -365,4 +367,52 @@ public void test_addCaseAt_throwsIndexOutOfBoundsException_whenPositionIsOutOfBo
...
+	
+	@DisplayName("should return the first case expression")
+	@Test
+	public void test_getCaseExpression_shouldReturnTheFirstCaseExpression() {
+		Factory factory = new Launcher().getFactory();
+
+		CtCase<Integer> ctCase = factory.createCase();
+		CtExpression<Integer> first = factory.createCodeSnippetExpression("1");
+		CtExpression<Integer> second = factory.createCodeSnippetExpression("2");
+		CtExpression<Integer> third = factory.createCodeSnippetExpression("3");
```
slarse, line 379: Either compile the expressions or create literals. I'd do the latter.

```suggestion
		CtExpression<Integer> first = factory.createLiteral(1);
		CtExpression<Integer> second = factory.createLiteral(2);
		CtExpression<Integer> third = factory.createLiteral(3);
```

In general you should avoid having raw code snippet expressions in a tested AST as they aren't type checked.

src/test/java/spoon/test/model/SwitchCaseTest.java
```diff
@@ -365,4 +367,52 @@ public void test_addCaseAt_throwsIndexOutOfBoundsException_whenPositionIsOutOfBo
...
+		@DisplayName("should remove all existing case expressions and insert the specified case expression")
+		@Test
+		public void test_setCaseExpression_removeExistingCaseExpressionsAndInsertTheSpecified() {
+			Factory factory = new Launcher().getFactory();
+
+			CtCase<Integer> ctCase = factory.createCase();
+			CtExpression<Integer> first = factory.createCodeSnippetExpression("1");
+			CtExpression<Integer> second = factory.createCodeSnippetExpression("2");
+			CtExpression<Integer> third = factory.createCodeSnippetExpression("3");
+			CtExpression<Integer> fourth = factory.createCodeSnippetExpression("4");
```
slarse, line 397: Same as above, use `createLiteral`.

src/test/java/spoon/test/model/SwitchCaseTest.java
```diff
@@ -365,4 +367,52 @@ public void test_addCaseAt_throwsIndexOutOfBoundsException_whenPositionIsOutOfBo
 			assertThrows(IndexOutOfBoundsException.class, () -> switchExpression.addCaseAt(3, onlyCase));
 		}
 	}
+	
+	@DisplayName("should return the first case expression")
+	@Test
+	public void test_getCaseExpression_shouldReturnTheFirstCaseExpression() {
```
slarse, line 373: Since you put the other tests in a nested type, I think this one should also be in a nested type, for consistency within this test class. It looks like all tests are categorized in nested types.

As you are testing different aspects of the same concept, I'd say put it into the new nested type you created. The other nested types don't categorize by method, but by concept, so that's fine in terms of consistency.

src/test/java/spoon/test/model/SwitchCaseTest.java
```diff
@@ -365,4 +367,52 @@ public void test_addCaseAt_throwsIndexOutOfBoundsException_whenPositionIsOutOfBo
...
+		@DisplayName("should clear the list of expressions when null is passed")
+		@Test
+		public void test_setCaseExpression_removeAllCaseExpressions() {
+			String code = "class A { public void f(int i) { switch(i) { case 1,2,3: break; } } }";
+			CtModel model = createModelFromString(code);
+			CtCase<?> ctCase = model.getElements(new TypeFilter<>(CtCase.class)).get(0);
+
+			ctCase.setCaseExpression(null);
+
+			assertTrue(ctCase.getCaseExpressions() instanceof EmptyClearableList);
```
slarse, line 415: Like we've discussed a few times before, I would if at all possible avoid using `assertTrue` for anything but asserting something that is inherently a boolean. What you want to assert here isn't a boolean itself, but rather you want to check that the expression list is empty. I would also argue that asserting specifically against `EmptyClearableList` is overspecifying the test: that's an implementation detail that's not visible through the API.

I would change to this:

```suggestion
			assertThat(ctCase.getCaseExpressions().size(), equalTo(0));
```

As we just have the core Hamcrest matchers. But if we add all Hamcrest matchers as of version 2.2 (I'll open an issue about that), this assertion could be even more clear.

```suggestion
			assertThat(ctCase.getCaseExpressions(), is(empty()));
```

slarse (changes requested):
Implementation LGTM. Some minor comments on the tests.

Also, about the display names, I would put the method under test into them as the first thing written, because that's the name seen when running the tests. And so it's nice to know which method is under test just by looking at the name. Since the other nested types in the class are not categorized by method, it's not sufficient to rely on the name of the nested type for this one case.

slarse:
> @monperrus @slarse I was also wondering if writing `contract` as a comment is necessary here as I have already written it using the `DisplayName` annotation. I could write it for the sake of consistency so let me know.

If the contract is adequately expressed in the display name, then no that's not necessary. But it should be evident what part/piece of funcionality is being tested, and under what circumstances, and what the expected result is.

src/test/java/spoon/test/model/SwitchCaseTest.java
algomaster99: > The other nested types don't categorize by method, but by concept, so that's fine in terms of consistency.

Right. Makes more sense.

algomaster99:
> I would put the method under test into them as the first thing written,

How about I omit the `DisplayName` annotation so that the method name only displays on the side panel? My reason for doing so is that prepending the method name would make the display name quite long and one would have to expand the IntelliJ side panel by a lot to read its purpose. However, it may be more readable when one runs the test using the terminal.

I propose that we omit the `DisplayName` annotation and write it as contract using inline comments.

slarse (approved):
This looks good to me now, but I'd like @monperrus to have a look as well since this is (technically speaking) a metamodel change. Although internal.

@monperrus Please see what you think of this, it looks like a good patch to me.

monperrus:
I like this simplification (simpler is better: this is purely internal, the API remains the same, and all tests are green.

:+1: 

Thanks a lot @algomaster99

slarse:
Thansk @algomaster99, nicely done!

PR #4024, opened 2021-07-02
fix: Fix equality for CtCodeSnippet

### LINKED ISSUE #4022: bug: Question about equality of `CtCodeSnippetExpression`
I created two expressions as shown below.

```java
CtExpression<Integer> firstExpression = factory.createCodeSnippetExpression("1");
CtExpression<Integer> secondExpression = factory.createCodeSnippetExpression("2");
```
When I execute `firstExpression.equals(secondExpression);`, I get `true` in return. Clearly, the value of these expressions is different so I cannot understand why they are passing the equality test.

slarse:
Fix #4022

This PR fixes equality between `CtCodeSnippet` instances. Previously, two code snippets would be considered equal regardless of their contained snippet (i.e. the raw string). With this PR, the raw snippet is now also considered part of the equality of code snippets (both statements and expressions).

### DIFF
--- a/src/main/java/spoon/support/visitor/equals/EqualsChecker.java
+++ b/src/main/java/spoon/support/visitor/equals/EqualsChecker.java
@@ -16,6 +16,7 @@ import spoon.reflect.code.CtOperatorAssignment;
 import spoon.reflect.code.CtStatement;
 import spoon.reflect.code.CtTextBlock;
 import spoon.reflect.code.CtUnaryOperator;
+import spoon.reflect.declaration.CtCodeSnippet;
 import spoon.reflect.declaration.CtElement;
 import spoon.reflect.declaration.CtMethod;
 import spoon.reflect.declaration.CtModifiable;
@@ -107,6 +108,15 @@ public class EqualsChecker extends CtInheritanceScanner {
 			setNotEqual(CtRole.MODIFIER);
 		}
 		super.scanCtModifiable(m);
+	}
+
+	@Override
+	public void scanCtCodeSnippet(CtCodeSnippet snippet) {
+		final CtCodeSnippet peek = (CtCodeSnippet) this.other;
+		if (!snippet.getValue().equals(peek.getValue())) {
+			setNotEqual(CtRole.SNIPPET);
+		}
+		super.scanCtCodeSnippet(snippet);
 	}
 
 	@Override
--- a/src/test/java/spoon/test/snippets/SnippetTest.java
+++ b/src/test/java/spoon/test/snippets/SnippetTest.java
@@ -30,7 +30,6 @@ import spoon.reflect.code.CtStatement;
 import spoon.reflect.code.CtBlock;
 import spoon.reflect.code.CtComment;
 import spoon.reflect.declaration.CtClass;
-import spoon.reflect.declaration.CtCodeSnippet;
 import spoon.reflect.declaration.CtElement;
 import spoon.reflect.declaration.CtMethod;
 import spoon.reflect.factory.Factory;
--- a/src/test/java/spoon/test/snippets/SnippetTest.java
+++ b/src/test/java/spoon/test/snippets/SnippetTest.java
@@ -243,13 +243,24 @@ public class SnippetTest {
 	}
 
 	@Test
-	public void testSnippetsWithDifferingValuesAreNotEqual() {
-		// contract: Two snippets with different values should be considered not equal
+	public void testCodeSnippetExpressionsWithNonEqualValuesAreNotEqual() {
+		// contract: Two code snippet expressions with non-equal values are not equal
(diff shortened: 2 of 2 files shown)


src/test/java/spoon/test/snippets/SnippetTest.java
```diff
@@ -239,4 +242,15 @@ public void testSnippetCompilationInUnnamedPackage() {
...
+
+	@Test
+	public void testSnippetsWithDifferingValuesAreNotEqual() {
+		// contract: Two snippets with different values should be considered not equal
+		Factory factory = new Launcher().getFactory();
+
+		CtExpression<Integer> one = factory.createCodeSnippetExpression("1");
+		CtExpression<Integer> two = factory.createCodeSnippetExpression("2");
+
+		assertThat(one, not(equalTo(two)));
```
algomaster99, line 254: Any reason why you have preferred `asserThat` over `assertEquals`?

Rohitesh-Kumar-Jain: > Any reason why you have preferred asserThat over assertEquals?

what about `assertNotEquals(one, two);` 😅

slarse: > Any reason why you have preferred asserThat over assertEquals?

Many reasons. First, I prefer the way you write the Hamcrest matchers, i.e. `assertThat(one, not(equalTo(expected)))`. It's clear which is the actual value, and which is the expected one. For `assertNotEquals(one, two)`, not so much (these ar actually reversed, because it's `assertNotEquals(expected, actual), but `assertThat(actual, someMatcher(expected))`).

Hamcerst matchers also almost always give more informative failure information than JUnit's built-in matchers when an assertion fails.

There's however a pretty big problem with this test that is unrelated to the assertion being used. Why is `one` the actual value, and `two` the expected value? Which method is actually under test here? Of course, neither of `one` or `two` is actual or expected, they are just two values. The method under test is `equals`, but that's hidden by the assertion.

The test should probably look like this instead:

```suggestion
		boolean areEqual = one.equals(two);

		assertFalse(areEqual);
```

slarse: Or more concisely, this would also be fine
```suggestion
		assertFalse(one.equals(two));
```

Asserting true and false is probably the one place where I prefer the built-in assertions to Hamcrest matchers, because the Hamcrest alternative is a bit more verbose than necessary:

```suggestion
		assertThat(one.equals(two), is(false));
```

However, compare the Hamcrest failure message:

```
java.lang.AssertionError: 
Expected: is <false>
     but: was <true>
[... stacktrace ...]
```

Is much more readable than the JUnit failure message:

```
java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:87)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at org.junit.Assert.assertFalse(Assert.java:65)
	at org.junit.Assert.assertFalse(Assert.java:75)
	at spoon.test.snippets.SnippetTest.testSnippetsWithDifferingValuesAreNotEqual(SnippetTest.java:255)
[... more stacktrace ...]
```

So actually I'll probably go with the Hamcrest one after all :)

algomaster99: > java.lang.AssertionError: 
Expected: is <false>
     but: was <true>
[... stacktrace ...]

Agree, this is clearly better.

src/main/java/spoon/support/visitor/equals/EqualsChecker.java
```diff
@@ -109,6 +110,15 @@ public void scanCtModifiable(CtModifiable m) {
 		super.scanCtModifiable(m);
 	}
 
+	@Override
+	public void scanCtCodeSnippet(CtCodeSnippet snippet) {
+		final CtCodeSnippet peek = (CtCodeSnippet) this.other;
+		if (!snippet.getValue().equals(peek.getValue())) {
+			setNotEqual(CtRole.SNIPPET);
```
algomaster99, line 117: @slarse This method throws an [exception](https://github.com/INRIA/spoon/pull/4024/files#diff-d991c6c34e1ca95a4c24d3ddec98465f7517bc76a50cbcf04a0f0c866fffdf6eR56) eventually so I am confused why it doesn't exit the JVM.

slarse: `EqualsChecker.scan()` is called within a try/catch: https://github.com/INRIA/spoon/blob/66db36cc8fb4e872597c51699e8aab1227a2542a/src/main/java/spoon/support/visitor/equals/EqualsVisitor.java#L38-L48

slarse:
@monperrus @nharrand ping for merge :)

monperrus:
Thanks a lot @slarse

PR #4025, opened 2021-07-02
review: chore: Try to list new errors in javadoc regression check

I-Al-Istannen:
## Motivation
Before this patch the CI Javadoc regression check would only look at the amount of violations and fail the build if they increased:

```
JAVADOC QUALITY SCORE (lower is better)
    Compare: 2584
    Current: 2592

Javadoc quality has deteriorated!
Run the chore/ci-checkstyle-javadoc.sh script locally to find errors
See https://github.com/inria/spoon/issues/3923 for details
```

This wasn't really helpful when actually finding the errors, as you would need to run the checkstyle command locally on both branches and diff the output yourself. 

## Changes
This patch adds a small python script that tries to (*somewhat* intelligently) perform this diff in the CI and print a more useful failure message.

The line numbers in the output will not be correct if the exact same violation message appears multiple times in a single file, but it is better than having no feedback.

## Help wanted
How do you test changes that affect your CI process? I tested them locally by:
1. Checking out this PR's branch
2. Randomly deleting some helpful javadoc
3. Committing the changes
4. Running `./chore/ci-checkstyle-javadoc.sh COMPARE_WITH_MASTER`

This, however, doesn't quite help me verify it doesn't crash in your CI environment.

## Example output
```
JAVADOC QUALITY SCORE (lower is better)
    Compare: 2584
    Current: 2592
    
Javadoc quality has deteriorated!
Run the chore/ci-checkstyle-javadoc.sh script locally to find errors
See https://github.com/inria/spoon/issues/3923 for details


You likely added the following new errors. Line numbers might be incorrect if there are multiple violations of the same type in the file.
  [ERROR] /tmp/spoon/src/main/java/spoon/reflect/visitor/CtVisitor.java:116:10: Erwartetes Tag @param für '<T>'. [JavadocMethod]
  [ERROR] /tmp/spoon/src/main/java/spoon/reflect/visitor/CtVisitor.java:543:54: Erwartetes Tag @param für 'pattern'. [JavadocMethod]
  [ERROR] /tmp/spoon/src/main/java/spoon/reflect/factory/CoreFactory.java:131: Fehlendes @return-Tag. [JavadocMethod]
  [ERROR] /tmp/spoon/src/main/java/spoon/reflect/factory/CoreFactory.java:136:10: Erwartetes Tag @param für '<T>'. [JavadocMethod]
  [ERROR] /tmp/spoon/src/main/java/spoon/reflect/code/CtTypePattern.java:40: Fehlendes @return-Tag. [JavadocMethod]
  [ERROR] /tmp/spoon/src/main/java/spoon/reflect/code/CtTypePattern.java:46: Fehlendes @return-Tag. [JavadocMethod]
  [ERROR] /tmp/spoon/src/main/java/spoon/reflect/code/CtTypePattern.java:47:10: Erwartetes Tag @param für '<C>'. [JavadocMethod]
  [ERROR] /tmp/spoon/src/main/java/spoon/reflect/code/CtTypePattern.java:47:71: Erwartetes Tag @param für 'variable'. [JavadocMethod]
```

### DIFF
--- a/chore/ci-checkstyle-javadoc.sh
+++ b/chore/ci-checkstyle-javadoc.sh
@@ -13,6 +13,8 @@ set -o pipefail
 
 COMPARE_BRANCH="master"
 JAVADOC_CHECKSTYLE_CONFIG="__SPOON_CI_checkstyle-javadoc.xml"
+OUTPUT_FILE_COMPARE="__SPOON_CI_checkstyle-javadoc-compare"
+OUTPUT_FILE_OWN="__SPOON_CI_checkstyle-javadoc-own"
 
 COMPARE_WITH_MASTER_ARG="COMPARE_WITH_MASTER"
 RELATED_ISSUE_URL="https://github.com/inria/spoon/issues/3923"
@@ -20,6 +22,14 @@ RELATED_ISSUE_URL="https://github.com/inria/spoon/issues/3923"
 function cleanup() {
     if [ -f "$JAVADOC_CHECKSTYLE_CONFIG" ]; then
         rm "$JAVADOC_CHECKSTYLE_CONFIG"
+    fi
+
+    if [ -f "$OUTPUT_FILE_COMPARE" ]; then
+        rm "$OUTPUT_FILE_COMPARE"
+    fi
+
+    if [ -f "$OUTPUT_FILE_OWN" ]; then
+        rm "$OUTPUT_FILE_OWN"
     fi
 }
 
@@ -39,11 +49,12 @@ function create_checkstyle_config() {
 
 function run_checkstyle() {
     create_checkstyle_config
-    mvn -B checkstyle:check --fail-never -Dcheckstyle.config.location="$JAVADOC_CHECKSTYLE_CONFIG"
+    mvn -B checkstyle:check --fail-never -Dcheckstyle.config.location="$JAVADOC_CHECKSTYLE_CONFIG" > "$1"
+    cat "$1"
 }
 
 function compute_num_errors() {
-     grep -Po '(?<=There are )\d+(?= errors reported by Checkstyle)' <<< `run_checkstyle`
+     grep -Po '(?<=There are )\d+(?= errors reported by Checkstyle)' <<< `run_checkstyle "$1"`
 }
 
 function main() {
@@ -51,11 +62,11 @@ function main() {
 
     # compute compare score
     git checkout --force "$COMPARE_BRANCH" &> /dev/null
-    compare_num_errors=`compute_num_errors`
+    compare_num_errors=`compute_num_errors "$OUTPUT_FILE_COMPARE"`
 
     # compute current score
     git checkout --force - &> /dev/null
-    current_num_errors=`compute_num_errors`
+    current_num_errors=`compute_num_errors "$OUTPUT_FILE_OWN"`
 
     echo "JAVADOC QUALITY SCORE (lower is better)
     Compare: $compare_num_errors
@@ -72,6 +83,8 @@ function main() {
         echo "Javadoc quality has deteriorated!"
         echo "Run the chore/ci-checkstyle-javadoc.sh script locally to find errors"
         echo "See $RELATED_ISSUE_URL for details"
+        echo -e "\n"
+        python3 chore/find_javadoc_regression.py "$OUTPUT_FILE_COMPARE" "$OUTPUT_FILE_OWN"
         exit 1
     else
         echo "Javadoc quality has not deteriorated"
@@ -108,5 +121,5 @@ if [ "$1" == "$COMPARE_WITH_MASTER_ARG" ]; then
 
     main
 else
-    grep "$1" <<< `run_checkstyle`
+    grep "$1" <<< `run_checkstyle "$OUTPUT_FILE_OWN"`
 fi
--- a/chore/find_javadoc_regression.py
+++ b/chore/find_javadoc_regression.py
@@ -1,0 +1,105 @@
+#!/usr/bin/env python3
+
+import re
+import sys
+from pathlib import Path
+from typing import List, Dict
+
+
+def filter_relevant_lines(lines: List[str]) -> List[str]:
+    """Tries to filter the checkstyle output to only include lines with violations."""
+    new_lines = []
+
+    for line in lines:
+        if re.search(r":\d+:", line) and "ERROR" in line:
+            new_lines.append(line)
+
+    return new_lines
+
+
+def strip_line_number(line: str) -> str:
+    return re.sub(r".java:\d+(:\d+)?:", ".java", line)
+
+
+def strip_line_numbers(lines: List[str]) -> Dict[str, int]:
+    """
+    Strips the violation line numbers to account for shifts.
+    This method returns a dictionary of "stripped line -> Count", as the same
+    error might appear multiple times.
+    """
+    output_dict: Dict[str, int] = dict()
+
+    for line in map(strip_line_number, lines):
+        if line in output_dict:
+            output_dict[line] += 1
+        else:
+            output_dict[line] = 1
+
+    return output_dict
+
+
+def try_match_lines(reference: Dict[str, int], other: Dict[str, int]) -> Dict[str, int]:
+    """Tries to find out which lines are *new* in other and which were already present in the reference."""
+    other = other.copy()
+    for line, amount in reference.items():
+        for _ in range(0, amount):
+            if line in other:
+                if other[line] > 1:
+                    other[line] -= 1
+                else:
+                    other.pop(line)
+
+    # Lines that are only in the reference aren't critical, as they were
+    # apparently fixed in "other"
+    return other
+
+
+def try_readd_line_numbers(without_numbers: Dict[str, int], with_numbers: List[str]) -> List[str]:
+    """
+    Tries to re-add the line numbers to a dict of strings.
+    This is not necessarily accurate, as it doesn't know which occurrence
+    in "with_numbers" is the real source. If the error message is unique within
+    a file it will find the exact match though.
+    """
+    output: List[str] = []
+
+    for line in with_numbers:
+        stripped = strip_line_number(line)
+        if stripped in without_numbers:
+            if without_numbers[stripped] == 1:
+                without_numbers.pop(stripped)
+            else:
+                without_numbers[stripped] -= 1
+            output.append(line)
+
+    for line in without_numbers:
+        output.append(line)
+
+    return output
+
+
+def main(file_reference: Path, file_other: Path):
+    with open(file_reference, "r") as file:
+        lines_reference = filter_relevant_lines(file.readlines())
+    with open(file_other, "r") as file:
+        lines_other = filter_relevant_lines(file.readlines())
+
+    new_errors = try_match_lines(
+        strip_line_numbers(lines_reference),
+        strip_line_numbers(lines_other)
+    )
+    with_line_numbers = try_readd_line_numbers(new_errors, lines_other)
+
+    print("You likely added the following new errors. "
+          "Line numbers might be incorrect if there are "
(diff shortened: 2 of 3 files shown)


slarse:
Hi @I-Al-Istannen,

This is definitely an improvement, but looking at your Python script, I think that going "all the way" isn't that much harder. By all the way, I mean the following redesign of the whole thing:

1. Run Javadoc checkstyle on the current branch
2. Run `git diff` with the master branch
3. Find all checkstyle violations that appear on _added_ lines present in the diff
    - If there are any violations, print violations and fail build
    - Otherwise, all good!

This isn't really _that_ hard to implement. It's relatively easy to parse the changed lines from a Git diff (see e.g. [this parse function](https://github.com/SpoonLabs/sorald-buildbreaker/blob/958ffc6b9d9bad0f091b3add13c73d6c8afa5f61/src/git.ts#L97-L127)), and from there all that remains is to filter by file and line range (see e.g. [this filtering function](https://github.com/SpoonLabs/sorald-buildbreaker/blob/958ffc6b9d9bad0f091b3add13c73d6c8afa5f61/src/action.ts#L106-L128)).

WDYT?

> How do you test changes that affect your CI process?

Activate GitHub Actions in your fork, and run it there. You can also change which branch you compare with if you don't want to mess up the master branch.

slarse:
Of course, if you feel like that's too much work, any improvement is of course welcome.

I'm also thinking it would be beneficial to convert the whole script into Python, I can't quite recall why I wrote it in bash :)

I-Al-Istannen:
That's an interesting approach, I'll try it out. I'll have to see where exactly checkstyle reports its errors, I image it might not catch the following. We'll see.
```diff
 /*
- * @param foobar a param
  *
  */
 public void foo(String foobar);
```

----

I wanted to avoid tinkering with your existing scripts too much, that's why it takes the checkstyle output files and keeps the bash script. I wasn't sure whether any other part of your project depends on them, sometimes bash scripts are used in interesting ways from different places :)

As you seem to be fine with a change there, I could try to remove the middleman.

> I'm also thinking it would be beneficial to convert the whole script into Python, I can't quite recall why I wrote it in bash :)

That's what I always end up with when hacking together CI pipelines without worrying too much about it, so maybe that was a simiiar motivation ;) Bash's quite convenient if you just shell out to other projects and combine some tools.

I will give diff parsing a go. Thankfully it isn't quite as high stakes as [this one](https://blog.ryotak.me/post/homebrew-security-incident-en/)...

slarse:
> I image it might not catch the following. We'll see.

A, right, that will of course not be caught! So, _added_ violations are easy, but actual regressions are harder. Need to think a bit about that.

I-Al-Istannen:
So far I am not convinced you can accurately attribute the regressions to their line numbers without investing quite a bit of effort.

----

* We can ignore deletions that happen inside a function, as they won't ever impact the javadoc coverage.
* We can ignore deletions that happen in the *declaration* of a function, as they will count as an "added" line *somewhere* and checkstyle should report them.
* We can *not* ignore deletions that happen anywhere in the javadoc, as they might cause a regression.

If we notice that a deleted line touches any javadoc, the diff might not have the corresponding function/class declaration checkstyle reports errors for (the context of a diff is limited to X lines). So we'd need to look at the original file on disk and find the corresponding function/class declaration. That is quite doable.
Using this we can assume that we know the exact line (and therefore all violations) checkstyle reported for any (relevant) touched function/class. If the whole element was deleted, checkstyle will not report anything and we can drop the match.

Now that we know the errors checkstyle reports for an element we know was touched in the diff, we need to figure out if they are *new*. To do this we need to figure out the corresponding element in the comparison branch. As we know this change only impacted Javadoc, the qualified name of the element is the same (iff the containing class didn't change its name). So we could try to find the element with the same name and parameters in the comparison branch and see if the errors existed there.

This would require special handling for types, fields and methods as their qualified name and signature is made up of different elements.

I think we'd end up with a considerably more complex algorithm that can very well contain some bugs. I am not sure that'd be worth it. The CI job should block all regressions and provide *useful starting points* for fixing them, it shouldn't be complex enough to easily contain bugs and doesn't have to *absolutely* perfect at highlighting the location.

----

We could also just assume that nobody in their right mind would introduce a regression in things they didn't actually add in the PR and perform the current dance of comparing absolute numbers, using the logic in this PR to match up error messages but be smarter about added methods, so that their location always corresponds to the actual change the contributor made. I am not sure how much extra work that'd be.

----

It is quite probable that there is a neat algorithm I couldn't think of. Matching elements against each other sounds like it should be a solved problem.

slarse:
Thanks for the analysis, I agree that it sounds like way more effort than it's worth.

And you know what, it's not only deletions that are problematic. Think about this case where we add a new parameter:

```diff
 /**
  * Sum numbers.
  * @param first The first number
  * @param second The second number
  * @return Sum of the provided numbers
  */
 public int sum(
     int first,
-    int second) {
+    int second,
+    int third) {
[...]
```

Even in this simple case of introducing a violation of the javadoc rule, we can't just look at the added lines, we need to match methods to added lines. So clearly, I didn't think this one through well enough, and my initial best-effort of just counting before-and-after is probably the most fool-proof way of an objective measure of "you made it worse".

On a side note, we could probably get to a pretty decent implementation using [gumtree-spoon](https://github.com/spoonlabs/gumtree-spoon-ast-diff) to match elements across revisions, but creating such a utility would be a project in and of itself. A possibly pretty cool one though, I'm not aware of any off-the-shelf tool for Java for tracking changes across revisions like this.

Sorry if I created extra work for you! In any case, let's go with your original idea, let me know when it's ready to review.

I-Al-Istannen:
> Sorry if I created extra work for you!

No harm done, it was an interesting tangent :) Gumtree-spoon sounds like the solution to the solved problem, that could be an interesting project. But I will pass on that for now ;)

----

I think the code works in the CI as well, as [I tested it on my fork](https://github.com/I-Al-Istannen/spoon/runs/2982907605?check_suite_focus=true) now. GH apparently formats the output, but it is still readable IMHO:
grafik


I can also convert the bash script and integrate it in the python script if you'd like. If not, I'd mark the PR as ready for review.

chore/find_javadoc_regression.py
```diff
@@ -0,0 +1,105 @@
...
+        for _ in range(0, amount):
+            if line in other:
+                if other[line] > 1:
+                    other[line] -= 1
+                else:
+                    other.pop(line)
+
+    # Lines that are only in the reference aren't critical, as they were
+    # apparently fixed in "other"
+    return other
```
slarse, line 54: If you use `collections.Counter`, this can be computed as a subtraction

```suggestion
def try_match_lines(reference: collections.Counter[str], other: collections.Counter[str]) -> Dict[str, int]:
    """Tries to find out which lines are *new* in other and which were already present in the reference."""
    return other - reference
```

Again, maybe it doesn't even need to be a function of its own anymore

chore/find_javadoc_regression.py
```diff
@@ -0,0 +1,105 @@
...
+            if without_numbers[stripped] == 1:
+                without_numbers.pop(stripped)
+            else:
+                without_numbers[stripped] -= 1
+            output.append(line)
+
+    for line in without_numbers:
+        output.append(line)
+
+    return output
```
slarse, line 78: This can also be simplified with a counter, something like this should do (I haven't tested this, may need tweaking):

```suggestion
    all_errors_without_line_numbers = collections.Counter(map(strip_line_number, with_numbers))
    errors_without_line_numbers_diff = all_errors_without_line_numbers - collections.Counter(without_numbers)
    return with_numbers + list(without_numbers_diff)
```

slarse (changes requested):
This looks mostly good! I think we can start with just adding this Python script, and then converting the whole thing into Python in a separate PR. This is kind of standalone.

Anyway, the thing I noted in your Python code is that it's mostly about computing amounts of occurrences and differences between those. There's built-in support for that in the stdlib, so I've added some comments on how the code can be simplified (probably, I haven't actually tested any of this :). Have a look and see what you think.

chore/find_javadoc_regression.py
slarse: Oh wait, I see that I misread the indentation there (I thought it was odd that each line with a number was added to the output; it wasn't :D).

chore/find_javadoc_regression.py
I-Al-Istannen: I think I'll leave it as a function because I like the comment trying to explain why we subtract them in that order. It is probably perfectly clear without it, but I am slightly more positive that I will understand what I did in a year with it present.

I-Al-Istannen:
Workflow run for dfcca17 with a failure can be found [here](https://github.com/I-Al-Istannen/spoon/runs/2997999247?check_suite_focus=true) and without a failure [here](https://github.com/I-Al-Istannen/spoon/runs/2998128066?check_suite_focus=true).

chore/find_javadoc_regression.py
```diff
@@ -0,0 +1,78 @@
...
+
+def find_added_violation_lines(reference: 'Counter[str]', other: 'Counter[str]') -> 'Counter[str]':
+    """
+    Tries to find out which violation lines are new in "other" and which
+    were already present in the "reference". These lines are added violations the
+    contributor should fix.
+    """
+    # Lines that are only in the reference aren't critical, as they were
+    # apparently fixed in "other"
+    return other.copy() - reference
```
slarse, line 27: Very minor note here: there's no need to copy `other`, as the subtraction is not in-place (i.e. does not mutate `other`).

```suggestion
    return other - reference
```

chore/ci-checkstyle-javadoc.sh
```diff
@@ -72,6 +83,8 @@ function main() {
         echo "Javadoc quality has deteriorated!"
         echo "Run the chore/ci-checkstyle-javadoc.sh script locally to find errors"
         echo "See $RELATED_ISSUE_URL for details"
+        echo -e "\n"
+        python3 chore/find_javadoc_regression.py "$OUTPUT_FILE_COMPARE" "$OUTPUT_FILE_OWN"
```
slarse, line 87: Here's a good reason why we'd want to compose this whole thing into Python: the name of the Python executable is something of a moving target right now. Ever since Python 2 went belly-up, the conventions have been in flux. Some systems now only have `python`.

We want it to be possible to run this script locally with as little effort as possible.

slarse (approved):
LGTM!

I had one teeny tiny note in the script, but it's not important. You can address it if you'd like, I'll merge tomorrow morning regardless.

If you feel like it, translating the entire script into Python (say Python 3.6 for good compatibility) is appreciated.

chore/find_javadoc_regression.py
I-Al-Istannen: I actually looked that up in the docs because I wasn't sure and didn't want to needlessly copy things aroound. The `subtract` method *does* modify the counter but the binary operator doesn't... It makes sense now that I think about it, I'll adjust it.

Thank you :)

chore/ci-checkstyle-javadoc.sh
I-Al-Istannen: Your `ci-extra` script actually broke in a docker container I tried it in because the ubuntu latest container symlinked `python` to `python2` and your scripts require python 3. That's why I opted for `python3` as the executable name here.

But yea, the whole situation is a bit unfortunate...

slarse: Well, `ci-extra.sh` requiring a little bit of tinkering to get running is fine IMO, as it's not meant to run locally. But as this script is meant to be run locally both to fix regressions and to actually improve the Javadoc quality as part of #3923, we'll want to make it as hassle-free as humanly possible.

Python 3.6+ should arguably be more hassle-free than bash, as the former also works on (native) Windows.

slarse:
Thanks @I-Al-Istannen, very nice work!

Issue #4021, opened 2021-07-01
bug: Combining CtSuperAccess and CtUnaryOperator lead to SpoonException with Sniper


gtoison:
Hello,

I'm trying to analyze and refactor some code containing the following line (note that is has a "super" access and an unary "minus" operator) :

`Date limitTradeDateBeforeIssue = super.computeSettleDate(getIssueDate(), -days);`

When pretty printing the code with Sniper I get an exception which seems to be coming from this specific line of code:

```
Exception in thread "main" spoon.SpoonException: Cannot compare this: [151, 192] with other: ["191", "196"]
	at spoon.support.sniper.internal.ElementSourceFragment.compare(ElementSourceFragment.java:371)
	at spoon.support.sniper.internal.ElementSourceFragment.add(ElementSourceFragment.java:285)
	at spoon.support.sniper.internal.ElementSourceFragment.addChild(ElementSourceFragment.java:317)
	at spoon.support.sniper.internal.ElementSourceFragment.addChild(ElementSourceFragment.java:243)
	at spoon.support.sniper.internal.ElementSourceFragment.access$000(ElementSourceFragment.java:51)
	at spoon.support.sniper.internal.ElementSourceFragment$1.enter(ElementSourceFragment.java:200)
	at spoon.reflect.visitor.CtScanner.visitCtUnaryOperator(CtScanner.java:808)
	at spoon.support.reflect.code.CtUnaryOperatorImpl.accept(CtUnaryOperatorImpl.java:36)
	at spoon.reflect.visitor.EarlyTerminatingScanner.doScan(EarlyTerminatingScanner.java:145)
	at spoon.reflect.visitor.EarlyTerminatingScanner.scan(EarlyTerminatingScanner.java:121)
	at spoon.reflect.visitor.CtScanner.scan(CtScanner.java:170)
	at spoon.reflect.visitor.EarlyTerminatingScanner.scan(EarlyTerminatingScanner.java:106)
	at spoon.reflect.visitor.EarlyTerminatingScanner.scan(EarlyTerminatingScanner.java:83)
	at spoon.reflect.visitor.CtScanner.visitCtInvocation(CtScanner.java:505)
	at spoon.support.reflect.code.CtInvocationImpl.accept(CtInvocationImpl.java:46)
	at spoon.reflect.visitor.EarlyTerminatingScanner.doScan(EarlyTerminatingScanner.java:145)
[...]

PR #4030, opened 2021-07-06
test: Test reproducing issue #4021

gtoison:
A simple unit test reproducing issue #4021, comment out the `Disabled` annotation and the test should fail with a `SpoonException`.

The issue seems to be triggered by a `super` invocation with a unary operator as argument (`-x`)


slarse (approved):
I've verified that this reproduces the correct issue. Thanks very much @gtoison!

PR #4032, opened 2021-07-07
review: chore: Migrate Javadoc regression script to python

I-Al-Istannen:
## Motivation
After #4025 we now have a python script that tries to find the new errors and a bash script that actually runs the checkstyle. This PR will try to consolidate this into a single python script that should also work on Windows.

## Content
This PR will expand the regression-finding Python script to also run the checkstyle and replace the existing bash script.

## Question: Paths
`subprocess` on Windows is apparently pretty broken and won't scan through the full (user + system) `PATH`. This might cause the `mvn` invocation to fail, if the user has installed maven into their user `PATH` (e.g. via scoop). I couldn't find any proper workaround for this, but I could set `shell=True` and pass all commands as simple strings. We control them so we should be able to avoid security vulnerabilities along the way.

## Question: Colors
Most terminal emulators and Github Actions (see [here](https://github.com/I-Al-Istannen/spoon/runs/3007418302?check_suite_focus=true) and [here](https://github.com/I-Al-Istannen/spoon/runs/3007404743?check_suite_focus=true)) do support ANSI escape codes (and we could disable it on windows by default). This allows us to color the output of the script to make it a bit easier to read. This is probably a bit controversial, but would you be opposed to that?

Current draft:
grafik
grafik
grafik

And if the user (likely) screwed up their maven setup:
grafik


slarse:
> Question: Paths [...] subprocess on Windows is apparently pretty broken and won't scan through the full (user + system) PATH

Ugh. Well, that's too much of a bother IMO. Unless someone says anything, let's leave it as explicitly *NIX compatible.

> but I could set shell=True and pass all commands as simple strings

As this runs in CI, I would tend to say no. Maybe it'll be fine but it triggers my spidey sense pretty badly.

> Question: Colors [...]

I don't think it's worth the bother. To my eyes it adds very little here since the output is so simple anyway. It makes the code just a little bit harder to read, and a little bit harder to edit (have to consider what color something should have).

That's just my opinion and I don't feel strongly enough about it to say nay, though. If you feel like the colors add enough value to be worth it, then go ahead.

I-Al-Istannen:
> Ugh. Well, that's too much of a bother IMO. Unless someone says anything, let's leave it as explicitly *NIX compatible.

I adjusted a few things while trying it out and it should work on windows, barring the potential PATH issue. So that might be good enough for now :)

> As this runs in CI, I would tend to say no. Maybe it'll be fine but it triggers my spidey sense pretty badly.

Yea, it will probably be alright but you don't exactly have to add an possible attack vector if you can help it...

> I don't think it's worth the bother. To my eyes it adds very little here since the output is so simple anyway. It makes the code just a little bit harder to read, and a little bit harder to edit (have to consider what color something should have).  
That's just my opinion and I don't feel strongly enough about it to say nay, though. If you feel like the colors add enough value to be worth it, then go ahead.

I probably wrote too many weird websites and programs but I think it's quite nice and helps. I initially added color support because I was having trouble reading the plain white output at a glance. I'll leave it in there and you can always rip it out if you feel it gets a bit cumbersome. I do hope nobody will be unfortunate enough to touch this area in the near future though :P


----
The CI in this PR here runs with my changes so you can find the output [here](https://github.com/INRIA/spoon/pull/4032/checks?check_run_id=3011313496). I probably didn't have to enable them in my fork, as I could have just looked at the output in here.

chore/check-javadoc-regressions.py
```diff
@@ -0,0 +1,272 @@
...
+        print()
+        print("Current status")
+        print(
+            f"  Violations on {hl('comparison')} branch ({target_branch}): "
+            f"{hl(str(reference_violation_count))}")
+        print(
+            f"  Violations on {hl('your')} branch         {' '*len(target_branch)}: "
+            f"{hl(str(other_violation_count))}"
+        )
+        print()
```
slarse, line 203: Why not just print one large string here? I think that'd make it easier to read (in the source code, that is).

```python
        print(f"""
Current status
  Violations on {hl('comparison')} branch ({target_branch}): {hl(str(reference_violation_count))}
  Violations on {hl('your')} branch         {' '*len(target_branch)}: {hl(str(other_violation_count))}
""")
```

I'd also suggest making these large blobs of text prints separate functions, so it the source reads e.g. `print_status(target_branch, reference_violation_count, other_violation_count)`

chore/check-javadoc-regressions.py
```diff
@@ -0,0 +1,272 @@
...
+
+        if reference_violation_count < other_violation_count:
+            print(warn("Javadoc quality has deteriorated!"))
+            print("Run the chore/ci-checkstyle-javadoc.sh script locally to find errors")
+            print(f"See {hl('https://github.com/inria/spoon/issues/3923')} for details")
+            print()
+            print_regression_lines(reference_output, other_output)
+            exit(1)
```
slarse, line 211: Same as above, I'd just print one large string. This string could even be a constant stored somewhere else.

chore/check-javadoc-regressions.py
```diff
@@ -0,0 +1,272 @@
...
+        elif reference_violation_count == other_violation_count:
+            print(success("Javadoc quality has not deteriorated!"))
+        else:
+            print(success("You increased the Javadoc quality! Thank you :)"))
```
slarse, line 215: This is a very minor note, but it seems potentially confusing to me to say that one has "increased" the quality, when that corresponds to the quality score decreasing (recall that the score is just the amount of warnings).

```suggestion
            print(success("You improved the Javadoc quality! Thank you :)"))
```

chore/check-javadoc-regressions.py
```diff
@@ -0,0 +1,272 @@
...
+
+
+@contextmanager
+def temporary_path(suffix: str):
+    """
+    Creates and returns a Path to a temporary file that will be deleted when
+    the context manager exits. This is needed as NamedTemporaryFile is
+    hopelessly broken on Windows due to mandatory file locks.
+    See https://bugs.python.org/issue14243.
+    """
```
slarse, line 27: A very minor note on [docstring conventions (PEP 257)](https://www.python.org/dev/peps/pep-0257/).

Docstrings that are "sufficiently short" go one line like so:

```python
def func():
    """A short docstring."""
```

Multi-line docstrings start with a summary line, and then have a body:

```python
def func():
    """This is a longer docstring that stretches over
    multiple lines, and starts with this summary line.

    Then, additional details are written in the body
    down here, and finally the terminating quotes are
    on a line of their own.
    """
```

Concretely, this docstring could look like so:

```suggestion
    """Creates and returns a Path to a temporary file that will be deleted
    when the context manager exits.

    This is needed as NamedTemporaryFile is hopelessly broken on Windows
    due to mandatory file locks. See https://bugs.python.org/issue14243.
    """
```

It doesn't particularly matter if you edit that for this script or not and it's not something that one must strictly adhere to when writing simple scripts, just thought I'd point it out as I spotted it.

chore/check-javadoc-regressions.py
```diff
@@ -0,0 +1,272 @@
...
+
+
+def find_added_violation_lines(reference: 'Counter[str]', other: 'Counter[str]') -> 'Counter[str]':
+    """
+    Tries to find out which violation lines are new in "other" and which
+    were already present in the "reference". These lines are added violations the
+    contributor should fix.
+    """
+    # Lines that are only in the reference aren't critical, as they were
+    # apparently fixed in "other"
```
slarse, line 120: This should either go into the function docstring or be removed, IMO.

chore/check-javadoc-regressions.py
```diff
@@ -0,0 +1,272 @@
...
+    )
+    with_line_numbers = try_readd_line_numbers(new_errors, lines_other)
+
+    print(warn("You likely added the following new errors. "), end=" ")
+    print("Line numbers might be incorrect if there are multiple violations of the same type in the file.")
+    for error in with_line_numbers:
+        print(error.strip())
+
+
+def print_compare_with_branch(target_branch: str) -> None:
```
slarse, line 168: This function is pretty large, could you split it into a few smaller ones?

slarse (changes requested):
Overall it LGTM, functionality seems like what we need. Definite improvement from the bash script. I just have a few comments on the code that I think can be improved, and a few general comments that you don't really need to worry about (more as FYIs).

chore/check-javadoc-regressions.py
I-Al-Istannen: Maybe that is nicer, I just don't really like that pythons text blocks consider *all* the whitespace to the left, forcing you to break the indentation of your program. This wouldn't be as bad in Java, but as indentation is semantically important in Python it feels a bit off to have a string hanging out at the start of the line.

I've changed it though.

chore/check-javadoc-regressions.py
I-Al-Istannen: I split it up a bit. It still does three distinct things, but they sadly need the output of the previous stages. I could maybe give that some more thought and try to restructure it completely, but I'd leave it as is if you don't want me to do that.

slarse: I think the new split is perfectly. If this was production code perhaps a more thorough redesign would be warranted, but this is a support script, and by support script standards it's very well put together IMO.

chore/check-javadoc-regressions.py
slarse: > I just don't really like that pythons text blocks consider all the whitespace to the left

That is a bit annoying, but working around it is pretty easy if one is prone to be bothered by such things.

```python
def dedent(s: str, start_char: str = "|") -> str:
    return "\n".join([substring(line, start_char) for line in s.split("\n") ])

def substring(s: str, from_char: str) -> str:
    start_idx = s.find(from_char) + 1
    return s[start_idx : ]

s = """Hello there
    |This is just a little test
    |of dedenting a string
"""

print(dedent(s))
```

That outputs 

```
Hello there
This is just a little test
of dedenting a string
```

But yeah, a bit overkill for a support script like this.

I-Al-Istannen: Thank you! :) Yea, it is probably fine here. The strings only appear in one place and it isn't too hard to follow, so I will probably keep your initial suggestion and leave them as-is.

Sometimes you just get a bit spoiled by features in other languages...

slarse (approved):
This looks good to me, will merge today! Thanks for taking the time, this is much better than a bash script.

I-Al-Istannen:
No worries, it was annoying me ;)

Maybe it would be worth it to update the referenced issue https://github.com/inria/spoon/issues/3923 a bit? Not sure if it's needed, but we still link to it and a few things have changed. It will e.g. fail to checkout if you have incompatible files:
```
$ git rev-parse --show-toplevel
$ git checkout master
Error ececuting native command: ['git', 'checkout', 'master']
Error output:
  error: Your local changes in the following files would be overwritten
  	chore/check-javadoc-regressions.py
  Please commit or stash your changes before switching branches.
  Abort
```
This won't happen in CI as it doesn't edit any files here and as a user it sounded a bit nicer than bold warnings in a document you might not read. But this and a few other things might benefit from some updating.

Now that I re-read the issue I think a `--help` flag that prints a high level summary would help. I'm sorry I missed that before but it would probably come in handy if you expect people to run that script on their own.

I'll also prepare an updated version of your issue and post it in a comment here, then you can decide whether you want to use some parts of it.

chore/check-javadoc-regressions.py
```diff
@@ -0,0 +1,327 @@
...
+if __name__ == "__main__":
+    if os.name == "nt":
+        fix_colors_windows_cmd()
+
+    if "--help" in sys.argv:
+        print_help()
+        exit(0)
+
+    if len(sys.argv) > 2:
+        print(f"Usage: {sys.argv[0]} [COMPARE_WITH_MASTER | regex file filter]")
```
slarse, line 312: Perhaps also add `--help` here? Otherwise it'll be hard to know one can use it. This is why I almost always go for using `argparse` whenever I have a script that's in any way non-trivial :)

```suggestion
        print(f"Usage: {sys.argv[0]} [COMPARE_WITH_MASTER | regex file filter] [--help]")
```

I-Al-Istannen: I was tempted to use it, but properly configuring it to reject invalid input here wasn't really trivial and the help page isn't what argparse would generate either (but hopefully easier to read, see image below). I was also clinging to the hope that the script was trivial, but the length of the conversations in the two issues might prove me wrong...

I will make it call `print_help()` in there, I think. Then the usage is consistent.

grafik

I-Al-Istannen:
Feel free to suggest changes for the help and change what you feel is reasonable (nothing to more than listed here) in your issue comment.

I hope this is finally *actually* ready to be merged... Sorry about that.

----

Help output:
grafik

Outdated parts of the comment:
```diff
- To facilitate this, we have a script to find checkstyle errors in method Javadoc at chore/ci-checkstyle-javadoc.
+ To facilitate this, we have a script to find checkstyle errors in method Javadoc at chore/check-javadoc-regressions.
```
----
The script compatibility is a bit of a mixed bag. It should work everywhere, but if you add maven and git to your user path settings it might fail on Windows. We should probably still caution against using it on plain Windows.

---

```diff
- Bug: At the time of writing, the script doesn't work at all if you're on the master branch, so checkout to another branch to use it!
```
It works, but it doesn't really make much sense to compare `master` to `master`.

----

```diff
- chore/ci-checkstyle-javadoc.sh
+ chore/check-javadoc-regressions.py
```
----
```diff
- but do note that running in this mode destroys uncommitted changes. So don't run it unless you've committed all changes!
+ 
```
----
And we might want to mention the `--help` flag somewhere.

slarse:
Thanks @I-Al-Istannen, very nice work! I'll update the issue.

PR #4037, opened 2021-07-11
review: fix(test): error output for CtInheritanceScanner test prints wrong class name

SirYwell:
When working on my other pull request, I came about this confusing output:

```cs
[ERROR] Failures: 
[ERROR]   SpoonArchitectureEnforcerTest.testInterfacesAreCtScannable:358 The following methods are missing in CtScanner: 
scanCtPattern
```
I wondered why it would expect me to add this method to the CtScanner as there are no other methods like that. Looking closer at the test revealed that the method is expected in CtInheritanceScanner instead. This PR uses the simple class name instead, so `CtInheritanceScanner` is now printed correctly.

It also seems like the CtScanner is not required as input resource. I guess that is a leftover from when this test was implemented.

**Note**: I'm not sure about this, but it the test case itself is probably named badly, indicating that it is about the CtScanner instead of the CtInheritanceScanner. If you agree on this, I can change the name of the test case too.


slarse (approved):
Good find @SirYwell. Will merge.

PR #4038, opened 2021-07-11
review: fix: Nested enums modifiers

### LINKED ISSUE #4036: Inheriting public/static modifier in interface member types
## Problem statement
```java
interface Foo {
  enum Bar {}
}
```
For this Java code `bar.isPublic()`/`bar.isStatic()` returns `false`, as the `Bar` enum does not specify the `public` (or `static`) modifier. However, §9.5 of the JLS specifies (Emphasis mine):
> Every member class or interface declaration in the body of an interface declaration is **implicitly public and static** ( §9.1.1.3). It is permitted to **redundantly** specify either or both of these modifiers

## Tangential problem
Additionally, §8.9 also states
> **A nested enum class is implicitly static.** That is, every member enum class and local enum class is static. It is permitted for the declaration of a member enum class to redundantly specify he static modifier, but it is not permitted for the declaration of a local enum class (§14.3)

So `bar.isStatic()` ought to return true for this reason as well.

## Question
I would expect `isPublic()` to return true there, as the Enum *does* possess the modifier - even if it is not spelled out explicitly. Many tools (like IntelliJ) will automatically remove / warn when specifying `public` for interface members and I wouldn't expect such a change to have semantic impact on Spoon's model.

So, should those types inherit the `public` and `static` modifiers? I do believe that is the expected behavior, but maybe you have a different opinion.

I-Al-Istannen:
## About
This PR tries to improve implicit modifier support in Spoon. For more details see #4036.

## Questions
Why does `CtModifierHandler#setExtendedModifiers` reject empty sets? I don't see any other way to remove the modifiers I added, so I removed the check for now. This looks a bit dangerous, but maybe I will find out why that's there while fixing tests.

*EDIT: I didn't, I don't see any reason why you'd want that in the if condition currently*

## Closes
Closes #4036

### DIFF
--- a/src/main/java/spoon/support/reflect/declaration/CtTypeImpl.java
+++ b/src/main/java/spoon/support/reflect/declaration/CtTypeImpl.java
@@ -16,6 +16,7 @@ import spoon.reflect.declaration.CtAnnotation;
 import spoon.reflect.declaration.CtAnnotationType;
 import spoon.reflect.declaration.CtCompilationUnit;
 import spoon.reflect.declaration.CtConstructor;
+import spoon.reflect.declaration.CtEnum;
 import spoon.reflect.declaration.CtExecutable;
 import spoon.reflect.declaration.CtField;
 import spoon.reflect.declaration.CtFormalTypeDeclarer;
@@ -60,6 +61,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 /**
  * The implementation for {@link spoon.reflect.declaration.CtType}.
@@ -248,12 +250,35 @@ public abstract class CtTypeImpl<T> extends CtNamedElementImpl implements CtType
 
 	@Override
 	public <N, C extends CtType<T>> C addNestedType(CtType<N> nestedType) {
-		return addTypeMember(nestedType);
+		addTypeMember(nestedType);
+
+		// Nested enums are implicitly static. We need to add the modifier, if it isn't already static.
+		if (nestedType instanceof CtEnum && !nestedType.isStatic()) {
+			Set<CtExtendedModifier> modifiers = new HashSet<>(nestedType.getExtendedModifiers());
+			modifiers.add(new CtExtendedModifier(ModifierKind.STATIC, true));
+			nestedType.setExtendedModifiers(modifiers);
+		}
+
+		return (C) this;
 	}
 
 	@Override
 	public <N> boolean removeNestedType(CtType<N> nestedType) {
-		return removeTypeMember(nestedType);
+		if (!removeTypeMember(nestedType)) {
+			return false;
+		}
+
+		// Nested enums are implicitly static. We might have added an implicit static modifier so we
+		// need to remove it again.
+		if (nestedType instanceof CtEnum) {
+			nestedType.setExtendedModifiers(
+					nestedType.getExtendedModifiers().stream()
+							.filter(mod -> mod.getKind() != ModifierKind.STATIC || !mod.isImplicit())
+							.collect(Collectors.toCollection(HashSet::new))
+			);
+		}
+
+		return true;
 	}
 
 	@Override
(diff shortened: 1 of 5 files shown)


I-Al-Istannen:
I am not quite sure how I should handle the test failures.

----

I believe
```
[ERROR]   AstCheckerTest.testPushToStackChanges:144 removeNestedType(spoon.reflect.declaration.CtType) on spoon.support.reflect.declaration.CtInterfaceImpl
removeNestedType(spoon.reflect.declaration.CtType) on spoon.support.reflect.declaration.CtTypeImpl
```
is a false positive. I never push, but the method I delegate to does. I just don't *only* delegate, so the `isDelegate` check fails.

----

I believe `testDefaultMethodInInterface` is actually an incorrect test. The method is implicitly public and therefore `public` should not be printed.

----

I believe `testNestedPrivateEnumValues` is actually an incorrect test. Nested enums should be static, so I have (or will) adjusted this test.

----

I believe `testFieldAndMethodInAnnotation` is actually an incorrect test. The members in an annotation are treated the same as in an interface, as Annotations are "just" special interfaces.

I-Al-Istannen:
I think the only remaining failing test is a false positive but I am not quite sure how to handle it.

```
[ERROR]   AstCheckerTest.testPushToStackChanges:144 removeNestedType(spoon.reflect.declaration.CtType) on spoon.support.reflect.declaration.CtInterfaceImpl
removeNestedType(spoon.reflect.declaration.CtType) on spoon.support.reflect.declaration.CtTypeImpl
```
I never push, but the method I delegate to does. I just don't *only* delegate, so the `isDelegate` check fails.

slarse:
I tend to agree that the test failures are false positives due to AstCheckerTest being too inflexible. I'll open a new issue about that and get back to this PR afterwards.

I-Al-Istannen:
I know, force pushes are discouraged but there were no reviews yet and it was the cleanest way to get the 48 commits I was lacking from master - this PR needed the latest one refactoring the AST check. I am also not sure why that even works as I didn't adjust it to the single-variable-return structure, but I will take it.

With that this PR is probably ready for review.

src/main/java/spoon/support/reflect/declaration/CtInterfaceImpl.java
```diff
@@ -109,4 +113,48 @@ public boolean isInterface() {
...
+		// We might have added implicit public static modifiers so we need to remove them again
+		Set<CtExtendedModifier> newModifiers = nestedType.getExtendedModifiers().stream()
+				.filter(modifier -> {
+					if (!modifier.isImplicit()) {
+						return true;
+					}
+					return modifier.getKind() != ModifierKind.STATIC
+							&& modifier.getKind() != ModifierKind.PUBLIC;
+				})
+				.collect(Collectors.toCollection(HashSet::new));
```
MartinWitt, line 154: Minor Style comment: convert the filter in 2 filter statements, which are shorter and don't need `{... }` brackets

I-Al-Istannen: That has different semantics but if you got confused and I needed a double-take, this code has to go. I will rewrite that with a `removeIf` later :)

slarse (commented):
Like I mentioned in #4036, this seems like two unrelated fixes to me. There's 1) nested types in interfaces, and 2) nested enums in general.

Is there any reason this cannot be 2 PRs, one for each of those problems?

I-Al-Istannen:
No, there isn't (Maybe the change in the CtModifierHandle but that is small enough I can just fix it if it conflicts). I thought it was small enough for a single PR, but I'll split it in two if you prefer.

I-Al-Istannen:
The test fails as this PR now depends on #4176. The test expects `public enum`, this PR now adds an implicit `static` modifier which gets lots when cloning and the printer therefore writes out `public static enum`.

I-Al-Istannen:
I split this up into three logical changes:
1. Keep implicit modifiers when cloning (by using set/get ExtendedModifiers #4176 
2. Make nested enums static #4038 
3. Fix modifiers of types nested in interfaces #4175 

Sadly 1. is blocked by a test failure that ought to have happened 2 years ago but didn't due to the bug that PR fixes and this PR depends on it for reasons outlined above.

I hope the three PRs are easier to review even though they have some interactions.

slarse:
@I-Al-Istannen I think you can go ahead here now?

I-Al-Istannen:
I don't quite care enough, so I'll close this PR. Enum staticness is not quite as important as correctly setting the visibility of types you add to interfaces so w/e.

SirYwell:
> @SirYwell can you confirm that, you're "explicitly" setting the implicit modifiers when building the model?

Yes, JDT provides the implicit modifiers in the corresponding `binding`, which *is* already covered for fields and methods but wasn't covered for types.

PR #4056, opened 2021-07-19
review: test: add test for removeAnnotation method

Rohitesh-Kumar-Jain:
#1818

Hi, I have added a new test for the method `removeAnnotation` :

These tests are intended to kill this mutation:

Screenshot 2021-07-19 at 11 14 20 PM

Link to #3967


src/test/java/spoon/test/ctElement/CtElementImplTest.java
```diff
@@ -0,0 +1,41 @@
...
+        spoon.buildModel();
+        Factory factory = spoon.getFactory();
+
+        CtClass<Annotated> annotatedClass = factory.Class().get(Annotated.class);
+        assertEquals(1, annotatedClass.getAnnotations().size());
+        CtAnnotation annotationToBeRemoved = annotatedClass.getAnnotations().get(0);
+
+        // act
+        Boolean expectedBoolean = annotatedClass.removeAnnotation(annotationToBeRemoved);
+        Boolean expectedBoolean2 = annotatedClass.removeAnnotation(annotationToBeRemoved);
```
MartinWitt, line 31: Small comments,

here, you could use `boolean` instead of `Boolean`. `CtElement#removeAnnotation` returns boolean. This change saves a reader the noise if the Boolean could be ever null.

Renaming `expectedBoolean` to something like `shouldBeTrue` and `expectedBoolean2`
to e.g. `shouldBeFalse` helps a reader and is a way more meaningful name.

Rohitesh-Kumar-Jain: Yes, you are right, thanks for the wonderful suggestions.

Feel free to contribute suggestions : )

src/test/java/spoon/test/ctElement/CtElementImplTest.java
```diff
@@ -0,0 +1,41 @@
...
+
+        // arrange
+        Launcher spoon = new Launcher();
+        spoon.addInputResource(new FileSystemFile("./src/test/java/spoon/test/ctElement/CtElementImplTest.java"));
+        spoon.buildModel();
+        Factory factory = spoon.getFactory();
+
+        CtClass<Annotated> annotatedClass = factory.Class().get(Annotated.class);
+        assertEquals(1, annotatedClass.getAnnotations().size());
+        CtAnnotation annotationToBeRemoved = annotatedClass.getAnnotations().get(0);
```
slarse, line 27:
```suggestion
        CtAnnotation<?> annotationToBeRemoved = annotatedClass.getAnnotations().get(0);
```

src/test/java/spoon/test/ctElement/CtElementImplTest.java
```diff
@@ -0,0 +1,41 @@
...
+        // contract: removeAnnotation returns true after removing an annotation of a class containing a single
+        // annotation, and returns false when a non existing annotation is tried to be removed
+
+        // arrange
+        Launcher spoon = new Launcher();
+        spoon.addInputResource(new FileSystemFile("./src/test/java/spoon/test/ctElement/CtElementImplTest.java"));
+        spoon.buildModel();
+        Factory factory = spoon.getFactory();
+
+        CtClass<Annotated> annotatedClass = factory.Class().get(Annotated.class);
```
slarse, line 25: We don't typically add resources inside of the actual test file, as that forces Spoon to parse a lot more stuff than just the intended resource. Arguably, it was a mistake by me to suggest it for the template tests.

Anyway, here your test resource is so small that you can just add it as inline text.

```suggestion
        CtClass<?> annotatedClass = Launcher.parseClass("@SuppressWarnings(\"unchecked\") class Annotated { }");
```

slarse (changes requested):
Hi @Rohitesh-Kumar-Jain,

Sorry about the slow response these days. I'm prioritizing your PRs as I know you're doing the GSOC thing, but I'm just not around that much during vacation times.

Anyway, the test here looks pretty good, but there are a couple of fundamental mistakes/miscomprehensions.

First of all, we don't typically test implementations, but rather test the public interfaces (e.g. `CtElement`). Second, you're not actually explicitly testing `CtElementImpl` in this test, you're testing the `CtClass` interface. How that interface's `removeAnnotation` method is implement is an unspecified implementation detail that could change at any time, making the test's specification non-robust.

To fix this, I would either a) place this test inside `CtClassTest` or b) explicitly instantiate a `CtElementImpl` and test that. I would favor a) as we don't typically test specific implementation details, but the choice is up to you.

src/test/java/spoon/test/ctElement/CtElementImplTest.java
slarse: Thanks @MartinWitt!

While this naming explains the expected result, it doesn't well represent _what_ the values are. One could for example consider this naming instead:

```java
        boolean firstRemovalSuccessful = annotatedClass.removeAnnotation(annotationToBeRemoved);
        boolean secondRemovalSuccessful = annotatedClass.removeAnnotation(annotationToBeRemoved);
```

This doesn't encode what the expected value is, but that is in fact encoded in the assertion. I'm not putting this in as a review comment as I think the names suggested by @MartinWitt are perfectly serviceable and read just fine, I just wanted to plant the thought.

slarse: On a side note: if you ever feel the need to enumerate variables (e.g. `something1`, `something2` etc), that's a clear indication that you need to reconsider the naming :)

Rohitesh-Kumar-Jain:
> Sorry about the slow response these days. I'm prioritizing your PRs as I know you're doing the GSOC thing, but I'm just not around that much during vacation times.

It's perfectly fine, do enjoy vacations 🎉, I am too going to grandma's house just after this ends

Thanks for prioritizing PRs : )

> First of all, we don't typically test implementations, but rather test the public interfaces

I will keep this in mind for future

>  How that interface's removeAnnotation method is implement is an unspecified implementation detail that could change at any time, making the test's specification non-robust.

That didn't came to my mind, thanks for letting me know

> To fix this, I would either a) place this test inside CtClassTest or b) explicitly instantiate a CtElementImpl and test that. 

a) seems fine to me as well

Rohitesh-Kumar-Jain:
> a) seems fine to me as well

For a) realized I first need to migrate `CtClassTest` it to Junit5

slarse:
> For a) realized I first need to migrate CtClassTest it to Junit5

This PR is already started, so I'm perfectly fine with you skipping that in this case.

slarse (approved):
Very good @Rohitesh-Kumar-Jain!

PR #4063, opened 2021-07-26
review: fix: Fix the insertion order for CtCaseImpl.insertBegin

### LINKED ISSUE #4060: bug: insertBegin method is inserting list in reverse order than it should be
## Context

Hi while writing test for [`CtCaseImpl.insertBegin`](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/reflect/code/CtCaseImpl.java), I feel that the core logic of this method is implemented incorrectly.

JavaDoc for this method states : 

Screenshot 2021-07-23 at 5 13 24 PM

## Problem 

`insertBegin` should add the list at the beginning of the block in the order of the elements inside the list, but it is adding in the reverse order of the elements inside the list.

What I mean is that after inserting a list consisting of statements : 

[first statement, second statement, third statement] they should be added like this : 

```
first statement 
second statement 
third statement
existing statement(s)
```

But they are being added like

```
third statement 
second statement 
first statement 
existing statement(s)
```
The existing code is working in the way as you are inserting individual statements, and not the whole list

It is based on the reasoning that the addition should be like this : 

```
list
existing statements(s)
```

And as the list is inserted in the beginning, the content of the list shouldn't be altered

## Failing test Case

 I have written this test to enforce this

```
@Test
    void testInsertBeginWithListOfStatements() {
        // contract: insertBegin add a list of statements at the beginning of a case, i.e above two already existing statements

        // arrange
        spoon.addInputResource(new FileSystemFile("./src/test/java/spoon/test/ctCase/testclasses/ClassWithSwitchExample.java"));
        spoon.buildModel();

        CtClass classWithSwitchExample = spoon.getFactory().Class().get(ClassWithSwitchExample.class);

        CtStatement firstStatementToBeInserted = classWithSwitchExample.getFactory().Code().createCodeSnippetStatement("first Statement");
        CtStatement secondStatementToBeInserted = classWithSwitchExample.getFactory().Code().createCodeSnippetStatement("second Statement");
        CtStatement thirdStatementToBeInserted = classWithSwitchExample.getFactory().Code().createCodeSnippetStatement("third Statement");
        CtCase testCase = classWithSwitchExample.getElements(new TypeFilter<>(CtCase.class)).get(0);

        CtStatementList ctStatementList =  new CtStatementListImpl<CtStatement>();

        ctStatementList.addStatement(firstStatementToBeInserted);
        ctStatementList.addStatement(secondStatementToBeInserted);
        ctStatementList.addStatement(thirdStatementToBeInserted);

        // act
        testCase.insertBegin(ctStatementList);

        // assert
        assertEquals(firstStatementToBeInserted, testCase.getStatement(0));
        assertEquals(secondStatementToBeInserted, testCase.getStatement(1));
        assertEquals(thirdStatementToBeInserted, testCase.getStatement(2));
    }
```

## Solution 

I will be happy to fix this bug, once given permission to work upon this from the community : )

I will be pushing this test along with other tests I have written for this Class.

Rohitesh-Kumar-Jain:
WIP

Fix #4060

Also, this test is killing this mutation

Screenshot 2021-07-26 at 11 32 58 AM

link to #3967


src/test/java/spoon/test/ctCase/CtCaseImplTest.java
```diff
@@ -0,0 +1,58 @@
...
+    private static CtClass getClassWithSwitchExample() {
+        Launcher spoon = new Launcher();
+        String pathToClassWithSwitchExample = "./src/test/java/spoon/test/ctCase/testclasses/ClassWithSwitchExample.java";
+        spoon.addInputResource(new FileSystemFile(pathToClassWithSwitchExample));
+        spoon.buildModel();
+
+        return spoon.getFactory().Class().get(ClassWithSwitchExample.class);
+    }
+
+    public static CtStatement createStatement(CtClass ctClass, String Statement) {
```
slarse, line 55: always start with a lowercase character for variable names

```suggestion
    public static CtStatement createStatement(CtClass ctClass, String statement) {
```

src/test/java/spoon/test/ctCase/CtCaseImplTest.java
```diff
@@ -0,0 +1,58 @@
...
+        return ctClass.getFactory().Code().createCodeSnippetStatement(Statement);
```
slarse, line 56: Always compile code snippet statements when using them to ensure that you actually have reasonable code (in this test, you actually don't).

```suggestion
        return ctClass.getFactory().Code().createCodeSnippetStatement(Statement).compile();
```

src/test/java/spoon/test/ctCase/CtCaseImplTest.java
```diff
@@ -0,0 +1,58 @@
...
+    @Test
+    void testInsertBeginWithListOfStatements() {
+        // contract: insertBegin adds a list of statements at the beginning of a case, i.e the list is added above two
+        // exiting statements, and the order of statements in the list remains the same
+
+        // arrange
+        CtClass classWithSwitchExample = getClassWithSwitchExample();
+        CtStatement firstStatementToBeInserted = createStatement(classWithSwitchExample, "firstStatement");
+        CtStatement secondStatementToBeInserted = createStatement(classWithSwitchExample, "secondStatement");
+        CtCase testCase = classWithSwitchExample.getElements(new TypeFilter<>(CtCase.class)).get(0);
```
slarse, line 27: Always prefer writing test cases without using a test resource when it's not too complicated to do so. That makes the test cases more self-contained. In this case, it's very easy to just create the `CtCase` you want to test with a factory.

```suggestion
        Factory factory = new Launcher().getFactory();
        CtCase<?> testCase = factory.createCase();
        testCase.addStatement(factory.createCodeSnippetStatement("int preexisting = 42;").compile());

        CtStatement firstStatementToBeInserted = factory.createCodeSnippetStatement("int first = 1;").compile();
        CtStatement secondStatementToBeInserted = factory.createCodeSnippetStatement("int second = 2;").compile();
```

slarse: Good names on the statements (e.g. `firstStatementToBeInserted`) by the way!

src/test/java/spoon/test/ctCase/CtCaseImplTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+        CtStatementList ctStatementList =  new CtStatementListImpl<CtStatement>() {
+            {
+                addStatement(firstStatementToBeInserted);
+                addStatement(secondStatementToBeInserted);
+            }
+        };
```
slarse, line 34: Don't create anonymous subtypes unless it's actually part of the test. Here, just use the factory.

```suggestion
        CtStatementList statementList = factory.createStatementList();
        statementList.addStatement(firstStatementToBeInserted);
        statementList.addStatement(secondStatementToBeInserted);
```

src/test/java/spoon/test/ctCase/CtCaseImplTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+        // act
+        testCase.insertBegin(ctStatementList);
+
+        // assert
+        int firstStatementIndex = 0;
+        int secondStatementIndex = 1;
```
slarse, line 41: Perhaps a bit overkill naming these indices :)

slarse (changes requested):
Implementation LGTM. The test can be improved.

First of all, like I mentioned in another review, we don't typically test the specific implementations directly, and in fact you're not testing `CtCaseImpl` directly here. We rely on the factories to create the implementations for us. There is already a test class for cases (called `SwitchCaseTest` for some reason instead of `CtCaseTest`, we should rename that). Put the test in there.

Second, the indentation in the test class is incorrect (should be tabs). Make sure that the test case, when moved, has tabs as indentation.

See also the comments on the code.

src/test/java/spoon/test/ctCase/CtCaseImplTest.java
Rohitesh-Kumar-Jain: I actually intended to initialize the list with those statements that's why created that way

src/test/java/spoon/test/ctCase/CtCaseImplTest.java
Rohitesh-Kumar-Jain: yes it seemed to me as well 😅

Rohitesh-Kumar-Jain:
> There is already a test class for cases (called SwitchCaseTest for some reason instead of CtCaseTest, we should rename that). Put the test in there.

Yeah it should be there, also I will migrate it along with the other test I am working parallel

> Second, the indentation in the test class is incorrect (should be tabs). Make sure that the test case, when moved, has tabs as indentation.

Yeah, sure IDEs have an auto refactoring feature for I will use that to get the best tab spacing

src/main/java/spoon/support/reflect/code/CtCaseImpl.java
```diff
@@ -166,10 +166,9 @@ private void ensureModifiableCaseExpressionsList() {
 	@Override
 	public <T extends CtStatementList> T insertBegin(CtStatementList statements) {
 		this.ensureModifiableStatementsList();
-		for (CtStatement statement : statements.getStatements()) {
-			statement.setParent(this);
-			this.addStatement(0, statement);
-		}
+		List<CtStatement> copy = new ArrayList<>(statements.getStatements());
+		statements.setStatements(null);
+		this.statements.addAll(0, copy);
```
SirYwell, line 171: I'm not sure how important this is, but the `addStatement(int, CtStatement)` method does a little bit more:
a) it calls `statement.setParent(this);` for each CtStatement
b) it calls `onListAdd`

I think it is necessary to loop over all statements to add at some point to keep that behavior. I'm not sure if `List#addAll(int, Collection)` is still prefered then.
@slarse WDYT?

slarse: You're absolutely right. I just had it in my mind that the statement list here was a model list (as it is in CtBlockImpl), but it's not. It's just an `ArrayList`.

It seems to me like the statement and expression lists in `CtCaseImpl` should be model lists as well, but I'm not sure about the design decision for them not being so currently. It could simply be that model lists were introduced at a later time, and were never added to `CtCaseImpl`. I think I'll want to wait for @monperrus or @nharrand to weigh in there before we proceed.

@Rohitesh-Kumar-Jain Can you add assertions to check that the parents of the inserted statements are set correctly? Something like so:

```java
        assertThat(firstStatementToBeInserted.getParent(), is(testCase));
        assertThat(secondStatementToBeInserted.getParent(), is(testCase));
```

Note the use of identity (`is`) as opposed to equality.

Rohitesh-Kumar-Jain:
> Second, the indentation in the test class is incorrect (should be tabs). Make sure that the test case, when moved, has tabs as indentation.

I don't get this one, also I ran IntelliJ's auto code reformat, but it failed to detect what are you pointing at ig

src/main/java/spoon/support/reflect/code/CtCaseImpl.java
Rohitesh-Kumar-Jain: Yes I have added these tests as well

Rohitesh-Kumar-Jain:
@slarse I have implemented all the suggested changes except for this

 > Second, the indentation in the test class is incorrect (should be tabs).

I am not able to understand this one

src/test/java/spoon/test/ctCase/CtCaseImplTest.java
I-Al-Istannen: I like [this answer](https://stackoverflow.com/a/27521360) by Lukas Eder. Double brace init can be the cause of subtile bugs and isn't really buying you much. The memory downsides don't really apply here as it is in a test but seeing it in real code always tingles my spidey-senses.

slarse (approved):
@Rohitesh-Kumar-Jain Nice fix, looks good to me.


> > Second, the indentation in the test class is incorrect (should be tabs).

> I am not able to understand this one

When you create a new file with IntelliJ, the default is to create it with spaces as indentation. If you look at your first commit, the test class has spaces as indentation. If you just run auto-formatting on that, it will still format with spaces.

When you pasted the test case into the existing test class, IntelliJ detected the use of tabs and changed the indentation automatically.

Rohitesh-Kumar-Jain:
> When you pasted the test case into the existing test class, IntelliJ detected the use of tabs and changed the indentation automatically.

Thanks for letting me, and I hope you enjoyed your vacations and the Olympics matches : )

slarse:
> > When you pasted the test case into the existing test class, IntelliJ detected the use of tabs and changed the indentation automatically.
> 
> Thanks for letting me, and I hope you enjoyed your vacations and the Olympics matches : )

Vacations: yes. Olympics: I'm just about vaguely aware they exist :)

PR #4068, opened 2021-07-27
fix: Fix compliance level support for old code

### LINKED ISSUE #4067: Compliance level is not used in keyword constraint
## Context

In [`CtReferenceImpl`](https://github.com/INRIA/spoon/blob/34cfb5d6178c63f34e851acf24b22f9c10fa175b/src/main/java/spoon/support/reflect/reference/CtReferenceImpl.java), the function `checkIdentiferForJLSCorrectness` checks names for correctness, which includes checking wether any part of it is a keyword:

https://github.com/INRIA/spoon/blob/34cfb5d6178c63f34e851acf24b22f9c10fa175b/src/main/java/spoon/support/reflect/reference/CtReferenceImpl.java#L123-L125

## Issue

Not all keywords have been in the Java specification since the beggining (see https://docs.oracle.com/javase/tutorial/java/nutsandbolts/_keywords.html). Namely, `enum` was only introduced in `5.0`.

The issue with setting keywords in this way is that there is no way of using the specified compliance level to build models that do use keywords like `enum` with lower than `5.0` compliance level.

Example: https://github.com/SpoonLabs/astor/blob/ff5dc0b1a4fbc1cb2fe72c68189a5ba02fa81a9b/examples/lang_63/src/test/java/org/apache/commons/lang/enum/EnumTest.java#L18

andre15silva:
This PR fixes #4067

Added a test case similar to the problem encountered with the example from `astor`.

### DIFF
--- a/src/main/java/spoon/support/reflect/reference/CtReferenceImpl.java
+++ b/src/main/java/spoon/support/reflect/reference/CtReferenceImpl.java
@@ -32,6 +32,13 @@ public abstract class CtReferenceImpl extends CtElementImpl implements CtReferen
 
 	private static final long serialVersionUID = 1L;
 
+	// See isKeyword for more information on keywords
+	private static final Collection<String> baseKeywords = fillWithBaseKeywords();
+	private static final Collection<String> java2Keywords = Stream.of("strictfp").collect(Collectors.toCollection(HashSet::new));
+	private static final Collection<String> java4Keywords = Stream.of("assert").collect(Collectors.toCollection(HashSet::new));
+	private static final Collection<String> java5Keywords = Stream.of("enum").collect(Collectors.toCollection(HashSet::new));
+	private static final Collection<String> java9Keywords = Stream.of("_").collect(Collectors.toCollection(HashSet::new));
+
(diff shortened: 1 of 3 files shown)


SirYwell (commented):
In regards of #4072, I've run the `CtGenerationTest` with your changes too, and it takes around 8-10s more time than with the master branch (on my machine).

It would be probably better to keep the static keyword set but remove the version specific ones and add checks for them in the `isKeyword` method directly.

andre15silva:
> In regards of #4072, I've run the `CtGenerationTest` with your changes too, and it takes around 8-10s more time than with the master branch (on my machine).
> 
> It would be probably better to keep the static keyword set but remove the version specific ones and add checks for them in the `isKeyword` method directly.

Hi @SirYwell 

You're absolutely right. Thank you and sorry for the regression.

I've updated the PR, it should be fixed now.

monperrus:
Nice fix, very helpful to analyze old Java code, which is definitely in our scope.

Will merge once CI is green.

Thanks @andre15silva

andre15silva:
Green @monperrus

PR #4072, opened 2021-07-31
review: perf: Precompile patterns for identifier checks

SirYwell:
@I-Al-Istannen and I started looking into ways to improve performance of https://github.com/I-Al-Istannen/JavadocApi.

When looking into profilings, I noticed that `Pattern.compile()` had a very high invocation count (in fact, it was the highest of all traced methods). I tracked down the callees and ended up in in methods of `CtReferenceImpl` where regular expressions were used in combination with `String#matches()`, `String#replaceAll()` and `String#split()`. Precompiling these patterns seems to have a very high impact in our case. Some of the comparisons we've done with and without this patch:

From @I-Al-Istannen with the linked JavadocApi:
```
/usr/lib/jvm/java-16-adoptopenjdk/bin/java -Xmx8G -jar target/JavadocApi.jar   1612,75s user 3,70s system 690% cpu 3:54,06 total
/usr/lib/jvm/java-16-adoptopenjdk/bin/java -Xmx8G -jar target/JavadocApi.jar   1042,61s user 3,32s system 673% cpu 2:35,21 total
```
(scanning JavaFX, upper is without patch, lower is with patch)

```
/usr/lib/jvm/java-16-adoptopenjdk/bin/java -Xmx8G -jar target/JavadocApi.jar   1276,71s user 11,40s system 424% cpu 5:03,47 total
/usr/lib/jvm/java-16-adoptopenjdk/bin/java -Xmx8G -jar target/JavadocApi.jar   978,21s user 10,06s system 452% cpu 3:38,48 total
```
(scanning OpenJDK 16, upper is without patch, lower is with patch)

With YourKit, I profiled the `CtGenerationTest`, this is a screenshot of the differences focused on the `checkIdentifierForJLSCorrectness` method:
yourkit result


src/main/java/spoon/support/reflect/reference/CtReferenceImpl.java
```diff
@@ -24,13 +24,17 @@
...
 import java.util.Objects;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import static spoon.reflect.path.CtRole.NAME;
 
 public abstract class CtReferenceImpl extends CtElementImpl implements CtReference, Serializable {
 
 	private static final long serialVersionUID = 1L;
+	private static final Pattern IS_ARRAY_OR_INSTANCE = Pattern.compile("\\[\\]|@");
```
slarse, line 35: I wonder if we could optimize this even further by creating the matcher right away. Otherwise, a new matcher object must be created on each invocation. So, something like this:

```suggestion
	private static final Matcher IS_ARRAY_OR_INSTANCE = Pattern.compile("\\[\\]|@").matches("");
```

And then you call `reset(String)` on the matcher instead of `matches(String)` on the pattern, for example:

```java
IS_ARRAY_OR_INSTANCE.reset(simpleName).replaceAll("");
```

Could you try that and see if we can squeeze out even more performance?

slarse (changes requested):
Hi @SirYwell,

This is great! I left one suggestion for how we might be able to squeeze out even more performance.

On a side note, if this turns out to still be a bottleneck after this optimization, we might want to try not using regex at all as it's kind of slow. It's probably even faster to just write custom search functions that operate directly on the string.

src/main/java/spoon/support/reflect/reference/CtReferenceImpl.java
SirYwell: Re-using the `Matcher` would make the code thread unsafe. I can look into it a little bit more, but I'd suggest to introduce JMH performance tests then for things like that.

slarse: > Re-using the Matcher would make the code thread unsafe.

Ah, good point. We could wrap it in a `ThreadLocal`, though. If you have time to try it (in the same fashion you did for this optimization, doesn't have to be a super robust performance test), then go ahead. Otherwise we'll just go with your current optimization.

slarse: Again, I think the be-all-end-all solution here is to not use regex at all. Using regex for performance-critical code is, in general, not the best solution.

slarse: On second thought, let's just ignore the matcher reuse optimization for now. We can get back to this later once we have a more robust way to evaluate performance.

src/main/java/spoon/support/reflect/reference/CtReferenceImpl.java
```diff
@@ -24,13 +24,17 @@
...
+	private static final Pattern IS_INNER_OR_GENERIC = Pattern.compile("\\.|<|>");
```
slarse, line 36: These two patterns are not used to identify arrays or generics, so I think the names are a bit misleading. Could you rename them appropriately? E.g. `IS_INNER_OR_GENERIC` would be more fittingly named `NESTED_OR_GENERIC_SPLITTER` or something like that.

src/main/java/spoon/support/reflect/reference/CtReferenceImpl.java
```diff
@@ -101,9 +105,9 @@ private void checkIdentiferForJLSCorrectness(String simplename) {
 		 */
 		//JDTTreeBuilderHelper.computeAnonymousName returns "$numbers$Name" so we have to skip them if they start with numbers
 		//allow empty identifier because they are sometimes used.
-		if (!simplename.matches("<.*>|\\d.*|^.{0}$")) {
+		if (!IS_INNER_OR_GENERIC_OR_EMPTY.matcher(simplename).matches()) {
```
slarse, line 108: The more I look at this, the more I'm thinking it's completely unnecessary to use a regex for it. The regex only actually checks the first and last characters so it seems redundant to use a regex at all. Seems to me like something like this should be faster and work just as well:

```java
	private static boolean isEmptyOrLocalOrGeneric(String identifier) {
		return identifier.isEmpty()
				|| Character.isDigit(identifier.charAt(0))
				|| (identifier.startsWith("<") && identifier.endsWith(">"));
	}
```

WDYT?

src/main/java/spoon/support/reflect/reference/CtReferenceImpl.java
MartinWitt: As @SirYwell said, the only real way to test performance is writing [JMH](https://github.com/openjdk/jmh) tests. But improving performance can't be done with a single PR, but every (small) step helps. So maybe try solutions like avoiding regex in another PR?

slarse (changes requested):
Let's ignore the optimization I suggested earlier for now. See my two new comments.

SirYwell:
I'll look into it in a few days again, thanks for your feedback so far.

src/main/java/spoon/support/reflect/reference/CtReferenceImpl.java
monperrus: oh nice, a human, static compilation of the regexp :) :+1:

SirYwell:
I spend a little bit more time on this and rewrote the logic without any regular expressions. I wrote microbenchmarks with JMH (<https://github.com/SirYwell/spoon-benchmark>, can be run with `./gradlew jmh`).

I changed the `ContractOnSettersparametrizedTest`, as it set the simple name of a `CtPackage` to `spoon.support.reflect.declaration.CtPackageImpl@1` which isn't a valid identifier (and it's the only case were a `@` was inserted). I'm not sure if the change there is appropriate but it makes the logic a little bit easier.

In general, the new implementation is more restrictive (e.g. `<<>>...@@@` was allowed as simple name before).

I also tried to add dense documentation to avoid further confusion.

src/main/java/spoon/support/reflect/reference/CtReferenceImpl.java
SirYwell: This was a misunderstanding on my side, it's actually not about generics but about names like `<init>` there. I re-used your code there, the check for anonymous/local classes was moved to the other part of the code.

src/main/java/spoon/support/reflect/reference/CtReferenceImpl.java
```diff
@@ -98,22 +98,92 @@ public boolean equals(Object o) {
...
+				default: // if we come across an illegal java identifier char here, it's not valid at all
+					if (start == i) { // first char of a part
+						if (!Character.isJavaIdentifierStart(name.charAt(i))) {
+							return false;
+						}
+					} else {
+						if (!Character.isJavaIdentifierPart(name.charAt(i))) {
+							return false;
+						}
+					}
```
slarse, line 166: Collapse?

```suggestion
					if (i == start && !Character.isJavaIdentifierStart(name.charAt(i))
						|| !Character.isJavaIdentifierPart(name.charAt(i))) {
						return false;
					}
```

src/main/java/spoon/support/reflect/reference/CtReferenceImpl.java
```diff
@@ -98,22 +98,92 @@ public boolean equals(Object o) {
...
+	 * this splits up the string into parts that need to be JLS compliant.
+	 */
+	private boolean checkAll(String name) {
+		int i = 0;
+		// leading digits come from anonymous/local classes. Skip them
+		while (i < name.length() && Character.isDigit(name.charAt(i))) {
+			i++;
+		}
+		int start = i; // used to mark the beginning of a part
+		char expectNext = 0; // 0 = do not expect anything
```
slarse, line 132: Avoid unnamed magic values :)

```suggestion
		final char anything = 0;
		char expectNext = anything;
```

src/main/java/spoon/support/reflect/reference/CtReferenceImpl.java
```diff
@@ -98,22 +98,92 @@ public boolean equals(Object o) {
...
+		for (; i < name.length(); i++) {
+			if (expectNext != 0) {
```
slarse, line 134: idem
```suggestion
			if (expectNext != anything) {
```

src/main/java/spoon/support/reflect/reference/CtReferenceImpl.java
```diff
@@ -98,22 +98,92 @@ public boolean equals(Object o) {
...
+				if (name.charAt(i) != expectNext) {
+					return false;
+				} else if (name.charAt(i) == expectNext) {
+					expectNext = 0; // reset
```
slarse, line 138:
```suggestion
					expectNext = anything;
```

slarse (changes requested):
Starting to look very good! This is significantly more explicit than using a regex. As we don't have any good performance regression tests at the moment, this PR becomes more valuable when it also improves the code itself. Very nice restructuring of the entire thing, as we already iterated over the entire string in `checkIdentifierChars` in addition to matching with regex, your restructuring is simply much better.

I have a few comments on the final code but after that I think we can merge.

As a general comment, I'd recommend toning down the use of inline comments. In many cases in this PR, they either cover up insufficiently detailed code (e.g. saying that `0` is a character to expect nothing in particular can be rewritten with a variable name) or state the obvious (e.g. that a keyword is not allowed in an identifier, when the code literally says `if this is a keyword return false`).

slarse (approved):
> I spend a little bit more time on this and rewrote the logic without any regular expressions. I wrote microbenchmarks with JMH (https://github.com/SirYwell/spoon-benchmark, can be run with ./gradlew jmh).

Completely missed this comment, but those numbers look very promising! I'm looking forward to trying this out with some of my dependent projects, based on some ad-hoc tests I ran it looks like a pretty substantial performance improvement.

Anyway, this all looks good to me now!

slarse:
Many thanks @SirYwell, this is an excellent contribution.

PR #4075, opened 2021-08-03
wip: fix: implicit modifiers reflective proxy

I-Al-Istannen:
## Blocked
This PR is blocked by #4038, as the reflective model now includes the relevant modifiers for nested types while the spoon model doesn't without that PR. This causes a test failure.

## Problem
Spoon adds implicit `public abstract` modifiers to methods (as they are present in the MethodBinding) but explicitly ignores them in the reflective model. This is IMHO quite unexpected and should be fixed.

## Solution in this PR
This PR just removes the checks added by https://github.com/INRIA/spoon/pull/1914. I couldn't really find any justification for ignoring them but maybe the MethodBinding was different back then. They were introduced in commit https://github.com/INRIA/spoon/pull/1914/commits/2fc84bb6f68ed678f71bcc4b7d78d341ead6912d if you want to have a look.

We *can not* mark the modifiers as implicit as far as I can tell, as reflection doesn't really tell us whether it was present in the source code or added implicitly.


monperrus:
done, and to stay of the positive side, let's celebrate that #4205 is merged.

Issue #3837, opened 2021-03-12
Discussion: Behavior on qualified name collisions


slarse:
After the discussion in #3827, I was intrigued by how we deal with name collisions on manipulating the AST. I've done some cursory investigations into this, and there's some truly odd behavior.

1. Creating a type or type member (e.g. with `Factory.createClass()` or `Factory.createMethod()`) that has a qualified name collision with some other element causes a silent overwrite of the original.
2. Creating a new package with `Factory.createPackage()` such that the qualified name collides with some other package, actually just returns the other package, without creating a new one. It's more or less a wrapper around `Package.getOrCreate()`.
3. Changing the simple name of a type, type member or package so as to make its qualified name collide with some other element causes two elements with the same qualified names to coexist.

This behavior is inconsistent and confusing.

The behavior of 1. and 2. should be consistent. An exception makes most sense to me, but returning the existing element is also an option (as is the case for 2.).

For 3., I think we should raise an exception.

monperrus:
Thanks for the analysis. I'm all for consistency and least confusion.

 > An exception makes most sense to me, but returning the existing element is also an option (as is 
the case for 2.).

Would prefer the latter, to return the element, it may be more backward compatible.

slarse:
> Would prefer the latter, to return the element, it may be more backward compatible.

Indeed, I think so to. To me, an exception is more "pure" in terms of design, but returning the existing value is more pragmatic given the current behavior.

PR #4076, opened 2021-08-04
wip: fix: Concurrent modification exception in modelset

### LINKED ISSUE #4071: bug: ConcurrentModificationException in CtPackageImpl ModelSet
## Problem statement
I wrote a trimmed down Scanner that processes the model in parallel, but sadly it *sometimes* fails with a concurrent modification in the CtPackageImpl / ModelSet code.
```
15:13:57.239 [Thread-6] WARN spoon.Launcher - cannot create shadow class: [Ljavax.sound.midi.VoiceStatus;
java.util.ConcurrentModificationException: null
	at java.base/java.util.TreeMap$PrivateEntryIterator.nextEntry(TreeMap.java:1480)
	at java.base/java.util.TreeMap$KeyIterator.next(TreeMap.java:1534)
	at spoon.support.util.ModelSet$Itr.next(ModelSet.java:146)
	at spoon.support.util.ModelSet$Itr.next(ModelSet.java:134)
	at spoon.support.reflect.declaration.CtPackageImpl.getType(CtPackageImpl.java:158)
	at spoon.support.visitor.java.JavaReflectionTreeBuilder.scan(JavaReflectionTreeBuilder.java:112)
	at spoon.reflect.factory.TypeFactory.get(TypeFactory.java:564)
	at spoon.support.reflect.reference.CtTypeReferenceImpl.getTypeDeclaration(CtTypeReferenceImpl.java:203)
	at de.ialistannen.javadocapi.spoon.JavadocElementExtractor.getPossiblyGenericType(JavadocElementExtractor.java:327)
	at de.ialistannen.javadocapi.spoon.JavadocElementExtractor.handleExecutable(JavadocElementExtractor.java:152)
	at de.ialistannen.javadocapi.spoon.JavadocElementExtractor.visitCtMethod(JavadocElementExtractor.java:113)
	at spoon.support.reflect.declaration.CtMethodImpl.accept(CtMethodImpl.java:58)
	at de.ialistannen.javadocapi.indexing.Indexer.lambda$main$0(Indexer.java:59)
	at de.ialistannen.javadocapi.spoon.filtering.ParallelProcessor.lambda$process$1(ParallelProcessor.java:48)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1130)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:630)
	at java.base/java.lang.Thread.run(Thread.java:831)
```

## Cause
The CtPackageImpl iterates over the set:
```java
	@Override
	public <T extends CtType<?>> T getType(String simpleName) {
		for (CtType<?> t : types) {
			if (t.getSimpleName().equals(simpleName)) {
				return (T) t;
			}
		}
		return null;
	}
```
This is problematic, as the `iterator` method delegates to the underlying set's iterator, which is a `Collections#synchronizedSet` instance. The Javadoc for that method states:
> It is imperative that the user manually synchronize on the returned collection when traversing it via [Iterator](https://docs.oracle.com/en/java/javase/16/docs/api/java.base/java/util/Iterator.html), [Spliterator](https://docs.oracle.com/en/java/javase/16/docs/api/java.base/java/util/Spliterator.html) or [Stream](https://docs.oracle.com/en/java/javase/16/docs/api/java.base/java/util/stream/Stream.html):
> 
> ``` java
> Set s = Collections.synchronizedSet(new HashSet());
> synchronized (s) {
>     Iterator i = s.iterator(); // Must be in the synchronized block
>     while (i.hasNext())
>         foo(i.next());
> }
> ```
> 
> Failure to follow this advice may result in non-deterministic behavior..

## Possible mitigations
The Java standard library doesn't offer a concurrent *ordered* set, so we can't simply transition to that. This sadly forces us to actually do what the javadoc states and synchronize on the returned set. Unfortunately, that would require us to leak a reference to the set to the caller and ensure they use it properly. I see the following easy options:

1. Leak the set in an `Object getLock()` getter and hope all clients use it (it is just CtPackage so we can adjust it easily). However, every caller must now take care to not misuse our set.
2. Make the iterator method violate the set contract and throw an exception upon invocation. Offer a callback based method like `<R> R acceptSynchronized(Function<Set<T>, R> action)` which can properly lock around the critical section.

## PRs
I have implemented 2) as I liked it a bit more and could PR it. The tests don't like it, so we'd probably end up with an iterator method that might throw a CME and an `acceptSynchronized` method that actually works.

I'd also be fine with implementing another option.

I-Al-Istannen:
Closes #4071.

## Problems
This PR involves tinkering with the ModelSet used in CtPackageImpl. I have absolutely no idea why certain invariants were chosen back then and if there are any I missed, but I tried to ensure it adheres to the same rules as before.

Sadly this also meant synchronizing access completely to prevent a race condition I think can occur in `add` and cause Thread A to miss an entry (`F`) it knows was added:
```
Thread A                 Thread B
add(F)
                         add(F)
                         set.remove(F)
for type in types:
  if type.equals...
 loopy de loop
                         set.add(F)
```

## Further notes
When running locally a coverage test for a `LovecraftException` failed and I think this summarizes the PR better than I ever could.


I-Al-Istannen:
And this created a few new great exceptions which I am pretty sure are triggered by the race condition in the iterator mentioned in the offpost. Solving these would require a lock to be held while iterating. Specializing the ModelSet a bit and letting it lookup types itself might be the nicest option.
```
Error in parallel worker
java.lang.NullPointerException: Cannot invoke "spoon.reflect.declaration.CtType.setFactory(spoon.reflect.factory.Factory)" because "newShadowClass" is null
    at spoon.reflect.factory.TypeFactory.get(TypeFactory.java:573)
    at spoon.support.reflect.reference.CtTypeReferenceImpl.getTypeDeclaration(CtTypeReferenceImpl.java:203)
    at de.ialistannen.javadocapi.spoon.JavadocElementExtractor.getPossiblyGenericType(JavadocElementExtractor.java:353)
    at de.ialistannen.javadocapi.spoon.JavadocElementExtractor.handleExecutable(JavadocElementExtractor.java:168)
    at de.ialistannen.javadocapi.spoon.JavadocElementExtractor.visitCtMethod(JavadocElementExtractor.java:129)
    at spoon.support.reflect.declaration.CtMethodImpl.accept(CtMethodImpl.java:58)
    at de.ialistannen.javadocapi.indexing.Indexer.lambda$main$0(Indexer.java:74)
    at de.ialistannen.javadocapi.spoon.filtering.ParallelProcessor.lambda$process$1(ParallelProcessor.java:49)
    at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
    at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1130)
    at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:630)
    at java.base/java.lang.Thread.run(Thread.java:831)
```

```
Removing boolean
20:31:37.316 [Thread-3] WARN spoon.Launcher - cannot create shadow class: boolean
spoon.reflect.declaration.ParentNotInitializedException: parent not initialized for boolean(class spoon.support.reflect.declaration.CtClassImpl) (unknown file)
	at spoon.support.reflect.declaration.CtElementImpl.getParent(CtElementImpl.java:363)
	at spoon.support.visitor.java.JavaReflectionTreeBuilder.scan(JavaReflectionTreeBuilder.java:116)
	at spoon.reflect.factory.TypeFactory.get(TypeFactory.java:564)
	at spoon.support.reflect.reference.CtTypeReferenceImpl.getTypeDeclaration(CtTypeReferenceImpl.java:203)
	at de.ialistannen.javadocapi.spoon.JavadocElementExtractor.getPossiblyGenericType(JavadocElementExtractor.java:353)
	at de.ialistannen.javadocapi.spoon.JavadocElementExtractor.lambda$handleExecutable$0(JavadocElementExtractor.java:155)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1625)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:913)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:682)
	at de.ialistannen.javadocapi.spoon.JavadocElementExtractor.handleExecutable(JavadocElementExtractor.java:158)
	at de.ialistannen.javadocapi.spoon.JavadocElementExtractor.visitCtMethod(JavadocElementExtractor.java:129)
	at spoon.support.reflect.declaration.CtMethodImpl.accept(CtMethodImpl.java:58)
	at de.ialistannen.javadocapi.indexing.Indexer.lambda$main$0(Indexer.java:74)
	at de.ialistannen.javadocapi.spoon.filtering.ParallelProcessor.lambda$process$1(ParallelProcessor.java:49)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1130)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:630)
	at java.base/java.lang.Thread.run(Thread.java:831)
Error in parallel worker
java.lang.NullPointerException: Cannot invoke "java.lang.Package.getName()" because the return value of "java.lang.Class.getPackage()" is null
	at spoon.reflect.factory.TypeFactory.get(TypeFactory.java:571)
	at spoon.support.reflect.reference.CtTypeReferenceImpl.getTypeDeclaration(CtTypeReferenceImpl.java:203)
	at de.ialistannen.javadocapi.spoon.JavadocElementExtractor.getPossiblyGenericType(JavadocElementExtractor.java:353)
	at de.ialistannen.javadocapi.spoon.JavadocElementExtractor.lambda$handleExecutable$0(JavadocElementExtractor.java:155)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1625)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:913)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:682)
	at de.ialistannen.javadocapi.spoon.JavadocElementExtractor.handleExecutable(JavadocElementExtractor.java:158)
	at de.ialistannen.javadocapi.spoon.JavadocElementExtractor.visitCtMethod(JavadocElementExtractor.java:129)
	at spoon.support.reflect.declaration.CtMethodImpl.accept(CtMethodImpl.java:58)
	at de.ialistannen.javadocapi.indexing.Indexer.lambda$main$0(Indexer.java:74)
	at de.ialistannen.javadocapi.spoon.filtering.ParallelProcessor.lambda$process$1(ParallelProcessor.java:49)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1130)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:630)
	at java.base/java.lang.Thread.run(Thread.java:831)
```

I-Al-Istannen:
Okay, I restructured the ModelSet and I want to give up. My new ModelSet is actually a Map (as the Set was used as a very inefficient Map anyways) and everything works just fine.

But sadly some tests just blindly call `type.setSimpleName` which causes the map to go out of sync. I am pretty sure this was already a problem in the previous implementation, as you silently changed the `compareTo` value of an Element in a TreeSet. This might cause the element to "disappear" from the TreeSet as it is no longer in a place the `TreeSet` expects it to be. This case was probably never observed as the code just straight up ignores the Set-iness and performs linear searches over the values of the Set.

IMHO the current behavior of silently adjusting the simple name is broken and only works in Spoon through sheer luck (or great foresight, though the whole ModelSet violates the contract of *every single data structure* it uses).

We could kinda save this by having the CtType check its parent package and remove and re-add itself when its simple name changes. Alternatively the package could subscribe to simple name changes in its types, but that doesn't seem to be possible with spoons listener model.

slarse:
I'll have a look at this and see if I can figure something out.

I-Al-Istannen:
As a small aside: My PR branch speeds up indexing JDK 16 by around a minute if I didn't screw anything up while measuring.

I assume using a Map with `O(1)` lookup for classes and packages vs the `O(n)` iteration the `ModelSet` did before does help with larger codebases.

slarse:
I'll have a close look at this as soon as I can, but right now it's failing two tests:

1. Documentation test, because some docs are out of date. To resolve this one just needs to run the test, let it fail, and commit the generated docs.
2. Generics test. This fails because the iteration order of types within a package has changed (expects to find `paramB` variable from `CelebrationLunch` type first, finds it from `Lunch` type instead). This is caused by returning hash sets everywhere where previously ordered tree sets would be returned.

The second point is a bit of a sore thumb. If possible, we don't want to change the iteration order of types/packages, but at the same time there's potentially quite a bit of overhead to sorting them for each "get" from the public API.

I-Al-Istannen:
> Documentation test, because some docs are out of date. To resolve this one just needs to run the test, let it fail, and commit the generated docs.

Actually it was because the order of the elements changed as it relied on the order spoon returned its elements.

> Generics test. This fails because the iteration order of types within a package has changed (expects to find paramB variable from CelebrationLunch type first, finds it from Lunch type instead). This is caused by returning hash sets everywhere where previously ordered tree sets would be returned.

Thanks :)

----

All tests pass and indexing now produces the same output as before :tada:

So this seems to be feasible in theory, though the parent package hack isn't great. I will need to have a closer look at performance of the ConcurrentSkipListMap later on.

I-Al-Istannen:
I am pretty sure the failure in Sorald is a bug (maybe because it doesn't understand intersection types):
```java
public abstract class KeyedModelSet<K extends Comparable<K> & Serializable, T extends CtElement>
		implements Serializable {

	// This can not be a "Map" as this class is Serializable and therefore Sorald wants this field
	// to be serializable as well (Rule 1948).
	// It doesn't seem smart enough to realize it is final and only assigned to a Serializable Map
	// in the constructor.
	private final ConcurrentSkipListMap<K, T> map;

	protected KeyedModelSet() {
		this.map = new ConcurrentSkipListMap<>();
	}
```

`ConcurrentSkipListMap` is Serializable, `T` is Serializable as `CtElement` is Serializable and `K` is Serializable because the `K` generic type is constrained with the intersection type `K extends Comparable<K> & Serializable`.

I don't see anything non-Serializable left that could not be a problem but maybe I am missing something. I could probably simplify the generic bound to `String` and omit it entirely as CtPackageImpl is the only caller and it is uses `String`s. That might be a nice enough solution.

I-Al-Istannen:
I spent too long getting Sonar-Java to compile, but the issue is fixed in `6.10.0.24201` via - I guess - https://jira.sonarsource.com/browse/SONARJAVA-3571.

Updating Sorald to use sonar-java `6.10.0.24201` should turn the last test for this PR green.

slarse:
@I-Al-Istannen We're working on updating Sonar-Java to v7+ in Sorald, but there are some complications.

I will disable sorald-buildbreaker for the moment because Sonar-Java 6.9 is just giving off way too many false positives (especially, for some reason, on dead stores).

slarse:
@I-Al-Istannen Is this ready for review?

I-Al-Istannen:
I think it is missing `review` because I wasn't sure the parent hacks can be done in a nicer way. It also has no test because this is a bit hard to nicely test - it only occurs when you throw enough cores on a large enough codebase.

It is ready enough for you to have a look at it, I am just not sure you want to merge it. But you can see the last part for yourself :)

slarse:
@I-Al-Istannen Is it correct that this PR is intended to fix a problem when you are mutating a model in parallel? I'm not sure that's actually a supported use case, @monperrus?

I-Al-Istannen:
Well, *I* am not mutating anything. My code runs after spoon has built its model and only queries it. Spoon does that all by itself while resolving queries.

If you want I can have a deeper look at *why* it does that but my guess would be resolving shadow classes.

slarse:
Ah, then I understand.

I'm a bit hesitant to put in synchronization in query logic, as that can severely hamper multi-threaded performance. Queries that mutate the AST are... problematic, so perhaps that can somehow be improved (although I suppose that's part of lazy initialization here, which is probably also desirable).

Could you share the code that (sometimes) reproduces the problem for you?

I-Al-Istannen:
It's in https://github.com/I-Al-Istannen/JavadocApi but it requires a frankenstein spoon version built up of some of my open PRs to actually run to completion (and even then I just swallow some exceptions because I don't want to fix the underlying Spoon bugs at the moment).

I will have a go at reducing the scope of my code tomorrow, I doubt you want to wade through the spoon part of my library. That part isn't pleasant to read as it works around *quite* a few weird spoon limitations. I gave up at some point and just started piling workarounds atop each other.

I-Al-Istannen:
As an aside: Note that the original implementation used the result of `Collections.synchronizedXX` (albeit incorrectly). My new implementation also converted the O(n) member lookups to O(1) and replaced the synchronized collection with a lockfree concurrent collection.

I would not expect this to degrade performance and it does at least not do that severely for me. I run my indexer which just queries Spoon on all cores with these changes applied.

Measurements with multiple threads are hard as the current solutions errors out but I will try to get some measurements with multiple threads tomorrow. The single threaded performance seems to be at least not spectacularly worse (master -> the last commit in this PR):
grafik

monperrus:
> when you are mutating a model in parallel? I'm not sure that's actually a supported use case, @monperrus?

There is no official support but there has been quite some work on parallelization, see https://github.com/INRIA/spoon/issues?q=is%3Aissue+sort%3Aupdated-desc+parallel, incl. closed roadmap item https://github.com/INRIA/spoon/issues/2659

I-Al-Istannen:
### Repro
Call the following code with `/usr/lib/jvm/java-16-openjdk/lib/src.zip` or wherever you stashed away the JDK source code. You can reproduce it with smaller codebases IIRC but that is a bit more hit and miss.
<details>
<summary>Code to reproduce</summary>

```java
package de.ialistannen.javadocapi;

import java.io.File;
import java.util.ArrayList;
import java.util.List;
import java.util.concurrent.BrokenBarrierException;
import java.util.concurrent.Callable;
import java.util.concurrent.CyclicBarrier;
import java.util.concurrent.ExecutorService;
import java.util.concurrent.Executors;
import java.util.concurrent.LinkedBlockingDeque;
import java.util.concurrent.TimeUnit;
import java.util.concurrent.TimeoutException;
import java.util.function.Consumer;
import spoon.Launcher;
import spoon.OutputType;
import spoon.reflect.CtModel;
import spoon.reflect.declaration.CtConstructor;
import spoon.reflect.declaration.CtField;
import spoon.reflect.declaration.CtModule;
import spoon.reflect.declaration.CtPackage;
import spoon.reflect.declaration.CtParameter;
import spoon.reflect.visitor.CtAbstractVisitor;
import spoon.support.compiler.ProgressLogger;
import spoon.support.compiler.ZipFolder;

public class Foo {

  public static void main(String[] args) throws Exception {
    ParallelProcessor processor = new ParallelProcessor(
        Runtime.getRuntime().availableProcessors()
    );

    launcher.getEnvironment().setShouldCompile(false);
    launcher.getEnvironment().disableConsistencyChecks();
    launcher.getEnvironment().setOutputType(OutputType.NO_OUTPUT);
    launcher.getEnvironment().setSpoonProgress(new ConsoleProcessLogger(launcher));
    launcher.getEnvironment().setCommentEnabled(false);
    launcher.getEnvironment().setComplianceLevel(15);
    if (args[0].endsWith(".zip")) {
      launcher.addInputResource(new ZipFolder(new File(args[0])));
    } else {
      launcher.addInputResource(args[0]);
    }

    Visitor visitor = new Visitor();
    for (CtModule module : model.getAllModules()) {
      processor.process(module.getRootPackage(), ctElement -> ctElement.accept(visitor));
    }

    processor.shutdown();
  }

  private static class Visitor extends CtAbstractVisitor {

    @Override
    public <T> void visitCtMethod(CtMethod<T> m) {
      CtType<T> declaration = m.getType().getTypeDeclaration();
      if (declaration != null && declaration.getQualifiedName().equals("foo")) {
        System.out.println("Please dear JIT, let my code survive");
      }
      for (CtParameter<?> parameter : m.getParameters()) {
        CtType<?> typeDeclaration = parameter.getType().getTypeDeclaration();
        if (typeDeclaration != null && typeDeclaration.getQualifiedName().equals("Bar")) {
          System.out.println("Please dear JIT, let my code survive");
        }
      }
    }

    @Override
    public <T> void visitCtConstructor(CtConstructor<T> c) {
      if (c != null && c.getType().getTypeDeclaration().getQualifiedName().equals("foo")) {
        System.out.println("Please dear JIT, let my code survive");
      }
      for (CtParameter<?> parameter : c.getParameters()) {
        CtType<?> typeDeclaration = parameter.getType().getTypeDeclaration();
        if (typeDeclaration != null && typeDeclaration.getQualifiedName().equals("Bar")) {
          System.out.println("Please dear JIT, let my code survive");
        }
      }
    }
  }

  public static class ParallelProcessor {

    private final ExecutorService pool;
    private final int parallelism;

    public ParallelProcessor(int parallelism) {
      this.parallelism = parallelism;

      this.pool = Executors.newFixedThreadPool(parallelism, r -> {
        Thread thread = new Thread(r);
        thread.setDaemon(true);
        return thread;
      });
    }

    public void process(CtElement root, Consumer<CtElement> consumer) {
      CyclicBarrier barrier = new CyclicBarrier(parallelism);
      LinkedBlockingDeque<CtElement> workQueue = new LinkedBlockingDeque<>();
      List<Callable<Void>> tasks = new ArrayList<>();

      workQueue.add(root);

      for (int i = 0; i < parallelism; i++) {
        tasks.add(() -> {
          while (true) {
            CtElement item;
            while ((item = workQueue.pollFirst()) != null) {
              if (!(item instanceof CtType) && !(item instanceof CtField)
                  && !(item instanceof CtMethod) && !(item instanceof CtConstructor)
                  && !(item instanceof CtPackage)
                  && !(item instanceof CtModule)) {
                continue;
              }
              try {
                consumer.accept(item);
              } catch (Exception e) {
                System.err.println("Error in parallel worker");
                e.printStackTrace();
              }
              workQueue.addAll(item.getDirectChildren());
            }

            try {
              barrier.await(100, TimeUnit.MILLISECONDS);
              return null;
            } catch (TimeoutException e) {
              barrier.reset();
            } catch (BrokenBarrierException ignored) {
            }
          }
        });
      }

      try {
        pool.invokeAll(tasks);
      } catch (InterruptedException e) {
        throw new RuntimeException(e);
      }
    }

    public void shutdown() {
      pool.shutdown();
    }

  }

  private static class ConsoleProcessLogger extends ProgressLogger {

    public int touchedClasses;

    public ConsoleProcessLogger(Launcher launcher) {
      super(launcher.getEnvironment());
      touchedClasses = 0;
    }

    @Override
    public void start(Process process) {
      System.out.println("Starting phase " + process);
      touchedClasses = 0;
    }

    @Override
    public void step(Process process, String task, int taskId, int nbTask) {
      touchedClasses++;
      if (touchedClasses % 1000 == 0) {
        System.out.println(
            "Phase " + process + " has discovered " + touchedClasses
                + " classes so far. Currently working on " + task
        );
      }
    }

    @Override
    public void end(Process process) {
      System.out.println("Phase " + process + " done! Discovered Classes: " + touchedClasses);
    }
  }

}
```
</details>


### Performance

I also benchmarked it and it seems be be on par in single-core mode. I consistently got crashes without this patch when using more than one thread so I do not have any real measurements for multi-core operation. The best I have is
```
/usr/lib/jvm/java-17-adoptopenjdk/bin/java -XX:ActiveProcessorCount=4 -Xmx8G   1019,58s user 6,69s system 372% cpu 4:35,87 total
  - with error

/usr/lib/jvm/java-17-adoptopenjdk/bin/java -XX:ActiveProcessorCount=4 -Xmx8G   1166,92s user 5,88s system 386% cpu 5:03,14 total
  - without error
```
That doesn't really help much though, as an error on a higher-level element will prune the tree below it or at least cause a lot less work in javadoc and method lookups and related things.

I then benchmarked indexing JavaFx which thankfully ran to completion without errors and resulted in:
```
/usr/lib/jvm/java-17-adoptopenjdk/bin/java -XX:ActiveProcessorCount=4 -Xmx8G   520,87s user 2,26s system 374% cpu 2:19,65 total
  - Master

/usr/lib/jvm/java-17-adoptopenjdk/bin/java -XX:ActiveProcessorCount=4 -Xmx8G   535,43s user 1,70s system 377% cpu 2:22,46 total
  - PR

```

src/main/java/spoon/support/reflect/declaration/CtNamedElementImpl.java
```diff
@@ -34,21 +36,35 @@ public String getSimpleName() {
 
 	@Override
 	public <T extends CtNamedElement> T setSimpleName(String simpleName) {
+		String oldName = this.simpleName;
+
 		Factory factory = getFactory();
 		if (factory == null) {
 			this.simpleName = simpleName;
+			updateParentPackage(oldName);
```
slarse, line 44: I don't like the idea of updating the parent package when setting the simple name, but I also can't really see a better solution. I've pondered it for a good 20 minutes now, and I can't figure something out that doesn't end up violating the contract of the set (or keys of the map, same thing) referencing the type/package. Mutability is terrible, isn't it? With the current implementation in Spoon, you can for example get two different types with the same name in a single package, which is all kinds of crazy.

Anyway, this is not the concern of `CtNamedElement`, but rather should be implemented in `CtTypeImpl` and `CtPackageImpl`. When you need to check the type of `this`, you're probably up to no good :)

Finally, there is the concern of what should happen if you change the name of a type/package such that it collides with another one in the same package. The current behavior appears to just clobber in case of a collision. This has been discussed before, and we found that an exception would be best: #3837

src/main/java/spoon/support/reflect/declaration/CtPackageImpl.java
```diff
@@ -236,4 +237,12 @@ public boolean hasPackageInfo() {
...
 		return getPackages().isEmpty() && getTypes().isEmpty();
 	}
+
+	public void updateName(CtType<?> newType, String oldName) {
+		types.updateKey(oldName, newType.getSimpleName());
+	}
+
+	public void updateName(CtPackage newPackage, String oldName) {
+		packs.updateKey(oldName, newPackage.getSimpleName());
+	}
```
slarse, line 247: These can both be package private.

Please name them more precisely as well: `updateTypeName` and `updatePackageName`. I'm of the opinion that overloads are generally not good for readability and discoverability.

src/main/java/spoon/support/util/KeyedModelSet.java
```diff
@@ -7,135 +7,136 @@
...
 	}
 
-	@Override
 	public Iterator<T> iterator() {
 		return new Itr();
 	}
 
+	public void updateKey(K oldName, K newName) {
+		T type = map.remove(oldName);
+		if (type != null) {
```
slarse, line 130: Updating a key that doesn't exist seems like an error to me.

slarse (changes requested):
Hi @I-Al-Istannen,

I'm convinced by your performance analysis.

Could finally set aside some time to look this over properly. In general, I don't like the idea of having the "update containing package" side effect on setting simple names, but I can't come up with a better idea. I don't really think there is one.

I have some comments on potential improvements on the implementation, but overall I think your approach here is the only sensible one.

src/main/java/spoon/support/util/KeyedModelSet.java
I-Al-Istannen: We are in a bit of a pickle here, as many elements are added *without a name*: The parent is set before the simple name while constructing the model...

src/main/java/spoon/support/reflect/declaration/CtNamedElementImpl.java
I-Al-Istannen: And this vomits errors all over the test suite because apparently this is done in multiple places. Another question is how this should interact with `setIgnoreDuplicateDeclarations`:
```java
	/**
	 * Set Spoon to ignore duplicate type declarations in a project.
	 *
	 * Setting this option to {@code true} causes Spoon to attempt to build a model even when
	 * the same qualified name appears for multiple types. This mode of operation makes Spoon less
	 * stable as duplicated types do not make sense in Java, and causes strange behavior in the
	 * underlying JDT compiler. Type resolution can become unpredictable as the order in which types
	 * are parsed becomes a determining factor in which duplicated type actually makes it into the
	 * model.
     *
	 * @param ignoreDuplicateDeclarations (default false)  set to true to allow spoon to create a model of a project that
	 *                                 contains multiple times the same class
	 */
	void setIgnoreDuplicateDeclarations(boolean ignoreDuplicateDeclarations);
```

slarse: Just to be clear, this changes behavior. Previously, if you changed the simple name of a type such that it collided with another in the same package, they would _both_ be in the set (and thus violating the contract of that set).

With your implementation here, one overwrites the other.

Anyway, since there turned out to be problems with the stricter enforcement, let's leave that for a separate PR. But, do document the current "overwriting" behavior with a test somehow. Show that if you rename such that there is a clash, the renamed element is the one that remains in the package. Understand what I mean?

I-Al-Istannen: I added a test: https://github.com/INRIA/spoon/pull/4076/commits/b2a2ca77b60f1e9bb72d3a873b522c57c5ae4afa

slarse: I was thinking of high-level tests on CtType.setSimpleName() and CtPackage.setSimpleName(). I.e. given a package with types A and B, calling B.setSImpleName("A") should result in the "original" A being removed from the package. Which, if I'm not misreading your code, is what happens with your implementation?

I-Al-Istannen: It doesn't change the simple name, it only validates the add method as that contained the logic the other tests depend on. I will add some tests that explicitly check the rename as well, that was another aspect of this implementation and not quite what we talked about earlier in this thread.

slarse: To be super clear about what I mean, try this code snippet on master and with this PR:

```java

CtType<?> bestClass = factory.createClass("BestClass");
CtType<?> otherCLass = factory.createClass("OtherClass");

System.out.println(model.getRootPackage().getTypes());

bestClass.setSimpleName(otherClass.getSimpleName());

System.out.println(model.getRootPackage().getTypes());
```

I-Al-Istannen:
> @I-Al-Istannen I'm happy with the changes now. And it only took two months (thanks for sticking with it :) ).

;)

> The PR title doesn't really explain what the PR does anymore (we fix nothing with ModelSet). For a merge commit message, would you agree that fix: Fix ConcurrentModificationException on parallel package/type queries is an adequate summary?

Sounds good to me :)

PR #4082, opened 2021-08-09
review: test: add-test-for-CtCase

Rohitesh-Kumar-Jain:
Hi, I have added new tests for the class `CtCase`, I had written them much earlier but pushing them now.

These tests are intended to kill these mutations

1)
Screenshot 2021-08-09 at 5 31 23 PM

2)
Screenshot 2021-08-09 at 5 31 54 PM

3)
Screenshot 2021-08-09 at 5 32 11 PM

4)
Screenshot 2021-08-09 at 5 32 34 PM

Link to #3967

PR #4089, opened 2021-08-11
review: test: add tests for CtStatementList

Rohitesh-Kumar-Jain:
Hi, I have added new tests for the class `CtStatementList`, I had written them earlier but to avoid working on conflicting branches I am pushing them now, this PR is similar to #4082, so I will automatically implement changes based on reviews on that PR.

These tests are intended to kill these mutations

1)

Screenshot 2021-08-11 at 9 58 10 AM

2)

Screenshot 2021-08-11 at 9 58 25 AM

3)
Screenshot 2021-08-11 at 9 58 38 AM

4)
Screenshot 2021-08-11 at 9 58 56 AM

Link to #3967

### DIFF
--- a/src/test/java/spoon/test/ctStatementList/CtStatementListTest.java
+++ b/src/test/java/spoon/test/ctStatementList/CtStatementListTest.java
@@ -1,3 +1,19 @@
+/**
+ * Copyright (C) 2006-2021 INRIA and contributors
+ * Spoon - http://spoon.gforge.inria.fr/
+ *
+ * This software is governed by the CeCILL-C License under French law and
+ * abiding by the rules of distribution of free software. You can use, modify
+ * and/or redistribute the software under the terms of the CeCILL-C license as
+ * circulated by CEA, CNRS and INRIA at http://www.cecill.info.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the CeCILL-C License for more details.
+ *
+ * The fact that you are presently reading this means that you have had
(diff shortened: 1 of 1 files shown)


src/test/java/spoon/test/ctStatementList/CtStatementListTest.java
```diff
@@ -1,35 +1,52 @@
...
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.CoreMatchers.is;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
 public class CtStatementListTest {
+    private static Factory factory;
+
+    private static CtStatementList getStatementListInitializedWithOneStatement() {
+        factory = new Launcher().getFactory();
```
slarse, line 34: Setting the `factory` field is a very unexpected side-effect of this method. I would in general avoid mutable fields in test classes (or any classes, for that matter) when possible. If a new test is added that uses this field without first calling the associated method, there could be an NPE if that test is executed first. Tests that depend on the order of test execution is a big nono.

If you really want this as a field, then set it up in an `@BeforeEach` method, but make it an instance field and not static. I'd personally avoid even that, though, and just create the factory in each test that needs it.

Just create the factory in each test instead.

```suggestion
    private static CtStatementList getStatementListInitializedWithOneStatement() {
        Factory factory = new Launcher().getFactory();
```

slarse (changes requested):
See comment

src/test/java/spoon/test/ctStatementList/CtStatementListTest.java
Rohitesh-Kumar-Jain: > Just create the factory in each test instead.

Thanks, I will do the same

slarse (approved):
LGTM, thanks @Rohitesh-Kumar-Jain!

PR #4091, opened 2021-08-12
review: test: add-tests-for-ConstructorCall

Rohitesh-Kumar-Jain:
Hi, I have added new tests for the interface `CtConstructorCall`

These tests are intended to kill these two mutations

1)

Screenshot 2021-08-12 at 5 52 51 PM

2)

Screenshot 2021-08-12 at 5 53 04 PM

Link to #3967


src/test/java/spoon/test/constructor/ConstructorCallTest.java
```diff
@@ -0,0 +1,66 @@
...
+
+public class ConstructorCallTest {
+
+    @Test
+    void testRemoveArgument() {
+        // contract: removeArgument removes the passed argument among two arguments in the call
+
+        Factory factory = new Launcher().getFactory();
+        CtExpression<?> argumentToBeRemoved = factory.Code().createCodeSnippetExpression("true").compile();
+        CtExpression<?> additionalArgument = factory.Code().createCodeSnippetExpression("false").compile();
```
slarse, line 37: Can be simplified

```suggestion
        CtExpression<?> argumentToBeRemoved = factory.createLiteral(true);
        CtExpression<?> additionalArgument = factory.createLiteral(false);
```

src/test/java/spoon/test/constructor/ConstructorCallTest.java
```diff
@@ -0,0 +1,66 @@
...
+        CtConstructorCall<?> call = factory.Code().createConstructorCall(
```
slarse, line 38: This code is not wrong in any way, but I wanted you to know that most of these methods are available on the core `Factory` object, and you don't need to access the actual subfactories.

```suggestion
        CtConstructorCall<?> call = factory.createConstructorCall(
```

Mentioning this as I've seen this a couple of times now in your recent PRs (where I've not bothered mentioning it because it's not a big deal).

src/test/java/spoon/test/constructor/ConstructorCallTest.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    @Test
+    void testRemoveActualTypeArgument() {
+        // contract: removeActualTypeArgument removes the passed type among two types
+
+        Factory factory = new Launcher().getFactory();
+        CtTypeReference<Boolean> typeToBeRemoved = factory.createCtTypeReference(Boolean.class);
+        CtTypeReference<Integer> additionalType = factory.createCtTypeReference(Integer.class);
+        CtConstructorCall<?> call = factory.Code().createConstructorCall(factory.createCtTypeReference(Object.class));
```
slarse, line 56: Same as above.

```suggestion
        CtConstructorCall<?> call = factory.createConstructorCall(factory.createCtTypeReference(Object.class));
```

slarse (changes requested):
I've suggested a few minor simplifications, otherwise LGTM!

src/test/java/spoon/test/constructor/ConstructorCallTest.java
```diff
@@ -0,0 +1,66 @@
...
+import spoon.Launcher;
+import spoon.reflect.code.CtConstructorCall;
+import spoon.reflect.code.CtExpression;
+import spoon.reflect.factory.Factory;
+import spoon.reflect.reference.CtTypeReference;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+public class ConstructorCallTest {
```
slarse, line 29: Sorry, I missed something. It's a bit inconsistent throughout the test suite, but I think we should always name the test classes precisely after the tested interfaces.

```suggestion
public class CtConstructorCallTest {
```

It just makes it easier to find the tests IMO, wdyt?

slarse (changes requested):
Missed something, see comment.

src/test/java/spoon/test/constructor/ConstructorCallTest.java
Rohitesh-Kumar-Jain: yes, it makes it easier to find tests. It's probably the naming convention as well I believe ?

slarse:
@Rohitesh-Kumar-Jain Don't forget to also rename the file :)

Rohitesh-Kumar-Jain:
> @Rohitesh-Kumar-Jain Don't forget to also rename the file :)

yes I have seen that failure, will fix that as well : )

slarse (approved):
LGTM, thanks @Rohitesh-Kumar-Jain

slarse:
@monperrus FYI I somehow failed to remove the `review:` tag from the squashed commit message. Oops. Just so you know there might be some strangeness in the generated change log.

PR #4093, opened 2021-08-12
fix: Make lambda type implicit to avoid auto-importing type targeted in lambda

### LINKED ISSUE #4079: Functional interfaces targeted in lambdas are imported when using auto-import printer
To me, that sounds like imports are added for functional interfaces even when they are used as lambda expression (and likely method references too?). Does that match your processed/printed code?

_Originally posted by @SirYwell in https://github.com/INRIA/spoon/issues/4070#issuecomment-893328459_

I've verified this behavior. When a functional interface is used as targeted in a lambda, it's imported by the auto-importer, even if it's never explicitly referenced. Here's a minimal example to reproduce the problem:

```java
// TargetsFunctionalInterface.java
public class TargetsFunctionalInterface {
        HasFunctionalInterface.supplyOne(i -> System.out.println(i));
    }
}

// HasFunctionalInterface.java
import java.util.function.IntFunction;

public class HasFunctionalInterface {
    public static void supplyOne(IntFunction intFunction) {
        intFunction.apply(1);
    }
}
```

When printing these files again, the line `import java.util.function.IntFunction;` is added to the top of `TargetsFunctionalInterface.java`. The **expected result** is of course that this does not happen.

Note that both `TargetsFunctionalInterface.java` _and_ `HasFunctionalInterface.java` must be added as resources. If only `HasFunctionalInterface.java` is not added, the `IntFunction` reference can't be resolved by Spoon, and consequently the import isn't added.

slarse:
Fix #4079 

See #4079 for a description of the problem.

### DIFF
--- a/src/main/java/spoon/support/reflect/code/CtLambdaImpl.java
+++ b/src/main/java/spoon/support/reflect/code/CtLambdaImpl.java
@@ -288,7 +288,9 @@ public class CtLambdaImpl<T> extends CtExpressionImpl<T> implements CtLambda<T> 
 
 	@Override
 	public <C extends CtTypedElement> C setType(CtTypeReference<T> type) {
-		type.setImplicit(true);
+		if (type != null) {
+			type.setImplicit(true);
+		}
 		return super.setType(type);
 	}
 }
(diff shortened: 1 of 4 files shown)


monperrus:
Clean and green, LGTM.

slarse:
@monperrus let's merge then, seems like a reasonable enough solution. I've double-checked and there's no way to explicitly specify return type of a lambda expression (which makes sense as lambdas can only target functional interfaces, where the type is already known).

monperrus:
Thanks for the fix @slarse

PR #4097, opened 2021-08-14
review: refactor: Simplify Set creation in CtReferenceImpl

I-Al-Istannen:
This PR just replaces the stream set creations with `Collections#singleton` - the `Set.of` for poor sods still using java 8.

I don't expect this to have any impact on correctness or performance.


slarse (approved):
LGTM.

> the Set.of for poor sods still using java 8.

:(

PR #4099, opened 2021-08-16
review: fix: Fix return type of setParent

### LINKED ISSUE #4098: CtElement.setParent's return type is incorrect, or all implementations are incorrect
`CtElement.setParent` has a return type that matches that of the input (i.e.

https://github.com/INRIA/spoon/blob/140daab011c9f0c4bd5a78468ff3dfc69c4548fd/src/main/java/spoon/reflect/declaration/CtElement.java#L264-L270

However, all implementations return the element itself, i.e. are implemented like so:

```
return (E) this;
```

This is consistent with the rest of the methods of `CtElement`, but it is clearly an error as given `e.setParent(p)`, it is unlikely that `e` and `p` are of the same type.

There are two options here.

1. We can simply change the implementations to return the parent instead of the element itself. That would make the return type make sense.

2. We can alter the interface like so:

```java
	/**
	 * Manually sets the parent element of the current element.
	 *
	 * @param parent
	 * 		parent reference.
	 */
	<E extends CtElement, P extends CtElement> E setParent(P parent);
```

Both fixes are breaking. Option 1 is breaking in terms of _behavior_, while option 2 is breaking in terms of typing. Therefore, option 1 will not break any compilation but may break at runtime, while option 2 may break some compilation but will not break at runtime. I therefore prefer option 2.

slarse:
Fix #4098 (see issue for description of problem)

**Breaking change!**

This PR implements option 2 presented in #4098. It's a _purely statically_ breaking change, meaning that some compilation may be affected, but behavior is not. However, compilation is pretty unlikely to be affected, and it should really only be the case if the parameterization of the method is explicitly provided.

### DIFF
--- a/src/main/java/spoon/support/reflect/code/CtTypeAccessImpl.java
+++ b/src/main/java/spoon/support/reflect/code/CtTypeAccessImpl.java
@@ -45,7 +45,7 @@ public class CtTypeAccessImpl<A> extends CtExpressionImpl<Void> implements CtTyp
 
 	@Override
 	public CtTypeReference<Void> getType() {
-		return getFactory().Type().VOID_PRIMITIVE.clone().setParent(this);
+		return (CtTypeReference<Void>) getFactory().Type().VOID_PRIMITIVE.clone().<CtTypeAccess>setParent(this);
 	}
 
 	@Override
--- a/src/main/java/spoon/support/reflect/code/CtTypeAccessImpl.java
+++ b/src/main/java/spoon/support/reflect/code/CtTypeAccessImpl.java
@@ -45,7 +45,7 @@ public class CtTypeAccessImpl<A> extends CtExpressionImpl<Void> implements CtTyp
 
 	@Override
 	public CtTypeReference<Void> getType() {
-		return (CtTypeReference<Void>) getFactory().Type().VOID_PRIMITIVE.clone().<CtTypeAccess>setParent(this);
+		return getFactory().Type().VOID_PRIMITIVE.clone().setParent(this);
 	}
 
 	@Override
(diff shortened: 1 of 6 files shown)


src/main/java/spoon/support/reflect/code/CtTypeAccessImpl.java
```diff
@@ -45,7 +45,7 @@ public void accept(CtVisitor visitor) {
 
 	@Override
 	public CtTypeReference<Void> getType() {
-		return (CtTypeReference<Void>) getFactory().Type().VOID_PRIMITIVE.clone().<CtTypeAccess>setParent(this);
+		return getFactory().Type().VOID_PRIMITIVE.clone().setParent(this);
```
slarse, line 48: Case and point in small amount of breakage: this was the only part of the entire Spoon codebase that needed to change, apart from the directly affected method headers.

slarse:
@I-Al-Istannen The improved Javadoc check pinpointed the correct error: https://github.com/INRIA/spoon/runs/3337956190?check_suite_focus=true#step:9:25 :)

monperrus:
Thanks for the analysis and proof that it's minor.

What about this alternative `<E extends CtElement> E setParent(CtElement parent)`, not changing the return type and type parameters?

slarse:
Ah, yes, that should also work. Technically speaking the return type is changed, as it is no longer bound to the argument. This means that there could potentially still be static breakage, but as the return type was incorrect we probably want that breakage.

Oh, and note that this is precisely the super sketchy use of type parameters in return types that I advocated against in #1846. It can cause stuff like this, where incorrect typing simply isn't detected. But it's already there so we need to roll with it.

monperrus:
Simple and to the point, thanks. Will merge.

monperrus:
Thanks @slarse

slarse:
This caused a regression in Spork (and likely will for any Kotlin dependents).

For any call previously looking like so:

```java
element.setParent(parent);
```

Must now be amended like so:

```java
element.setParent<CtElement>(parent);
```

For the reasons outlined in #1846

Issue #4074, opened 2021-08-02
Local classes/enums/interfaces/records


SirYwell:
As Java 16 introduced local enums and local interfaces additionally to (local) records, I started looking into how they are handled in Spoon.

My first observation was that local interfaces are ignored but local enums are found and included in the model correctly. This seems to come from https://github.com/INRIA/spoon/blob/78d64fe348c282a190a353c17f8855684a0cb1b5/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java#L752 this check.

Only when writing tests to cover local types I saw that local classes are actually prefixed with numbers. Looking into that, I found its cause in the check here https://github.com/INRIA/spoon/blob/78d64fe348c282a190a353c17f8855684a0cb1b5/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java#L756 (the part after the `||`). 

In the same commit this check was introduced, this behavior was also documented in `CtType` (https://github.com/INRIA/spoon/commit/9e0de7530627784e04545fce947a461b4f7309a7#diff-493d9e2eaf0b87f5449ea2a10488544623189dcbad95afb9e68029e1d48cb753R39).
This causes several issues: 
- it differs from the output of [Class#getSimpleName()](https://docs.oracle.com/en/java/javase/16/docs/api/java.base/java/lang/Class.html#getSimpleName()) (note: "Returns the simple name of the underlying class **as given in the source code.**", highlighting added by me)
- the printer currently prints it wrong for enums (and for interfaces and records once supported)
- creating a new local class with spoon does not require this naming

My suggestion would be to just use the name which is used in the source code too, but that is clearly breaking the contract of the `getSimpleName()` method. Otherwise, the printer would need to be fixed in the appropriate places to strip numbers from interfaces and enums too.

To support local interfaces, `CtInterface` would also need to extend `CtStatement`.

I-Al-Istannen:
Additionally, the simple name returned by `Class#getSimpleName` for anonymous classes is `""` - spoon returns the number. The name of an anonymous class isn't as relevant (it is never printed out again), but being consistent with `getSimpleName` **might** be useful.

andrewbwogi:
The behavior that names are prefixed with numbers is inconsistent with the meaning of `getSimpleName`, which I understand is to get the element name as it is in the source code. That is why `setSimpleName` can undo it. 

> My suggestion would be to just use the name which is used in the source code too

[...]

PR #4108, opened 2021-08-19
review: feature: Add support for local interfaces

SirYwell:
In regards of #4074 and  #3653, this adds support for local interfaces and enums.

I also added tests for local classes as this doesn't seem to be covered pretty well for now (see below).

Enums are basically already supported, but they are printed wrong, as the leading number is not removed.

I also added the layout for a future test covering local records, this needs to be finished at some point. I disabled it therefore for now.

~~Note that the only test that succeeds currently is the `testLocalClassExists` one. This is wrong *if* the javadocs are right (see mentioned issue above) as there is no number prefixing the actual name. Looking into that, I found out that JDT reports the names somewhat inconsistently, `MyClass` has the constant pool name `X$1$MyClass` while `MyEnum` and `MyInterface` have the constant pool names `X$1MyEnum` and `X$1MyInterface` respectively...~~

Not sure if this makes it better or worse, but the constant pool name only differs if the compliance level is set to <=4. Are those lower levels relevant for spoon?

TODO:
- if the number prefix is kept in the simple name, the `DefaultJavaPrettyPrinter` needs to adapt that for local enums and local interfaces
- if the number prefix is removed, the code handling that in `DefaultJavaPrettyPrinter` can be removed

(I guess that should be tested too, independent of what we decide on)

Edit: The `CommentTest#testDocumentationContract()` fails now, should I add an example there?


src/main/java/spoon/support/reflect/declaration/CtInterfaceImpl.java
```diff
@@ -51,6 +56,41 @@ public boolean isInterface() {
 		return Collections.unmodifiableList(l);
 	}
 
+	@Override
+	public <R extends CtCodeElement> R partiallyEvaluate() {
+		return null;
+	}
```
raghav-deepsource, line 62: wouldn't this be better as an exception? Is there any use case where partially evaluating a local interface will make sense?

SirYwell: Good catch, seems like I missed that method. I'll implement it the same way `CtClassImpl` implements it.

raghav-deepsource: ahh, yup. I misunderstood how partial evaluation worked

slarse:
> Not sure if this makes it better or worse, but the constant pool name only differs if the compliance level is set to <=4. Are those lower levels relevant for spoon?

I don't think so, @monperrus?

On a side note, the names discussed here are the _binary names_ as specified in [JLS 13.1](https://docs.oracle.com/javase/specs/jls/se16/html/jls-13.html#jls-13.1).

> The CommentTest#testDocumentationContract() fails now, should I add an example there?

Any type that is a statement or expression must have a code snippet showing its use. I think the reason for only requiring this for statements is that Spoon supports on-the-fly compilation of these. Anyway, to fix the test error, you need to add an example interface to the doc and increase the compliance level in the test to 16. For example:

```java
/**
 * This element defines an interface declaration.
 *
 * <pre>
 *     // an interface definition
 *     interface Foo {
 *        void bar();
 *     }
 * </pre>
 */
```

SirYwell:
The printer now removes leading digits from local enums and local interfaces too, I also added a test case for that.

monperrus:
>> if the compliance level is set to <=4. Are those lower levels relevant for spoon?
> I don't think so, @monperrus?

I'd say it is relevant. AFAIK, Spoon is also used for analyzing old legacy code, incl. in very old Java versions.

SirYwell:
I wasn't able to find any proper specification for local class names for pre Java 5 versions, it's calculated [here](https://git.eclipse.org/c/jdt/eclipse.jdt.core.git/tree/org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/lookup/CompilationUnitScope.java#n292) in JDT. I tried to look into a few commits around that code but that wasn't very revealing for me.

Would it make sense to change the Javadocs to state that the simple name is the binary name in that case? Or should we strip out the last `$` if below Java 5? WDYT?

slarse:
> I'd say it is relevant. AFAIK, Spoon is also used for analyzing old legacy code, incl. in very old Java versions.

@monperrus It would be good to add at least one such project to the smoke tests on Jenkins. Know of any?

> I wasn't able to find any proper specification for local class names for pre Java 5 versions.

@SirYwell Java 5 and before is kind of a "water under the bridge" era. I've not found any specs for that either.

> Would it make sense to change the Javadocs to state that the simple name is the binary name in that case?

I think it does. Isn't that actually the current behavior for local classes?

SirYwell:
> Isn't that actually the current behavior for local classes?

It is, but the current comment explicitely states
```
	 * Following the compilation convention, if the type is a local type,
	 * the name starts with a numeric prefix (e.g. local class Foo has simple name 1Foo).
```
which is only correct for Java >=5 as it seems.

monperrus:
> @monperrus It would be good to add at least one such project to the smoke tests on Jenkins. Know of any?

Great idea. Any old commit of old projects, such as an old Java 1.4 commit of Apache Commons Lang or Apache Commons Math.

src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
```diff
@@ -742,7 +743,12 @@ public void visitCtDo(CtDo doLoop) {
 	@Override
 	public <T extends Enum<?>> void visitCtEnum(CtEnum<T> ctEnum) {
 		visitCtType(ctEnum);
-		printer.writeKeyword("enum").writeSpace().writeIdentifier(ctEnum.getSimpleName());
+		printer.writeKeyword("enum").writeSpace();
+		if (ctEnum.isLocalType()) {
+			printer.writeIdentifier(stripLeadingDigits(ctEnum.getSimpleName()));
+		} else {
+			printer.writeIdentifier(ctEnum.getSimpleName());
+		}
```
slarse, line 751: I just have one thought, and that's if we should simplify this like so:

```suggestion
		printer.writeIdentifier(stripLeadingDigits(ctEnum.getSimpleName()));
```

This should effectively provide the same behavior, as non-local types don't have leading digits. By that logic, we might even want to push down the stripping of leading digits into `DefaultTokenWriter.writeIdentifier`. We never want to print an identifier with leading digits, as that's not a valid Java identifier.

What do you think? The solution in here is not bad by any means, and is the preexisting one, but the repetition across methods just got me thinking.

slarse (commented):
I just have one pretty minor thing to discuss, rest LGTM.

src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
SirYwell: > but the repetition across methods just got me thinking

Yeah I thought about that too, especially as it's currently done in #4134 too - having it consistent in one place would make more sense. I'll look into it.

SirYwell: calling the method in the DefaultTokenWriter fails due to https://github.com/INRIA/spoon/blob/a03c436e4f3b892cebc12d0a1b5ddad83e8e5c7c/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java#L1883 writing the numbers of anonymous classes. As an example, the following test fails (the error is thrown in the stripLeadingDigits method, but the test would fail with an empty string too) and it's also expected to have a number there.

If that's the only case, I could just return the string itself instead of throwing the exception. This would make writing such stuff a bit slower but as it tends to be only a few chars (as long as you don't have hundreds of anonymous classes in one class) it shouldn't really matter.

slarse: Test case you linked there is a pretty strange one. The method is called `testAnonymousClassesHaveAnEmptyStringForItsNameInNoClasspath`, the contract probably by accident states that all references should be empty strings, and the test then goes on to verify that the simple name of an anonymous class is a digit. And a bunch of _other_ fully qualified names. So, the test IMO is overspecified and hardly does what it says, but the question still remains: what should the fully qualified name of an anonymous class be, if it does not contain that digit?

Let's be pragmatic: the beautiful solution caused complications, let's go with the less beautiful one of doing `printer.writeIdentifier(stripLeadingDigits(ctEnum.getSimpleName());` all over the place :)

SirYwell: > what should the fully qualified name of an anonymous class be, if it does not contain that digit?

As long as I didn't miss anything, there should be no case where printing the name of an anonymous type would be correct. So yeah, that's a good question.

I simplified the code now and it works with the tests that failed before, but it does not throw an exception anymore when writing anonymous type names as it seems to cause more headache than necessary I think.

slarse (changes requested):
Hi @SirYwell,

Implementation looks great. Now on my final pass review, I just have a little bit of a concern with the `LocalTypesTest` test class. I just feel like it might not be found when the next person comes in and has a problem with local, say, classes. Or enums.

I'd personally go looking in their dedicated test classes, `CtClassTest`, `EnumsTest` and `InterfaceTest`. What do you think about that? Do you agree that the tests would be more discoverable if they were in those, perhaps more expected test classes? You could put the helper methods in `SpoonTestHelpers`.

SirYwell:
> Do you agree that the tests would be more discoverable if they were in those, perhaps more expected test classes?

Yes, I agree with you. That said, those tests are currently written with JUnit 4. I already opened #4138 independently, I can probably also update the other tests too. Would that be preferred over mixing JUnit 4 and JUnit 5 (or writing the test for local types in JUnit 4)?

slarse:
Adding JUnit4 tests to an existing JUnit4 test class is fine, we don't want to make it too complicated to contribute stuff that isn't strictly test-related :). I'd say add the tests here in whichever version the test classes use, convert to JUnit5 later if you've time left over.

slarse (approved):
LGTM, thanks @SirYwell!

monperrus:
Super cool, thanks a lot.

PR #4114, opened 2021-08-22
review: doc: add new releases to README

h4iku:
Added new releases to README, so they are also shown on the homepage.


monperrus:
Thanks a lot. `8.4.0` was a CD bug (#4116), I removed it from the changelog.

Will merge.

Issue #4100, opened 2021-08-17
Need to expand AstCheckerTest to support more patterns for delegation/surcharging


slarse:
As evidenced by #4038, the AstCheckerTest is too conservative. I'll use #4038 as grounds for the following patterns we probably want to add support for.

First of all, the PushStackInIntercessionChecker needs a refactoring, it's pretty hard to understand what's expected e.g. for a method to count as delegated.

[`CtInterfaceImpl.removeNestedType`](https://github.com/INRIA/spoon/pull/4038/files#diff-aebac4f9ac007ac937bc14a49470399a1ce6edf5ed772bfbd0c187cc58d2f175R85-R105) should be detected as a delegated method, but it isn't as the method invocation isn't in any location expected by `PushStackInIntercessionChecker.isDelegate`, so that should be expanded. I'm thinking we might want to cover this pattern:

```java
// store return value of delegated-to method in a variable
@Override
public SomeType method() {
    SomeType val = super.method();
    [...]
    if (something) {
        // all returns must use val
        return val;
    }

    return val;
}
```

Now, the pattern used in this PR is more like this:

```java
// use delegated method in early return if condition
@Override
public boolean method() {
    if (!super.method()) {
        return false;
    }
    [...]

    return true;
}
```

But supporting that pattern is a little bit more difficult as it's harder to verify that the return value of the method is actually used. It is used implicitly here, and it's possible to verify, it's just some amount of work to do so, whereas verifying that all return values reference a particular local variable is trivial.

[`CtTypeImpl.removeNestedType`](https://github.com/INRIA/spoon/pull/4038/files#diff-c27f4bbf924c0556ebbba08b2f902851adc380122ba9fe15c1fcd8be50ce08c0R265-R281) has essentially the same problem, but for the surcharge instead of the delegate check. I think supporting the pattern I suggested above should cover this as well.

Obviously the pattern doesn't make for ideal code, but it's easy to verify and applicable to all kinds of return values.

slarse:
@I-Al-Istannen WDYT?

I-Al-Istannen:
> First of all, the PushStackInIntercessionChecker needs a refactoring, it's pretty hard to understand what's expected e.g. for a method to count as delegated.

[...]

PR #4119, opened 2021-08-23
test(AstCheckerTest): Refactor isSurcharged

slarse:
#4100

This is a refactor of `AstCheckerTest#isSurcharged()`. It should be noted that I don't understand what "surcharged" means in the context, and I also don't understand the intent of the function. The mechanics are however clear to me, see the invariants in 32de37d.

An invocation `v` is a potential surcharge delegate if it:

* Is an invocation that is the only statement in the body.
* Is an invocation that is the first statement in the body, and the last statement is a return _without_ an invocation.
* Is an invocation in a return occurring as the last statement in the body.

`v` is determined to be a surcharge delegate if it fulfills any of the 3 above points, and returns false for `isToBeProcessed(v)`.

### DIFF
--- a/src/test/java/spoon/reflect/ast/AstCheckerTest.java
+++ b/src/test/java/spoon/reflect/ast/AstCheckerTest.java
@@ -46,6 +46,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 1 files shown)


slarse:
I was going to keep going here but I think this refactoring will be sufficient.

Ping @monperrus for merge, I need to get on with the rest of this class so that I unblock @I-Al-Istannen's PRs #4038 which has been blocked for way too long.

monperrus:
ack, thanks for the test improvement!

PR #4121, opened 2021-08-24
chore: Manually set the renovate GHA tags

slarse:
As all but one of the PRs opened by renovate on GHA versions have _not_ pinned to a tag, I'm now manually pinning to tags [as suggested by the docs here](https://docs.renovatebot.com/modules/manager/github-actions/). The auto-pinning simply does not work, at least not reliably.


monperrus:
LGTM, will merge. Thanks @slarse

Issue #4033, opened 2021-07-07
CtExtendedModifiers are stored in a HashSet - leads to non-deterministic pretty-printing


slarse:
The `ElementPrinterHelper` prints out modifiers by fetching the extended modifiers. These are in turn stored inside a `HashSet`, which is not ordered. The printer-helper sections them in `[visibility] [abstract/static] [everything else]`, but there is no internal ordering in the `abstract/static` and `everything else` parts.

So, for example, a class with the modifiers `abstract static` could be written either `abstract static` or `static abstract` depending on the machine state. That's not desirable.

Either we store the modifiers in an ordered collection, or the printer helper must order them when printing. Non-deterministic printing is a hassle.

raghav-deepsource:
replace `HashSet` with `LinkedHashSet`?

monperrus:
@raghav-deepsource thanks for the suggestion, would you be able to make a tentative PR?

raghav-deepsource:
yup, was planning to do just that

raghav-deepsource:
or rather, CtModifierHandler?

I-Al-Istannen:
I would order them when printing (unless that breaks your sniper). Some elements do not offer `addModifier` or `addExtendedModifier` or `removeModifier` methods. This forces users to fetch the elements, store them in their own collection, mutate it and and use `setModifiers`. 

I personally wouldn't expect this to make printing non-deterministic, as all methods just take and return a Set and normal sets do not make any ordering guarantees.

raghav-deepsource:
The intent is to avoid worrying about ordering from the time modifiers are parsed to the point where they are printed out again. it would be better to alter the internal representation to account for this.

I-Al-Istannen:
The problem is that you can't do that by just using a LinkedHashSet as some elements miss remove/add methods and users are forced to temporarily store it in their own collections.

So using a LinkedHashSet would still lead to non-deterministic printing, if any API user ever touches the API and throws it in a HashSet (which most will do, it is the most common Set).
If you switch to a LinkedHashSet you also need to adjust every single caller of `setXXModifiers` to make sure it keeps the order, if there was one. A LinkedHashSet only treats one symptom but not really the root cause.

If you really want to keep the internal representation the same during the whole life I'd suggest a SortedSet (like TreeSet) and make sure you order the modifiers following conventions. You really want a distinct `List` there and a SortedSet comes pretty close.
[...]

PR #4122, opened 2021-08-24
wip: fix: change how modifiers are handled to use an ordered LinkedHashSet

raghav-deepsource:
I should probably move the linkedHashSetCollector method elsewhere (a utility class for example). 

this PR addresses #4033

### DIFF
--- a/src/main/java/spoon/support/reflect/CtModifierHandler.java
+++ b/src/main/java/spoon/support/reflect/CtModifierHandler.java
@@ -58,7 +58,7 @@ public class CtModifierHandler implements Serializable {
 	}
 
 	public Set<ModifierKind> getModifiers() {
-		return modifiers.stream().map(CtExtendedModifier::getKind).collect(linkedHashSetCollector());
+		return modifiers.stream().map(CtExtendedModifier::getKind).collect(Collectors.toCollection(LinkedHashSet::new));
 	}
 
 	public boolean hasModifier(ModifierKind modifier) {
--- a/src/main/java/spoon/support/reflect/CtModifierHandler.java
+++ b/src/main/java/spoon/support/reflect/CtModifierHandler.java
@@ -16,8 +16,9 @@ import spoon.support.reflect.declaration.CtElementImpl;
 import java.io.Serializable;
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.Set;
-import java.util.stream.Collectors;
+import java.util.stream.Collector;
 
 import static spoon.reflect.path.CtRole.MODIFIER;
 
@@ -44,7 +45,7 @@ public class CtModifierHandler implements Serializable {
 		if (extendedModifiers != null && !extendedModifiers.isEmpty()) {
 			getFactory().getEnvironment().getModelChangeListener().onSetDeleteAll(element, MODIFIER, this.modifiers, new HashSet<>(this.modifiers));
 			if (this.modifiers == CtElementImpl.<CtExtendedModifier>emptySet()) {
-				this.modifiers = new HashSet<>();
+				this.modifiers = new LinkedHashSet<>();
 			} else {
 				this.modifiers.clear();
 			}
@@ -57,7 +58,7 @@ public class CtModifierHandler implements Serializable {
 	}
 
 	public Set<ModifierKind> getModifiers() {
-		return modifiers.stream().map(CtExtendedModifier::getKind).collect(Collectors.toSet());
+		return modifiers.stream().map(CtExtendedModifier::getKind).collect(linkedHashSetCollector());
 	}
 
 	public boolean hasModifier(ModifierKind modifier) {
@@ -199,5 +200,20 @@ public class CtModifierHandler implements Serializable {
 			return false;
 		}
 		return getModifiers().containsAll(other.getModifiers());
+	}
+
+	private <T> Collector<T, LinkedHashSet<T>, LinkedHashSet<T>> linkedHashSetCollector() {
+		return Collector.of(
+				LinkedHashSet::new,
+				LinkedHashSet::add,
+				(left, right) -> {
+					if (left.size() < right.size()) {
+						right.addAll(left);
+						return right;
+					} else {
+						left.addAll(right);
+						return left;
+					}
+				});
 	}
 }
--- a/src/main/java/spoon/support/reflect/CtModifierHandler.java
+++ b/src/main/java/spoon/support/reflect/CtModifierHandler.java
@@ -19,6 +19,7 @@ import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.Set;
 import java.util.stream.Collector;
+import java.util.stream.Collectors;
 
 import static spoon.reflect.path.CtRole.MODIFIER;
 
@@ -67,7 +68,7 @@ public class CtModifierHandler implements Serializable {
 
 	public CtModifierHandler setModifiers(Set<ModifierKind> modifiers) {
 		if (modifiers == null) {
-			modifiers = Collections.emptySet();
+			modifiers = new LinkedHashSet<>();
 		}
 		getFactory().getEnvironment().getModelChangeListener().onSetDeleteAll(element, MODIFIER, this.modifiers, new HashSet<>(this.modifiers));
 		this.modifiers.clear();
@@ -79,7 +80,7 @@ public class CtModifierHandler implements Serializable {
 
 	public CtModifierHandler addModifier(ModifierKind modifier) {
 		if (this.modifiers == CtElementImpl.<CtExtendedModifier>emptySet()) {
-			this.modifiers = new HashSet<>();
+			this.modifiers = new LinkedHashSet<>();
 		}
 		getFactory().getEnvironment().getModelChangeListener().onSetAdd(element, MODIFIER, this.modifiers, modifier);
 		// we always add explicit modifiers, then we have to remove first implicit one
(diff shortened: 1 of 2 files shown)


src/main/java/spoon/support/reflect/CtModifierHandler.java
```diff
@@ -57,7 +58,7 @@ public CtModifierHandler setExtendedModifiers(Set<CtExtendedModifier> extendedMo
 	}
 
 	public Set<ModifierKind> getModifiers() {
-		return modifiers.stream().map(CtExtendedModifier::getKind).collect(Collectors.toSet());
+		return modifiers.stream().map(CtExtendedModifier::getKind).collect(linkedHashSetCollector());
```
slarse, line 61: I suggest using [`Collectors.toCollection()`](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Collectors.html#toCollection-java.util.function.Supplier-) instead of writing a collector from scratch.

```suggestion
		return modifiers.stream().map(CtExtendedModifier::getKind).collect(Collectors.toCollection(LinkedHashSet::new));
```

slarse:
@raghav-deepsource As we don't want a regression of this behavior, could you add tests that verify that the insertion order of modifiers is preserved? Insertion order for all of `setExtendedModifiers`, `setModifiers` and `addModifier` must be reflected in `getModifiers` and `getExtendedModifiers`.

src/test/java/spoon/support/reflect/CtModifierHandlerTest.java
```diff
@@ -0,0 +1,67 @@
...
+    @Before
+    public void setUp() {
+        Launcher l = new Launcher();
+        Factory f = l.createFactory();
+        handler = new CtModifierHandler(f.createClass("org.test.Test"));
+    }
+
+    @Test
+    public void testAddModifier() {
+        inputModifierKinds.forEach(handler::addModifier);
```
raghav-deepsource, line 40: I'm actually enjoying java for once thanks to this syntax!

src/test/java/spoon/support/reflect/CtModifierHandlerTest.java
```diff
@@ -0,0 +1,67 @@
...
+    LinkedHashSet<CtExtendedModifier> inputExtendedModifiers = inputModifierKinds.stream().map(CtExtendedModifier::new).collect(Collectors.toCollection(LinkedHashSet::new));
+
+    // these will be overwritten in each test.
+    Set<ModifierKind> modifiers;
+    Set<CtExtendedModifier> modifiersExt;
+    CtModifierHandler handler;
+
+    @Before
+    public void setUp() {
+        Launcher l = new Launcher();
```
I-Al-Istannen, line 33: Why not name them `launcher` and `factory`? One letter names make this code here a lot harder to read than would be necessary.

src/test/java/spoon/support/reflect/CtModifierHandlerTest.java
```diff
@@ -0,0 +1,67 @@
...
+    LinkedHashSet<ModifierKind> inputModifierKinds = new LinkedHashSet<>(Arrays.asList(
+            ModifierKind.STATIC,
+            ModifierKind.PUBLIC,
+            ModifierKind.ABSTRACT,
+            ModifierKind.NATIVE
+    ));
+    LinkedHashSet<CtExtendedModifier> inputExtendedModifiers = inputModifierKinds.stream().map(CtExtendedModifier::new).collect(Collectors.toCollection(LinkedHashSet::new));
+
+    // these will be overwritten in each test.
+    Set<ModifierKind> modifiers;
```
I-Al-Istannen, line 27: I know this is a test, but do all those variables really need to be package private?

src/test/java/spoon/support/reflect/CtModifierHandlerTest.java
```diff
@@ -0,0 +1,67 @@
...
+import spoon.Launcher;
+import spoon.reflect.declaration.ModifierKind;
+import spoon.reflect.factory.Factory;
+
+import java.util.Arrays;
+import java.util.LinkedHashSet;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import static org.junit.Assert.*;
```
I-Al-Istannen, line 14: Can you migrate this test to Junit 5? (`Assertions`)

src/test/java/spoon/support/reflect/CtModifierHandlerTest.java
I-Al-Istannen: Why do you use class/instance fields and not local variables for `inputModifierKinds`, `modifiers`, `modifiersExt` and so on?

src/test/java/spoon/support/reflect/CtModifierHandlerTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+public class CtModifierHandlerTest {
+
+    LinkedHashSet<ModifierKind> inputModifierKinds = new LinkedHashSet<>(Arrays.asList(
```
I-Al-Istannen, line 18: Why not use Lists here if the order is important? That would also ensure the test *actually fails* if the order is wrong! Currently it just accepts any order as the equals method doesn't take order into account (see the comment further down).

```suggestion
    List<ModifierKind> inputModifierKinds = Arrays.asList(
```

src/test/java/spoon/support/reflect/CtModifierHandlerTest.java
```diff
@@ -0,0 +1,67 @@
...
+            ModifierKind.STATIC,
+            ModifierKind.PUBLIC,
+            ModifierKind.ABSTRACT,
+            ModifierKind.NATIVE
+    ));
+    LinkedHashSet<CtExtendedModifier> inputExtendedModifiers = inputModifierKinds.stream().map(CtExtendedModifier::new).collect(Collectors.toCollection(LinkedHashSet::new));
```
I-Al-Istannen, line 24: Maybe chop this down?
```suggestion
    LinkedHashSet<CtExtendedModifier> inputExtendedModifiers = inputModifierKinds
        .stream()
        .map(CtExtendedModifier::new)
        .collect(Collectors.toCollection(LinkedHashSet::new));
```

src/test/java/spoon/support/reflect/CtModifierHandlerTest.java
```diff
@@ -0,0 +1,67 @@
...
+        handler = new CtModifierHandler(f.createClass("org.test.Test"));
+    }
+
+    @Test
+    public void testAddModifier() {
+        inputModifierKinds.forEach(handler::addModifier);
+        modifiersExt = handler.getExtendedModifiers();
+        modifiers = handler.getModifiers();
+
+        assertEquals("Order of modifiers was not preserved.", inputModifierKinds, modifiers);
```
I-Al-Istannen, line 44: I am pretty sure this doesn't work as the LinkedHashSet uses the equals from HashSet:
```
jshell> Set<String> a = new LinkedHashSet<>(List.of("a", "b"))
a ==> [a, b]

jshell> Set<String> b = new LinkedHashSet<>(List.of("b", "a"))
b ==> [b, a]

jshell> a.equals(b)
$3 ==> true

```

src/test/java/spoon/support/reflect/CtModifierHandlerTest.java
```diff
@@ -0,0 +1,67 @@
...
+        assertEquals("Order of extended modifiers was not preserved.", inputExtendedModifiers, modifiersExt);
+    }
+
+    @Test
+    public void testSetModifiers() {
+        handler.setModifiers(inputModifierKinds);
+        modifiersExt = handler.getExtendedModifiers();
```
I-Al-Istannen, line 51: You could maybe extract lines `51-55` into a single `assertOrderUnchanged()` method and just call that from your multiple test methods?

I-Al-Istannen (changes requested):

src/test/java/spoon/support/reflect/CtModifierHandlerTest.java
raghav-deepsource: Wasn't aware of this. I'll rewrite the tests with this in mind.

src/test/java/spoon/support/reflect/CtModifierHandlerTest.java
raghav-deepsource: That was to reduce duplication. I had all of these tests within a single test function at one point which I refactored into its present form.

src/test/java/spoon/support/reflect/CtModifierHandlerTest.java
raghav-deepsource: I could use lists, but the problem with this is that I need to convert the input data into a set at some point so I can feed it into the setModifiers and setExtendedModifiers methods, which only accept sets. But maybe I could add overloads for these methods that accept lists

src/test/java/spoon/support/reflect/CtModifierHandlerTest.java
raghav-deepsource: I'll make those changes separately, since there's other stuff I'm changing as well

src/test/java/spoon/support/reflect/CtModifierHandlerTest.java
I-Al-Istannen: If the API only takes in a Set I would probably rather order them while printing. Otherwise you need to change all callsites and  users might accidentally break it:
Some types lack `addModifier` methods so users need to fetch them, add them to their own set and re-set them. I wouldn't expect this to make printing non-deterministic. I will raise this concern in the attached issue though.

src/test/java/spoon/support/reflect/CtModifierHandlerTest.java
raghav-deepsource: I'll make everything private

raghav-deepsource:
This PR was rather misguided, seeing as spoon generates modifiers as an unordered set anyway.  A better method has been brought up in #4033.

PR #4139, opened 2021-08-29
review: Use matching package with most contained types

### LINKED ISSUE #4051: bug: Classes in overlapping modules are not resolved correctly
## Problem statement
Jigsaw projects are allowed to define modules with *overlapping packages*. A real-world example of this is `java.desktop` and `java.datatransfer`, which share the `java.awt` package.
A self-contained example is this:
<details>
<summary>Example</summary>

```
.
├── bar
│  ├── common
│  │  └── Bar.java
│  └── module-info.java
└── foo
   ├── common
   │  └── Foo.java
   └── module-info.java
```

<details>
<summary>Bar module info</summary>

```java
module bar {
  exports common;

  requires foo;
}
```
</details>

<details>
<summary>Foo module info</summary>

```java
module foo {
  exports common;
}
```
</details>

<details>
<summary>Bar</summary>

```java
package common;

/**
 * I am bar and I am a class.
 */
public class Bar {
  public static void main(String args[]) {
    Foo.main(args);
  }
}
```
</details>

<details>
<summary>Foo</summary>

```java
package common;

/**
 * I am bar and I am a class.
 */
public class Foo {
  public static void main(String args[]) {
  }
}
```
</details>

</details>

([as a zip](https://github.com/INRIA/spoon/files/6830715/Sadness.zip))
This will cause the Javadoc class lookup for at least one of the classes to fail.

## Cause
I spent a few too many hours trying to make sense of JDT, Spoon and my current conclusion is the following:

When resolving comments spoon needs to know where in the file given source code elements appear. This is all fine, but when looking up a class in an overlapping module, spoon will call `PackageFactory#get` which will in turn iterate over all modules *and choose the first package it finds!* Spoon will retrieve the package from the wrong module, can't find the class in it and create a dummy class without any positions. This causes the comment insertion to attribute *all* comments to the surrounding `CtCompilationUnit`.

## Experimental validation
I added a print to the `PackageFactory#get` method, reporting when two or more packages were found and it returned this:
```
Oh no (x times)

Javadoc for Bar is:
  [/**
   * I am bar and I am a class.
   */]

Oh no (x times)

Javadoc for Foo is:
  []
```
As you can see it swallowed one of them, as it tried to look for `Foo` in the wrong module.

## Impact
My Javadoc Indexer fails to find any Javadoc for classes in `java.awt` due to the overlap mentioned at the beginning. This is a bit unfortunate.

## Possible solutions
1. We could merge all packages into one instead of picking a random one, which would at least allow us to find them here. This might, however, create conflicting qualified names which isn't awesome.
2. We could return a List of packages from `get(QualifiedName)`
    * We could check for any `requires` clause and only return packages in our module and any required one
3. We could look up types directly without taking the `Package()` factory middle-man which would solve the problem for unambiguous qualified names and if we also know the module context (and only look in our modules + required) this should work all the time

I have no idea how we can save the `CtPackage get(QualifiedName)` method, the current code seems to query for packages first and types second, which just doesn't work anymore in a post-module world.

I-Al-Istannen:
## About this PR
This PR closes #4051 by always using the matching package with most contained types.  This ensures we will skip "synthetic" packages in overlapping modules and fetch the package from the correct module.

## Considerations
- The `getContainedTypeCount` getter was added as `CtPackage#getAllTypes` copies the full set just to call `size` on it. The getter does pollute the `CtPackage` interface a bit though.

## Further Work
- Without the `Cache packages in CtModule` (https://github.com/INRIA/spoon/pull/4139/commits/e1b82d336c8c1ca5bfc8432b03bb88d2d4101f5e) commit the PR significantly increases runtime on codebases with modules. This happens as it calls `getPackageFromModule` on quite a few modules and that is a *costly* operation -- it iterates through the whole package chain each time:
    ```
   moduleA
     - fr
       - inria
         - spoon
           - a
   moduleB
     - fr
       - inria
         - spoon
           - b
   moduleC
     - fr
       - inria
         - spoon
           - c
    ```
    In this  example every type lookup of `fr.inria.spoon.*` would walk down the chain *in every module this application has*, causing a *lot* of unnecessary work.  
   The cache would need to be kept up-to-date though, which this PR *doesn't* implement as of now and might pose a challenge. Therefore I would trim off the commit and open a different PR for it after this one is merged.

### DIFF
--- a/src/main/java/spoon/reflect/declaration/CtPackage.java
+++ b/src/main/java/spoon/reflect/declaration/CtPackage.java
@@ -146,8 +146,6 @@ public interface CtPackage extends CtNamedElement, CtShadowable {
 	@Override
 	CtPackage clone();
 
-	int getContainedTypeCount();
-
 	/**
 	 * Returns {@code true} if this is an <i>unnamed</i> Java package.
 	 * See JLS §7.4.2. Unnamed Packages.
--- a/src/main/java/spoon/reflect/declaration/CtPackage.java
+++ b/src/main/java/spoon/reflect/declaration/CtPackage.java
@@ -146,6 +146,8 @@ public interface CtPackage extends CtNamedElement, CtShadowable {
 	@Override
 	CtPackage clone();
 
+	int getContainedTypeCount();
+
 	/**
 	 * Returns {@code true} if this is an <i>unnamed</i> Java package.
 	 * See JLS §7.4.2. Unnamed Packages.
--- a/src/main/java/spoon/reflect/factory/PackageFactory.java
+++ b/src/main/java/spoon/reflect/factory/PackageFactory.java
@@ -172,17 +172,21 @@ public class PackageFactory extends SubFactory {
 		//
 		// To solve this we look for the package with at least one contained type, effectively
 		// filtering out any synthetic packages.
+		CtPackage lastPackage = null;
 		for (CtModule module : factory.getModel().getAllModules()) {
 			CtPackage aPackage = getPackageFromModule(qualifiedName, module);
 			if (aPackage == null) {
 				continue;
 			}
+			lastPackage = aPackage;
 			if (!aPackage.getTypes().isEmpty()) {
 				return aPackage;
 			}
 		}
 
-		return null;
+		// return the last empty package we found. This is an educated guess but can be wrong if the
+		// module actually declaring the package has no types in the package.
+		return lastPackage;
 	}
 
 	/**
--- a/src/main/java/spoon/reflect/factory/PackageFactory.java
+++ b/src/main/java/spoon/reflect/factory/PackageFactory.java
@@ -12,6 +12,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.StringTokenizer;
+import spoon.SpoonException;
 import spoon.reflect.declaration.CtModule;
 import spoon.reflect.declaration.CtPackage;
 import spoon.reflect.declaration.CtPackageDeclaration;
@@ -172,6 +173,7 @@ public class PackageFactory extends SubFactory {
 		//
 		// To solve this we look for the package with at least one contained type, effectively
 		// filtering out any synthetic packages.
+		int foundPackageCount = 0;
 		CtPackage lastPackage = null;
 		for (CtModule module : factory.getModel().getAllModules()) {
 			CtPackage aPackage = getPackageFromModule(qualifiedName, module);
@@ -180,8 +182,17 @@ public class PackageFactory extends SubFactory {
 			}
 			lastPackage = aPackage;
 			if (!aPackage.getTypes().isEmpty()) {
-				return aPackage;
+				foundPackageCount++;
 			}
+		}
+
+		if (foundPackageCount > 1) {
+			throw new SpoonException(
+					"Ambiguous package name detected. If you believe the code you analyzed is correct, please"
+							+ " file an issue and reference https://github.com/INRIA/spoon/issues/4051. "
+							+ "Error details: Found " + foundPackageCount + " non-empty packages with name "
+							+ "'" + qualifiedName + "'"
+			);
 		}
 
 		// return the last empty package we found. This is an educated guess but can be wrong if the
--- a/src/main/java/spoon/reflect/factory/PackageFactory.java
+++ b/src/main/java/spoon/reflect/factory/PackageFactory.java
@@ -174,14 +174,16 @@ public class PackageFactory extends SubFactory {
 		// To solve this we look for the package with at least one contained type, effectively
 		// filtering out any synthetic packages.
 		int foundPackageCount = 0;
-		CtPackage lastPackage = null;
+		CtPackage packageWithTypes = null;
+		CtPackage lastNonNullPackage = null;
 		for (CtModule module : factory.getModel().getAllModules()) {
 			CtPackage aPackage = getPackageFromModule(qualifiedName, module);
 			if (aPackage == null) {
 				continue;
 			}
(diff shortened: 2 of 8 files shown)


I-Al-Istannen:
Okay, the version in this PR is slightly optimized (max instead of reverse sorting) and clocks in at ~ the same time for projects without modules (it shouldn't do *anything* extra there, as the comparator is not called for single element streams and we should not have duplicates there) and slows down JDK indexing by ~40 seconds (so ~20%).

That is probably more acceptable, though I don't know where the slowdown even comes from. Maybe it increases the lock contention on the concurrent skip list map.

Though running it with one instead of 8 threads (I have an 8 thread CPU) yields:
```
With this PR
java16 -Xmx8G -XX:ActiveProcessorCount=1   619,10s user 3,65s system 111% cpu 9:17,34 total

without this PR
java16 -Xmx8G -XX:ActiveProcessorCount=1   420,27s user 4,09s system 119% cpu 5:54,36 total
```
The output without this PR is wrong though and correctness trumps performance, it just looks so weird.

If you want I can detail the steps you'd need to do to try to reproduce this but your time is probably quite a bit more valuable.

I-Al-Istannen:
Most of the slowdown is due to the following: With the comparator java needs to evaluate `getPackageFromModule` for all modules and not just lazily for the first. That method is *really* heavy, causing substantial regressions in the JDK code where quite a few packages are shared between modules.

I optimized the `getPackageFromModule` visible in #4145 resulting in the following change:
image
Most of that time is actually spent in ConcurrentSkipListMap:
image
Changing it to a ConcurrentHashMap is quite a bit faster but breaks the ordering guarantees:
image

The last commit (`Cache packages in CtModule`) introduces a package cache in CtModule  - which I am sure will make you cry and destroy quite a few usecases where people decide to move packages between modules or change package instances or who knows what. If we could somehow cleanly maintain that it would bring down the time to around where it was before this PR, but with the correct behaviour now :)
```
/usr/lib/jvm/java-16-adoptopenjdk/bin/java -Xmx8G  -XX:+DebugNonSafepoints     1080,13s user 7,14s system 533% cpu 3:23,86 total
```

Code without modules *is not affected at all* from this PR as far as I can see (see [here](https://velcom.aaaaaaah.de/compare/fce1143f-ed03-4a1b-9ccf-f04727204ff8/to/63c409b7-60db-4ac3-88f1-ec31b5a25636?hash1=9e38a1a86339270e90950dc0315838233a7d5384)):
image

monperrus:
Thanks! The optimization is good, but I'm not sure we need a new method in the interface. WDYT?

I-Al-Istannen:
That only exists to prevent a useless set copy from being created every time, but I think with some stuff from SirYwells PR we can sidestep that issue and maintain the cleaner API.

I will remove that method in this PR, we can always re-add some smarter set creation later on.

monperrus:
LGTM. Could you document the core idea and contract also directly in the methods in ``PackageFactory``?

I-Al-Istannen:
While writing the comment you requested I noticed a possible simplification:
`javac` does not allow overlapping packages. So each package either has zero types in it (if it is synthetic) or one or more if it is the real one. This means we could remove the `max` calculation and just pick the first package that has a non-zero amount of types in it.

Sadly, this does not seem to be part of the java spec but is something `javac` enforces. JDT does not. With JDT you can have the *exact same class* in *the same package* in different modules. I am not sure how I should handle that case.

I could:
1. Assume overlapping packages won't ever occur and use the first package with a non-empty package
2. Throw an error if overlap is detected
3. Search through *all overlapping packages* to find a given type (Though this becomes impossible if you defined the same class in the same package in multiple modules)

I would prefer `1` as I do not believe that any real-world program would be written to contain overlapping packages if `javac` doesn't allow it.

What do you think?

slarse:
> `javac` does not allow overlapping packages. So each package either has zero types in it (if it is synthetic) or one or more if it is the real one. This means we could remove the `max` calculation and just pick the first package that has a non-zero amount of types in it.

`javac` does allow overlapping packages in modules, it just doesn't allow overlapping _exported_ packages. At least this was the case last time I tried, with JDK 15 I believe. See my comment here: https://github.com/INRIA/spoon/issues/4051#issuecomment-881546232

I'm not quite sure what to make of that.

I-Al-Istannen:
Right, I misremembered that. Then we are kinda screwed, no? If we need to look up a type we'd need to give that information to the `getPackage` method because it needs to resolve overlaps. I am not sure we can retrofit that into the current architecture.

This PR is probably fine as-is then but only addresses the problem partially. Using the package with most types might statistically have a few more hits than taking the first, but I am not sure that's worth it...

slarse:
> Right, I misremembered that. Then we are kinda screwed, no? 

Well, in terms of compile time, I suppose so. But as the module system system doesn't seem to be onboard with overlapping packages at runtime, I think it's pretty safe to assume that no one will write code like that in a single project.

That being said, it shouldn't be particularly expensive to check for overlap and throw an exception if there is overlap, right? I think doing that would be a nice safeguard, and good for usability in case someone does something crazy, _or_ we have misunderstood something about the module system.

I-Al-Istannen:
I had that originally but JDT seems to allow having overlapping packages and that caused the exception to trigger. If you are fine with rejecting such code I will re-add the check.

slarse (approved):
> I had that originally but JDT seems to allow having overlapping packages and that caused the exception to trigger. If you are fine with rejecting such code I will re-add the check.

Yes, I think we'll go with that first as it's an easy solution, and having overlapping packages seems to be entirely incompatible with the module system at runtime. So, we'll do it like that and... see what happens.

Otherwise, LGTM, thanks @I-Al-Istannen !

PR #4146, opened 2021-09-04
chore: Remove sorald-buildbreaker job

### LINKED ISSUE #4144: Disable sorald-buildbreaker
The version of SonarJava it uses is too prone to false positives, especially on dead stores. We need to upgrade the version of SonarJava in Sorald to 7+ before it can be useful.

slarse:
Fix #4144 

sorald-buildbreaker has had more false positives than true positives, and so we'll need to remove it. I just don't have time to maintain buildbreaker anymore. Instead of disabling the job, I'm going to just remove it, as otherwise I fear we may have it disabled indefinitely. The job is pretty trivial and clearly described in [the buildbreaker README](https://github.com/spoonlabs/sorald-buildbreaker), so there's really no need to keep it.

@algomaster99 If you get around to resuming work on buildbreaker, see this PR for the workflow config.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -136,17 +136,3 @@ jobs:
         run: ./chore/ci-extra.sh
       - name: Run Javadoc quality check
         run: ./chore/check-javadoc-regressions.py COMPARE_WITH_MASTER
-
-  sorald-buildbreaker:
-    runs-on: ubuntu-latest
-    name: Run Sorald Buildbreaker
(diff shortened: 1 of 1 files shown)


monperrus:
Thanks @slarse

algomaster99:
> false positives than true positives

@slarse could you please link some PRs where we faced a problem?

slarse:
I don't have too many off the top of my head, but see for example #4076 and #4134

PR #4148, opened 2021-09-06
Review: feat: add JLSViolation exception

MartinWitt:
As discussed in #4134 we need a more specific exception for JLS correctness.  See the record pr for the reason behind this.

### DIFF
--- a/src/main/java/spoon/JLSViolationException.java
+++ b/src/main/java/spoon/JLSViolationException.java
@@ -1,0 +1,24 @@
+/*
+ * SPDX-License-Identifier: (MIT OR CECILL-C)
+ *
+ * Copyright (C) 2006-2019 INRIA and contributors
+ *
+ * Spoon is available either under the terms of the MIT License (see LICENSE-MIT.txt) of the Cecill-C License (see LICENSE-CECILL-C.txt). You as the user are entitled to choose the terms under which to adopt Spoon.
+ */
+package spoon;
+/**
+ * This exception is thrown when an operation on a {@link spoon.reflect.declaration.CtElement} transfers it in an invalid state.
+ * <p>
+ * An invalid state is a state that is not conform to the JLS. For example, a {@link spoon.reflect.declaration.CtMethod} is in an invalid state if it is abstract and has a body.
+ */
+public class JLSViolationException extends SpoonException {
+
+	/**
+	 * Creates a new JLSViolationException with the given message.
+	 * @param msg  the reason of the exception.
+	 */
+	public JLSViolationException(String msg) {
+		super(msg);
+	}
+
+}
--- a/src/main/java/spoon/SpoonException.java
+++ b/src/main/java/spoon/SpoonException.java
@@ -10,6 +10,7 @@ package spoon;
 import java.lang.invoke.MethodHandles;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import spoon.compiler.Environment;
 import spoon.processing.FactoryAccessor;
 
 /** is a generic runtime exception for Spoon */
@@ -27,13 +28,19 @@ public class SpoonException extends RuntimeException {
 	public SpoonException(String msg, Throwable e) {
 		super(msg, e);
 	}
-
+	/**
+	 * Handles a JLSViolation according to the environment settings. If {@link Environment#getIgnoreSyntaxErrors()} is set to true, the exception is ignored.
+	 * Otherwise, the exception is thrown.
+	 * @param holder  an object with access to factory and environment.
+	 * @param reason  the reason for the exception.
+	 */
 	public static void handleJLSViolation(FactoryAccessor holder, String reason) {
-			if (holder.getFactory() != null && holder.getFactory().getEnvironment() != null
-					&& holder.getFactory().getEnvironment().getIgnoreSyntaxErrors()) {
-				throw new JLSViolation(reason);
-			} else {
-				LOGGER.info("An element is not compliant to the JLS. See: {}", reason);
-			}
+		if (holder != null && holder.getFactory() != null
+				&& holder.getFactory().getEnvironment() != null
+				&& !holder.getFactory().getEnvironment().getIgnoreSyntaxErrors()) {
+			throw new JLSViolation(reason);
+		} else {
+			LOGGER.info("An element is not compliant to the JLS. See: {}", reason);
+		}
 	}
 }
--- a/src/main/java/spoon/SpoonException.java
+++ b/src/main/java/spoon/SpoonException.java
@@ -10,8 +10,6 @@ package spoon;
 import java.lang.invoke.MethodHandles;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import spoon.compiler.Environment;
-import spoon.processing.FactoryAccessor;
 
 /** is a generic runtime exception for Spoon */
 public class SpoonException extends RuntimeException {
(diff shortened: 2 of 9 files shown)


src/main/java/spoon/reflect/declaration/CtElement.java
```diff
@@ -421,4 +421,10 @@
 	// overriding the return type
 	@Override
 	ElementSourceFragment getOriginalSourceFragment();
+
+	/**
+	 * Throws an {@link spoon.JLSViolationException} if enabled, logs otherwise.
+	 * @param reason the reason why this element is not compliant with the JLS.
+	 */
+	void throwJLSViolationException(String reason);
```
monperrus, line 429: This method is not related to the element's state. I feel we should have it elsewhere.

Proposal: as static method in SpoonException?

monperrus (commented):
Interesting and well-scoped baby PR.

* see comment
* we need a test case

Thanks!

src/main/java/spoon/reflect/declaration/CtElement.java
MartinWitt: The problem with moving it away form `CtElement` is that we need the environment or the factory for the check if we need to throw the exception. I'm unsure if passing a factory or environment to the exception creation produces good code.
```java
public void throwJLSViolation(Factory fac, String reason)
```
looks a bit strange to me.

slarse: I agree with @monperrus, this method should not be on `CtElement`. It's a helper method for Spoon internals, exposing it in the public API is not desirable.

I also don't really like the idea of having a method named `throwX`, that may not actually throw X. A better name would probably be something like `handleJLSViolation`, but I'm not entirely sold on the necessity for the method to begin with.

src/main/java/spoon/JLSViolationException.java
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * Spoon is available either under the terms of the MIT License (see LICENSE-MIT.txt) of the Cecill-C License (see LICENSE-CECILL-C.txt). You as the user are entitled to choose the terms under which to adopt Spoon.
+ */
+package spoon;
+/**
+ * This exception is thrown when an operation on a {@link spoon.reflect.declaration.CtElement} transfers it in an invalid state.
+ * <p>
+ * An invalid state is a state that is not conform to the JLS. For example, a {@link spoon.reflect.declaration.CtMethod} is in an invalid state if it is abstract and has a body.
+ */
+public class JLSViolationException extends SpoonException {
```
slarse, line 14: This might be a controversial idea, but why not just call it `JLSViolation`? When thrown, we'll know it's an exception because only exceptions can be thrown.

src/main/java/spoon/compiler/Environment.java
```diff
@@ -589,6 +589,16 @@
...
+	/**
+	 * Whether Spoon currently ignores JLS correctness.
+	 * @return true if spoon ignores JLS correctness, false if spoon throws {@link spoon.JLSViolationException} when it detects JLS violation.
+	 */
+	boolean isIgnoreJLSCorrectness();
+	/**
+	 * Sets whether Spoon should ignore JLS violations.
+	 * @param ignoreJLSCorrectness true if spoon should ignore JLS violations, false if spoon should throw {@link spoon.JLSViolationException} when it detects JLS violation.
+	 */
+	void setIgnoreJLSCorrectness(boolean ignoreJLSCorrectness);
```
slarse, line 601: This concept is extremely close to `setIgnoreSyntaxErrors`. A JLS violation is in most situations a syntax error. If this is to be added as a separate mode, it needs to be clearly distinguished from `setIgnoreSyntaxErrors`.

The key thing about `setIgnoreSyntaxErrors` is that it's a filter to ignore files with syntax errors in the input. This new proposed flag could perhaps distinguish itself by being explicitly about _maintaining_ JLS compliance in the Spoon model.

What I'm trying to get across is that if I were to read these docs, the distinction between this flag and `setIgnoreSyntaxErrors` would not be clear to me.

Finally, I'd argue this should be called "ignore JLS violations" or "skip JLS correctness checks".  I understand what "ignore correctness" is meant to imply, but it's a bit of a brain twister for me.

monperrus:
About the concept here there are two things.

- introducing a new exception
- having this as configurable behavior

The former is OK to me. We can already put in this PR some usages.

As @slarse I'm not enthusiastic about the latter because 1) configurable behavior is hard to understand for clients 2) configurable behavior is harder to test and maintain 2) configurable behavior for exceptions is especially caveat.

WDYT?

MartinWitt:
My intention with allowing non JLS compliant code was that some (maybe existing) transformations have steps which produce non JLS code. I think we shouldn't remove the ability to write transformations freely. But as @slarse pointed out correctly, the JLS correctness is pretty close to `tIgnoreSyntaxError` and the existing flag should be used.

> We can already put in this PR some usages.

This could be hard as Spoon ignores a lot of JLS restrictions and is more focused on building any modell.

src/main/java/spoon/compiler/Environment.java
MartinWitt:
```java
	/**
	 * Sets the option ignore-syntax-errors to remove files with any syntax errors from the compilation batch.
	 *
	 * @param ignoreSyntaxErrors whether Spoon should ignore files with any syntax errors or JLS violations
	 */
	void setIgnoreSyntaxErrors(boolean ignoreSyntaxErrors);
```
After reading these docs, I think there is a problem with merging these 2 flags, because building a model with a wrong identifier is possible, but the flag removes the file completely. I'm unsure if ignoring = removing is expected behavior.

src/main/java/spoon/SpoonException.java
```diff
@@ -21,4 +28,19 @@ public SpoonException(Throwable e) {
...
+	 * @param reason  the reason for the exception.
+	 */
+	public static void handleJLSViolation(FactoryAccessor holder, String reason) {
+		if (holder != null && holder.getFactory() != null
+				&& holder.getFactory().getEnvironment() != null
+				&& !holder.getFactory().getEnvironment().getIgnoreSyntaxErrors()) {
+			throw new JLSViolation(reason);
+		} else {
+			LOGGER.info("An element is not compliant to the JLS. See: {}", reason);
+		}
```
MartinWitt, line 44: Here we could say ignoring means no logging at all. Any opinions on just log it, even if the user said `gnoreSyntaxErrors`?

src/main/java/spoon/JLSViolationException.java
MartinWitt: Renamed the exception

src/main/java/spoon/reflect/declaration/CtElement.java
MartinWitt: I moved the Exception to `SpoonException` in a static until method. 

> but I'm not entirely sold on the necessity for the method to begin with.

The goal of this method is to not repeat the check if the exception should be logged or thrown at every place we have JLSViolations.

monperrus:
Thanks for the update. LGTM.

src/main/java/spoon/SpoonException.java
slarse: It's fine to log IMO.

src/main/java/spoon/SpoonException.java
```diff
@@ -21,4 +28,19 @@ public SpoonException(Throwable e) {
 	public SpoonException(String msg, Throwable e) {
 		super(msg, e);
 	}
+	/**
+	 * Handles a JLSViolation according to the environment settings. If {@link Environment#getIgnoreSyntaxErrors()} is set to true, the exception is ignored.
+	 * Otherwise, the exception is thrown.
+	 * @param holder  an object with access to factory and environment.
+	 * @param reason  the reason for the exception.
+	 */
+	public static void handleJLSViolation(FactoryAccessor holder, String reason) {
```
slarse, line 37: I don't think this method belongs here either. This makes `SpoonException` dependent on its subtype, which is undesirable and IMO violates SRO (`SpoonException` might have to change if `JLSViolation` changes).

If this method should exist, and should be the prime way of throwing a `JLSViolation`, I would put it as a static method on `JLSViolation` and make its constructor private. I.e. like so:

```java
public class JLSViolation extends SpoonException {

	private JLSViolation(String msg) {
		super(msg);
	}

	public static void throwIfSyntaxErrorsAreNotIgnored(FactoryAccessor holder, String reason) {
		// logic    
	}

}
```

Then usage becomes a bit easier to read, I think:

```java
// current
SpoonException.handleJLSViolation(...);

// my proposition
JLSViolation.throwIfSyntaxErrorsAreNotIgnored(...);
```

What do you think? Agree, disagree?

slarse (changes requested):
See my comment on the code.

In addition, this changes the semantics of `setIgnoreSyntaxErrors`, and its docs must be updated. It's no longer only about ignoring syntax errors in files, but also in transformations, which must be expressed in the docs.

src/main/java/spoon/SpoonException.java
MartinWitt: 100% correct, I moved the method and changed the name to your better name.

slarse (approved):
LGTM

I think using more specific exceptions is a step in the right direction for Spoon overall.

slarse:
Thanks @MartinWitt

Issue #4151, opened 2021-09-08
reproducible builds for Spoon


monperrus:
concept: the creation of Jar files should be deterministic, i.e. we should have a reproducible build, see https://reproducible-builds.org/

maybe it is already, maybe it is not, to be checked.

then, we have different solutions:

* https://maven.apache.org/guides/mini/guide-reproducible-builds.html
* https://zlika.github.io/reproducible-build-maven-plugin/

Tasks: 

* [x] build is reproducible (manually)
* [x] check reproducibility in CI (see https://github.com/INRIA/spoon/pull/4653)
* [x] create buildspec file, see https://github.com/INRIA/spoon/issues/4151#issuecomment-1079887720
* [ ] periodically rebuild past releases

related to #3867

monperrus:
See also thread "Reproducible builds on Java", https://lists.reproducible-builds.org/pipermail/rb-general/2021-September/002374.html

I-Al-Istannen:
```patch
diff --git a/spoon-pom/pom.xml b/spoon-pom/pom.xml
index 7a352b09..eef300ae 100644
--- a/spoon-pom/pom.xml
+++ b/spoon-pom/pom.xml
@@ -24,6 +24,7 @@
         <java.test.version>1.8</java.test.version>
         <runtime.log>target/velocity.log</runtime.log>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <project.build.outputTimestamp>1</project.build.outputTimestamp>
     </properties>
 
     <dependencies>
@@ -108,6 +109,11 @@
     <build>
         <defaultGoal>clean install</defaultGoal>
         <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <version>3.2.0</version>
+            </plugin>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-surefire-plugin</artifactId>
```
Yields builds that consistently have the same hash for me. So at least in the core that seems to work (after updating the jar plugin).

I haven't checked the other submodules.

monperrus:
oh nice! would you make a PR?

monperrus:
Thanks a lot @I-Al-Istannen for #4152!

I would say that the next step is to check reproducibility in CI, by packaging twice.

Thoughts?

seintur:
Hi Martin @monperrus 

[...]

PR #4152, opened 2021-09-09
chore: setup reproducible builds for spoon-core

I-Al-Istannen:
This patch upgrades the maven jar plugin to 3.2.0 and disables timestamps in maven to make the spoon-core (and maybe others) build reproducible.

See https://maven.apache.org/guides/mini/guide-reproducible-builds.html

Partially addresses #4151.


monperrus:
Thanks, LGTM. Unwip?

I-Al-Istannen:
Ah right, I tested it locally without running tests as those just don't work on my computer for reasons I am not privy too, but running tests shouldn't change anything. I will unwip it.

monperrus:
LGTM, will merge.

PR #4157, opened 2021-09-13
review: fix: Handle nested `}` in inline tag

### LINKED ISSUE #4154: bug: Javadoc inline tag parsing is too primitive
## Expectation
Parsing the following javadoc (SwitchPoint)
```java
/**
 * Here is an example of a switch point in action:
 * <pre>{@code
 * MethodHandle MH_strcat = MethodHandles.lookup()
 *     .findVirtual(String.class, "concat", MethodType.methodType(String.class, String.class));
 * SwitchPoint spt = new SwitchPoint();
 * assert(!spt.hasBeenInvalidated());
 * // the following steps may be repeated to re-use the same switch point:
 * MethodHandle worker1 = MH_strcat;
 * MethodHandle worker2 = MethodHandles.permuteArguments(MH_strcat, MH_strcat.type(), 1, 0);
 * MethodHandle worker = spt.guardWithTest(worker1, worker2);
 * assertEquals("method", (String) worker.invokeExact("met", "hod"));
 * SwitchPoint.invalidateAll(new SwitchPoint[]{ spt });
 * assert(spt.hasBeenInvalidated());
 * assertEquals("hodmet", (String) worker.invokeExact("met", "hod"));
 * }</pre>
 */
```
should return a `@code` tag that spans the entire quoted range (minus the `<pre>` stuff) as the code ends there.

## Problem
It sadly returns a code tag ending **at the first encountered `}`**. Here is an illustrative image of what spoon parses (the same as the Neovim code highlighting and also the Github highlighting in the block above. Looks like about any tool is broken, but such is the life of parsing Javadoc):
grafik

## Proposed solution
We should try to end the code tag *at the matching closing curly brace*. This seems to be what the standard HTML doclet does, as it renders
```java
/**
 * Hello. {@code This is some code with a void() { method here}}.
 * <p>
 * And some more with broken indent: {@code Hello{}. This is outside?
 * <p>
 * Hmm}. Random closing.
 */
```
to
```html
<div class="block">Hello. <code>This is some code with a void() { method here}</code>.
 <p>
 And some more with broken indent: <code>Hello{}. This is outside?
 &lt;p&gt;
 Hmm</code>. Random closing.</p></div>
```

I-Al-Istannen:
An inline tag does *not* end at the next `}` bracket but at the next
*matching* `}` bracket. This can be especially noticeable in code tags:

```
                  Tag doesn't end here
                         |
                         v
{@code public class Foo {}}
                          ^
                    Tag ends here
```

Closes #4154.

### DIFF
--- a/src/main/java/spoon/javadoc/internal/Javadoc.java
+++ b/src/main/java/spoon/javadoc/internal/Javadoc.java
@@ -146,8 +146,32 @@ public class Javadoc implements Serializable {
 		if (index == -1) {
 			return null;
 		}
-		// we are interested only in complete inline tags
-		int closeIndex = text.indexOf("}", index);
+
+		// Find the corresponding end curly brace:
+		//
+		//                 Tag doesn't end here
+		//                            |
+		//                            v
+		//   {@code public class Foo {}}
+		//   ^                         ^
+		//   |                         |
+		// index                 Tag ends here
+
+		int closeIndex = -1;
+		int nesting = 1;
+		for (int i = index + 2; i < text.length(); i++) {
+			char read = text.charAt(i);
+			if (read == '{') {
+				nesting++;
+			} else if (read == '}') {
+				nesting--;
+			}
+			if (nesting == 0) {
+				closeIndex = i;
+				break;
+			}
+		}
+
 		if (closeIndex == -1) {
 			return null;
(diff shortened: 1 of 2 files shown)


monperrus:

PR #4158, opened 2021-09-13
review: refactor: Replace MapUtils#getOrCreate with Map#computeIfAbsent

SirYwell:
The removed class was only used in those two calls, which can simply be replaced with the computeIfAbsent method provided by the JDK.


monperrus:
Nice. LGTM.

slarse:
`+6 -52` diff, very nice!

monperrus:
Thanks @SirYwell

PR #4160, opened 2021-09-14
chore: Pin setup-java action to tag

slarse:
#4156 showed that I'd missed pinning one of the setup-java actions to a tag. This PR fixes that, which makes renovate only suggest updates when a new version is available (as opposed to on every new push to the main branch of the action).


monperrus:
Thanks @slarse

PR #4164, opened 2021-09-15
wip: chore(deps): update jdt to Java 17

SirYwell:
Supersedes #4161.

This JDT release adds support for Java 17. There was one breaking change with spoon that was simple to fix as the representation was just unified.

**Edit**: Oh well, that worked locally because I'm running it with Java 16. Seems like JDT now requires Java 11 as minimum version. How do we want to deal with that? @monperrus 

Blocked by ~~#4169~~ #4173

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -48,7 +48,7 @@
     <dependency>
       <groupId>org.eclipse.jdt</groupId>
       <artifactId>org.eclipse.jdt.core</artifactId>
-      <version>3.26.0</version>
+      <version>3.27.0</version>
       <exclusions>
         <exclusion>
           <groupId>org.eclipse.platform</groupId>
--- a/src/main/java/spoon/support/compiler/jdt/ParentExiter.java
+++ b/src/main/java/spoon/support/compiler/jdt/ParentExiter.java
@@ -473,7 +473,7 @@ public class ParentExiter extends CtInheritanceScanner {
 		if (node instanceof CaseStatement) {
 			caseStatement.setCaseKind(((CaseStatement) node).isExpr ? CaseKind.ARROW : CaseKind.COLON);
 		}
-		if (node instanceof CaseStatement && ((CaseStatement) node).constantExpression != null && child instanceof CtExpression
+		if (node instanceof CaseStatement && ((CaseStatement) node).constantExpressions != null && child instanceof CtExpression
 				&& caseStatement.getCaseExpressions().size() < ((CaseStatement) node).constantExpressions.length) {
 			caseStatement.addCaseExpression((CtExpression<E>) child);
 			return;


slarse:
> Edit: Oh well, that worked locally because I'm running it with Java 16. Seems like JDT now requires Java 11 as minimum version. How do we want to deal with that? @monperrus

Well. That's a problem. Here's a discussion about it: https://www.eclipse.org/lists/cross-project-issues-dev/msg18507.html

As I see it, we don't have much more of a choice than to drop support for running Spoon with Java 8. That's unfortunate as Java 8 is supposed to have LTS support from Oracle until 2030. But it's either that, or we don't support any features after Java 16 (i.e we stagnate). Given the rapid release cycle of Java versions, I just don't see that as an option.

SirYwell:
From my *personal* view, Java 11 is totally fine! From a project view, statistics of java versions spoon actually runs with would be interesting. This is likely something you have more information than me. 
Looking into projects in <https://github.com/INRIA/spoon/network/dependents> might be useful (though it only brings up projects using maven afaik), but even then you'd need to filter projects that weren't updated in a while I guess.

monperrus:
I agree, it's more strategic to move on together with JDT. Ok for me to drop Java 8.

spoon-control-flow/pom.xml
```diff
@@ -29,8 +29,8 @@
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>3.8.1</version>
                 <configuration>
-                    <source>1.8</source>
-                    <target>1.8</target>
+                    <source>${java.src.version}</source>
+                    <target>${java.src.version}</target>
```
MartinWitt, line 33: Switch to maven.compiler.release flag? JDK8 is gone [https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html#release](https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html#release)

spoon-pom/pom.xml
```diff
@@ -20,8 +20,8 @@
     <inceptionYear>2007</inceptionYear>
 
     <properties>
-        <java.src.version>1.8</java.src.version>
-        <java.test.version>1.8</java.test.version>
+        <java.src.version>11</java.src.version>
+        <java.test.version>11</java.test.version>
```
MartinWitt, line 24: Switch to `maven.compiler.release` flag? JDK8 is gone [https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html#release](https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html#release)

slarse:
> From my _personal_ view, Java 11 is totally fine! From a project view, statistics of java versions spoon actually runs with would be interesting. This is likely something you have more information than me.
> Looking into projects in https://github.com/INRIA/spoon/network/dependents might be useful (though it only brings up projects using maven afaik), but even then you'd need to filter projects that weren't updated in a while I guess.

In the open source world, I would imagine moving to Java 11 is a non-issue. It's potential commercial use cases where JDK8 still lives and thrives. And we know that there are commercial users of Spoon as we sometimes get bug reports where the reporters are unable to share the actual projects.

But like @monperrus said, we move with JDT. And from a Spoon dev point of view it's going to be really nice to have some of the QOL features that have been introduced since Java 8.

SirYwell:
I changed a few things now:

- Github Actions are running with Java 11 and Java 16 right now, using the [Temurin](https://adoptium.net/) builds now. There are no Java 17 builds from Adoptium yet, we might want to wait for that and add it too?
- Extra checks and Coverage run with Java 16, as Adoptium does not provide Java 15 builds and AdoptOpenJDK won't have Java 17 builds
- Dataflow was a little bit harder to make it work with Java 16, as I needed to update it to Gradle 7 (thanks @MartinWitt)

There are more things to do, which I would think are more appropriate in separate PRs:

- Remove tests that only run with Java <11
- General refactoring (I found https://github.com/INRIA/spoon/blob/2fff4c50266f79bc06b961e3b01ca6c6e0fcc1ab/src/main/java/spoon/support/visitor/java/reflect/RtMethod.java#L153-L175 which seems to be even pre Java 8)
- Implement support for Java 17 features in spoon

monperrus:
That's really good work @SirYwell, kudos an thanks. The PR now contains many different things.

Can we put the Gradle update in a separate PR? And maybe the CI update as well? (both to be merged before this one)

Thanks!

SirYwell:
> Can we put the Gradle update in a separate PR? And maybe the CI update as well?

Sure. This PR now only contains the initial commits - it should build once the other PRs #4168 and #4169 are merged (and applied to this branch)

monperrus:
@SirYwell could you rebase to see how it looks here? Thanks!

SirYwell:
@monperrus done, it seems like LGTM runs with Java 8 and therefore fails to build. Everything else looks fine I guess

monperrus:
LTGM, OK to merge.

(LGTM does not LGTM)

I-Al-Istannen:
We can probably make LGTM look gooder to us by setting the java version manually (see [here](https://lgtm.com/help/lgtm/java-extraction#customizing-index))?

monperrus:
Thanks a lot @I-Al-Istannen

monperrus:
> We can probably make LGTM look gooder to us by setting the java version manually (see here)?

Good idea, would you give it a try?

PR #4165, opened 2021-09-15
review: Correctly parse inline tag names

I-Al-Istannen:
Spoon previously expected e.g. `docRoot` to be `doc_root`, which is not correct. This PR fixes this in a backwards-compatible way by setting the keyword to its actual value.


slarse (changes requested):

slarse (approved):
Nice, thanks @I-Al-Istannen

PR #4168, opened 2021-09-17
review: chore: Update gradle in dataflow

SirYwell:
This updates the gradle wrapper to gradle 7.2 and make the project use modern gradle features.

PR #4169, opened 2021-09-17
review: chore: upgrade to java 11

SirYwell:
This upgrades the projects and the ci to java 11.

Currently, there is no temurin build for Java 17. It probably makes sense to wait for that before merging.

(See https://github.com/adoptium/adoptium/issues/74 for the current status)

~~Blocked by:  #4168~~

See #4164 for more context


slarse:
We don't need to wait for a JDK 17 to be available in GitHub Actions. We test across multiple JDK versions as a precaution, but in the vast majority of cases testing on only Java 11 is sufficient. It was far more crucial with testing e.g. JDK 8 and JDK 11, as they had some massive differences.

slarse:
Btw you can proceed here now, the blocker is merged!

SirYwell:
Okay, I guess this is ready for review then.

slarse (approved):
Nice, thanks @SirYwell

slarse:
@monperrus Note that this is breaking, the next release of Spoon will not support Java 8.

slarse:
@monperrus Aaaand just after merging this I realize that all of the snapshot builds will break as they run with JDK 8. Oops. I'll have to deal with that later. Can you install Java 11 on the Ubuntu 16 agent?

slarse:
@SirYwell I didn't quite think ahead (as you can see above) and had to revert this PR. Can you open a new one?

SirYwell:
Yes, see #4173

monperrus:
done

PR #4170, opened 2021-09-19
review: refactor: Remove Java 7 specific code

SirYwell:
I found a spot where reflections where used to call Java 8 methods, for source compatibility with Java 7. This is no longer relevant and can be simplified.


slarse (approved):
Thanks @SirYwell, will merge!

PR #4173, opened 2021-09-21
chore: Move CI to Java >=11

SirYwell:
There we go.

~~This is blocked until https://github.com/INRIA/spoon/pull/4169#issuecomment-924351912 is resolved.~~ Not blocked anymore.


monperrus:
Created https://github.com/INRIA/spoon/issues/4179 for discussing the Java 8 issue?

spoon-pom/pom.xml
```diff
@@ -20,8 +20,7 @@
     <inceptionYear>2007</inceptionYear>
 
     <properties>
-        <java.src.version>1.8</java.src.version>
-        <java.test.version>1.8</java.test.version>
+        <maven.compiler.release>11</maven.compiler.release>
```
monperrus, line 23: what happens if we put "8" here as compiler target bytecode?

monperrus, line 23: What breaks if we keep 8?

I-Al-Istannen: The `target` flag does not set the bootstrap classpath correctly so things might break. Additionally, the JDT class files are compiled for Java 11, so lowering Spoon's target compatibility will have no real effect for end users.

slarse: Java 11 source code cannot target Java 8. Java 9, 10 and 11 introduced several changes to the class file format, so 11 is binary incompatible with 8.

`javac` will refuse to compile if you try to set source to `1.8` and target to `11`.

What's more, we depend on JDT whose binaries are compiled for Java 11, which a Java 8 JVM cannot execute. There are simply no tricks we can pull here that would allow us to run Spoon on a Java 8 JVM.

monperrus:
According to the comments on #4179, #4164  and here, it seems we'll proceed with merging this one soon.

slarse:
@monperrus Not quite yet, all of the builds on Jenkins need to be configured to use JDK 11 (you installed it, but all builds running on the ubuntu 16 node still use JDK 8).

Some builds will probably fail.

monperrus:
> all of the builds on Jenkins need to be configured to use JDK 11

We will do this after, it will be interesting to see what actually fails and why.

slarse:
> > all of the builds on Jenkins need to be configured to use JDK 11
> 
> We will do this after, it will be interesting to see what actually fails and why.

Almost every build will fail if we don't reconfigure them to use JDK 11, including the snapshot deployer. Afaik, only 2 or 3 builds are configured to use JDK 11. The rest will fail.

What I meant is that some builds will probably fail when we reconfigure them to use JDK 11, as they depend on JDK 8 behavior. In fact, now that I think about it, most of those projects are probably configured to target Java 8, and thus will not be possible to compile with Spoon after we require Java 11, regardless of which JDK is used to build.

So to summarize, without reconfiguring builds to use JDK11, we have guaranteed failures for most. With reconfiguring to JDK11, we'll probably have failures anyway.

monperrus:
We'll see. In terms of precedence, I'd go for merging this one first.

slarse:
Well, okay. Go ahead then. Just note that the Snapshot Deployer will fail immediately, so you may want to switch that over as soon as you see it go red.

monperrus:
Big jump!

Thanks a lot @SirYwell 

This will surely ripple for some time.

PR #4175, opened 2021-09-22
review: Fix implicit modifiers for nested interface member types

I-Al-Istannen:
Extracted from #4038.

Nested types in interfaces should be implicitly `public static `.


monperrus:
Nice, LGTM.

slarse (approved):
LGTM as well so I'll go ahead and merge :)

Thanks @I-Al-Istannen

Issue #4136, opened 2021-08-28
Adding and removing extended modifiers


SirYwell:
I was wondering why it isn't possible to directly add or remove extended modifiers. The only way is to copy the set, add your modifier and set it again. This is currently done in #4038 and #4134, and I'm currently trying to figure out some more issues with implicit modifiers where this functionality would make things a lot easier.

Is there any specific reason why such methods don't exist? If it just wasn't considered as useful before, I'd be happy to open a pull request for that.

slarse:
> Is there any specific reason why such methods don't exist?

No, it just hasn't been done I suppose. See discussion in #4033.

Ping @raghav-deepsource, I think you were perhaps looking at this as well?

raghav-deepsource:
Yup. Any thoughts on adding a way to pass a `List` of extended modifiers, instead of just a set? It would be useful for cases where the user wants to preserve the order of modifiers they assign to an element.

slarse:
> Any thoughts on adding a way to pass a List of extended modifiers, instead of just a set?

I'm not sure that's necessary. There are ordered sets. Really, the use cases where the order of (semantically) unordered elements matter are very few. We want to keep the API as small as possible.

If there's a concrete use case that can't be served by the current API (plus the now planned `add` and `remove` methods for extended modifiers), then we can consider it. Otherwise, I'd personally say no.

raghav-deepsource:
That makes sense.

SirYwell:
The problem with passing a List is, that it does not have an effect on the underlying implementation. Spoon only exposes a `Set` with no guarantees about order. One could stricten that requirement to return an ordered set but a method *taking* a List only makes sense if the underlying implementation can also return a List with the exact same order.

Also, the required order probably depends on the context - for the sniper printer the insertion order is more relevant, for another printer a strict order that does not depend on the insertion order is more relevant - I basically agree with what @I-Al-Istannen mentioned in #4033. (Though, lets not mix up the discussion about that in two issues now)

----

Another issue with extended modifiers is that, for example, records and enums are always static (the static modifier can be implicit or explicit if the reord is a nested class, but not if is a top level class) but the final modifier always needs to be implicit. I'm not sure in what extent this should be validated, but any validation there would be a lot easier with an `add` method too.

raghav-deepsource:
> Spoon only exposes a Set with no guarantees about order. 

I've changed the implementation of CtModifierHandler from a HashSet to a LinkedHashSet in #4122  which could change this.

[...]

PR #4176, opened 2021-09-22
review: Keep implicit modifiers when cloning

I-Al-Istannen:
Extracted from #4038.

When cloning elements the *CtExtendedModifiers* should be kept instead of the modifiers - otherwise the `implicit` state is lost.


I-Al-Istannen:
The test fails because it finally *can actually test what the contract states* but that functionality is broken. Not copying the implicit modifier accidentally fixed the test and it now successfully fails.

The problematic part is that the functionality the test wanted *is not implemented* as far as I can tell. `toString` and `toDebugString` seem to call basically the same code and excluding implicit modifiers is hard-coded in `ElementPrinterHelper#printModifiers`.

Should I just delete the test or rewrite the printer helper / printer to do what the contract of the test wants?

slarse:
> Should I just delete the test or rewrite the printer helper / printer to do what the contract of the test wants?

IMO the expected value is incorrect, remove `public` from it.

I-Al-Istannen:
That was a bit of an unexpected rabbithole but I think I have all of the weird cases I found covered now.

src/test/java/spoon/reflect/ast/CloneTest.java
```diff
@@ -265,4 +269,53 @@ public void testIssue3389() {
...
+				method.getExtendedModifiers()
+		);
+
+		CtMethod<?> clone = method.clone();
+		method.getExtendedModifiers().iterator().next().setImplicit(false);
+
+		assertEquals(
+				expectedModifiers,
+				clone.getExtendedModifiers()
+		);
```
slarse, line 319: Given that implicitness may be removed from the equality check of extended modifiers, this is rather fragile. I'd suggest altering this test to check that there are the expected amount of unique objects. Something like this:

```suggestion
		Set<CtExtendedModifier> methodModifiers = method.getExtendedModifiers();
		Set<CtExtendedModifier> clonedMethodModifiers = method.clone().getExtendedModifiers();

		Set<CtExtendedModifier> distinctModifiers = Collections.newSetFromMap(new IdentityHashMap<>());
		distinctModifiers.addAll(methodModifiers);
		distinctModifiers.addAll(clonedMethodModifiers);

		assertEquals(methodModifiers, clonedMethodModifiers);
		assertEquals(distinctModifiers.size(), methodModifiers.size() + clonedMethodModifiers.size());
```

The check on distinct modifiers could arguably be extracted into a separate helper method, but I think you get the idea.

src/test/java/spoon/reflect/ast/CloneTest.java
```diff
@@ -265,4 +269,53 @@ public void testIssue3389() {
...
+
+		assertEquals(
+				expectedModifiers,
+				method.getExtendedModifiers()
+		);
+
+		assertEquals(
+				expectedModifiers,
+				method.clone().getExtendedModifiers()
+		);
```
slarse, line 293: Same thing here, this might be rather fragile if extended modifiers are altered. Could you make this test more explicit in checking implicitness in particular?

I'd personally not parse this from source code, but instead just create anything modifiable, explicitly add modifiers to it, and then check that for the given modifiers the implicitness doesn't change when cloning. That is to say, don't use equals of the modifier object, but explicitly check the value of `isImplicit()` for some given modifier kind.

src/test/java/spoon/generating/CloneVisitorGenerator.java
```diff
@@ -417,15 +417,30 @@ private boolean avoidThrowUnsupportedOperationException(CtMethod<?> candidate) {
...
+					CtExecutableReference<Object> cloneReference = factory.Executable()
+							.createReference(
+									"Set<CtExtendedModifier> #clone(Set<CtExtendedModifier>)"
+							);
+					invocation = factory.Code().createInvocation(
+							null, cloneReference, invocation
+					);
+				}
+
+				return invocation;
```
slarse, line 435: Without the context of the explanation you provided in #4136, this would have me very confused. Even knowing that I was just slightly confused reading this. The big problem for me is that the method doesn't do what it says: it can either create a getter invocation, or create a clone invocation, but it only implies it does the former.

Try to refactor this such that the method name doesn't lie, and make sure to include an explanation for why it's necessary to create a clone call instead of a getter call for extended modifiers. I don't think I would have been able to figure that out.

One option would be to do something like so:

1. Leave `createGetterInvocation` unchanged.
2. Create a new method `createGetterOrCloneInvocation` and call that at the place where `createGetterInvocation` is now called. In this method, either delegate directly to `createGetterInvocation()` or return a clone invocation as appropriate. Make sure to provide an explanation for why the clone invocation is necessary.

That's just one way to do it, you may chose to do something different. As long as the methods do what they say, and there's an explanation for the clone call, it's all good.

slarse (changes requested):
I see your pain now with the mutable extended modifiers. Please see my comments.

src/test/java/spoon/generating/CloneVisitorGenerator.java
I-Al-Istannen: My motivation to mess with these results of ancient civilizations pursuing forbidding knowledge wasn't especially high and I didn't want the card house to collapse, but I refactored it slightly.

I moved the code out of the getter method as that was complex enough on its own and inlined it in the only callsite with an explanatory comment. I considered extracting it in a method but I found no name I felt actually helped so I left it in the current state. I am very open to suggestions on that front.

slarse: I fiddled around with a few different variations and landed on something that I think is somewhat clear. See comment.

src/test/java/spoon/generating/CloneVisitorGenerator.java
```diff
@@ -251,11 +251,40 @@ private CtInvocation createFactoryInvocation(CtVariableAccess<CtElement> element
...
+						);
+					}
+
+					CtInvocation<?> setterInvocation = createSetterInvocation(
+							element.getParameters().get(0).getType(),
+							setterOfField,
+							fetchValueInvocation
+					);
+
+					List<CtMethod<?>> methodsToAvoid = getCtMethodThrowUnsupportedOperation(setterOfField);
```
slarse, line 287: I suggest something like this:

```suggestion
					CtParameter<?> receiver = element.getParameters().get(0);
					CtMethod<?> setter = getSetterOf(ctField);
					CtMethod<?> getter = getGetterOf(ctField);
					CtInvocation<?> setterInvocation = createSetterInvocationWithGetterReturnValueAsArgument(
							receiver, getter, setter);
							
					List<CtMethod<?>> methodsToAvoid = getCtMethodThrowUnsupportedOperation(setter);
```

With the methods defined like so:

```java
			private CtInvocation<?> createFetchValueInvocation(CtParameter<?> receiver, CtMethod<?> getter) {
			    CtInvocation<?> getterInvocation = createGetterInvocation(receiver, getter);

				if (getter.getSimpleName().equals("getExtendedModifiers")) {
					// TODO explain here why we wrap the getter in a clone
					return wrapExtendedModifiersGetterInvocationInClone(getterInvocation);
				} else {
					return getterInvocation;
				}
			}

			private CtInvocation<?> wrapExtendedModifiersGetterInvocationInClone(CtInvocation<?> getterInvocation) {
				CtExecutableReference<Object> cloneReference = factory.Executable()
						.createReference(
								"Set<CtExtendedModifier> #clone(Set<CtExtendedModifier>)"
						);
				return factory.Code().createInvocation(null, cloneReference, getterInvocation);
			}
```

It's essentially what you initially wrote, but with the branching on `getExtendedModifiers` pulled up just one level above `createGetterInvocation. WDYT?

I-Al-Istannen: I don't love those names but as I said earlier I didn't come up with anything pretty either. I used your solution and added some (probably too verbose) comments which are hopefully helpful enough.

slarse: The original factoring of the method is, let's be honest, not great. To better isolate the behaviors and put more descriptive names on them, I'd say the whole thing would need to be refactored.

slarse: And I agree with the verbosity of the comments being a tad bit high. But, this is not a part of the code base that is touched often, and so the risk of them going out of date isn't that large.

As a note, I personally prefer putting a brief comment and then adding a link to the related issue for more context.

slarse (approved):
Thanks @I-Al-Istannen, let's go with this.

PR #4183, opened 2021-09-24
refactor(getActualClass): Improve design and reability

MartinWitt:
This PR tries to remove some old code and improve the code style. The method has the same behavior as before, but doesn’t need to build a model for array types anymore. In a feature PR there should be an enum for primitive types, but this is out of scope for this PR.


monperrus:
Thanks @MartinWitt. It seems that handling concurrency is the main goal here, right?

MartinWitt:
The main goal was to remove the `Launcher.parseClass` call, but because I was already refactoring the code, I fixed the rest also.

monperrus:
The diff is hard to decipher, but it is overall OK, plus it is code much executed by the test suite which is green.

OK to merge.

Next time, it'd be great to split in smaller and clearer PR :)

monperrus:
thanks @MartinWitt

PR #4185, opened 2021-09-26
review: refactor: Replace getOrCreate in Metamodel with computeIfAbsent

SirYwell:
In #4158 I removed the utility class and its usages, but it seems like the code was duplicated and used in the Metamodel too, I refactored the code to use computeIfAbsent there now too.


monperrus:
Thanks  @SirYwell.  LGTM, will merge.

PR #4202, opened 2021-10-04
wip: feature: Add static factory methods for CtExtendedModifier

SirYwell:
This adds two methods to `CtExtendedModifier` that allow to create instances without having to pass boolean values.

I deprecated the constructor where the implicitness defaulted to `false`, as the result is less obvious compared to the named method.

I also replaced the usages of the deprecated constructor and the usages of the other constructor where the second parameter was a boolean literal.

The usage as API might be more relevant with methods to directly add and remove extended modifiers, see #4136 for context.

### DIFF
--- a/src/main/java/spoon/support/reflect/CtExtendedModifier.java
+++ b/src/main/java/spoon/support/reflect/CtExtendedModifier.java
@@ -25,13 +25,46 @@ public class CtExtendedModifier implements SourcePositionHolder, Serializable {
 	private ModifierKind kind;
 	private SourcePosition position;
 
+	/**
+	 * Creates a new extended modifier of the given kind.
+	 *
+	 * @param kind the kind of this modifier.
+	 * @deprecated use {@link #explicit(ModifierKind)} to create an explicit modifier.
+	 */
+	@Deprecated
 	public CtExtendedModifier(ModifierKind kind) {
 		this.kind = kind;
 	}
 
+	/**
+	 * Creates a new extended modifier of the given kind with the given implicitness.
+	 *
+	 * @param kind the kind of this modifier.
+	 * @param implicit whether this modifier should be implicit.
+	 */
 	public CtExtendedModifier(ModifierKind kind, boolean implicit) {
 		this(kind);
 		this.implicit = implicit;
+	}
+
+	/**
+	 * Creates an extended modifier of the given kind that is explicit.
+	 *
+	 * @param kind the kind of the created modifier.
+	 * @return an explicit extended modifier.
+	 */
+	public static CtExtendedModifier explicit(ModifierKind kind) {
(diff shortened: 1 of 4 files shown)


src/main/java/spoon/support/reflect/CtExtendedModifier.java
```diff
@@ -25,15 +25,48 @@
 	private ModifierKind kind;
 	private SourcePosition position;
 
+	/**
+	 * Creates a new extended modifier of the given kind.
+	 *
+	 * @param kind the kind of this modifier.
+	 * @deprecated use {@link #explicit(ModifierKind)} to create an explicit modifier.
+	 */
+	@Deprecated
```
MartinWitt, line 34: Maybe add a `since` value like 9.2? [https://download.java.net/java/GA/jdk14/docs/api/java.base/java/lang/Deprecated.html#since()](https://download.java.net/java/GA/jdk14/docs/api/java.base/java/lang/Deprecated.html#since())

slarse:
> I would not deprecate the existing constructor to maximize backward compatibility (we love backward compatibility in Spoon). It just assumes a default implicitness, which we document and that's fine.

I disagree. By providing the `explicit` static method, we provide two ways of achieving the same thing, namely creating an explicit modifier. As a general rule, I think it's best to provide _one_ way of doing something, or at least one canonical way. By deprecating the constructor, we clearly show the preferred way of doing it, without actually affecting backward compatibility (we never need to _act_ on the deprecation, not really).

Just my two cents, I'm in favor of the deprecation _given_ that these static methods should be added at all (which I don't have a strong opinion about).

slarse:
> > we never need to _act_ on the deprecation
> 
> `@Deprecated` got a `forRemoval` value which defaults to false, so _if_ we want to remove it in future, we could use that - but I don't see a reason for that either.

I've never actually seen that used :). But it does further prove my point that deprecation does not mean "planned for removal". According to the Javadoc of `@Deprecated` (emphasis mine):

> A program element annotated @Deprecated is one that programmers are discouraged from using, typically because it is dangerous, **or because a better alternative exists**.

So I vote for deprecation of that constructor.

> The other constructor is currently used in places where the implicitness is given a boolean _variable_ (or parameter). This should still be possible without having deprecation warnings. But if we have a static factory method for that too, we could deprecate it I think.

I suppose that's fair, rewriting that code would be a bit awkward.

SirYwell:
> I've never actually seen that used

I forgot to mention that this was added in Java 9, so that might be the reason

monperrus:
> we never need to act on the deprecation

I love that. We use deprecation for signaling better alternatives (free added value for clients), but not for signaling removal (forced labor for clients).

Merging! Thanks  a lot!

PR #4204, opened 2021-10-05
review: fix: extract enum constant modifiers from the right place

SirYwell:
Previously, the enum values' modifiers were read from the declaring class by both the JDTTreeBuilder and the JavaReflectionTreeBuilder. This does not match the produced bytecode. For more context, see #4138.


monperrus:
Thanks @SirYwell 

LGTM. @slarse your knowledge about modifier design is more recent than mine, would you double-check and merge?

slarse (approved):
Nice, very clean PR, thanks @SirYwell!

For future reference, the "workaround" that is removed here is incorrect, enum constants (or rather the corresponding fields) _are_ always `public static final` (see https://github.com/INRIA/spoon/pull/4138#issuecomment-919305285)

PR #4205, opened 2021-10-05
wip: fix: extract implicit type modifiers correctly

SirYwell:
Previously, no implicit modifiers were extracted from types. I basically copied the way it is done for methods and fields.

I first thought it would only apply to enums, but it's also relevant for nested/local types and interfaces. I added assertions to the local types and test cases for enums and interfaces.

#4134 should probably also add test cases for records validating the modifiers.


monperrus:
LGTM! Thanks @SirYwell

PR #4207, opened 2021-10-05
refactor(*): replace equals/hashcode implementations with proper ones

MartinWitt:
This fixes three LGTM issues, `equals without type check`. I replaced the old equals/hashcode directly because they only compared `toString()` which seems kinda strange and not correct.


monperrus:
Thanks @MartinWitt

PR #4208, opened 2021-10-05
refactor(CtReferenceImpl): remove not needed abstract method and interface declaration

MartinWitt:
While looking at LGTM issues my IDE was a bit unhappy for the not needed interface and already in a superclass declared abstract method. This PR removes both.


monperrus:
thanks @MartinWitt

PR #4209, opened 2021-10-06
wip: refactor: use java 11 immutable lists where appropriate

SirYwell:
This replaces usages of `Collections.unmodifiableX()` when it was used to create an *immutable* collection (in most cases by wrapping a newly creating collection). 

This does not touch places where `Collections.unmodifiableX()` is used to expose an unmodifiable view for a collection that might be changed internally.


slarse (approved):
Very good, thanks @SirYwell

PR #4210, opened 2021-10-06
wip: fix: scan packages and extract their annotations

SirYwell:
Previously, packages where only visited from type references. This means the package of the type given to the `scan` method isn't necessarily visited, and its annotations aren't visited either.

This fixes it. The test case assumes that the package has the `@Deprecated` annotation declared in the `package-info.java`: https://github.com/INRIA/spoon/blob/63104c7b659c07fda43f1b3b5aa16898c3ec2e45/src/test/java/spoon/test/pkg/package-info.java#L5-L6

When debugging, I saw that packages (`java.util`, `java.lang`) are visited multiple times, as they are always visited whenever a type (new) or type reference (before) is visited. As the values of a package don't change at runtime (well, at least within a module), this doesn't seem necessary. Should we skip that by making sure that packages are only visited once here: https://github.com/INRIA/spoon/blob/9ff819461eefb1569cec29a92b6bbbb437eb47de/src/main/java/spoon/support/visitor/java/JavaReflectionTreeBuilder.java#L121-L129 ? I'm not really sure if this would cause any issues.

### DIFF
--- a/src/main/java/spoon/support/visitor/java/JavaReflectionVisitorImpl.java
+++ b/src/main/java/spoon/support/visitor/java/JavaReflectionVisitorImpl.java
@@ -29,12 +29,15 @@ import java.util.List;
 class JavaReflectionVisitorImpl implements JavaReflectionVisitor {
 	@Override
 	public void visitPackage(Package aPackage) {
+		for (Annotation annotation : aPackage.getDeclaredAnnotations()) {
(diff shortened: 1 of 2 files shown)


monperrus:
LGTM! Could you add the contract in the test? Thanks.

monperrus:
Thanks a lot @SirYwell

PR #4215, opened 2021-10-09
review: Rewrite type adaption

### LINKED ISSUE #4140: bug: Cannot resolve type erasure of the type parameter reference, which is not able to found it's declaration.
## Problem statement
Spoon fails to get the executable declaration of `crashy` in `Fooy` in the following code:
```java
package de.ialistannen.javadocapi;

public interface Test<F> {

  void crashy(F... paramTypes);
}

package de.ialistannen.javadocapi;

public interface Fooy extends Test<String> {

  @Override
  void crashy(String... paramTypes);
}
```
Please make sure the two interfaces are in *different java files*.

## Example test case
```java
	@Test
	public void test() {
		launcher.addInputResource("./src/test/java/spoon/test/generics/testclasses7/Fooy.java");
		launcher.addInputResource("./src/test/java/spoon/test/generics/testclasses7/Test.java");

		final Collection<CtType<?>> types = ctModel.getAllTypes();
		for (CtType<?> type : types) {
			System.out.println("Executables: " + type.getAllExecutables());
			for (CtExecutableReference<?> reference : type.getAllExecutables()) {
				System.out.println(reference);
				System.out.println(reference.getExecutableDeclaration());
			}
		}
	}
```

## Stacktrace
`crashy(java.lang.Object[])` is a synthetic bridge method. It is needed as a caller owning a `Test<F>` can not know that `Fooy` is specialized to `String` and therefore the `Object[]` (erasure of `F` in `Test`) bridge method is generated. The method just checks the cast and forwards the call to the specialized `String...` method.
```
Type: Fooy
Executables: [crashy(java.lang.Object[]), crashy(java.lang.String[])]
crashy(java.lang.Object[])

spoon.SpoonException: Cannot resolve type erasure of the type parameter reference, which is not able to found it's declaration.

	at spoon.support.reflect.reference.CtTypeParameterReferenceImpl.getTypeErasure(CtTypeParameterReferenceImpl.java:176)
	at spoon.support.reflect.declaration.CtTypeImpl.isSameParameter(CtTypeImpl.java:752)
	at spoon.support.reflect.declaration.CtTypeImpl.hasSameParameters(CtTypeImpl.java:734)
	at spoon.support.reflect.declaration.CtTypeImpl.getMethod(CtTypeImpl.java:718)
	at spoon.support.reflect.reference.CtExecutableReferenceImpl.getCtExecutable(CtExecutableReferenceImpl.java:121)
	at spoon.support.reflect.reference.CtExecutableReferenceImpl.getExecutableDeclaration(CtExecutableReferenceImpl.java:113)
	at spoon.test.generics.GenericsTest.test(GenericsTest.java:1580)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:78)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:567)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:63)
	at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
	at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:413)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:69)
	at com.intellij.rt.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:33)
	at com.intellij.rt.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:235)
	at com.intellij.rt.junit.JUnitStarter.main(JUnitStarter.java:54)
```

## Probable cause
The type reference `F` it tries to resolve has no parent reference at all. Therefore, it can not find the declaration and that causes `getTypeErasure` to fail.

The same error occurs when you use `[]` instead of varargs.

I-Al-Istannen:
I was baited into rewriting type adaption a few weeks ago (damn you @MartinWitt and @SirYwell) and after suffering for a few too many working days it seems to be mostly compatible :)

Fixes #4140.


I-Al-Istannen:
@slarse Do you have any comments (especially for the questions in the OP) by chance?

slarse:
@I-Al-Istannen Oof, I've never touched any of this code, so I'd need to go through this all very carefully to answer with any amount of certainty. I'll put it on my TODO list (which is perilously long, no promises when I'll get around to it :( ). Frankly, you know more about this part of Spoon than I do at this point.

Perhaps @nharrand or @monperrus can provide some feedback? Otherwise I'll get around to it as soon as I can, I just can't promise when that is.

> The current type adaption apparently implements wildcard "contains" to resolve subtypes with respect to a type reference (e.g. List<?> is not a subtype of List<Object>).

I'm not entirely sure what you mean, but `List<?>` isn't a subtype of `List<Object>`, so that seems correct to me? `List<Object>` is however a subtype of `List<?>`.

> The current type adaption allows you to use an Invocation as context. This is never used exception in the generics test and I am not sure I see the utility. Do we need to keep that?

I have no idea.

slarse:
If you can point me to specific parts of the code base that you are uncertain about (with a nice hyperlink), then I might get around to this quicker :)

I-Al-Istannen:
All good :D

> I'm not entirely sure what you mean, but List<?> isn't a subtype of List<Object>, so that seems correct to me?

It's correct, but spoon AFAICT never uses *reference subtyping*, so it isn't *strictly* needed for spoon itself. You probably want to keep it anyways to ensure potential users don't break?

>  The current type adaption allows you to use an Invocation as context. This is never used exception in the generics test and I am not sure I see the utility. Do we need to keep that?

There is an API receiving a `CtInvocation` able to adapt another method to the invocation's type arguments ([here](https://github.com/INRIA/spoon/pull/4215/files#diff-86b3ab44f387e5aa35c2a86b9bb72c515395553345f36039377ed9603da2db69L1054-L1063)). I don't quite see the utility and the only usage was the linked test...

MartinWitt:
Could you push an empty commit for a qodana run?

slarse:
@MartinWitt you should be able to restart workflows manually now

slarse:
@I-Al-Istannen Sorry to keep you waiting here. It's unfortunate that these larger PRs get less attention than the smaller ones, seeing the amount of work behind them.

@monperrus @nharrand If either of you find the time to help out here, that'd be great. I don't know this part of the code base well (read: at all).

With any luck, I should have time to look at this during the coming weekend, but I'm not super confident I'll have all the answers :)

I-Al-Istannen:
> @I-Al-Istannen Sorry to keep you waiting here. It's unfortunate that these larger PRs get less attention than the smaller ones, seeing the amount of work behind them. It's just incredibly difficult and time consuming to review a PR of this size. With any luck, I should have time to look at this during the coming weekend, but I'm not super confident I'll have all the answers.

It's fine :)

---- 

And `@everyone taking a look`:

I am leaning towards re-implementing question 1 or delegating to the old implementation if I detect that I can not handle it properly. But a short look and some opinions on question 2 would be nice :)

You should probably read the markdown documentation I wrote up and can review just that if you are a bit short on time. It should contain enough context to answer the questions in the OP if you know the current spoon parts :)

I-Al-Istannen:
While this is probably not in a condition for *merging*, it now passes the last test (by delegating to the old impl if it detects that generic parameter analysis is needed).

### Todo-suggestion:
1. Is this something you'd even want in spoon? It refactors things to be (IMHO) easier than before and seems to provide a significant performance boost.
2. Does the approach seem reasonable?
3. Should I test the old *and new* implementation in parallel? I don't know how I would reasonably do that, but as we need to keep the old implementation around this might be something to look at. Currently I switched all tests to the new implementation.
4. Normal code review stuff

I-Al-Istannen:
I've rebased the PR on master (Thanks `@MartinWitt` for causing conflicts in every test :P That was a fun rebase...)

I've also compared it against the master branch by indexing JDK 17:

### Execution environment
Linux based, Intel i7-7700HQ (8) @ 3.800GHz, Laptop. Rough indexing trace:
grafik
The first part is building the spoon model. When all cores go to 100% my parallel Processor takes over and converts the model by inspecting classes, methods and fields.

Run command:
`time -v java -Xmx8G -jar target/JavadocApi.jar src/main/resources/Jdk-sample-config.json`

### Master branch
* Time taken: 4 min 46 sec
* gc-easy report: [Here](https://gceasy.io/my-gc-report.jsp?p=YXJjaGl2ZWQvMjAyMi8wMi85Ly0tYXBpLTAzOWViZDY3LTI4MDMtNDBjMy1hMmNjLWI4YmUxOTE3ZjQ4Yy50eHQ=&channel=API)
* You can also see the garbage collector kicking in periodically to drastically reduce the heap usage
grafik
* You can also see that quite a few objects are created
grafik


### This branch
* Time taken:  2 min 13 sec
* gc-easy report: [here](https://gceasy.io/my-gc-report.jsp?p=YXJjaGl2ZWQvMjAyMi8wMi85Ly0tYXBpLTJkMjZlODA1LTY4NGEtNDYyYi04Yzk3LWI4NzBmYzQwYjNmNy50eHQ=&channel=API)
* You can see that the heap usage is a lot more stable 
grafik
* And you can see that it created way fewer objects 
grafik


This seems to indicate that the changes in this PR do have some merit besides (hopefully) being easier to understand. The implementation also delegates to the old one if it isn't sure it can handle the request so correctness shouldn't suffer.

MartinWitt:
I had a call with @I-Al-Istannen explaining to me the problem this code solves. Sadly, this is not really separable in multiple PRs.
The following steps are:
- clean some Diffs stuff.
- add feature flag (and instead deprecated it)
- mark the whole API as internal

slarse:
@MartinWitt Are those the next steps to take?

MartinWitt:
Yes

I-Al-Istannen:
And to answer your original question: The problem with splitting up this PR is that it already only replaces more or less a single feature.

The different ways I could come up with for splitting this up have some drawbacks:
- I could leave out modifying tests, but then this would introduce untested changes into master.
- I could only implement subtyping, then make a second PR for adapting types, another for methods and so on. This would introduce lots of unfinished and half-tested APIs in master as the whole thing wouldn't *quite* be ready to replace the current implementation at each commit. Additionally, lots of the underlying machinery is needed for *all* the APIs this offers so I am not convinced this would get substantially smaller.

I will rebase this PR a bit to make it at least possible to review the changes a bit more easily, verify I did not mangle any test by accident and do what Martin outlined above. Maybe it will at least *look* a bit less daunting then :)

I-Al-Istannen:
Okay @slarse, I cleaned up the history of this PR into a series of independent commits that progressively convert more and more of the PR.

This should ease reviewing as each commit introduces exactly one new concept and you can focus on verifying it does what it says on the tin. It is still quite a large PR and if you'd rather have me split it up despite the negatives in my last comment, I'd be happy to oblige.

I also added a feature toggle (and deprecated it). I'd be interested in your opinion on how to *test this*. Currently the tests use the new implementation (and a select few verify some parts of the old implementation), but we would need to re-run essentially the *whole test suite* with it on/off - type adaption is implicitly used in a *large* chunk of the codebase.

slarse:
Alright, I will start diving into this. The cleaned up commit history should help indeed.

Fair warning, this will take me a while. I need to both understand the original code, and the new code.

@MartinWitt Will you also be reviewing this? Only due to the size, I will inevitably miss things, so I think doubling up on the review is preferable.

slarse:
> I also added a feature toggle (and deprecated it). I'd be interested in your opinion on how to _test this_. Currently the tests use the new implementation (and a select few verify some parts of the old implementation), but we would need to re-run essentially the _whole test suite_ with it on/off - type adaption is implicitly used in a _large_ chunk of the codebase.

Without looking at any of the implementation, my spontaneous thought would be to just add another job to the build matrix in CI that runs the legacy implementation.

I-Al-Istannen:
> This could be the largest performance patch in spoons history, so yes.

As a word of caution: It likely won't match the regex optimization @SirYwell cooked up while ago (which was the largest bottleneck of my indexer at the time :P) and the gains depend on how heavily type adaption is used in your program. Anything querying methods or doing override/inheritance/subtype checks should receive a significant boost though. My indexer is particularly heavy on that, as it needs to query all inherited and declared executables for *every class* it explores.

I should probably look into building a performance test suite again, but after trying to properly benchmark a compiler we were writing I don't really have much hope left that running in Github actions can provide meaningful results *at all*.

----

> Fair warning, this will take me a while. I need to both understand the original code, and the new code.

Understandable :) I tried to document it and ease the burden but it is still quite a large change. I also didn't completely understand every facet of the original implementation as I got lost a lot in there. The code in this PR was mostly built by looking at the JLS, existing tests, building new tests and diffing the output of my JDK indexer. This might fall onto the wrong side of Chesterton's fence but it is hopefully okay.

----

> Without looking at any of the implementation, my spontaneous thought would be to just add another job to the build matrix in CI that runs the legacy implementation.

Hm, alright. Doubling the number of jobs sounds a bit scary CI-time wise, maybe we can get away with only running it for one Java version.

slarse:
> I should probably look into building a performance test suite again, but after trying to properly benchmark a compiler we were writing I don't really have much hope left that running in Github actions can provide meaningful results at all.

It probably can't. Jenkins is a better choice, but it's still imperfect. One would really want bare-metal machine for that.

> I  tried to document it and ease the burden but it is still quite a large change. 

I have started looking at it, and the documentation does help! There's just a lot to wrap my head around  regardless :)

> Hm, alright. Doubling the number of jobs sounds a bit scary CI-time wise, maybe we can get away with only running it for one Java version.

I wasn't particularly clear here, but what I meant was to add a single job to the matrix, not add a new variable to it. For example, running only on Linux with the oldest supported version of Java (so, 11).

doc/type_adaption.md
```diff
@@ -0,0 +1,276 @@
...
+
+Alas, this is still not enough. Consider this:
+```java
+interface Foo<T> {
+  void foo(T t);
+}
+interface SubFoo extends Foo<String> {
+  void foo(String t);
+}
+```
```
slarse, line 184: I'm a bit confused here. This section is talking about methods that:

> *conflict*, i.e. whether you can not declare *both* in the same type.

But the above example is fine, it's just an override. What precisely do you mean by "conflict"?

slarse (commented):
@I-Al-Istannen  I've now carefully read through your explaining doc again and agree with both assumptions and conclusions. I just had one question about what "conflicting" means.

In addition, I'm wondering if you've looked at what facilities exist in JDT? It seems like _all_ of this must be performed somewhere in JDT (it's a compiler after all). Is there any way to leverage JDT more so we don't have to do so much heavy lifting?

MartinWitt:
> In addition, I'm wondering if you've looked at what facilities exist in JDT? It seems like all of this must be performed somewhere in JDT (it's a compiler after all). Is there any way to leverage JDT more so we don't have to do so much heavy lifting?

From my understanding these methods exist, see https://git.eclipse.org/c/jdt/eclipse.jdt.core.git/tree/org.eclipse.jdt.compiler.apt/src/org/eclipse/jdt/internal/compiler/apt/model/TypesImpl.java#n519. But if we change the AST, we don't have access to JDT elements and their API any more. For example, you create a new `CtMethod` and want to check if it overrides an existing method. This is currently not possible using JDT methods, because we simply don't have the JDT element any more. I'm unsure how easily we could convert a `CtMethod` back to the JDT counterpart.

doc/type_adaption.md
I-Al-Istannen: This definition:
> It is sometimes necessary to find out whether two methods *conflict*, i.e. whether you can not declare *both* in the same type.

The Javadoc for the method is probably a bit more helpful:
```java
	 * Checks if two given methods are conflicting, i.e. they can not both be declared in the same
	 * class. This happens if the erasure of the methods is the same or one overrides the other. This
	 * method is used to remove methods that were already visited or were overwritten/shadowed by a
	 * subclass in various places.
```
It's currently used in the metamodel classes to figure out if something was inherited (which seems like a weird way to do that, but I didn't touch the Metamodel).  
It is also used in `getAllMethods` to reject duplicated methods (which presumably can occur if a class is visited twice or a superclass method is found after an overriding one in the subclass).

src/main/java/spoon/support/adaption/TypeAdaptor.java
```diff
@@ -0,0 +1,598 @@
...
+	public CtTypeReference<?> adaptType(CtTypeReference<?> superRef) {
+		if (superRef.getFactory().getEnvironment().useOldAndSoonDeprecatedClassContextTypeAdaption()) {
+			if (startMethod != null) {
+				return new MethodTypingContext()
+					.setClassTypingContext(getOldClassTypingContext())
+					.setMethod(startMethod)
+					.adaptType(superRef);
+			}
+			return getOldClassTypingContext().adaptType(superRef);
+		}
```
slarse, line 428: For all methods that have a legacy alternative, I think it's a good thing to minimize the noise. So, something like this:

```suggestion
	public CtTypeReference<?> adaptType(CtTypeReference<?> superRef) {
		if (useLegacyTypeAdaption(superRef)) {
			return legacyAdaptType(superRef);
		}
```

And then have a helper `useLegacyTypeAdaption` with the `SuppressWarnings` on it. Even if the code in the legacy implementation is a one-liner, I think it helps to separate it into a method of its own just for clarity.

src/main/java/spoon/support/adaption/TypeAdaptor.java
```diff
@@ -0,0 +1,598 @@
...
+	 */
+	@SuppressWarnings("removal")
+	public boolean isSubtypeOf(CtTypeReference<?> superRef) {
+		if (superRef.getFactory().getEnvironment().useOldAndSoonDeprecatedClassContextTypeAdaption()) {
+			return getOldClassTypingContext().isSubtypeOf(superRef);
+		}
+		// We have no declaration so we can't really do any subtype queries
+		if (hierarchyStart == null) {
+			return false;
+		}
```
slarse, line 95: Is it wise for `null` to be allowed for `hierarchyStart`? It seems like nonsense.

src/main/java/spoon/support/adaption/TypeAdaptor.java
```diff
@@ -0,0 +1,598 @@
...
+	 * @return true if base extends/implements the super type
+	 */
+	@SuppressWarnings("removal")
+	public static boolean isSubtype(CtType<?> base, CtTypeReference<?> superRef) {
+		if (base.getFactory().getEnvironment().useOldAndSoonDeprecatedClassContextTypeAdaption()) {
+			return new TypeAdaptor(base).isSubtypeOf(superRef);
+		}
+		String superRefFqn = superRef.getTypeErasure().getQualifiedName();
+
+		// Everything inherits from object
```
slarse, line 129:
```suggestion
```

src/main/java/spoon/support/adaption/TypeAdaptor.java
```diff
@@ -0,0 +1,598 @@
...
+		if (superRef.getQualifiedName().equals("java.lang.Object")) {
+			return true;
+		}
+
+		// Types are subtypes of themselves
```
slarse, line 134:
```suggestion
```

src/main/java/spoon/support/adaption/TypeAdaptor.java
```diff
@@ -0,0 +1,598 @@
...
+			CtTypeReference<?> next = workQueue.poll();
+			if (next.getQualifiedName().equals(superRefFqn)) {
+				return true;
+			}
+
+			if (next.getSuperclass() != null) {
+				workQueue.add(next.getSuperclass());
+			}
+			workQueue.addAll(next.getSuperInterfaces());
+		}
```
slarse, line 153: This seems like a very specific thing. As you want to comment _what_ this does, I'd suggest making it a method of its own.

slarse (changes requested):
@I-Al-Istannen So I've gotten around to the implementation now. I've kind of wrapped my head around `TypeAdaptor` and have some preliminary comments to clean it up a bit. So far, I've not found any functional problems. But overall it's looking very good, I must say, the code is of very high quality.

I'm not entirely done with `TypeAdaptor` and I'll want to give it another pass, but overall it looks very good to me.

src/main/java/spoon/support/adaption/TypeAdaptor.java
slarse: I would also suggest renaming the flag to `useLegacyTypeAdaption`.

slarse:
I intend to have a look at the node classes next, I think that will kind of wrap up my first pass

src/main/java/spoon/support/adaption/TypeAdaptor.java
slarse: I would say these comments are superfluous, but if you really want to keep them, put them _inside_ the if-statement they are related to. That connects them more strongly to the statement they relate to.

src/main/java/spoon/support/adaption/TypeAdaptor.java
I-Al-Istannen: The context can be initialized from a reference to a type not on the classpath. In that case we don't have a corresponding declaration. The current implementation just rejects such queries as they are rather pointless. I will amend the comment, do you have a better idea for handling this situation? We could store an optional but I am not sure whether that's strictly better.

src/main/java/spoon/support/adaption/TypeAdaptor.java
I-Al-Istannen: I have deleted a few and moved a few more into ifs. I hope the ones still left have enough merit to earn their stay.

src/main/java/spoon/support/adaption/TypeAdaptor.java
I-Al-Istannen: I tried, but coming up with a descriptive name was non trivial. I am not perfectly happy with what I ended up with, but it hopefully works. As always, I am open for suggestions though :)

I-Al-Istannen:
Thank you for the review! I was away for a few days and didn't have a chance to look at it before today. I think I addressed your concerns but I will leave the conversations for you to resolve if you are satisfied with the changes :)

----

Qodana complains about the deprecated method *at its declaration*. I can add a `@SuppressWarnings` there to placate it, but that strikes me as odd.

slarse:
> Thank you for the review! I was away for a few days and didn't have a chance to look at it before today. I think I addressed your concerns but I will leave the conversations for you to resolve if you are satisfied with the changes :)

No worries about that, this week has been very busy for me after work (with a lot of after works...) so I've had minimal time for open sourcering anyway :). I will get back to this on Sunday.

> Qodana complains about the deprecated method _at its declaration_. I can add a `@SuppressWarnings` there to placate it, but that strikes me as odd.

I think that's intentional, in the sense "hey you deprecated this method is but you're still using it in the same code base in which it is defined". @MartinWitt can perhaps weigh in on it.

slarse:
@I-Al-Istannen I think your changes look good and overall I'm pretty happy with the interface of TypeAdaptor.

I'm currently considering the node classes and AdaptationVisitor. I've mostly wrapped my head around what's going on and I think the solution is sound, but I think there is some amount of complexity here that may not be necessary. For one, the `Node` class and its subclasses hint at an arbitrary tree-like structure, but the tree that's actually built is far from arbitrary: GlueNodes always have DeclarationNodes as children, and vice versa. I feel like more of that could be encapsulated into types, and I think it might be better not to let `GlueNode` and `DeclarationNode` extend the same parent class.

I need to give this some more thought, though, so I don't have any concrete feedback right now.

I-Al-Istannen:
@slarse 
> I'm still working on understanding the ins and outs of this :)
Now, I think I'm pretty close to grasping everything. I've been poking around for hours, rewriting code, trying to understand what makes everything tick. It's rather hard to just eyeball everything and understand it, but I'd say I'm pretty close now.

I feel *slightly* guilty here, especially after I just reworked the tree building implementation... Thank you though :)

> 1.

Yea, that is correct and absolute garbage now that I think about it further - it can't handle rawtypes properly! If you have a class extending another as a raw type, the type parameter of the parent should be translated to `Object`!

> 3.

That is indeed a relict. Looking at the history it has existed for so long that I have forgotten why it was needed in the first place. Maybe it was carried along when I extracted the tree structure out of the TypeAdaptor class, but I am not sure.

----

Thanks for those comments, they were really helpful and made me realize that I do not properly handle rawtypes. I fixed that now and in the process I separated the two node types into distinct classes and now enforce the back-and-forth of them at type level.

----

I also need to add contracts to the new tests and maybe clean them up a bit, but that is something for later.

src/main/java/spoon/support/adaption/NodePrintHelper.java
```diff
@@ -0,0 +1,39 @@
...
+	}
+
+	/**
+	 * Converts a list of values to a crudely formatted JSON-y array string.
+	 * Special-purpose helper for {@link DeclarationNode} and {@link GlueNode}.
+	 *
+	 * @param input the input list
+	 * @return the list as a quoted array
+	 */
+	public static String quote(List<?> input) {
```
slarse, line 31: The method name is not sufficiently descriptive imo, and I generally think that overloads should be used exceedingly sparingly. Ought to be something like:

```suggestion
	public static String toJSONArray(List<?> input) {
```

or

```suggestion
	public static String toStringifiedJSONArray(List<?> input) {
```

slarse (commented):
@I-Al-Istannen So, I was going to handle this on my vacation, but then I figured I'd actually go on vacation for real and not work even on open source. So I played video games for two weeks instead.

Anyhow, I think this is very close to mergeable now. You handled my comments well, and in rewriting the tree building you also removed a lot of stuff that I was going to comment on.

In terms of the code I just have some minor suggestions and a question on the fixme.

I'm also thinking, could some part of this be considered breaking as the old type adaptor stuff is all public? I'm thinking not since it's all still there and usable, but we should have another think about it.

Next step I think is for @MartinWitt to browse through this. I don't have anything more to add.

slarse:
Oh, and one or more files are missing the license header, could you add it? There's a command to add it with the Maven plugin (I forget what it is), but copy-pasting also works.

src/main/java/spoon/support/adaption/NodePrintHelper.java
I-Al-Istannen: It isn't a general-purpose Json-ify method, as it does not handle embedded quotes and control char escaping and so on. It just wraps quotes around the input which is enough to turn class names into valid Json. I only made it mostly Json-like so editors can neatly fold, highlight and organize the output.

I've renamed it and added a further disclaimer in the doc.

src/main/java/spoon/support/reflect/declaration/CtTypeParameterImpl.java
```diff
@@ -244,15 +244,22 @@ public boolean isSubtypeOf(CtTypeReference<?> superTypeRef) {
 		if (superTypeRef instanceof CtTypeParameterReference) {
 			//the type is type parameter too. Use appropriate sub type checking algorithm
 			CtTypeParameter superTypeParam = (CtTypeParameter) superTypeRef.getDeclaration();
-			return isSubtypeOf(getFactory().Type().createTypeAdapter(getTypeParameterDeclarer()), this, superTypeParam);
+			// FIXME: Method parameters
```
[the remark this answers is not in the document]
I-Al-Istannen: Oh *god*, please no. That was trying to remind me that generic parameters on methods are legal which I already handled, but I was also reminded that *this* is legal as well:
```java
public class Fpp<R> {
	public <T> FooT t) {
		new <String>RawBottom<String>("hello");
	}
}
```
I have never seen a generic constructor (or the call of one :^) in real code and it has *zero* coverage in the current Spoon test suite. Nevertheless, that would throw a ClassCastException. I added support and tests for it now, but I am not sure how elegant the solution is. I don't think it's *too* bad, but your mileage may vary.

I-Al-Istannen:
> So, I was going to handle this on my vacation, but then I figured I'd actually go on vacation for real and not work even on open source. So I played video games for two weeks instead.

That sounds nice! I hope it went well and was rejuvenating :)

> In terms of the code I just have some minor suggestions and a question on the fixme.

Yea, that was another corpse I accidentally buried apparently. Thank you, your reviews dug up quite a few landmines I'd missed or not considered deeply enough...


> I'm also thinking, could some part of this be considered breaking as the old type adaptor stuff is all public? I'm thinking not since it's all still there and usable, but we should have another think about it.

The old stuff is public and untouched but it is no longer used by default. Any breaking change is a bug, but that *might* happen of course (as we've seen with the last fixme. That could have blown up somebody's code if they have constructors declaring generic type parameters and do `instanceOf` checks on them against other type parameter references).  
Apart from that, I am not quite sure what the best course of action is. The main risk are bugs in the new implementation that (more or less) subtly change the outcome but I am not sure that really counts.

src/main/java/spoon/support/adaption/DeclarationNode.java
```diff
@@ -0,0 +1,144 @@
...
+
+		this.lowerNodes = new HashSet<>();
+	}
+
+	/**
+	 * Adds a glue node as a child of this declaration node.
+	 *
+	 * @param node the node to add as a child
+	 */
+	public void addLower(GlueNode node) {
```
slarse, line 44:
```suggestion
	public void addChild(GlueNode node) {
```

src/main/java/spoon/support/adaption/DeclarationNode.java
```diff
@@ -0,0 +1,144 @@
...
+
+	private final List<CtTypeParameter> formalArguments;
+	private final CtTypeReference<?> inducedBy;
+	private final Set<GlueNode> lowerNodes;
+
+	protected DeclarationNode(CtTypeReference<?> inducedBy) {
+		this.inducedBy = inducedBy;
+		this.formalArguments = inducedBy.getTypeDeclaration().getFormalCtTypeParameters();
+
+		this.lowerNodes = new HashSet<>();
```
slarse, line 36:
```suggestion
		this.children = new ArrayList<>();
```

src/main/java/spoon/support/adaption/GlueNode.java
```diff
@@ -0,0 +1,122 @@
...
+import java.util.stream.Collectors;
+
+import static spoon.support.adaption.NodePrintHelper.quote;
+import static spoon.support.adaption.NodePrintHelper.toJsonLikeArray;
+
+class GlueNode {
+
+	private final List<CtTypeReference<?>> actualArguments;
+	private final CtTypeReference<?> inducedBy;
+	private final Set<DeclarationNode> lowerNodes;
```
slarse, line 29: Same comment as on `DeclarationNode`

```suggestion
	private final List<DeclarationNode> children;
```

src/main/java/spoon/support/adaption/GlueNode.java
```diff
@@ -0,0 +1,122 @@
...
+	/**
+	 * Creates a new glue node for the given reference.
+	 *
+	 * @param inducedBy the reference this node is built from
+	 */
+	GlueNode(CtTypeReference<?> inducedBy) {
+		this.inducedBy = inducedBy;
+		this.actualArguments = inducedBy.getActualTypeArguments();
+
+		this.lowerNodes = new HashSet<>();
```
slarse, line 40:
```suggestion
		this.children = new ArrayList<>();
```

src/main/java/spoon/support/adaption/GlueNode.java
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	/**
+	 * Adds a declaration node as a child of this glue node.
+	 *
+	 * @param node the node to add as a child
+	 */
+	public void addLower(DeclarationNode node) {
```
slarse, line 48:
```suggestion
	public void addChild(DeclarationNode node) {
```

slarse (changes requested):
Some final nitpicks on `GlueNode` and `DeclarationNode`. After that I've really done everything I can to validate this PR.

I think @MartinWitt and/or @monperrus should also have at least a quick browse through before we merge this, but I'm personally pretty convinced by the tests and the new code. I think we should aim at merging next week.

slarse:
Oh, and thanks for the massive amount of work here @I-Al-Istannen. This must have been a good few hours of work, considering just reviewing it has been... a good few hours of work :)

src/main/java/spoon/support/adaption/DeclarationNode.java
```diff
@@ -0,0 +1,144 @@
...
+import java.util.stream.Collectors;
+
+import static spoon.support.adaption.NodePrintHelper.quote;
+import static spoon.support.adaption.NodePrintHelper.toJsonLikeArray;
+
+class DeclarationNode {
+
+	private final List<CtTypeParameter> formalArguments;
+	private final CtTypeReference<?> inducedBy;
+	private final Set<GlueNode> lowerNodes;
```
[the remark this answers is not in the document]
I-Al-Istannen: > but as far as I can tell, the same (identity-wise and content-wise) node cannot be added twice anyway

That is precisely the reason why I made it a set. I don't need any order, I don't care about index based access and I do not intend do ever add duplicates. It just sounds like a good usecase for a set to me.

In its current form it does not guard against creating multiple nodes for the same reference as you noticed, so this is mostly of symbolic value. I am not sure I agree with it, but I will change it to a list.

src/main/java/spoon/support/reflect/declaration/CtTypeParameterImpl.java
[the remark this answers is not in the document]
I-Al-Istannen: > Buuut, shouldn't the FIXME be removed now?

Okay, that's embarrassing :P Fixed.

src/main/java/spoon/support/adaption/DeclarationNode.java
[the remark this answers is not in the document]
slarse: > That is precisely the reason why I made it a set. I don't need any order, I don't care about index based access and I do not intend do ever add duplicates. It just sounds like a good usecase for a set to me.

I would have to disagree on that. Using a set indicates that uniqueness (and consequently deduplication of "careless" insertions) is semantically important.

Using an explicitly identity-based set here would be mildly misleading as it hints that there is some form of importance to the object identities (which there is not), but it would at least be clear that the set operates on identity.

Using a `HashSet` might be significantly misleading as someone coming in later to edit the code might assume that the objects contained in the `HashSet` have some concept of equality beyond identity. I would.

There's also the fact that the primary use case for the child list is to iterate over it. Operations on a `HashSet` (most prominently inserting and iterating) are several times slower than on an `ArrayList`. Compounding all of that, I think a list is a much better choice.

> In its current form it does not guard against creating multiple nodes for the same reference as you noticed, so this is mostly of symbolic value.

This is precisely what is misleading to me.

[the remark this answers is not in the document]
I-Al-Istannen: > Using a set indicates that uniqueness (and consequently deduplication of "careless" insertions) is semantically important

I agree with the first part and that is the case here, bit I never considered the part in parenthesis to be of relevance when picking a collection. Interesting, I will keep that in mind when choosing something next time.

A general `Collection` randomly instantiated to an `ArrayList` (as that is the "default" collection in most cases) might be a better solution then, but it also isn't really important. It is declared as a `List` now, that should be good enough. Most people do not necessarily think "is in a certain order" when they see `List` as much as they think "must be unique" when seeing a `Set`.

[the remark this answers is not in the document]
slarse: > A general Collection randomly instantiated to an ArrayList

Yes, I agree that would be even better in flagging the the point of "this is just a collection with no particular characteristics to rely on". If I had thought of that I would have made that my suggestion :)

slarse (approved):
From my perspective, this is now good for merge. I would however like @MartinWitt and/or @monperrus to at least do a sanity check due to the magnitude of the changes.

Thanks @I-Al-Istannen for sticking in there!

I-Al-Istannen:
Okay, I lied. After your affirmative comment I did quickly change the `List` to a `Collection` :) I hope that is a small enough change to not invalidate the approval.

> This must have been a good few hours of work, considering just reviewing it has been... a good few hours of work :)

That is one way to put it :D It seems like it was worth it at least and I hope it will make the life of somebody else in the long term a bit easier.

> From my perspective, this is now good for merge. Thanks @I-Al-Istannen for sticking in there!

Sounds good :) This PR took a while as it is was rather large and would definitely be worse off without your feedback. Thank you for sticking with it as well :)

slarse (approved):
Just for the sake of propriety, another approve on the final changes.

:tada:

monperrus:
Fully agree, thanks a lot for the hard work put here! :100:

src/main/java/spoon/compiler/Environment.java
```diff
@@ -583,6 +583,22 @@ public interface Environment {
 	 */
 	void setPrettyPrinterCreator(Supplier<PrettyPrinter> creator);
 
+	/**
+	 * Whether to use the new or legacy and (soon to be) deprecated type adaption. You should not use this method.
+	 *
+	 * @return true if spoon uses the old and (soon to be) deprecated type adaption.
+	 */
+	@SuppressWarnings("DeprecatedIsStillUsed")
+	@Deprecated(forRemoval = true, since = "10.0.1")
```
MartinWitt, line 592: Change this to the current version please.

src/main/java/spoon/compiler/Environment.java
```diff
@@ -583,6 +583,22 @@ public interface Environment {
...
+	boolean useLegacyTypeAdaption();
+
+	/**
+	 * Sets whether to use the new or legacy and (soon to be) deprecated type adaption. You should not use this method.
+	 *
+	 * @param useLegacyTypeAdaption whether to use the old type adaption implementation
+	 */
+	@Deprecated(forRemoval = true, since = "10.0.1")
```
MartinWitt, line 600: Please update this to the most recent version.

src/main/java/spoon/metamodel/MMMethod.java
```diff
@@ -10,11 +10,11 @@
 import spoon.SpoonException;
 import spoon.reflect.declaration.CtMethod;
 import spoon.reflect.reference.CtTypeReference;
-import spoon.support.visitor.MethodTypingContext;
 
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import spoon.support.adaption.TypeAdaptor;
```
MartinWitt, line 17: Please fix the import order to the old style.

src/main/java/spoon/metamodel/Metamodel.java
```diff
@@ -41,8 +41,8 @@
 import spoon.reflect.visitor.filter.TypeFilter;
 import spoon.support.DefaultCoreFactory;
 import spoon.support.StandardEnvironment;
+import spoon.support.adaption.TypeAdaptor;
```
MartinWitt, line 44: Please fix import order

src/main/java/spoon/support/adaption/NodePrintHelper.java
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * Spoon is available either under the terms of the MIT License (see LICENSE-MIT.txt) of the Cecill-C License (see LICENSE-CECILL-C.txt). You as the user are entitled to choose the terms under which to adopt Spoon.
+ */
+package spoon.support.adaption;
+
+import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+final class NodePrintHelper {
```
MartinWitt, line 14: Add a short sentence with the information that this an utility class, and it's provided API.
As this is a package private final class, you don't need to write a lot of text.

MartinWitt (approved):
A minor doc issue aside, this is an excellent PR. Thank you for your time @I-Al-Istannen.

MartinWitt:
Looks like all integrators are happy with this PR, so after @I-Al-Istannen changes the import thing we can merge it finally.

I-Al-Istannen:
Thanks, adjusted :) My IDE has a different (grouped) sorting order.

slarse:
If no one objects, I will merge tonight.

@monperrus @MartinWitt

slarse (approved):
Thanks @I-Al-Istannen, massive work that finally gets merged. Cheers!

monperrus:
Excellent!

I-Al-Istannen:
Thank you, that was a long journey :) I will make sure to address any potential fallout once it is in a public release, though I do hope that won't be needed. Changes will hopefully be a bit easier, as I think the new structure is more understandable than the previous solution - I am biased though ;)

My own profiles also look a lot nicer now (I guess I need to have a look at Javadoc handling next...) and maybe yours will be a bit faster as well :) If you query methods a lot, I'd expect quite some improvements based on my testing.

PR #4222, opened 2021-10-12
fix: support modifiers in annotation methods

I-Al-Istannen:
I noticed that annotation methods did not have any modifiers previously, which was probably a mistake. This patch just adds them to annotation methods.

The logic is now duplicated for methods and annotation methods but I didn't find a good place to share it. The Helper only has `create` methods and the builder only `visit` and I didn't want to be the first one to break that order.


monperrus:
Thanks a lot @I-Al-Istannen. Feel free to improve the tests in a subsequent PR :)

PR #4228, opened 2021-10-15
feat: add CtAnnotation#getName

monperrus:
Just wasted some time to find the useful method I was sure to find.

So I make this hallucination real.


monperrus:
Suggestion applied, CI now runs with checkstyle-javadoc

PR #4229, opened 2021-10-16
chore: Add MissingJavadocMethod checkstyle module to Javadoc regression test

slarse:
As discussed in https://github.com/INRIA/spoon/pull/4228#discussion_r730278058, we should also check for missing Javadoc. Currently, one can get around the Javadoc regression test by simply not adding a Javadoc comment, or by misplacing it (e.g. in between method annotation and method header).


slarse:
Ping @monperrus

PR #4237, opened 2021-10-19
feature: Support of shift left and right operation inside VisitorPartialEvaluator

thimo-seitz:
Support for evaluation of shift left and right operations:
- `1L << 53`
- `8 >> 2`

I've found no matching issue for this.


slarse:
@monperrus Hold that thought, I want to have a look :)

PR #4244, opened 2021-10-21
Review: add qodana analyzer

MartinWitt:
This PR adds Qodana as static code analyzer and fails if a PR adds a bad smell. All new bad smells are printed in the CI log. We need 2 qodana runs, because first we create a baseline with master-branch and then for the PR branch.  
There is a list of all possible inspections [https://www.jetbrains.com/help/idea/list-of-java-inspections.html](https://www.jetbrains.com/help/idea/list-of-java-inspections.html). @I-Al-Istannen had a list of checks he was using, which is a good starting point. Are there any inspections in the list you want to activate?

Here is an example output of a found bad smell. [https://github.com/INRIA/spoon/runs/4107300966?check_suite_focus=true](https://github.com/INRIA/spoon/runs/4107300966?check_suite_focus=true)


monperrus:
nice! What happens if we have a new violation? a failing build? a message on the PR?

MartinWitt:
> What happens if we have a new violation?

The build fails and in the CI step the problem with location is printed. See [https://github.com/INRIA/spoon/runs/4107300966?check_suite_focus=true](https://github.com/INRIA/spoon/runs/4107300966?check_suite_focus=true) for example output.

monperrus:
Nice. However, in https://github.com/INRIA/spoon/runs/4107300966?check_suite_focus=true,

```
Run VIOLATIONS=$(jq '.runs | .[0].results | map(select (.baselineState == "new")) | length' /home/runner/work/spoon/spoon/result/qodana.sarif.json)
  VIOLATIONS=$(jq '.runs | .[0].results | map(select (.baselineState == "new")) | length' /home/runner/work/spoon/spoon/result/qodana.sarif.json)
  if [ "$VIOLATIONS" -gt "0" ]; then exit 1; fi
  shell: /usr/bin/bash -e {0}
  env:
    JAVA_DISTRIBUTION: temurin
    MAVEN_OPTS: -Dhttp.keepAlive=false -Dmaven.wagon.http.pool=false
    JAVA_HOME: /opt/hostedtoolcache/Java_Temurin-Hotspot_jdk/16.0.2-7/x64

```


 I don't see the actual violation. Where it is?

MartinWitt:
In the step `print Problems`

monperrus:
ah yes. it's a little counter-intuitive that the problems are printed before the failing step instead of inside the failing step, at the point of failure. any option to do that?

slarse:
The problems are incredibly hard to read (JSON is  not really meant for human consumption).

It seems to me like all we really need to see for each problem is the startline and the explanatory text, possibly with a tiny bit of extra metainfo such as severity. Could you perhaps select only those things into the failing step, and keep `Print problems` as is?

MartinWitt:
How about the new `print problem(short)` step?

monperrus:
it's nice!

MartinWitt:
@I-Al-Istannen had some more JQ magic and now the output should be readable for humans.

slarse:
@MartinWitt new output looks good to me

I think it's about time you revert that smelly little snippet of code and get this PR merged!

@monperrus You know how I pondered the problem of how to avoid posting the _same_ code style comment over and over again in a PR? The LGTM comments show what it looks like when you don't ponder that problem...

monperrus:
> @monperrus You know how I pondered the problem of how to prevent an SE code quality bot from posting the same code correction comment over and over again in a PR? The LGTM comments show what it looks like when you don't ponder that problem...

Indeed, not nice. What's the status in flacocobot @dginelli @andre15silva @martinezmatias @danglotb? Can we make sure we don't pollute evolving PRs with the same messages again and again?

dginelli:
Hi @monperrus,

Based on the current implementation, flacocobot works as follows.

Assume that flacocobot starts now (November, 10th) to inspect project X.

At the first iteration, it searches for the open failed pull requests created in the last 7 days (so, between November, 3rd and November, 10th). If it finds any suspicious lines, it adds a review comment with the 5 most suspicious lines.

For the next iterations (every 15 minutes), flacocobot searches for new open failed pull requests or for updates in the other pull requests created after November, 3rd. If it detects an update in one of these pull requests (it does not consider updates related to the addition of new comments or review comments), it will add a new review comment on the pull request. The content will be similar, the only difference is associated with the lines identified by flacoco.

If we decide to set a maximum of review comments per pull request that flacocobot can add, we should keep the information to avoid that flacocobot runs again on a certain pull request.

slarse:
@dginelli It's not about the amount of review comments, but about repeated comments. In this PR, LGTM repeated the same comment over and over again, even though @MartinWitt didn't even touch that file in the related commits. That's because LGTM compares with the main branch, disregarding any comments it's already posted.

A more user-friendly solution would be to compare with the main branch only if there are no prior review comments (by the bot). If the bot has already posted review comments, it should compare with the commit on which it last posted comments.

That comparison can be more or less advanced, but a bare minimum effort would be to not post review comments on/related to files that haven't changed since last time the bot commented.

I don't know if this applies to flacoco, but in terms of LGTM that's what they ought to do. Perhaps it's something we can configure, even?

This is all very off topic, though, this PR is done and should be merged. Thanks @MartinWitt!

slarse (approved):
We'll try this out and see how it goes.

@monperrus On a related note, I vote for discontinuing LGTM, it's annoying me more than a little bit. I prefer something like this where we get all of the info locally and can tweak it however we like.

MartinWitt:
Now it's too late but can you keep all authors while squashing? @I-Al-Istannen did the JQ magic.

slarse:
Hmm, I did not see a `co-authored by` tag. Must have missed among the liberal amount of commits.

We can revert and redo to put the record straight.

Issue #4219, opened 2021-10-12
SniperPrettyPrinter - Abstract keyword getting deleted


abhi22493:
When using SniperPrettyPrinter, for the below input the expected output is not generated. Transformation is working as expected but abstract keyword is getting deleted from another method of the same class.

Input :
`public @interface CtAnnotationMethodProcessor {
	String totals();
	abstract String subjectName() default "CSE";
}`

Expected :
`public @interface CtAnnotationMethodProcessor {
	int totals();
	abstract String subjectName() default "CSE";
}`

Actual :
`public @interface CtAnnotationMethodProcessor {
	String totals();
	 String subjectName() default "CSE";
}`

Spoon print config :
launcher.getEnvironment().setPrettyPrinterCreator(() -> new SniperJavaPrettyPrinter(launcher.getEnvironment()));

Spoon version : 9.1.0-beta-10

nharrand:
Hello @abhi22493 ,

Can you provide the code of your transformation? I can't reproduce the bug. Simply using `setType(CtTypeReference ref)`  and calling Sniper pretty printer print the abstract modifier  as expected for me.

PR #4263, opened 2021-11-02
test(TestSniperPrinter): make the method testSniper able to handle non class types

nharrand:
This fix allows testing the SniperPrettyPrinter with types that are not classes (Interfaces, AnnotationTypes, etc). (As for example #4219)


monperrus:
LGTM, will merge. thanks @nharrand

PR #4272, opened 2021-11-06
fix: force execution of `mvn -version` in CMD in case of Windows

algomaster99:
Reference: https://github.com/SpoonLabs/sorald/issues/612

@slarse, I think the scheduled tests were failing because the "Run tests" step on windows was failing. For [other platforms](https://github.com/SpoonLabs/sorald/runs/4126994132?check_suite_focus=true#step:11:535), that step passed.

I have also tried to execute tests on Sorald on Windows after using the edited spoon build with the change in this PR and all tests passed :heavy_check_mark: .

EDIT: I saw the tests for macOS and Ubuntu pass in one of the workflows where they ran before the Windows' tests. I am not sure how to run all jobs for an individual OS in the matrix.

### DIFF
--- a/src/main/java/spoon/support/compiler/SpoonPom.java
+++ b/src/main/java/spoon/support/compiler/SpoonPom.java
@@ -525,6 +525,9 @@ public class SpoonPom implements SpoonResource {
 		try {
 			String[] cmd;
 			if (System.getProperty("os.name").contains("Windows")) {
+				// We launch a subshell in case of Windows because the subprocess below is unable to look up the PATH
+				// on Windows.
+				// See https://github.com/SpoonLabs/sorald/runs/4193699004?check_suite_focus=true#step:8:11.
 				cmd = new String[]{"cmd", "/c", "\"mvn -version\""};
 			} else if (System.getProperty("os.name").contains("Mac")) {
 				cmd = new String[]{"sh", "-c", "mvn -version"};
--- a/src/main/java/spoon/support/compiler/SpoonPom.java
+++ b/src/main/java/spoon/support/compiler/SpoonPom.java
@@ -525,7 +525,7 @@ public class SpoonPom implements SpoonResource {
 		try {
 			String[] cmd;
 			if (System.getProperty("os.name").contains("Windows")) {
-				cmd = new String[]{"mvn.cmd", "-version"};
+				cmd = new String[]{"cmd", "/c", "\"mvn -version\""};
 			} else if (System.getProperty("os.name").contains("Mac")) {
 				cmd = new String[]{"sh", "-c", "mvn -version"};
 			} else {


slarse:
Hi @algomaster99,

I do not believe that there's a problem in Spoon. If `mvn.cmd` can't be found, most likely it's not available on the path. Spoon's Windows tests are passing fine without this change.

Launching a subshell is bad practice due to the risk of injection attacks (I see that the Mac version is doing that for some reason, we should change that). There's really no such risk here as there's no user input, but I don't see any reason we should need to launch a subshell.

Check the system path of the affected build and ensure that it includes Maven first.

algomaster99:
> Check the system path of the affected build and ensure that it includes the Maven bin directory.

I thought I should do this, however, the workflow is [able to execute](https://github.com/SpoonLabs/sorald/runs/4123271163?check_suite_focus=true#step:11:1) `mvn tests` so it has to be in `PATH`.

slarse:
> > Check the system path of the affected build and ensure that it includes the Maven bin directory.
> 
> I thought I should do this, however, the workflow is [able to execute](https://github.com/SpoonLabs/sorald/runs/4123271163?check_suite_focus=true#step:11:1) `mvn tests` so it has to be in `PATH`.

There are different paths in Windows. As I don't use Windows often, I don't know this by heart, but I _think_ that when you run a subprocess from Java, it only does binary lookup on the system path. When you instead launch a full subshell, the shell itself probably resolves the full process path (user path + system path, perhaps some other path I don't know about), which is also what happens when you just run `mvn test`.

You could go down that avenue and investigate the details, but I'm pretty sure that if a binary is on the system path, it's "globally" discoverable. The user path is not always loaded.

@MartinWitt @I-Al-Istannen didn't one of you have a problem with the path in Windows recently? Can you recall the details?

algomaster99:
@slarse Thanks for the inputs! I will try debugging this again.

algomaster99:
> but I think that when you run a subprocess from Java, it only does binary lookup on the system path.

I tried this yesterday by creating a `RunMvnSubprocess` java class in sorald package and executed its main method. It was able to output the path to the executable of maven in Windows.

```java
package sorald;

import java.io.BufferedReader;
import java.io.IOException;
import java.io.InputStreamReader;

class RunMvnSubprocess {
    public static void main(String[] args) throws IOException, InterruptedException {
        String[] cmd = new String[]{"mvn.cmd", "-version"};
        Process p = Runtime.getRuntime().exec(cmd);
        String mavenVersionParsing = "Maven home: ";
        try (BufferedReader output = new BufferedReader(new InputStreamReader(p.getInputStream()))) {
            String line;

            while ((line = output.readLine()) != null) {
                if (line.contains(mavenVersionParsing)) {
                    System.out.println(line);
                }
            }
        }
        p.waitFor();
    }
}
```
```
>>>
Maven home: C:\Users\91730\apache-maven-3.8.1-bin\apache-maven-3.8.1\bin\..

Process finished with exit code 0
 
```

I would also like to bring one more thing to your attention which I discovered yesterday, but did not post it here because I thought it might be unrelated. I noticed that all the 4 tests which failed in the build had one thing in common about them. Each of them ran the subprocess from a `tempDir` created by `@TempDir` annotation provided by Junit5. I didn't explore this avenue much because I thought there should be no issue to execute `mvn -version` form `\C\Users\<name>\AppData\Local\Temp`. I can probably look along these lines too.

monperrus:
Overall, LGTM.

My understanding is that `guessMavenHome` never worked correctly on Windows. Still all tests pass.

@algomaster99  could you add a new test for `guessMavenHome` and ensure it's run by the windows build?

slarse:
> Overall, LGTM.

I strongly disagree, launching an entire shell as a subprocess is not a good solution. We should only execute the Maven binary.

> My understanding is that guessMavenHome never worked correctly on Windows. Still all tests pass.

It worked fine up until a few days ago, and has been working fine for months. In Sorald's CI, that is. We have no reports of it not working elsewhere. I _really_ don't think this is a problem in Spoon, I think it's a problem with paths, or potentially even that the Maven binary has been renamed (but I looked it up and it still seems to be called `mvn.cmd` in the very latest version).

I looked through the logs of the [first build that failed](https://github.com/SpoonLabs/sorald/runs/4087248573?check_suite_focus=true), and it appears that the virtual environment version for Windows has changed in respect to the [prior build](https://github.com/SpoonLabs/sorald/runs/4087248573?check_suite_focus=true) (which passed).

Here's the passing build's virtual environment:

```
Virtual Environment
  Environment: windows-2019
  Version: 20211018.0
  Included Software: https://github.com/actions/virtual-environments/blob/win19/20211018.0/images/win/Windows2019-Readme.md
  Image Release: https://github.com/actions/virtual-environments/releases/tag/win19%2F20211018.0
```

Here's the failing build's virtual environment:

```
Virtual Environment
  Environment: windows-2019
  Version: 20211102.4
  Included Software: https://github.com/actions/virtual-environments/blob/win19/20211102.4/images/win/Windows2019-Readme.md
  Image Release: https://github.com/actions/virtual-environments/releases/tag/win19%2F20211102.4
```

I'm not sure that's it but it's a start.

algomaster99:
> I really don't think this is a problem in Spoon

More debugging has led me to this conclusion too. I think there is a problem with Junit maybe. I will explain why.

I created a simple test case in spoon and sorald and tried executing them on my local machine using Windows.
```java
@Test
public void test() {
    System.out.println(SpoonPom.guessMavenHome());
}
```
This works as expected in spoon, however it STDOUTs `null` in Sorald. And the same piece of code (without test annotation) works well in the main package for both repositories. This led me to think that probably Junit 5 doesn't allow executing scripts _within_ it. I am thinking to question the Junit team about it.

monperrus:
> We should only execute the Maven binary.

if we are able to find it with guarantees yes :)

> launching an entire shell as a subprocess is not a good solution. 

that's only when calling `guessMavenHome` which I expect it to be called rarely and never in hot code

slarse:
> > launching an entire shell as a subprocess is not a good solution.
> 
> that's only when calling `guessMavenHome` which I expect it to be called rarely and never in hot code

It's not about performance, it's about a) knowing _why_ one thing works while the other does not, and b) launching a full subshell opens up a full host of injection attack vectors. Who knows, perhaps at some point someone decides the Maven launcher should take user input, does not sanitize it, and opens up a very unnecessary attack vector.

If it turns out the only way we can make this work reliably on Windows is to launch a subshell, then sure, we'll do that, but only after we actually know it's the only way, and why it's the only way. Since it worked fine until version 20211102.4 of the Windows virtual environment here on GH, I feel fairly confident it's not the only way.

algomaster99:
[The list](https://github.com/actions/virtual-environments/compare/win19/20211102.4...win19/20211018.0) of changes between the two images specified above. Maybe we will have to inspect version changes of software which might have affected this.

slarse:
@algomaster99 Have you checked the output from the subprocess? If there's nothing on stdout, there should be something on stderr.

algomaster99:
@slarse I checked the error stream as well. There is no byte there too. I even tried to call `read` over the stream and it returned `-1` so it confirmed that the error stream has no bytes to read.

I also used `ProcessBuilder`, as suggested in this [answer](https://stackoverflow.com/a/69903411), however I don't get any output in the input stream even after merging the two streams. Moreover, error stream didn't have any bytes without merging too.

algomaster99:
@slarse They answered why the builds are failing.
> your project still uses M2_HOME which was deprecated and removed along with Win 20211102.4 release (more details here: TheCakeIsNaOH/chocolatey-packages#17). Probably you need to redesign your scripts to avoid M2_HOME usage.

I think we always relied on `M2_HOME` for getting the path to `mvn` on Windows and we never really needed to execute the `mvn.cmd -version` command. Since usage of `M2_HOME` is deprecated, we should do two things:
1. Eliminate usage of `M2_HOME` in `MavenLauncer` to get path to the executable.
2. Figure out how to execute Java subprocess in Windows.

slarse:
> @slarse They answered why the builds are failing.
> 
> > your project still uses M2_HOME which was deprecated and removed along with Win 20211102.4 release (more details here: [TheCakeIsNaOH/chocolatey-packages#17](https://github.com/TheCakeIsNaOH/chocolatey-packages/issues/17)). Probably you need to redesign your scripts to avoid M2_HOME usage.
> 
> I think we always relied on `M2_HOME` for getting the path to `mvn` on Windows and we never really needed to execute the `mvn.cmd -version` command. Since usage of `M2_HOME` is deprecated, we should do two things:

Ah, that is quite possible. Thus, when `M2_HOME` was removed, it stopped working.

>     1. Eliminate usage of `M2_HOME` in `MavenLauncer` to get path to the executable.

Sounds reasonable but not urgent, it's just a fallback. What's more urgent is figuring out why `mvn.cmd` is not on the path.

>     2. Figure out how to execute Java subprocess in Windows.

Well, this is not a problem. The current version executes a subprocess. It's just that `mvn.cmd` is not on the (relevant) path.

slarse:
@algomaster99 I did a little bit of debugging, and the Maven bin directory is _not_ on the system path, it's on the user path (and process path), [see here](https://github.com/SpoonLabs/sorald/runs/4192956706?check_suite_focus=true). Like I mentioned [up here](https://github.com/INRIA/spoon/pull/4272#issuecomment-962659893), I bet that Java only checks the system path on Windows.

slarse:
Okay, so it's really pretty simple: Java doesn't do any lookup on the path at ALL when creating a subprocess. We gotta provide the absolute path to the Maven executable. [We could do it like they do for Java Language Server](https://github.com/georgewfraser/java-language-server/commit/40b76143b17f9b0fcd3d09106d621f45f6c61cea).

Here's debugging showing that the absolute path works fine: https://github.com/SpoonLabs/sorald/runs/4193323097?check_suite_focus=true

I can add a fix for this tomorrow.

algomaster99:
> Java doesn't do any lookup on the path at ALL when creating a subprocess.

Some questions about this statement.

Q.1 How is this happening in Linux then? We don't use `M2_HOME` there and I think we get the path to binary of `mvn` by executing `mvn --version`. And we also don't open a subshell for executing this command in case of Linux.

Q.2 Why are there two outputs of `mvn --version` [here](https://github.com/SpoonLabs/sorald/runs/4193323097?check_suite_focus=true#step:8:21)? One comes from executing the `mvn` binary via an absolute path. Where does the other come from if binaries in `mvn.cmd --version` or `mvn --version` are not being looked up?

> but it will be the most reliable way of finding the mvn.cmd executable as it is guaranteed to respect the full process path.

Yes, I am also not much in favour of walking through the `bin` directory and finding the path when we know that the launching a subshell here is not prone to arbitrary code execution.

slarse (changes requested):
Let's go with launching a subshell, at least now we understand why it's the easiest solution.

I'd request two additional things here:

1. A test for `guessMavenHome` to see that it actually finds _something_ on all platforms.
2. A comment stating why we launch a subshell (Runtime.exec() doesn't do binary lookup on the path on Windows, or at least that's the working theory)

slarse:
> > Java doesn't do any lookup on the path at ALL when creating a subprocess.
> 
> Some questions about this statement.
> 
> Q.1 How is this happening in Linux then? We don't use `M2_HOME` there and I think we get the path to binary of `mvn` by executing `mvn --version`. And we also don't open a subshell for executing this command in case of Linux.

It does binary lookup on path on Linux, I've tested that locally.

> Q.2 Why are there two outputs of `mvn --version` [here](https://github.com/SpoonLabs/sorald/runs/4193323097?check_suite_focus=true#step:8:21)? One comes from executing the `mvn` binary via an absolute path. Where does the other come from if binaries in `mvn.cmd --version` or `mvn --version` are not being looked up?

That's a good question. There seems to be some overlap in output here, but looking again, there actually seems to be only one exception. Hold on, I'll try again.

algomaster99:
> There seems to be some overlap in output here, but looking again, there actually seems to be only one exception. Hold on, I'll try again.

I think the subprocess exits with code `1` but it doesn't raise a Java exception and that's why the tests just passed even though we didn't get the desired result. Running only [`mvn.cmd` confirms it](https://github.com/SpoonLabs/sorald/runs/4198668159?check_suite_focus=true).

slarse:
> > There seems to be some overlap in output here, but looking again, there actually seems to be only one exception. Hold on, I'll try again.
> 
> I think the subprocess exits with code `1` but it doesn't raise a Java exception and that's why the tests just passed even though we didn't get the desired result. Running only [`mvn.cmd` confirms it](https://github.com/SpoonLabs/sorald/runs/4198668159?check_suite_focus=true).

Possibly. It's just absolutely incomprehensible to me why it doesn't do binary lookup on the path. On Windows. It _clearly_ does so on *NIX systems, well at least Linux. I'll try on macOS when I next get my hands on a Mac.

Let's just go with launching a subshell anyway. We've spent enough time here :)

algomaster99:
> Let's just go with launching a subshell anyway. We've spent enough time here :)

Great! I will make the two changes you suggested above by tomorrow.

algomaster99:
@slarse , the tests for `guessMavenHome` [already exists](https://github.com/INRIA/spoon/blob/master/src/test/java/spoon/MavenLauncherTest.java#L211) and it [passes on Windows](https://github.com/INRIA/spoon/runs/4201008971?check_suite_focus=true#step:10:47) too which is very weird because the virtual environment of this build is as follows
```
Environment: windows-2019
  Version: 20211110.1
  Included Software: https://github.com/actions/virtual-environments/blob/win19/20211110.1/images/win/Windows2019-Readme.md
  Image Release: https://github.com/actions/virtual-environments/releases/tag/win19%2F20211110.1
```
And according to https://github.com/actions/virtual-environments/issues/4485#issuecomment-967145878, the test should have failed because I am assuming that if `M2_HOME` is removed in 20211102.4, it should also not exist in  20211110.1.

Another reason which supports that the test should fail because we already reached a conclusion (perhaps a correct one) that Java subprocess is not able to lookup path in Windows system and so eventually, it should not be able to find `mvnHome`.

slarse:
You're right, there is a test. I missed it because it is loading the method dynamically, maybe it was private at some point.

Anyway, yes, that test works fine, and I tried it on my local machine as well. Windows _does_ do binary lookup on the path, the reason it didn't work in my "debugging" session was that I had forgotten to remove a line of code that added the actual binary to the path. Which caused strangeness.

So, we're back to "we don't actually need to launch a subshell", I'm afraid. There's something wrong in Sorald's environment.

slarse:
I've finally narrowed it down. This has something to do with JUnit5 on Windows. On my local Windows machine, this works fine:

```java
@org.junit.Test
public void testGuessMavenHomeReturnsString() {
    assertThat(SpoonPom.guessMavenHome(), is(notNullValue(String.class)));
}
```

But if I switch `@org.junit.Test` for `@org.junit.jupiter.api.Test`, it fails.

slarse:
Found the problem, it's the `SystemExitHandler` that registers a custom `SecurityManager`, which causes strangeness in executing a subprocess.

algomaster99:
We decided to refrain from launching a sub-command prompt and found a [better workaround](https://github.com/SpoonLabs/sorald/pull/617).

slarse:
So, I've considered this some more, and I think we should either a) launch a subshell, or b) actually do the lookup of the full path for the binary. The former is a bit nasty but _probably_ okay, the latter is the proper way.

It doesn't really make sense for Spoon's MavenLauncher to suddenly stop working on Windows in the presence of a `SecurityManager`. If we say we support Windows, then we should aim for feature parity between Windows and *NIX.

algomaster99:
> It doesn't really make sense for Spoon's MavenLauncher to suddenly stop working on Windows in the presence of a SecurityManager.

It's not sudden, right? It had never been working because we always fell back on `M2_HOME` which corroborates your point that we should have a more *permanent* way because we say we support windows. In that case, maybe we should look up the `maven` binary the proper way. I can open an issue for that and submit a PR when I get time.

slarse:
> It's not sudden, right? 

No, it's probably never worked. But `M2_HOME` has always been set in our test environments and has historically been widely used.

> In that case, maybe we should look up the maven binary the proper way. I can open an issue for that and submit a PR when I get time.

That would be great.

PR #4277, opened 2021-11-10
Revert "chore: add qodana analyzer"

slarse:
Reverts INRIA/spoon#4244

Failed to attribute all co-authors


MartinWitt:
After some chatting with @I-Al-Istannen he seems to value a clean master higher than one commit or less with his coauthor ship. 
We can close this and preserve the impression of a clean master.

slarse:
I think contributor attribution outweighs the (frankly quite small) clutter of an extra two commits, but if that's his wish then sure.

@I-Al-Istannen sorry about losing your contrbution

PR #4279, opened 2021-11-11
fix: add missing space after modifiers in sniper mode

### LINKED ISSUE #4220: SniperPrettyPrinter - Modifiers are printed without trailing white spaces
When using SniperPrettyPrinter, for the below input the expected output is not generated. Modifiers are printed without trailing white spaces.

Input :
```java
public class CtFieldWriteProcessor {
	
		private static final java.lang.String x;
		
		CtFieldWriteProcessor()
		{
			x = "Test";
		}
	}
```
Expected :
```java
public class CtFieldWriteProcessor {
	
		private static final int x;
		
		CtFieldWriteProcessor()
		{
			x = "Test";
		}
	}
```
Actual :
```java
public class CtFieldWriteProcessor {
	
		private static finaljava.lang.String x;
		
		CtFieldWriteProcessor()
		{
			x = "Test";
		}
	}
```
Spoon print config :
```java
launcher.getEnvironment().setPrettyPrinterCreator(() -> new SniperJavaPrettyPrinter(launcher.getEnvironment()));
```

Spoon version : 9.1.0-beta-10

algomaster99:
Fixes #4220 

I haven't fixed this till now, but I have found the cause, where this bug, might emanate from.

While trying to print the modified `CtField` instance in this test, one of the functions called is [`getGroupedChildrenFragments`](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/sniper/internal/ElementSourceFragment.java#L490). It generates `flatChildren` of the field as follows:

```java
0 = {ElementSourceFragment@4115} "|58, 65|private|"
1 = {TokenSourceFragment@4128} "| |"
2 = {ElementSourceFragment@4129} "|66, 72|static|"
3 = {TokenSourceFragment@4130} "| |"
4 = {ElementSourceFragment@4131} "|73, 78|final|"
5 = {TokenSourceFragment@4132} "| |"
6 = {ElementSourceFragment@4133} "java.lang.String|79, 95|java.lang.String|"
7 = {TokenSourceFragment@4134} "| |"
8 = {TokenSourceFragment@4135} "|x|"
9 = {TokenSourceFragment@4136} "|;|"
```
Until now, there is a space after `final`, so we are good and can possibly expect to print the space between `final` and `java.lang.String`. However, upon reaching [this line](https://github.com/INRIA/spoon/blob/7052a94bb094ee96fb33637636255e15552b68a3/src/main/java/spoon/support/sniper/internal/ElementSourceFragment.java#L520), the value of `lastOfSameRole` is set to 4 so `childrenInSameCollection` becomes:
```java
0 = {ElementSourceFragment@4115} "|58, 65|private|"
1 = {TokenSourceFragment@4128} "| |"
2 = {ElementSourceFragment@4129} "|66, 72|static|"
3 = {TokenSourceFragment@4130} "| |"
4 = {ElementSourceFragment@4131} "|73, 78|final|"
```
Please note that space token  _after_ `final` is excluded. Now we enter the function for [writing modifiers](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/reflect/visitor/ElementPrinterHelper.java#L106-L116) where I noticed an anomaly.

While it was [at this line](https://github.com/INRIA/spoon/blob/7052a94bb094ee96fb33637636255e15552b68a3/src/main/java/spoon/reflect/visitor/ElementPrinterHelper.java#L107), I expected it to print only the items in in `firstPostion` (only `private`). However, it printed contents of `secondPosition` and `thirdPosition` too. Internally, it transformed the prettyprinted string from
```java
package sniperPrint;

public class SpaceAfterFinal {

    
```
to
```java
package sniperPrint;

public class SpaceAfterFinal {

    private static final
```
I checked why it would print all of those modifiers at once. I found out it is because of [this line here](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/sniper/SniperJavaPrettyPrinter.java#L234). Note also that there is no space after `final` in the latter code snippet even though there is a writeSpace call after `writeKeyword` [here](https://github.com/INRIA/spoon/blob/7052a94bb094ee96fb33637636255e15552b68a3/src/main/java/spoon/reflect/visitor/ElementPrinterHelper.java#L107). Apparently, the `muted` attribute is changed to `true` for this space created by `writeSpace` and [it doesn't print](https://github.com/INRIA/spoon/blob/7052a94bb094ee96fb33637636255e15552b68a3/src/main/java/spoon/support/sniper/internal/AbstractSourceFragmentPrinter.java#L52).

Anyway, moving on to parts of code where elements of `secondPostion` and `thirdPostion` are [attempted to print](https://github.com/INRIA/spoon/blob/7052a94bb094ee96fb33637636255e15552b68a3/src/main/java/spoon/reflect/visitor/ElementPrinterHelper.java#L110-L116). The reason why `static` and `final` are not printed again is, again, because `muted` is set to `true`.

A natural way to debug this correctly is to check how the space is being printed between `public` and `class`. [There also the keyword `class` is written, and then `writeSpace` is called explicitly (which doesn't work too)](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java#L646), but when we go inside `writeIdentifier`, it writes the space and then appends the identifier.

Finally, I suspect that the space is not being printed because when `writeSpace` was called in the loop for `firstPostion`, `muted` was set to `true`. If I am right, I think I need to find out how the `muted` behaviour changes throughout the code and how exactly its value is decided.

### DIFF
--- a/src/main/java/spoon/support/sniper/internal/AbstractSourceFragmentPrinter.java
+++ b/src/main/java/spoon/support/sniper/internal/AbstractSourceFragmentPrinter.java
@@ -278,7 +278,7 @@ abstract class AbstractSourceFragmentPrinter implements SourceFragmentPrinter {
 		CtRole role = event.getRole();
 		if (role != null) {
 			if ((event.getElement() instanceof CtModifiable && event.getElement().getPosition().isValidPosition())
-					|| role == CtRole.MODIFIER) {
+					|| role == CtRole.MODIFIER || role == CtRole.TYPE) {
 				// using only roles for handling modifiers and preexisting modifiables correctly
 				return findIndexOfNextChildTokenOfRole(childFragmentIdx + 1, role);
 			}
--- a/src/test/java/spoon/test/prettyprinter/TestSniperPrinter.java
+++ b/src/test/java/spoon/test/prettyprinter/TestSniperPrinter.java
@@ -797,6 +797,21 @@ public class TestSniperPrinter {
 		testSniper("superCall.SuperCallSniperTestClass", deleteForUpdate, assertContainsSuperWithUnaryOperator);
 	}
 
+	@Test
+	@GitHubIssue(issueNumber = 4220)
+	void testSniperAddsSpaceAfterFinal() {
+		Consumer<CtType<?>> modifyField = type -> {
+			Factory factory = type.getFactory();
+			CtField field = type.filterChildren(CtField.class::isInstance).first();
+			field.setType(factory.Type().integerType());
+		};
+
+		BiConsumer<CtType<?>, String> assertContainsSpaceAfterFinal = (type, result) ->
+				assertThat(result, containsString("private static final java.lang.Integer x;"));
+
+		testSniper("sniperPrint.SpaceAfterFinal", modifyField, assertContainsSpaceAfterFinal);
+	}
+
 	/**
 	 * 1) Runs spoon using sniper mode,
 	 * 2) runs `typeChanger` to modify the code,
--- a/src/test/java/spoon/test/prettyprinter/TestSniperPrinter.java
+++ b/src/test/java/spoon/test/prettyprinter/TestSniperPrinter.java
@@ -811,6 +811,21 @@ public class TestSniperPrinter {
 		testSniper("ArithmeticExpression", noOpModifyFieldAssignment, assertPrintsRoundBracketsCorrectly);
 	}
 
+	@Test
+	@GitHubIssue(issueNumber = 4220)
+	void testSniperAddsSpaceAfterFinal() {
+		Consumer<CtType<?>> modifyField = type -> {
+			Factory factory = type.getFactory();
+			CtField field = type.filterChildren(CtField.class::isInstance).first();
+			field.setType(factory.Type().integerType());
+		};
+
+		BiConsumer<CtType<?>, String> assertContainsSpaceAfterFinal = (type, result) ->
+				assertThat(result, containsString("private static final java.lang.Integer x;"));
+
+		testSniper("sniperPrint.SpaceAfterFinal", modifyField, assertContainsSpaceAfterFinal);
+	}
+
 	/**
 	 * 1) Runs spoon using sniper mode,
 	 * 2) runs `typeChanger` to modify the code,
--- a/src/test/resources/sniperPrint/SpaceAfterFinal.java
+++ b/src/test/resources/sniperPrint/SpaceAfterFinal.java
@@ -1,0 +1,11 @@
+package sniperPrint;
+
+public class SpaceAfterFinal {
+
+    private static final java.lang.String x;
+
+    public SpaceAfterFinal()
+    {
+        x = "Test";
+    }
+}


algomaster99:
@nharrand or @slarse , could you please throw some light here? And maybe try to give more insights related to this bug.

slarse:
I might have some time this weekend to look at this, I'll try to make some time at least.

algomaster99:
@slarse that's very helpful. Thanks!

I think I am currently having some trouble related to fragments. I don't understand how they are generated and printed. I feel if I get a hang of it, some issues related to sniper printer may become easier.

monperrus:
> If I am right, I think I need to find out how the muted behaviour changes throughout the code and how exactly its value is decided.

sounds like a good trail to follow

algomaster99:
@slarse , if you get time today, please have a look at it.

slarse:
@algomaster99, sorry, I spent too much time on the whole Windows thing, which seemed more urgent. Don't think I'll have time to look at this :(

algomaster99:
Not a problem at all. Thanks for your time after the working hours!

algomaster99:
The patch https://github.com/INRIA/spoon/pull/4279/commits/1efa523bf4c3c13fe0dbe8dcc0a5b24fd9dcdade works! But I am not sure why it is working. I just had an intuition that this piece of code would be responsible for pretending spaces before printing modifier (and now type) so I just added another boolean condition for that.

Once I try to figure how this works, I will open the PR for reviews.

algomaster99:
Okay, so I think I know why this works and this is indeed the correct way to resolve this bug. Let me try to explain what was happening after:
```java
package sniperPrint;

public class SpaceAfterFinal {

    private static final
```
We have to print the `type` of the field which starts [here](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java#L783). Note that we print the string fragment by fragment so following is the list of fragments we are supposed to print in this iteration.
```java
0 = {CollectionSourceFragment@4046} "[|58, 65|private|, | |, |66, 72|static|, | |, |73, 78|final|]"
1 = {TokenSourceFragment@4047} "| |"
2 = {ElementSourceFragment@4048} "java.lang.String|79, 95|java.lang.String|"
3 = {TokenSourceFragment@4049} "| |"
4 = {TokenSourceFragment@4050} "|x|"
5 = {TokenSourceFragment@4051} "|;|"
```
Back to `scan(type)`, when we go into it, we encounter `AbstractSourceFragmentPrinter.print` which is supposed to print spaces and the current fragment. To do so, it keeps track of two things:
1. `prevIndex`: the index of the fragment which has just been printed - `0` in our case.
2. `event`: what has to be printed. This is decided by finding its index in the above fragment list.

For step number `2`, we call [`update`](https://github.com/INRIA/spoon/blob/3ebf4720d438646ab98fda35887b535223c479fd/src/main/java/spoon/support/sniper/internal/AbstractSourceFragmentPrinter.java#L57) which is supposed to fetch the index of the event (in our case `TYPE->java.lang.Integer`) and **this is where the bug is**. `update(event)` returned `-1` and while doing that, it enters [this block](tAbhttps://github.com/INRIA/spoon/blob/master/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java#L783). This block prints the return type _without printing the space before_.

Why it returned `-1` in the first place? That's because it tried to [find `java.lang.Integer` in the fragment which does not exist](tAbhttps://github.com/INRIA/spoon/blob/master/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java#L783). So my change ensures that we try to find the fragment (index `2`) by providing a role. Since role of `java.lang.String` and `java.lang.Integer` is same, we get the correct index [here](https://github.com/INRIA/spoon/blob/3ebf4720d438646ab98fda35887b535223c479fd/src/main/java/spoon/support/sniper/internal/AbstractSourceFragmentPrinter.java#L57). Thus, we enter into the subsequent block and print a space before we print the return type.

algomaster99:
@monperrus @slarse this is ready for review. I have also tried some edge cases like having a `type` in a method or a variable declaration and it works there. Should I also push tests for that?

slarse:
@nharrand Can you take this one?

nharrand:
@slarse I'll try to have a look before the end of the week. I am a bit busy witting my thesis these days.

monperrus:
looked at it with @algomaster99

I'll merge once the conflict is resolved.

@nharrand this saves you precious time for writing your PhD thesis.

PR #4018, opened 2021-06-30
review: test: add tests for substitution class

Rohitesh-Kumar-Jain:
#1818

Hi, I have added 3 new tests for  methods of the `Substitution` class : 

1) `testSubstituteMethodBody` should kill this mutation : 
Screenshot 2021-06-30 at 4 01 07 PM

2) `testSubstituteStatement` should kill this mutation :

Screenshot 2021-06-30 at 4 02 09 PM

3) `testSubstituteFieldDefaultExpression`should kill this mutation :
Screenshot 2021-06-30 at 4 03 34 PM

Link to #3967

PR #4282, opened 2021-11-12
fix: remove annotation from the pretty-printed string if it is deleted from the model

### LINKED ISSUE #4218: SniperPrettyPrinter - Annotation not getting deleted on transformation
When using SniperPrettyPrinter, for the below input the expected output is not generated. Spoon is not deleting the annotation with delete api.

Input :
`@abc.def.xyz
public class CtAnnotationProcessor extends AbstractSessionBaseEjb3 { 
public void print(String str) {
		System.out.println(str);
}`

Expected output :
`public class CtAnnotationProcessor extends AbstractSessionBaseEjb3 { 
public void print(String str) {
		System.out.println(str);
}`

Actual Output :
`@abc.def.xyz
public class CtAnnotationProcessor { 
public void print(String str) {
		System.out.println(str);
}`

Code used for this transformation :
`@Override
public void process(final CtAnnotation<?> element) {
element.delete();
}`

Spoon print config :
`launcher.getEnvironment().setPrettyPrinterCreator(() -> new SniperJavaPrettyPrinter(launcher.getEnvironment()));`

Spoon version : 9.1.0-beta-10

algomaster99:
Fixes #4218 

I am still looking for a fix for this, but I think the problem could be with the [`changeCollector`](https://github.com/INRIA/spoon/blame/master/src/main/java/spoon/support/sniper/SniperJavaPrettyPrinter.java#L94). The size of the map `elementToChangeRole` is `0`, which is unusual as we have altered the type by deleting the annotation attached to it. Thus, I am concluding that it fails to listen for deletion of annotation in the model build for the resource file in this PR. 

EDIT:

`onChange` is not being called when I delete an annotation, however, it is called when I delete an instance of `CtField` inside a `CtType`. This should confirm that what I am saying above is right.

### DIFF
--- a/src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
+++ b/src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
@@ -201,9 +201,9 @@ public abstract class CtElementImpl implements CtElement, Serializable {
 	public <E extends CtElement> E setAnnotations(List<CtAnnotation<? extends Annotation>> annotations) {
 		if (annotations == null || annotations.isEmpty()) {
 			this.annotations = emptyList();
+			getFactory().getEnvironment().getModelChangeListener().onListDeleteAll(this, CtRole.ANNOTATION, this.annotations, new ArrayList<>(this.annotations));
 			return (E) this;
 		}
-		getFactory().getEnvironment().getModelChangeListener().onListDeleteAll(this, CtRole.ANNOTATION, this.annotations, new ArrayList<>(this.annotations));
 		this.annotations.clear();
 		for (CtAnnotation<? extends Annotation> annot : annotations) {
 			addAnnotation(annot);
--- a/src/test/java/spoon/test/prettyprinter/TestSniperPrinter.java
+++ b/src/test/java/spoon/test/prettyprinter/TestSniperPrinter.java
@@ -28,6 +28,7 @@ import spoon.reflect.code.CtLocalVariable;
 import spoon.reflect.code.CtStatement;
 import spoon.reflect.code.CtThrow;
 import spoon.reflect.cu.SourcePosition;
+import spoon.reflect.declaration.CtAnnotation;
 import spoon.reflect.declaration.CtClass;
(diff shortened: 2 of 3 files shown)


algomaster99:
Any ideas related to this @nharrand and @slarse ?

monperrus:
> onChange is not being called when I delete an annotation, however, it is called when I delete an instance of CtField inside a CtType. This should confirm that what I am saying above is right.

seems you have the core idea of the fix :)

algomaster99:
@monperrus, and here comes this fix!

When all annotations are deleted, `this.annotations` is set to `emptyList`, but we also need to register this change in the environment. The patch does exactly that.

algomaster99:
@monperrus @slarse @nharrand need your reviews. The tests have passed.

monperrus:
nice, LGTM. thanks @algomaster99 

there are many more test cases in #4018, does this fix make them all pass?

algomaster99:
@monperrus I think I linked the wrong issue. :man_facepalming: Let me find the correct one.

algomaster99:
@monperrus I found it. It was https://github.com/INRIA/spoon/issues/4218 instead of 4018.

algomaster99:
@monperrus Some certifcate has expired which is making tests fail.
```
Run actions/setup-java@8db439b6b47e5e12312bf036760bbaa6893481ac
  with:
    java-version: 16
    distribution: temurin
    java-package: jdk
    architecture: x64
    check-latest: false
    server-id: github
    server-username: GITHUB_ACTOR
    server-password: GITHUB_TOKEN
    overwrite-settings: true
    job-status: success
  env:
    JAVA_DISTRIBUTION: temurin
    MAVEN_OPTS: -Djava.src.version=16 -Dhttp.keepAlive=false -Dmaven.wagon.http.pool=false
Trying to resolve the latest version from remote
Error: certificate has expired
```
Do you know what certificate are they talking about? [Failing step](https://github.com/INRIA/spoon/runs/4189881964?check_suite_focus=true#step:4:1).

PR #4283, opened 2021-11-12
fix: printing of generic records (issue #4281)

### LINKED ISSUE #4281: [Bug] generic records output is wrong
**Describe the bug**
Generic record output is wrong, the generic type is printed to the right of the argument list.

**To Reproduce**

Input:
```java
public record A<T>(int a, int b) {}
```

Processing with Spoon:
```java
        String code = "public record A<T>(int a, int b) {}";

        System.out.println(model.getRootPackage().getType("A").toString());
```

Output:
```java
public record A(int a, int b)<T> {}
```

* OS: MacOS 12.0.1
* JDK: same result with Oracle OpenJDK 17 and GraalVM CE 21.3.0 
* Spoon version: same result with 10.0.0 and 10.0.1-beta-1 (maven central repository)

xzel23:
Fix #4281 

This PR moves the type argument to the correct location when printing generic record declarations. A new unit test is included.

### DIFF
--- a/src/test/java/spoon/test/record/CtRecordTest.java
+++ b/src/test/java/spoon/test/record/CtRecordTest.java
@@ -148,7 +148,7 @@ public class CtRecordTest {
 
 	@Test
 	void testGenericTypeParametersArePrintedBeforeTheFunctionParameters() {
-		// a record with generic type arguments should be printed correctly 
+		// contract: a record with generic type arguments should be printed correctly 
 		String code = "src/test/resources/records/GenericRecord.java";
 		CtModel model = createModelFromPath(code);
 		Collection<CtRecord> records = model.getElements(new TypeFilter<>(CtRecord.class));
--- a/src/test/java/spoon/test/record/CtRecordTest.java
+++ b/src/test/java/spoon/test/record/CtRecordTest.java
@@ -147,7 +147,7 @@ public class CtRecordTest {
 	}
 
 	@Test
-	void printGenericRecord() {
+	void testGenericTypeParametersArePrintedBeforeTheFunctionParameters() {
 		// a record with generic type arguments should be printed correctly 
 		String code = "src/test/resources/records/GenericRecord.java";
 		CtModel model = createModelFromPath(code);
--- a/src/test/java/spoon/test/record/CtRecordTest.java
+++ b/src/test/java/spoon/test/record/CtRecordTest.java
@@ -146,6 +146,16 @@ public class CtRecordTest {
 		return model;
 	}
 
+	@Test
+	void printGenericRecord() {
+		// a record with generic type arguments should be printed correctly 
+		String code = "src/test/resources/records/GenericRecord.java";
+		CtModel model = createModelFromPath(code);
+		Collection<CtRecord> records = model.getElements(new TypeFilter<>(CtRecord.class));
+		CtRecord record = head(records);
+		assertEquals("public record GenericRecord<T>(T a, T b) {}", record.toString());
+	}
+
 	private <T> T head(Collection<T> collection) {
 		return collection.iterator().next();
 	}
--- a/src/test/resources/records/GenericRecord.java
+++ b/src/test/resources/records/GenericRecord.java
@@ -1,0 +1,3 @@
+package records;
+
+public record GenericRecord<T>(T a, T b) {}
--- a/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
+++ b/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
@@ -2197,8 +2197,8 @@ public class DefaultJavaPrettyPrinter implements CtVisitor, PrettyPrinter {
 		context.pushCurrentThis(recordType);
 		visitCtType(recordType);
 		printer.writeKeyword("record").writeSpace().writeIdentifier(stripLeadingDigits(recordType.getSimpleName()));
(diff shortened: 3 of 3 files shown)


src/test/resources/records/GenericRecord.java
```diff
@@ -0,0 +1,3 @@
+package records;
+
+public record GenericRecord<T>(T a, T b) {}
```
algomaster99, line 3: I think this line is syntactically incorrect. Instead of `record`, there should be `class`, `interface`, or even `enum`. I tried running the tests after replacing `record` with `class` and it returned the expected string.

algomaster99 (commented):
Hi @xzel23 !

Thank you for the pull request. Please look at my comment.

src/test/resources/records/GenericRecord.java
algomaster99, line 3: Could you please add this test case under `src/test/resources/records`? It's where we put all the tests related to records.

src/test/java/spoon/test/record/CtRecordTest.java
```diff
@@ -146,6 +146,16 @@ private CtModel createModelFromPath(String code) {
 		return model;
 	}
 
+	@Test
+	void printGenericRecord() {
```
algomaster99, line 150: We try to name new tests we create by describing what is happening in the test. I think a better name would be `testGenericTypeParametersArePrintedBeforeTheFunctionParameters`.

algomaster99 (approved):
It looks good to me. Thanks, @xzel23 ! I am sorry I gave a misinformed review first.

Anyway, @monperrus @slarse , you can also have a look.

xzel23:
@algomaster99 Yes, I have seen code in Spoon using "record" as a variable name in the test classes. That code will probably break when compiling in Java 14+ where it was introduced as a keyword. The code in the unit test actually is more or less taken from one of my source files that compile perfectly but cannot be handled by spoon.

src/test/resources/records/GenericRecord.java
xzel23: Hm... I think the suggested path is the same as where the file is currently located. But I think I should update the package name.

src/test/java/spoon/test/record/CtRecordTest.java
xzel23: OK, will update the PR

algomaster99:
> That code will probably break when compiling in Java 14+

@xzel23 We don't compile our test resources before using them in tests and probably that's why all the [tests are green when executed in ubuntu/Java 16 environment](https://github.com/INRIA/spoon/runs/4188656082?check_suite_focus=true). Should we consider compiling our test resources, @slarse? We do it in `diffmin`, but I know it will really hard to do that in `spoon` because of the number of resources we have in this project.

> cannot be handled by spoon.

Spoon doesn't compile code before building the model. You can even create a model for syntactically incorrect code. For example,
```java
public class A {
    x;
}
```
The return type of field is missing.

However, we guarantee correct transformations only for syntactically correct code. What you encountered was a bug in the printer (thanks for the fix). In other words, spoon created a valid [model](https://spoon.gforge.inria.fr/mvnsites/spoon-core/apidocs/spoon/reflect/CtModel.html) for your code. It just failed to print it correctly when you called `toString`.

xzel23:
@algomaster99 I just learned that using `record` as a variable name is indeed legal even in Java 14+ where it is a keyword. The JLS defines it as a [contextual keyword](https://docs.oracle.com/javase/specs/jls/se17/html/jls-3.html#jls-3.9).

algomaster99:
> variable name is indeed legal even in Java 14+ where it is a keyword.

Oh, nice. Our test resources are free of at least one possible error :)

Thanks for renaming the test. Could you also shift the resource to the suggested directory?

xzel23:
@algomaster99 I would do it, but I'm afraid I can't ;-)

Please compare the paths. First one is the current resource path, second one the suggestzed directory:

`src/test/resources/records/GenericRecord.java`
`src/test/resources/records`

I don't see any difference. Where should I move the file to?

algomaster99:
Oh, right. I mistook the path because of diffs in multiple file. 😅 I think we are good to go. Let maintainers approve your PR for workflow and then we will review further, if needed.

src/test/java/spoon/test/record/CtRecordTest.java
```diff
@@ -146,6 +146,16 @@ private CtModel createModelFromPath(String code) {
 		return model;
 	}
 
+	@Test
+	void testGenericTypeParametersArePrintedBeforeTheFunctionParameters() {
+		// a record with generic type arguments should be printed correctly
```
MartinWitt, line 151: Could you add a `// contract:  ... a record with generic type arguments should be printed correctly`?

MartinWitt (approved):

slarse:
> We don't compile our test resources before using them in tests 

This is by design, a lot of the test resources can't compile as they are missing dependencies. It would however be neat to check the syntax. There's already something along the lines of a syntax check implemented in `TreeBuilderCompiler.ignoreSyntaxErrors`, could probably start from there.

Approved the workflow, getting back to this later.

src/test/java/spoon/test/record/CtRecordTest.java
xzel23: done

slarse:
@MartinWitt you approved with a comment, are you happy with the PR? if so I merge.

MartinWitt:
yes you can merge.

slarse:
@MartinWitt okidoki

Thanks @xzel23 for the fix!

PR #4284, opened 2021-11-12
fix: removed annotation should not appear in sniper mode

### LINKED ISSUE #4218: SniperPrettyPrinter - Annotation not getting deleted on transformation

algomaster99:
Duplicate of #4282 

Fixes #4218

Opening this PR again in the hope that the tests will pass. Some weird issue came in the linked PR.

### DIFF
--- a/src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
+++ b/src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
@@ -200,12 +200,12 @@ public abstract class CtElementImpl implements CtElement, Serializable {
 	@Override
 	public <E extends CtElement> E setAnnotations(List<CtAnnotation<? extends Annotation>> annotations) {
 		if (annotations == null || annotations.isEmpty()) {
-			this.annotations = emptyList();
 			getFactory().getEnvironment().getModelChangeListener().onListDeleteAll(this, CtRole.ANNOTATION, this.annotations, new ArrayList<>(this.annotations));
+			this.annotations = emptyList();
 			return (E) this;
 		}
-		this.annotations.clear();
 		getFactory().getEnvironment().getModelChangeListener().onListDeleteAll(this, CtRole.ANNOTATION, this.annotations, new ArrayList<>(this.annotations));
+		this.annotations.clear();
 		for (CtAnnotation<? extends Annotation> annot : annotations) {
 			addAnnotation(annot);
 		}
--- a/src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
+++ b/src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
@@ -201,9 +201,9 @@ public abstract class CtElementImpl implements CtElement, Serializable {
 	public <E extends CtElement> E setAnnotations(List<CtAnnotation<? extends Annotation>> annotations) {
 		if (annotations == null || annotations.isEmpty()) {
 			this.annotations = emptyList();
+			getFactory().getEnvironment().getModelChangeListener().onListDeleteAll(this, CtRole.ANNOTATION, this.annotations, new ArrayList<>(this.annotations));
 			return (E) this;
 		}
-		getFactory().getEnvironment().getModelChangeListener().onListDeleteAll(this, CtRole.ANNOTATION, this.annotations, new ArrayList<>(this.annotations));
 		this.annotations.clear();
 		for (CtAnnotation<? extends Annotation> annot : annotations) {
 			addAnnotation(annot);
--- a/src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
+++ b/src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
@@ -205,6 +205,7 @@ public abstract class CtElementImpl implements CtElement, Serializable {
 			return (E) this;
 		}
 		this.annotations.clear();
+		getFactory().getEnvironment().getModelChangeListener().onListDeleteAll(this, CtRole.ANNOTATION, this.annotations, new ArrayList<>(this.annotations));
 		for (CtAnnotation<? extends Annotation> annot : annotations) {
 			addAnnotation(annot);
 		}
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -159,14 +159,14 @@ jobs:
       - name: "Create folder"
         run: "mkdir -p ${{ github.workspace }}/result/master"
       - name: Qodana - Code Inspection
-        uses: JetBrains/qodana-action@v3.2.1
+        uses: JetBrains/qodana-action@cd7dc31b49427ce740ebf6255319b267801d08cf # renovate: tag=v3.2.1
         with:
           linter: qodana-jvm-community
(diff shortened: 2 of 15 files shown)


algomaster99:
@slarse Any idea what [certificate](https://github.com/INRIA/spoon/runs/4189980038?check_suite_focus=true#step:4:17) are the workflows talking about?

slarse:
@algomaster99 Looks like someone forgot to renew a domain name, restarting the workflow.

algomaster99:
You mean this domain - https://spoon.gforge.inria.fr/?

algomaster99:
@monperrus ping for reviews.

src/main/java/spoon/support/reflect/declaration/CtElementImpl.java
```diff
@@ -201,9 +201,9 @@ public int hashCode() {
 	public <E extends CtElement> E setAnnotations(List<CtAnnotation<? extends Annotation>> annotations) {
 		if (annotations == null || annotations.isEmpty()) {
 			this.annotations = emptyList();
+			getFactory().getEnvironment().getModelChangeListener().onListDeleteAll(this, CtRole.ANNOTATION, this.annotations, new ArrayList<>(this.annotations));
 			return (E) this;
 		}
-		getFactory().getEnvironment().getModelChangeListener().onListDeleteAll(this, CtRole.ANNOTATION, this.annotations, new ArrayList<>(this.annotations));
```
slarse, line 206: This doesn't seem right to me. The list is cleared on the next line, but the listener is not called now.

algomaster99: I tested this. If you delete any number of annotations, it registers what annotations need to be added (the ones which are left after deletion) in the for-loop which follows.

slarse: > I tested this. If you delete any number of annotations, it registers what annotations need to be added (the ones which are left after deletion) in the for-loop which follows.

Well, the problem I see is that if you set a non-empty list of annotations, then the change listener is only informed of additions. I'm not certain of precisely how this plays with the sniper printer, but this is strictly incorrect. Don't think of this in terms of the sniper printer alone (the change listener interface knows nothing of the sniper printer), but in a general sense: If the list is cleared, `onListDeleteAll` _must_ be called.

algomaster99: > Don't think of this in terms of the sniper printer alone (the change listener interface knows nothing of the sniper printer),

Okay, I get what you are trying to say, and this makes sense. The listener could be used for something else in future, and "If the list is cleared, onListDeleteAll must be called" might be necessary there. I think this was just a fortuitous event.

I will revert the deletion.

algomaster99:
@monperrus @slarse I think this is ready for merge.

Just as a note - changes like https://github.com/INRIA/spoon/pull/4284/commits/1bdf92bd36d2adc8671b630693d102cfe0116cb6 would also be needed in methods like [setTypeMembers](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/reflect/declaration/CtTypeImpl.java#L165-L179).

algomaster99:
@monperrus The order of registering an event listener and performing the action [doesn't matter for `ChangeCollector`](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/modelobs/ChangeCollector.java#L204) because it just ignores the `oldValue`. It is just concerned with what roles has been changed. Example,

```java
getFactory().getEnvironment().getModelChangeListener().onListDeleteAll(this, CtRole.ANNOTATION, this.annotations, new ArrayList<>(this.annotations));
this.annotations.clear();
```
and
```java
this.annotations.clear();
getFactory().getEnvironment().getModelChangeListener().onListDeleteAll(this, CtRole.ANNOTATION, this.annotations, new ArrayList<>(this.annotations));
```
are the same thing for registering a change in a `ChangeCollector`.

Therefore, it is hard to assert the difference in change collector before and after the transformation because both of the above cases will have the same effect.

Maybe, we could try writing a test case which invokes [ActionBasedChangeListener.onListDeleteAll](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/modelobs/ActionBasedChangeListenerImpl.java#L85) instead of [`ChangeCollector.onListDeleteAll`](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/modelobs/ChangeCollector.java#L204). I will try to look at coverage reports and figure out where ActionBasedChangeListener is used.

algomaster99:
@monperrus [`FineModelChangeListener`](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/modelobs/FineModelChangeListener.java) is implemented by only three classes:
1. [ActionBasedChangeListenerImpl.java](https://github.com/INRIA/spoon/blob/3d6f585543e62f50b5e18bd98a1434054337a1cd/src/main/java/spoon/support/modelobs/ActionBasedChangeListenerImpl.java#L31): Only this class uses the `field` and `oldValue` parameters, but it's only being used in a [test](https://github.com/INRIA/spoon/blob/3d6f585543e62f50b5e18bd98a1434054337a1cd/src/test/java/spoon/test/parent/ContractOnSettersParametrizedTest.java#L84).

2. [ChangeCollector.java](https://github.com/INRIA/spoon/blob/3d6f585543e62f50b5e18bd98a1434054337a1cd/src/main/java/spoon/support/modelobs/ChangeCollector.java#L162): Ignores the value of `field` and `oldValue`.

3. [EmptyModelChangeListener.java](https://github.com/INRIA/spoon/blob/3d6f585543e62f50b5e18bd98a1434054337a1cd/src/main/java/spoon/support/modelobs/EmptyModelChangeListener.java#L22): Does not do anything.

I say, that we skip writing test for now because it will only improve the robustness of `ActionBasedChangeListenerImpl.java` which is just used once in spoon and that too in tests. Not sure if it is used in other repositories, but for their sake, we could open a separate issue to write tests for ActionBasedChangeListener.

For this PR, let's move forward with merging it. I have followed the contract of the listener and made changes in accordance with that.

monperrus:
Thanks a lot @algomaster99  for the fix and @slarse for spotting the unspecified behavior.

Per our discussion today, @algomaster99 will open an issue and a PR later on properly specifying and testing onDelete methods

PR #4286, opened 2021-11-13
fix: Test failure in TestModule.testModuleNames() (#4285)

### LINKED ISSUE #4285: [Bug] Test failure in TestModule.testModuleNames()
**Describe the bug**
Test fails in `testModuleNames()` when doing `assertTrue(list.contains("bar"))`.

**To Reproduce**
Check out Spoon; run `mvn compile` then `mvn test`

**Notes**
In that test case, an array of compilation units is retrieved and then the names of the two first entries of that list are inserted into a list. Then a test is made to check if both "foo" and "bar" are contained in that list. When I run this test on my machine, the list contains four entries, two with a module name of "bar" and two with a module name of "foo", so the test result depends on the ordering of rresults.

An easy fix would be to create a list consisting of all four module names, so instead of creating the list like this:

		List list = new ArrayList<>(Arrays.asList(new String[]{String.copyValueOf(cu[0].module),String.copyValueOf(cu[1].module)}));

it should be done like this:

		List<String> list = Arrays.asList(cu).stream().map(CompilationUnit::getModuleName).map(String::copyValueOf).collect(Collectors.toList());

However I am not familiar enough with the Spoon source to know if there really should be four compilation units in that list, or if it should be two.

Path to the test files: `src/test/resources/spoon/test/module/code-multiple-modules`

* OS: MacOS 12.0.1
* JDK: openjdk 17.0.1 2021-10-19 GraalVM CE 21.3.0 (build 17.0.1+12-jvmci-21.3-b05)
* Spoon version: Source on Github, revison 15bd93f5a685474f39a8957ffa4cb94bfe171eea

xzel23:
This fixes #4285. The test relied on the order of entries returned by `File#listFiles()`, which is unspecified (thank you @SirYwell ).

### DIFF
--- a/src/test/java/spoon/test/module/TestModule.java
+++ b/src/test/java/spoon/test/module/TestModule.java
@@ -54,6 +54,7 @@ import java.util.List;
 import java.util.Collections;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import static org.junit.Assert.assertEquals;
@@ -313,7 +314,7 @@ public class TestModule {
 		batchCompiler.configure(args);
 
 		CompilationUnit[] cu = batchCompiler.getCompilationUnits();
-		List list = new ArrayList<>(Arrays.asList(new String[]{String.copyValueOf(cu[0].module),String.copyValueOf(cu[1].module)}));
+		List<String> list = Arrays.stream(cu).map(CompilationUnit::getModuleName).map(String::copyValueOf).collect(Collectors.toList());
 		assertTrue(list.contains("foo"));
 		assertTrue(list.contains("bar"));
 	}
--- a/src/test/java/spoon/test/module/TestModule.java
+++ b/src/test/java/spoon/test/module/TestModule.java
@@ -17,6 +17,8 @@
 package spoon.test.module;
 
 import org.eclipse.jdt.internal.compiler.batch.CompilationUnit;
+import org.hamcrest.CoreMatchers;
+import org.hamcrest.MatcherAssert;
 import org.junit.AfterClass;
 import org.junit.BeforeClass;
 import org.junit.Ignore;
@@ -54,12 +56,16 @@ import java.util.List;
 import java.util.Collections;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import static org.hamcrest.CoreMatchers.*;
+import static org.hamcrest.MatcherAssert.*;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertSame;
+import static org.junit.Assert.assertThat;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
@@ -314,9 +320,8 @@ public class TestModule {
 		batchCompiler.configure(args);
 
 		CompilationUnit[] cu = batchCompiler.getCompilationUnits();
-		List<String> list = Arrays.stream(cu).map(CompilationUnit::getModuleName).map(String::copyValueOf).collect(Collectors.toList());
-		assertTrue(list.contains("foo"));
-		assertTrue(list.contains("bar"));
+		Set<String> list = Arrays.stream(cu).map(CompilationUnit::getModuleName).map(String::copyValueOf).collect(Collectors.toSet());
+		MatcherAssert.assertThat(list, is(Set.of("foo", "bar")));
 	}
 
 	@org.junit.jupiter.api.Test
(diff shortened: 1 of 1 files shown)


src/test/java/spoon/test/module/TestModule.java
```diff
@@ -313,7 +314,7 @@ public void testModuleNames() {
 		batchCompiler.configure(args);
 
 		CompilationUnit[] cu = batchCompiler.getCompilationUnits();
-		List list = new ArrayList<>(Arrays.asList(new String[]{String.copyValueOf(cu[0].module),String.copyValueOf(cu[1].module)}));
+		List<String> list = Arrays.stream(cu).map(CompilationUnit::getModuleName).map(String::copyValueOf).collect(Collectors.toList());
 		assertTrue(list.contains("foo"));
 		assertTrue(list.contains("bar"));
```
slarse, line 319: While you're poking about here anyway, how about improving it just a little bit more? This test is a bit weak (because it leaves the possibility of there being other module names in the list), and it also isn't entirely clear about its purpose.

I propose we assert on the entire list. And instead of a list use a set, because we care only about distinct module names.

```suggestion
		Set<String> moduleNames = Arrays.stream(cu)
				.map(CompilationUnit::getModuleName)
				.map(String::valueOf)
				.collect(Collectors.toSet());
		assertThat(moduleNames, equalTo(Set.of("bar", "foo")));
```

slarse (changes requested):
Hi @xzel23, thanks for fixing the flaky test! I propose we go one step further to improve it while we're at it, see comment :)

src/test/java/spoon/test/module/TestModule.java
xzel23: done. I replaced the suggested deprecated assertThat with a non-deprecated one.

src/test/java/spoon/test/module/TestModule.java
```diff
@@ -17,6 +17,8 @@
 package spoon.test.module;
 
 import org.eclipse.jdt.internal.compiler.batch.CompilationUnit;
+import org.hamcrest.CoreMatchers;
+import org.hamcrest.MatcherAssert;
```
slarse, line 21: Not needed

```suggestion
```

src/test/java/spoon/test/module/TestModule.java
```diff
@@ -54,11 +56,16 @@
 import java.util.Collections;
 import java.util.ArrayList;
```
slarse, line 57: I don't think `ArrayList` is needed anymore.

```suggestion
```

src/test/java/spoon/test/module/TestModule.java
```diff
@@ -54,11 +56,16 @@
...
+import java.util.Set;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import static org.hamcrest.CoreMatchers.*;
+import static org.hamcrest.MatcherAssert.*;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertSame;
+import static org.junit.Assert.assertThat;
```
slarse, line 68:
```suggestion
```

src/test/java/spoon/test/module/TestModule.java
```diff
@@ -313,9 +320,8 @@ public void testModuleNames() {
 		batchCompiler.configure(args);
 
 		CompilationUnit[] cu = batchCompiler.getCompilationUnits();
-		List list = new ArrayList<>(Arrays.asList(new String[]{String.copyValueOf(cu[0].module),String.copyValueOf(cu[1].module)}));
-		assertTrue(list.contains("foo"));
-		assertTrue(list.contains("bar"));
+		Set<String> list = Arrays.stream(cu).map(CompilationUnit::getModuleName).map(String::copyValueOf).collect(Collectors.toSet());
+		MatcherAssert.assertThat(list, is(Set.of("foo", "bar")));
```
slarse, line 324:
```suggestion
		assertThat(list, is(Set.of("foo", "bar")));
```

slarse (changes requested):

src/test/java/spoon/test/module/TestModule.java
slarse: Hmm, I didn't suggest the deprecated `assertThat`, I just didn't specify which one to use (in fact I didn't even remember that JUnit 4 had a built-in `assertThat`). You've imported both, so I suspect your IDE auto-imported a bunch of stuff. See review comments.

src/test/java/spoon/test/module/TestModule.java
xzel23: My IDE had somehow lost track of the imports and marked everything as missing, so I didn't see the unused exports. Fixed after restart. I have cleaned it up and added a static import for the hamcrest assertThat as suggested by the junit deprecationcomment. Hope it's alright now.

slarse: Yeah, I feel you. I find that most IDEs with default settings are over-eager when it comes to auto-importing stuff, it gets rather confusing rather quickly.

slarse (approved):
Thanks @xzel23, this is a much more robust (and non-flaky) test!

PR #4292, opened 2021-11-17
doc: fix Javadoc search feature

h4iku:
Since Spoon is now using Java 11, the Javadoc has a search field. However, using it results in a "Page Not Found" like this:
https://spoon.gforge.inria.fr/mvnsites/spoon-core/apidocs/undefined/spoon/reflect/declaration/CtMethod.html
Removing the `undefined` part from this URL loads the correct page. The problem is a known bug in JDK and seems to be fixed in Java 12:
https://bugs.openjdk.java.net/browse/JDK-8215291

For Java 11, adding the `--no-module-directories` option to `maven-javadoc-plugin` fixes the issue.
see: https://stackoverflow.com/questions/52326318/maven-javadoc-search-redirects-to-undefined-url

### DIFF
--- a/spoon-pom/pom.xml
+++ b/spoon-pom/pom.xml
@@ -303,6 +303,7 @@
                     <detectJavaApiLink>false</detectJavaApiLink>
                     <!-- ref: https://github.com/INRIA/spoon/pull/4253 -->
                     <additionalJOption>-notimestamp</additionalJOption>
+                    <additionalJOption>--no-module-directories</additionalJOption>
                 </configuration>
                 <reportSets>
                     <reportSet>


SirYwell:
That option [was removed in Java 13 again](https://bugs.openjdk.java.net/browse/JDK-8215582), and the Extra checks are using Java 16 currently. Not sure about this, but it would probably be fine to generate the Javadocs for the website with a newer Java version instead?

h4iku:
Thank you, I didn't see that the option was removed. Some projects also use POM profiles to build Javadocs for Java 9-11.

slarse:
Hi @h4iku, thanks for raising this issue.

I presume this problem would be solved simply by generating the Javadoc with JDK 12+? We could do that for our online Javadoc.

slarse:
> Using JDK 16 even generates links for standard Java classes:

Ah, neat. Let's go with JDK 16, then! Thanks a lot.

PR #4296, opened 2021-11-18
fix: ensure sniper pretty-printer prints modifiers and type separated by a space

algomaster99:
Reference: https://github.com/SpoonLabs/sorald/issues/603

This patch is a consequence of what has been discussed [here](https://github.com/SpoonLabs/sorald/issues/603#issuecomment-971539825). The source position of a field's type was not being set correctly. I realised that the source position are stored in the AST nodes created by JDT compiler and the function in this patch did not have access to one. So, I have modified the function definition so I am able to build and set the position.

There are things left in this PR:
- [x] Tests for all the changes. I can assert the model built to check if one of its field's type contains the source position now.
~- [ ] Look for an alternative way because just by reading the code, my intuition is telling me that `ReferenceBuilder` is a wrong place to set the position. It should probably be done somewhere in a related file.~

EDIT: The second task shall be handled in https://github.com/INRIA/spoon/issues/4297.

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
+++ b/src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
@@ -147,7 +147,7 @@ public class ReferenceBuilder {
 		CtTypeReference<T> accessedType = buildTypeReference((TypeReference) type, scope);
 		final TypeBinding receiverType = type != null ? type.resolvedType : null;
 		if (receiverType != null) {
-			final CtTypeReference<T> ref = getQualifiedTypeReference(type.tokens, receiverType, receiverType.enclosingType(), new JDTTreeBuilder.OnAccessListener() {
+			final CtTypeReference<T> ref = getQualifiedTypeReference(type.tokens, type, receiverType, receiverType.enclosingType(), new JDTTreeBuilder.OnAccessListener() {
 				@Override
 				public boolean onAccess(char[][] tokens, int index) {
 					return true;
@@ -260,12 +260,13 @@ public class ReferenceBuilder {
 	 * Builds a type reference from a qualified name when a type specified in the name isn't available.
 	 *
 	 * @param tokens        Qualified name.
+	 * @param receiverNode  AST node of last type in the qualified name.
 	 * @param receiverType  Last type in the qualified name.
 	 * @param enclosingType Enclosing type of the type name.
 	 * @param listener      Listener to know if we must build the type reference.
 	 * @return a type reference.
 	 */
-	<T> CtTypeReference<T> getQualifiedTypeReference(char[][] tokens, TypeBinding receiverType, ReferenceBinding enclosingType, JDTTreeBuilder.OnAccessListener listener) {
+	<T> CtTypeReference<T> getQualifiedTypeReference(char[][] tokens, ASTNode receiverNode, TypeBinding receiverType, ReferenceBinding enclosingType, JDTTreeBuilder.OnAccessListener listener) {
 		if (enclosingType != null && Collections.disjoint(PUBLIC_PROTECTED, JDTTreeBuilderQuery.getModifiers(enclosingType.modifiers, false, ModifierTarget.NONE))) {
 			String access = "";
 			int i = 0;
@@ -283,13 +284,18 @@ public class ReferenceBuilder {
 			final TypeBinding accessBinding = searchTypeBinding(access, units);
 			if (accessBinding != null && listener.onAccess(tokens, i)) {
 				final TypeBinding superClassBinding = searchTypeBinding(accessBinding.superclass(), CharOperation.charToString(tokens[i + 1]));
+				CtTypeReference<T> typeReference;
 				if (superClassBinding != null) {
-					return this.getTypeReference(superClassBinding.clone(accessBinding));
+					typeReference = this.getTypeReference(superClassBinding.clone(accessBinding));
 				} else {
-					return this.getTypeReference(receiverType);
+					typeReference = this.getTypeReference(receiverType);
 				}
+				typeReference.setPosition(this.jdtTreeBuilder.getPositionBuilder().buildPositionCtElement(typeReference, receiverNode));
+				return typeReference;
 			} else {
-				return this.getTypeReference(receiverType);
+				CtTypeReference<T> typeReference = this.getTypeReference(receiverType);
+				typeReference.setPosition(this.jdtTreeBuilder.getPositionBuilder().buildPositionCtElement(typeReference, receiverNode));
+				return typeReference;
 			}
 		}
 		return null;
--- a/src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
+++ b/src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
@@ -283,13 +283,11 @@ public class ReferenceBuilder {
 			final TypeBinding accessBinding = searchTypeBinding(access, units);
 			if (accessBinding != null && listener.onAccess(tokens, i)) {
 				final TypeBinding superClassBinding = searchTypeBinding(accessBinding.superclass(), CharOperation.charToString(tokens[i + 1]));
-				CtTypeReference<T> typeReference;
 				if (superClassBinding != null) {
-					typeReference = this.getTypeReference(superClassBinding.clone(accessBinding));
+					return this.getTypeReference(superClassBinding.clone(accessBinding));
 				} else {
-					typeReference = this.getTypeReference(receiverType);
+					return this.getTypeReference(receiverType);
 				}
-				return typeReference;
 			} else {
 				return this.getTypeReference(receiverType);
 			}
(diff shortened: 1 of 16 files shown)


algomaster99:
For now, I have created a test case that will be sufficient to fix Sorald's issue. However, that is not the best way. There are still some type references whose source position is not set. For example, when we execute this [test case](https://github.com/INRIA/spoon/blob/master/src/test/java/spoon/test/fieldaccesses/FieldAccessTest.java#L338), the model built misses out setting position of `Foo` (instance `of `CtTypeReference`) [is this method declaration](https://github.com/INRIA/spoon/blob/master/src/test/java/spoon/test/fieldaccesses/testclasses/Tacos.java#L18).

It is safe to say that this pull request will _only_ fix Sorald's issue. For type references like above, a further inspection is needed. I can open a ticket for that and deal with that later.

algomaster99:
@monperrus this is ready for review.

src/test/java/spoon/test/serializable/SourcePositionTest.java
```diff
@@ -71,4 +75,17 @@ public void testSourcePosition() throws IOException {
 		CtField<?> elem2 = typeFromFile.getField("a");
 		assertTrue(elem1.getPosition().getFile().equals(elem2.getPosition().getFile()));
 	}
+
+	@Test
+	public void test_sourcePositionOfNestedTypeInFieldExists() {
+		final Launcher launcher = new Launcher();
```
monperrus, line 81: could you add a contract? thanks!

monperrus:
Overall LGTM, @slarse do you want to look at it and merge?

src/test/java/spoon/test/serializable/SourcePositionTest.java
```diff
@@ -71,4 +76,18 @@ public void testSourcePosition() throws IOException {
...
 		assertTrue(elem1.getPosition().getFile().equals(elem2.getPosition().getFile()));
 	}
+
+	@Test
+	public void test_sourcePositionOfTypeInFieldExists() {
+		// contract: the type reference of type of the field should have a source position
+		final Launcher launcher = new Launcher();
+		launcher.addInputResource("./src/test/java/spoon/test/fieldaccesses/testclasses/SourcePosition.java");
+		launcher.addInputResource("./src/test/java/spoon/test/fieldaccesses/testclasses/SourcePartitionValidator.java");
+		CtModel model = launcher.buildModel();
```
MartinWitt, line 86: Could you move the test resources to `src/test/resources/...`

src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
```diff
@@ -260,12 +260,13 @@ private boolean isTypeArgumentExplicit(TypeReference[][] typeArguments) {
...
 	 *
 	 * @param tokens        Qualified name.
+	 * @param receiverNode  AST node of last type in the qualified name.
 	 * @param receiverType  Last type in the qualified name.
 	 * @param enclosingType Enclosing type of the type name.
 	 * @param listener      Listener to know if we must build the type reference.
 	 * @return a type reference.
 	 */
-	<T> CtTypeReference<T> getQualifiedTypeReference(char[][] tokens, TypeBinding receiverType, ReferenceBinding enclosingType, JDTTreeBuilder.OnAccessListener listener) {
+	<T> CtTypeReference<T> getQualifiedTypeReference(char[][] tokens, ASTNode receiverNode, TypeBinding receiverType, ReferenceBinding enclosingType, JDTTreeBuilder.OnAccessListener listener) {
```
MartinWitt, line 269: Adding a new parameter here is fine from my side, because this isn't real public api. Any other opinions?

src/test/java/spoon/test/serializable/SourcePositionTest.java
```diff
@@ -71,4 +76,18 @@ public void testSourcePosition() throws IOException {
...
+
+	@Test
+	public void test_sourcePositionOfTypeInFieldExists() {
+		// contract: the type reference of type of the field should have a source position
+		final Launcher launcher = new Launcher();
+		launcher.addInputResource("src/test/resources/spoon/test/sourcePosition/FieldType.java");
+		launcher.addInputResource("src/test/resources/spoon/test/sourcePosition/SourcePartitionValidator.java");
+		CtModel model = launcher.buildModel();
+
+		CtField field = (CtField) model.getElements(
```
MartinWitt, line 88: Could you add the wildcard `<?>` here to suppress unchecked warnings?

src/test/java/spoon/test/serializable/SourcePositionTest.java
```diff
@@ -71,4 +76,18 @@ public void testSourcePosition() throws IOException {
...
+				element -> element instanceof CtField &&
+						((CtField) element).getSimpleName().equals("pleaseAttachSourcePositionToMyType")).stream().findFirst().get();
+		assertThat(field.getType().getPosition(), not(instanceOf(NoSourcePosition.class)));
```
MartinWitt, line 91: Can`t you check here with https://spoon.gforge.inria.fr/mvnsites/spoon-core/apidocs/spoon/reflect/cu/SourcePosition.html#isValidPosition()?

MartinWitt (approved):
I added some small style comments otherwise looks fine for me. Adding a new parameter for a method is breaking behaviour, but the referencebuilder isn't part of the public api from my pov.

src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
algomaster99: I needed access to AST node to build the source position so for now, I just passed it as a parameter. I have opened an issue https://github.com/INRIA/spoon/issues/4297 to find a more generic solution to this but that would require how `CtTypeReference` are built and inserted into the model. This PR is kind of like a workaround but it fixes a very important issue in Sorald.

I will come back to https://github.com/INRIA/spoon/issues/4297 once I am done dealing with some important issues in sorald.

src/test/java/spoon/test/serializable/SourcePositionTest.java
algomaster99: Oh, right. I wasn't aware of this API method. Thanks!

algomaster99:
@MartinWitt incorporated your changes.

MartinWitt:
Merging from my side is fine. Let's wait the 24 hours and merge with `
fix(sniper): ensure sniper pretty-printer prints modifiers and type separated by a space` ?

slarse:
Having a peek.

src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
slarse: It's generally fine to add parameters to these internal methods (anything that isn't public, really), but here I think it's not the right solution. The added parameter is used in exactly the same way in two different places, which indicates to me that it should be used at the call site of the method rather than being passed in. So I think the code changes to this method should be reverted.

I think the place to solve this particular problem is in `JDTTreeBuilder.visit(QualifiedTypeReference, BlockScope)`, and change the very end of the method like so:

```diff
-		context.enter(factory.Code().createTypeAccessWithoutCloningReference(references.buildTypeReference(qualifiedTypeReference, scope)), qualifiedTypeReference);
+		CtTypeReference<?> typeRef = references.buildTypeReference(qualifiedTypeReference, scope);
+		if (typeRef != null) {
+			typeRef.setPosition(position.buildPositionCtElement(typeRef, qualifiedTypeReference));
+		}
+		context.enter(factory.Code().createTypeAccessWithoutCloningReference(typeRef), qualifiedTypeReference);
```

slarse (changes requested):
This was a rather tricky one, took me a while to evaluate it.

I don't think this is the right solution. The `ReferenceBuilder` isn't concerned with setting source positions, and for good reason: references are often implicit. For example, a variable read has a type, which is represented as a type reference, but the type is contained in the declaration of the variable, and not at the site of the variable read. For that reason, I think it's pretty important we deal with setting source positions further up the stack.

See comment on the source code.

algomaster99:
@slarse , yes that makes sense. Thanks for the patch. I think we will need to replicate such a thing for many type references and I have kept that for #4297 .

algomaster99:
@slarse This is causing some problems. I inspected [one of them](https://github.com/INRIA/spoon/runs/4284285393?check_suite_focus=true#step:10:514). It turns out that the patch tries to build a source position of the element `java.lang.Exception` which doesn't even exist in the test case resources. I am looking further into it.

algomaster99:
@nharrand need your opinions about disabling https://github.com/INRIA/spoon/pull/3359 's fix because it is interfering with setting source position for `Exception` [here](https://github.com/INRIA/spoon/blob/a9aa2296f4858baa6188955b9be63a849135d8de/src/test/java/spoon/test/annotation/testclasses/AnnotationsAppliedOnAnyTypeInAClass.java#L8). Reason:

> It tries to look up previous annotations while building source position for type reference (spoon element) java.lang.Exception however, sourceStart points to the beginning of the qualified type reference of the @TypeAnnotation and there are no annotations before that. findPrevAnnotations returns -1, and sourceStart is set to that, because it can't find any annotations before j in java.lang. ... @TypeAnnotation.

monperrus:
Works for me. Actions:

1. ignoring the failing assertion with a detailed comment
2.  puts high priority on #3358 on your backlog

algomaster99:
@monperrus I ignored the annotation position test for now. I will come back to it after we are done with sorald's paper submission.

monperrus:
ack, thanks @algomaster99

algomaster99:
@MartinWitt Something broke in the code quality test after this PR was merged. Can you take a look at it?

MartinWitt:
Sure, taking a look, but at the first glance this looks like some maven-central related problem.

algomaster99:
@MartinWitt thanks!

Issue #4290, opened 2021-11-16
Rewrite `SpoonPom.guessMavenHome` so that it correctly looks up the path to `mvn` executable


algomaster99:
`SpoonPom.guessMavenHome` abruptly stopped working on Windows in sorald (one of spoon's user). Its reason has been documented [here](https://github.com/INRIA/spoon/pull/4272#issuecomment-968293590) and for now, we have implemented a [workaround](https://github.com/SpoonLabs/sorald/pull/617) in sorald which lets our tests pass. However, we could re-implement `guessMavenHome` and overall improve [SpoonPom.java](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/compiler/SpoonPom.java) in terms of [error messages](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/compiler/SpoonPom.java#L568) and how it fetches the path. We can follow how it is implemented in [Java Language Server](https://github.com/georgewfraser/java-language-server/commit/40b76143b17f9b0fcd3d09106d621f45f6c61cea). This would perform the exact look up of the executable and won't rely on executing `mvn -version`.

slarse:
> SpoonPom.guessMavenHome abruptly stopped working on Windows in sorald (one of spoon's user).

Correction: It never worked on Windows in the presence of a `SecurityManager` :)

PR #4298, opened 2021-11-19
fix: lookup absolute path to maven executable

algomaster99:
Fixes https://github.com/INRIA/spoon/issues/4290

Initially, it fetched "maven home" which is the path to the installation directory of maven. I, instead, try to look up the absolute path of the maven executable. I saw that the path was used to [build class path ](https://github.com/INRIA/spoon/blob/a9aa2296f4858baa6188955b9be63a849135d8de/src/main/java/spoon/support/compiler/SpoonPom.java#L571) and we can directly use the executable there.

I shall open this PR for review once I test the patch in sorald on a Windows machine.

### DIFF
--- a/src/main/java/spoon/support/compiler/SpoonPom.java
+++ b/src/main/java/spoon/support/compiler/SpoonPom.java
@@ -29,7 +29,12 @@ import spoon.compiler.SpoonFolder;
 import spoon.compiler.SpoonResource;
 import spoon.compiler.SpoonResourceHelper;
 
-import java.io.*;
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
--- a/src/main/java/spoon/support/compiler/SpoonPom.java
+++ b/src/main/java/spoon/support/compiler/SpoonPom.java
@@ -48,7 +48,6 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 public class SpoonPom implements SpoonResource {
-	static String mavenVersionParsing = "Maven home: ";
 	static String spoonClasspathTmpFileName = "spoon.classpath.tmp";
 	static String spoonClasspathTmpFileNameApp = "spoon.classpath-app.tmp";
 	static String spoonClasspathTmpFileNameTest = "spoon.classpath-test.tmp";
@@ -517,39 +516,23 @@ public class SpoonPom implements SpoonResource {
 	}
 
 	/**
-	 * Try to guess Maven home when none is provided.
-	 * @return the path toward maven install on the local machine.
+	 * Lookup absolute path to maven executable.
+	 * @return the path toward maven executable on machine.
 	 */
 	public static String guessMavenHome() {
-		String mvnHome = null;
-		try {
-			String[] cmd;
-			if (System.getProperty("os.name").contains("Windows")) {
-				cmd = new String[]{"mvn.cmd", "-version"};
-			} else if (System.getProperty("os.name").contains("Mac")) {
-				cmd = new String[]{"sh", "-c", "mvn -version"};
-			} else {
-				cmd = new String[]{"mvn", "-version"};
-			}
-			Process p = Runtime.getRuntime().exec(cmd);
-			try (BufferedReader output = new BufferedReader(new InputStreamReader(p.getInputStream()))) {
-				String line;
-
-				while ((line = output.readLine()) != null) {
-					if (line.contains(mavenVersionParsing)) {
-						return line.replace(mavenVersionParsing, "");
-					}
-				}
-			}
-
-			p.waitFor();
-		} catch (IOException e) {
-			throw new SpoonException("Maven home detection has failed.");
-		} catch (InterruptedException e) {
-			Thread.currentThread().interrupt();
-			throw new SpoonException("Maven home detection was interrupted.");
+		String executableName;
+		if (System.getProperty("os.name").contains("Windows")) {
+			executableName = "mvn.cmd";
+		} else {
+			executableName = "mvn";
 		}
-		return mvnHome;
+		for (String dirname : System.getenv("PATH").split(File.pathSeparator)) {
+			File file = new File(dirname, executableName);
+			if (file.isFile() && file.canExecute()) {
+				return file.getAbsolutePath();
+			}
+		}
+		return null;
 	}
 
 	/**
--- a/src/main/java/spoon/support/compiler/SpoonPom.java
+++ b/src/main/java/spoon/support/compiler/SpoonPom.java
@@ -34,7 +34,6 @@ import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.FileReader;
 import java.io.IOException;
-import java.io.InputStreamReader;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
--- a/src/main/java/spoon/support/compiler/SpoonPom.java
+++ b/src/main/java/spoon/support/compiler/SpoonPom.java
@@ -29,11 +29,7 @@ import spoon.compiler.SpoonFolder;
 import spoon.compiler.SpoonResource;
 import spoon.compiler.SpoonResourceHelper;
 
-import java.io.BufferedReader;
-import java.io.File;
-import java.io.FileNotFoundException;
-import java.io.FileReader;
-import java.io.IOException;
+import java.io.*;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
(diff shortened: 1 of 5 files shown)


algomaster99:
@slarse Qodana tells [what the error is](https://github.com/INRIA/spoon/runs/4265820752?check_suite_focus=true#step:8:188), but I am not sure where to look at it. What is the source position of the error? Can we figure that out?

MartinWitt:
There was an update to qodana, and we had a solution planed. I pushed the (hopefully) fix to your branch as a tempoary solution. I hope to create a PR this weekend fixing this with a GitHub action created by @SirYwell

MartinWitt:
Qodana doesn't like the return of null instead of throwing an exception. Returning null instead of a String seems wrong. Maybe add an exception instead of null?

src/main/java/spoon/support/compiler/SpoonPom.java
```diff
@@ -564,9 +547,6 @@ public static String guessMavenHome() {
 	public String[] buildClassPath(String mvnHome, MavenLauncher.SOURCE_TYPE sourceType, Logger LOGGER, boolean forceRefresh) {
 		if (mvnHome == null) {
 			mvnHome = guessMavenHome();
-			if (mvnHome == null) {
-				throw new SpoonException("M2_HOME must be initialized to use this MavenLauncher constructor.");
-			}
```
algomaster99, line 569: No longer needed because `guessMavenHome` throws an exception if it does not return a string.

algomaster99:
@MartinWitt 
> Qodana doesn't like the return of null instead of throwing an exception.

That shouldn't be the case because `null` is returned by a lot of methods in Spoon. Moreover, severity level of "Return of `null`" in the job is "warning" so I don't think that the tests would _fail_ because of returning `null`. Regardless, I threw an exception instead, but it didn't work.

Opening this PR for review as I have tested this patch in Sorald on a Windows laptop. We would no longer need [this workaround](https://github.com/SpoonLabs/sorald/pull/617) there.

MartinWitt:
@I-Al-Istannen had the fix, and he is preparing a separate PR for this.

algomaster99:
> Returning null is a bad smell, either we return Optional or throw Exceptions

I agree that returning `null` is a bad smell. However, we had PRs passing quality checks after qodana was configured even though `return null` existed in the codebase.

> We should try to get away from it.

Completely agree.

Thanks for the fix! Do you plan to review it now or should we wait for @I-Al-Istannen to submit the PR and then I will incorporate that patch into this PR?

src/main/java/spoon/support/compiler/SpoonPom.java
```diff
@@ -517,39 +516,23 @@ private void generateClassPathFile(File mvnHome, MavenLauncher.SOURCE_TYPE sourc
...
+		String executableName;
+		if (System.getProperty("os.name").contains("Windows")) {
+			executableName = "mvn.cmd";
+		} else {
+			executableName = "mvn";
+		}
+		for (String dirname : System.getenv("PATH").split(File.pathSeparator)) {
+			File file = new File(dirname, executableName);
+			if (file.isFile() && file.canExecute()) {
+				return file.getAbsolutePath();
```
slarse, line 532: This doesn't seem right to me. "Maven home" is the path to the Maven install directory, not the path to the Maven executable.

Does this actually work for running the `MavenLauncher`?

I would also advice extracting this into a separate method ('Path getExecutablePath(String name)`) or something like that, and throw the exception from there instead.

algomaster99: @slarse `Maven home` and `Maven executable` are indeed different things. I tried to look up Maven home in PATH, but it did not exist, so I checked how the path to maven home is being used in the codebase. I found out that there were only two places.
1. Used to [generate a classpath file](https://github.com/INRIA/spoon/blob/a9aa2296f4858baa6188955b9be63a849135d8de/src/main/java/spoon/support/compiler/SpoonPom.java#L571). I did not meticulously read the method definition, but it seems like it tries to build the classpath if it doesn't exist.
2. Used to [run checkstyle](https://github.com/INRIA/spoon/blob/a9aa2296f4858baa6188955b9be63a849135d8de/src/test/java/spoon/test/prettyprinter/DefaultPrettyPrinterTest.java#L525) in this test, although it is ignored.

From these two usages, I felt that looking up the path to executable would suffice, and maven home is not actually needed. 

> Does this actually work for running the MavenLauncher?

Yes, it does, and this corroborated my patch.

algomaster99: I decided to keep the intended behaviour. Pushed the changes for it.

algomaster99:
@monperrus ping for reviews.

algomaster99:
@MartinWitt I am extracting your changes to the workflow in a separate PR.

MartinWitt:
We have a PR already here, https://github.com/INRIA/spoon/pull/4301. I thought @slarse is merging it after his comment.

src/test/java/spoon/MavenLauncherTest.java
```diff
@@ -208,12 +208,11 @@ public void mavenLauncherTestMultiModulesAndVariables() {
...
-		String mvnHome = (String) method.invoke(null, new Object[]{});
-		File mvnDir = new File(mvnHome);
-		assertTrue(mvnDir.exists());
-		assertTrue(mvnDir.isDirectory());
+	public void testGuessMavenHome() {
+		// contract: it should correctly fetch path to maven home
+		String pathToMavenHome = SpoonPom.guessMavenHome();
+		File mavenHome = new File(pathToMavenHome);
+		assertTrue(mavenHome.exists());
+		assertTrue(mavenHome.isDirectory());
```
algomaster99, line 216: Please note that I also removed using reflection to invoke `guessMavenHome`.

algomaster99:
@monperrus ping for merge.

slarse:
> We have a PR already here, #4301. I thought @slarse is merging it after his comment.

Well, that was the idea. But I forgot and this week is a bit heavy.

Will be AWOL until the weekend, someone else tag in for this PR or I'll get to it on Saturday.

src/main/java/spoon/support/compiler/SpoonPom.java
```diff
@@ -541,15 +535,30 @@ public static String guessMavenHome() {
...
 		}
-		return mvnHome;
+		throw new SpoonException("Couldn't find path to maven home.");
+	}
+
+	private static String getPathToMavenExecutable() {
+		String executableName;
+		if (System.getProperty("os.name").contains("Windows")) {
+			executableName = "mvn.cmd";
+		} else {
```
MartinWitt, line 552: Does this work on Mac? You removed the Mac case?

src/test/java/spoon/MavenLauncherTest.java
```diff
@@ -208,12 +208,11 @@ public void mavenLauncherTestMultiModulesAndVariables() {
...
-		method.setAccessible(true);
-		String mvnHome = (String) method.invoke(null, new Object[]{});
-		File mvnDir = new File(mvnHome);
-		assertTrue(mvnDir.exists());
-		assertTrue(mvnDir.isDirectory());
+	public void testGuessMavenHome() {
+		// contract: it should correctly fetch path to maven home
+		String pathToMavenHome = SpoonPom.guessMavenHome();
+		File mavenHome = new File(pathToMavenHome);
+		assertTrue(mavenHome.exists());
```
MartinWitt, line 215: Can you add here a String explaining what failed? The error message of `assertTrue` is kinda bad.

src/main/java/spoon/support/compiler/SpoonPom.java
algomaster99: Yes. I tested it [here](https://github.com/SpoonLabs/sorald/runs/4309903908?check_suite_focus=true#step:5:8).

1. I get the absolute path to the maven executable as I have done in this PR.
2. Execute `</absolute/path/to/maven> -version`.

src/test/java/spoon/MavenLauncherTest.java
algomaster99: I throw a `SpoonException` in `guessMavenHome` (message: Couldn't find path to maven home.). We will never reach this line if the test fails.

Is this the correct to way to write this test because it doesn't make sense to `assertDoesNotThrow`?

monperrus: Agree, assertTrue with message is a best practice, but in this case we can proceed.

Keeping trace of this in good first issue #4307

monperrus:
Thanks a lot @algomaster99 for the solution and @slarse @MartinWitt for the review.

src/test/java/spoon/MavenLauncherTest.java
slarse: I've been thinking to introduce the full range of Hamcrest matchers into Spoon, they truly bring some wonderful expressiveness to tests. The perfect matcher here would be `aFileNamed()`.

algomaster99: Completely agree. Moreover, it would also help refactor our tests. We currently use Junit4, Junit5, and hamcrest. It's best to have just one library handling it all.

slarse: Hamcrest is just the matchers, it's not a testing framework. So, it would be Junit5 + Hamcrest matchers.

PR #4301, opened 2021-11-20
review: fix: Ignore Qodana action exit code

I-Al-Istannen:
This PR just ignores failures in the qodana action (as its exit codes were a bit weird) and instead relies on the jq buildbreaker.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -167,9 +167,7 @@ jobs:
           results-dir:  "${{ github.workspace }}/result/master"
       - name: Qodana - Code Inspection
         uses: JetBrains/qodana-action@cd7dc31b49427ce740ebf6255319b267801d08cf # renovate: tag=v3.2.1
-        # the action's exit codes are a bit unreliable at the moment
         continue-on-error: true
-        id: qodana # later used by the sarif annotator step
         with:
           linter: qodana-jvm-community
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -167,6 +167,7 @@ jobs:
           results-dir:  "${{ github.workspace }}/result/master"
       - name: Qodana - Code Inspection
         uses: JetBrains/qodana-action@cd7dc31b49427ce740ebf6255319b267801d08cf # renovate: tag=v3.2.1
+        continue-on-error: true
```
algomaster99, line 170: Doesn't this make usage of qodana-action redundant? Even if it fails, we shall ignore the exit code and still move on with merging a PR.

I-Al-Istannen: Not quite, as the print (short) step later does the actual checking and breaks the build.

slarse (approved):
LGTM, will merge.

algomaster99:
@monperrus This PR has changes that need to be excluded from https://github.com/INRIA/spoon/pull/4298. Please merge it.

MartinWitt:
@I-Al-Istannen thanks for the fix.

slarse:
I of course meant LGTM, will forget to merge.

monperrus:
Thanks for the great group work.

PR #4303, opened 2021-11-20
fix: NPE in Launcher.prettyprint() (#4302)

### LINKED ISSUE #4302: [Bug] NPE in Launcher.prettyPrint() due to not checking return of File.getParent()
**Describe the bug**
Program throws NPE in `Launcher.prettyPrint()` when the path to the current resource has no parent.

**To Reproduce**

Input:
I think the actual input doesn't matter much because the problem lies in manipulating Strings to create paths instead of using the JDK classes for that purpose. I ran into this bug with the following input:
```java
package nested_generic_class;

import java.util.function.Consumer;

public class Foo implements Consumer<Foo.Bar> {
    enum Bar {
        A, B;
    }

    @Override
    public void accept(Bar bar) {
        System.out.println(bar.name());
    }
}
```

Processing with Spoon:
```java
        launcher.getEnvironment().setPreserveLineNumbers(true);
        
        launcher.prettyprint();
```

Output:
```java
Exception in thread "main" java.lang.NullPointerException: Cannot invoke "String.substring(int)" because "resourceParentPath" is null
	at spoon.Launcher.prettyprint(Launcher.java:816)
	at spoon_issues.Main.main(Main.java:27)
```

* OS: MacOS 12.0.1 (Monterey)
* JDK: 17 - OpenJDK Runtime Environment GraalVM CE 21.3.0 (build 17.0.1+12-jvmci-21.3-b05)
* Spoon version: current Github

xzel23:
fix #4302


SirYwell:
You should probably use JUnit 5 (the `@Test` annotation needs to be fully qualified like `@org.junit.jupiter.api.Test` then) and `assertDoesNotThrow(launcher::prettyprint);` instead of the try catch with the `assertTrue`.

There is also a `DefaultJavaPrettyPrinterTest` class (note the `Java`) which already uses JUnit 5, but I'm not sure if it belongs there...

xzel23:
Ok, will Look into this later.

slarse:
For a test class that uses JUnit4, it's fine to add more JUnit4 tests. That actually makes migrating easier than if we mix.

As the test is for the `Launcher` class (it really has nothing to do with the actual printing anyway), the test should go into `LauncherTest`.

Note that I've made no review here, I just looked at the files changed

@MartinWitt can you review this PR?

src/test/java/spoon/test/prettyprinter/DefaultPrettyPrinterTest.java
```diff
@@ -475,6 +477,28 @@ public void testElseIf() {
...
+
+		boolean done;
+		try {
+			Launcher launcher = new Launcher();
+			launcher.addInputResource(new VirtualFile(code));
+			launcher.getEnvironment().setNoClasspath(true);
+			launcher.getEnvironment().setAutoImports(true);
+			launcher.getEnvironment().setPreserveLineNumbers(true);
+	
+			launcher.prettyprint();
```
MartinWitt, line 493: here you should use `assertDoesNotThrow(() -> Launcher.prettyPrint())`

MartinWitt (changes requested):

xzel23:
I have changed the test to use assertDoesNotThrow(). In a second commit, I have moved the test from DefaultPrettyPrinterTest to LauncherTest as suggested by slarse. I noticed that the test failed due to Spoon trying to write to an unwritable directory when I simply move the test method, so I have set the source output directory to a temp dir.

src/test/java/spoon/LauncherTest.java
```diff
@@ -119,4 +121,22 @@ public void testLLauncherBuildModelReturnAModel() {
...
+
+		Launcher launcher = new Launcher();
+		launcher.addInputResource(new VirtualFile(code));
+		launcher.getEnvironment().setNoClasspath(true);
+		launcher.getEnvironment().setAutoImports(true);
+		launcher.getEnvironment().setPreserveLineNumbers(true);
+
+		File tmpDir = Files.createTempDirectory("spoonTestPrettyPrintWithVirtualFileInput").toFile();
+		tmpDir.deleteOnExit();
+		launcher.setSourceOutputDirectory(tmpDir);
```
MartinWitt, line 138: https://spoon.gforge.inria.fr/mvnsites/spoon-core/apidocs/spoon/compiler/Environment.html#setOutputType(spoon.OutputType) Here you can simply set the `OutputType` to none, which is better than creating temp files.

src/test/java/spoon/test/prettyprinter/DefaultPrettyPrinterTest.java
```diff
@@ -86,6 +82,7 @@
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
```
MartinWitt, line 85: Can you restore the file to the original state, for a minimal diff?

MartinWitt:
LGTM will merge after 24 hours.

slarse:
@MartinWitt You don't need to wait 24 hours after the final change, but from the point that the PR was opened. This is more to let other integrators have their say about the intent of the PR rather than the final content. After that we trust the judgement of individual integrators.

Of course, if there are multiple reviewers on a PR, then it may be good to wait for the other(s) to cast their votes. But otherwise, you're free to merge at your discretion after the PR has been open for 24 hours.

MartinWitt:
Thanks @xzel23

PR #4306, opened 2021-11-24
fix: prevent appending semicolon after enum constant if it is not there initially

algomaster99:
Printing enum constants with sniper printer enforces `;` at the end even though it is not present before. For example,
```java
public enum Days {
    SUNDAY, MONDAY, TUESDAY, WEDNESDAY,
    THURSDAY, FRIDAY, SATURDAY
}
```
is changed to
```java
public enum Days {
    SUNDAY, MONDAY, TUESDAY, WEDNESDAY,
    THURSDAY, FRIDAY, SATURDAY;
}
```

`enum` is compilable with or without the `;` if there are no method declarations afterwards. Source: https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html.
> when there are fields and methods, the list of enum constants must end with a semicolon.

I think we should configure the sniper printer to check for `;` and print it only if required. The current behaviour is not causing a compilation error, but it helps minimise the diff so we should consider it.

### DIFF
--- a/src/test/java/spoon/test/prettyprinter/TestSniperPrinter.java
+++ b/src/test/java/spoon/test/prettyprinter/TestSniperPrinter.java
@@ -841,6 +841,17 @@ public class TestSniperPrinter {
 		testSniper("sniperPrint.SpaceAfterFinal", modifyField, assertContainsSpaceAfterFinal);
 	}
 
+	@Test
+	void testSniperShouldNotAddSemicolonAfterEnumIfItIsNotPresent() {
+		Consumer<CtType<?>> noOpModifyFieldAssignment = type ->
+				type.descendantIterator().forEachRemaining(TestSniperPrinter::markElementForSniperPrinting);
+
+		BiConsumer<CtType<?>, String> assertPrintsRoundBracketsCorrectly = (type, result) ->
+				assertThat(result, not(containsString("SATURDAY;")));
+
+		testSniper("sniperPrinter.Days", noOpModifyFieldAssignment, assertPrintsRoundBracketsCorrectly);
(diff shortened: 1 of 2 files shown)


algomaster99:
I have figured out [where](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java#L760) this happens. However, I am not sure how to fix this yet. I know that this fix would go into SniperJavaPrettyPrinter. This is the place where we [build the context](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/sniper/SniperJavaPrettyPrinter.java#L71), but it spoils the genericity if we also check for such an edge case. @slarse @MartinWitt @nharrand , your thoughts?

slarse:
The place where most of the sniper printing magic happens is in `AbstractSourceFragmentPrinter::print`, and the methods transitively called from there. That's usually where I start debugging when I'm trying to figure stuff like this out.

algomaster99:
@slarse thanks for the pointers.

Keeping this in my backlog.

algomaster99:
Debugged this further. I think the following is a wrong assumption.
https://github.com/INRIA/spoon/blob/fa8f97c6d1d78e6e8657e5aa4dd4b45bbaa05a92/src/main/java/spoon/support/sniper/internal/AbstractSourceFragmentPrinter.java#L90-L93

The comment says that the token is not present in the original sources, however, it always prints them. We need to have a way that prevents that.

slarse:
Closing this as stale per the contributing guidelines, feel free to reopen if you can invest time in this again.

PR #4309, opened 2021-11-25
fix: print each resource in `CtTryWithResource` exactly once and retain separator

algomaster99:
Reference: https://github.com/SpoonLabs/sorald/issues/600

The second resource in the catch block, `BufferedWriter writer = newBufferedWriter(outputFilePath, charset)`, is printed twice. For example,

```java
        try (ZipFile zf = new ZipFile(zipFileName);
             BufferedWriter writer = newBufferedWriter(outputFilePath, charset)) { }
```
is changed to
```java
        try (ZipFile zf = new ZipFile(zipFileName);
             BufferedWriter writer = newBufferedWriter(outputFilePath, charset);BufferedWriter writer = newBufferedWriter(outputFilePath, charset)) { }
```
when printed using the sniper printer.

I think it is happening because we add a separator, `;`, after first `BufferedWriter writer = newBufferedWriter(outputFilePath, charset)` which changes `muted` from `true` to `false`. Although the separator is not needed here, it is still printed. This may be similar to #4306 .

Please note that `muted` should remain true till we print the second resource because all resources are [printed at once](https://github.com/INRIA/spoon/blob/f898632bbdaf5c671b74a652c04d10dfc65328c3/src/main/java/spoon/support/sniper/SniperJavaPrettyPrinter.java#L327) because of this line [here](https://github.com/INRIA/spoon/blob/f898632bbdaf5c671b74a652c04d10dfc65328c3/src/main/java/spoon/support/sniper/SniperJavaPrettyPrinter.java#L327). Basically, if a collection fragment is not modified, we print it at once and set `muted` to `true` for upcoming elements in the collection.

### DIFF
--- a/src/main/java/spoon/support/sniper/internal/AbstractSourceFragmentContextCollection.java
+++ b/src/main/java/spoon/support/sniper/internal/AbstractSourceFragmentContextCollection.java
@@ -8,7 +8,6 @@
 package spoon.support.sniper.internal;
 
 import java.util.List;
-import java.util.Optional;
 
 import spoon.SpoonException;
 import spoon.reflect.declaration.CtCompilationUnit;
@@ -53,7 +52,10 @@ abstract class AbstractSourceFragmentContextCollection extends AbstractSourceFra
 				return false;
 			} else if (tpe.getType() == TokenType.IDENTIFIER) {
 				return findIndexOfNextChildTokenByType(TokenType.IDENTIFIER) >= 0;
-			} else if (tpe.getToken().equals(";") && childFragmentHasSpecifiedRoleInParent(CtRole.TRY_RESOURCE)) {
+			}
+			// We check for ; printed using printList in DJPP#visitCtTryWithResource because if we do not, the context
+			// gets popped from the stack and printer messes up printing of the latter elements.
+			else if (tpe.getToken().equals(";") && anyChildFragmentHasRole(CtRole.TRY_RESOURCE)) {
 				return true;
 			}
 			return findIndexOfNextChildTokenByValue(tpe.getToken()) >= 0;
@@ -71,15 +73,14 @@ abstract class AbstractSourceFragmentContextCollection extends AbstractSourceFra
 		throw new SpoonException("Unexpected PrintEvent: " + event.getClass());
 	}
 
-	private boolean childFragmentHasSpecifiedRoleInParent(CtRole roleInParent) {
-		Optional<SourceFragment> optionSourceFragment = childFragments.stream()
-				.filter(fragment -> fragment instanceof ElementSourceFragment)
-				.findFirst();
-		if (optionSourceFragment.isPresent()) {
-			ElementSourceFragment elementSourceFragment = (ElementSourceFragment) optionSourceFragment.get();
-			return  elementSourceFragment.getRoleInParent() == roleInParent;
-		}
-		return false;
+	private boolean anyChildFragmentHasRole(CtRole role) {
+		return childFragments.stream()
+				.filter(ElementSourceFragment.class::isInstance)
+				.map(ElementSourceFragment.class::cast)
+				.map(ElementSourceFragment::getRoleInParent)
+				.map(role::equals)
+				.findAny()
+				.orElse(false);
 	}
 
 	@Override
--- a/src/main/java/spoon/support/sniper/internal/AbstractSourceFragmentContextCollection.java
+++ b/src/main/java/spoon/support/sniper/internal/AbstractSourceFragmentContextCollection.java
@@ -52,11 +52,10 @@ abstract class AbstractSourceFragmentContextCollection extends AbstractSourceFra
 				return false;
 			} else if (tpe.getType() == TokenType.IDENTIFIER) {
 				return findIndexOfNextChildTokenByType(TokenType.IDENTIFIER) >= 0;
-			}
-			// We check for ; printed using printList in DJPP#visitCtTryWithResource because if we do not, the context
-			// gets popped from the stack and printer messes up printing of the latter elements.
-			// See: https://github.com/INRIA/spoon/pull/4309.
-			else if (tpe.getToken().equals(";") && anyChildFragmentHasRole(CtRole.TRY_RESOURCE)) {
+			} else if (tpe.getToken().equals(";") && anyChildFragmentHasRole(CtRole.TRY_RESOURCE)) {
+				// We check for ; printed using printList in DJPP#visitCtTryWithResource because if we do not, the context
+				// gets popped from the stack and printer messes up printing of the latter elements.
+				// See: https://github.com/INRIA/spoon/pull/4309.
 				return true;
 			}
 			return findIndexOfNextChildTokenByValue(tpe.getToken()) >= 0;
--- a/src/main/java/spoon/support/sniper/internal/AbstractSourceFragmentContextCollection.java
+++ b/src/main/java/spoon/support/sniper/internal/AbstractSourceFragmentContextCollection.java
@@ -8,7 +8,7 @@
 package spoon.support.sniper.internal;
 
 import java.util.List;
-import java.util.stream.Collectors;
+import java.util.Optional;
 
 import spoon.SpoonException;
 import spoon.reflect.declaration.CtCompilationUnit;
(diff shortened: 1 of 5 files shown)


algomaster99:
The context is pushed before printing the resources, however that same context is [not used when we are printing the separator](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/sniper/SniperJavaPrettyPrinter.java#L380). Apparently, `sfc.knowsHowToPrint` returns `false` when we are printing `;`. Probably, the problem lies there.

algomaster99:
> The context is pushed before printing the resources, however that same context is not used when we are printing the separator. Apparently, sfc.knowsHowToPrint returns false when we are printing ;. Probably, the problem lies there.

I went along these lines and pushed a fix for it.

I debugged where the context was lost and it was in `knowsHowToPrint` method. With the changes, I am ensuring that context doesn't get lost if we are parsing a `;` and we are printing resources. I have a feeling that such a fix will also help patch #4306 .

One major problem I see in this patch is that I am hard-coding `;` in an _abstract_ code that doesn't sound logically coherent. There is a probably a better way to do it and I hope to discover that through reviews.

@slarse @MartinWitt @monperrus , could you please review it?

src/main/java/spoon/support/sniper/internal/AbstractSourceFragmentContextCollection.java
```diff
@@ -68,6 +73,17 @@ public boolean knowsHowToPrint(PrinterEvent event) {
...
+	private boolean checkIfPartOfRole(CtRole role) {
+		List<SourceFragment> elementSourceFragments = childFragments.stream()
+				.filter(fragment -> fragment instanceof ElementSourceFragment)
+				.collect(Collectors.toList());
+		for (SourceFragment sourceFragment: elementSourceFragments) {
+			ElementSourceFragment elementSourceFragment = ((ElementSourceFragment) sourceFragment);
+			return elementSourceFragment.getRoleInParent() == role;
+		}
+		return false;
+	}
```
MartinWitt, line 85: This is really hard to read and should be refactored in a single loop. A loop that always returns and doesn't loop is kinda unexpected.

src/main/java/spoon/support/sniper/internal/AbstractSourceFragmentContextCollection.java
```diff
@@ -52,6 +53,10 @@ public boolean knowsHowToPrint(PrinterEvent event) {
 				return false;
 			} else if (tpe.getType() == TokenType.IDENTIFIER) {
 				return findIndexOfNextChildTokenByType(TokenType.IDENTIFIER) >= 0;
+			} else if (tpe.getToken().equals(";")) {
+				if (checkIfPartOfRole(CtRole.TRY_RESOURCE)) {
```
slarse, line 57: Collapse

```suggestion
			} else if (tpe.getToken().equals(";") && checkIfPartOfRole(CtRole.TRY_RESOURCE)) {
```

src/main/java/spoon/support/sniper/internal/AbstractSourceFragmentContextCollection.java
```diff
@@ -68,6 +73,17 @@ public boolean knowsHowToPrint(PrinterEvent event) {
 		throw new SpoonException("Unexpected PrintEvent: " + event.getClass());
 	}
 
+	private boolean checkIfPartOfRole(CtRole role) {
```
slarse, line 76: Just reading the invocation of this method, I have no idea of what its intention is. Check if what is part of a role? What does it mean to be part of a role?

`check` is also usually reserved for methods that throw. For methods that are meant to return a boolean, I'd recommend starting with `is`, `has` etc.

Looking at the method body, this method returns true if the first `ElementSourceFragment` of `childFragments` matches the provided role. Otherwise it returns false. Can you explain those semantics? I also agree with @MartinWitt that the method needs refactoring, but first I want to try to understand its intent.

slarse (changes requested):

algomaster99:
@MartinWitt @slarse The intent of the method is to check if the role of the current `ElementSourceFragment`s of `childFragments` is the passed `role` (in our case - `CtRole.TRY_RESOURCE`). 

> Can you explain those semantics?

I assumed that if the first element has the role `CtRole.TRY_RESOURCE`, then the rest elements will also have the same role and I guess it's safe to do so because you can only used objects inside it which implement `java.lang.AutoCloseable`. Thus, making all of them `TRY_RESOURCE`.

I have explained why this code was required in the first place [here](https://github.com/INRIA/spoon/pull/4309#issuecomment-980028086).

algomaster99:
I have refactored the method to what @slarse and @MartinWitt suggested. I named the method `childFragmentHasSpecifiedRole` because I think it describes the intent cleary although it doesn't start with `has`, `does`, or `is`.

src/main/java/spoon/support/sniper/internal/AbstractSourceFragmentContextCollection.java
```diff
@@ -68,6 +71,17 @@ public boolean knowsHowToPrint(PrinterEvent event) {
...
+	private boolean childFragmentHasSpecifiedRoleInParent(CtRole roleInParent) {
+		Optional<SourceFragment> optionSourceFragment = childFragments.stream()
+				.filter(fragment -> fragment instanceof ElementSourceFragment)
+				.findFirst();
+		if (optionSourceFragment.isPresent()) {
+			ElementSourceFragment elementSourceFragment = (ElementSourceFragment) optionSourceFragment.get();
+			return  elementSourceFragment.getRoleInParent() == roleInParent;
+		}
+		return false;
+	}
```
slarse, line 83: > I assumed that if the first element has the role CtRole.TRY_RESOURCE, then the rest elements will also have the same role and I guess it's safe to do so because you can only used objects inside it which implement

That's probably safe, yes. The thingy that builds the `CollectionSourceFragment` (which is later fed into the collection context) groups the children by role. I don't have a great understanding of how often this check will run, but I don't think a solitary `;` is printed very often. In the entire sniper printer unit test suite, it only happens 4 times, 2 of which are in your tests. So probably this is fine from a performance perspective, but it bears keeping in mind.

Reading the method title, I'm still thinking "which child fragment?". And the answer to that is of course _any_ child fragment. I would therefore suggest a renaming like below, and also further idiomatic use of the stream.

```suggestion
	private boolean anyChildFragmentHasRole(CtRole role) {
		return childFragments.stream()
				.filter(ElementSourceFragment.class::isInstance)
				.map(ElementSourceFragment.class::cast)
				.map(ElementSourceFragment::getRoleInParent)
				.map(role::equals)
				.findAny()
				.orElse(false);
	}
```

I'm using `findAny()` here to emphasize that there's nothing special about the first element, we just assume that if any element has a given role, the others do too.

slarse (changes requested):
Added a further suggestion for clarification now that I understand the intent.

src/main/java/spoon/support/sniper/internal/AbstractSourceFragmentContextCollection.java
```diff
@@ -52,6 +53,8 @@ public boolean knowsHowToPrint(PrinterEvent event) {
 				return false;
 			} else if (tpe.getType() == TokenType.IDENTIFIER) {
 				return findIndexOfNextChildTokenByType(TokenType.IDENTIFIER) >= 0;
+			} else if (tpe.getToken().equals(";") && childFragmentHasSpecifiedRoleInParent(CtRole.TRY_RESOURCE)) {
```
slarse, line 56: With the suggested renaming, this becomes somewhat easier to interpret, I think:

```suggestion
			} else if (tpe.getToken().equals(";") && anyChildFragmentHasRole(CtRole.TRY_RESOURCE)) {
```

algomaster99:
> So probably this is fine from a performance perspective, but it bears keeping in mind.

We can add a comment for this because it might confound future developers to see such a thing. It's true for future-me at least.

> And the answer to that is of course any child fragment.

Right. We could do this with `anyMatch`. Thank you for refactoring the stream. I often get confused about how to club the intermediate operations. I am working on it.

I will push the changes.

algomaster99:
@slarse @MartinWitt incorporated your changes. I have also added a short comment explaining the `else-if` block. Please review.

algomaster99:
@slarse reminder to merge this PR, if you are okay with the changes. :)

slarse (approved):

PR #4312, opened 2021-11-26
doc: fix spelling of @slarse 's surname

algomaster99:
I have always wondered why there is an `é` in your name even though this alphabet is not part of the Swedish lexicon.

Anyway, please proceed to merge if this is indeed the correct spelling. :)


slarse:
It's not a special character, really, it's just an e with an acute accent. It indicates pronunciation, but is otherwise relatively unimportant.

But yes it's correct so I'll merge :)

PR #4318, opened 2021-11-30
fix: print `abstract` modifier once and with proper spaces

algomaster99:
This is an extremely weird issue because of the randomness involved.

If you directly run this test, it will fail because it modifies the file as such:
```diff
package sniperPrinter;

public class PrintAbstractOnceWithProperSpacing {
-     static abstract class AbstractFoo
+     abstractstatic abstract class AbstractFoo
    {

    }
}
```

However, when you want to debug why this happens, the test gets a bit flaky. In other words, sometimes it passes and sometimes it fails.

If you directly press debug, without setting any breakpoints in code, the tests sometimes fail and it passes the other times. I haven't been able to figure out what is going wrong simply because of the random behaviour.

Now if you set a breakpoint at [here](https://github.com/INRIA/spoon/blob/ab1edeeb1b9ef4bfdc4f900acf0160017214dc8f/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java#L1699), the program will stop there twice - 1) for printing modifiers of `PrintAbstractOnceWithProperSpacing` 2) for printing modifiers of `AbstractFoo`. Just press "Resume Program (F9)" twice and this test will **pass** :exploding_head: .

@khaes-kth is suggesting that it probably has to do how the program is executed in threads. Could you please elaborate?


algomaster99:
@MartinWitt @slarse have a look at this when you get time.

algomaster99:
Screencast of the flaky behaviour:
abstract-static

khaes-kth:
Cool and worrying screenshot :)

I-Al-Istannen:
I have absolutely no idea how anything works but this issue seems to stem from the undefined order of modifiers in `CtModifierHandle`. In `ElementPrinterHelper#writeModifiers` the `secondPosition` list is either `[static, abstract]` in which case it can match that onto the source (`static abstract class`) or `[abstract, static]` in which case it can not and creates that garbled output.

The order seems to depend on the other of an unspecified set in the `CtModifierHandle` which *probably* ends up depending on the hashcode order of the underlying enum, which varies from JVM start to JVM start and hence creates some flakey output. In my unlucky periods I hit that issue after retrying 10 or more times, sometimes instantly on the first try but it was always reproducible.

Maybe this can serve as a starting point for Slarse or Martin :)

slarse:
Probably related to #4033

monperrus:
conflict here

algomaster99:
@monperrus I have reproduced the issue in this PR. I have not worked on fix for this.

algomaster99:
I will restart work on this when I feel it's blocking something because it is a tough one to deal with.

PR #4322, opened 2021-12-01
refactor: replace Collections.unmodifiableList with List.of to fix Qodana warnings

MartinWitt:
Because we finally have Java 11, we can use the newer API. Fixes 2 qodana warnings.


MartinWitt:
Maybe I run in code I never wanted to see.

monperrus:
thanks @MartinWitt for looking at Qodana problems.

PR #4324, opened 2021-12-01
refactor: replace with bulk operation

MartinWitt:
Replace for loop with new bulk operation


monperrus:
Thanks @MartinWitt 

An example of code simplification @fermadeiral @lyxell

PR #4325, opened 2021-12-05
refactor: remove useless defensive programming

MartinWitt:
fixes qodana issue

### DIFF
--- a/src/main/java/spoon/support/util/ModelSet.java
+++ b/src/main/java/spoon/support/util/ModelSet.java
@@ -84,10 +84,7 @@ public abstract class ModelSet<T extends CtElement> extends AbstractSet<T> imple
 
 		// we make sure that the element is always the last put in the set
 		// for being least suprising for client code
-		if (set.contains(e)) {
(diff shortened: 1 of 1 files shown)


src/main/java/spoon/support/util/ModelSet.java
```diff
@@ -84,10 +84,7 @@ public boolean add(T e) {
 
 		// we make sure that the element is always the last put in the set
 		// for being least suprising for client code
```
algomaster99, line 86: Maybe we can remove these comments as well? I don't know why were they added in the first place. The order is anyway controlled by the comparator.

monperrus:
LGTM, will merge

src/main/java/spoon/support/util/ModelSet.java
MartinWitt: Spoon does a lot of unholy things with sets and relaying on their order. I would leave the comment there, because it helps to understand this madness.

monperrus:
thanks @MartinWitt

PR #4327, opened 2021-12-06
Wip: refactor: replace old stream api with new

MartinWitt:
Fix qodana issue. As always, we had the neat case of nulls in Collections(why?). I removed them because this is never the wanted behaviour.


MartinWitt:
Sadly, the null is important for the state and I currently know no good solution closed.

PR #4328, opened 2021-12-06
refactor: use String#repeat

MartinWitt:
For this refactoring i`m unsure if it's better readable. Some opinions?

### DIFF
--- a/src/main/java/spoon/reflect/visitor/PrinterHelper.java
+++ b/src/main/java/spoon/reflect/visitor/PrinterHelper.java
@@ -144,7 +144,7 @@ public class PrinterHelper {
 				if (env != null) {
 					indentationSize = env.getTabulationSize();
 				}
-				sbf.append(Character.toString(' ').repeat(indentationSize));
+				sbf.append(" ".repeat(indentationSize));
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


src/main/java/spoon/reflect/visitor/PrinterHelper.java
```diff
@@ -144,9 +144,7 @@ private void writeTabsInternal() {
 				if (env != null) {
 					indentationSize = env.getTabulationSize();
 				}
-				for (int j = 0; j < indentationSize; j++) {
-					sbf.append(' ');
-				}
+				sbf.append(Character.toString(' ').repeat(indentationSize));
```
algomaster99, line 147: This refactoring is definitely better than `for-loop`. I would suggest we can further shorten this to:
```java
sb.append("\s".repeat(indentationSize));
```

I am unsure if `\s` escape sequence works like `\t` does.

algomaster99 (changes requested):

src/main/java/spoon/reflect/visitor/PrinterHelper.java
monperrus, line 147:
```suggestion
				sbf.append(" ".repeat(indentationSize));
```

MartinWitt: I merge @monperrus suggestions, but yours looks fine too.

algomaster99 (approved):

PR #4331, opened 2021-12-06
fix: prevent removal of static field import

algomaster99:
The failing test case reproduces https://github.com/SpoonLabs/sorald/issues/602.

### DIFF
--- a/src/main/java/spoon/reflect/visitor/ImportCleaner.java
+++ b/src/main/java/spoon/reflect/visitor/ImportCleaner.java
@@ -13,6 +13,7 @@ import spoon.javadoc.internal.JavadocDescriptionElement;
 import spoon.javadoc.internal.JavadocInlineTag;
 import spoon.reflect.code.CtExpression;
 import spoon.reflect.code.CtFieldAccess;
+import spoon.reflect.code.CtFieldRead;
 import spoon.reflect.code.CtInvocation;
 import spoon.reflect.code.CtJavaDoc;
 import spoon.reflect.code.CtJavaDocTag;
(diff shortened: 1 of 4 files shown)


algomaster99:
The `handleTargetedExpression` ignored instances of `CtFieldRead`. This caused the removal of import statements which imported static fields. I have added a method `isReferenceToFieldPresentInImports` which checks if the field reference is present in imports, and if it does, it is added to `computedImports`.

algomaster99:
@slarse @MartinWitt @monperrus could you please look at this?

algomaster99:
At another glance, I think the fix should belong [here](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/reflect/visitor/ImportCleaner.java#L333). I am unsure about it though.

monperrus:
Cool! Do we need the three changes at line 90, 179 and 196 to fix the single failing test case?

algomaster99:
Yes. Let me explain it case by case.

> [Line 90](https://github.com/INRIA/spoon/pull/4331/files#diff-ad1a70b692b135a81092c7b5d601d9a05f3864824c325d5f021ef43d316ba501R90)

This allows the instance of `FieldRead` to be passed to `addImport`. 

> [Line 179](https://github.com/INRIA/spoon/pull/4331/files#diff-ad1a70b692b135a81092c7b5d601d9a05f3864824c325d5f021ef43d316ba501R179)

Since we are not in `addImport` (courtesy of line 90), we need to do something about it. This conditions checks if the `ref` is present in any of the import statements. If it is not present, we `return` so that the import is not added to the final output.

> [Line 196](https://github.com/INRIA/spoon/pull/4331/files#diff-ad1a70b692b135a81092c7b5d601d9a05f3864824c325d5f021ef43d316ba501R196)

This bit is quite tricky, but I have two reasons to support this diff.
1. When the import, `import static fieldImport.UtilityLibrary.FirstUtility.LIMIT;`, is present, we move forwards to check other conditions if it needs to be added in `computedImports`. The previous condition, without my diff, passed because package of the compilation unit and the field reference were same. We need to pass this condition when the references to types are considered.
2. The comment below the line changed further supports the diff because we only need to check for `TypeReference`s in same packages and not field references.

monperrus:
Thanks for the explanation, merging.

PR #4333, opened 2021-12-06
refactor: remove unneeded continue

MartinWitt:
Fixes another qodana issue. I'm unsure if I should create one PR for each code smell type, one for each file, or one for each smell categorie in one file?


slarse (approved):
LGTM

> I'm unsure if I should create one PR for each code smell type, one for each file, or one for each smell categorie in one file?

Smaller PRs are easier to review, but there's of course some overhead if they get too small. One PR per file is fine if the amount of changes aren't that large, but it's not desirable to have say 20 unrelated changes in one PR.

One PR per code smell is fine if you don't touch too many files.

I think the general idea is that "one PR per X" is fine as long as there aren't too many unrelated changes.

One PR per individual fix is of course always OK.

monperrus:
thanks @MartinWitt

PR #4334, opened 2021-12-06
refactor: fix style issues in CtQueryImpl

MartinWitt:
This feels like only surface fixes, the class needs a rewrite badly.


src/main/java/spoon/reflect/visitor/chain/CtQueryImpl.java
```diff
@@ -30,6 +30,11 @@
  */
 public class CtQueryImpl implements CtQuery {
 
+	/**
+	 *
+	 */
```
algomaster99, line 35: Is this done to pass Qodana or checkstyle checks? I am unsure why it is required. Type members with `private` modifiers are not part of the public API and can be changed at will. Thus, documenting them should not be necessary.

MartinWitt: no, it was a mistake by my IDE, that expected a comment there. But I really don't know any code comment for this.

monperrus:
thanks @MartinWitt

PR #4336, opened 2021-12-07
fix(MavenLauncher): maximum recovery from missing libraries

henryhchchc:
The way Spoon build the class path for a Maven project in `MavenLauncher` is to launch `mvn dependencies:build-classpath` to generate a text file containing class path. However, for multi-module projects, when there is a module whose dependencies can not be resolved, Maven will skip the remaining modules and report an error.

In this PR, I added the `-fn` flag to the Maven invocation request such that Maven can continue to resolve dependencies for the remaining modules even though there is an error occurred. This can help resolve as many libraries as possible.


monperrus:
Thanks for the contribution  @henryhchchc

LGTM.

@nharrand may be the expert integrator for this.

monperrus:
thanks @henryhchchc for your contribution

PR #4339, opened 2021-12-08
review: refactor: refactor printingContext

MartinWitt:
Because the diff is kinda hard to read, here is what I did:

I refactored the old long vector for states into an inner `Enum` and `Enumset` as replacement. This is way better readable and should deliver the same performance. 

I moved the type members in the natural Java ordering of members. This means fields, methods and  inner classes. Fields spread all over the class are really hard to read. 

I tried to add at least some doc to the class. But it's still for the full class, because I don't understand all of this magic.


monperrus:
LGTM.

slarse:
Thanks @SirYwell

PR #4340, opened 2021-12-08
refactor: produce static inner classes and remove useless catch for ReplacementVisitor

MartinWitt:
Change generation to produce static inner classes and remove useless catch. Add quality of life to ReplacementVisitorGeneration adding a newline after generation to the file, because checkstyle enforces it.


monperrus:
Thanks @MartinWitt

PR #4341, opened 2021-12-09
fix: print square brackets in array initalisation

### LINKED ISSUE #4315: [Bug] Sniper Pretty Printer issue : Square brackets are missing in array initialization
**Describe the bug**
When sniper pretty printer writes a file, if a array initialization contains a square bracket after variable name , it removes it from the code. 

**To Reproduce**

Input:
```/*Code being processed by spoon and leading to the bug.*/
String str[] = new String[10];
```
Processing with Spoon:
```/*Code calling spoon on the input and triggering the bug.*/
launcher.getEnvironment().setPrettyPrinterCreator(() -> new SniperJavaPrettyPrinter(launcher.getEnvironment()));
```
Output:
```/*Output code or stacktrace if unexpected Exception is raised*/
String str = new String[10];
```
You can use any code to reproduce this issue as long as sniper pretty printer rewrites the file containing the above mentioned array initialization.

* OS: Windows
* JDK: JDK11
* Spoon version:9.1.0-beta-10

algomaster99:
Fixes #4315

### DIFF
--- a/src/main/java/spoon/pattern/internal/node/ElementNode.java
+++ b/src/main/java/spoon/pattern/internal/node/ElementNode.java
@@ -20,7 +20,6 @@ import spoon.reflect.declaration.CtElement;
 import spoon.reflect.meta.ContainerKind;
 import spoon.reflect.path.CtRole;
 import spoon.reflect.reference.CtExecutableReference;
-import spoon.support.reflect.reference.CtArrayTypeReferenceImpl;
 import spoon.support.util.ImmutableMap;
 
 import java.util.ArrayList;
@@ -265,9 +264,7 @@ public class ElementNode extends AbstractPrimitiveMatcher {
 		@SuppressWarnings("rawtypes")
 		CtElement clone = generator.getFactory().Core().create((Class) elementType.getMetamodelInterface().getActualClass());
 		generateSingleNodeAttributes(generator, clone, parameters);
-		if (this.templateElement instanceof CtArrayTypeReferenceImpl) {
-			((CtArrayTypeReferenceImpl<?>) clone).setDeclarationKind(((CtArrayTypeReferenceImpl<?>) this.templateElement).getDeclarationKind());
-		}
+		clone.setAllMetadata(templateElement.getAllMetadata());
 		generator.applyGeneratedBy(clone, generator.getGeneratedByComment(templateElement));
 		result.addResult((U) clone);
 	}
--- a/src/main/java/spoon/pattern/internal/node/ElementNode.java
+++ b/src/main/java/spoon/pattern/internal/node/ElementNode.java
@@ -20,6 +20,7 @@ import spoon.reflect.declaration.CtElement;
 import spoon.reflect.meta.ContainerKind;
 import spoon.reflect.path.CtRole;
 import spoon.reflect.reference.CtExecutableReference;
+import spoon.support.reflect.reference.CtArrayTypeReferenceImpl;
 import spoon.support.util.ImmutableMap;
 
 import java.util.ArrayList;
@@ -264,6 +265,9 @@ public class ElementNode extends AbstractPrimitiveMatcher {
 		@SuppressWarnings("rawtypes")
 		CtElement clone = generator.getFactory().Core().create((Class) elementType.getMetamodelInterface().getActualClass());
 		generateSingleNodeAttributes(generator, clone, parameters);
+		if (this.templateElement instanceof CtArrayTypeReferenceImpl) {
+			((CtArrayTypeReferenceImpl<?>) clone).setDeclarationKind(((CtArrayTypeReferenceImpl<?>) this.templateElement).getDeclarationKind());
+		}
 		generator.applyGeneratedBy(clone, generator.getGeneratedByComment(templateElement));
 		result.addResult((U) clone);
 	}
--- a/src/test/java/spoon/test/prettyprinter/TestSniperPrinter.java
+++ b/src/test/java/spoon/test/prettyprinter/TestSniperPrinter.java
@@ -877,6 +877,7 @@ public class TestSniperPrinter {
 	}
 
 	@Nested
+	@GitHubIssue(issueNumber = 4315)
 	class SquareBracketPrintingInArrayInitialisation {
 		private Consumer<CtType<?>> markFieldForSniperPrinting() {
 			return type -> {
@@ -920,6 +921,19 @@ public class TestSniperPrinter {
 					"sniperPrinter.arrayInitialisation.MultiDimension",
 					markFieldForSniperPrinting(),
 					assertPrintsBracketForArrayInitialisation("String array[][][] = new String[1][2][3];"));
+		}
+
+		@Test
+		void test_bracketsShouldBePrintedForArrayInitialisedInLocalVariable() {
+			Consumer<CtType<?>> noOpModifyLocalVariable = type -> {
+				CtMethod<?> method = type.getMethod("doNothing");
+				TestSniperPrinter.markElementForSniperPrinting(method.getBody().getStatement(0));
+			};
+
+			testSniper(
+					"sniperPrinter.arrayInitialisation.AsLocalVariable",
+					noOpModifyLocalVariable,
+					assertPrintsBracketForArrayInitialisation("int array[] = new int[]{ };"));
 		}
 	}
 
--- a/src/test/java/spoon/test/prettyprinter/TestSniperPrinter.java
+++ b/src/test/java/spoon/test/prettyprinter/TestSniperPrinter.java
@@ -876,6 +876,45 @@ public class TestSniperPrinter {
 		}
 	}
 
+	@Nested
+	class SquareBracketPrintingInArrayInitialisation {
+		private Consumer<CtType<?>> markFieldForSniperPrinting() {
+			return type -> {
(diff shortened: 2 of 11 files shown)


algomaster99:
I tried [this approach](https://github.com/INRIA/spoon/pull/4341/commits/bb98a8df074582d8e9591eabf65d567daef77e84) today, but it fails to work for elements that have no source position. Maybe a better way to go about it is to just add metadata to `CtArrayTypeReference` elements using `putMetadata` while building the tree.

algomaster99:
@slarse I wanted to ask a question. I managed to attach metadata in `CtArrayTypeReference` which would help us to distinguish the `DECLARATION_STYLE`. However, there were some test cases where the metadata is removed when a `type` inside a model is fetched. For example, consider https://github.com/INRIA/spoon/blob/7c01e6296417fe9ae3754ca17ca1fc83e1cf849f/src/test/java/spoon/test/prettyprinter/TestSniperPrinter.java#L108.

The model is built with metadata attached however, the `type` fetched using `type = f.Class().get(testClass);` does not have the above metadata present. Is this the intended behaviour?

algomaster99:
I needed some help with this pull request. Only one test is failing and I cannot figure out why. I shall first explain the changes to get better insights.

I have added a _psuedo metamodel_ element in `CtArrayTypeReferenceImpl` to store the declaration style of array. I set the type of declaration in `JDTTreeBuilder` as metadata of `CtArrayTypeReferenceImpl` and then I use this metadata for printing the brackets after type or identifier. Note that I have ignored cases like `int[] array[]` because it is a quite bizarre way to declare an array and @slarse and I feel it's safe to ignore it.

However, this approach works for all the existing test cases and the new ones introduced except `CtGenerationTest.testGenerateRoleHandler`. I tried to debug the reason and I found out that it is because the metadata of [`RoleHandler[]`](https://github.com/INRIA/spoon/blob/2d8c550940ed5ed53a85bb4070a01dfb1c8b2761/src/main/java/spoon/reflect/meta/impl/ModelRoleHandlers.java#L118) **is removed** while [prettyprinting](https://github.com/INRIA/spoon/blob/2d8c550940ed5ed53a85bb4070a01dfb1c8b2761/src/main/java/spoon/Launcher.java#L737) it. I cannot follow why it would have been erased because I am pretty sure (I checked too) that the metadata existed once the model was built, or until this line [here](https://github.com/INRIA/spoon/blob/2d8c550940ed5ed53a85bb4070a01dfb1c8b2761/src/main/java/spoon/Launcher.java#L737).

@slarse @MartinWitt @I-Al-Istannen any thoughts?

algomaster99:
The issue seems to be with the following line.
https://github.com/INRIA/spoon/blob/a47ada1b18b7cf2824ec6821502a9d21b8db9595/src/main/java/spoon/support/compiler/jdt/JDTBasedSpoonCompiler.java#L500
I think this invocation removes the metadata out of the type of field.

src/main/java/spoon/pattern/internal/node/ElementNode.java
```diff
@@ -264,6 +265,9 @@ public void forEachParameterInfo(BiConsumer<ParameterInfo, RootNode> consumer) {
 		@SuppressWarnings("rawtypes")
 		CtElement clone = generator.getFactory().Core().create((Class) elementType.getMetamodelInterface().getActualClass());
 		generateSingleNodeAttributes(generator, clone, parameters);
+		if (this.templateElement instanceof CtArrayTypeReferenceImpl) {
+			((CtArrayTypeReferenceImpl<?>) clone).setDeclarationKind(((CtArrayTypeReferenceImpl<?>) this.templateElement).getDeclarationKind());
+		}
```
algomaster99, line 270: The problem was here. The metadata of the declaration style was lost here.

Should we consider copying the metadata as well? We can make it generic and do it for all `ElementNode`s.

algomaster99:
```suggestion
		clone.setAllMetadata(this.templateElement.getAllMetadata());
```
Something like this can be done. What do you think?

algomaster99:
This is ready for review. @monperrus @slarse @MartinWitt @I-Al-Istannen , can you have a look?

Summary of the changes.
1. Add an [enum](https://github.com/INRIA/spoon/pull/4341/files#diff-553b1c5aa30c52fb924be41a5287946f25681d6e70896bb3f1de71386d87aad3R130) in `CtArrayTypeReferenceImpl`. This enum is always used in the metadata.
2. Add two public APIs there which will set and get metadata.
3. Modify `JDTTreeBuilder` to set the declaration style.
4. Modify `DefaultJavaPrettyPrinter` to conditionally print square brackets.
5. Add tests.

A bit unrelated change:
[Ensure metadata of all template elements is copied while copying them.](https://github.com/INRIA/spoon/pull/4341/commits/96180889cec4e2b0b3fec88225f9644d2189d1be)

src/main/java/spoon/pattern/internal/node/ElementNode.java
slarse: I'm not sure that's a good idea, propagating metadata in a clone is a major change to semantics.

@monperrus is there a conscious decision for not "cloning" metadata?

monperrus: > is there a conscious decision for not "cloning" metadata?

no, seems to me that it is reasonable to clone the metadata as well?

monperrus:
LGTM, will merge, thanks @algomaster99

src/test/java/spoon/test/prettyprinter/TestSniperPrinter.java
```diff
@@ -876,6 +876,69 @@ void test_retainSemiColonAfterTheLastResource() {
 		}
 	}
 
+	@Nested
+	@GitHubIssue(issueNumber = 4315)
+	class SquareBracketPrintingInArrayInitialisation {
+		// contract: square brackets should be printed *only* after the identifier of the field or local variable
```
algomaster99, line 882: @monperrus note that I have added the contract for the nested case instead of individual test cases because all of them are testing the same thing but resources are different.

src/main/java/spoon/pattern/internal/node/ElementNode.java
slarse: > no, seems to me that it is reasonable to clone the metadata as well?

I'm unsure how it could affect e.g. gumtree-spoon-ast-diff, which uses metadata to map Gumtree nodes to Spoon nodes. And surely there is a reliance somewhere on not propagating metadata, so I'd label this as a breaking change if we use it.

algomaster99: > I'm unsure how it could affect e.g. gumtree-spoon-ast-diff, which uses metadata to map Gumtree nodes to Spoon nodes.

I can check if gumtree-spoon is affected in any way.

> And surely there is a reliance somewhere on not propagating metadata

That could be an issue but our test cases are not able to catch that. I was also wondering that there are so many places where we clone elements. We should also discuss if they need to be changed as well. Otherwise, the definition of `clone` shall vary from place to place.

slarse: > > I'm unsure how it could affect e.g. gumtree-spoon-ast-diff, which uses metadata to map Gumtree nodes to Spoon nodes.
> 
> I can check if gumtree-spoon is affected in any way.
> 
> > And surely there is a reliance somewhere on not propagating metadata
> 
> That could be an issue but our test cases are not able to catch that. I was also wondering that there are so many places where we clone elements.

I don't mean in Spoon, but in some other app or library that uses Spoon. It's a behavioral change, and thus potentially breaking to dependents as something may be relying on the old behavior.

> We should also discuss if they need to be changed as well. Otherwise, the definition of `clone` shall vary from place to place.

It might be good to add an explicit test to check that metadata is propagated in a clone.

> Nothing breaks on gumtree-spoon.

That's good.

> I added `topLevelClone.setAllMetadata(topLevelElement.getAllMetadata())` [here](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/visitor/equals/CloneHelper.java#L120) and all the tests still pass. Retaining metadata does not seem to affect anything.

In libraries we control. Which is good but not sufficient to not mark this a breaking change.

algomaster99: > It might be good to add an explicit test to check that metadata is propagated in a clone.

```java
@Test
public void test_cloneCopiesMetadata() {
  String c1 = "public class A { }";
  CtType < ? > type = Launcher.parseClass(c1);
  type.putMetadata("meaning of life", 42);

  CtType < ? > typeClone = type.clone();

  assertEquals(1, typeClone.getAllMetadata().size());
}
```
This test case passed without the changes, which means that we have been cloning metadata as well, and this is [the line](https://github.com/INRIA/spoon/blob/673a5e93e8bf9af8894d5a99c76f1d61e1c23e49/src/main/java/spoon/support/visitor/clone/CloneBuilder.java#L60) that does it.

Cloning via `clone` API is different from cloning via `ElementNode`. Metadata was propagated for the former, however, it was ignored for the latter. I am unsure of the exact distinction between their semantics.

algomaster99: @slarse I added the change for cloning element via template [here](https://github.com/INRIA/spoon/pull/4396).

PR #4344, opened 2021-12-10
WIP: feat(CtArrayTypeReference): store printing information of array declaration in AST

algomaster99:
This PR adds another meta-model information to `CtArrayTypeReference` - `ARRAY_DECLARATION_KIND`. Java has two ways to place `[]` for array declaration.
1. `int[] array`
2. `int array[]`

Although the convention discourages the latter form of declaration, it is still allowed.

Currently, Spoon is unable to print `[]` after the variable name. See #4315 . Storing this information in the meta-model is a nice solution to it, but I also agree with @slarse [here](https://github.com/INRIA/spoon/issues/4315#issuecomment-990086568). Printing information has always been kept out of AST and that's why we have implemented a sniper printer in the first place.

Another solution is to add a function called `printBracketsIfRequired` [here](https://github.com/INRIA/spoon/blob/fd8060c7e53ed986e2fbdf09b36bb11a4d6f57cf/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java#L786) in `DJPP::visitCtField`. See ["how it prints identifier"](https://github.com/INRIA/spoon/issues/4315#issuecomment-989814914).


algomaster99:
I tried to modify the printers to incorporate for printing `[]` based on array declaration and it works for simpler cases like the ones in #4341  - `int x[]` and `String x[][][]`.

However, I came to know about the fact that Java also compiles the following declaration :angry:  :
```java
int[] x[] = new int[42][10]
```
This really points to the fact that we should not store this information in the AST because the possibility of bracket placement goes factorially high as the dimensions increase. Moreover, if we proceed to store such information in the AST, it might just pollute it with formatting information and it no longer will be _abstract_ syntax tree . A better way is to let the sniper printer deal with this. I am not sure how it will happen, but I can look for a fix; hopefully, by next week.

P.S. curse the people who declare multi-dimensional arrays like that.

PR #4349, opened 2021-12-12
review: fix: add junit extension for line endings

MartinWitt:
This PR adds a junit 5 extension for setting the line separator before a testcase. This solves a lot of issues with unit tests we don't run, because of windows line endings. Running test cases with this extension is better than not running them at all.
See #3752


slarse (approved):
That's pretty neat, thanks @MartinWitt

PR #4353, opened 2021-12-14
fix(ImportCleaner): resolve imports of parent types

algomaster99:
Reference: https://github.com/SpoonLabs/sorald/issues/598

Type `B` in `A.B x = new A.B()` was not being resolved to `import somePackage.A` even though this import statement was present in the compilation unit.

I apologise for the bad test case. I wasn't able to reproduce the bug in a smaller test file. The reason was that whenever it used to visit `B` in `A.B x`, the visit was ended due to this [condition](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/reflect/visitor/AstParentConsistencyChecker.java#L21). However, this line didn't interfere with the original test resource which brought about this bug.

### DIFF
--- a/src/main/java/spoon/reflect/visitor/ImportCleaner.java
+++ b/src/main/java/spoon/reflect/visitor/ImportCleaner.java
@@ -176,13 +176,19 @@ public class ImportCleaner extends ImportAnalyzer<ImportCleaner.Context> {
 				// we would like to add an import, but we don't know to where
 				return;
 			}
-			if (ref instanceof CtFieldReference<?>
-					&& !isReferenceToFieldPresentInImports((CtFieldReference<?>) ref)) {
+			if (ref instanceof CtFieldReference<?> && !isReferencePresentInImports(ref)) {
 				return;
 			}
 			CtTypeReference<?> topLevelTypeRef = typeRef.getTopLevelType();
 			if (typeRefQNames.contains(topLevelTypeRef.getQualifiedName())) {
 				//it is reference to a type of this CompilationUnit. Do not add it
+				return;
+			}
+			if (ref instanceof CtTypeReference<?>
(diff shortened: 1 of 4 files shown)


algomaster99:
@slarse @MartinWitt @I-Al-Istannen can you please look at it?

I-Al-Istannen:
I think this breaks the following:
```java
import java.util.Map.Entry;

class Hello {
	public void hey() {
		Entry<String, String> f = null;
	}
}

```

algomaster99:
Thanks for pointing it out. I am looking into it.

algomaster99:
> I think this breaks the following class (import is removed and Entry fully qualified):

You were right about it. I also added a condition which would check if the subtype is present. However, the reference to subtype here is `java.util.Map.Entry<String, String>` while the reference to import is `java.util.Map.Entry` (without type arguments). Thus, I skipped checking for `TYPE_ARGUMENT` while I compute their equality. Maybe we can put that method in `EqualVisitor` which allows you to skip for certain roles while getting the equality. What do you think?

monperrus:
thanks @algomaster99

slarse:
> Maybe we can put that method in `EqualVisitor` which allows you to skip for certain roles while getting the equality. What do you think?

@algomaster99 I don't think that's a good idea. What you're suggesting is some form of partial comparison, i.e. not equality.

algomaster99:
I thought it would be beneficial in comparing references. For example, in this case, I needed to compare two references - reference to an import and a field's type. I just wanted to skip comparison for `TYPE_ARGUMENT` role because type argument cannot be there in import references.

slarse:
I get the use case, but that's not equality, it's a different kind of comparison. So I don't think it's appropriate for the `EqualVisitor`.

Put simply, a reference with type arguments is not equal to a reference without type arguments, even if the referenced type is the same.

PR #4358, opened 2021-12-16
fix #4357: spoon doesn't handle paths containinig spaces correctly

### LINKED ISSUE #4357: [Bug] StandardEnvironment does not work if path contains spaces
**Describe the bug**
StandardEnvironment does not work if classpath contains spaces.

**To Reproduce**

check out spoon from git to a location where the path contains a space character and run `mvn compile test`.  test will fail in: 
 - JDTImportBuilderTest because of an issue in StandardEnvironment
 - JavaReflectionTreeBBuilderTest because of incorrectly converting URL to String
 
I think, the correct fix would be to refactor *all* methods taking a String argument for a path or filename should  be refactored to use Path, in a first step by introducing a Path overload and deprecating the String versions.

Input: N/A

Processing with Spoon: N/A

Output: N/A

* OS: MacOS Monterey (12.1)
* JDK: GraalVM CE (JDK 17.0.1)
* Spoon version: current github (2021-12-16)

xzel23:
Fix #4357

please review and apply


slarse:
Hi @xzel23,

Could you add a test case to demonstrate that this fix solves the problem? As per the [contributing guidelines](https://github.com/INRIA/spoon/blob/master/CONTRIBUTING.md), bugfix PRs must contain a test case that reproduces the bug.

src/main/java/spoon/support/StandardEnvironment.java
```diff
@@ -418,7 +420,7 @@ public void setInputClassLoader(ClassLoader aClassLoader) {
 				if (onlyFileURLs) {
 					List<String> classpath = new ArrayList<>();
 					for (URL url : urls) {
-						classpath.add(url.getPath());
+						classpath.add(URLDecoder.decode(url.getPath(), StandardCharsets.UTF_8));
```
I-Al-Istannen, line 423: I am not sure if this is better or something like
```java
classpath.add(Path.of(url.toURI()).toAbsolutePath().toString());
```

xzel23: I am unconvinced. Now that you remind me: Path.of(URL) will throw an exception if the URL is not pointing to something that lies on a filesystem or if no FileSystemProvider supporting the URL's protocol is present. For example to create a Path instance that points to a file located inside a Jar file (not the Jar itself), there has to be a FileSystem instance for the Jar file. In the case of the "jar:" protocol, Path.of(URL) will make a disk access and try to read the Jar contentand fail if the file is not present or cannot be read. You can try this with `jshell`:

```java
jshell> Path.of(new URL("jar:file:/hello.jar!/hello.txt").toURI())
|  Exception java.nio.file.FileSystemNotFoundException
|        at ZipFileSystemProvider.getFileSystem (ZipFileSystemProvider.java:156)
|        at ZipFileSystemProvider.getPath (ZipFileSystemProvider.java:142)
|        at Path.of (Path.java:208)
|        at (#3515 
```

xzel23:
> Could you add a test case to demonstrate that this fix solves the problem? As per the [contributing guidelines](https://github.com/INRIA/spoon/blob/master/CONTRIBUTING.md), bugfix PRs must contain a test case that reproduces the bug.

Test case added.

src/test/java/spoon/test/issue4357/TestIssue4357.java
```diff
@@ -0,0 +1,27 @@
...
+    public void testClasspathURLWithSpaces() throws MalformedURLException {
+        // contract: launcher can handle spaces in classpath URL
+        Launcher launcher = new Launcher();
+        URL[] classpath = {
+                Paths.get("src/test/resources/path with spaces").toUri().toURL()
+        };
+        ClassLoader classLoader = new URLClassLoader(classpath);
+        launcher.getEnvironment().setInputClassLoader(classLoader);
+        launcher.getEnvironment().setNoClasspath(false);
+        assertDoesNotThrow(() -> launcher.buildModel());
```
monperrus, line 25: a stronger assert would be checking for the presence of CtType Foo in the model. WDYT?

src/main/java/spoon/support/StandardEnvironment.java
I-Al-Istannen: I doubt the existing code can handle files inside JARs either? URL-Decoding would return `file:/hello.jar!/hello.txt` which likely isn't handled correctly. I'd need to have a look at it, but failing early instead of *maybe* failing later on sounds like a plus to me?

src/test/java/spoon/test/issue4357/TestIssue4357.java
xzel23: OK, that's next on my list.

src/main/java/spoon/support/StandardEnvironment.java
slarse: Hmm, this assumes utf8 encoding. How does this work on Windows when there's a non-ASCII character in the filepath?

xzel23: @slarse Maybe someone with access to a windows box could test. In any case, URLs *should* use UTF-8. Qutoe taken from [W3.org](https://www.w3.org/TR/html40/appendix/notes.html#non-ascii-chars):

> **Note.** Some older user agents trivially process URIs in HTML using the bytes of the character encoding in which the document was received. Some older HTML documents rely on this practice and break when transcoded. User agents that want to handle these older documents should, on receiving a URI containing characters outside the legal set, first use the conversion based on UTF-8. Only if the resulting URI does not resolve should they try constructing a URI based on the bytes of the character encoding in which the document was received.

I-Al-Istannen: I am not fully convinced, that JDT accepts file paths referring to files *inside a jar* or other weird things correctly, so I am not sure raw URL decoding really is a benefit here over stricter Path.of validation/conversion.

Either way, I tried it out @slarse and the URLDecode javadoc states that it only uses the charset for percent-encoded data. `new URL("öäüß")` is not stored percent encoded, so the string is returned unchanged in the default JVM windows charset. The UTF-8 setting has no relevance for it, but if the URL does contain other percent encoded data it could maybe end up with a String using *two* encodings.

slarse: @xzel23 I did not know that, good to know what the spec says (although I doubt all file systems follow that, a filepath is not necessarily a URI).

@I-Al-Istannen Thanks for checking it out, that's 100% confusing :D

src/test/java/spoon/test/issue4357/TestIssue4357.java
xzel23: Ah, this was a bit more complicated than expected because sppon doesn't load java files via classloaders. So I added a jar file that is used during the test. After some trial and error (I accidentally compiled the jar using JDK 17 and other oversights) the test finally passed in all tested configurations.

src/test/java/spoon/test/issue4357/TestIssue4357.java
```diff
@@ -0,0 +1,33 @@
...
+        };
+        launcher.getEnvironment().setNoClasspath(false);
+        launcher.getEnvironment().setShouldCompile(true);
+        ClassLoader classLoader = new URLClassLoader(classpath);
+        launcher.getEnvironment().setInputClassLoader(classLoader);
+        launcher.addInputResource(Paths.get("./src/test/resources/path with spaces/Foo.java").toAbsolutePath().toString());
+        CtModel model = launcher.buildModel();
+
+        assertTrue(model.getAllTypes().stream().anyMatch(ct -> ct.getQualifiedName().equals("Foo")), 
+                "CtTxpe 'Foo' not present in model");
```
slarse, line 31: Optional change: This only indirectly verifies that we correctly handle spaces in classpath URLs. I'd suggest actually verifying that the reference to `bar.Bar` is correctly resolved.

```suggestion
        launcher.buildModel();

        CtType<?> foo = launcher.getFactory().Type().get("Foo");
        CtMethod<?> methodWithBarReturnType = foo.getMethod("foo");
        CtTypeReference<?> barRef = methodWithBarReturnType.getType();

        assertThat(barRef.getQualifiedName(), equalTo("bar.Bar"));
        assertThat(barRef.getTypeDeclaration(), notNullValue());
        assertTrue(barRef.getTypeDeclaration().isShadow(), "expected bar.Bar to be a shadow class");
```

Another alternative would be to test this with a unit test of `StandardEnvironment` instead, which is where the problem actually lies. Since this is tested so "far away" from the problem, the assertion should be very specific to IMO.

src/test/java/spoon/test/issue4357/TestIssue4357.java
```diff
@@ -0,0 +1,33 @@
...
+import java.net.URL;
+import java.net.URLClassLoader;
+import java.nio.file.Paths;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class TestIssue4357 {
+    
+    @Test
+    public void testClasspathURLWithSpaces() throws MalformedURLException {
```
slarse, line 17: Required change: Put this test in an existing test class instead of creating a new one. Due to the phrasing of the contract, I'd put this in `LauncherTest`.

slarse (changes requested):
Production code looks fine to me. As `URL.toURI()` throws a checked exception, using `URLDecoder.decode()` seems more convenient to use.

The test should be moved to an existing test class (required change), and I strongly suggest making the assertion more specific to the scenario being tested (optional change).

> Ah, this was a bit more complicated than expected because sppon doesn't load java files via classloaders. So I added a jar file that is used during the test. After some trial and error (I accidentally compiled the jar using JDK 17 and other oversights) the test finally passed in all tested configurations.

Class loaders are only for binary code (e.g. jars and class files), not for source code. You could have added a single `.class` file as well. I don't think it matters if it's a `.jar` or a `.class`, both are easy to inspect, so you can keep it as-is.

src/test/java/spoon/test/issue4357/TestIssue4357.java
xzel23: @slarse done.

slarse (approved):
LGTM, thanks @xzel23

PR #4359, opened 2021-12-16
fix #4313: do not strip super class from qualified name

### LINKED ISSUE #4313: [Bug] Spoon removes super class name when printing nested class name resulting in invalid output
**Describe the bug**
Spoon sometimes removes the super class name when  printing a nested class name which in some cases leads to incorrect output. In this example, `public class Foo implements Iterable<Foo.Bar<?>>` is changed by Spoon to `public class Foo implements Iterable<Bar<?>>`. Note the missing `Foo.` in the generic argument. While it is sufficient to use `Bar<?>` when *inside* the class declaration of `Foo`, it `Bar`  has to be qualified as `Foo.Bar` when used as the generic parameter to its own super class. 

Note that I use `VirtualFile`in the example code which might fail before #4302 is fixed/#4303 is applied - I actually opened that one  while trying to create this minimal example. ;-) 

I will look into this later and if I find out what's wrong hopefully create a unit test and a patch.

**To Reproduce**

Input:
```java
    package nested_generic_class;
                    
    import java.util.ArrayList;
    import java.util.Iterator;
    import java.util.List;
                    
    public class Foo implements Iterable<Foo.Bar<?>> {
        private final List<Bar<?>> bars = new ArrayList<>();

        public static final class Bar<T> {
        }
                    
        @Override
        public Iterator<Foo.Bar<?>> iterator() {
            return bars.iterator();
        }
                    
    }
```

Processing with Spoon:
```java
        launcher.getEnvironment().setPreserveLineNumbers(true);

        launcher.getEnvironment().setOutputType(OutputType.NO_OUTPUT);

        launcher.buildModel();

        CtCompilationUnit cu = launcher.getFactory().Type().get(qualifiedName)
                .getPosition().getCompilationUnit();

        PrettyPrinter autoImportPrettyPrinter = launcher.getEnvironment().createPrettyPrinterAutoImport();
        String output = autoImportPrettyPrinter.prettyprint(cu);
        
        System.out.println(output);
```

Output:
```java
package nested_generic_class;

import java.util.ArrayList;
import java.util.Iterator;
import java.util.List;

public class Foo implements Iterable<Bar<?>> {
    private final List<Bar<?>> bars = new ArrayList<>();

    public static final class Bar<T> {
    }

    @Override
    public Iterator<Bar<?>> iterator() {
        return bars.iterator();
    }

}
```

* OS: MacOs 12.0.1
* JDK: 17
* Spoon version: current Git (as of Nov 27 2021)

xzel23:
I am not really happy with this as there should be a more elegant solution. But since the change only applies to `ProblemReferenceBinding` at least the name is somewhat telling.

Instead of stripping everything before the class name, this patch assumes a standard Java naming convention, i. e. package names start with a lower case letter, and that upper case names refer to classes. Then, only the package names are stripped so that "foo.bar.Joe" will be mapped to "Joe", but "foo.Bar.Joe" will be mapped to "Bar.Joe".

I also had to change one of the unit tests,  but I think it's not worse than before: in TryCatchTest, the returned type name for an inner class was expected to be "GenericType" and now is "NonClosableGenericInTryWithResources.GenericType"

Please review and provide feedback. I had this lying around for too long already (too busy in my day job).


src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
```diff
@@ -1056,7 +1056,17 @@ private void setSuperInterfaceDerivedBoundingTypeOnWildcard(
...
-		String simpleName = readableName.substring(Math.max(0, readableName.lastIndexOf('.') + 1));
+		// cut package name, assuming package names start with a lowercase letter
+		String n = readableName;
+		while (Character.isLowerCase(n.charAt(0))) {
+			int idx = n.indexOf('.');
+			if (idx < 0) {
+				break;
+			}
+			n = n.substring(idx + 1);
+		}
```
algomaster99, line 1067:
```suggestion
		String fullClassName = stripPackageName(readableName)
```
Let's not use `simpleName` now because that refers to just the name of the class and does not include its parent types. Maybe something like `fullClassName` is clearer. I also suggest abstracting this loop to a separate function just for the sake of readability.

```java
private static String stripPackageName(String fullyQualifiedName) {
	if (Character.isLowerCase(fullyQualifiedName.charAt(0))) {
		int indexToInnerPackage = fullyQualifiedName.indexOf('.');
		return stripPackageName(fullyQualifiedName.substring(indexToInnerPackage + 1));
	}
	return fullyQualifiedName;
}
```
I thought a recursive call might be clearer to read and that is why I refactored like above. Note that I removed the condition `idx < 0` because you are writing this function with an assumption "assuming package names start with a lowercase letter" so if the first character is indeed a lowercase, you will never reach this condition because you are traversing a package right now and types are yet to follow _after the delimiter_.

src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
```diff
@@ -1056,7 +1056,17 @@ private void setSuperInterfaceDerivedBoundingTypeOnWildcard(
...
+
+		String simpleName = n;
```
algomaster99, line 1069: If I replace `simpleName` with `readableName`, the tests passes which implies that we don't need to strip the package name in the first place. Thus, I also suggest adding another test case that emphasises its significance.

I think the input in #4313 is sufficient for a test case.

algomaster99 (changes requested):
Hi @xzel23 !

Thanks for your contribution. I just left some comments which might improve the readability. Feel free to drop your opinions as well. :)

src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
xzel23: The existing test case is declared in the unnamed package, that's why it passes with readableName. 

Hmmm... if you want it clear and concise, how about using a regex with named groups? It's a one-liner (the named group for package is not used, it just makes the intention clearer:

```
private static String stripPackageName(String fullyQualifiedName) {
    return replaceFirst("^(?<package>([a-z][^.]*\\.)*)(?<class>[A-Z].*)", "${class}");
}
```

algomaster99: As @slarse would agree, RegEx are hard to maintain so we avoid using it as a solution if there is a way without it. Most of the time people have to use RegEx testers to understand what it is trying to do and it can be hard sometimes. Fortunately, we do have a nice solution here.

For me, the expression is very concise, but not very clear as opposed to what you did in the loop or I did in a recursive call.

> The existing test case is declared in the unnamed package, that's why it passes with readableName.

Okay. Sounds good. But do introduce a test that would fail without this patch.

src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
slarse: I agree that this should be extracted into a separate function. Whether or not it should be recursive is a matter of taste, I think @xzel23 can use their preference.

src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
slarse: Yes, do introduce a test case reproducing the bug. This is required for bugfix PRs, see the [contributing guidelines](https://github.com/INRIA/spoon/blob/master/CONTRIBUTING.md)

slarse (changes requested):
Hi @xzel23, thanks for the PR.

I agree with @algomaster99's comments on a test case and extracting the fix into a separate method, but whether or not to make that recursive is up to you.

> I am not really happy with this as there should be a more elegant solution. But since the change only applies to ProblemReferenceBinding at least the name is somewhat telling.

A `ProblemReferenceBinding` appears when JDT cannot resolve a reference. This can only happen when Spoon runs in noclasspath mode. When that happens, we have to make a best guess as to what the reference represents, and going by naming convention is most often going to work out. So that's a good guess.

I-Al-Istannen:
Why is the reference binding a `ProblemBinding`? I can't seem to be able to reproduce your error. If this PR made your test pass, Spoon (or rather JDT) thought your code was *invalid* and does a best effort lookup from what I gathered.

If I throw your code into spoon, the method you adjusted in this PR isn't even called.

From glancing over it and running it in the debugger once, I'd say the problem is that the `Foo.` type reference *is marked implicit* and therefore not printed.

slarse:
> Why is the reference binding a `ProblemBinding`? I can't seem to be able to reproduce your error. If this PR made your test pass, Spoon (or rather JDT) thought your code was _invalid_ and does a best effort lookup from what I gathered.

I just assumed the referenced type wasn't available on the classpath. A test case reproducing the problem will clarify things either way :)

I-Al-Istannen:
@slarse The bad part is that I can reproduce it *with all of them on the classpath* :P

slarse:
> @slarse The bad part is that I can reproduce it _with all of them on the classpath_ :P

But then there's no `ProblemReferenceBinding`, right? It's as you said an incorrect implicit mark (not the first time).

I-Al-Istannen:
@slarse After a wild ride I arrived at the following: The type reference is marked as "implicit" by the ForceImportProcessor. Maybe a diff like
```patch
diff --git a/src/main/java/spoon/reflect/visitor/ForceImportProcessor.java b/src/main/java/spoon/reflect/visitor/ForceImportProcessor.java
index 89ea70ed1..3fba16d86 100644
--- a/src/main/java/spoon/reflect/visitor/ForceImportProcessor.java
+++ b/src/main/java/spoon/reflect/visitor/ForceImportProcessor.java
@@ -46,6 +46,9 @@ public class ForceImportProcessor extends ImportAnalyzer<LexicalScope> {
                                CtType<?> topLevelType = contextType.getTopLevelType();
                                CtTypeReference<?> referenceDeclaringType = reference.getDeclaringType();
                                if (referenceDeclaringType != null && referenceDeclaringType.getQualifiedName().equals(topLevelType.getQualifiedName())) {
+                                       if (reference.getQualifiedName().startsWith(referenceDeclaringType.getQualifiedName())) {
+                                               return;
+                                       }
                                        //the reference to direct child type has to be made implicit
                                        reference.setSimplyQualified(true);
                                        return;

```

Would do? This solves the issue for me, but checking the names is not really nice. I was not sure whether I am able to use `getTypeDeclaration()` there, as that could be quite costly and load stuff via reflection.

xzel23:
Hi all, sorry, I have been busy and can now continue with this.

I run into this problem while creating a gradle plugin that uses spoon. I did a lot of debugging and finally arrived at the ProblemBinding in the debugger, and the file was handled correctly after I applied the above change. So that's where the ProblemBinding comes in.

After fixing several other bugs and testing with my own branch that contains this and the other fixes (that in the meantime also made it upstream), I have now observed the same problem with other sorts of bindings in different input files. So the root cause might be something else.

I will try to create some small self-contained test cases that reprocude the issue with and without ProblemBinding and then report back.

xzel23:
@I-Al-Istannen your suggestion fixesthe error I get when running with a classpath. When I submitted this, I had been running in no-classpath mode. I will update the branch later. There is a unit test-failure however in `ImportTest.testNestedAccessPathWithTypedParameterWithImports()`. I will provide details later when I had the time to update the branch and create test cases.

I-Al-Istannen:
@xzel23 Do you want to cover both cases in this PR or should I split that off in my own (once I looked at whether the change actually makes enough sense to include)

xzel23:
@I-Al-Istannen I will update the PR to include your changes.

xzel23:
I have updated the PR with the changes suggested by @I-Al-Istannen.  Note that I also had to change ImportTest.testNestedAccessPathWithTypedParameterWithImports(), but that looks OK to me.

xzel23:
Sorry, Ipushed to early. Still have to add a test case and cleanup something here.

xzel23:
OK, I hope this PR is OK now:

- I extracted the code as suggested by @I-Al-Istannen 
- I did not use recursion because that led to an infinite recursion and before looking deeper into the cause, I thought about all the unnecessary temporary string objects created in both the recursive and "old" iterative version and so I used a slightly improved iterative version.
- I added two test cases, one of which is fixed by the change to ForceImportProcessor.java, the other by the change to ReferenceBuilder.java (and vice versa).

I have to admit that I used a fowl trick to create the second test case: I made the java file used in that test uncompilable by declaring the child class private, so that I will end up with a ProblemReferenceBinding. I am 100% sure that I havea test case with compilable Java code, but to find that one, I have to create a spoon version with all the other PRs I have submitted so far because it was a single file in a >100 files project that showed up when applying a custom developed gradle plugin that uses spoon innternally, and I forgot which file that was.

MartinWitt:
LGTM. Are you okay with adding @I-Al-Istannen as co-author because he had a wild ride for finding this fix(https://github.com/INRIA/spoon/pull/4359#issuecomment-996265045)?

xzel23:
@MartinWitt Yes, of course.

PR #4362, opened 2021-12-17
refactor(ContextBuilder): remove Qodana warnings

MartinWitt:
Field Initializer  false for field isBuildLambda in type spoon.support.compiler.jdt.ContextBuilder was redundant and was removed.
Field Initializer  false for field isBuildTypeCast in type spoon.support.compiler.jdt.ContextBuilder was redundant and was removed.
Field Initializer  false for field ignoreComputeImports in type spoon.support.compiler.jdt.ContextBuilder was redundant and was removed.
Empty String check was written as String.equals("") and refactored to String.isEmpty()

### Comment 

Because I'm kind of lazy, I'm currently writing a tool for auto generation of these commits. This is the first result and should fix for ContextBuilder some issues.

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/ContextBuilder.java
+++ b/src/main/java/spoon/support/compiler/jdt/ContextBuilder.java
@@ -66,11 +66,11 @@ public class ContextBuilder {
 
 	CompilationUnit compilationUnitSpoon;
 
-	boolean isBuildLambda = false;
+	boolean isBuildLambda;
 
-	boolean isBuildTypeCast = false;
+	boolean isBuildTypeCast;
 
-	boolean ignoreComputeImports = false;
+	boolean ignoreComputeImports;
 
 	/**
 	 * Stack of all parents elements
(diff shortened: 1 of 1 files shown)


monperrus:
Thanks @MartinWitt 

I would object removing the `= false`. It's redundant for the compiler but not for the human: it makes behavior explicit and 100% understandable, especially for novice developers.

I-Al-Istannen:
Personal opinion below :)

Why are aren't *all* fields initialized to `null` and `0` then, even just a few lines above in the same class? Do you want to change every field in the codebase to explicitly initialize to `null`, `0` and `false`, unless the constructor assigns it? Most things in the model are initialized *lazily*, so we'd need to add `= null` for the majority of field initializations.
If you do not want to do this for numbers and references, then booleans stick out and catch the attention quite a bit more. It'd make me wonder *why* they are explicitly initialized when the rest is not - maybe there was a reason.

IMHO knowing that a boolean is initialized to `false`, a number to `0` and a reference to `null` is *so* elementary, that I can't image it trips people up?

MartinWitt:
> it makes behavior explicit and 100% understandable, especially for novice developers.

As @I-Al-Istannen pointed out, then we should add it to every field. For me, the default value of primitives and objects is really basic understanding and explicitly stating it contains not useful information.

slarse:
> then we should add it to every field

I agree with this sentiment. Either rely on defaults or initialize all fields.

Furthermore, I really dislike splitting initialization between the constructor and the field declarations. As there's already a non-default constructor initializing the tree builder, if there are to be initializations of other fields I'd put them there.

monperrus:
I hear support here, we proceed!

thanks @MartinWitt

PR #4363, opened 2021-12-17
review: chore: update qodana action to v4

SirYwell:
With v4, the way how the linter is defined was changed. I adapted this manually.

Supersedes #4361.


MartinWitt:
LGTM, faster merge because we want to keep the CI pipeline in as good quality as possible, and it's only a dependency bump.

MartinWitt:
Thanks @SirYwell

PR #4366, opened 2021-12-17
fix: fix pretty-printing ordering for record parameters

### LINKED ISSUE #4365: [Bug] spoon reorders record parameters
**Describe the bug**
Record parameters are output in the wrong order, i. e. first and second are switched in this example (first line of code).

**To Reproduce**

Input:
```java
record Pair(int first, float second) {
    public static Pair of(int first, float second) {
        return new Pair(first, second);
    }
}
```

Processing with Spoon:
```java
java -classpath ${DIR}/lib/*.jar spoon.Launcher -i "${DIR}/issue" -o "${DIR}/issue/processed" --compliance 16
```

Output:
```java
record Pair(float second, int first) {
    public static Pair of(int first, float second) {
        return new Pair(first, second);
    }
}
```

* OS: MacOS 12.1
* JDK: GraalVM CE 21.3 (JDK 17.0.1)
* Spoon version: spoon-core-10.0.1-beta-2-jar-with-dependencies.jar

UPDATE: I changed the second parameter so that it can be observed that not just the names but also the types are swapped.

xzel23:
Instead of creating a new test case, I added the missing checks to the existing one. The most important one is the one with the `head(records).toString()`: changing the Set to LinkedHashSet makes all the other checks pass, but the output is still in reverse order.

The fix in CloneHelper is less than ideal, but I cannot think of a better solution. IMHO the visitor pattern is overused in spoon, i. e. I cannot see a valid reason to use it for cloning. The visitor pattern is necessary for double dispatch, but if I didn't miss anything, cloning just needs a single dispatch.

Please review and comment.

### DIFF
--- a/src/main/java/spoon/support/reflect/declaration/CtRecordImpl.java
+++ b/src/main/java/spoon/support/reflect/declaration/CtRecordImpl.java
@@ -14,6 +14,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
 import spoon.JLSViolation;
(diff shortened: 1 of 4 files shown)


xzel23:
changed the test code to make it compatible with windows

monperrus:
Thanks a lot @xzel23

PR #4369, opened 2021-12-19
review: refactor(*): Remove redundant supertypes

MartinWitt:
Removed a lot of not needed supertypes in the code, where other supertypes already have the interface as supertype.
 
Removed interface java.io.Serializable from type spoon.reflect.cu.position.NoSourcePosition, because it's redundant
Removed interface spoon.reflect.declaration.CtTypeMember from type spoon.reflect.declaration.CtConstructor, because it's redundant
Removed interface spoon.reflect.declaration.CtTypeMember from type spoon.reflect.declaration.CtMethod, because it's redundant
Removed interface java.io.Serializable from type spoon.support.reflect.cu.position.BodyHolderSourcePositionImpl, because it's redundant
Removed interface java.io.Serializable from type spoon.support.reflect.cu.position.CompoundSourcePositionImpl, because it's redundant
Removed interface java.io.Serializable from type spoon.support.reflect.cu.position.DeclarationSourcePositionImpl, because it's redundant
Removed interface java.io.Serializable from type spoon.support.reflect.cu.position.SourcePositionImpl, because it's redundant
Removed interface java.io.Serializable from type spoon.support.reflect.declaration.CtElementImpl, because it's redundant
Removed interface spoon.reflect.visitor.Filter from type spoon.reflect.visitor.filter.AbstractReferenceFilter, because it's redundant

### DIFF
--- a/src/main/java/spoon/reflect/cu/position/NoSourcePosition.java
+++ b/src/main/java/spoon/reflect/cu/position/NoSourcePosition.java
@@ -12,12 +12,11 @@ import spoon.reflect.cu.SourcePosition;
 import spoon.support.reflect.cu.CompilationUnitImpl;
 
 import java.io.File;
-import java.io.Serializable;
 
 /**
  * This class represents the position of a program element in a source file.
  */
-public class NoSourcePosition implements SourcePosition, Serializable {
+public class NoSourcePosition implements SourcePosition {
 
 	private static final long serialVersionUID = 1L;
 
--- a/src/main/java/spoon/reflect/declaration/CtConstructor.java
+++ b/src/main/java/spoon/reflect/declaration/CtConstructor.java
@@ -17,7 +17,7 @@ import spoon.support.UnsettableProperty;
 /**
  * This element defines a constructor declaration.
  */
-public interface CtConstructor<T> extends CtExecutable<T>, CtTypeMember, CtFormalTypeDeclarer, CtShadowable {
+public interface CtConstructor<T> extends CtExecutable<T>, CtFormalTypeDeclarer, CtShadowable {
 
 	/**
 	 * Always returns "&lt;init&gt;".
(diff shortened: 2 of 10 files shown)


src/main/java/spoon/reflect/meta/impl/ModelRoleHandlers.java
```diff
@@ -1,4 +1,4 @@
-/*
+/**
```
I-Al-Istannen, line 1: Why is this a javadoc comment now? The other license headers seem to be normal comments.

MartinWitt: as you see in the CI the header is wrong, I just need to apply the license plugin

I-Al-Istannen: Alright :) Looked a bit odd at first glance...

monperrus:
Interesting, thanks @MartinWitt

In programming, explicitness is good. Here, the supertypes are explicit, and there is no risk of error because the compiler entirely checks for consistency.

So I'm not in favor of this PR because explicitness is good.

MartinWitt:
I have a hard time in believing the information a programmer editing a class extending `AbstractFilter<T>` needs the information, that this class implements `Filter<T>`. But if we want really this explicitness, we should (again) declare them everywhere or nowhere. A codebase following any rule is better than no standard way.

slarse:
I have to agree with @MartinWitt here. This is both not done consistently, and there's also very little value in redeclaring superinterfaces/implemented interfaces. A modern IDE will quickly tell you all inherited/implemented interfaces, and Javadoc always lists all of them for any given class or interface. See  for example [NoSourcePosition](https://spoon.gforge.inria.fr/mvnsites/spoon-core/apidocs/spoon/reflect/cu/position/NoSourcePosition.html).

So I vote for merging this.

monperrus:
> we should (again) declare them everywhere or nowhere

I disagree on this statement. We make an informed decision based on cost/benefit analysis. If removing or adding everywhere is too costly for too little benefits, it's not good. This is best-effort open source.

> I have to agree with @MartinWitt here. 

then go ahead :)

MartinWitt:
Here is the link to the related sonar rule for documentation propose https://rules.sonarsource.com/java/RSPEC-1939

PR #4370, opened 2021-12-19
chore(qodana): pin version

MartinWitt:
Because JetBrains releases doesn't seem that polished(see 6 hour CI runs), we should stop using latest and pin versions.


monperrus:
thanks!

PR #4373, opened 2021-12-21
Update Qodana Scan action version

tiulpin:
Saw that you are using the latest version by the commit SHA :)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -160,14 +160,14 @@ jobs:
       - name: "Create folder"
         run: "mkdir -p ${{ github.workspace }}/result/master"
       - name: Qodana - Code Inspection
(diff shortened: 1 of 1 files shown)


MartinWitt:
Hi, thank you for your commit. LGTM, will merge.
#### Interesting Qodana issue
We had some trouble with timing out CIs(6h limit) and fixed it with using an older action  + Qodana-JVM container version(`2021:2` instead of `2021:3`. See for example [https://github.com/INRIA/spoon/runs/4574191472?check_suite_focus=true](https://github.com/INRIA/spoon/runs/4574191472?check_suite_focus=true).
There is currently no bug report made by us because we have zero clue, what was wrong. You may want to have a look, if you feel interested.

tiulpin:
Yeah, I've already looked at it and we started investigating. Sorry for this, we will find out what's wrong and fix this 🙂

PR #4381, opened 2021-12-23
review: chore(gradle): update gradle wrapper

MartinWitt:
From time to time we should regenerate the gradle wrapper, if we have it inside the repo.


spoon-dataflow/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.3.3-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.3.3-all.zip
```
slarse, line 3: Why change this to `all`? I don't see a need to pull down sample code and docs and stuff like that?

MartinWitt: I more or less copied the command from [https://docs.gradle.org/current/userguide/gradle_wrapper.html](https://docs.gradle.org/current/userguide/gradle_wrapper.html), but I can change it back.

slarse (approved):
LGTM

PR #4382, opened 2021-12-23
review: fix: add usage of String#lines

### LINKED ISSUE #4105: The regular expression for line separators in CommentHelper is wrong
While reading through some source code of Spoon I stumbled upon this regex

https://github.com/INRIA/spoon/blob/d18c944c48725eaccf6568ec22cd3a75d4496a3c/src/main/java/spoon/reflect/visitor/CommentHelper.java#L25-L28

This looks wrong to me, since neither `\\n\\r` (LF CR) nor `\\r` (CR) are valid line separators on any operating system I know of.

This should probably be replaced with:
```java
private static final Pattern LINE_SEPARATORS_RE = Pattern.compile("\\r\\n|\\n"); 
```
where `\r\n` (CR LF) is the line separator on Windows systems and `\n` (LF) is the line separator on Unix systems.

If the design of the current regex is intentional a comment should probably be added to clarify why these values are correct in this context.

Edit: Apparently a single `\r` is a valid line separator on a few old-school systems such as the C64, the ZX Spectrum and the Apple II, however, none of these systems run Java, see https://en.wikipedia.org/wiki/Newline#Representation for a complete list

MartinWitt:
As @I-Al-Istannen already pointed out, Java 11 has the nice `String#lines` method. Removes the regex and replaces it with this.
fixes #4105

### DIFF
--- a/src/main/java/spoon/reflect/visitor/CommentHelper.java
+++ b/src/main/java/spoon/reflect/visitor/CommentHelper.java
@@ -78,7 +78,7 @@ public class CommentHelper {
 	static void printCommentContent(PrinterHelper printer, CtComment comment, Function<String, String> transfo) {
 		CtComment.CommentType commentType = comment.getCommentType();
 		String content = comment.getContent();
-		
+
 		String[] lines = content.lines().toArray(String[]::new);
 		for (String com : lines) {
 			if (commentType == CtComment.CommentType.BLOCK) {
--- a/src/main/java/spoon/reflect/visitor/CommentHelper.java
+++ b/src/main/java/spoon/reflect/visitor/CommentHelper.java
@@ -14,7 +14,6 @@ import spoon.support.Internal;
 
 import java.util.Collection;
 import java.util.function.Function;
-import java.util.regex.Pattern;
 
 /**
  * Computes source code representation of the Comment literal
@@ -22,10 +21,6 @@ import java.util.regex.Pattern;
 @Internal
 public class CommentHelper {
 
-	/**
-	 * RegExp which matches all possible line separators
-	 */
-	private static final Pattern LINE_SEPARATORS_RE = Pattern.compile("\\n\\r|\\n|\\r");
 
 	private CommentHelper() {
 	}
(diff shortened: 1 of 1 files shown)


src/main/java/spoon/reflect/visitor/CommentHelper.java
```diff
@@ -113,7 +109,7 @@ static void printJavaDocTag(PrinterHelper printer, CtJavaDocTag docTag, Function
 			printer.write(docTag.getParam()).writeln();
 		}
 
-		String[] tagLines = LINE_SEPARATORS_RE.split(docTag.getContent());
+		String[] tagLines = docTag.getContent().lines().toArray(String[]::new);
```
slarse, line 112: How about making this a `forEach`?

```suggestion
		docTag.getContent().lines().forEach(line -> {
			if (docTag.getType().hasParam()) {
				printer.write(transfo.apply("\t\t"));
			}
			printer.write(line.trim()).writeln();
		});
```

src/main/java/spoon/reflect/visitor/CommentHelper.java
```diff
@@ -83,7 +78,8 @@ static void printComment(PrinterHelper printer, CtComment comment) {
 	static void printCommentContent(PrinterHelper printer, CtComment comment, Function<String, String> transfo) {
 		CtComment.CommentType commentType = comment.getCommentType();
 		String content = comment.getContent();
-		String[] lines = LINE_SEPARATORS_RE.split(content);
+
+		String[] lines = content.lines().toArray(String[]::new);
```
slarse, line 82: How about making this a forEach?

```suggestion
		content.lines().forEach(line -> {
			if (commentType == CtComment.CommentType.BLOCK) {
				printer.write(line);
				if (lines.length > 1) {
					printer.write(CtComment.LINE_SEPARATOR);
				}
			} else {
				printer.write(transfo.apply(line)).writeln(); // removing spaces at the end of the space
			}
		});
```

slarse (commented):
Another regex bites the dust, neat.

How about using the streams directly with for each terminators instead of going back to an array and iterating over that? I think it removes some visual noise. But it's kind of a preference thing so if you'd rather keep the array iteration that's also fine.

slarse (approved):
LGTM, thanks @MartinWitt

PR #4387, opened 2021-12-27
review: refactor(*): Convert StringTests to junit5

MartinWitt:
#3919 

In type spoon.test.strings.StringLiteralTest:
- Replaced junit 4 test annotation with junit 5 test annotation in testSnippetFullClass

In type spoon.test.strings.StringTest:
- Replaced junit 4 test annotation with junit 5 test annotation in testModelBuildingInitializer


SirYwell:
I think the assertions should be updated to JUnit 5 too

slarse:
> I think the assertions should be updated to JUnit 5 too

Indeed, see #3919

slarse (changes requested):
Also replace assertions with JUnit5 equivalents

slarse (approved):

PR #4412, opened 2022-01-06
test: migrate CtVisitorTest to junit 5

MartinWitt:
#3919
Replaced junit 4 test annotation with junit 5 test annotation in testMethodsInVisitor


slarse (approved):
Thanks @MartinWitt, fast-merging this as it's so tiny.

PR #4436, opened 2022-01-10
fix: Fix square bracket printing in array field declarations

### LINKED ISSUE #4421: [Bug] Brackets omitted for arrays of generic types
Dear all,

The Inria CI for project Juliac bumped into an issue 2 days ago. By figuring out what is going out, the issue can be trimmed down to the following simple case.

**To Reproduce**

Input:
```java
class C { Class<?>[] foo; }
```

Processing with Spoon:
```java
System.out.println( Launcher.parseClass("class C { Class<?>[] foo; }") );
```

Output:
```java
class C {
    Class<?> foo;
}
```

Note that the issue does not arise with array of raw types, e.g. Class[] is correctly pretty-printed.

* OS: macOS
* JDK: 13
* Spoon version: latest from github

algomaster99:
Fixes #4421

I missed attaching metadata to generic types in array type references in https://github.com/INRIA/spoon/pull/4341 which resulted in #4421. The changes in this PR fix it.

Also, I slightly changed the test case to modify the source position of type reference instead of the entire field because the former is more precise.


monperrus:
LGTM, will merge.

PR #4448, opened 2022-01-10
fix: no parentheses for single lambda parameters

MartinWitt:
Before the fix DJPP printed  `((Predicate<Person>) (( p) -> p.age > 10))` and now `((Predicate<Person>) (p -> p.age > 10))`
The removed whitespace and parentheses for single parameter lambdas without type is much closer to normal style.


monperrus:
nice, LGTM. Thanks @MartinWitt

PR #4486, opened 2022-01-13
test: migrate GetBinaryFilesTest to JUnit 5

MartinWitt:
#3919
# Change Log
## JUnit4-@Test
The JUnit 4 `@Test` annotation should be replaced with JUnit 5 `@Test` annotation.
## JUnit4Assertion
The JUnit4 assertion should be replaced with JUnit5 Assertions.

### JUnit4-@Test
- Replaced junit 4 test annotation with junit 5 test annotation in `testSingleBinary`
- Replaced junit 4 test annotation with junit 5 test annotation in `testExistingButNotBuiltBinary`
- Replaced junit 4 test annotation with junit 5 test annotation in `testMultiClassInSingleFile`
- Replaced junit 4 test annotation with junit 5 test annotation in `testNestedTypes`
- Replaced junit 4 test annotation with junit 5 test annotation in `testAnonymousClasses`
### JUnit4Assertion
- Transformed junit4 assert to junit 5 assertion in `testSingleBinary`
- Transformed junit4 assert to junit 5 assertion in `testExistingButNotBuiltBinary`
- Transformed junit4 assert to junit 5 assertion in `testMultiClassInSingleFile`
- Transformed junit4 assert to junit 5 assertion in `testNestedTypes`
- Transformed junit4 assert to junit 5 assertion in `testAnonymousClasses`

### DIFF
--- a/src/test/java/spoon/test/compilationunit/GetBinaryFilesTest.java
+++ b/src/test/java/spoon/test/compilationunit/GetBinaryFilesTest.java
@@ -23,7 +23,6 @@ import java.nio.file.Path;
 import java.util.List;
 import java.util.Map;
 
-import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.io.TempDir;
 import spoon.Launcher;
@@ -41,9 +40,6 @@ public class GetBinaryFilesTest {
 
 	@TempDir
 	Path tmpFolder;
-
-	@BeforeEach
-
 
 	@Test
 	public void testSingleBinary() {
--- a/src/test/java/spoon/test/compilationunit/GetBinaryFilesTest.java
+++ b/src/test/java/spoon/test/compilationunit/GetBinaryFilesTest.java
@@ -16,21 +16,22 @@
  */
 package spoon.test.compilationunit;
 
-import org.junit.Rule;
-import org.junit.Test;
-import org.junit.rules.TemporaryFolder;
-import spoon.Launcher;
-import spoon.SpoonModelBuilder;
-import spoon.reflect.cu.CompilationUnit;
 
 import java.io.File;
 import java.io.IOException;
 import java.util.List;
 import java.util.Map;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertTrue;
+import org.junit.Rule;
+import org.junit.jupiter.api.Test;
+import org.junit.rules.TemporaryFolder;
+import spoon.Launcher;
+import spoon.SpoonModelBuilder;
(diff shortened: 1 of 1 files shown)


src/test/java/spoon/test/compilationunit/GetBinaryFilesTest.java
```diff
@@ -16,36 +16,41 @@
...
  * Tests {@link CompilationUnit#getBinaryFiles()}.
  */
 public class GetBinaryFilesTest {
 
-	@Rule
-	public TemporaryFolder tmpFolder = new TemporaryFolder();
+	@TempDir
+	Path tmpFolder;
+
+	@BeforeEach
```
slarse, line 45: This looks strange?

slarse (changes requested):
Weird annotation.

src/test/java/spoon/test/compilationunit/GetBinaryFilesTest.java
MartinWitt: Sadly, JUnit 5 has no longer the `@Rule` annotations but extensions, see https://junit.org/junit5/docs/current/user-guide/#migrating-from-junit4-rule-support. The replacement for the old temporary folder concept is https://junit.org/junit5/docs/5.4.2/api/org/junit/jupiter/api/io/TempDir.html, https://www.baeldung.com/junit-5-temporary-directory.

slarse: Maybe I'm dense here, but where does `@BeforeEach` come into the picture?

I actuall quite like the new `@TempDir` construct, allows you to remove the field (ugh, fields!) and do something like this:

```java
@Test
void testSomething(@TempDir workdir) {
    // use injected temporary directory
}
```

MartinWitt: Oh the `@BeforeEach` should be removed you are right

MartinWitt: > I actuall quite like the new @TempDir construct, allows you to remove the field (ugh, fields!) and do something like this:

Yes, if you use this style it's nice, but then the whole codebase should do it this way and not mix of different styles

slarse: > Yes, if you use this style it's nice, but then the whole codebase should do it this way and not mix of different styles

Good practices have to start somewhere :). That's kind of the thing with legacy, good practices have to be adopted gradually. And often, they are never _fully_ adopted, but we can be watchful for future PRs.

If someone were to submit a PR with a JUnit5 test that had a field like this, I'd request a change. In this case, it's already there, so it's ok.

slarse (approved):

PR #4491, opened 2022-01-13
test: migrate TestCtBlock to JUnit 5

MartinWitt:
#3919
# Change Log
## JUnit4-@Test
The JUnit 4 `@Test` annotation should be replaced with JUnit 5 `@Test` annotation.
## JUnit4Assertion
The JUnit4 assertion should be replaced with JUnit5 Assertions.

### JUnit4-@Test
- Replaced junit 4 test annotation with junit 5 test annotation in `testRemoveStatement`
- Replaced junit 4 test annotation with junit 5 test annotation in `testAddStatementInBlock`
- Replaced junit 4 test annotation with junit 5 test annotation in `testAddStatementInCase`
### JUnit4Assertion
- Transformed junit4 assert to junit 5 assertion in `testRemoveStatement`
- Transformed junit4 assert to junit 5 assertion in `testAddStatementInBlock`
- Transformed junit4 assert to junit 5 assertion in `testAddStatementInCase`

### DIFF
--- a/src/test/java/spoon/test/ctBlock/TestCtBlock.java
+++ b/src/test/java/spoon/test/ctBlock/TestCtBlock.java
@@ -16,7 +16,10 @@
  */
 package spoon.test.ctBlock;
 
-import org.junit.Test;
+
+import java.util.List;
+
+import org.junit.jupiter.api.Test;
 import spoon.Launcher;
 import spoon.reflect.code.CtBlock;
 import spoon.reflect.code.CtCase;
@@ -28,11 +31,9 @@ import spoon.reflect.declaration.CtMethod;
 import spoon.reflect.visitor.filter.NamedElementFilter;
 import spoon.test.ctBlock.testclasses.Toto;
 
-import java.util.List;
-
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotSame;
-import static org.junit.Assert.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotSame;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 /**
  * Created by urli on 15/03/2017.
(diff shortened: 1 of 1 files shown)


src/test/java/spoon/test/ctBlock/TestCtBlock.java
```diff
@@ -83,10 +84,10 @@ public void testAddStatementInBlock() {
 
 		assertEquals(5, block.getStatements().size());
 
-		assertEquals(originalBlock.getStatement(0), block.getStatement(0));
+		assertEquals((CtStatement) originalBlock.getStatement(0), block.getStatement(0));
```
slarse, line 87: Is this cast really necessary?

src/test/java/spoon/test/ctBlock/TestCtBlock.java
```diff
@@ -83,10 +84,10 @@ public void testAddStatementInBlock() {
...
 
 		for (int i = 1; i < 4; i++) {
-			assertEquals(originalBlock.getStatement(i), block.getStatement(i + 1));
+			assertEquals((CtStatement) originalBlock.getStatement(i), block.getStatement(i + 1));
```
slarse, line 90: Same as above.

src/test/java/spoon/test/ctBlock/TestCtBlock.java
MartinWitt: image Looks like otherwise the compiler has some type problems.

slarse: There's gotta be something wrong here, how in the world would the overload resolve to `Short`? I'll have a look and see if my compiler can give us a second opinion, but it'll have to wait until tomorrow :)

slarse: Argh, okay now I get it. `CtStatementList.getStatement()` is typed with a type parameter that's only used in the return type, see: #1846

It can possibly be solved, but it's just not worth the effort.

slarse (approved):

PR #4509, opened 2022-01-14
fix: fix printing of joint declaration of fields and local variable

### LINKED ISSUE #3386: bug(sniper): fields and local variables with joint declaration cannot be snipered

algomaster99:
Fixes #3386 

I think the problem lies in how `PositionBuilder` computes the source position of the individual elements in the joint declaration. I am still looking for a fix.

### DIFF
--- a/src/main/java/spoon/reflect/declaration/CtVariable.java
+++ b/src/main/java/spoon/reflect/declaration/CtVariable.java
@@ -49,4 +49,16 @@ public interface CtVariable<T> extends CtNamedElement, CtTypedElement<T>, CtModi
 	 */
 	@DerivedProperty
 	boolean isPartOfJointDeclaration();
+
+	/**
+	 * Returns true if field or a local variable are part of the same joint declaration.
+	 * Eg:
+	 * ```
+	 * int a,b;
+	 * int c;
+	 * ```
+	 * a and b are part of the same joint declaration.
+	 * a and c are not part of the same joint declaration.
+	 */
+	boolean isInSameJointDeclarationAs(CtVariable<?> variable);
 }
--- a/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
+++ b/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
@@ -100,6 +100,7 @@ import spoon.reflect.declaration.CtProvidedService;
 import spoon.reflect.declaration.CtRecord;
 import spoon.reflect.declaration.CtRecordComponent;
 import spoon.reflect.declaration.CtType;
+import spoon.reflect.declaration.CtTypeMember;
 import spoon.reflect.declaration.CtTypeParameter;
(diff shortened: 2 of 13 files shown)


algomaster99:
A declaration like `int f1, f2, f3, f4 = 2` is counted as one instance of `ElementSourceFragment`, but I expected it to be a `CollectionSourceFragment` containing the 4 field declarations and the tokens in between them. This gave me an idea to modify the source position of each declaration so that all fields have disjoint intervals of source position.

Proceeding with this idea, I modified the `PositionBuilder` but it broke some implementations:
1. `CtLocalVariable.isPartOfJointDeclaration()`
2. Since `f2` is now in its own discrete interval, say `55-56`, it fails the consistency check for positions [here](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/sniper/internal/ElementSourceFragment.java#L472). The child of `f2` is supposed to be `int` (its type) which is way before 55 and so it throws an error complaining about inconsistent positions.

I shall now look at how I can modify the printer to fix this issue. The `isPartOfJointDeclaration` API works so at least we have a way to know how the variable is declared.

algomaster99:
I tried to modify the printer, but somehow that does not seem to be the best way. It prints something like this:
```java
public class Field {
	String f1,

	 f2,

	 f3 = "42",

	 f4;
}
```
It prints some extra new lines between the declaration. The more code I wrote to prevent it, the more I felt that I am working around this problem. Like I said in the previous comment, the root problem is with how the source fragments are classified. If we modify that, then it breaks the `isPartOfJointDeclarationAPI`.

algomaster99:
I will reopen this when I resume my work [here](https://github.com/algomaster99/spoon/tree/joint-declaration-esf).

PR #4512, opened 2022-01-15
style: refactor `@TempDir` to method parameter in GetBinaryFilesTest

MartinWitt:
Yesterday, I had a small conversation with @slarse, and we agreed we should refactor `@TempDir` as method parameter when possible. Removing fields from test classes and converting them into method parameters makes test way easier to read.


algomaster99 (approved):

slarse (approved):
Thanks @MartinWitt

PR #4536, opened 2022-01-18
fix: fix qodana warning with proper call to Array.toString()

MartinWitt:
# Change Log
## ArraysToString
`array.toString()` is not the best way to print an array. Use `Arrays.toString(array)` instead.
- https://rules.sonarsource.com/java/RSPEC-2116

### ArraysToString
- Replaced `str.toString()` with `Arrays.toString(str)`.

### DIFF
--- a/src/main/java/spoon/support/sniper/internal/ElementSourceFragment.java
+++ b/src/main/java/spoon/support/sniper/internal/ElementSourceFragment.java
@@ -721,7 +721,7 @@ public class ElementSourceFragment implements SourceFragment {
 			}
 			// nothing has matched, best effort token
 			if (longestMatcher == null) {
-				// this case only exists currently because we have problems with sniper printing and joint variable declarations
+				consumer.accept(new TokenSourceFragment(str.toString(), TokenType.CODE_SNIPPET));
 				return;
 			}
 			consumer.accept(new TokenSourceFragment(longestMatcher.toString(), longestMatcher.getType()));
--- a/src/main/java/spoon/support/sniper/internal/ElementSourceFragment.java
+++ b/src/main/java/spoon/support/sniper/internal/ElementSourceFragment.java
@@ -7,6 +7,7 @@
  */
 package spoon.support.sniper.internal;
 
+
 import spoon.SpoonException;
 import spoon.reflect.code.CtComment;
 import spoon.reflect.code.CtLiteral;
@@ -29,6 +30,7 @@ import spoon.reflect.visitor.EarlyTerminatingScanner;
 import spoon.support.Experimental;
 import spoon.support.reflect.CtExtendedModifier;
 import spoon.support.reflect.cu.position.SourcePositionImpl;
+
 
 import java.util.ArrayDeque;
 import java.util.ArrayList;
@@ -338,7 +340,7 @@ public class ElementSourceFragment implements SourceFragment {
 		OTHER_IS_BEFORE,
 		OTHER_IS_AFTER,
 		OTHER_IS_CHILD,
-		OTHER_IS_PARENT
+		OTHER_IS_PARENT;
 	}
 
 	/**
@@ -720,7 +722,7 @@ public class ElementSourceFragment implements SourceFragment {
 			}
 			// nothing has matched, best effort token
 			if (longestMatcher == null) {
-				consumer.accept(new TokenSourceFragment(str.toString(), TokenType.CODE_SNIPPET));
+				consumer.accept(new TokenSourceFragment(Arrays.toString(str), TokenType.CODE_SNIPPET));
 				return;
 			}
 			consumer.accept(new TokenSourceFragment(longestMatcher.toString(), longestMatcher.getType()));
--- a/src/main/java/spoon/support/sniper/internal/ElementSourceFragment.java
+++ b/src/main/java/spoon/support/sniper/internal/ElementSourceFragment.java
@@ -694,14 +694,14 @@ public class ElementSourceFragment implements SourceFragment {
 			consumer.accept(new TokenSourceFragment(buff.toString(), TokenType.SPACE));
 			return;
 		}
-		char[] str = new char[buff.length()];
-		buff.getChars(0, buff.length(), str, 0);
+		char[] charArray = new char[buff.length()];
+		buff.getChars(0, buff.length(), charArray, 0);
 		int off = 0;
-		while (off < str.length) {
+		while (off < charArray.length) {
 			//detect java identifier or keyword
-			int lenOfIdentifier = detectJavaIdentifier(str, off);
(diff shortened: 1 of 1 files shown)


MartinWitt:
Do you want a list of all we have or all qodana has?

algomaster99:
@MartinWitt Both would be good. I will see if we can fix any violations using Sorald.

MartinWitt:
For all we have, see here https://github.com/INRIA/spoon/pull/4481/files (we disabled this view in a recent PR) or  run `sudo docker run --rm -it -p 8080:8080 -v $PWD/:/data/project/ jetbrains/qodana-jvm-community:2021.3 --show-report`
For all rules, see here https://www.jetbrains.com/help/idea/list-of-java-inspections.html
Maybe you can open the project in IntelliJ and run qodana/the inspections there.

src/main/java/spoon/support/sniper/internal/ElementSourceFragment.java
```diff
@@ -7,6 +7,7 @@
  */
 package spoon.support.sniper.internal;
 
+
```
algomaster99, line 10: You accidentally added this new line.

algomaster99 (approved):
Delete the new line and then we can merge.

src/main/java/spoon/support/sniper/internal/ElementSourceFragment.java
```diff
@@ -720,7 +721,7 @@ private void onCharSequence(CharType type, StringBuilder buff, Consumer<SourceFr
 			}
 			// nothing has matched, best effort token
 			if (longestMatcher == null) {
-				consumer.accept(new TokenSourceFragment(str.toString(), TokenType.CODE_SNIPPET));
+				consumer.accept(new TokenSourceFragment(Arrays.toString(str), TokenType.CODE_SNIPPET));
```
monperrus, line 724: very confusing to have an array named str, `charArray` is better?

MartinWitt: Yes, i renamed it

src/main/java/spoon/support/sniper/internal/ElementSourceFragment.java
MartinWitt: Yes, I still have problems of added newlines while sniper printing. I removed it.

src/main/java/spoon/support/sniper/internal/ElementSourceFragment.java
```diff
@@ -714,13 +715,13 @@ private void onCharSequence(CharType type, StringBuilder buff, Consumer<SourceFr
...
 			for (StringMatcher strMatcher : matchers) {
-				if (strMatcher.isMatch(str, off)) {
+				if (strMatcher.isMatch(charArray, off)) {
 					longestMatcher = strMatcher.getLonger(longestMatcher);
 				}
 			}
 			// nothing has matched, best effort token
 			if (longestMatcher == null) {
-				consumer.accept(new TokenSourceFragment(str.toString(), TokenType.CODE_SNIPPET));
+				consumer.accept(new TokenSourceFragment(Arrays.toString(charArray), TokenType.CODE_SNIPPET));
```
monperrus, line 724: this looks like this best-effort token never worked. I guess we don't have any single test here :)?

MartinWitt: Looking at the coverage, there should exist a testcase for this line https://coveralls.io/builds/45751341/source?filename=src%2Fmain%2Fjava%2Fspoon%2Fsupport%2Fsniper%2Finternal%2FElementSourceFragment.java#L724. But I have not found a way in the UI to find it.

monperrus: > Looking at the coverage, there should exist a testcase for this line coveralls.io/builds/45751341/source?filename=src%2Fmain%2Fjava%2Fspoon%2Fsupport%2Fsniper%2Finternal%2FElementSourceFragment.java#L724. But I have not found a way in the UI to find it.

argh, this is https://github.com/INRIA/spoon/issues/4541

monperrus: anyway, I'd be in favor in writing an assertion on the content of this token. Otherwise, we fix broken code by something which may be buggy as well

would you be able to add such an assertion?

MartinWitt: After some debugging, I found the testcase https://github.com/INRIA/spoon/blob/f98e33b7c669e9c11d6a916333bcac13444b1e5d/src/test/java/spoon/test/prettyprinter/TestSniperPrinter.java#L187 using this line with a string `0,`

monperrus: great, could you write an assertion on the content of the token impacted by this PR?

MartinWitt: I believe the current input token is simply wrong. See for example https://www.baeldung.com/java-array-to-string.

> The default implementation of the toString() method on an array returns something like Ljava.lang.String;@74a10858 > which only informs us of the object's type and hash code.
```java
public TokenSourceFragment(String source, TokenType type) {
	this.source = source;
	this.type = type;
}
```
The source String should contain the original source token and not a string like `"[C@2d96543c"`. It's pretty hard to assert against something internal and defect.

monperrus: > I believe the current input token is simply wrong.

Agree, that's why we need a new test or a new assertion to assert the actual correct content. Would you write this?

MartinWitt: I can only trigger this line with joint declarations but currently the sniper is not able to work with them, see https://github.com/INRIA/spoon/issues/3386. It looks like either this token was never relevant or is needed in a future version and we can't great asserts for the value currently.

monperrus: If we are unable to specify this line, then the best option is to simply remove it. Could you give it a try?

MartinWitt: It looks like the token is for some strange reasons important for the sniper printing and removing it adds unwanted whitespaces. So we more or less have to keep it and fix this after fixing the linked problem.

monperrus: ack, then let's revert this last commit and proceed with the change, it's not harmful. we replace code we don't understand by code we still don't understand but less confusing.

monperrus:
Thanks @MartinWitt

PR #4537, opened 2022-01-18
refactor: String.equals("") instead  of String.isEmpty()

MartinWitt:
# Change Log
## EmptyStringCheck
Checking if a string is empty should be done by `String#isEmpty` instead of `String.equals("")`

### EmptyStringCheck
- Empty String check was written as `sb.toString().equals("")` and refactored to `sb.toString().isEmpty()`

### DIFF
--- a/src/main/java/spoon/support/compiler/SpoonPom.java
+++ b/src/main/java/spoon/support/compiler/SpoonPom.java
@@ -29,7 +29,6 @@ import spoon.compiler.SpoonFolder;
 import spoon.compiler.SpoonResource;
 import spoon.compiler.SpoonResourceHelper;
 
-
 import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileNotFoundException;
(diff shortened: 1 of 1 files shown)


src/main/java/spoon/support/compiler/SpoonPom.java
```diff
@@ -503,7 +503,7 @@ private void generateClassPathFile(File mvnHome, MavenLauncher.SOURCE_TYPE sourc
 					sb.append(line);
 					line = br.readLine();
 				}
-				if (!"".equals(sb.toString())) {
+				if (!sb.toString().isEmpty()) {
```
SirYwell, line 506: This can be further simplified to `!sb.isEmpty()`

SirYwell (changes requested):

src/main/java/spoon/support/compiler/SpoonPom.java
MartinWitt: Good catch time for the next refactoring

MartinWitt: https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/CharSequence.html#isEmpty() This is a JDK15 method

PR #4569, opened 2022-01-23
test: migrate ContractOnSettersParametrizedTest to JUnit 5

MartinWitt:
#3919
while refactoring, I removed an unused field and moved the fields to the top

### DIFF
--- a/src/test/java/spoon/test/parent/ContractOnSettersParametrizedTest.java
+++ b/src/test/java/spoon/test/parent/ContractOnSettersParametrizedTest.java
@@ -16,7 +16,7 @@
  */
 package spoon.test.parent;
 
-import org.junit.Test;
+import org.junit.jupiter.api.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
 import spoon.SpoonException;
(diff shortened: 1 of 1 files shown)


src/test/java/spoon/test/parent/ContractOnSettersParametrizedTest.java
```diff
@@ -49,23 +48,29 @@
...
+	private ModelChangeListener changeListener = new ModelChangeListener();
 
-	@Parameterized.Parameters(name = "{0}")
-	public static Collection<Object[]> data() {
-		return createReceiverList();
+	@TestFactory
+	public  Collection<DynamicTest> data() {
+		List<DynamicTest> values = new ArrayList<>();
+		for (CtType<?> t : allInstantiableMetamodelInterfaces) {
+			if (!(CtReference.class.isAssignableFrom(t.getActualClass()))) {
```
monperrus, line 69: why do we need this extra condition?

MartinWitt: I copied the called method from here https://github.com/INRIA/spoon/blob/c8ec87305409fd864ec68ca07a94932c91fced99/src/test/java/spoon/test/parent/ContractOnSettersParametrizedTest.java#L76 and adapted it to create DynamicTests. The method `createReceiverList` is used in other test classes and therefore wasn't easy to change.

monperrus: ok thanks

PR #4584, opened 2022-02-03
chore(qodana): use new flag "changes"

MartinWitt:
Qodana got a new flag `changes` which as you (don't) expect `Inspect uncommitted changes and report new problems.`. Uncommitted means here uncommitted and not the expected behaviour changed files in the PR. Therefore, we reset the git branch with `git reset --soft master`. This reduces the qodana runtime in less than half.
Qodana has an annotator included. We don't need @I-Al-Istannen JQ magic any more and use the included from now. See https://github.com/INRIA/spoon/pull/4584/files#annotation_2688909324.
The only risk is that for some reason `git reset --soft master` fails.

Edit: The bad smell and failing CI is currently only there for showcasing and gets removed after approval.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -146,19 +146,10 @@ jobs:
         uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # renovate: tag=v2.4.0
         with:
           path: pull_request
-      - name: Checkout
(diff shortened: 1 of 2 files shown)


monperrus:
LGTM thanks.

SirYwell (approved):

PR #4585, opened 2022-02-03
refactor: remove unnecessary return

MartinWitt:
`return` is unnecessary as the last statement in a 'void' method

As always, auto generated.

### DIFF
--- a/src/main/java/spoon/support/sniper/SniperJavaPrettyPrinter.java
+++ b/src/main/java/spoon/support/sniper/SniperJavaPrettyPrinter.java
@@ -221,7 +221,6 @@ public class SniperJavaPrettyPrinter extends DefaultJavaPrettyPrinter implements
 			public void printSourceFragment(SourceFragment fragment, ModificationStatus isModified) {
 				if (isModified == ModificationStatus.UNKNOWN || isModified == ModificationStatus.MODIFIED) {
 					printAction.run();
-					return;
 				} else {
 					if (fragment instanceof CollectionSourceFragment) {
 						//we started scanning of collection of elements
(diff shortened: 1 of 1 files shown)


monperrus:
thanks @MartinWitt

Indeed it's equivalent for the machine, but not for the human. Removing the return makes it harder for the human developer to understand what happens.

With `return`, the human is 100% sure to return just by reading the if branch.

Without `return`, the human needs to read the rest of the method.

MartinWitt:
> Removing the return makes it harder for the human developer to understand what happens.

Understanding of source code can be measured in complexity of the code. Less complex code is easier to read. I'm aware of  two metrics for measuring code complexity, either Allen[1] or cyclomatic complexity(McCabe)[2]. As the first is rather unknown and not often used,  we may look at the second. And from my understanding of the metric, each return increases the complexity. So these returns increase the complexity to read the source code. 

> Without return, the human needs to read the rest of the method.

But aren't you ignoring the case that typically humans need to understand the whole method and not only some parts of it. 
Personally, while debugging, I try to understand the whole method where I think the bug is inside. What leads you to your opinion? 



[1] https://dl.acm.org/doi/abs/10.1007/s11219-006-9010-3 
[2] https://perso.ensta-paris.fr/~diam/java/online/notes-java/principles_and_practices/complexity/complexity-java-method.html

src/main/java/spoon/support/sniper/SniperJavaPrettyPrinter.java
```diff
@@ -221,7 +221,6 @@ public void onTokenWriterWrite(TokenType tokenType, String token, CtComment comm
 			public void printSourceFragment(SourceFragment fragment, ModificationStatus isModified) {
 				if (isModified == ModificationStatus.UNKNOWN || isModified == ModificationStatus.MODIFIED) {
 					printAction.run();
-					return;
 				} else {
 					if (fragment instanceof CollectionSourceFragment) {
 						//we started scanning of collection of elements
```
slarse, line 226: IMO this thing is much worse than that early return

```suggestion
				} else if (fragment instanceof CollectionSourceFragment) {
						//we started scanning of collection of elements
```

monperrus:
> Understanding of source code can be measured in complexity of the code.

Fully agree.

Yet, 1) the metrics you mentioned, which I know very well, are backed up with little empirical evidence of usefulness 2) complexity is highly context dependent and cannot be captured by a single metric.

slarse:
I'm gonna come in with another suggestion here: I think this would be a good place to use guard clauses. I.e. along these lines:

```java
void method() {
    if (something) {
        return;
    }

    // invariant: something is false
    // do other stuff
}
```

slarse:
I'm closing this per our contributing guidelines (3 months of inactivity). And I'm not counting my own comment there.

PR #4588, opened 2022-02-07
test(UnresolvedBug): Replace Junit 4 Annotation with Junit 5 Annotation

MartinWitt:
This should be the last junit 4 PR for spoon-core. I replaced the old junit 4 test runner with a new test extension and improved the annotation to a meta annotation[1]. The `GithubIssue` annotation now combines `@Test` and `UnresolvedBugExtension` extension. The `UnresolvedBugExtension` is currently an inner class because I see no reason, it should be part of the API or used anywhere else. 

Edit:
What we have lost is the ability to use the `@GithubIssue` annotation on type level. This has implementation reasons but could be fixed in another PR if this feature is really needed. 

[1} https://junit.org/junit5/docs/current/user-guide/#writing-tests-meta-annotations


monperrus:
Great, I'm learning that we can `@Test` a new annotation.

Did you check that surefire does actually run all those tests if we remove `@Test`?

MartinWitt:
> [](https://github.com/MartinWitt)Did you check that surefire does actually run all those tests if we remove @Test?

I have added a commit to showcase it. They should run and fail because of the thrown exception. See https://github.com/INRIA/spoon/runs/5092067694?check_suite_focus=true

> Great, I'm learning that we can @Test a new annotation.

In JUnit 5 there is a meta annotation concept, and you can combine more or less all annotations to create new powerful, expressive ones. In read about the concept about extensions and these meta annotations in [1]. I believe there is a lot of potential to write more extensions and annotations, improving the test classes.

[1] https://www.oreilly.com/library/view/mastering-software-testing/9781787285736/

src/test/java/spoon/test/GitHubIssue.java
```diff
@@ -11,10 +11,75 @@
...
+	/**
+	 * The github number of the issue.
+	 */
 	int issueNumber();
+
+	/**
+	 * If true, the test will be marked as success if the test fails.
+	 * Signals that the issue is not fixed yet.
+	 */
+	boolean open() default true;
```
slarse, line 45: This is probably just me thinking too much about this, but wouldn't "fixed" be more appropriate?

I'm thinking that 1) when you first mark it such that the test should pass (in a PR), the issue itself is most likely still open. And 2), there may be issues that have multiple symptoms, and a PR could fix one such symptom without actually warranting a closing of the issue.

CONTRIBUTING.md
```diff
@@ -60,7 +60,7 @@ Guidelines for feature pull-requests:
 Other kinds of pull-requests:
 
 1. Pull requests with passing test cases only are welcome, they specify previously unspecified behavior and are prefixed by "test:".
-1. Pull requests with failing test cases only are welcome, they reproduce bugs and are very useful for maintainers to fix them. You can prevent failing the CI with adding the annoation `@Ignore("UnresolvedBug")` and `@GitHubIssue(issueNumber = <your-issue-number>)`.
+1. Pull requests with failing test cases only are welcome, they reproduce bugs and are very useful for maintainers to fix them. You can prevent failing the CI with adding the annoation `@GitHubIssue(issueNumber = <your-issue-number>)`.
```
slarse, line 63: Mention what to do when fixing an existing `GitHubIssue`-marked (and currently failing) test?

slarse:
Very cool! I'll have a closer look tomorrow, just had some initial very superficial thoughts.

src/test/java/spoon/test/GitHubIssue.java
MartinWitt: > I'm thinking that 1) when you first mark it such that the test should pass (in a PR), the issue itself is most likely still open, so it will feel a bit strange

Yes, open is not a good wording here, `fixed` works better. My intention here was to leave the trace from test case to issue in the code even after the issue was fixed. Test cases for an issue are always way more specific, and having a linked issue helps in understanding them. Git blame can provide this information too, but over the time the git history gets worse and worse.

slarse: > My intention here was to leave the trace from test case to issue in the code even after the issue was fixed.

Yeah, I get that, I wasn't in any way objecting to the idea. Just the name :)

src/test/java/spoon/test/UnresolvedBugTest.java
```diff
@@ -45,19 +32,19 @@
...
 				.collect(Collectors.toList());
 	}
 
 	/**
 	 * Checks if every githubIssue annotation has an open github issue.
 	 */
-	@Ignore
+	
 	@Test
+	@Disabled
```
slarse, line 47: I'm somewhat skeptical to keeping these tests. I would bet they are absolutely never executed. IMO if it doesn't run in CI, it may as well not exist.

src/test/java/spoon/test/GitHubIssue.java
slarse: I'm also thinking that we might not want a default for this. If there's no default, then you have to set e.g. `fixed = false` explicitly when first annotating a test, and that makes it obvious for someone coming in later what to do when the related issue has been fixed.

slarse (commented):
I think it all looks very good. Just have some minor comments.

MartinWitt:
As this is a pretty neat utility I would like to use in other projects, are you fine with keeping this code here and me providing a separate artefact on central with this code?

monperrus:
> are you fine with keeping this code here and me providing a separate artefact on central with this code?

yes!

MartinWitt:
FYI the artifact is now on central located here https://search.maven.org/artifact/io.github.martinwitt/GithubIssue

monperrus:
Merged! The artifact is meant for opportunistic reuse elsewhere.

Thanks a lot @MartinWitt

PR #4591, opened 2022-02-08
WIP: add pitest

MartinWitt:
Currently testing, I will add some reasoning for it later today.


MartinWitt:
Sadly, pitest with real git support is only usable with a commercial license.

Issue #3890, opened 2021-04-26
continuous delivery CD pipeline for releases


monperrus:
Requirements
  * [x] a new script for normal releases (shell or python) (spec below)  
  * [x] add CI job to handle changes in the release spec https://github.com/SpoonLabs/spoon-deploy/pull/6  

Normal release is everything which is not a beta

monperrus:
FYI, CD of gumtree-spoon is also still Travis-based;

slarse:
I've got this on next week's TODO list.

slarse:
By next week I of course meant next month when Travis CI has already stopped working.

Looking at this right now. It's of course possible to just port the CD script to some other platform, but while we're at it, we may want to consider simplifying the CD process. The script is fairly complex and fragile. One example of extreme fragility is that it relies on the version number in `pom.xml` being on line 21. If I were to run the same formatting on the `spoon-core` pom file as I do on the parent pom file in #3994, then the version would be bumped down to line 22, breaking the CD script. That seems unreasonably fragile to me.

So there are two goals here: make the CD less fragile, and make it work on some other platform. We can deploy either from GitHub Actions or Jenkins. As this is a critical process, I think we should run it on Jenkins for complete control.

@monperrus I suggest the following changes.

### Separate the beta releases and normal releases into different scripts/jobs

These are two fundamentally different concerns. We want to do a beta release every X days, while we want to do a normal release on specific commits (bump commits).

### Permanently fix the Javadoc plugin problems

Currently, the CD script is injecting stuff into the pom to fix some crash on Travis. I don't see why we can't fix this permanently by just fixing the pom file. According to the linked issue, a workable solution is to just [set the `maven.compiler.source` property to `8`](https://github.com/assertj/assertj-parent-pom/commit/9b521bc33ac81b9a809de39c00599a9f928934d8).

### Normal release changes

Currently, the normal release is where we have the severe fragility I mentioned above. We can get rid of that by running the normal release script on every commit, but exiting immediately if the commit in question is not a bump commit. Then, instead of searching for changes on a specific line, we just compare the results of `mvn help:evaluate -Dexpression=project.version -q -DforceStdout` on the current and previous commit. If they differ, we have a bump!

Instead of semi-manually fiddling around in the pom with xmlstarlet, I also think it would be much better to use the Maven release plugin.

### Current priorities

[...]

PR #4595, opened 2022-02-11
chore: use GitHub actions for beta releases

algomaster99:
Reference: https://github.com/INRIA/spoon/issues/3890#issuecomment-1035118530

I have shifted the scripts from [spoon-deploy](https://github.com/SpoonLabs/spoon-deploy) to this repository itself. It now also uses GitHub actions instead of Travis.

Before merging this PR, there are some secrets that the maintainers need to configure. `SONATYPE_USER`, `SONATYPE_PASSWORD`, and `SPOONBOT_PARAPHRASE`. The first two are the credentials for ossrh server and the paraphrase is for decrypting the gpg key submitted in this pull request. @monperrus I sent you all these credentials over an encrypted channel in October last year. I think the credentials will be the same for Spoon as well. If indeed that is the case, could you configure them as secrets of this repository? I shall forward you that e-mail once again.

I skipped configuring [this script](https://github.com/SpoonLabs/spoon-deploy/blob/master/gpg-create-key.sh). We can do that in a separate PR.


monperrus:
Great!

As before, we would to put the CD job in the spoon-deploy repo. Would you move the PR there?

algomaster99:
> As before, we would to put the CD job in the spoon-deploy repo. Would you move the PR there?

Any specific reason why? The deployment is related to spoon and in my opinion, it should be a part of workflow this repository instead.

slarse:
I'm also curious about this. Putting deployments in a separate repo adds complexity, so there must be some form of payoff for it to be worthwhile. Especially when using GitHub Actions.

monperrus:
> Any specific reason why?
> so there must be some form of payoff for it to be worthwhile

The advantage is decoupling. Since CD is meant to be executed only for commits on master, it's really hard to test and debug. That's why you would see so many commits "trying CD on master" / "merging for CD", etc.

When you externalize CD to another repo, development, configuration and maintenance is easier. This is why we created https://github.com/SpoonLabs/spoon-deploy in the first place (which indeed contains many trial-and-error commits)

Some called such external CI/CD repo a "meta-repo" see https://medium.com/@marcguilera/effective-meta-repo-ci-cd-pipelines-c18676443d

algomaster99:
> it's really hard to test and debug. That's why you would see so many commits "trying CD on master" / "merging for CD", etc.

That's a really good point. Debugging is indeed easier.

PR #4598, opened 2022-02-12
test(smpl): migrate SMPL tests to JUnit5

MartinWitt:
Let me try to reduce the diff first, before anyone reviews. ~~And fix my git usage....~~

Edit 1: Git usage fixed.

### DIFF
--- a/spoon-smpl/src/test/java/spoon/smpl/C4JSetTextSizeTest.java
+++ b/spoon-smpl/src/test/java/spoon/smpl/C4JSetTextSizeTest.java
@@ -22,13 +22,14 @@
 
 package spoon.smpl;
 
-import org.junit.BeforeClass;
-import org.junit.Test;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
 import spoon.reflect.declaration.CtConstructor;
 
-import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.fail;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.fail;
 
 /**
  * Note: the patch used in this test has been significantly modified from the C4J original.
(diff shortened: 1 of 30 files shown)


algomaster99:
@MartinWitt don't force push the reduced diff. I would want to analyse what are the code elements that can be removed to minimise diff.

spoon-smpl/src/test/java/spoon/smpl/C4JSetTextSizeTest.java
```diff
@@ -22,52 +22,23 @@
...
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.fail;
 
 /**
  * Note: the patch used in this test has been significantly modified from the C4J original.
  */
 public class C4JSetTextSizeTest {
 	private static ZippedCodeBaseTestContext ctx = null;
 
-    /*@BeforeClass
```
MartinWitt, line 39: was this comment important?

algomaster99: I don't think so. But there was a change like this:
```diff
-(method) ->
+method ->
```

Paranthesis were removed.

MartinWitt: This is a result of the new DJPP with lambda fix https://github.com/INRIA/spoon/pull/4448

algomaster99: > This is a result of the new DJPP with lambda fix #4448

Right. I am trying to add an option to GitHub which removes such changes as well. We feel that minimising such changes will help aid the review process.

MartinWitt: PR with large diffs which could be reduced? I would mine huge git repos from scientific non industry grade software, because there should huge time pressure and  code quality is often not a goal there. Larger merges are probably more typically filled with these noises. Another good candidate could be students developing software having their first contact with git. There could be data, a problem, but at KIT I believe we have for starters a programming course where some use git horribly. Other universities could have this data too. Maybe you could ask them? If you need help with mining GitHub, feel free to ask or mention me.

MartinWitt: > Right. I am trying to add an option to GitHub which removes such changes as well. We feel that minimising such changes will help aid the review process.

I strongly believe creating a list of add/deleted/modified AST elements should be possible with spoon. 
Then you map from git diffs to AST elements using their position(fair warning here, I'm unsure if this is possible). Then you should have list of changed AST elements, which are either semantic or formatting changed. To detect the formatting changes, you could compare their `toString()` results. Using the DJPP for printing should produce deterministic results.

slarse: > I strongly believe creating a list of add/deleted/modified AST elements should be possible with spoon.

Indeed it is :) https://github.com/spoonlabs/gumtree-spoon-ast-diff

That being possible was the basis of my master's thesis where I made a merge tool on top of Spoon, see https://github.com/kth/spork

algomaster99: We already have a way to get the AST diffs like Simon said.

> Then you map from git diffs to AST elements using their position(fair warning here, I'm unsure if this is possible).

That's a really great idea. Thanks! Getting AST element from source position is something what Sorald also does. See [`BestFitScanner`](https://github.com/SpoonLabs/sorald/blob/master/src/main/java/sorald/sonar/BestFitScanner.java).

Thanks for the idea about mining GitHub repositories. I will look for datatsets which focus on "bad code quality".

MartinWitt:
The diffs should be now clean, only the imports are still an open problem.

PR #4601, opened 2022-02-13
rewview: chore(Qodana): Let Qodana fail CI again

MartinWitt:
While adding the new flag, `--changes` I forgot to remove the `continue-on-error:` flag. @I-Al-Istannen found this(see https://github.com/INRIA/spoon/runs/5170049523?check_suite_focus=true). Classical how to test your CI problem.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -162,4 +162,35 @@ jobs:
           inspected-dir: ./src/main/java # only main spoon project at first
           results-dir:  "${{ github.workspace }}/result"
           changes: true
-          fail-threshold: 0
+          fail-threshold: 1
+          upload-result: false
+      - name: Print problems(full)
+        run: jq '.runs | .[0].results | map(select (.baselineState == "new"))' ${{ github.workspace }}/result/qodana.sarif.json
+      - name: Print problems(short)
+        run: |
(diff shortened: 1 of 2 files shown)


I-Al-Istannen (approved):

MartinWitt:
As @I-Al-Istannen was pointing me out, the GitHub permission system for code quality checks is bad at best. Pull requests from fork repos can write the sarif annotations to the PR. This wasn't detected by me, because I had write access. To solves this problem I asked the Qodana team for access to their app(see https://www.jetbrains.com/help/qodana/qodana-github-application.html#how-to-start-github-app).
I believe they will answer very fast, so we don't really need a short term solution. A short term solutions would be using 2 workflows and a github token(see https://blog.pitest.org/pitest-pr-setup/)

MartinWitt:
Okay, JetBrains has answered me, and it looks like their app is deprecated. So we have to go back to the nice JQ code from @I-Al-Istannen.

MartinWitt:
https://github.com/INRIA/spoon/runs/5191847123?check_suite_focus=true as we see here we have the old behaviour back and it works for all PRs again.

MartinWitt:
#4608 provides a way better fix. Closed

PR #4604, opened 2022-02-14
chore: add GPG plugin

algomaster99:
The [deployment might be failing](https://github.com/SpoonLabs/spoon-deploy/runs/5187204138?check_suite_focus=true) because spoon did not have a gpg plugin installed. Its docs [here](https://maven.apache.org/plugins/maven-gpg-plugin/usage.html) also suggest to configure it first before signing package with the GPG key.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -305,6 +305,20 @@
            </execution>
         </executions>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-gpg-plugin</artifactId>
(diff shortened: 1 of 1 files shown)


algomaster99:
@monperrus [This error](https://github.com/SpoonLabs/spoon-deploy/runs/5187204138?check_suite_focus=true#step:4:255) is most likely arising due to the absence of a gpg plugin.

algomaster99:
Okay, so the missing plugin is not a problem. Thanks!

PR #4605, opened 2022-02-14
fix: fix signing of spoon artifact

algomaster99:
I am not very sure about the problem, but I found an [article](https://myshittycode.com/2017/08/07/maven-gpg-plugin-prevent-signing-prompt-or-gpg-signing-failed-no-such-file-or-directory-error/) quite relevant to the error shown [here](https://github.com/SpoonLabs/spoon-deploy/runs/5187204138?check_suite_focus=true#step:4:254).

### DIFF
--- a/spoon-pom/pom.xml
+++ b/spoon-pom/pom.xml
@@ -397,6 +397,12 @@
                                 <goals>
                                     <goal>sign</goal>
                                 </goals>
+                                <configuration>
+                                    <gpgArguments>
+                                        <arg>--pinentry-mode</arg>
+                                        <arg>loopback</arg>
+                                    </gpgArguments>
+                                </configuration>
                             </execution>
                         </executions>
                     </plugin>


algomaster99:
I thought of another solution. I will reopen this if need be.

slarse:
@algomaster99 I again recommend looking at Sorald's deploy workflow, I went through all of these problems about a year ago :)

algomaster99:
@slarse I am referring to gumtree-spoon and sorald's deploy workflow. The process they all follow is more or less the same. The only difference in all three is how they compute the next version for deployment.

> I thought of another solution. I will reopen this if need be.

When I wrote this, the error in the CI was that it could not find the public key corresponding to the secret Spoon Bot's GPG key we are using so I just uploaded it to `keyserver.ubuntu.com`. Now it is complaining about [`Nexus connection problem`](https://github.com/SpoonLabs/spoon-deploy/runs/5197873792?check_suite_focus=true#step:4:280). I thought rerunning the job would fix it, but that is not the case.

monperrus:
> so I just uploaded it to keyserver.ubuntu.com.

:+1:

PR #4606, opened 2022-02-14
doc: Add security policy

### LINKED ISSUE #4603: How to report vulnerabilities?
As the title says: I don't know how to report issues that might impact security? I think it's not advisable to open a publicly visible issue for such things, so how should the maintainers be contacted?

slarse:
Fix #4603 

Here's a proposal for a security policy.

I also added emails and the PGP key fingerprints hat I know of in the CONTRIBUTING.md file. I refer to them as GPG keys, though,  because I think that's more common nowadays, even though it's technically incorrect. I mean, even I say GPG key, and I'm pretty uptight about being correct ...

@monperrus @nharrand @MartinWitt please have a look.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,9 +18,15 @@ How to become integrator? The integrators are the developers who have made signi
 Current integrators:
 
 - Simon Larsén [@slarse](https://github.com/slarse/)
+    - Email: slarse@slar.se
+    - GPG fingerprint: 70345F27028C791C8D82D3DE1CEF06827CEA5C29
 - Nicolas Harrand [@nharrand](https://github.com/nharrand/)
+    - Email: nicolas.harrand@gmail.com
 - Martin Monperrus [@monperrus](https://github.com/monperrus/)
+    - Email: martin.monperrus@gnieh.org
+    - GPG fingerprint: 074F73B36D8DD649B132BAC18035014A2B7BFA92
 - Martin Wittlinger [@MartinWitt](https://github.com/MartinWitt)
+    - Email: wittlinger.martin@gmail.com
 
 Guidelines for pull requests
 ----------------------------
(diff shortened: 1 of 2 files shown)


MartinWitt (approved):

intrigus-lgtm:
Do you know [Github Security Advisories](https://docs.github.com/en/code-security/security-advisories/about-github-security-advisories)?
Might be easier to use, because using PGP can be pretty annoying.

The flow would look like this:
1. Reporter informs this project about the fact that they found a security problem (**without** disclosing any information about the vulnerability). For example, when I report security vulnerabilities I often create an issue, asking the maintainer to create an security advisory. Or I ask them via email.
2. A security advisory is created. All admins are invited to it. Additionally an admin invites the reporter to the advisory. (Any github user can be invited)
3. The reporter and admin can discuss the vulnerability in private without having to use PGP/mail. You can also collaborate on a private branch on the fix of the problem.
4. If the vulnerability is indeed a vulnerability, you can request a CVE from github using pretty much 1 click.

slarse:
Hi @intrigus-lgtm,

I knew in passing about GitHub security advisories, but I don't think any of us were aware that one could invite people like that. But I agree that sounds easier for most. We still want to keep the email route open because we don't assume that all users of Spoon have a GitHub account, but for those that to that sounds pretty good.

@monperrus @MartinWitt thoughts? Add the suggested workflow?

monperrus:
Yes, GitHub security advisories are also welcome.

FTR, to create a new advisory, navigate to https://github.com/INRIA/spoon/security/advisories/new

slarse:
@monperrus Only admins can do that, though. I don't have access, for example.

So the flow would be like @intrigus-lgtm suggest, that a user somehow (e.g. through email or an issue) contacts an admin to say that there _IS_ a security problem, without detailing what that is. The admin then opens a security advisory, and invites the reporter to it.

I will add this to the security policy tonight. Thanks @intrigus-lgtm for the suggestion.

PR #4608, opened 2022-02-15
Use Codescanning instead of annotations.

intrigus-lgtm:
The Github Codescanning API is accessible by **forks** and also allows creating annotations based on a SARIF file.

So instead of using the Checks API (not accessible by forks) to create annotations we using Codescanning.
I think this should probably give us the same results.

As an example see
https://github.com/intrigus-lgtm/spoon/pull/10/files.
The PR has been created by @intrigus (i.e. is from a fork) and successfully received annotations :tada:

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -164,3 +164,7 @@ jobs:
           results-dir:  "${{ github.workspace }}/result"
           changes: true
           fail-threshold: 0
+          use-annotations: false
+      - uses: github/codeql-action/upload-sarif@v1
+        with:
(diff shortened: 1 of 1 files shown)


intrigus-lgtm:
Qodana now fails as expected on violations:
https://github.com/intrigus-lgtm/spoon/runs/5222530347?check_suite_focus=true

MartinWitt (approved):

MartinWitt:
> I had no idea about this, very cool!

Me neither, but sometimes knowing an expert is helpful :D

MartinWitt:
Thanks @intrigus-lgtm

PR #4610, opened 2022-02-16
Fix incorrect import

ceki:


MartinWitt:
Good catch, thank you @ceki. As well known persons don't come by often, what is your spoon use case?

ceki:
@MartinWitt  Nothing overly exciting I am afraid. I am merely evaluating various Java code analysis tools.

PR #4612, opened 2022-02-19
review: refactor: remove useless toString calls

MartinWitt:
This removes not needed toString calls. I'm still unsure about when a `toString()` call is not needed, but these changes seem correct. The detection was done by Qodana and this fixes 4 issues. Calling `toString` in String concatenation adds zero value to the code and should be removed.

### DIFF
--- a/src/main/java/spoon/experimental/SpoonifierVisitor.java
+++ b/src/main/java/spoon/experimental/SpoonifierVisitor.java
@@ -257,12 +257,26 @@ public class SpoonifierVisitor extends CtScanner {
 		@Override
 		public void visitCtLiteral(CtLiteral element) {
 			if (element.getType().isPrimitive()) {
-				result.append(printTabs() + variableName + ".setValue((" + element.getType().getSimpleName() + ") " + element + ");\n");
+				result.append(printTabs())
+							.append(variableName)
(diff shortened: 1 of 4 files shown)


src/main/java/spoon/experimental/SpoonifierVisitor.java
```diff
@@ -225,7 +225,11 @@ public void exit(CtElement element) {
 		if (!roleContainer.peek().isEmpty()) {
 			for (CtRole role: roleContainer.peek().keySet()) {
 				String variableName = roleContainer.peek().get(role);
-				result.append(printTabs() + parentName.peek() + ".setValueByRole(CtRole." + role.name() + ", " + variableName + ");\n");
+				result.append(printTabs())
+						.append(parentName.peek())
+						.append(".setValueByRole(CtRole.")
+						.append(role.name()).append(", ")
+						.append(variableName).append(");\n");
```
slarse, line 232:
```suggestion
						.append(role.name())
						.append(", ")
						.append(variableName)
						.append(");\n");
```

slarse (changes requested):
@MartinWitt Just one piece of inconsistent formatting, otherwise LGTM.

slarse (approved):

PR #4613, opened 2022-02-20
test: fix junit extension and false flags

MartinWitt:
It appears that my IDE fooled me, and failing test cases from a 'Testwatcher' is not permitted (the exception is only logged), but failing test cases from a 'AfterEachCallback' is permitted. It should now work properly. Using a 'Testwatcher' may work depending on your junit-engine version (see https://github.com/junit-team/junit5/issues/1775). I've added four test methods to this PR so that the reviewer can confirm that it now works correctly. Following that, the methods will be removed.
And this PR moves the state from the extension to the junit store mechanism.


slarse:
@MartinWitt What's the state of this PR?

slarse:
@MartinWitt Looks passing now, ready for review?

MartinWitt:
Yes

I-Al-Istannen (commented):
I'd suggest testing that, but I am not sure if you can even do that without causing surefire to fail… Looks similar enough to my fix though :)

slarse (commented):
@MartinWitt Confirmed that it works. If you remove the test cases I can approve for merge!

MartinWitt:
The test case is already removed, or did I miss something?

slarse:
> The test case is already removed, or did I miss something?

I read the opening post and assumed that the parameterized test that you split up were the "test tests" :face_exhaling: 

But now I see that was just the split of the pre-existing test.

slarse:
@MartinWitt Ready to merge, yes? Looks all spiffy to me.

MartinWitt:
Yes ready2merge.

PR #4614, opened 2022-02-20
chore: check reproducible builds in CI

MartinWitt:


MartinWitt:
There was an issue a time ago, and it was never implemented, see https://github.com/INRIA/spoon/issues/4151. I found the maven plugin and thought this is easy. Turns out the plugin is near to unusable and is missing basic features.

algomaster99:
Do we need to use a maven plugin for this? It seems that we only need to check if the contents of the artefacts created are equal. How about if we use `cmp` for that?

MartinWitt:
> How about if we use cmp for that?

As I see the state of the plugin, this is a way better way. Sadly, my Unix/shell command knowledge is limited. You are welcome to create a PR checking the content of 2 artefacts in a CI step.

algomaster99:
Sure. I can try having a look at this later this evening or tomorrow perhaps.

algomaster99:
I ran the following command, however it exits with `1`. There are some changes in the file, but I am not sure if they are relevant.
```bash
cmp build/spoon-core-10.0.1-SNAPSHOT-jar-with-dependencies.jar target/spoon-core-10.0.1-SNAPSHOT-jar-with-dependencies.jar
>>> differ: byte 11, line 2
```

I-Al-Istannen:
> @MartinWitt I have got it to work with `cmp`. I just needed to upgrade the `maven-assembly-plugin` to `3.2.0` and change its goal from `attached` to `single` and it worked. Maybe you can try doing that in this PR. If that does not work, we can resort to using `cmp`.

@algomaster99 That's an interesting question. My PR bumped the jar plugin and made the jar the `maven-jar-plugin` creates reproducible. This jar contains the artifacts for spoon itself.
Your suggestions does the same thing for the assembled artifact *with dependencies*. This jar is not distributed by maven (I hope?) but is one the end user can link against themselves.

Which JAR(s) do we want to verify?

algomaster99:
> My PR bumped the jar plugin and made the jar the maven-jar-plugin

Ah, okay. Then the jar without dependencies would be the same at the bytecode level not the assembled one.

> Which JAR(s) do we want to verify?

I chose the jar with dependencies because you can execute it as-is. Meanwhile, `spoon.jar` can also be executed but you need to explicitly provide its dependencies in the classpath. 

> This jar is not distributed by maven (I hope?) 

It is distributed. See https://repo1.maven.org/maven2/fr/inria/gforge/spoon/spoon-core/10.0.1-beta-9/. Is that not correct practice?

I-Al-Istannen:
> I chose the jar with dependencies because you can execute it as-is. Meanwhile, spoon.jar can also be executed but you need to explicitly provide its dependencies in the classpath.

Yea, we need all artifacts from maven central anyways. I opened a PR bumping the assembly plugin (linked in the github history in this PR).

> It is distributed. See [repo1.maven.org/maven2/fr/inria/gforge/spoon/spoon-core/10.0.1-beta-9](https://repo1.maven.org/maven2/fr/inria/gforge/spoon/spoon-core/10.0.1-beta-9/). Is that not correct practice?

It's the first time I saw that. Maven does dependency resolution for you and it allows you to override which versions are used or exclude certain transitive dependencies. I suppose it does no harm to additionally distribute the standalone jar for people not using it though.

algomaster99:
> Maven does dependency resolution for you

Yes, that makes sense. But if someone just wants to take Spoon for a spin they can always do so:
```
java -jar spoon-jar-with-dependencies.jar Main
```
where Main is a class that is dependent upon Spoon.

PR #4618, opened 2022-02-21
test: migrate dataflow to Junit 5

MartinWitt:


MartinWitt:
@tiulpin another case where Qodana runs endless, :( see https://github.com/INRIA/spoon/runs/5275032392?check_suite_focus=true for the run.

tiulpin:
@MartinWitt thank you. It's sad to hear, we will investigate – https://youtrack.jetbrains.com/issue/QD-2136
Our previous attempt to find the problem wasn't successful, but we'll try again!

MartinWitt:
Currently, we more or less have to hope your tests catch all bugs before you release a new qodana version for the same version tag. Have you thought about offering more fine granular versions? Normal CI tools offer pinnable versions (see [sonar-java](https://search.maven.org/artifact/org.sonarsource.java/sonar-java-plugin))sometimes even down their hash for reproducibility(see more or less any github action), and not cases like this?

tiulpin:
@MartinWitt I'm sorry for the inconvenience. 

>  Have you thought about offering more fine granular versions? Normal CI tools offer pinnable versions (see [sonar-java](https://search.maven.org/artifact/org.sonarsource.java/sonar-java-plugin))sometimes even down their hash for reproducibility(see more or less any github action), and not cases like this?

You can pin it to the hash, because every Qodana linter is packed inside a Docker image: you can set the linter in the action to `jetbrains/qodana-jvm-community@sha256:c388ddd7b852296f71b1a786d3b30f458a993eb252d5cd4963422007e5b824db`.

> we more or less have to hope your tests catch all bugs before you release a new qodana version for the same version tag

Yes, we do test Qodana on a lot of projects, but, unfortunately, this bug was not easy to catch for us. But we finally reproduced it: the build hanged because of the incorrect caches usage from the previous versions. 

As a workaround, you can still use the latest 2021.3 version, but you'll need to drop the caches (that can be disabled in the action for one run and enabled again). We'll fix the problem ASAP and will add more tests for update scenarios. 

Thank you for reporting this!

slarse (approved):
LGTM, thanks @MartinWitt

slarse:
Thanks for the input @tiulpin, we've now disabled the cache with  #4621 for the time being :)

PR #4621, opened 2022-02-22
chore: pinn Qodana version and disable cache

MartinWitt:
I disabled the cache because Qodana is fast enough, and I favour a working CI over a fast but unreliable CI.


slarse (approved):
Thanks @MartinWitt 

Going to merge fast-merge this as it seems to be the only surefire way of having Qodana not get stuck (see https://github.com/INRIA/spoon/pull/4618#issuecomment-1047782236).

PR #4624, opened 2022-02-23
refactor: provide access to JDT and Spoon parent in ParentExiter

monperrus:


monperrus:
@algomaster99  It's not necessary, its a refactoring for improving readability. Further improved the refactoring thanks to your comment.

algomaster99 (approved):

MartinWitt:
This is interestingly another Qodana bug, `stack.peek()` should only return null if it is empty. Which we check before.

I-Al-Istannen:
I had a look at find cause:
grafik

The message does *not* occur if you move the peek call to be the first statement after the check:
grafik

Apparently IJ is a bit paranoid about the method calls there, even though they currently do not affect `stack` in any way. @monperrus could move it up there to please the analysis or we merge it anyways (which should not break master as the analysis only checks the delta between master and the tested commit).

monperrus:
> which should not break master as the analysis only checks the delta between master and the tested commit

good to know, we can safely merge then

monperrus:
ping @MartinWitt @slarse for merge

monperrus:
thanks @MartinWitt

Issue #4274, opened 2021-11-10
Questions about Sniper prettyprinter


algomaster99:
Hi! I was trying to look into #4220, and while debugging it, I came up with some questions about the sniper printer. I will first write down how I reproduced it and then ask questions about things I didn't understand while debugging it.

### Test case

```java
public class SniperTest {
    @Test
    public void test() {
        Launcher launcher = new Launcher();
        Environment environment = launcher.getEnvironment();

        // enable sniper pretty-printer
        environment.setPrettyPrinterCreator(() -> new SniperJavaPrettyPrinter(environment));
        launcher.addInputResource("src/test/resources/SpaceBetweenModifiers.java");

        CtModel model = launcher.buildModel();
        CtField field = (CtField) model.getElements(e -> e.getClass().equals(CtFieldImpl.class)).get(0);
        // change the type of field to `Integer` from `String`
        field.setType(launcher.getFactory().Type().integerType());

        // print a compilation unit
        CtCompilationUnit cu = List.copyOf(model.getUnnamedModule().getFactory().CompilationUnit().getMap().values()).get(0);
        PrettyPrinter printer = cu.getFactory().getEnvironment().createPrettyPrinter();
        System.out.println(printer.printCompilationUnit(cu)); // I will refer to this line as output
    }
}
```

### Test resource

**Input**
```java
class SpaceBetweenModifiers {

    private static final java.lang.String x;

    public TestSniper()
    {
        x = "Test";
    }
}
```
**Output**
```java
class SpaceBetweenModifiers {

    private static finaljava.lang.Integer x;

    public TestSniper()
    {
        x = "Test";
    }
}
```
Note that there is no space between `final` and `java.lang.Integer` as pointed out by the issue linked above.

## Questions

- What does the `launcher.prettyprint()` do? It returns void, although I expected it to return a `String` because of the method's name - pretty _print_. I can print the compilation unit using the sniper pretty print even without calling this method. So what is its purpose?
- How is `CtElement.toString()` different from prettyprinting? For example, I checked the output of `model.getAllTypes().stream().toArray()[0].toString()` and it returned the string representation of the type `SpaceBetweenModifiers` with _all spaces intact_. Moreover, it could be compiled without errors.
- What are [`listener`] and [`delegate`](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/sniper/internal/TokenWriterProxy.java#L34-L35) fields for?

### Question related specifically to the issue.

> I will link this later to #4220 

[...]

PR #4637, opened 2022-03-10
doc: Remove incorrect Maven usage of spoon-control-flow

LakshyAAAgrawal:


monperrus:
Thanks! Since the two changes are very different in nature, would you split them in two PRs?

monperrus:
Thanks a lot @LakshyAAAgrawal

monperrus:
For the record, should you be ready to create another PR, the FAQ suggestion is

>  What does launcher.prettyprint() do?It returns void, although I expected it to return a String representation of the model.
```launcher.prettyprint``` prints the model (could be transformed or not) in an output directory which one can set up with ```launcher.setSourceOutputDirectory()``` (there is a default one too). If you want the output in the STDOUT, use toString method defined for every CtElement. It will first get the prettyprinter from the environment and return the result.
Source: https://github.com/INRIA/spoon/issues/4274

LakshyAAAgrawal:
Thanks a lot @monperrus and sorry for the late response, I had been away for the past few days. I will create another PR.

PR #4646, opened 2022-03-22
doc: document our versioning process

monperrus:
Our versioning convention is not documented. Here it is.

### DIFF
--- a/doc/_release/README.md
+++ b/doc/_release/README.md
@@ -2,11 +2,11 @@
 
 ## Versioning
 
-Spoon uses a three digit version number MAJOR.MINOR.HOTPATCH
+Spoon uses a three digit version number MAJOR.MINOR.HOTPATCH. We follow semantic versioning with the exception that we may increment the major version number for significant new features even when there are no breaking changes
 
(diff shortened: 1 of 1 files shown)


doc/_release/README.md
```diff
@@ -1,4 +1,14 @@
-# How to release Spoon?
+# Spoon Releases
+
+## Versioning
+
+Spoon uses a three digit version number MAJOR.MINOR.HOTPATCH
+
+* we bump MAJOR when there is a big new feature or a strongly breaking change. We use it mostly for new features since we take special care to backward compatibility.
+* we bump HOTPATCH when a normal release is not operational.
```
MartinWitt, line 8: Maybe change the order of the points to follow the order above `MAJOR.MINOR.HOTPATCH` ?

slarse:
The only difference between this versioning and SemVer 2 is that we also increment the major version number for significant new features. IMO we could replace this text with: "We follow semantic versioning, with the exception that we may increment the major version number for significant new features even when there are no breaking changes"

monperrus:
added @slarse 's proposed sentence and switched the two bullets per @MartinWitt 's suggestion.

ready to merge.

MartinWitt:
Thanks @monperrus

PR #4652, opened 2022-03-28
chore: Ensure assembly jar with dependencies is reproducible

I-Al-Istannen:
This can be achieved by just updating the assembly plugin.

This is preparation for #4151 and #4614.


monperrus:
LGTM, thanks @I-Al-Istannen.

PR #4655, opened 2022-03-30
refactor: make use of InputStream#transferTo where appropriate

SirYwell:
I found places where spoon manually read from an input stream just to write to an output stream. Since Java 9, there is a [InputStream#transferTo](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/io/InputStream.html#transferTo(java.io.OutputStream)) method that provides the same functionality and is easier to read.

To make it more consistent, I also replaced usages of `IOUtils#copy`, as it does the same than the new method.

### DIFF
--- a/src/main/java/spoon/compiler/SpoonFile.java
+++ b/src/main/java/spoon/compiler/SpoonFile.java
@@ -7,7 +7,6 @@
  */
 package spoon.compiler;
 
-import org.apache.commons.io.IOUtils;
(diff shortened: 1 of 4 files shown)


MartinWitt (approved):
Using newer API and reducing dependencies is always good. LGTM, will merge 👍

monperrus:
:+1:

MartinWitt:
Thanks @SirYwell

PR #4657, opened 2022-03-31
review: refactor: replace usages of IOUtils with Java NIO API

SirYwell:
I replaced remaining usages of IOUtils with modern Java APIs.

From what I've seen, `commons-io` isn't in use anymore, so we can probably remove it in a following PR.

### DIFF
--- a/src/test/java/spoon/test/imports/ImportTest.java
+++ b/src/test/java/spoon/test/imports/ImportTest.java
@@ -17,6 +17,7 @@
 package spoon.test.imports;
 
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
 import spoon.Launcher;
 import spoon.SpoonException;
 import spoon.SpoonModelBuilder;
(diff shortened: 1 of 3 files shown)


src/test/java/spoon/test/imports/ImportTest.java
```diff
@@ -1034,8 +1033,8 @@ public void testmportInCu() throws  Exception{
 		launcher.addInputResource("./src/test/java/"+path);
 		launcher.run();
 
-		File output = new File("target/testmportInCu/"+path);
-		String code = IOUtils.toString(new FileReader(output));
+		Path output = Path.of("target/testmportInCu/" + path);
```
slarse, line 1036:
```suggestion
		Path output = Path.of("target/testmportInCu/").resolve(path);
```

src/test/java/spoon/test/imports/ImportTest.java
```diff
@@ -1047,7 +1046,7 @@ public void testmportInCu() throws  Exception{
 		assertThat(code, containsString("ArrayList<String> list = new ArrayList<>()"));
 
 		// cleaning
-		output.delete();
+		Files.delete(output);
```
slarse, line 1049: Test cleanup after asserts isn't great, this should be in a `finally`. Or better yet we should use a tempfile from junit5.

slarse (changes requested):
Looks good, but I think we can improve one of those test cases just a li'l bit more.

slarse (approved):
LGTM, thanks @SirYwell

PR #4671, opened 2022-04-10
review: fix: Use thread-local caching for actual class lookup

### LINKED ISSUE #4668: A question regarding how CtTypeReferenceImpl retrieves its type.
I was going through the code of CtTypeReferenceImpl to see how it resolves its type declaration, and I noticed that there is a lock operation going on in the `findClass` method:

https://github.com/INRIA/spoon/blob/17e000b58363dee4ee1687e582f13ef993b387e0/src/main/java/spoon/support/reflect/reference/CtTypeReferenceImpl.java#L163-L169

The `lastClassLoader` and `classByQName` fields are static, but the lock itself is not, meaning there will be a different lock for each type reference created...

https://github.com/INRIA/spoon/blob/17e000b58363dee4ee1687e582f13ef993b387e0/src/main/java/spoon/support/reflect/reference/CtTypeReferenceImpl.java#L56-L59

I feel like this may cause multithreading issues. Was this intended?

slarse:
Fix #4668

This is a quick-fix to avoid the cross-thread race condition when invalidating the cache and then doing a lookup in it. This is a problem as the current lock is per-instance but the cache is static, so in theory a thread A could check the cache integrity and state that it's fine, then there's a context switch to thread B which does the same thing and invalidates the cache and inserts something into the cache, and then we're back to thread B which does a lookup on a "bad cache" (according to thread A's view of the world).

By using thread-local storage, this situation cannot occur.

### DIFF
--- a/src/main/java/spoon/support/reflect/reference/CtTypeReferenceImpl.java
+++ b/src/main/java/spoon/support/reflect/reference/CtTypeReferenceImpl.java
@@ -160,14 +160,16 @@ public class CtTypeReferenceImpl<T> extends CtReferenceImpl implements CtTypeRef
 			}
 			typeReference = componentTypeReference;
 		}
-		ClassLoader classLoader = getFactory().getEnvironment().getInputClassLoader();
+		Class<T> actualClass = getClassFromThreadLocalCacheOrLoad(typeReference);
+		return isArray() ? arrayType(actualClass) : actualClass;
+	}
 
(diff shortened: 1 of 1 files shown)


MartinWitt (approved):

MartinWitt:
LGTM as a short-term fix.

MartinWitt:
thanks, @slarse

PR #4672, opened 2022-04-10
review: refactor: clean up `generateProcessedSourceFilesUsingCUs` method

SirYwell:
The method contained some duplicated code and didn't make use of modern file system APIs. 

There also was a list which had no practical use as the elements were only added (with a contains check to prevent duplicates). I'm not sure if there were any intentions behind that, but removing it does not change behavior.

PR #4674, opened 2022-04-10
review: refactor: Use NIO to read compilation unit source code

I-Al-Istannen:
This was discovered while fixing #4672. I don't believe the `available` method makes any guarantees about its length, this looks like it could silently truncate the read source code.


slarse (approved):
Agreed, the docs clearly say that `available()` returns an _estimate_. I'd wager that it's actually exact as this has been working for so long, but that's an undocumented implementation detail that shouldn't be relied upon even if we can confirm it to be the case.

Will merge.

PR #4676, opened 2022-04-10
wip: refactor: use JDK nullOutputStream

SirYwell:
Construction a `NullOutputStream` is already deprecated, and we can use a JDK method here.

Note that the implementations are slighly different, the output stream provided by the JDK does not allow to write to it after it was closed. This shouldn't be an issue though, as this is the expected behavior for other output stream implementations anyways.


slarse (approved):
Thanks @SirYwell, will merge.

PR #4679, opened 2022-04-11
shore(deps) Update DepClean to version 2.0.2

cesarsotovalero:
Version 2.0.2 of DepClean has [been released](https://github.com/castor-software/depclean/releases/tag/2.0.2), which includes several improvements such as static analysis of dependencies at the source code level. For Spoon, this means that the exclusion of `org.kohsuke.metainf-services:metainf-services` is no longer needed in the `pom` of Spoon because now DepClean correctly identifies this dependency as used.


monperrus:
Thanks @cesarsotovalero

PR #4682, opened 2022-04-11
wip: refactor: use String#stripTrailing instead of custom method

SirYwell:
This replaces a custom method to remove trailing whitespaces with the new method provided by `String` directly.


monperrus:
thanks @SirYwell

PR #4685, opened 2022-04-12
refactor: use StringJoiner where appropriate

SirYwell:
Using StringJoiner allows to simplify a few places in code and make them more clear.


MartinWitt:
thanks @SirYwell

PR #4689, opened 2022-04-13
refactor: use EnumMap instead of dealing with ordinals for array indexes

SirYwell:
This array was used like an enum map, so I replaced it.


MartinWitt (approved):

MartinWitt:
LGTM will merge after 24h

MartinWitt:
thanks, @SirYwell

PR #4691, opened 2022-04-14
wip: refactor: Add ModelTest annotation

I-Al-Istannen:
This aims to clean up tests by adding a `ModelTest` annotation to provide a launcher, factory and model.

The usages were mostly found using intellijs structural search (template):
```xml
<searchConfiguration name="@Test&#10;public $ReturnType$ $Method$() {…" uuid="0cbc6611-f554-3bd0-809a-388dc95a615b" text="@Test&#10;public $ReturnType$ $Method$() {&#10;&#9;$Statements$;&#10;&#9;Launcher $launcher$ = new Launcher();&#10;&#9;$launcher$.addInputResource(&quot;$input$&quot;);&#10;&#9;$Statements2$;&#10;}" recursive="true" type="JAVA" pattern_context="member" search_injected="false" case_sensitive="true">
  <constraint name="__context__" within="" contains="" />
  <constraint name="ReturnType" within="" contains="" />
  <constraint name="Method" within="" contains="" />
  <constraint name="Statements" minCount="0" maxCount="2147483647" within="" contains="" />
  <constraint name="launcher" within="" contains="" />
  <constraint name="input" within="" contains="" />
  <constraint name="Statements2" minCount="0" maxCount="2147483647" within="" contains="" />
</searchConfiguration>
```
And then replaced using its structural replacement
```xml
<replaceConfiguration name="@Test&#10;public $ReturnType$ $Method$() {…" uuid="0cbc6611-f554-3bd0-809a-388dc95a615b" text="@Test&#10;public $ReturnType$ $Method$() {&#10;&#9;$Statements$;&#10;&#9;Launcher $launcher$ = new Launcher();&#10;&#9;$launcher$.addInputResource(&quot;$input$&quot;);&#10;&#9;$Statements2$;&#10;}" recursive="false" type="JAVA" pattern_context="member" search_injected="false" reformatAccordingToStyle="true" shortenFQN="true" replacement="&#9;@ModelTest(&quot;$input$&quot;)&#10;&#9;public $ReturnType$ $Method$(spoon.Launcher launcher) {&#10;&#9;&#9;$Statements$; &#10;&#9;&#9;$Statements2$;&#10;&#9;}" case_sensitive="true">
  <constraint name="__context__" within="" contains="" />
  <constraint name="ReturnType" within="" contains="" />
  <constraint name="Method" within="" contains="" />
  <constraint name="Statements" minCount="0" maxCount="2147483647" within="" contains="" />
  <constraint name="launcher" within="" contains="" />
  <constraint name="input" within="" contains="" />
  <constraint name="Statements2" minCount="0" maxCount="2147483647" within="" contains="" />
</replaceConfiguration>
```

grafik
As you can see on the lefthand side, we went through quite a few iterations and the reproduced queries above are just an excerpt.

Replacement was done automatically but cleaned up manually, as usages varied too widely and new flags needed to be added to the annotation. Finally, the whole thing was manually committed using `git add -p`, as intellij fixed thousands of absolutely awful formatting mistakes in the code that would have bloated the diff.

### DIFF
--- a/src/test/java/spoon/CompilationUnitPrintTest.java
+++ b/src/test/java/spoon/CompilationUnitPrintTest.java
@@ -10,15 +10,15 @@ import spoon.reflect.declaration.CtMethod;
 import spoon.reflect.factory.Factory;
 import spoon.reflect.visitor.DefaultJavaPrettyPrinter;
 import spoon.support.JavaOutputProcessor;
+import spoon.testing.utils.ModelTest;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
 public class CompilationUnitPrintTest {
 
-    @Test
-    public void test() {
-
-        /* Testing scenario:
+    @ModelTest("src/test/java/spoon/CompilationUnitPrintTest.java")
+    public void test(Factory factory) {
+		/* Testing scenario:
             Build a model with a class
             Clone this class
             Modify the clone by adding a method
@@ -28,12 +28,6 @@ public class CompilationUnitPrintTest {
          */
 
         // build original class
-        Launcher launcher = new Launcher();
-        launcher.addInputResource("src/test/java/spoon/CompilationUnitPrintTest.java");
-        launcher.getEnvironment().setNoClasspath(true);
-        launcher.buildModel();
-        final Factory factory = launcher.getFactory();
-
         // clone
         final CtClass<?> compilationUnitPrintTest = factory.Class().get("spoon.CompilationUnitPrintTest");
         final CtClass<?> clone = compilationUnitPrintTest.clone();
@@ -61,7 +55,7 @@ public class CompilationUnitPrintTest {
         assertEquals(1 , compilationUnitPrintTest.getMethods().size());
 
         // building now a new model from the java file outputted just before
-        launcher = new Launcher();
+        Launcher launcher = new Launcher();
         launcher.addInputResource("target/spoon/CompilationUnitPrintTest.java");
         launcher.getEnvironment().setNoClasspath(true);
         launcher.buildModel();
--- a/src/test/java/spoon/LauncherTest.java
+++ b/src/test/java/spoon/LauncherTest.java
@@ -33,6 +33,7 @@ import spoon.reflect.CtModel;
 import spoon.reflect.visitor.DefaultJavaPrettyPrinter;
 import spoon.support.JavaOutputProcessor;
 import spoon.support.compiler.VirtualFile;
+import spoon.testing.utils.ModelTest;
 
 import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertEquals;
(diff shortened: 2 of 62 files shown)


I-Al-Istannen:
Before you ban me from this project for only creating massive PRs, this wasn't all my own doing and @SirYwell and @MartinWitt are equally guilty.

We originally just wanted to tackle the `spooned` folder issue but somehow we ended up shaving this yak instead, as previously launchers were created all over the place. The next step would be removing all the manual calls to `setXXOutputDirectory` or rewiring them to a temp dir.

monperrus:
Overall, I like the concept of this PR very much.

The only problem is that it is very tedious to review.

Ideally, I would like to see an "assertion trace difference": the assertions that  are executed before and after this PR are strictly the same. Then, it would mean that it is a fully backward compatible test refactoring.

That's doable with a little bit of instrumentation. It's possibly a thesis topic and a research paper.

cc/ @algomaster99 @khaes-kth

I-Al-Istannen:
> Ideally, I would like to see an "assertion trace difference": the assertions that are executed before and after this PR are strictly the same. Then, it would mean that it is a fully backward compatible test refactoring.

Isn't this already guaranteed by virtue of them still existing and _passing_? I am not sure what exactly you are proposing here. My initial understanding was that you want to ensure the inputs to the assertions are the same, but that doesn't really make sense. If the assertion does not verify something, it is allowed to (and might!) differ from run to run.
Comparing assertion arguments is also not really possible, as the inputs might not be serializable (using Java's built-in mechanism or gson or w/e) and can therefore not be persisted between runs, preventing comparison.

Do you just want to see if any assertion was deleted (i.e. grepping for `- .+assert.*` in the git diff) and whether any assertions are no longer executed now? So a slightly more sophisticated version of comparing total executed assertion counts with and without the changes?

slarse:
@I-Al-Istannen I had a quick look and no assertions have been altered, so I do not share @monperrus concern. _However_, the test setup has obviously been altered (that's the point of the entire PR), and this means that the tests may not test exactly the same scenarios as before. They still pass, obviously, but it's entirely possible that a test could have been weakened or made completely vacuous.

For example, imagine a regression test that checks for a bug that appeared under very specific circumstances, and the test setup ensures those circumstances are actually met. Even a tiny change to the test setup might mean that the test no longer reproduces the bug (if reintroduced).

Like @monperrus I'm in favor of this change, but it needs to be brought in in much smaller chunks as reviewers need to validate that the test setup results in the same state as before. I propose that you make an initial PR with the new annotation and use it for just a few tests where we can validate the core principles of the new annotation and how its used, and perhaps find ways to improve it. Then you can follow that up with a couple of more PRs that modify a limited amount of tests each (let's say <10 as an initial idea). @monperrus do you agree with this approach?

monperrus:
yes, it's a good plan. we love baby PRs.

MartinWitt:
I close this as we concluded with smaller PRs.

PR #4697, opened 2022-04-20
review: fix: search in previous cases for variable declarations

### LINKED ISSUE #4696: [Bug] The scope of variable in switch blocs is not properly parsed
**Describe the bug**
When a variable is declared within a switch bloc of code, its scope extends to the whole switch bloc, for instance the declaration of a second variable with the same name will raise a compilation error.

**To Reproduce**
1. Parse the following code with spoon ;
2. Check the resulting model.

Input:
```java
public class A {

  public void function(int value) {

    switch (value) {
      case 1:
        String test;
        test = "first";
        break;
      case 2:
      case 3:
      default:
        test = "not first";
    }

  }

}
```

Processing with Spoon:
```java
spoon = new Launcher();
spoon.addInputResource(path);
spoon.getEnvironment().setComplianceLevel(11);
spoon.getModel()
```

Output:
When parsing the model with a custom scanner, the local variable reference of the assignment ```test = "first"``` is properly defined, that is referencing the preceding declaration ```String test```, but the local variable reference of the assignment ```test = "not first"``` is null instead of referencing the declaration ```String test```.

* OS: Windows 10
* JDK: 11
* Spoon version: 10.0.0

SirYwell:
The naive approach for searching variable declarations only considered siblings and parents, which is not enough for switch statements. There, we need to look through all preceding cases as well.

Fixes #4696

### DIFF
--- a/src/main/java/spoon/reflect/visitor/filter/PotentialVariableDeclarationFunction.java
+++ b/src/main/java/spoon/reflect/visitor/filter/PotentialVariableDeclarationFunction.java
@@ -7,11 +7,15 @@
  */
 package spoon.reflect.visitor.filter;
 
+import spoon.reflect.code.CaseKind;
 import spoon.reflect.code.CtBodyHolder;
+import spoon.reflect.code.CtCase;
 import spoon.reflect.code.CtCatch;
 import spoon.reflect.code.CtCatchVariable;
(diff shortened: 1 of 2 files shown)


monperrus:
thanks a lot @SirYwell

Issue #4698, opened 2022-04-20
[Bug] The role in parent of java.util.ArrayList is wrongly null


dhmoclex:
**Describe the bug**
The role in parent of a java.util.ArrayList is null.
Strangely, sometimes, when printed to the console it is "CONTAINED_TYPE" which is correct, but later in my code, or in debug mode, it is wrongly null.

**To Reproduce**

Any ArrayList in some code may reproduce this bug.

Input:
```java
import java.util.ArrayList;

public class B {

  ArrayList<String> test;

}
```

Processing with Spoon:
```java
spoon = new Launcher();
spoon.addInputResource(path);
spoon.getEnvironment().setComplianceLevel(11);
spoon.getModel()
```

Output:
When parsing the model with a custom scanner, calling ```arrayListTypeReference.getTypeDeclaration().getRoleInParent()``` returns null instead of ```CONTAINED_TYPE```.

The problem is that the CtClassImpl representing the ArrayList is not the same instance of the one present in the CtPackage ```java.util```.

Replacing the element wise comparison in the CtElementImpl.getRoleInParent() may correct the problem. I'm not sure if this fix is valid or may introduce performance issue.

```java
	@Override
	public CtRole getRoleInParent() {
		if (isParentInitialized()) {
			EarlyTerminatingScanner<CtRole> ets = new EarlyTerminatingScanner<CtRole>() {
				@Override
				public void scan(CtRole role, CtElement element) {
					if (element.equals(CtElementImpl.this)) { // <== instead of element == CtElementImpl.this
						setResult(role);
						terminate();
					}
					//do not call super.scan, because we do not want scan children
				}
			};
			getParent().accept(ets);
			return ets.getResult();
		}
		return null;
	}
```

**Operating system, JDK and Spoon version used**

* OS: Windows 10
* JDK: 11
* Spoon version: 10.0.0

dhmoclex:
This pull request break 12 additional tests on my local execution (43 -> 55), but I'm not sure how these tests are related to my fix.

algomaster99:
Hi @dhmoclex !

Thank you for the bug report. The type declaration's (instance of `CtClass`) role should indeed be `CONTAINED_TYPE` as it belongs to the `java.util` package. The fix that you have suggested does make sense. Could you report your investigations on why other tests are failing? You can select any number of them. I think it would help us understand better how roles are computed.

dhmoclex:
Hi @algomaster99 !

[...]

PR #4703, opened 2022-04-21
test: add a test to reproduce the bug #4698

dhmoclex:


slarse:
Hi @dhmoclex,

It seems like this bug was fixed pretty rapidly after you reported it (see #4709). I'm closing this PR as the bug has been fixed. Feel free to reopen it if you feel like your problem was not resolved.

On a side note, as per the contributing guidelines, we typically don't look at PRs unless they are prefixed with `review:`, which is likely why this was glossed over for a couple of weeks.

Thanks for your reporting the bug!

dhmoclex:
Hi @slarse,

Thank you for the info. I was not aware of that.

PR #4704, opened 2022-04-22
wip: test: Fix GitHubIssue annotation not failing surefire tests

I-Al-Istannen:
The previous implementation threw an exception in `testSuccessful` to fail the test. This was dutifully ignored by Junit, as the TestWatcher API explicitly *cannot* impact test results from its callbacks.

This patch switches to the AfterTestExecutionCallback, which does not specify anything in its javadoc and seems to work.

### DIFF
--- a/src/test/java/spoon/test/GitHubIssue.java
+++ b/src/test/java/spoon/test/GitHubIssue.java
@@ -15,6 +15,7 @@ import java.util.HashSet;
 import java.util.Set;
 import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.AfterTestExecutionCallback;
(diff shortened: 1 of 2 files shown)


I-Al-Istannen:
Okay, this exposed two cases where the `GithubIssue` contract was violated and tests *succeeded* where they should not.

The first one was apparently fixed in late 2021 (https://github.com/inria/spoon/issues/4218) and the second one is still open (https://github.com/inria/spoon/issues/3811) but apparently fixed?
If somebody with a bit more experience in the Sniper would have a look whether the test is broken or the `fixed` was just never adjusted, I'd be grateful.

I-Al-Istannen:
Looks like @MartinWitt did make a PR (#4613) after we annoyed him and I just missed it.

PR #4712, opened 2022-04-25
review: meta: Add reproducible build badge to README

I-Al-Istannen:
After #4151, we can now add the badge without lying (and it is green!). As long as we update the buildspec files in reproducible-central, it will even continue to reflect reality.


slarse (approved):
looks green to me, will merge!

PR #4719, opened 2022-05-01
fix: remove unnecessary junit dep from spoon-smpl pom.xml

mkforsb:
This fix stops depclean complaining about unused dependencies when building the spoon-smpl submodule. Running tests still works. I guess junit gets pulled in from the parent `../spoon-pom/pom.xml`?

I also removed a commented-out dependency which I should not have committed in the first place.


monperrus:
thanks @mkforsb

Issue #4716, opened 2022-04-29
[Bug] Creating SmPLMethodCFG causes error if CtExecutable contains for loop


J-Merle:
Trying to create a new SmPLMethodCFG causes an error if the CtExecutable contains a for loop

Here is a minimal code sample to reproduce

Input:
```java
public class FailingProcessor extends AbstractProcessor<CtExecutable<?>> {
    @Override
    public void process(CtExecutable<?> element) {
        SmPLMethodCFG cfg = new SmPLMethodCFG(element);
    }

    private void forLoop() {
        for (int i : Arrays.asList(1, 2, 3, 4)) {
            System.out.println(i);
        }
    }

    public static void main(String[] args) {
        Launcher launcher = new Launcher();
        launcher.addInputResource(new FileSystemFile(new File("./src/main/java/FailingProcessor.java"))); // Change the path to match the current file path
        launcher.addProcessor(new FailingProcessor());
        launcher.run();
    }
}

```

Output:
```java
Exception in thread "main" java.lang.ClassCastException: class spoon.support.reflect.code.CtForEachImpl cannot be cast to class spoon.reflect.code.CtIf (spoon.support.reflect.code.CtForEachImpl and spoon.reflect.code.CtIf are in unnamed module of loader 'app')
	at spoon.smpl.SmPLMethodCFG.<init>(SmPLMethodCFG.java:341)
	at processors.FailingProcessor.process(FailingProcessor.java:15)
	at processors.FailingProcessor.process(FailingProcessor.java:12)
	at spoon.support.visitor.ProcessingVisitor.scan(ProcessingVisitor.java:72)
	at spoon.reflect.visitor.CtScanner.scan(CtScanner.java:184)
	at spoon.reflect.visitor.CtScanner.scan(CtScanner.java:149)
	at spoon.reflect.visitor.CtScanner.visitCtClass(CtScanner.java:357)
	at spoon.support.reflect.declaration.CtClassImpl.accept(CtClassImpl.java:58)
	at spoon.reflect.visitor.CtScanner.scan(CtScanner.java:194)
	at spoon.support.visitor.ProcessingVisitor.scan(ProcessingVisitor.java:68)
	at spoon.reflect.visitor.CtScanner.scan(CtScanner.java:184)
	at spoon.reflect.visitor.CtScanner.scan(CtScanner.java:149)
	at spoon.reflect.visitor.CtScanner.visitCtPackage(CtScanner.java:677)
[...]

PR #4720, opened 2022-05-01
fix: add for-loop elements to UnsupportedElementSwapper and add warning messages (issue #4716)

mkforsb:
This adds `CtFor` and `CtForEach` to the set of elements swapped out by spoon-smpl's `UnsupportedElementSwapper` along with a warning log message for each replaced element.

Should be a slight improvement on the current user experience (see #4716).

### DIFF
--- a/spoon-smpl/src/main/java/spoon/smpl/SmPLMethodCFG.java
+++ b/spoon-smpl/src/main/java/spoon/smpl/SmPLMethodCFG.java
@@ -62,8 +62,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
-import java.util.function.BiFunction;
-import java.util.function.Supplier;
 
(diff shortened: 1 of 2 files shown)


monperrus:
@mkforsb Would we able to add a test case to specify this behavior?

mkforsb:
> @mkforsb Would we able to add a test case to specify this behavior?

Two simple tests added.

MartinWitt:
Could you document the current implementation state inside the subproject? Missing elements are a factory when you choose your tool, and we should be as honest as possible about the limitations.

monperrus:
@MartinWitt the reference doc is Mikael Forsberg's master's thesis ["Design and Implementation of Semantic Patch Support for the Spoon Java Transformation Engine' at KTH Royal Institute of Technology"](http://urn.kb.se/resolve?urn=urn:nbn:se:kth:diva-291226).

Issue #4746, opened 2022-06-11
[Bug] Erroenous AST for type in a Jigsaw module


Auties00:
**Describe the bug**
In short, calling CtElement<P extends CtElement>#getParent(P) using as a parameter CtModule.class always returns the unnamed module even when using types such as String that are definitely in a valid jigsaw module. You can find an example below and various snippets I tried.

**To Reproduce**

Input:
```java
public class TestInput {
    private static final String ABC = String.valueOf("");
}
```

Processing with Spoon:
```java
public class FieldTransformer extends AbstractProcessor<CtField<?>> {
    @Override
    public boolean isToBeProcessed(CtField<?> element) {
        return Processing.shouldProcess(element);
    }

    @Override
    public void process(CtField<?> element) {
        if (element.getAssignment() == null) {
            return;
        }

        element.getParent(CtModule.class); // Unnamed module 
        element.getType().getParent(CtModule.class); // Unnamed module 
        element.getType().getPackage().getParent(CtModule.class); // Unnamed module 
    }
}
```

* OS: Windows 11
* JDK: openjdk version "16" 2021-03-16
* Spoon version: 10.2.0-beta-10

SirYwell:
It would be good to know your Launcher configuration. 

Also, note that the parent of a Ct*Reference is the context it appears in. In your example, the code should return the module the `TestInput` class belongs to for all cases.

Auties00:
> It would be good to know your Launcher configuration.
> 
> Also, note that the parent of a Ct*Reference is the context it appears in. In your example, the code should return the module the `TestInput` class belongs to for all cases.

Here it is:
```java
        var launcher = new Launcher();
        launcher.addInputResource(".../TestInput.class");
        launcher.setSourceOutputDirectory(".../test-classes/");
        launcher.addProcessor(new FieldTransformer());
        launcher.run();
```
(The path is not complete, but I can guarantee that both the input file and output directory are recognized correctly).
Also, even TestInput is in a module, here is the module-info:
```java
module it.auties.example {
    requires spoon.core;
}
```
So I'm almost sure that this is a bug. Also is there any way to hypothetically get the module for the type like I was trying to do?

SirYwell:
You need to set the compliance level to >=9, (e.g. `launcher.getEnvironment().setComplianceLevel(9)`). Otherwise, spoon should throw an exception as soon as it loads a module-info file. This file must be present in the input sources (I'd personally recommend to just pass the source folder as input resource).

[...]

PR #4747, opened 2022-06-12
review: fix: Stop adding duplicated modules when cloning them

I-Al-Istannen:
This issue was discovered while trying to reproduce #4746. Printing a module caused it to be cloned, leading to a duplicate insertion into the global module list. This in turn triggered the "overlapping modules found" sanity check (thankfully!).


slarse (approved):
Thanks @I-Al-Istannen, lgtm.

For future reference for anyone else confused about why pretty printing is calling `createModule()`, it's because we clone before printing and the `CloneVisitor` class creates a new module as part of cloning.

PR #4750, opened 2022-06-16
tests: add a failing test case for MavenLauncher

algomaster99:
[Sorald](https://github.com/SpoonLabs/sorald) has a [test case](https://github.com/SpoonLabs/sorald/blob/master/sorald/src/test/java/sorald/miner/WarningMinerTest.java#L187) that depends upon resolving classpath using `MavenLauncher`. However, it has been failing for a few days. I narrowed down the reason for its failure and it is the inability of the `MavenLauncher` to resolve the classpath correctly. I have reproduced the behaviour there as a failing test case here. Could anyone tell me what could be the reason for the empty classpath returned by the method? It is weird that such a behaviour started happening all of a sudden without any changes to our source code. Also, we use Spoon `10.1.1`.

Please note that I thought I would directly open a PR with a test case rather than an issue as I feel it is much easier to explain my issue with a test case.

### DIFF
--- a/src/test/java/spoon/MavenLauncherTest.java
+++ b/src/test/java/spoon/MavenLauncherTest.java
@@ -219,4 +219,13 @@ public class MavenLauncherTest {
 		assertTrue(mavenHome.exists());
 		assertTrue(mavenHome.isDirectory());
 	}
+
+	@Test
+	public void testClasspathShouldBeResolvedForATestResourceInsideSorald() {
+		// contract: it should resolve the classpath
+		File c = new File("src/test/resources/maven-launcher/classpath-dependent-project");
+		MavenLauncher launcher =
+				new MavenLauncher(c.getAbsolutePath(), MavenLauncher.SOURCE_TYPE.ALL_SOURCE);
+		assertTrue(launcher.getEnvironment().getSourceClasspath().length > 0);
+	}
 }
--- a/src/test/resources/maven-launcher/classpath-dependent-project/pom.xml
+++ b/src/test/resources/maven-launcher/classpath-dependent-project/pom.xml
@@ -1,0 +1,22 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>sorald.test</groupId>
+  <artifactId>sorald.test.project</artifactId>
+  <properties>
+    <maven.compiler.target>8</maven.compiler.target>
+    <maven.compiler.source>8</maven.compiler.source>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+  </properties>
+  <packaging>jar</packaging>
+  <version>1.0-SNAPSHOT</version>
+  <name>sorald.test.project</name>
+  <url>http://maven.apache.org</url>
+  <dependencies>
+    <dependency>
+      <groupId>fr.inria.gforge.spoon.labs</groupId>
+      <artifactId>gumtree-spoon-ast-diff</artifactId>
+      <version>1.24</version>
+    </dependency>
+  </dependencies>
+</project>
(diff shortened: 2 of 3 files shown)


algomaster99:
@MartinWitt could you please help me with this?

slarse:
@algomaster99 I had a look, the MavenLauncher is swallowing the same error that is causing a lot of stuff to fail right now:

```
[ERROR] Failed to execute goal on project sorald.test.project: Could not resolve dependencies for project sorald.test:sorald.test.project:jar:1.0-SNAPSHOT: Failed to collect dependencies at fr.inria.gforge.spoon.labs:gumtree-spoon-ast-diff:jar:1.24 -> fr.inria.gforge.spoon:spoon-core:jar:8.0.0 -> org.eclipse.jdt:org.eclipse.jdt.core:jar:3.16.0 -> org.eclipse.platform:org.eclipse.core.resources:jar:3.12.0 -> org.eclipse.platform:org.eclipse.core.expressions:jar:3.5.100 -> org.eclipse.platform:org.eclipse.core.runtime:jar:3.12.0 -> org.eclipse.platform:org.eclipse.equinox.preferences:jar:3.10.0 -> org.osgi.service:org.osgi.service.prefs:jar:[1.1.0,1.2.0): No versions available for org.osgi.service:org.osgi.service.prefs:jar:[1.1.0,1.2.0) within specified range -> [Help 1]
```

The problem in particular is this:

```
No versions available for org.osgi.service:org.osgi.service.prefs:jar:[1.1.0,1.2.0) within specified range -> [Help 1]
```

Trying to figure out what's going on here.

slarse:
If I go into the directory with the example project and just run `mvn dependency:build-classpath` I get the same error.

There's some kind of strange dependency resolution error here. If I add this dependency as a sole dependency and run `mvn dependency:build-classpath`, it will correctly resolve and fetch the dependency:

```
<dependency>
  <groupId>org.osgi</groupId>
  <artifactId>org.osgi.service.prefs</artifactId>
  <version>1.1.2</version>
</dependency>
```

After that, the test passes.

algomaster99:
@slarse Ah yes, we faced a similar problem in repairnator as well. We resolved it using this [SO post](https://stackoverflow.com/a/72626648). I can provide a fix to it.

slarse:
Oh. Right. `org.osgi.service` is not the same as `org.osgi` :)

slarse:
> @slarse Ah yes, we faced a similar problem in repairnator as well. We resolved it using this [SO post](https://stackoverflow.com/a/72626648). I can provide a fix to it.

I don't really know how you would fix it in Spoon. The problem is not in `MavenLauncher`. If you run `MavenLauncher` on something that does not have that broken transitive dependency, it will work just fine.

We should however not swallow the error like that, that's incredibly confusing.

algomaster99:
@slarse We are ignoring this specific transitive dependency. See https://github.com/SpoonLabs/sorald/pull/837. This ensures that the classpath is built. However, I am not sure how it worked because that transitive dependency is still needed and I did not add it anywhere explicitly.

algomaster99:
> The problem is not in MavenLauncher. If you run MavenLauncher on something that does not have that broken transitive dependency, it will work just fine.

I think we can close this pull request for this reason. :) Thanks for the help.

slarse:
> @slarse We are ignoring this specific transitive dependency. See [SpoonLabs/sorald#837](https://github.com/SpoonLabs/sorald/pull/837). This ensures that the classpath is built. However, I am not sure how it worked because that transitive dependency is still needed and I did not add it anywhere explicitly.

I see. But yeah, the dependency is not from Spoon, I checked the dependency tree and it's nowhere to be seen. Not to mention that Spoon's builds are working just fine :)

The smoke tests are failing all over the place, unfortunately, but I suspect it will "blow over" once projects move past the broken version of that library.

algomaster99:
> I checked the dependency tree and it's nowhere to be seen

For https://github.com/SpoonLabs/sorald/pull/837, it seems that the broken dependency was terminating from gumtree-spoon. However, gumtree-spoon's build is passing as well. Not sure which repository it emanates from. 

> The smoke tests are failing all over the place, unfortunately, but I suspect it will "blow over" once projects move past the broken version of that library.

Right. :+1

slarse:
> For https://github.com/SpoonLabs/sorald/pull/837, it seems that the broken dependency was terminating from gumtree-spoon. However, gumtree-spoon's build is passing as well. Not sure which repository it emanates from.

Yeah, I noted that as well. But perhaps it was that old version of gumtree-spoon (you were on like 1.24) that had a transitive dependency that had the bad dependency. Somehow. It's all a bit of a mess.

PR #4768, opened 2022-06-27
test: Add ModelTest annotation to spoon.test.jdtimportbuilder

I-Al-Istannen:


src/test/java/spoon/test/jdtimportbuilder/ImportBuilderTest.java
```diff
@@ -16,6 +16,7 @@
  */
 package spoon.test.jdtimportbuilder;
 
+import spoon.reflect.factory.Factory;
```
MartinWitt, line 19: The import order is more or less hopeless in test files, but is there any reasoning for the placement of the import?

MartinWitt (approved):

src/test/java/spoon/test/jdtimportbuilder/ImportBuilderTest.java
I-Al-Istannen: There are multiple possible places to insert it here, as the imports seem to be random. I manually wrangled my local diff to prevent IJ from fixing it, but I can move it down.

PR #4770, opened 2022-06-27
fix: Enable creation of packages with repeated simple names in qualified name

### LINKED ISSUE #4764: [Bug] Can't create CtPackage which the parent package name equals child package name
**Describe the bug**
When use PackageFactory to create package. If the parent package name equals the child package name. Only one level of package will be created.

**To Reproduce**

Processing with Spoon:
```java
    @Test
    void testCreatePackage(){
        CtClass<?> dClass = Launcher.parseClass("class A{}");
        CtPackage ctPackage1 = dClass.getFactory().Package().getOrCreate("duplicated.non_duplicated");
        Assertions.assertEquals("duplicated.non_duplicated", ctPackage1.getQualifiedName());//successful
        CtPackage ctPackage2 = dClass.getFactory().Package().getOrCreate("duplicated.duplicated");
        Assertions.assertEquals("duplicated.duplicated", ctPackage2.getQualifiedName());//failed
        CtPackage ctPackage3 = dClass.getFactory().Package().getOrCreate("duplicated.duplicated.duplicated");
        Assertions.assertEquals("duplicated.duplicated.duplicated", ctPackage3.getQualifiedName());//failed
    }
```

Output:
```java
expected: <duplicated.duplicated> but was: <duplicated>
Expected :duplicated.duplicated
Actual   :duplicated
```

* OS: MacOS
* JDK: Java 11
* Spoon version: 10.1.1

slarse:
Fix #4764 

The problem was a check that I can't quite make sense of, see comment in code.

### DIFF
--- a/src/main/java/spoon/support/reflect/declaration/CtPackageImpl.java
+++ b/src/main/java/spoon/support/reflect/declaration/CtPackageImpl.java
@@ -45,13 +45,7 @@ public class CtPackageImpl extends CtNamedElementImpl implements CtPackage {
 
 		@Override
 		public CtPackage put(String simpleName, CtPackage pack) {
-			if (pack == null) {
-				return null;
-			}
-			// they are the same
-			if (CtPackageImpl.this.getQualifiedName().equals(pack.getQualifiedName())) {
-				addAllTypes(pack, CtPackageImpl.this);
-				addAllPackages(pack, CtPackageImpl.this);
+			if (pack == null || pack == CtPackageImpl.this) {
 				return null;
 			}
 
--- a/src/test/java/spoon/reflect/factory/PackageFactoryTest.java
+++ b/src/test/java/spoon/reflect/factory/PackageFactoryTest.java
@@ -1,17 +1,16 @@
 package spoon.reflect.factory;
 
-import org.junit.jupiter.api.Test;
-
 import static org.hamcrest.CoreMatchers.equalTo;
 import static org.hamcrest.CoreMatchers.sameInstance;
 import static org.hamcrest.MatcherAssert.assertThat;
 
 import spoon.Launcher;
 import spoon.reflect.declaration.CtPackage;
+import spoon.test.GitHubIssue;
 
 class PackageFactoryTest {
 
-	@Test
+	@GitHubIssue(issueNumber = 4764, fixed = true)
 	void getOrCreate_returnsNestedPackageStructure_whenQualifiedNameRepeatsSimpleName() {
 		// contract: A qualified name that is simply repetitions of a single simple name results in the expected
 		// nested package structure
--- a/src/test/java/spoon/reflect/factory/PackageFactoryTest.java
+++ b/src/test/java/spoon/reflect/factory/PackageFactoryTest.java
@@ -1,0 +1,32 @@
+package spoon.reflect.factory;
+
+import org.junit.jupiter.api.Test;
+
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.hamcrest.CoreMatchers.sameInstance;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+import spoon.Launcher;
+import spoon.reflect.declaration.CtPackage;
+
+class PackageFactoryTest {
+
+	@Test
+	void getOrCreate_returnsNestedPackageStructure_whenQualifiedNameRepeatsSimpleName() {
+		// contract: A qualified name that is simply repetitions of a single simple name results in the expected
+		// nested package structure
+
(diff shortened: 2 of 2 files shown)


src/main/java/spoon/support/reflect/declaration/CtPackageImpl.java
```diff
@@ -45,13 +45,7 @@ protected CtRole getRole() {
 
 		@Override
 		public CtPackage put(String simpleName, CtPackage pack) {
-			if (pack == null) {
-				return null;
-			}
-			// they are the same
-			if (CtPackageImpl.this.getQualifiedName().equals(pack.getQualifiedName())) {
-				addAllTypes(pack, CtPackageImpl.this);
-				addAllPackages(pack, CtPackageImpl.this);
```
slarse, line 54: The comment implies that the package A to which we are trying to add the package B are the same by comparing their qualified names. This simply does not make sense, as the act of add B as a subpackage to A _changes_ its qualified name to `A.B`.

It's simply impossible for `A` and `B` to end up with the same qualified name after adding `B` to `A` as the qualified name of `A` will always be prefixed to `B`.

I-Al-Istannen: I was quite confused about the check as well when adjusting this class. It sounded like this was meant to merge equal packages. If I didn't miss anything, the two names *can* be the same at the time of the check, if you have two parallel package hierarchies. The linking is only done in the `super` call, so the parent isn't yet replaced.

I think this could happen with the module system (though the overlap is purely due to spoon's creation of synthetic packages). 

I was afraid to touch it back then, but it doesn't seem particularly likely that code depends on this.

slarse: > I was quite confused about the check as well when adjusting this class. It sounded like this was meant to merge equal packages. If I didn't miss anything, the two names _can_ be the same at the time of the check, if you have two parallel package hierarchies.

Absolutely, the names can be the same (e.g. `A = B`, which is precisely the bug reported in #4764). It doesn't matter, however, because the semantics of this method is to add `B` as a subpackage of `A`. A subpackage of `A` cannot have the same qualified name as `A`, as `A`'s qualified name is prepended to all of its supbackages. There simply cannot be a name collision; the action that this method performs makes that impossible.

It just.. doesn't make sense.

I-Al-Istannen: What I was trying to convey is, that you can hit this code path in existing code (e.g.):
```java
		CtModule mod = factory.Module().getOrCreate("second");
		CtPackage pack1 = factory.Package().getOrCreate("test", factory.Module().getUnnamedModule());
		CtPackage pack2 = factory.Package().getOrCreate("test", mod);

		pack1.addPackage(pack2);
```
or if you obtain a second package another way. The existing code would merge this, your code would nest them. This is an observable semantic difference. I just wanted to point out, that this *can happen* and is *not* impossible and why I left that code in there when refactoring the ModelSet.

I do agree though, that the new semantics are likely closer to what the user intended and probably won't do any harm. You just *might* break something, as you change observable behavior.

slarse: For clarity, if you run `Launcher().getFactory().Package().getOrCreate("A.A")`, you get back a package with the qualified name `A` (before removal of the discussed code path).

slarse: Oh no need to apologize, I appreciate your attention to detail :)

src/test/java/spoon/reflect/factory/PackageFactoryTest.java
```diff
@@ -0,0 +1,32 @@
...
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.hamcrest.CoreMatchers.sameInstance;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+import spoon.Launcher;
+import spoon.reflect.declaration.CtPackage;
+
+class PackageFactoryTest {
+
+	@Test
```
MartinWitt, line 14: Could you add either a comment or the GitHub issue linking to the issue of the bug? This could help to understand test cases in the future.

MartinWitt (changes requested):

src/test/java/spoon/reflect/factory/PackageFactoryTest.java
slarse: Good catch, I routinely forget that we have that new annotation. Fixed.

MartinWitt (approved):

MartinWitt:
Thanks @slarse

Issue #4332, opened 2021-12-06
[Bug] Failure on comments inside enum values


plcarmel:
When there are comments inside enum values, Spoon fails to perform a transformation on the type.

```java
package org.example;

enum Example {

    ENGLISH( /* boom */
            "Hello",
            "World"
    ),
    FRENCH(
            "Bonjour",
            "l'univers"
    );

    String hello;
    String world;

    public Example(
            String hello,
            String world
    ) {
        this.hello = hello;
        this.world = world;
    }
}
```

```
Exception in thread "main" spoon.SpoonException: Cannot compare this: [46, 115] with other: ["55", "185"]
        at spoon.support.sniper.internal.ElementSourceFragment.compare(ElementSourceFragment.java:371)
        at spoon.support.sniper.internal.ElementSourceFragment.add(ElementSourceFragment.java:285)
        at spoon.support.sniper.internal.ElementSourceFragment.addChild(ElementSourceFragment.java:317)
        at spoon.support.sniper.internal.ElementSourceFragment.addChild(ElementSourceFragment.java:243)
        at spoon.support.sniper.internal.ElementSourceFragment$1.enter(ElementSourceFragment.java:200)
        at spoon.reflect.visitor.CtScanner.visitCtEnumValue(CtScanner.java:442)
        at spoon.support.reflect.declaration.CtEnumValueImpl.accept(CtEnumValueImpl.java:18)
        at spoon.reflect.visitor.EarlyTerminatingScanner.doScan(EarlyTerminatingScanner.java:145)
        at spoon.reflect.visitor.EarlyTerminatingScanner.scan(EarlyTerminatingScanner.java:121)
        at spoon.reflect.visitor.CtScanner.scan(CtScanner.java:184)
        at spoon.reflect.visitor.EarlyTerminatingScanner.scan(EarlyTerminatingScanner.java:106)
        at spoon.reflect.visitor.EarlyTerminatingScanner.scan(EarlyTerminatingScanner.java:83)
        at spoon.reflect.visitor.CtScanner.visitCtEnum(CtScanner.java:415)
        at spoon.support.reflect.declaration.CtEnumImpl.accept(CtEnumImpl.java:45)
        at spoon.reflect.visitor.EarlyTerminatingScanner.doScan(EarlyTerminatingScanner.java:145)
        at spoon.reflect.visitor.EarlyTerminatingScanner.scan(EarlyTerminatingScanner.java:121)
[...]

PR #4780, opened 2022-07-04
Adjust the source position for annotated enum values

gtoison:
- Unit tests illustrating the issue that the source position is not computed correctly, this leads to a SpoonException with message `"Cannot compare this:...` when trying to print the code with Sniper
- Adjust the calculation of the source start to account for annotations

The fix seems to work but I have very very little experience on the project so this might very well be misguided, there really seem to be a bug though!

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/PositionBuilder.java
+++ b/src/main/java/spoon/support/compiler/jdt/PositionBuilder.java
@@ -421,12 +421,12 @@ public class PositionBuilder {
 			if (allocationExpression.enumConstant != null) {
 				FieldDeclaration fieldDeclaration = allocationExpression.enumConstant;
 
-                Annotation[] annotations = allocationExpression.enumConstant.annotations;
-                if (annotations != null && annotations.length > 0) {
-                    Annotation lastAnnotation = annotations[annotations.length - 1];
-                    sourceStart = findNextNonWhitespace(contents, sourceEnd, lastAnnotation.sourceEnd);
-                }
-                
+				Annotation[] annotations = allocationExpression.enumConstant.annotations;
+				if (annotations != null && annotations.length > 0) {
+					Annotation lastAnnotation = annotations[annotations.length - 1];
+					sourceStart = findNextNonWhitespace(contents, sourceEnd, lastAnnotation.sourceEnd);
+				}
+
 				//1) skip comments
 				sourceStart = findNextNonWhitespace(contents, sourceEnd, sourceStart);
 				//2) move to beginning of enum construction
--- a/src/main/java/spoon/support/compiler/jdt/PositionBuilder.java
+++ b/src/main/java/spoon/support/compiler/jdt/PositionBuilder.java
@@ -420,6 +420,13 @@ public class PositionBuilder {
 			AllocationExpression allocationExpression = (AllocationExpression) node;
 			if (allocationExpression.enumConstant != null) {
 				FieldDeclaration fieldDeclaration = allocationExpression.enumConstant;
+
+                Annotation[] annotations = allocationExpression.enumConstant.annotations;
+                if (annotations != null && annotations.length > 0) {
+                    Annotation lastAnnotation = annotations[annotations.length - 1];
+                    sourceStart = findNextNonWhitespace(contents, sourceEnd, lastAnnotation.sourceEnd);
+                }
+                
 				//1) skip comments
 				sourceStart = findNextNonWhitespace(contents, sourceEnd, sourceStart);
 				//2) move to beginning of enum construction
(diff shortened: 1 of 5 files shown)


src/test/java/spoon/test/position/PositionTest.java
```diff
@@ -1458,4 +1460,16 @@ public void testLambdaParameterPosition1() {
 				)
 		);
 	}
+	
+	@Test
+	@GitHubIssue(issueNumber = 4779, fixed = true)
```
SirYwell, line 1465: I don't fully understand this test case. It looks like it does not hit the modified code in `PositionBuilder`. Am I missing something?
(also, you're currently building the model twice, with the variables `build` and `classContent` being unused)

src/main/java/spoon/support/compiler/jdt/PositionBuilder.java
```diff
@@ -420,6 +420,13 @@ SourcePosition buildPositionCtElement(CtElement e, ASTNode node) {
 			AllocationExpression allocationExpression = (AllocationExpression) node;
 			if (allocationExpression.enumConstant != null) {
 				FieldDeclaration fieldDeclaration = allocationExpression.enumConstant;
+
+				Annotation[] annotations = allocationExpression.enumConstant.annotations;
```
SirYwell, line 424: There are comments below, starting with `1)`. I think it would make sense to add one for annotations too and change the numeration accordingly.

src/test/java/spoon/test/prettyprinter/SniperAnnotatedEnumTest.java
```diff
@@ -0,0 +1,59 @@
...
+public class SniperAnnotatedEnumTest {
+  private static final Path INPUT_PATH = Paths.get("src/test/java/");
+  private static final Path OUTPUT_PATH = Paths.get("target/test-output");
+
+  @BeforeClass
+  public static void setup() throws IOException {
+    FileUtils.deleteDirectory(OUTPUT_PATH.toFile());
+  }
+
+  @Test
```
SirYwell, line 30: `GithubIssue` already serves as test annotation, so `@Test` can be removed.

src/test/java/spoon/test/prettyprinter/SniperAnnotatedEnumTest.java
```diff
@@ -0,0 +1,59 @@
+package spoon.test.prettyprinter;
+
+import org.apache.commons.io.*;
+import org.hamcrest.*;
+import org.junit.*;
+import static org.junit.Assert.*;
```
SirYwell, line 6: You should avoid using JUnit 4. The `assertThat` method should be imported from `org.hamcrest.MatcherAssert` instead.

src/test/java/spoon/test/prettyprinter/SniperAnnotatedEnumTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+    ctClass.addComment(launcher.getFactory().Code().createComment("test", CommentType.BLOCK));
+
+    launcher.process();
+    launcher.prettyprint();
+    // Verify result file exists and is not empty
+    assertThat("Output file for " + path + " should exist", OUTPUT_PATH.resolve(path).toFile().exists(),
+        CoreMatchers.equalTo(true));
+
+    String content = new String(Files.readAllBytes(OUTPUT_PATH.resolve(path)), StandardCharsets.UTF_8);
```
SirYwell, line 56: Very minor nitpick: You can use `Files.readString(...)` here, which also defaults to UTF-8 directly.

SirYwell (changes requested):
Thanks for your work. The fix looks good, but I got a question about one test and a few things that could be changed.

src/test/java/spoon/test/position/PositionTest.java
gtoison: I think I got a bit confused here and I agree that this test is not good
If that's OK I'll remove it

src/main/java/spoon/support/compiler/jdt/PositionBuilder.java
gtoison: Agreed, I have updated the comments as proposed

src/test/java/spoon/test/prettyprinter/SniperAnnotatedEnumTest.java
gtoison: OK, I have removed the redundant test annotation

SirYwell (approved):
I think we need a contract comment on the first line of the method, though I'm not sure if it's needed in combination with the `@GitHubIssue` annotation. I'll let an integrator decide :p.

Otherwise, this looks good to me.

MartinWitt:
Fixes #4332 btw.
Thank you for your PR. LGTM will merge.

PR #4781, opened 2022-07-05
test: upgrade sniper unit tests to JUnit 5

gtoison:
Following the discussion on https://github.com/INRIA/spoon/pull/4780 the new unit test I've added was copied from other existing JUnit 4 tests.
This PR upgrades these similar tests to JUnit 5

### DIFF
--- a/src/test/java/spoon/test/prettyprinter/SniperAssertTest.java
+++ b/src/test/java/spoon/test/prettyprinter/SniperAssertTest.java
@@ -1,27 +1,30 @@
 package spoon.test.prettyprinter;
 
-import org.apache.commons.io.*;
-import org.hamcrest.*;
-import org.junit.*;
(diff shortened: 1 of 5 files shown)


MartinWitt (approved):

MartinWitt:
LGTM, will merge. Thank you @gtoison

MartinWitt:
It seems like some dependency is providing JUnit 4 as transitive dependency. And the really strange package import were not matched by my analysis. I never thought anyone creates static package imports.

gtoison:
I swear I'm innocent!
I looked into it because https://github.com/INRIA/spoon/blob/master/CONTRIBUTING.md says that PR migrating to JUnit 5 are welcome, so maybe this should be amended too?

slarse:
See #4792

PR #4793, opened 2022-07-09
review: doc: Update bug report issue template with new YAML format

slarse:
There's [new templating functionality for issues](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository) that uses YAML to produce much better structured forms for issues. I figured we should try it out. One of the biggest perks is that code formatting is automatic, reporters just need to paste in code, no need for backticks.

This can be considered a draft, feel free to suggest changes.

@MartinWitt @monperrus @SirYwell @I-Al-Istannen


.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -0,0 +1,67 @@
...
+      placeholder: System.out.println("Hello, world!");
+      render: Java
+    validations:
+      required: false
+  - type: textarea
+    id: version
+    attributes:
+      label: Spoon Version
+      description: What version of Spoon are you using? Please note that we only actively support the latest major release. If you are using an older major version, your issue may be closed as non-actionable.
+      placeholder: 10.1.0
```
I-Al-Istannen, line 59: We should add updating this template to the release workflow then.

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -0,0 +1,67 @@
...
+      label: Expected output
+      description: If you know what kind of output you expect to get, please provide it here. This will be automatically formatted, so no need for backticks.
+      placeholder: System.out.println("Hello, world!");
+      render: Java
+    validations:
+      required: false
+  - type: textarea
+    id: version
+    attributes:
+      label: Spoon Version
```
I-Al-Istannen, line 57: Maybe also provide a "java compliance level" or something field? Not sure if that is helpful enough, we don't want to burden the user too much.

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -0,0 +1,67 @@
...
+      label: Source code for your Spoon processing
+      description: Provide your processing code that uses Spoon and causes the bug to appear. This will be automatically formatted, so no need for backticks.
+      placeholder: System.out.println("Hello, world!");
+      render: Java
+    validations:
+      required: false
+  - type: textarea
+    id: output
+    attributes:
+      label: Output
```
I-Al-Istannen, line 39: Maybe make the distinction between expected and actual a bit clearer? `Your`/`Actual`/`Observed`/... I don't know.
```suggestion
      label: Actual output
```

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -0,0 +1,67 @@
...
+      placeholder: System.out.println("Hello, world!");
+      render: Java
+    validations:
+      required: false
+  - type: textarea
+    id: spoon-code
+    attributes:
+      label: Source code for your Spoon processing
+      description: Provide your processing code that uses Spoon and causes the bug to appear. This will be automatically formatted, so no need for backticks.
+      placeholder: System.out.println("Hello, world!");
```
I-Al-Istannen, line 32: Maybe we should be a bit more suggestive and actually have a spoon snippet here (of something we'd like them to include)?

```suggestion
      placeholder: Launcher launcher = new Launcher();
```

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -0,0 +1,67 @@
...
+      label: Describe the bug
+      description: Provide a general description of the bug, in plain text. Use the dedicated fields below for code snippets.
+      placeholder: Description of the bug
+    validations:
+      required: true
+  - type: textarea
+    id: java-input
+    attributes:
+      label: Source code you are trying to analyze/transform
+      description: Provide a minimal example of source code you are trying to analyze/transform that causes the bug to appear.
```
I-Al-Istannen, line 22: Is this also automatically formatted? (I think so, but do we want to explicitly specify it here as well?)

I-Al-Istannen (commented):
Thanks! I was looking at that every few weeks whenever an issue was opened :P I didn't realize they are out of beta yet.

I have left a few things that came to mind while skimming the new template.

.github/ISSUE_TEMPLATE/bug_report.yaml
slarse: I'm a little bit hesitant here, it's easily forgotten. I _wanted_ to put a dropdown here, but it would be really bad the if we forgot to update it with relevant versions. Even beta versions, which are automatically released.

To cut ourselves loose from a potentially manual step here, perhaps the placeholder should be something like `X.Y.Z` instead?

.github/ISSUE_TEMPLATE/bug_report.yaml
slarse: Agreed, that's a good distinction.

.github/ISSUE_TEMPLATE/bug_report.yaml
slarse: Yes, it's the "render" attribute. I've just forgotten to add the little text snippet about the formatting.

slarse:
```suggestion
      description: Provide a minimal example of source code you are trying to analyze/transform that causes the bug to appear. This will be automatically formatted, so no need for backticks.
```

.github/ISSUE_TEMPLATE/bug_report.yaml
slarse: It's not a required input for using Spoon, so I think we shouldn't include that. I wanted to put in something like "noclasspath", but that's also not required input so I left it out. Like you said, we don't want to make it too much of a hassle to report a problem.

MartinWitt:
How about adding the used JVM version string as needed information?

slarse:
> How about adding the used JVM version string as needed information? Otherwise, looks wonderful.

Of course that should be in there as well. It just somehow slipped my mind. Will add.

.github/ISSUE_TEMPLATE/bug_report.yaml
I-Al-Istannen: `X.Y.Z` sounds fine, yea. It's probably understandable enough and decidedly less out of date :)

slarse:
@MartinWitt Ready for re-review!

MartinWitt (approved):

MartinWitt:
LGTM, let's see how the users like it? Always highlighted source code seems astonishing.

slarse:
>  Edit: I have merged it now because we can easily change this anytime we want, if we see there is something missing.

Yep. I had a quick peek and it looks pretty reasonable. Perhaps a bit large with all of the different "input, spoon code, output, expected output" stuff.

PR #4794, opened 2022-07-09
doc: Clarify stance on supported versions

slarse:
We still get a lot of issues using JDK 8 and old versions of Spoon. I think we should clarify that we don't intend to support JDK 8 or older versions of Spoon that uses it.

The new bug report template I've added in #4793 also takes some steps in this direction.


MartinWitt (approved):

MartinWitt:
Looks good for me. This should clarify it.

PR #4795, opened 2022-07-10
review: fix: clearer message when cannot compare fragments

gtoison:
The error message `Cannot compare this: [46, 115] with other: ["55", "185"]` is slightly unclear because of the double quotes for the other element. This might suggest that the types are different while the issue is really with the start and end positions.
This PR removes the unnecessary double quotes.


MartinWitt:
Good catch, will merge.

MartinWitt (approved):

PR #784, opened 2016-08-11
Refactors DefaultJavaPrettyPrinter and JDTTreeBuilder 🎉

GerardPaligot:

PR #4809, opened 2022-07-20
fix: minimise bracket printing in case of string operands of binary operator

### LINKED ISSUE #4807: Extra Parenthesis after Spoon getAssignment()
Hi Team,

I have an issue where I am getting extra parenthesis after getAssignment().

Below is the replication of the problem -

**My code during build model**

```
import spoon.Launcher;
import spoon.compiler.Environment;
import spoon.reflect.CtModel;
import spoon.support.sniper.SniperJavaPrettyPrinter;

launcher.addInputResource(dirPath);
final Environment e = launcher.getEnvironment();
e.setPrettyPrinterCreator(() -> new SniperJavaPrettyPrinter(e));
launcher.buildModel(); 
```

**Input which I am giving**

ResultSet rs1 = stmt.executeQuery(QUERY1 + getTableName() + QUERY2 + "id=" + javaFileName.getEmployeeId() + QUERY3);

And trying to extract what is inside executeQuery so the 
**Expected output -** QUERY1 + getTableName() + QUERY2 + "id=" + javaFileName.getEmployeeId() + QUERY3
**Output I am getting -** ((((QUERY1 + getTableName()) + QUERY2) + "id=") + javaFileName.getEmployeeId()) + QUERY3

Please suggest how can I get rid of the extra parenthesis (Excluding parenthesis of methods like getTableName())...!!!

algomaster99:
Fixes #4807

I have tried to reproduce #4807 with a related test case.
```
Expected: java.lang.System.out.println("1" + "2" + "3" + "4")
Actual: java.lang.System.out.println("1" + ("2" + "3" + "4"))
```

I have noticed that the test pass when I change each operand in the above expression to an integer, thanks to #3823.

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/ParentExiter.java
+++ b/src/main/java/spoon/support/compiler/jdt/ParentExiter.java
@@ -445,10 +445,11 @@ public class ParentExiter extends CtInheritanceScanner {
 			} else if (jdtTreeBuilder.getContextBuilder().stack.peek().node instanceof StringLiteralConcatenation) {
 				CtBinaryOperator<?> op = operator.getFactory().Core().createBinaryOperator();
 				op.setKind(BinaryOperatorKind.PLUS);
-				op.setLeftHandOperand(operator.getRightHandOperand());
-				op.setRightHandOperand((CtExpression<?>) child);
+				op.setLeftHandOperand(operator.getLeftHandOperand());
+				op.setRightHandOperand(operator.getRightHandOperand());
 				op.setType((CtTypeReference) operator.getFactory().Type().STRING.clone());
-				operator.setRightHandOperand(op);
+				operator.setLeftHandOperand(op);
+				operator.setRightHandOperand(((CtExpression<?>) child));
 				int[] lineSeparatorPositions = jdtTreeBuilder.getContextBuilder().getCompilationUnitLineSeparatorPositions();
 				SourcePosition leftPosition = op.getLeftHandOperand().getPosition();
 				SourcePosition rightPosition = op.getRightHandOperand().getPosition();
--- a/src/test/java/spoon/test/comment/CommentTest.java
+++ b/src/test/java/spoon/test/comment/CommentTest.java
@@ -465,9 +465,9 @@ public class CommentTest {
 
 
 		CtLocalVariable ctLocalVariableString = m1.getBody().getStatement(12);
-		assertEquals(createFakeComment(f, "comment multi line string"), ((CtBinaryOperator) ((CtBinaryOperator) ctLocalVariableString.getDefaultExpression()).getRightHandOperand()).getLeftHandOperand().getComments().get(0));
-		assertEquals("\"\" + (\"\"// comment multi line string" + newLine
-				+ " + \"\")", ctLocalVariableString.getDefaultExpression().toString());
+		assertEquals(createFakeComment(f, "comment multi line string"), (((CtBinaryOperator) ctLocalVariableString.getDefaultExpression()).getLeftHandOperand()).getComments().get(0));
+		assertEquals("(\"\" + \"\")// comment multi line string" + newLine
+				+ " + \"\"", ctLocalVariableString.getDefaultExpression().toString());
 
 		ctLocalVariable1 = m1.getBody().getStatement(13);
 		assertEquals("boolean c = (i == 1) ? // comment before then boolean CtConditional" + newLine
--- a/src/test/java/spoon/reflect/ast/AstCheckerTest.java
+++ b/src/test/java/spoon/reflect/ast/AstCheckerTest.java
@@ -22,13 +22,14 @@ import spoon.reflect.code.CtStatement;
 import spoon.reflect.declaration.CtClass;
 import spoon.reflect.reference.CtExecutableReference;
 import spoon.support.modelobs.FineModelChangeListener;
+import spoon.reflect.code.CtBinaryOperator;
 import spoon.reflect.code.CtBlock;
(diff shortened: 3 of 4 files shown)


algomaster99:
I think I have figured out why it is happening. To understand it, let's consider a working case first - "1 + 2 + 3 + 4".

### 1 + 2 + 3 + 4

The above expression is modelled as:
```
                                +
                    +                    4
           +               3
     1           2
```
So everytime [nestedOperatorRequiresRoundBrackets](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/reflect/visitor/RoundBracketAnalyzer.java#L60) is called for each nested operator, the [`associativity`](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/reflect/visitor/RoundBracketAnalyzer.java#L65) and [`positionInParent`](positionInParent) are equal and thus no brackets are printed (the operator precedence is obviously equal).

### "1" + "2" + "3" + "4"

This is the failing test case. We expect `"1" + "2" + "3" + "4"` to be printed, however, `"1" + ("2" + "3" + "4")` is printed. I feel this has to do with how spoon models this expression.

```
                                 +
                    1                     +
                                  +               4
                           2           3
```

For the first nested operator, besides `"4"` in this case, there is a difference in `associativity` and `positionInParent` and so the brackets are put to enclose `"2" + "3" + "4"`.

The resolution would be to see why there is a difference in AST modelling of the two expressions and probably fix it if it turns out to be a bug. @slarse do you also agree that the difference in behaviour arises from how Spoon is modelling nested binary operators?

src/test/java/spoon/test/comment/CommentTest.java
```diff
@@ -465,9 +465,9 @@ public void testInLineComment() {
 
 
 		CtLocalVariable ctLocalVariableString = m1.getBody().getStatement(12);
-		assertEquals(createFakeComment(f, "comment multi line string"), ((CtBinaryOperator) ((CtBinaryOperator) ctLocalVariableString.getDefaultExpression()).getRightHandOperand()).getLeftHandOperand().getComments().get(0));
-		assertEquals("\"\" + (\"\"// comment multi line string" + newLine
-				+ " + \"\")", ctLocalVariableString.getDefaultExpression().toString());
+		assertEquals(createFakeComment(f, "comment multi line string"), (((CtBinaryOperator) ctLocalVariableString.getDefaultExpression()).getLeftHandOperand()).getComments().get(0));
+		assertEquals("(\"\" + \"\")// comment multi line string" + newLine
```
algomaster99, line 469: The `minimizeRoundBrackets` was `false` for this test case. Hence, I rearranged the brackets.

src/main/java/spoon/support/compiler/jdt/ParentExiter.java
```diff
@@ -445,10 +445,11 @@ public <T> void visitCtBinaryOperator(CtBinaryOperator<T> operator) {
...
 				CtBinaryOperator<?> op = operator.getFactory().Core().createBinaryOperator();
 				op.setKind(BinaryOperatorKind.PLUS);
-				op.setLeftHandOperand(operator.getRightHandOperand());
-				op.setRightHandOperand((CtExpression<?>) child);
+				op.setLeftHandOperand(operator.getLeftHandOperand());
+				op.setRightHandOperand(operator.getRightHandOperand());
 				op.setType((CtTypeReference) operator.getFactory().Type().STRING.clone());
-				operator.setRightHandOperand(op);
+				operator.setLeftHandOperand(op);
+				operator.setRightHandOperand(((CtExpression<?>) child));
```
algomaster99, line 452: I have done exactly what I have explained [here](https://github.com/INRIA/spoon/pull/4809#issuecomment-1189865743). I ensured that the left operand takes the nested operator whenever possible.

algomaster99:
I have not put the exact test case as listed in the issue, but I feel the one I added in the PR captures the essence and should fix it. @slarse @MartinWitt @I-Al-Istannen @SirYwell please review it when you get time.

P.S. Is there a team for maintainers so that I can collectively tag them?

I-Al-Istannen:
I don't see any justification for the reordering in the introducing commit, which apparently was meant for switch case statements: https://github.com/INRIA/spoon/pull/784/commits/25156e14862fdccb7cb2f1c2510db3d4131e5da6.

Based on your example the current behaviour seems questionable. It should be *semantically correct*, as `+` on string literals is well-defined and associative, but the AST differs from the underlying code. I am not sure if this can be classified as a *bug* necessarily, but I'd welcome a change.

slarse:
@algomaster99 

> P.S. Is there a team for maintainers so that I can collectively tag them?

No, there isn't. There isn't really a strong use case for it. We go over PRs for review if they are prefixed with `review:`, and usually maintainers will ping each other if they know one or the other is more well equipped for any particular review.

@I-Al-Istannen 

> I am not sure if this can be classified as a bug necessarily, but I'd welcome a change.

Well, if it builds a non-equivalent AST (even if semantically equivalent), I'd say it's a bug. It makes string concatenation look right associative, which it technically isn't. Although it also technically doesn't matter because there's only one operator involved in concatenation of literal strings.

slarse (changes requested):
@algomaster99 Thanks for the PR. Could you add a minimal test case demonstrating the problem on the AST level? That is to say, given a literal string concatenation `"a" + "b" + "c"`, we expect the AST `binop(binop("a", "b", +), "c", +)`.

slarse (approved):
Thanks @algomaster99 , lgtm!

I'm adjusting the commit message from the PR title as this isn't really about pretty-printing (the printer was doing the right thing by the AST all along).

PR #4828, opened 2022-08-05
review: doc: Remove return type in sample getSignature() output

### LINKED ISSUE #4821: [doc bug]: how to determine two methods are equal (all upto sigature) in the sense of what javac believe?
I saw this:

https://spoon.gforge.inria.fr/mvnsites/spoon-core/apidocs/spoon/reflect/declaration/CtExecutable.html#getSignature()

it returns a string, can I compare the returned string to determine if two methods are equal (all upto sigature) in the sense of what javac believe?

BTW, according to the doc:
"""
 The signature is composed of the method name and the parameter types, all fully-qualified, eg "int foo(java.lang.String)".
"""

But I tried some simple code, the method's return type is not part of the returned string, e.g.:

```
equals(java.lang.Object)
```

Is this a bug?

I-Al-Istannen:
Closes #4821

### DIFF
--- a/src/main/java/spoon/reflect/declaration/CtExecutable.java
+++ b/src/main/java/spoon/reflect/declaration/CtExecutable.java
@@ -7,13 +7,13 @@
  */
 package spoon.reflect.declaration;
 
+import spoon.reflect.annotations.PropertyGetter;
+import spoon.reflect.annotations.PropertySetter;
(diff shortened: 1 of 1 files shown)


slarse (approved):

slarse:
Thanks @I-Al-Istannen

Issue #4817, opened 2022-08-01
How to pass environment Variables to MavenLauncher


AryanMansani:
I have a project, where I am fetching code from a rep and then building a CtModel from it that repo.

Now in some project there are some environment variables in dependencies like - 

```
<dependency>
<groupID>common-annotation</groupID>
<artifactID>common-annotation</artifactID>
<version>${revision}<version>
</dependency>
```

Here, revision is that variable, now when building a model, spoon gives an error.

Is there a way I can pass this variable to Maven Launcher or any other work around.

I-Al-Istannen:
Probably not the best solution, but if they are present as environment variables, the spawned `mvn` process will inherit them. This means that `revision=foo java YourProgram` should work.

Setting that from inside your program is not really portable though. The maven launcher would need to expose an interface and use the `addShellEnvironment` method to pass that along to the invocation request.

I-Al-Istannen:
Ah, I just thought of another slightly more realistic workaround:
You could edit the pom to add them to the `properties` section. Either using an XML library or some well-placed regular expressions :)

AryanMansani:
That's a really good idea, but for it, we need to beforehand identify either manually or by XML parser if this variable is not present in the properties, so for a single module it sounds okay but for a multi-modular project it sounds like a lot of work, is there a way we can automatically identify which variables are not present using any library?

I-Al-Istannen:
You need to know which environment variables need to be set in any case. I guess we could also add that to the MavenLauncher API, setting environment variables sounds useful enough.

See #4830.

I-Al-Istannen:
The new API you want to look at is `setEnvironmentVariable` in the `MavenLauncher`. Do let us know if this does not solve your issue @AryanMansani :)

PR #4830, opened 2022-08-07
feat: Allow passing environment variables in maven launcher invocations

I-Al-Istannen:
Some maven builds might need environment variables to complete successfully and we should offer an API for this.

As the methods I overloaded are part of the public API, I enclosed the environment variables in an opaque settings object. This should allow us to add more options later without another overload layer per feature. I am not sure if this is a good idea, but it might ease the burden a bit.

Closes: #4817

### DIFF
--- a/src/main/java/spoon/MavenLauncher.java
+++ b/src/main/java/spoon/MavenLauncher.java
@@ -177,8 +177,8 @@ public class MavenLauncher extends Launcher {
 	 * @param value its value
 	 * @see #rebuildClasspath() #rebuildClasspath() for changes to take effect
 	 */
-	public void addEnvironmentVariable(String key, String value) {
-		mavenOptions.addEnvironmentVariable(key, value);
+	public void setEnvironmentVariable(String key, String value) {
+		mavenOptions.setEnvironmentVariable(key, value);
 	}
 
 	private void init(String mavenProject, String[] classpath, Pattern profileFilter) {
--- a/src/main/java/spoon/MavenLauncher.java
+++ b/src/main/java/spoon/MavenLauncher.java
@@ -20,11 +20,13 @@ import java.util.regex.Pattern;
  * Create a Spoon launcher from a maven pom file
  */
 public class MavenLauncher extends Launcher {
+	private static final Logger LOGGER  = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+
 	private String mvnHome;
 	private SOURCE_TYPE sourceType;
 	private SpoonPom model;
 	private boolean forceRefresh = false;
-	private static final Logger LOGGER  = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+	private final SpoonPom.MavenOptions mavenOptions = SpoonPom.MavenOptions.empty();
 
 	/**
 	 * @return SpoonPom corresponding to the pom file used by the launcher
@@ -166,6 +168,19 @@ public class MavenLauncher extends Launcher {
 		init(mavenProject, classpath, profileFilter);
 	}
 
+	/**
+	 * Adds an environment variable to the maven invocation.
+	 * <p>
+	 * <br><strong>Note that you need to call {@link #rebuildClasspath()} after calling this method for changes to take effect.</strong>
+	 *
+	 * @param key the name of the environment variable
+	 * @param value its value
+	 * @see #rebuildClasspath() #rebuildClasspath() for changes to take effect
+	 */
+	public void addEnvironmentVariable(String key, String value) {
+		mavenOptions.addEnvironmentVariable(key, value);
+	}
+
 	private void init(String mavenProject, String[] classpath, Pattern profileFilter) {
 		File mavenProjectFile = new File(mavenProject);
 		if (!mavenProjectFile.exists()) {
@@ -195,7 +210,7 @@ public class MavenLauncher extends Launcher {
 		}
 
 		if (classpath == null) {
-			classpath = model.buildClassPath(mvnHome, sourceType, LOGGER, forceRefresh);
+			classpath = model.buildClassPath(mvnHome, sourceType, LOGGER, forceRefresh, mavenOptions);
 			LOGGER.info("Running in FULLCLASSPATH mode. Source folders and dependencies are inferred from the pom.xml file (doc: http://spoon.gforge.inria.fr/launcher.html).");
 		} else {
 			LOGGER.info("Running in FULLCLASSPATH mode. Classpath is manually set (doc: http://spoon.gforge.inria.fr/launcher.html).");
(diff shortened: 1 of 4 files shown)


I-Al-Istannen:
Okay, I am stumped. I am not sure why the maven invocation fails, but it only seems to fail on windows. I don't have a usable windows install right now, so I'd be grateful if somebody else could help here :)

src/test/java/spoon/MavenLauncherTest.java
```diff
@@ -44,9 +48,23 @@
...
+		try (Stream<Path> paths = Files.walk(Path.of("src/test/resources/maven-launcher"))) {
+			List<Path> filesToDelete = paths
+				.filter(it -> it.getFileName().toString().equals("spoon.classpath.tmp"))
+				.collect(Collectors.toList());
+
+			for (Path path : filesToDelete) {
+				Files.deleteIfExists(path);
+			}
+		}
+	}
```
slarse, line 63: I think it would be better if the tests were executed in a temporary directory instead (e.g. using `@TempDir` from JUnit5).

src/test/java/spoon/MavenLauncherTest.java
```diff
@@ -219,4 +237,22 @@ public void testGuessMavenHome() {
...
+	@Test
+	void mavenLauncherPassesEnvironmentVariables() {
+		// 10.1.0
+		MavenLauncher launcher = new MavenLauncher("./src/test/resources/maven-launcher/with-environment-variables", MavenLauncher.SOURCE_TYPE.ALL_SOURCE);
+		launcher.addEnvironmentVariable("spoonVersion", "10.1.0");
+		launcher.rebuildClasspath();
+
+		boolean containsSpoonDependency = Arrays
+			.stream(launcher.getEnvironment().getSourceClasspath())
+			.anyMatch(it -> it.matches(".*fr.inria.gforge.spoon.spoon-core.10.1.0.spoon-core-10.1.0.jar"));
```
slarse, line 250: Are you entirely certain the classpath will _end_ with this dependency? Perhaps the test fails on Windows because it's somewhere else?

```suggestion
			.anyMatch(it -> it.matches(".*fr.inria.gforge.spoon.spoon-core.10.1.0.spoon-core-10.1.0.jar.*"));
```

src/main/java/spoon/support/compiler/SpoonPom.java
```diff
@@ -654,4 +688,39 @@ public String getPath() {
 	public File toFile() {
 		return pomFile;
 	}
+
+	/**
+	 * Additional options for maven invocations.
+	 */
+	@Internal
+	public static class MavenOptions {
```
slarse, line 696: What's the reasoning for making this a nested class in `SpoonPom`? Barring the fact that it's used _by_ `SpoonPom`, it doesn't seem intrinsically related.

slarse (commented):
A few observations.

I can try on Windows when I get home this weekend, I have one sitting around.

src/main/java/spoon/support/compiler/SpoonPom.java
I-Al-Istannen: I wanted to keep this an internal implementation detail and the directory SpoonPom resides in didn't feel adequate (it doesn't have anything to do with compilers). Additionally it is quite closely related so I hoped I could get away with it.

If you feel that's too much of a stretch I can extract it with a more verbose name :)

src/test/java/spoon/MavenLauncherTest.java
slarse: This is something of a systematic error we're doing (i.e. just running in the local directory), but perhaps this PR would be a good driving force to actually start running tests appropriately isolated.

src/test/java/spoon/MavenLauncherTest.java
I-Al-Istannen: That's a good spot, but:
It is an array with one entry per dependency, so it *should* be fine. The [classpath printed in the CI](https://github.com/INRIA/spoon/runs/7711796963?check_suite_focus=true#step:10:579) is an empty array, so nothing is found. I could observe that happening locally if the maven invocation failed and did not produce any output, but the invocation request does not seem to throw an exception in such cases.

src/main/java/spoon/support/compiler/SpoonPom.java
slarse: > the directory SpoonPom resides in didn't feel adequate

That we can agree on. It would have to go into an internal subpackage, and that'd put it far away from it's currently singular use. I mostly wanted to see if you'd thought it through, which you have, so let's leave it where it is :)

src/test/java/spoon/MavenLauncherTest.java
I-Al-Istannen: Yea, that was a bit lazy… In exchange I have converted all launcher tests (that aren't disabled) to use temporary folders. The diff is a bit more annoying now but I hope still readable enough.

src/test/java/spoon/MavenLauncherTest.java
slarse: It probably failed then. We've had trouble with Maven on Windows before. I'll have a peek once I get home then!

I-Al-Istannen:
@slarse Okay, I think I solved it. The docs say that only `env.xx` is valid, but without `env.` it works as well in practice. I have no idea why, but I either missed it or that is an undocumented feature.

The problem is that maven normalizes all env variables to upper case – but apparently not the occurences in the pom. So `env.foo` works on linux, but on windows it is matched against `FOO`, which fails. This is in the name of "reliability", but after wasting a few hours I disagree quite strongly with that…

slarse (approved):
I was unable to get around to testing this over the weekend, other things came up. But I see you worked it out, and LGTM.

src/main/java/spoon/MavenLauncher.java
```diff
@@ -166,6 +168,19 @@ public MavenLauncher(String mavenProject, SOURCE_TYPE sourceType, String[] class
...
+	 * <p>
+	 * <br><strong>Note that you need to call {@link #rebuildClasspath()} after calling this method for changes to take effect.</strong>
+	 *
+	 * @param key the name of the environment variable
+	 * @param value its value
+	 * @see #rebuildClasspath() #rebuildClasspath() for changes to take effect
+	 */
+	public void addEnvironmentVariable(String key, String value) {
+		mavenOptions.addEnvironmentVariable(key, value);
+	}
```
slarse, line 182: Oh, actually, I see one more thing that irks me a little bit. Could we rename this to `set`? To me, `add` implies no deduplication, but that makes no sense for environment variables.

```suggestion
	public void setEnvironmentVariable(String key, String value) {
		mavenOptions.setEnvironmentVariable(key, value);
	}
```

slarse (changes requested):
Actually, one last thing ... :thinking:

src/main/java/spoon/MavenLauncher.java
I-Al-Istannen: Yea, I just copied the name of the maven invoker method, but you are right. I adjusted it.

slarse (approved):
Thanks @I-Al-Istannen

PR #4832, opened 2022-08-08
fix: Maintain insertion of types in a package

### LINKED ISSUE #4827: [Bug]: reverse order for classes in source file that declares multiple classes
### Describe the bug

The order of AST nodes does not seem right when a class declares multiple classes. I tried to correct the order in my CtScanner but was not able to do it.

I tried all combinations and enabled/disabled these methods:
```java
    @Override
    public void visitCtCompilationUnit(CtCompilationUnit compilationUnit) {
        var declared = compilationUnit.getDeclaredTypes();
        Collections.reverse(declared);
        compilationUnit.setDeclaredTypes(declared);
        super.visitCtCompilationUnit(compilationUnit);
    }
    @Override
    public void visitCtPackage(CtPackage ctPackage) {
        var declared = ctPackage.getTypes().stream().toList();
        Collections.reverse(declared);
        ctPackage.setTypes(new HashSet<>(declared));
        super.visitCtPackage(ctPackage);
    }
```

```Java
// Main.java
class Main { }
class Shape { }
class Rectangle extends Shape { }
```

_No response_

### Actual output

```Java
// AST:
RootPackage
  Rectangle
  Shape
  Main
```

### Expected output

```Java
// AST:
RootPackage
  Main
  Shape
  Rectangle
```

### Spoon Version

10.1.0

### JVM Version

11

Windows

slarse:
Fix #4827 

Adjusts the `ElementNameMap` to maintain an insertion order of it's entries, which in turn allows us to maintain insertion order of contained types. This _will_ result in a performance hit on `entrySet()` due to the sorting, but I don't think it should be all that bad.

If we want to improve performance further we'll need to implement a concurrent version of a `LinkedHashMap`, or something similar.

@I-Al-Istannen Thoughts?


src/main/java/spoon/support/util/internal/ElementNameMap.java
```diff
@@ -33,22 +40,42 @@
...
 
-	private static final long serialVersionUID = 1L;
+	private static final long serialVersionUID = 2L;
 
 	// This can not be a "Map" as this class is Serializable and therefore Sorald wants this field
 	// to be serializable as well (Rule 1948).
 	// It doesn't seem smart enough to realize it is final and only assigned to a Serializable Map
 	// in the constructor.
-	private final ConcurrentSkipListMap<String, T> map;
+	private final ConcurrentHashMap<String, InsertOrderWrapper<T>> map;
```
slarse, line 56: The hash map generally performs better. The ordering of the skip list map wasn't what we wanted.

src/test/java/spoon/support/util/internal/ElementNameMapTest.java
```diff
@@ -0,0 +1,56 @@
...
+                .collect(Collectors.toList());
+
+        var map = new SimpleElementNameMap(factory.createClass());
+        entries.forEach(entry -> map.put(entry.getKey(), entry.getValue()));
+
+        var fetchedEntries = new ArrayList<>(map.entrySet());
+
+        assertEquals(fetchedEntries, entries);
+    }
+}
```
I-Al-Istannen, line 56: I'd probably add a trailing newline

src/test/resources/noclasspath/MultipleClasses.java
```diff
@@ -0,0 +1,7 @@
+class A {}
+
+class D {}
+
+class C {}
+
+class B {}
```
I-Al-Istannen, line 7: And here maybe as well

src/main/java/spoon/support/util/internal/ElementNameMap.java
```diff
@@ -124,15 +166,21 @@ public boolean containsKey(Object key) {
...
 	@Override
 	public Set<Entry<String, T>> entrySet() {
-		return map.entrySet();
+		return entriesByInsertionOrder().collect(Collectors.toCollection(LinkedHashSet::new));
+	}
+
+	private Stream<Entry<String, T>> entriesByInsertionOrder() {
+		return map.entrySet().stream()
+				.sorted(Comparator.comparing(entry -> entry.getValue().insertionNumber))
+				.map(entry -> new AbstractMap.SimpleEntry<>(entry.getKey(), entry.getValue().value));
```
I-Al-Istannen, line 183:
```suggestion
				.map(entry -> Map.entry(entry.getKey(), entry.getValue().value));
```

Unless you need serializability or adjust its value.

I-Al-Istannen (approved):
Looks good to me, maybe a single nitpick

src/main/java/spoon/support/util/internal/ElementNameMap.java
slarse: I didn't know that utility method existed. My brain runs on Java 8.

MartinWitt:
Thanks, @slarse

PR #4833, opened 2022-08-09
Create RoundBracketPrinter

bhatnagar444:


algomaster99:
@bhatnagar444 This is not how you reproduce a bug via PR. Notice that my [PR](https://github.com/INRIA/spoon/pull/4750/files#diff-58bbf79c2ce4fdad0caa639f676efe4e47e9cdc44175a037e3dad3f86fa649bc) has a valid JUnit test case and it _asserts_ some conditions. You also need to make sure that your test run and fails our CI build. If you use IDEs like Eclipse or IntelliJ, they have an option for running test from GUI.

If you do not have the time, at least try to be very descriptive about it. Elaborately speaking, mention what do you expect and what you actually got. I always say this because a lot of people are working on Spoon so they can also help you out.

For now, please write the JUnit test properly and if Spoon has a bug, it will fail our CI. I will take it from there. If you have not worked with JUnit testing, I feel it is good to learn that and I promise it will take a matter of minutes.

MartinWitt:
Closing this, as this does not even have correct source files and information is missing.

PR #4836, opened 2022-08-11
fix: attach comment to `CtLiteral` nodes

algomaster99:
I was trying to reproduce https://github.com/INRIA/spoon/issues/4727 and I came across another possible bug in the Spoon model builder. It seems that when comments  are added to `CtCaseExpression`s, they are not parsed and stored in the model. Is this a bug or does Spoon has some configuration that I need to enable to include comments in the model? Note that I already set one in my knowledge - `launcher.getEnvironment().setCommentEnabled(true);`.

### DIFF
--- a/src/test/java/spoon/reflect/ast/AstCheckerTest.java
+++ b/src/test/java/spoon/reflect/ast/AstCheckerTest.java
@@ -44,6 +44,7 @@ import spoon.support.UnsettableProperty;
 import spoon.support.comparator.CtLineElementComparator;
 import spoon.support.util.internal.ElementNameMap;
 import spoon.support.util.ModelList;
+import spoon.testing.utils.ModelTest;
 
 import java.util.Arrays;
 import java.util.Collection;
@@ -59,15 +60,9 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 
 public class AstCheckerTest {
 
-	@Test
-	void ctLiteralsInCtCaseExpressionShouldHaveCommentsAttached() {
+	@ModelTest("src/test/resources/comment/CommentsOnCaseExpression.java")
+	void ctLiteralsInCtCaseExpressionShouldHaveCommentsAttached(CtModel model) {
 		// contract: literal nodes should have comments attached to them.
-		// arrange
-		Launcher launcher = new Launcher();
-		launcher.addInputResource("src/test/resources/comment/CommentsOnCaseExpression.java");
-		launcher.getEnvironment().setCommentEnabled(true);
-		CtModel model = launcher.buildModel();
-
 		// act
 		List<CtComment> comments = model.getElements(new TypeFilter<>(CtComment.class));
 
--- a/src/test/java/spoon/reflect/ast/AstCheckerTest.java
+++ b/src/test/java/spoon/reflect/ast/AstCheckerTest.java
@@ -22,8 +22,10 @@ import spoon.reflect.code.CtStatement;
 import spoon.reflect.declaration.CtClass;
 import spoon.reflect.reference.CtExecutableReference;
 import spoon.support.modelobs.FineModelChangeListener;
+import spoon.reflect.CtModel;
 import spoon.reflect.code.CtBinaryOperator;
 import spoon.reflect.code.CtBlock;
+import spoon.reflect.code.CtComment;
 import spoon.reflect.code.CtExpression;
 import spoon.reflect.code.CtFieldRead;
 import spoon.reflect.code.CtIf;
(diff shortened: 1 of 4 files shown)


SirYwell:
This doesn't look specific to case expressions or am I missing something? It seems like the comment isn't attached to the CtLiteral, but I didn't really figure out yet how the JDZCommentBuilder works (or is supposed to work). Adding a
```java
			@Override
			public <T> void visitCtLiteral(CtLiteral<T> e) {
				e.addComment(comment);
			}
``` 
to this visitor https://github.com/INRIA/spoon/blob/2fff4c50266f79bc06b961e3b01ca6c6e0fcc1ab/src/main/java/spoon/support/compiler/jdt/JDTCommentBuilder.java#L215
seems to do the trick, however I'm not sure if that's an appropriate fix at all (and it prints the comment in front of the literal, but that might be rather a printer issue - I don't now).

algomaster99:
> This doesn't look specific to case expressions or am I missing something?

I observed it only for literals inside `CtCaseExpression`, so I did not want to generalise it. The fix you have proposed seems to intimate that this is a general problem for `CtLiteral`s.

> however I'm not sure if that's an appropriate fix at all (and it prints the comment in front of the literal, but that might be rather a printer issue - I don't now).

Let's fix the AST in this PR and then we can focus on printing comments.

src/test/java/spoon/test/fieldaccesses/FieldAccessTest.java
```diff
@@ -415,7 +416,7 @@ class CounterScanner extends CtScanner {
 			@Override
 			public <T> void visitCtFieldWrite(CtFieldWrite<T> fieldWrite) {
 				visited++;
-				assertEquals("a", ((CtVariableWrite) fieldWrite.getTarget()).getVariable().getSimpleName());
+				assertEquals("a", ((CtVariableRead<?>) fieldWrite.getTarget()).getVariable().getSimpleName());
```
algomaster99, line 419: This test was failing as I said [here](https://github.com/INRIA/spoon/pull/4836#issuecomment-1212940313). I think it is supposed to be `CtVariableRead` because `fieldWrite.getTarget` evaluates to `a` in `a.l` in this [file](https://github.com/INRIA/spoon/blob/ae305c31620d3a3fdd704119babe006636779593/src/test/resources/noclasspath/FieldAccessRes.java#L5).

algomaster99:
@SirYwell Can you have a look again? I added your suggested fix which fixes the AST parsing.

src/test/resources/comment/CommentsOnCaseExpression.java
```diff
@@ -0,0 +1,19 @@
+public class CommentsOnCaseExpression {
+    public void commentsShouldBeAttachedToCtLiteral(int stage) {
+        String stageStr;
+        boolean fullStatus;
+        switch (stage) {
+            case (1/*org.apache.coyote.Constants.STAGE_PARSE*/):
```
MartinWitt, line 6: Could you improve the test case and a comment before a case, e.g.,`/*org.apache.coyote.Constants.STAGE_PARSE*/1`

algomaster99: I added inline comment to add variety.

src/test/java/spoon/test/fieldaccesses/FieldAccessTest.java
algomaster99: Probably I had some local changes and that is why it was failing. I have reverted them in the latest [commit](https://github.com/INRIA/spoon/pull/4836/commits/59ccc5a6c5e454197ae795b72cfcbcdfa874ed8e).

src/test/java/spoon/reflect/ast/AstCheckerTest.java
```diff
@@ -57,6 +59,22 @@
 
 public class AstCheckerTest {
 
+	@Test
```
SirYwell, line 62: You could use the `@ModelTest` annotation here instead. This way, you don't need the arrange block.
There are a few examples already you can look into if needed.

SirYwell (commented):
Fix looks good, although I think the whole class needs some attention in future.

I left a comment how you can simplify the test.

src/test/java/spoon/reflect/ast/AstCheckerTest.java
```diff
@@ -57,6 +59,22 @@
...
+	void ctLiteralsInCtCaseExpressionShouldHaveCommentsAttached() {
```
MartinWitt, line 63: You could use the new `ModelTest` Annotation here. This removes all of the arrange code.

MartinWitt (approved):
Seems fine for me. If you want you can use the `ModelTest` way to write test cases.

SirYwell (approved):

MartinWitt:
Thanks @algomaster99

PR #4842, opened 2022-08-15
review: chore: Add maven installed question

MartinWitt:
This PR simply adds the question if the maven has maven installed. This is for many failing tests a critical information.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -88,5 +88,5 @@ body:
       options:
         - label: "yes"
         - label: "no"
-  validations:
-    required: true
+    validations:
+      required: true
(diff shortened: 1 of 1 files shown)


I-Al-Istannen:
I doubt this really warrants a point here: it is quite specialized and even if the user set it up, it might only be visible in a shell session after the profile/bashrc are loaded (or windows equivalents, as far as they exist). Therefore, the user's expression might not be really all that helpful.

If we want to go down that route, we should probably use a [checkbox](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-githubs-form-schema#checkboxes) though.

slarse:
I also feel this is too specialized for the general issue template. And like @I-Al-Istannen pointed out, it'd probably be better to use a checkbox.

But I'd personally not include this in the template at all.

PR #4843, opened 2022-08-15
review: Use official qodana github action

MartinWitt:
Like ever enterprise project, the developers make horrible decisions at someday and remove the usability.
The Qodana action no longer supports the options we gave it. We need to provide these options as a single string. This is a result of the change in the action using Qodana CLI. Qodana only gave a warning in the log for this behavior, and the last runs since this change could have produced wrong results.
Maybe someone with more advanced YAML skills has a solution for this hideous long string?


.github/workflows/tests.yml
```diff
@@ -157,19 +158,13 @@ jobs:
...
-          project-dir: "${{ github.workspace }}/pull_request"
-          inspected-dir: ./src/main/java # only main spoon project at first
-          results-dir:  "${{ github.workspace }}/result"
-          changes: true
-          fail-threshold: 0
+          args: --linter, ${{ env.QODANA_LINTER }}, project-dir,${{ github.workspace }}/pull_request,changes,true,fail-threshold,0
           use-annotations: false
           use-caches: false # we disable cache for consistent results
-      - uses: github/codeql-action/upload-sarif@3f62b754e23e0dd60f91b744033e1dc1654c0ec6 # tag=v2
+      - uses: github/codeql-action/upload-sarif@v2
```
slarse, line 164: Don't remove the commit pins. Pinning on tag is not safe as they can move. But most prominently it leads to unexpected failures when devs introduce breaking changes and move tags to those.

.github/workflows/tests.yml
```diff
@@ -157,19 +158,13 @@ jobs:
       - name: Qodana - Code Inspection
         uses: JetBrains/qodana-action@ab695492d98902c49b33382782175107ec63ca28 # tag=v2022.2.1
         with:
-          linter: ${{ env.QODANA_LINTER }}
-          project-dir: "${{ github.workspace }}/pull_request"
-          inspected-dir: ./src/main/java # only main spoon project at first
-          results-dir:  "${{ github.workspace }}/result"
-          changes: true
-          fail-threshold: 0
+          args: --linter, ${{ env.QODANA_LINTER }}, project-dir,${{ github.workspace }}/pull_request,changes,true,fail-threshold,0
```
slarse, line 161: You could split this line on multiple lines (if GHA supports this YAML syntax, but I think it should):

```suggestion
          args: >-
              --linter,
              ${{ env.QODANA_LINTER }},
              project-dir,
              ${{ github.workspace }}/pull_request,
              changes,
              true,
              fail-threshold,
              0
```

src/main/java/spoon/support/compiler/ZipFolder.java
```diff
@@ -39,6 +39,9 @@ public ZipFolder(File file) throws IOException {
 		if (!file.isFile()) {
 			throw new IOException(file.getName() + " is not a valid zip file");
 		}
+		if (false==true) {
+			String s = new String[0].toString();
+		}
```
slarse, line 44: Guessing this is just to trigger an error?

slarse (changes requested):

src/main/java/spoon/support/compiler/ZipFolder.java
MartinWitt: Yes

MartinWitt:
This ports Qodana to the new style. We fail the CI if any problems are found. If we see this is too strict, we can disable it.

I-Al-Istannen:
I'd be happy to merge if you could quickly test whether it really fails the check. There were significant changes since the last time this was tested in this PR, if I did not miss anything.

slarse:
@MartinWitt :point_up: conflict :o

I've been incredibly busy with other things for the past couple of weeks and feel a bit out of the loop, is there anything keeping us from merging this (except for the conflict)?

I-Al-Istannen:
> I've been incredibly busy with other things for the past couple of weeks and feel a bit out of the loop, is there anything keeping us from merging this (except for the conflict)?

Your guess is as good as mine, but I don't think so. I would really love to get this merged though, as the current pinned hash is now expired and qodana runs fail with:
```
This IDE build has expired. Please download a new build from JetBrains official site
✗  Qodana exited with code 7
```
See https://github.com/INRIA/spoon/actions/runs/3505578524/jobs/5871965639.

MartinWitt:
I fixed the conflict. Shall I push some bad to showcase Qodana is working?

I-Al-Istannen:
> I fixed the conflict. Shall I push some bad to showcase Qodana is working?

I'd appreciate it :)

MartinWitt:
@I-Al-Istannen CI is red as wanted.

I-Al-Istannen:
Looks good to me :)

I-Al-Istannen (approved):

slarse:
Looks like CI extra checks are failing as snapshots can't be fetched from OW2 repository :/

I-Al-Istannen:
@slarse Fixed now, I think you can review/merge it. Your negative review is greying out the merge button — though I might be able to overrule you.

slarse (approved):

slarse:
@I-Al-Istannen I'd be more comfortable if you merged as you were the last to look at it, I've approved so there shouldn't be any problems now.

I-Al-Istannen:
> I'd be more comfortable if you merged as you were the last to look at it, I've approved so there shouldn't be any problems now.

A wise decision, as qodana no longer seems to perform delta comparisons. I was a bit suspicious but the test here seems to have worked so I accepted it. I did *not* realize that @MartinWitt was actually **NOT testing his changes here** — the target was not `pull_request_target`. One day I'll remember all the pitfalls of Github Actions.

So, @MartinWitt could you make a new PR with actual delta checks?

slarse:
> A wise decision, as qodana now failed on master :)

Jackpot.

Issue #4834, opened 2022-08-09
[Bug]: CtExecutableReference cannot use CtTypeParameterReference. Use boundingType of CtTypeParameterReference instead.


dotSp0T:
### Describe the bug

Updating an older project from spoon 5.5.x to latest version. Getting exception I have issues understanding the problem with...

Stacktrace:

```
Exception in thread "main" spoon.SpoonException: CtExecutableReference cannot use CtTypeParameterReference. Use boundingType of CtTypeParameterReference instead.
	at spoon.support.reflect.reference.CtExecutableReferenceImpl.checkMethodParameterTypeRef(CtExecutableReferenceImpl.java:184)
	at spoon.support.reflect.reference.CtExecutableReferenceImpl.addParameter(CtExecutableReferenceImpl.java:176)
	at spoon.support.reflect.reference.CtExecutableReferenceImpl.setParameters(CtExecutableReferenceImpl.java:167)
	at spoon.reflect.factory.CodeFactory.createConstructorCall(CodeFactory.java:192)
	at spoon.reflect.factory.FactoryImpl.createConstructorCall(FactoryImpl.java:479)
	at org.crumbleworks.forge.spoon_openglparser.LwjglOpenGLAPIProcessor.createCustomPointerBufferAndWrapper(LwjglOpenGLAPIProcessor.java:1312)
	at org.crumbleworks.forge.spoon_openglparser.LwjglOpenGLAPIProcessor.process(LwjglOpenGLAPIProcessor.java:772)
	at spoon.support.QueueProcessingManager.process(QueueProcessingManager.java:115)
	at spoon.support.QueueProcessingManager.process(QueueProcessingManager.java:132)
	at spoon.support.compiler.jdt.JDTBasedSpoonCompiler.process(JDTBasedSpoonCompiler.java:194)
	at spoon.Launcher.process(Launcher.java:791)
	at org.crumbleworks.forge.spoon_openglparser.LwjglOpenGLAPIProcessor.main(LwjglOpenGLAPIProcessor.java:1478)
```

Code:
```
wrapMethod.setBody(getFactory().Core().createReturn().setReturnedExpression(
        getFactory().createConstructorCall(
                cwEnginePointerBufferWrapper.getReference(),
                methodCall)
        ));
```

I cannot provide the full sourcecode as of this moment, as there's a bunch of proprietory stuff in there. I am trying to narrow it down to a self-contained executable sample

### Spoon Version

10.1.1

### JVM Version

18

[...]

PR #4844, opened 2022-08-16
feat: Use inferred generic type in invocations

I-Al-Istannen:
If no explicit type arguments are given, the type arguments JDT inferred are now used for getActualTypeArguments.

This addresses #4834, but I am not sure we actually want to use the inferred information. It seems quite useful, but it may also be incomplete if JDT is not able to infer things correctly.

### DIFF
--- a/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
+++ b/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
@@ -132,6 +132,9 @@ import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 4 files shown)


SirYwell (commented):
I think it makes sense to use the information in the model. While it makes the printer a bit more complex, it helps with analysing code. I think we should look into more places where we don't use implicit information.

For the `PrintTypeArguments` I'd prefer to either find better suited names or document it properly. It also looks like the `writeActualTyoeArguments()` method is public, so I'm not sure if we can just add a parameter to it.

I-Al-Istannen:
I overloaded it, so existing callers and subclasses should be fine. I also renamed them and added some documentation, maybe that's nicer now?

SirYwell (approved):
Yes, that's better

MartinWitt:
Thanks @I-Al-Istannen

PR #4845, opened 2022-08-18
review: refactor: SizeReplaceableByIsEmpty

MartinWitt:
# Changelog
## SizeReplaceableByIsEmpty
Checking if a string is empty should be done by `String#isEmpty` instead of `String.length==0`

### SizeReplaceableByIsEmpty
- Replaced string.length empty check with string.isEmpty


SirYwell (approved):

I-Al-Istannen (approved):

slarse (approved):
Thanks @MartinWitt !

PR #4848, opened 2022-08-19
review: perf: avoid collection copy in CtPackageImpl

SirYwell:
I saw that when running this test https://github.com/INRIA/spoon/blob/0fcb28878671d90daede513326bff96274068b32/src/test/java/spoon/test/architecture/SpoonArchitectureEnforcerTest.java#L157

most of the time is spent on `CtMethodImpl#getTopDefintions()`. I did some profiling and encountered this:
lhs

That's an immense amount of time of copying elements just to check if the set is empty. As a solution, I propose to introduce a `CtPackage#hasTypes()` method. This way, we can ask the ElementNameMap directly instead of doing costly copying.

In my first measurements, this reduced the runtime of the mentioned test from ~20s to ~16s.

(I rebased after that, and the runtime of the test on the current master takes >1m on my machine. With this patch, it is still faster, but I have a second PR ready to get back to the ~16s)

### DIFF
--- a/src/main/java/spoon/reflect/declaration/CtPackage.java
+++ b/src/main/java/spoon/reflect/declaration/CtPackage.java
@@ -59,6 +59,8 @@ public interface CtPackage extends CtNamedElement, CtShadowable {
 
 	/**
 	 * Gets the set of included child packages.
+	 * This method might take linear time (regarding the amount of packages in this package).
+	 * For emptiness-checks, {@link #hasPackages()} should be preferred.
 	 */
 	@PropertyGetter(role = SUB_PACKAGE)
 	Set<CtPackage> getPackages();
@@ -91,6 +93,8 @@ public interface CtPackage extends CtNamedElement, CtShadowable {
 
 	/**
 	 * Returns the set of the top-level types in this package.
+	 * This method might take linear time (regarding the amount of types in this package).
+	 * For emptiness-checks, {@link #hasTypes()} should be preferred.
 	 */
 	@PropertyGetter(role = CONTAINED_TYPE)
 	Set<CtType<?>> getTypes();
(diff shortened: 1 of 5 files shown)


src/main/java/spoon/support/reflect/declaration/CtPackageImpl.java
```diff
@@ -242,7 +242,12 @@ public boolean hasPackageInfo() {
 
 	@Override
 	public boolean isEmpty() {
-		return getPackages().isEmpty() && getTypes().isEmpty();
+		return getPackages().isEmpty() && !hasTypes();
```
slarse, line 245: Doesn't `getPackages()` have precisely the same problem (allocates a new `LinkedHashSet` on each invocation)? Perhaps kill two birds with one stone in this PR?

SirYwell: Yes, `getPackages()` has the same problem, but as it didn't really show up in my profiles (as it's called far less), I didn't include an extra method for the empty check in this PR. If you want, I can do so though.

slarse: It's probably not a performance problem because the amount of packages is typically far fewer than the amount of types in any given package.

I think from an API standpoint it makes sense to have both. Since there are methods `getTypes()` and `getPackages()`, I probably would expect a `hasTypes()` to be accompanied by a `hasPackages()`. What do you think?

src/main/java/spoon/reflect/declaration/CtPackage.java
```diff
@@ -161,4 +161,10 @@ public interface CtPackage extends CtNamedElement, CtShadowable {
 	 * @return true if the package contains no types nor any other packages
 	 */
 	boolean isEmpty();
+
+	/**
+	 * @return true if the package contains any types.
+	 * @see #getTypes()
+	 */
```
slarse, line 168: Now that I think about it, we should probably add a requirement here that this is O(1). That's really the reason for it to exist in the first place.

On that note, we maybe should add notes to `getTypes()` and `getPackages()` that they may be linear in the amount of types/packages, and one should use `isEmpty()`, `hasPackages()`  (if we add it) or `hasTypes()` to check for "emptiness".

slarse (changes requested):

SirYwell:
I added some runtime behavior documentation and the `hasPackages()` method.

slarse (approved):
Thanks @SirYwell!

PR #4850, opened 2022-08-19
fix: Fix getOriginalSourceFragment and getRawJavadoc for zip file inputs

I-Al-Istannen:
## Description of the problem
Previously, zip files were flattened and stored as `<name><random number>.java`. This caused all methods that retrieve code
at runtime, such as `getOriginalSourceFragment` or `getRawJavadoc,` to fail.

## Chosen solution
This patch keeps the structure of the zip file intact while extracting it into a temporary folder. Consequently, the aforementioned methods are now able to find the source files.

There were multiple ways to proceed here. As JDT needs a copy anyways, I chose to proactively copy the files to disk instead of storing them in memory and copying them when the sources are prepared. Then I set `isActualFile` to true, tricking the relevant parts of spoon into treating the zip file as a regular file (which it mostly is by now, as it has an on-disk representation). The old constructor is kept in place (but will cause problems) as the class is public. Nobody should have used it in the first place, so maybe we could also delete it?

Instead of changing the folder structure we could change the compilation unit name named passed to JDT, This sounds a bit deadly and is also absolutely not compatible with existing code looking at the compilation unit layer. Every compilation unit would be named `/tmp/MyClass381891.java` or similiar, which is not that helpful.

## Further possible features
Allow the user to specify where the temporary files should be stored when creating a `ZipFolder`. This can be easily retrofitted in the future though.

### DIFF
--- a/src/main/java/spoon/support/Internal.java
+++ b/src/main/java/spoon/support/Internal.java
@@ -13,10 +13,10 @@ import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
 /**
- * Tells that a class or method is not in the public API (even if it has Java visibility "public")
+ * Tells that an element is not in the public API (even if it has Java visibility "public")
  * Required because package-visibility is too coarse-grained.
  */
 @Retention(RetentionPolicy.RUNTIME)
-@Target({ ElementType.METHOD, ElementType.TYPE })
+@Target({ElementType.CONSTRUCTOR, ElementType.METHOD, ElementType.TYPE })
 public @interface Internal {
 }
--- a/src/main/java/spoon/support/compiler/ZipFile.java
+++ b/src/main/java/spoon/support/compiler/ZipFile.java
@@ -29,7 +29,7 @@ public class ZipFile implements SpoonFile {
 	private final byte[] content;
 
 	/**
-	 * Creates a new zip file. Should never be called manually.#
+	 * Creates a new zip file. Should never be called manually.
 	 *
 	 * @param parent the parent folder
 	 * @param name the name of the file
--- a/src/main/java/spoon/support/compiler/ZipFile.java
+++ b/src/main/java/spoon/support/compiler/ZipFile.java
@@ -7,32 +7,68 @@
  */
 package spoon.support.compiler;
 
+import spoon.SpoonException;
+import spoon.compiler.SpoonFile;
+import spoon.compiler.SpoonFolder;
+import spoon.support.Internal;
+
 import java.io.ByteArrayInputStream;
 import java.io.File;
+import java.io.IOException;
 import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.Arrays;
 import java.util.Objects;
 
-import spoon.compiler.SpoonFile;
-import spoon.compiler.SpoonFolder;
-
 public class ZipFile implements SpoonFile {
 
-	byte[] content;
+	private final String name;
+	private final ZipFolder parent;
+	private final Path tempFile;
+	private final byte[] content;
 
-	String name;
-
-	ZipFolder parent;
-
+	/**
+	 * Creates a new zip file. Should never be called manually.#
+	 *
+	 * @param parent the parent folder
+	 * @param name the name of the file
+	 * @param content the content of the file
+	 * @deprecated use {@link ZipFile#ZipFile(ZipFolder, String, Path)}
+	 */
+	@Deprecated
 	public ZipFile(ZipFolder parent, String name, byte[] content) {
 		this.content = content;
 		this.name = name;
 		this.parent = parent;
+		this.tempFile = null;
+	}
+
+	/**
+	 * Creates a new zip file. Should never be called manually.
+	 *
+	 * @param parent the parent folder
+	 * @param name the name of the file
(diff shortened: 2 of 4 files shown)


src/main/java/spoon/support/Internal.java
```diff
@@ -13,10 +13,10 @@
 import java.lang.annotation.Target;
 
 /**
- * Tells that a class or method is not in the public API (even if it has Java visibility "public")
+ * Tells that an element is not in the public API (even if it has Java visibility "public")
  * Required because package-visibility is too coarse-grained.
  */
 @Retention(RetentionPolicy.RUNTIME)
-@Target({ ElementType.METHOD, ElementType.TYPE })
+@Target({ElementType.CONSTRUCTOR, ElementType.METHOD, ElementType.TYPE })
```
I-Al-Istannen, line 20: This was needed as I wanted to mark the new constructor of the ZipFile internal.

src/main/java/spoon/support/compiler/ZipFolder.java
```diff
@@ -57,22 +65,47 @@ public List<SpoonFile> getFiles() {
...
-				ByteArrayOutputStream output = new ByteArrayOutputStream(2048)) {
-
-				ZipEntry entry;
-				while ((entry = zipInput.getNextEntry()) != null) {
-					zipInput.transferTo(output);
-					files.add(new ZipFile(this, entry.getName(), output.toByteArray()));
-					output.reset();
+			try (FileSystem zip = FileSystems.newFileSystem(URI.create("jar:" + file.toURI()), Map.of())) {
+				Path tempFolder = Files.createTempDirectory("spoon-zip-file-proxy");
+				// Try to clean up - not guaranteed to work!
```
I-Al-Istannen, line 70: This wasn't guaranteed before as well and depends on the mood of the JVM. `File#deleteOnExit()` which was used before in Sources also registers a shutdown hook.

src/main/java/spoon/support/compiler/ZipFolder.java
```diff
@@ -57,22 +65,47 @@ public List<SpoonFile> getFiles() {
...
 			files = new ArrayList<>();
-			try (ZipInputStream zipInput = new ZipInputStream(new BufferedInputStream(new FileInputStream(file)));
-				ByteArrayOutputStream output = new ByteArrayOutputStream(2048)) {
-
-				ZipEntry entry;
-				while ((entry = zipInput.getNextEntry()) != null) {
-					zipInput.transferTo(output);
-					files.add(new ZipFile(this, entry.getName(), output.toByteArray()));
-					output.reset();
+			try (FileSystem zip = FileSystems.newFileSystem(URI.create("jar:" + file.toURI()), Map.of())) {
```
I-Al-Istannen, line 68: This can be made nicer in Java 13+, but for now we are stuck with it. URI.toString() should round-trip, so this *should* be safe for arbitrary paths.

src/main/java/spoon/support/compiler/ZipFolder.java
```diff
@@ -57,22 +65,47 @@ public List<SpoonFile> getFiles() {
...
+					}
+				}));
+
+
+				for (Path directory : zip.getRootDirectories()) {
+					copyFolder(directory, tempFolder);
 				}
 			} catch (Exception e) {
-				Launcher.LOGGER.error(e.getMessage(), e);
+				Launcher.LOGGER.error("Error copying zip file contents", e);
```
I-Al-Istannen, line 84: `LOGGER.error` crashes with `null` messages, which is not ideal if you are trying to log such a message.

I-Al-Istannen (commented):
I could add the reviews here as comments, but they are probably a bit too meta.

src/test/java/spoon/support/compiler/ZipFileTest.java
```diff
@@ -0,0 +1,67 @@
...
+		// contract: Fetching an original source code fragment from a zip folder input works
+		Launcher launcher = new Launcher();
+		launcher.addInputResource(new ZipFolder(createZip(tempDir).toFile()));
+		launcher.buildModel();
+
+		CtType<?> type = launcher.getFactory().Type().get("a.Test");
+		CtMethod<?> method = type.getMethod("foo");
+		assertNotNull(method.getOriginalSourceFragment().getSourceCode());
+		assertTrue(method.getOriginalSourceFragment().getSourceCode().contains("/**"));
+		assertTrue(method.getOriginalSourceFragment().getSourceCode().contains("foo()"));
```
SirYwell, line 34: Using hamcrest here might be better

SirYwell (commented):
LGTM

src/main/java/spoon/support/compiler/ZipFile.java
```diff
@@ -7,32 +7,68 @@
...
-
-	ZipFolder parent;
-
+	private final String name;
+	private final ZipFolder parent;
+	private final Path tempFile;
+	private final byte[] content;
+
+	/**
+	 * Creates a new zip file. Should never be called manually.#
```
intrigus, line 32:
```suggestion
	 * Creates a new zip file. Should never be called manually.
```

MartinWitt (approved):

MartinWitt:
Thanks @I-Al-Istannen

PR #4851, opened 2022-08-19
refactor: fix bad smell "Non-Strict-Comparison-Can-Be-Equality"

MartinWitt:
# Changelog
## Non-Strict-Comparison-Can-Be-Equality
Inequality conditions that, according to data flow analysis, can be satisfied only for a single operand value. Such conditions could be replaced with equality conditions to make the code clearer.

### Non-Strict-Comparison-Can-Be-Equality
- Replaced inequality with simpler equality

### DIFF
--- a/src/main/java/spoon/support/comparator/DeepRepresentationComparator.java
+++ b/src/main/java/spoon/support/comparator/DeepRepresentationComparator.java
@@ -28,7 +28,7 @@ public class DeepRepresentationComparator implements Comparator<CtElement>, Seri
 		}
 		String current = getDeepRepresentation(o1);
 		String other = getDeepRepresentation(o2);
-		if (current.length() == 0 || other.length() == 0) {
+		if (current.isEmpty() || other.isEmpty()) {
 			throw new ClassCastException("Unable to compare elements");
 		}
 		return current.compareTo(other);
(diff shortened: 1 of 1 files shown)


I-Al-Istannen (approved):

src/main/java/spoon/support/comparator/DeepRepresentationComparator.java
```diff
@@ -28,7 +28,7 @@ public int compare(CtElement o1, CtElement o2) {
 		}
 		String current = getDeepRepresentation(o1);
 		String other = getDeepRepresentation(o2);
-		if (current.length() <= 0 || other.length() <= 0) {
+		if (current.length() == 0 || other.length() == 0) {
```
slarse, line 31: ?

```suggestion
		if (current.isEmpty() || other.isEmpty()) {
```

slarse (changes requested):

src/main/java/spoon/support/comparator/DeepRepresentationComparator.java
I-Al-Istannen: And now @MartinWitt has to run his tool until it finds a fixpoint or cycle :rocket:

MartinWitt: You are correct, I had not the ability to fix style issues with a fix point algorithm.

monperrus: FYI, fixed point fixing is the main feature of https://github.com/lyxell/logifix by @lyxell 
cc/ @fermadeiral

slarse (approved):
Thanks @MartinWitt

PR #4862, opened 2022-08-25
perf: Don't clone method body when adapting it for isOverriding

I-Al-Istannen:
Cloning is quite slow and method bodies can be large - this results in the `isOverriding` check taking quite a bit longer than needed.

Removing the type adaption is not possible as the javadoc explains: If we do not adapt the method, we have no `CtTypeParameter` declaration lying around our references can attach to. This results in `getTypeErasure()` potentially entering an infinite recursion.

There are some more gains to be had if one really wants to, but I currently don't see the need. This one is quite unobtrusive and doesn't really make the code more confusing.

### Side effects
This creates an unnecessary change event for the Sniper printer (body changed), but it seems to ignore that and properly print it anyways.

### DIFF
--- a/src/main/java/spoon/support/adaption/TypeAdaptor.java
+++ b/src/main/java/spoon/support/adaption/TypeAdaptor.java
@@ -9,6 +9,7 @@ package spoon.support.adaption;
 
 import spoon.SpoonException;
 import spoon.processing.FactoryAccessor;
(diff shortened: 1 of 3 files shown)


SirYwell (approved):
LGTM. Makes sense to avoid costs that depend on the method body here.

MartinWitt:
Thanks @I-Al-Istannen

PR #4870, opened 2022-08-28
review: chore: Fix javadoc regression script with newer checkstyle version

### LINKED ISSUE #4869: Javadoc quality check does not work with checkstyle-maven-plugin 3.2.0
After merging #4859 CI started failing on the master branch. So something about it is incompatible with the javadoc style script. Here's the output:

```
Run ./chore/check-javadoc-regressions.py COMPARE_WITH_MASTER
$ git rev-parse --show-toplevel
$ git checkout master
$ mvn -B checkstyle:check --fail-never -Dcheckstyle.config.location=/tmp/tmppz3dbyx5.xml
$ git checkout -
$ mvn -B checkstyle:check --fail-never -Dcheckstyle.config.location=/tmp/tmppz3dbyx5.xml
The Checkstyle run for the reference branch did not yield any result.
```

https://github.com/INRIA/spoon/runs/8057134947?check_suite_focus=true#step:9:21

So we had to roll back for the moment

I-Al-Istannen:
It seems like I added a nicer error message (as you can see in the screenshot) but did not print it for comparison failures. Judging by slarse opening an issue, this indeed was the case and I still failed :P
```py
# If the user has a maven problem that prevents checkstyle from running they'd likely
# appreciate *some* output
```

Following that suggestion, the error message was
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.2.0:check (default-cli) on project spoon-core: Failed during checkstyle configuration: cannot initialize module TreeWalker - cannot initialize module JavadocMethod - Property 'scope' does not exist, please check the documentation -> [Help 1]
```

---

Note that the CI status will not be significant as it runs with the version on master. Also note that running `COMPARE_WITH_MASTER` will *fail* on this PR, as the master state was reverted and uses the older checkstyle version, breaking the check:
grafik

Closes #4869

### DIFF
--- a/chore/check-javadoc-regressions.py
+++ b/chore/check-javadoc-regressions.py
@@ -58,6 +58,21 @@ def hl(text: str) -> str:
     return colored(text, _COLOR_HIGHLIGHT)
 
 
+def print_checkstyle_not_parseable_error():
+    """
+    If the user has a maven problem that prevents checkstyle from running they'd likely
+    appreciate *some* output. If we can't extract a violation count the checkstyle plugin
+    probably didn't run successfully. Let's warn them.
+    """
+    print()
+    print(warn("Checkstyle output doesn't contain a violation count. Maybe your setup has an error?"))
+    if len(sys.argv) > 1:
+        print(
+            f" {hl('Suggestion')}: Run this script without arguments and inspect the unfiltered "
+            "checkstyle output"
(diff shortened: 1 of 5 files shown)


slarse (approved):
Nice, thanks @I-Al-Istannen

PR #4873, opened 2022-08-29
fix: fix top definitions filtering

SirYwell:
With multiple independent interfaces declaring the same method, `getTopDefinitions()` returned wrong results.

The filtering was broken due to `TypeAdaptor#isOverriding(m, m) == true` *and* the wrong initialization of the method (as all methods were already added to the result).

### DIFF
--- a/src/main/java/spoon/support/reflect/declaration/CtMethodImpl.java
+++ b/src/main/java/spoon/support/reflect/declaration/CtMethodImpl.java
@@ -29,6 +29,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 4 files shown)


MartinWitt:
Do I understand that correct, that before we could only find a single top definition even if the method has multiple, and now we find all?

SirYwell:
No, it returned *all* methods that were overridden, not just the top definitions. In my test case, it returned the methods from `C`, `B`, `A1` and `A2`. Now it only returns the methods from `A1` and `A2`, as in `B` and `C`, the method is only redefined. Basically, the whole loop at the bottom of `getTopDefinitions()` did *not* change the result anymore.

PR #4893, opened 2022-09-09
test: Introduce a custom temp dir to work around windows file locking

### LINKED ISSUE #4877: MavenLauncherTest fails on Windows - can't delete temporary directory
Out of the blue, one of the MavenLauncher tests just started failing on Windows with this error message:

```
Error:  Errors: 
Error:    MavenLauncherTest.spoonMavenLauncherTest(Path) � IO Failed to delete temp directory C:\Users\RUNNER~1\AppData\Local\Temp\junit13011363813463001138. The following paths could not be deleted (see suppressed exceptions for details): , src, src\test, src\test\resources, src\test\resources\maven-launcher, src\test\resources\maven-launcher\java-11
```

See https://github.com/INRIA/spoon/runs/8101496901?check_suite_focus=true

I-Al-Istannen:
Adds a test execution listener that maintains a `spooned` temporary directory. This is intended as a replacement to the normal Junit `@TempDir`, which fails to clean up files due to file locking on Windows.

### Cons
- The temporary folder will be created before *every test*, which doesn't sound pleasant. Maybe we should only delete it if it actually contains files?

### Further steps
We need to update every test using `@TempDir` or `"spooned"` directly to use it. This could be directly done in this PR.

Closes #4877.

### DIFF
--- a/src/test/java/spoon/MavenLauncherTest.java
+++ b/src/test/java/spoon/MavenLauncherTest.java
@@ -30,7 +30,6 @@ import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
-import org.junit.jupiter.api.io.TempDir;
 import spoon.compiler.SpoonResource;
 import spoon.reflect.declaration.CtElement;
 import spoon.reflect.reference.CtTypeReference;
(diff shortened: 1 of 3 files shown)


MartinWitt:
> We need to update every test using @TempDir or "spooned" directly to use it. This could be directly done in this PR.

Can we first use this on the maven tests only? These are somewhat hard failing now. If it works, we can use it for all in another big PR.

I-Al-Istannen:
We can absolutely do this in multiple steps and merge it in its current form as a band-aid fix first.

MartinWitt (approved):
LGTM let's merge this tomorrow and give it a try

slarse:
Ah, nice, thanks spearheading the war against Windows' file system @I-Al-Istannen

PR #4904, opened 2022-09-15
doc: change "upcall" to "call"

chammp:
Just fixing a small typo in a couple of Javadoc comments. Methods just _call_ each other 😉

### DIFF
--- a/src/main/java/spoon/processing/Processor.java
+++ b/src/main/java/spoon/processing/Processor.java
@@ -45,12 +45,12 @@ public interface Processor<E extends CtElement> extends FactoryAccessor {
 	boolean isToBeProcessed(E candidate);
 
 	/**
-	 * A callback method upcalled by the meta-model scanner to perform a
-	 * dedicated job on the currently scanned element. The way Spoon upcalls
+	 * A callback method called by the meta-model scanner to perform a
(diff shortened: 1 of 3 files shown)


I-Al-Istannen (approved):
There are upcalls between functions, but AFAIK they typically are up from a deeper layer in e.g. a hierarchical architecture. [One use I know is for scheduler upcalls when using a hybrid threading model](https://lkml.iu.edu/hypermail/linux/kernel/9809.3/0922.html).

Spoon is probably not *quite* at that level yet :P Thanks, will merge :)

slarse:
"Upcall" is sometimes used in the sense of "calling a callback function". I think the word is esoteric at best and don't mind seeing it removed.

I-Al-Istannen:
Thanks @chammp!

As an aside:
Your commit was authored by a mail not associated with your GitHub account so GitHub insisted on changing it to one that is. If you do not want this to happen, you should probably either add the gmail address to your GitHub profile or commit with the GitHub mail. If the discrepancy is intended, we might be able to figure something out next time :)

chammp:
> Thanks @chammp!
> 
> As an aside: Your commit was authored by a mail not associated with your GitHub account (I assume) 

True. That was a mistake on my side when manually configuring the email address for this repo with git command line X_x. Thanks for the heads up :)

PR #4903, opened 2022-09-14
style: fix Bad Smells PointlessBooleanExpression

MartinWitt:
# Repairing Code Style Issues
## PointlessBooleanExpression
Boolean expressions shouldn't be overcomplex. 
# Repairing Code Style Issues
* PointlessBooleanExpression (95x)

@I-Al-Istannen and I concluded creating a single PR for all bad smells of this kind. This touches ~50 files but changes less than 100 lines.
Edit: after we merged this we should add `PointlessBooleanExpression` as Qodana-Rule

PR #4906, opened 2022-09-15
chore: Add PointlessBooleanExpression rule to qodana checks

MartinWitt:
All violations of this rule are fixed in #4903. I see zero cases where we want code like `.. == true` or `... == false`, so we should stop such code in the quality check. As a reminder, we only run the checks on `src/main/java`.


I-Al-Istannen (approved):
LGTM

PR #4910, opened 2022-09-19
doc: update contributor list

monperrus:


MartinWitt:
Technical there are some Qodana reports about the py quality but as the script works and the results are correct, I'm for ignoring these. LGTM will merge

MartinWitt (approved):

PR #4912, opened 2022-09-20
refactor: simplify usage of setType

monperrus:
yesterday, I did some Spoon.

And I got a compilation error for a simple, intuitive:
```java
CtLocalVariable<?> lv = localVariables.get(0);
lv.setType(c.getFactory().Type().createReference(Object.class));
```

That was not satisfying.

Here is a fix.

Fun part: the fix is done with Spoon itself and the Spoon transformation is part of the PR.

@slarse I'm sure you'll have an opinion on this generics related PR :)


monperrus:
Note: `ModelRoleHandlers` had to be regenerated.

MartinWitt:
Could you readd the maven license header with `mvn com.mycila:license-maven-plugin:4.1:format`? Otherwise, I'm all for merging because in had this problem often and always needed this ugly cast. I'm unsure if this is breaking for anyone who uses this generics in a useful way. But I cant really think of any case where this works.

src/test/java/spoon/test/variable/VariableTest.java
```diff
@@ -37,6 +44,64 @@
...
+        // it's source level backward compatible for Spoon itself
+        CtLocalVariable<?> lv = localVariables.get(0);
+        lv.setType(c.getFactory().Type().createReference(Object.class));
+
+        CtLocalVariable lv2 = localVariables.get(0);
+        lv2.setType(c.getFactory().Type().createReference(Object.class));
+
+    }
+
+    @Test
```
MartinWitt, line 67: As a note for @I-Al-Istannen `@modeltest` and Sniper printer don't work together. We need to set the printer before building the model.

src/test/java/spoon/test/variable/VariableTest.java
```diff
@@ -22,12 +22,19 @@
...
 import spoon.reflect.code.CtLocalVariable;
+import spoon.reflect.declaration.CtClass;
+import spoon.reflect.declaration.CtElement;
+import spoon.reflect.declaration.CtMethod;
+import spoon.reflect.declaration.CtParameter;
 import spoon.reflect.factory.TypeFactory;
 import spoon.reflect.visitor.filter.TypeFilter;
+import spoon.support.sniper.SniperJavaPrettyPrinter;
 
+import javax.swing.plaf.IconUIResource;
```
MartinWitt, line 37: Is this import really used?

src/main/java/spoon/reflect/meta/impl/ModelRoleHandlers.java
```diff
@@ -1,4 +1,4 @@
-/*
+/**
```
MartinWitt, line 1: This destroys the license check in the CI btw

MartinWitt (changes requested):

src/test/java/spoon/test/variable/VariableTest.java
I-Al-Istannen: This also won't be a `@ModelTest` as it actually calls `run` instead of buildModel. I omitted support for doing more involved operations on the launcher (run processes it, prettyprints and optionally checks for compilation errors). I am not sure we want to support that, but we probably could allow a `sniper = true` setting if tests need that.

src/test/java/spoon/test/variable/VariableTest.java
```diff
@@ -37,6 +44,64 @@
...
+        lv.setType(c.getFactory().Type().createReference(Object.class));
+
+        CtLocalVariable lv2 = localVariables.get(0);
+        lv2.setType(c.getFactory().Type().createReference(Object.class));
+
+    }
+
+    @Test
+    public void refactor() {
+        // rectoring all setType(CtTypeParameter<T>) -> setType(CtTypeParameter)
```
I-Al-Istannen, line 69: I am not sure if I am a huge fan of a test overwriting *the very source code I am currently editing*. That sounds a bit scary.

What exactly is the purpose of this test? Should it enforce that nobody re-introduces this pattern? In that case, I would probably make this a model test and just *verify*, without printing anything. We have a few of such tests checking various properties of spoon. The test could then be something along these lines:
```java
@ModelTest("src/main/java")
void ensureNoWildcardUsedForSetType...(Model model) {
  model.processWith(new Abstract....() {
    if (isToBeProcessed()) { assertFalse("message"); }
  })
}
```

If we want to automatically refactor it continuously (similar to the CloneBuilder CtGeneration tests), we might want to think about using a similar pattern.

If this is a one-off, I am not sure the test is worth including in master. Maybe it can live in this PR as documentation instead?

src/main/java/spoon/reflect/meta/impl/ModelRoleHandlers.java
```diff
@@ -2350,7 +2350,7 @@ public <T, U> void setValue(T element, U value) {
 		}
 	}
 
-	static class CtSynchronized_BODY_RoleHandler extends SingleHandler<CtSynchronized, CtBlock<? extends Object>> {
```
I-Al-Istannen, line 2353: This change also removes redundant upper bounds in many places. Is this a shortcoming of the sniper printer or intentional?

src/test/java/spoon/test/variable/VariableTest.java
```diff
@@ -37,6 +44,64 @@
...
+                System.out.println(ctParameter.getType().getActualTypeArguments().size());
+                if (!ctParameter.getType().getSimpleName().equals("CtTypeReference")) return false;
+                if (ctParameter.getType().getActualTypeArguments().size()!=1) return false;
+                System.out.println(ctParameter.getType().toString());
+                return true;
+            }
+
+            @Override
+            public void process(CtMethod<?> element) {
+                System.out.println(element.getSignature());
```
I-Al-Istannen, line 95: We can probably remove the prints now (here and in various places) :)

I-Al-Istannen (changes requested):
If I understand correctly, the problem you are solving is that the generic in `setType` is a burden without much use? You have a `CtLocalVariable<?>` and want to set its type. The signature of `setType` forces you to set it to *the same type it already is*. This seems a bit weird, but is the only way you can ensure your expressions stay true to their type parameter.
This is a bit unfortunate, as you want to *change* the type (from an unnamed wildcard capture to `String`).

This means that the following is now possible
```java
CtLocalVariable<String> lv = localVariables.get(0);
lv.setType(c.getFactory().Type().createReferences(int.class));
// lv still thinks it is a String, but it really is not
```

So the  type parameter at `lv` is now wrong and outdated. If you intend to bypass it, why not declare it as a rawtype in the first place?:
```java
CtLocalVariable lv = localVariables.get(0);
lv.setType(c.getFactory().Type().createReference(Object.class));
```
This code should also compile and make it obvious that you are destroying some guarantees.

----

The root cause is probably *that we even have a generic parameter* there, which makes it hard for transformations changing the type to typecheck. I do agree that this offers a sizable UX benefit, as I recently also got really annoyed at CtMethods for behaving the same… At the same time, it also silently corrupts the model — I am not sure that anybody really makes use of the generics in the first place though.

src/test/java/spoon/test/variable/VariableTest.java
monperrus: Ack, removed the @Test annotation per the comment below.

src/test/java/spoon/test/variable/VariableTest.java
monperrus: done

src/main/java/spoon/reflect/meta/impl/ModelRoleHandlers.java
monperrus: Done, probably a bug of sniper.

src/test/java/spoon/test/variable/VariableTest.java
monperrus: Fully agree, removed @Test

src/main/java/spoon/reflect/meta/impl/ModelRoleHandlers.java
monperrus: None of them. This class is generated from CtGenerationTest which does not use sniper.

src/test/java/spoon/test/variable/VariableTest.java
monperrus: done

monperrus:
> If I understand correctly, the problem you are solving is that the generic in setType is a burden without much use?

yes

>  I do agree that this offers a sizable UX benefit

indeed.

> This means that the following is now possible

correct. In theory, those generics are very elegant and make things safer. In  practice, they are hardly usable and actually never used. It's a tradeoff.

src/main/java/spoon/reflect/meta/impl/ModelRoleHandlers.java
MartinWitt: Kind of scary because I cannot think of any reason why this happens, but I'm fine with it.

src/test/java/spoon/test/variable/VariableTest.java
```diff
@@ -37,6 +42,58 @@
...
 public class VariableTest {
 
+    @Test
+    public void testSetType() {
+        // contract: one can use setType without having a very hard generics type checking error
+
+        // to get this done, we use the refactoring below
+        CtClass c = Launcher.parseClass("class C{ void f() {Object i=null; }}");
+        List<CtLocalVariable> localVariables = c.getElements(new TypeFilter<>(CtLocalVariable.class));
+        System.out.println(localVariables);
```
MartinWitt, line 52: I'm not a massive fan of system outs in test cases. This only fills the log. How about removing this?

MartinWitt (changes requested):
Some small comments and additions, otherwise LGTM. If @I-Al-Istannen has no bad feelings for the one guy on earth using spoon generics in a useful way, we can merge it.

src/test/java/spoon/test/variable/VariableTest.java
monperrus: done

src/main/java/spoon/reflect/meta/impl/ModelRoleHandlers.java
I-Al-Istannen: Git bisect found the culprit:
```
0fcb28878671d90daede513326bff96274068b32 is the first bad commit
commit 0fcb28878671d90daede513326bff96274068b32
Author: I-Al-Istannen <I-Al-Istannen@users.noreply.github.com>
Date:   Tue Apr 19 22:46:53 2022 +0200

    refactor: Rewrite type adaption (#4215)

binary search found first bad commit
```

I plead the fifth and refuse any comment. I don't event want to think about why this changed whether wildcard bounds are printed or not, that sounds like a quite unimportant hole for more masochistic moods.

I-Al-Istannen (approved):
The model corruption is *probably* an acceptable tradeoff here. It's a bit weird that the test doesn't actually assert anything, but I don't know a nicer way to write "this should compile" tests either.

I am inclined to agree with MartinWitt.

src/main/java/spoon/reflect/meta/impl/ModelRoleHandlers.java
monperrus: :rofl:

MartinWitt (approved):

slarse:
Sorry, I'm a little bit late to the party, been busy with a trainee program. But it looks like you've all got this under control, so I'm just going to silently back out of here :))

MartinWitt:
Thanks @monperrus

PR #4915, opened 2022-09-23
review: fix: set correct executable reference type for signature polymorphic methods

### LINKED ISSUE #4863: [Bug]: Inconsistency in modelling of invocations of signature polymorphic methods
### Describe the bug

*Another sweet, very specific edge case*
Java has a few [signature polymorphic methods](https://docs.oracle.com/javase/specs/jls/se18/html/jls-15.html#jls-15.12.3).

In my opinion, there are two ways how spoon could model invocations of such methods:

1. Use the return type and the the `Object...` (or `Object[]`) type parameter of the declared method
2. Use the return type and the parameters from the invocation context as described in the JLS.

However, spoon currently picks the third one: The return type is invocation-specific but the parameter is always `Object[]`.

Both solutions are possible as JDT provides the needed information. This issue is rather for discussion which solution is more appropriate for spoon.

```Java
package invoke;

import spoon.Launcher;
import spoon.reflect.CtModel;
import spoon.reflect.code.CtInvocation;

import java.lang.invoke.MethodHandle;
import java.lang.invoke.MethodHandles;

public class SignaturePolymorphic {
  private static final MethodHandle MH = MethodHandles.identity(String.class);
  public static void main(String[] args) throws Throwable {
    Launcher launcher  = new Launcher();
    launcher.addInputResource("src/main/java/invoke/SignaturePolymorphic.java");
    for (CtMethod<?> method : ctModel.getAllTypes().iterator().next().getMethodsByName("a")) {
      CtInvocation<?> lastStatement = method.getBody().getLastStatement();
      lastStatement = (CtInvocation<?>) lastStatement.getArguments().get(0);
      System.out.println(lastStatement.getType());
      System.out.println(lastStatement.getExecutable().getParameters());
    }
  }

  private void a() throws Throwable {
    System.out.println((String) MH.invoke("Hello World"));
  }
}
```

```Java
(contained in the example above)
```

### Actual output

```Java
java.lang.String
[java.lang.Object[]]
```

### Expected output

```Java
Either

java.lang.String
[java.lang.String]

or

java.lang.Object
[java.lang.Object[]]
```

### Spoon Version

10.2.0-beta-16

### JVM Version

17

Windows 11

SirYwell:
Fixes #4863 

As we decided to go the easier route, the change is pretty straightforward: We only need to correct the return type of the reference to always be `java.lang.Object`.

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
+++ b/src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
@@ -457,7 +457,7 @@ public class ReferenceBuilder {
 			//  therefore we use the original one here
 			//  see https://github.com/INRIA/spoon/issues/4863
 			if (exec.isPolymorphic()) {
-				ref.setType(getTypeReference(exec.original().returnType, true));
+				ref.setType(getTypeReference(exec.original().returnType));
 			}
 			final List<CtTypeReference<?>> parameters = new ArrayList<>(exec.original().parameters.length);
 			for (TypeBinding b : exec.original().parameters) {
--- a/src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
+++ b/src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
@@ -452,9 +452,13 @@ public class ReferenceBuilder {
 		}
 
 		// original() method returns a result not null when the current method is generic.
-		// if polymorphic, the original parameter list differs from the actual parameter list
-		//  see https://github.com/INRIA/spoon/issues/4863
-		if (exec.original() != null && !exec.isPolymorphic()) {
+		if (exec.original() != null) {
+			// if polymorphic, the original return type differs from the actual return type
+			//  therefore we use the original one here
+			//  see https://github.com/INRIA/spoon/issues/4863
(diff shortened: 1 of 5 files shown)


SirYwell:
Let the review begin :)

src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
```diff
@@ -453,6 +453,12 @@ <T> CtExecutableReference<T> getExecutableReference(MethodBinding exec, int sour
 
 		// original() method returns a result not null when the current method is generic.
 		if (exec.original() != null) {
+			// if polymorphic, the original return type differs from the actual return type
+			//  therefore we use the original one here
+			//  see https://github.com/INRIA/spoon/issues/4863
+			if (exec.isPolymorphic()) {
+				ref.setType(getTypeReference(exec.original().returnType, true));
```
I-Al-Istannen, line 460:
```suggestion
				ref.setType(getTypeReference(exec.original().returnType));
```
Why do we need to resolve generics here?

I-Al-Istannen (approved):

monperrus:
thanks a lot @SirYwell

PR #4916, opened 2022-09-25
style: fix Bad Smells in spoon.support.util.ImmutableMapImpl

MartinWitt:
# Repairing Code Style Issues
## UnnecessaryToStringCall
The `toString()` method is not needed in cases the underlying method handles the conversion. Also calling toString() on a String is redundant. Removing them simplifies the code.
## Changes: 
* Remove redudant `toString()` call in `parent.toString()`

### DIFF
--- a/src/main/java/spoon/support/util/ImmutableMapImpl.java
+++ b/src/main/java/spoon/support/util/ImmutableMapImpl.java
@@ -88,7 +88,7 @@ public class ImmutableMapImpl implements ImmutableMap {
 		appendMap(sb, map);
 		if (parent != null) {
 			sb.append("\nparent:\n");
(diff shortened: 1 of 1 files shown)


SirYwell (approved):
Looks good.

As a side note, do you think it would make sense to write transformations that would transform code from
```java
sb.append(a);
sb.append(b);
```
to
```java
sb.append(a).append(b);
```
?
(probably with better formatting)

MartinWitt:
> As a side note, do you think it would make sense to write transformations that would transform code from

Writing formatting transformation is a possibility, for sure. But I'm unsure if this field is already solved by spotless and other tools. Also, formatting is very opinionated and not as strict as bad smells. So, I skipped this until now.

PR #4921, opened 2022-09-26
fix: properly print type parameter reference in instanceof

SirYwell:
Since Java 16, it is allowed to have type parameters other than `<?>` in instanceof checks. This is allowed in cases where the generic type can be carried over from the type on the left side.

I added a test with three such examples, one with pattern matching (the main reason this feature was added), one normal instanceof check with a type parameter as type argument and one with String as type argument.

The current solution is to not enforce wildcard in the printer for compliance mode >= 16. I would personally prefer to not change the printing output there at all, but there are tests relying on it - namely this one:
https://github.com/INRIA/spoon/blob/59990fe39101fdb7fcb13d4ad6bbaf0085cb8509/src/test/java/spoon/test/template/TemplateTest.java#L1096
And it also might break existing code.

I'm not sure if that's the best solution. If you have any better ideas, please let me know.

### DIFF
--- a/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
+++ b/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
@@ -537,7 +537,7 @@ public class DefaultJavaPrettyPrinter implements CtVisitor, PrettyPrinter {
 		printer.writeSpace();
 		try (Writable _context = context.modify()) {
 			if (operator.getKind() == BinaryOperatorKind.INSTANCEOF) {
-				_context.forceWildcardGenerics(true);
+				_context.forceWildcardGenerics(env.getComplianceLevel() < 16);
 			}
 			scan(operator.getRightHandOperand());
 		}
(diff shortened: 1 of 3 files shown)


src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
```diff
@@ -537,7 +537,7 @@ public <T> void visitCtBinaryOperator(CtBinaryOperator<T> operator) {
 		printer.writeSpace();
 		try (Writable _context = context.modify()) {
 			if (operator.getKind() == BinaryOperatorKind.INSTANCEOF) {
-				_context.forceWildcardGenerics(true);
+				_context.forceWildcardGenerics(env.getComplianceLevel() < 16);
```
MartinWitt, line 540: Can you please move this in a method containing an explanation why we have this condition?

src/test/resources/patternmatching/InstanceofGenerics.java
```diff
@@ -0,0 +1,22 @@
...
+import java.util.List;
+import java.util.Set;
+
+class InstanceofGenerics {
+	<T> boolean check(Iterable<T> iterable) {
+		return iterable instanceof Set<T>;
+	}
+
+	<T> T first(Iterable<T> iterable) {
+		if (iterable instanceof List<T> list) {
```
MartinWitt, line 11: Could you add a test with `List<T<U>>` or something more complex?

MartinWitt (changes requested):
I add some inline comments other LGTM

MartinWitt (approved):

MartinWitt:
LGTM will merge tomorrow.

MartinWitt:
Thanks @SirYwell. Almost forget merging this PR. Please don't hesitate to ping me next time.

Issue #4924, opened 2022-09-27
Get rid of warnings in Spoon's build


slarse:
I think it's about time we clean up the compile output and actually start to care about warnings, or even better, treat warnings as errors and selectively suppress the ones that we don't care about.

I've been bothered by the warnings in the build output for a good while. All of the garbage warnings obscure "real" warnings, such as deprecations in libraries we depend upon.

I propose two things:

1. We get rid of the current warnings.
2. We treat warnings as errors so we can keep the build output clean, and actually act upon new warnings.

MartinWitt:
How about pinning this issue to the top during the hacktober fest? Should get some more views, and maybe someone is searching for easy PRs?

> We get rid of the current warnings.

If we feel risky, someone can try to fix many generic warnings with IntelliJ. There is a fix all button for inspections. This could save a lot of time adding `<?>` to raw types.

I-Al-Istannen:
And we potentially also want to use the magic tempdir, though that is not as critical.

slarse:
One question is how we verify that warnings ? I tried to write a script to check warnings before/after (with some simple `grep | wc -l` stuff), but it just didn't work as warnings that I fixed didn't reappear when I checked out to the master branch (does Maven cache build outputs somewhere else than in `target`??), didn't get the warnings to appear until I changed something in the concerned file and recompiled. Then they appeared.

I-Al-Istannen:
I actually can not reproduce that. If I `git stash` the numbers change:
```
❯ mvn clean package -DskipDepClean=true -Dmaven.test.skip=true 2>&1 | grep -c '[WARNING]'
236
❯ git stash
Arbeitsverzeichnis und Index-Status WIP on master: 0b498bfec refactor: fix bad smell Non-Protected-Constructor-in-Abstract-Class (#4853) gespeichert.
❯ mvn clean package -DskipDepClean=true -Dmaven.test.skip=true 2>&1 | grep -c '[WARNING]'
237
```

TheMarvelFan:
Hi can someone provide some guidance as to what I should do to solve this issue?

Ao-senXiong:
Hi @slarse , I have tried to work on this issue. I finished suppressing or removing warnings. We can add this code to pom to treat the warning as error. I think we should add suppress warnings to a lot of code and remove them gradually.  
```
<plugin>
        <groupId>org.apache.maven.plugins</groupId>
        <artifactId>maven-compiler-plugin</artifactId>
[...]

PR #4925, opened 2022-09-27
refactor: Fix warnings in CtTypeReferenceImpl

slarse:
#4924 

Most of the fixes are just suppressing of "intentional" warnings.

### DIFF
--- a/src/main/java/spoon/support/reflect/reference/CtTypeReferenceImpl.java
+++ b/src/main/java/spoon/support/reflect/reference/CtTypeReferenceImpl.java
@@ -114,7 +114,7 @@ public class CtTypeReferenceImpl<T> extends CtReferenceImpl implements CtTypeRef
 	}
 
 	@Override
-
+	@SuppressWarnings("deprecation")
 	public Class<T> getActualClass() {
 			return getPrimitiveType(this).orElseGet(this::findClass);
 	}
@@ -311,6 +311,7 @@ public class CtTypeReferenceImpl<T> extends CtReferenceImpl implements CtTypeRef
 	}
 
 	@Override
+	@SuppressWarnings("unchecked")
 	public <C extends CtActualTypeContainer> C setActualTypeArguments(List<? extends CtTypeReference<?>> actualTypeArguments) {
 		if (actualTypeArguments == null || actualTypeArguments.isEmpty()) {
 			this.actualTypeArguments = CtElementImpl.emptyList();
@@ -328,6 +329,7 @@ public class CtTypeReferenceImpl<T> extends CtReferenceImpl implements CtTypeRef
 	}
 
 	@Override
+	@SuppressWarnings("unchecked")
 	public <C extends CtTypeReference<T>> C setDeclaringType(CtTypeReference<?> declaringType) {
 		if (declaringType != null) {
 			declaringType.setParent(this);
@@ -338,6 +340,7 @@ public class CtTypeReferenceImpl<T> extends CtReferenceImpl implements CtTypeRef
 	}
 
 	@Override
+	@SuppressWarnings("unchecked")
 	public <C extends CtTypeReference<T>> C setPackage(CtPackageReference pack) {
 		if (pack != null) {
 			pack.setParent(this);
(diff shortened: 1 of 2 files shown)


SirYwell:
An alternative idea for the unchecked casts: Maybe it would make sense to add a method like
```java
protected final <T> T self() {
	return (T) this;
}
```
to `CtElementImpl` and replace all the unchecked casts by calls to that method? This way, other (unrelated) unchecked casts don't get supressed accidentally. Additionally, we could write a test to make sure this method is only used in return statements (and with a bit more work, we can likely also make sure it is only called in return statements of methods that should return `this`).

WDYT?

slarse:
@SirYwell Hmm, well on the one hand it's nice for the reason you mentioned. On the other hand, it would also hide something very nasty that we shouldn't be doing in the first place. The explicit '@SuppressWarnings` make it a bit more obvious that in that method, we're up to no good.

Perhaps we could make a helper that's slightly more of an eye-sore, just to make it clear that we really shouldn't be doing that thing. Something like this:

```java
/**
 * Make an unchecked cast of the provided object to the type of the expression it is used in.
 *
 * This should primarily be used to coerce <code>this</code> to the expected return type in methods that are generic
 * in the return type.
 * 
 * THIS IS VERY BAD, use only where legacy requires it to avoid having to litter classes with <code>SuppressWarnings("unchecked")</code>
 */
public static <T> T uncheckedCastToInferredReturnType(Object object) {
    return (T) object;
}
```

Something like that. It's obviously a horrendous helper mothod. Having it as a separate helper method, rather than providing it through inheritance, also makes it a bit less discoverable to future us (or other integrators who lack this context). It's less likely to be used in a spur of the moment "well that looks neat".

WDYT?

SirYwell:
> On the other hand, it would also hide something very nasty that we shouldn't be doing in the first place.

That's very true, and basically the reason why I suggested to introduce checks to only allow the method to be called in the specific places.
Sadly, we won't get rid of whatever dirty solution we decide on without breaking basically everything, so it's probably worth to invest some work into that.

I don't have a strong opinion on naming and where the method is actually declared. My idea behind inheritance was to avoid unintended usage (as we can pass everything to a method that takes an object), but I can also fully understand your reasoning. So, feel free to implement it the way you suggested.

To prevent misuse, I'd still go with some architecture test on top of it (if you don't want to include it here, we can do that separately I guess), and probably even create a list of the specific methods in which it is allowed (or whatever is suitable to prevent *new* methods from being implemented that way, without slipping through a code review) (?).

src/main/java/spoon/support/reflect/reference/CtTypeReferenceImpl.java
```diff
@@ -841,12 +842,12 @@ public boolean isShadow() {
...
 		getFactory().getEnvironment().getModelChangeListener().onObjectUpdate(this, IS_SHADOW, isShadow, this.isShadow);
 		this.isShadow = isShadow;
-		return (E) this;
+		return Unchecked.castToReturnType(this);
 	}
 
 	@Override
 	public CtTypeReference<T> clone() {
-		return (CtTypeReference<T>) super.clone();
+		return Unchecked.castToReturnType(super.clone());
```
slarse, line 850: @SirYwell This also seems like an appropriate place to use this, wdyt?

slarse:
@SirYwell I've pushed a new suggestion up. How about that (or something like that, feel free to come with suggestions), and then an architecture test to check that `Unchecked.castToReturnType` is only a) used on arguments `this` or `super.clone()`, and b) is not imported statically (so it's clearly used with the `Unchecked` "namespace").

There are any amount of variations to that strategy, this is just what I came up with.

SirYwell:
> @SirYwell I've pushed a new suggestion up. How about that (or something like that, feel free to come with suggestions)

I like that.

> and then an architecture test to check that `Unchecked.castToReturnType` is only a) used on arguments `this` or `super.clone()`, and b) is not imported statically (so it's clearly used with the `Unchecked` "namespace").

That sounds good. I'd additionally only allow it in return statements, I think that works for our use case (?) and is easy to check.

src/main/java/spoon/support/reflect/reference/CtTypeReferenceImpl.java
SirYwell: Yep, that makes sense aswell.

slarse:
Closing this for now, perhaps I'll revisit it later.

Issue #4269, opened 2021-11-05
Add support for the Gradle build scan using the Maven extension


cesarsotovalero:
The [Gradle Build Scan](https://scans.gradle.com/#maven) is a free service that provides useful insights regarding how the project build (e.g., dependencies, plugins, tasks performance, tests, infrastructure used, etc.)
Once configured, the information is accessible via a randomly generated link, printed in the logs at the end of the build. 
This information can be valuable for the community to better understand how Spoon is build.
For example, here's a view of the service after executing `mvn compile` on Spoon:

![image](https://user-images.githubusercontent.com/13948717/140495220-3e82547f-60af-4910-8e43-1f2721fdc088.png)

This service can be easily configured integrated as part of the Spoon build using the official [Maven Extension](https://docs.gradle.com/enterprise/maven-extension/#build_scans_captured_information).

My question is the following: **Is this valuable for the Spoon project?**

monperrus:
thanks for the constructive proposal.

:+1:  for installing the Maven extension

slarse:
Can we disable this for local builds? I a) don't like data being uploaded somewhere when I do a build locally, and b) just had failures running with this extension enabled (which is the only reason I figured out I was uploading stuff in the first place).

Ping @MartinWitt @SirYwell @I-Al-Istannen Thoughts?

MartinWitt:
If you find an option for this, sure go ahead. Tbh I never looked at build scans and if I need one I create one local. Build scans contain no information I can get from my IDE.

I-Al-Istannen:
I am in the same boat. I have never looked at one, but I have also never tried to optimize spoon's maven build. My local builds are also more often than not quite weird with some plugins ripped out to speed things up.

slarse:
I think I wasn't entirely clear, I mostly meant what your opinions were on any user cloning this repo and build Spoon unknowingly upload data to some third party, and in that process also unknowingly accept their [TOS](https://gradle.com/legal/terms-of-use/):
https://github.com/INRIA/spoon/blob/38b60dbbe82be7b3dfa7dfe683444286d0a9172f/.mvn/gradle-enterprise.xml#L3-L7

I find that pretty concerning, and I have no idea what we're using this for. As a quick fix I will disable uploads

PR #4927, opened 2022-09-28
chore: Disable automatic Gradle Dashboard upload

slarse:
See #4269 for discussion

See [this part of the docs](https://docs.gradle.com/enterprise/maven-extension/#publishing_on_demand) for on-demand publishing.

### DIFF
--- a/.mvn/gradle-enterprise.xml
+++ b/.mvn/gradle-enterprise.xml
@@ -1,6 +1,6 @@
 <gradleEnterprise>
   <buildScan>
-    <publish>ALWAYS</publish>
+    <publish>ON_DEMAND</publish>
     <termsOfService>
       <url>https://gradle.com/terms-of-service</url>
       <accept>true</accept>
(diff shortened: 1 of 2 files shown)


slarse:
@monperrus Regarding static analysis noise, is that :point_up:  going to show up on every PR now?

monperrus:
No! See https://github.com/INRIA/spoon/pull/4928

I-Al-Istannen (commented):
If I knew how the build scan should be used, I would be better able to tell whether skipping tests is a problem here. If this PR just wants to be a change to on-demand, it should probably edit the workflow file too.

If you don't think it's worthwhile that's fine with me too.

I-Al-Istannen:
Dito. If I understood [their advertisement](https://blog.gradle.org/introducing-build-scans) correctly it is for helping maintainers debug why builds for clients (or themselves) are slow or failing. Instead of looking at your log you can look at their website version of the log and users can share (deep-)links to the reports. This might be easier than having users upload the full maven build log, which might even miss information. 

So far, I haven't really seen people complain that their *spoon build* is failing — they normally just use it.  
If no integrator has ever used it either (Current count: 4/6) then maybe just deleting it in the first place really is the better solution.

MartinWitt (approved):

MartinWitt:
I merged this for now, and we can discuss the removal of the plugin in the issue where we already started it.

PR #4930, opened 2022-09-29
doc: add submodule policy in CONTRIBUTING.md

monperrus:


CONTRIBUTING.md
```diff
@@ -83,3 +83,20 @@ The public API is composed of all public classes and methods, except those for w
...
+
+1) the functionality would bloat the spoon-core binary too much
+
+OR
+
+2) the functionality is too experimental / too unstable to meet the high quality standard of spoon-core
+
+OR
+
+3) the integrator team welcomes the contribution, but cannot commit to maintaining it.
```
slarse, line 102:
```suggestion
3) the integrator team welcomes the contribution, but cannot commit to maintaining it.

Submodules are **not publlished to Maven central**, and so need to be built from source.
```

slarse (commented):
I think it'd be appropriate to at least mention that we don't publish submodules?

monperrus:
agree. merged suggestion.

slarse (approved):
This is good for me.

@SirYwell @I-Al-Istannen @MartinWitt anything to add? I think the CONTRIBUTING.md file is pretty important so I don't want to merge without you getting the chance to say your piece.

MartinWitt:
I thought about adding a hint for [jitpack.io](jitpack.io), but sadly this is pretty unstable and no one should use this for production code. So lgtm you can merge.

MartinWitt (approved):

I-Al-Istannen:
I have found jitpack to be quite reliable for other projects, but users can probably find something nice on their own here.

I-Al-Istannen (approved):

slarse:
I also think that Jitpack is a little bit beyond the scope of these docs (although I concur with @I-Al-Istannen in that I've found it to be reliable when I've used it).

Merging this now so we get it in, we can always update it later!

slarse:
Thanks @monperrus

PR #4931, opened 2022-09-29
review: test: add two test runs for newest and ea jdk

MartinWitt:
This PR adds two new check runs in the CI. One for newest stable JDK(19) and early access JDK(20-ea). The goal is to know earlier if something will break in the future. Furthermore, a goal of this PR is to finally send a mail to [quality-outreach-programm](https://wiki.openjdk.org/display/quality/Quality+Outreach#QualityOutreach-HowtojointheQualityOutreachProgram) from openjdk. One test case was simply broken in future JDKs because we test against internal methods, and they got changed over the time.
I'm unsure if we should allow failure for these test runs. Allowing failure means we simply move the problem 6/12 months in the future.

1) Pinning of action hashes should be done by renovate after the merge, or shall I search for the hashes?
2) This checks need to be updated every 6 month to new versions, which shouldn't be too high an effort.

Edit: It seems like JDK-EA is around 50% slower than a released JDK 10 vs. 7 min runtime. This should still be fine.

### DIFF
--- a/src/test/java/spoon/test/model/TypeTest.java
+++ b/src/test/java/spoon/test/model/TypeTest.java
@@ -7,6 +7,8 @@
  */
 package spoon.test.model;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.DisabledForJreRange;
+import org.junit.jupiter.api.condition.JRE;
 import spoon.Launcher;
 import spoon.compiler.ModelBuildingException;
 import spoon.reflect.CtModel;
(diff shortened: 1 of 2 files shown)


slarse:
> Pinning of action hashes should be done by renovate after the merge, or shall I search for the hashes?

When I first introduced pinning of action hashes, I had _very_ sporadic success in letting Renovate find and pin stuff on its own. For the most part, I had to pin manually and then updates would work. I'm in favor of pinning manually immediately as it's not much effort and it's a but uncomfortable to have GHA versions only locked by tags in the master branch, especially if Renovate doesn't fix it fast.

> This checks need to be updated every 6 month to new versions, which shouldn't be too high an effort.

Agreed that it's not much effort, but how do we remember to do it? :)

MartinWitt:
I will later add the pinned hashes.

> Agreed that it's not much effort, but how do we remember to do it? :)

We could use a bot like https://github.com/probot/reminders#probot-reminders or simple belief in ourselves. 

> Is there any reason we can't just add these to the "normal" build matrix?

The exclude list grows as we only run them on Ubuntu. Furthermore, I thought marking them as allowed failure is easier without the matrix.

MartinWitt:
> By allowed failure, do you mean not marking it as a required check for GitHub PRs?

No, I thought it could be handy to mark, e.g., the java 20-EA job as neutral check if needed. I never really tested against an early access version of the JDK, and maybe it is not production ready.

slarse:
> No, I thought it could be handy to mark, e.g., the java 20-EA job as neutral check if needed. I never really tested against an early access version of the JDK, and maybe it is not production ready.

I'm not entirely following here, when you say "neutral check", what do you mean? If you mean a normal non-required GitHub PR check, then there's no benefit that I see to splitting it out from the normal build matrix. It will appear as a standalone check where you select required checks regardless.

The build is a little bit complex, and it's easy to get things wrong. For example, your new builds don't daily expiration of the cache working correctly. The key used for the cache looks like this:

```
Linux--deb318a75988295fefd49dba4e251c96035c13a3bfe8d149715e46f350e2cb9b
```

But it should look like this:

```
Linux-20220929-deb318a75988295fefd49dba4e251c96035c13a3bfe8d149715e46f350e2cb9b
```

The reason it's incorrect is that the `get-date` step isn't defined in the new jobs. But that's hard to spot.

Would you mind just including these builds in the existing build matrix? Or is there something I'm not seeing here?

src/test/java/spoon/test/model/TypeTest.java
```diff
@@ -61,24 +63,23 @@ public void testGetDifferentElements() throws Exception {
 	}
 
 	@Test
+	@DisabledForJreRange(min = JRE.JAVA_14)
+	/*
+	* This annotation is needed because in java.lang.object the method registerNative was removed.
+	* See https://bugs.openjdk.java.net/browse/JDK-8232801 for details.
+	*/
```
slarse, line 70: Should this be considered a bug (for which we need a bug ticket)?

slarse (approved):
Thanks @MartinWitt, LGTM.

See the comment as well, it relates to your opening post.

src/test/java/spoon/test/model/TypeTest.java
MartinWitt: No, we simply tested against a real internal method `private native registerNative()` which was a relict of the sun java time. We should stop testing against these internals which are not really stable.

> In this case we are deleting a private method from the Object class. There are no specification implications here as it is a private method. However there is, as described, a behavioural change, because the now deleted method will no longer be considered during method resolution and so code that would currently generate an error, no longer will.

slarse: Ah, I misunderstood, I thought it was the production code that used an internal method.

PR #4937, opened 2022-09-30
review: style: convert set to enumset

MartinWitt:
Enumset for enum values is strictly better than set, and this fixes one of many locations with this bad smell.


slarse (approved):
LGTM, will merge.

slarse:
@MartinWitt "style" is not a prefix I recognize (and it's not in `CONTRIBUTING.md`), is that something we're using now? I would otherwise have put this under "refactor".

MartinWitt:
I started using style because refactoring sounds like way more changes than this small single line bad smell removals. But I can change back to `refactor:`

slarse:
I will merge with `refactor` because this is absolutely a refactoring. A tiny one, to be sure, but it definitely fits the definition.

PR #4938, opened 2022-09-30
review: chore: move plugin to parent

MartinWitt:
As discussed in #4911 I moved the plugin to the parent pom.


slarse (approved):
LGTM. The thing that changes here is that a bunch of ignores are removed from spoon-visualization, but it appears as if they're not needed. So that's good, I suppose :)

Issue #4342, opened 2021-12-09
[Bug] Annotations are not added to the correct in the model


SirYwell:
**Describe the bug**

I stumbled across a few issues with annotations:

1. Annotations with ElementType.TYPE_USE are added to fields and methods, rather than to the type references. From my understanding, this is a [JDT bug](https://bugs.eclipse.org/bugs/show_bug.cgi?id=544288).
2. Annotations on array types are fully ignored. E.g. `Object @Nonnull [] @Nullable []` becomes `Object[][]`. The information seems to be provided correctly by JDT, but I wasn't able to figure out how to integrate it in the spoon code properly.
3. Annotations in front of method references, e.g. `Function<?, ?> f = @Nonnull Object::toString;` is added to the model but it isn't printed anymore.

I started looking into 2. and 3. but that was over my head for now. Maybe someone else knows a little bit more about the code that needs to be changed for that.

**To Reproduce**

Use the code fragements from above in your code and inspect and/or print it.

Input:
```java
/*Code being processed by spoon and leading to the bug.*/

// The definition of the used annotations should look like that

@Target({ElementType.TYPE_USE})
@Retention(RetentionPolicy.RUNTIME)
@interface Nullable { }

@Target(ElementType.TYPE_USE)
@Retention(RetentionPolicy.RUNTIME)
@interface Nonnull { }

// that's the class we want to inspect/print

class M {
	private @Nullable Object @Nonnull [] field;
	public void method(final @Nullable Object @Nonnull [] @Nullable [] value) {
		Function<Object, String> toString = @Nonnull Object::toString;
	}
}
```

Processing with Spoon:
```java
/*Code calling spoon on the input and triggering the bug.*/
```

Output:
```java
/*Output code or stacktrace if unexpected Exception is raised*/
class M {
    @Nullable
    private Object[] field;

    public void method(@Nullable
    final Object[][] value) {
        Function<Object, String> toString = Object::toString;
    }
}
```

Note that you can also open a pull request reproducing the issue and reference that, instead of writing additional information here.

**Operating system, JDK and Spoon version used**

* OS: Windows 10
* JDK: Temurin 17.0.1
* Spoon version: master branch as of https://github.com/INRIA/spoon/commit/12cc1a530e5801829ec4e5f674135f704040f419

slarse:
Ouch, that output is pretty crazy.

[...]

PR #4939, opened 2022-09-30
test: add tests for type annotations on extends/implements

SirYwell:
This is the first part emerging from my work on #4342. It's meant to discuss the overall design of the tests. I included tests for the first two scenarios, as the third one already requires a bug fix.

The design:

- I'm following the structure of the JLS where types are used to cover all the cases where type annotations can appear.
- I'm using nested classes/nested tests to clarify the structure given by the JLS.
- I'm using display names to pretty much reuse the texts from the JLS to clarify the intentions.
- The tests test both the model and the printer.

The follow-up tests will be submitted in batches, depending on whether they require bug fixes or not.


MartinWitt:
Could you move the test files to the resources folder? Otherwise, these tests look amazing.

MartinWitt (approved):

MartinWitt:
Thanks @SirYwell

PR #4940, opened 2022-09-30
review: style: change set to enumset

MartinWitt:
Enumset for enum values is strictly better than set, and this fixes one of many locations with this bad smell.


slarse (approved):
Oh my, that return statement is a mouthful. But I suppose what you did here makes it ever so slightly better.

PR #4947, opened 2022-10-06
test: clean up MainTest#testGenericContract

SirYwell:
This test seems to be outdated. It used compliance level 8 on spoon.
Additionally, there are files which are filtered as they caused bugs before, but it seems like that's not the case for most of them anymore.
I also removed the manual launcher args as it's cleaner to do that in code directly (and most options are not needed).

### DIFF
--- a/src/test/java/spoon/test/main/MainTest.java
+++ b/src/test/java/spoon/test/main/MainTest.java
@@ -27,6 +27,8 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.stream.Stream;
+
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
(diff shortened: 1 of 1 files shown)


src/test/java/spoon/test/main/MainTest.java
```diff
@@ -38,63 +38,28 @@ public void testGenericContract() throws IOException {
...
-		}
-		String systemClassPath = classpath.substring(0, classpath.length() - 1);
-
 		launcher = new Launcher();
-
-		launcher.setArgs(new String[] {
-				"-o", "target/spooned",
-				"--destination","target/spooned-build",
-				"--source-classpath", systemClassPath,
-				"--compile", // compiling Spoon code itself on the fly
```
MartinWitt, line 59: Did we lose this flag?

src/test/java/spoon/test/main/MainTest.java
```diff
@@ -38,63 +38,28 @@ public void testGenericContract() throws IOException {
...
-				// bug 4: remove the filter to trigger it
-				.filter(path -> !filePathContains(path, "lambda/testclasses/Bar"))
-
-				// bug 5: remove the filter to trigger it
-				.filter(path -> !filePathContains(path, "LambdaRxJava"))
-
-				// bug 6: remove the filter to trigger it
-				.filter(path -> !filePathContains(path, "Tapas"))
-
 				.forEach(x -> {
```
MartinWitt, line 63: Please remove the curly brackets for this lambda.

MartinWitt (changes requested):
2 open points otherwise LGTM

src/test/java/spoon/test/main/MainTest.java
SirYwell: They really weren't necessary, we only want to build the model and verify the classes.

src/test/java/spoon/test/main/MainTest.java
SirYwell: Replaced and also wrapped the stream with a TWR as required by the docs

MartinWitt:
Thanks @SirYwell

PR #4950, opened 2022-10-11
chore: Specify version of default plugins explicitly

I-Al-Istannen:
We used some truly ancient versions of a few default plugins and renovate did not pick them up. Now we specify them explicitly and renovate will hopefully keep them up to date in the future.

It will be interesting to see whether CI accepts this PR.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -227,7 +227,6 @@
         <plugin>
             <groupId>org.apache.maven.plugins</groupId>
             <artifactId>maven-assembly-plugin</artifactId>
-            <version>3.4.2</version>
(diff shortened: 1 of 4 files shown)


I-Al-Istannen:
Color me surprised, but this seems to have been accepted. I think this could unblock one or more of the renovate PRs which are currently failing.

MartinWitt (approved):

slarse (approved):
:+1: on actually getting around to doing this :)

MartinWitt:
Thanks @I-Al-Istannen

PR #4951, opened 2022-10-11
refactor: use modifier masks instead of additional checks

SirYwell:
Instead of manually checking if a modifier is applicable to a `CtModifiable`, we can just use the masks provided by Java to only have the allowed bits set.


MartinWitt (approved):

MartinWitt:
Thanks @SirYwell

PR #4953, opened 2022-10-12
refactor: Object#getClass will never be <primitive>.class

SirYwell:
The checks are (assumingly(?)) meant to avoid cost of toString + parse when the object is already of the expected type. As we'll never have primitive instances, the class comparison will never be true.

Not sure if this is rather a bug fix then a refactoring, but as there is no case in which the *result* will differ (besides checking the actual instance), it's hard to write an appropriate test so I'd rather label it as refactoring.


MartinWitt (approved):

MartinWitt:
thanks @SirYwell

PR #4954, opened 2022-10-12
feat: Add `CtTry.addCatcherAt`

algomaster99:
Reference: #3884 

The order of catchers is important because the thrown exception is matched with the parameters of the catchers in the order they appear, hence it is semantically important.

### DIFF
--- a/src/main/java/spoon/reflect/code/CtTry.java
+++ b/src/main/java/spoon/reflect/code/CtTry.java
@@ -48,6 +48,12 @@ public interface CtTry extends CtStatement, TemplateParameter<Void>, CtBodyHolde
 	<T extends CtTry> T addCatcher(CtCatch catcher);
 
 	/**
+	 * Adds a catch block at the specified position in the <code>try</code> statement.
+	 */
+	@PropertySetter(role = CATCH)
+	<T extends CtTry> T addCatcherAt(int index, CtCatch catcher);
+
+	/**
 	 * Removes a catch block.
 	 */
 	@PropertySetter(role = CATCH)
--- a/src/main/java/spoon/reflect/code/CtTry.java
+++ b/src/main/java/spoon/reflect/code/CtTry.java
@@ -49,9 +49,14 @@ public interface CtTry extends CtStatement, TemplateParameter<Void>, CtBodyHolde
 
 	/**
 	 * Adds a catch block at the specified position in the <code>try</code> statement.
+	 *
+	 * @param <T> the type of the try statement
+	 * @param position the position at which the <code>catcher</code> is to be inserted
+	 * @param catcher the catch statement to be inserted
+	 * @return this try statement
 	 */
 	@PropertySetter(role = CATCH)
-	<T extends CtTry> T addCatcherAt(int index, CtCatch catcher);
+	<T extends CtTry> T addCatcherAt(int position, CtCatch catcher);
 
 	/**
 	 * Removes a catch block.
--- a/src/main/java/spoon/reflect/code/CtTry.java
+++ b/src/main/java/spoon/reflect/code/CtTry.java
@@ -51,14 +51,13 @@ public interface CtTry extends CtStatement, TemplateParameter<Void>, CtBodyHolde
 	 * Adds a catch block at the specified position in the <code>try</code> statement.
 	 * Behaves similarly to {@link java.util.List#add(int, Object)}.
 	 *
-	 * @param <T> the type of the try statement
 	 * @param position the position at which the <code>catcher</code> is to be inserted
 	 * @param catcher the catch statement to be inserted
 	 * @return this try statement
 	 * @throws IndexOutOfBoundsException if the position is out of range (position < 0 || position > number of catchers)
 	 */
 	@PropertySetter(role = CATCH)
-	<T extends CtTry> T addCatcherAt(int position, CtCatch catcher);
+	CtTry addCatcherAt(int position, CtCatch catcher);
 
 	/**
 	 * Removes a catch block.
--- a/src/main/java/spoon/reflect/code/CtTry.java
+++ b/src/main/java/spoon/reflect/code/CtTry.java
@@ -49,11 +49,13 @@ public interface CtTry extends CtStatement, TemplateParameter<Void>, CtBodyHolde
 
 	/**
 	 * Adds a catch block at the specified position in the <code>try</code> statement.
+	 * Behaves similarly to {@link java.util.List#add(int, Object)}.
 	 *
 	 * @param <T> the type of the try statement
 	 * @param position the position at which the <code>catcher</code> is to be inserted
 	 * @param catcher the catch statement to be inserted
 	 * @return this try statement
+	 * @throws IndexOutOfBoundsException if the position is out of range (position < 0 || position > number of catchers)
 	 */
 	@PropertySetter(role = CATCH)
 	<T extends CtTry> T addCatcherAt(int position, CtCatch catcher);
--- a/src/test/java/spoon/test/trycatch/TryCatchTest.java
+++ b/src/test/java/spoon/test/trycatch/TryCatchTest.java
@@ -498,7 +498,7 @@ public class TryCatchTest {
 
 			// act & assert
 			assertThrows(IndexOutOfBoundsException.class,
-					() -> tryStatement.addCatcherAt(2, catcherAtWrongPosition));
+					() -> tryStatement.addCatcherAt(1, catcherAtWrongPosition));
 		}
 	}
 }
--- a/src/test/java/spoon/test/trycatch/TryCatchTest.java
+++ b/src/test/java/spoon/test/trycatch/TryCatchTest.java
@@ -26,6 +26,7 @@ import java.util.List;
 import java.util.Set;
 
 import org.apache.commons.lang3.StringUtils;
+import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 import spoon.Launcher;
@@ -36,7 +37,6 @@ import spoon.reflect.code.CtCatchVariable;
 import spoon.reflect.code.CtResource;
 import spoon.reflect.code.CtTry;
 import spoon.reflect.code.CtTryWithResource;
-import spoon.reflect.code.CtVariableRead;
 import spoon.reflect.declaration.CtClass;
 import spoon.reflect.declaration.CtMethod;
 import spoon.reflect.declaration.CtVariable;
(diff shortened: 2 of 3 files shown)


algomaster99:
@slarse please have a look at this when you get time.

src/test/java/spoon/test/trycatch/TryCatchTest.java
```diff
@@ -442,4 +444,60 @@ public void testTryWithVariableAsResource() {
...
+			tryStatement.addCatcherAt(1, second);
+
+			// assert
+			assertThat(tryStatement.getCatchers(), contains(first, second, third));
+		}
+
+		@Test
+		void throwsOutOfBoundsException_whenPositionIsOutOfBounds() {
+			// contract: `addCatcherAt` should throw an out-of-bounds exception when the specified position is out of
+			// bounds of the catcher collection
```
SirYwell, line 487: That's not exactly correct, as `collection.size()` *is* out of bounds, but only `collection.size() + 1` and above (and < 0) throws an exception. More important than the contract here, this should probably also documented in the javadoc of the new method. Maybe we can just say that the behavior corresponds to `List#add(int, E)`?

algomaster99: > Maybe we can just say that the behavior corresponds to List#add(int, E)

I think this should be sufficient. They explicitly say [here](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/List.html#add(int,E)) when `IndexOutOfBoundsException` is thrown.

src/main/java/spoon/reflect/code/CtTry.java
```diff
@@ -47,6 +47,19 @@ public interface CtTry extends CtStatement, TemplateParameter<Void>, CtBodyHolde
...
+	 * Behaves similarly to {@link java.util.List#add(int, Object)}.
+	 *
+	 * @param <T> the type of the try statement
+	 * @param position the position at which the <code>catcher</code> is to be inserted
+	 * @param catcher the catch statement to be inserted
+	 * @return this try statement
+	 * @throws IndexOutOfBoundsException if the position is out of range (position < 0 || position > number of catchers)
+	 */
+	@PropertySetter(role = CATCH)
+	<T extends CtTry> T addCatcherAt(int position, CtCatch catcher);
```
algomaster99, line 61: I am using `position` to refer to `index` in the `catchers` collection because it has been used for all analogous methods. See https://github.com/INRIA/spoon/blob/76ffd3353a535b0ce6edf0bf961a05236a40d3a1/src/main/java/spoon/reflect/code/CtAbstractSwitch.java#L72 as an example.

src/test/java/spoon/test/trycatch/TryCatchTest.java
SirYwell: Yep, that's good I guess.

src/test/java/spoon/test/trycatch/TryCatchTest.java
```diff
@@ -442,4 +444,60 @@ public void testTryWithVariableAsResource() {
...
+			second.setParameter(factory
+					.createCatchVariable()
+					.setMultiTypes(List.of(
+							factory.Type().createReference(ExceptionInInitializerError.class),
+							factory.Type().createReference(NoSuchFieldException.class))));
+
+			CtCatch third = factory.createCatch();
+			third.setParameter(factory
+					.createCatchVariable()
+					.setType(factory.Type().createReference(NoSuchMethodException.class)));
```
slarse, line 473: This is kind of a mouthful to read, and there is as far as I can tell unnecessary complexity in having a catcher that catches multiple types (I can't see how this influences insertion?).

I'd suggest something like this:

```suggestion
			CtCatch first = createCatch(factory, IOException.class);
			CtCatch second = createCatch(factory, ExceptionInInitializerError.class);
			CtCatch third = createCatch(factory, NoSuchMethodException.class);
```

With a helper like so:

```java
		private <T> CtCatch createCatch(Factory factory, Class<T> typeToCatch) {
			CtTypeReference<T> typeReference = factory.Type().createReference(typeToCatch);
			return factory.createCatch().setParameter(
				factory.createCatchVariable().setType(typeReference)
			);
		}
```

src/test/java/spoon/test/trycatch/TryCatchTest.java
```diff
@@ -442,4 +444,60 @@ public void testTryWithVariableAsResource() {
 		tryStmt.removeResource(ctResource);
 
 	}
+
+	@Nested
+	class AddCatcherAt {
+		@Test
+		void addsCatcherAtTheSpecifiedPosition() {
```
slarse, line 451: Note: This test really tests three separate cases in one go:

1. Inserting into an empty list.
2. Inserting at the beginning of a non-empty list.
3. Inserting at a position that is neither first nor last.

Given that this functionality is neither complicated nor critical to Spoon, that's probably fine. But it always bears thinking about how much is tested in any single test. The fact that we never insert at the end of a non-empty list also isn't a big deal considering what the implementation looks like.

src/test/java/spoon/test/trycatch/TryCatchTest.java
```diff
@@ -442,4 +444,60 @@ public void testTryWithVariableAsResource() {
...
+			// bounds of the catcher collection
+
+			// arrange
+			Factory factory = createFactory();
+
+			CtTry tryStatement = factory.createTry();
+			CtCatch catcherAtWrongPosition = factory.createCatch();
+			catcherAtWrongPosition.setParameter(factory
+					.createCatchVariable()
+					.setType(factory.Type().createReference(Exception.class)));
```
slarse, line 496: Also use the proposed helper here :)

src/test/java/spoon/test/trycatch/TryCatchTest.java
```diff
@@ -442,4 +444,60 @@ public void testTryWithVariableAsResource() {
...
+
+			// act & assert
+			assertThrows(IndexOutOfBoundsException.class,
+					() -> tryStatement.addCatcherAt(2, catcherAtWrongPosition));
```
slarse, line 500: If you test only one negative case, then experience tells me that it's best to test the boundary. Boundaries are highly prone to errors (think off-by-one errors).

```suggestion
					() -> tryStatement.addCatcherAt(1, catcherAtWrongPosition));
```

slarse (changes requested):
Some notes and some stuff I think we should change :)

src/test/java/spoon/test/trycatch/TryCatchTest.java
```diff
@@ -442,4 +444,60 @@ public void testTryWithVariableAsResource() {
...
+
+			CtCatch third = factory.createCatch();
+			third.setParameter(factory
+					.createCatchVariable()
+					.setType(factory.Type().createReference(NoSuchMethodException.class)));
+
+			// act
+			tryStatement.addCatcherAt(0, third);
+			tryStatement.addCatcherAt(0, first);
+			tryStatement.addCatcherAt(1, second);
```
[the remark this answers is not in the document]
algomaster99: Yes. This is a better idea. It also eliminates the "unused return" warning.

src/main/java/spoon/reflect/code/CtTry.java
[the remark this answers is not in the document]
algomaster99: That is indeed confusing, but why does it throw the class cast exception? `CtTryWithResource` extends from `CtTry` so it should be valid during runtime as well. Is it because type erasure replaces `<T extends CtTry>` to `CtTry`? I do not understand what exactly is happening at runtime.

src/test/java/spoon/test/trycatch/TryCatchTest.java
algomaster99: > createCatch(Factory, Class<T> )

Any reason why we pass the same instance of `Factory` and not create a new one whenever we want to call `createCatch`?

src/test/java/spoon/test/trycatch/TryCatchTest.java
algomaster99: > But it always bears thinking about how much is tested in any single test.

Right. Ideally, we should test only one case per _unit_ test. I will keep this in mind. We can skip refactoring for this feature because, like you said, it is not that critical and is rather a very simple feature.

> The fact that we never insert at the end of a non-empty list also isn't a big deal considering what the implementation looks like.

Tests of `addCatcher` would already test that so we do not need to think about them I guess.

algomaster99:
I pushed an empty commit to give you co-authorship, but now I realise that I have accepted some of your suggestions so you will be a co-author anyway if this PR gets squashed and merged :sweat_smile:

src/test/java/spoon/test/trycatch/TryCatchTest.java
```diff
@@ -442,4 +444,52 @@ public void testTryWithVariableAsResource() {
...
+
+			CtTry tryStatement = factory.createTry();
+			CtCatch catcherAtWrongPosition = createCatch(factory, Exception.class);
+
+			// act & assert
+			assertThrows(IndexOutOfBoundsException.class,
+					() -> tryStatement.addCatcherAt(1, catcherAtWrongPosition));
+		}
+
+		private <T extends Throwable> CtCatch createCatch(Factory factory, Class<T> typeToCatch) {
```
algomaster99, line 488: I made it type stricter by replacing `T` with `<T extends Throwable>`

src/main/java/spoon/reflect/code/CtTry.java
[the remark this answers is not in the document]
slarse: > CtTryWithResource extends from CtTry so it should be valid during runtime as well.

Yes, but the actual dynamic type of the `tryStatement` variable is `CtTryImpl`, which does not implement `CtTryWithResource`. So, there's an exception when the runtime tries to cast to the latter.

I wrote a really elaborate explanation for this, and then the page refreshed and I lost it :D. I'll give it a go again tomorrow, but the very short explanation is that the problems occur with the unchecked cast in the return statement (i.e. `(T)`). Basically, it's _not_ a cast at all, it's just telling the compiler "trust me mate, this is fine".

The actual cast here occurs when we assign to the variable with type `CtTryWithResource` after the method has returned a value. That's where the crash occurs. Java is hailed as a statically typed language, but it's actually very much a dynamically typed language as well, and implicit type checks occur all the time wherever generics are involved.

I'll try to circle back around to the rather nasty ramifications this stuff can have: a confusing `ClassCastException` is just the tip of the iceberg.

I spent way too much time on writing (and losing) the explanation here so I'll have to get back to the actual PR at a later time :)

slarse:
I think it will be a couple of days before I can return to this, I was away during the weekend which ate up most of what would normally be my OSS time :)

algomaster99:
Understandable. Take your time. Not very urgent. :)

src/test/java/spoon/test/trycatch/TryCatchTest.java
slarse: > Tests of addCatcher would already test that so we do not need to think about them I guess.

Precisely, "considering what the implementation looks like" :). White box testing has merits in that we can omit redundant testing, but it has drawbacks to match (e.g. creating tests based on the implementation rather than the specification).

src/test/java/spoon/test/trycatch/TryCatchTest.java
slarse: :+1:

slarse (approved):
Thanks @algomaster99 , sorry about the delay! Turned out I was away this weekend as well :s

Issue #3879, opened 2021-04-13
Jenkins continuous integration (CI) jobs and configuration


slarse:
The Jenkins jobs running [over here](https://ci.inria.fr/sos/) test the latest snapshot of Spoon on a variety of projects, in a variety of ways. There are two general kinds of jobs:

* For projects that use Spoon
    - Inject the latest Spoon snapshot and run project's test suite
    - Uses the new [inject_spoon_snapshot.py script](https://spoon.gforge.inria.fr/jenkins/inject_spoon_snapshot.py)
* For projects that don't use Spoon
    - Parse and pretty-print the project with spoon-maven-plugin using the latest Spoon snapshot
    - Then run the project's test suite

There's a bunch of tiny details surrounding this as well. In order to facilitate for someone else to manage Jenkins (or just myself next week when I've forgotten all about this), we should 1) streamline the jobs to make it very easy to add a new one, and 2) properly document how everything works.

monperrus:
FTR, Jenkins nodes sos-builder01 (aka Ubuntu 16) now has Java 11 under '/usr/lib/jvm/java-11-openjdk-amd64/'

```
$ /usr/lib/jvm/java-11-openjdk-amd64/bin/java -version
Picked up JAVA_TOOL_OPTIONS: -Dfile.encoding=UTF8
openjdk version "11.0.2" 2019-01-15
OpenJDK Runtime Environment 18.9 (build 11.0.2+9)
OpenJDK 64-Bit Server VM 18.9 (build 11.0.2+9, mixed mode)

```

Installed from https://download.java.net/java/GA/jdk11/9/GPL/openjdk-11.0.2_linux-x64_bin.tar.gz

monperrus:
FYI, https://ci.inria.fr/sos/job/Spoon-Snapshot-Deployer/ is now configured with Java 11

Ref: https://github.com/SpoonLabs/spoon-ci-config/commit/e14f7f709d4f7706375140a43deb69cb8355141e

monperrus:
@seintur the full disk on `sos-builde02` issue is solved, there were hundreds of temp dirs in /tmp/.

For some reasons, `df` reported incorrect free space.

monperrus:
FYI, removed outdated certificate `DST_Root_CA_X3.pem` (Digital Signature Trust Co., CN = DST Root CA X3) on sos-builder01, which resulted in failing to curl

```
curl https://spoon.gforge.inria.fr/jenkins/build.sh
[...]

PR #4955, opened 2022-10-12
chore: add workflow for creating and storing SBOM

siberov:
This new workflow creates a software bill of materials (SBOM) formatted as an in-toto attestation. This covers the content of the source tree, as well as the compiled files. The SBOM is pushed to a remote repository under the chains-project organisation for long-term storage.

**NOTE:** To work, this requires that a secret called `SBOM_DEPLOY_SSH` is added to the Spoon repo. It must contain the private part of an SSH key, where the public key is added as a deploy key to the [chains-project/sbom-files](https://github.com/chains-project/sbom-files) repository.

This should not be merged before the note above has been addressed.

### DIFF
--- a/.github/workflows/sbom.yml
+++ b/.github/workflows/sbom.yml
@@ -6,9 +6,6 @@
 
 name: sbom
 on:
-  pull_request:
-    branches:
-      - master
   push:
     branches:
       - master
--- a/.github/workflows/sbom.yml
+++ b/.github/workflows/sbom.yml
@@ -66,8 +66,8 @@ jobs:
           git config --global user.name "${{ env.GIT_USER }}"
           git config --global user.email "${{ env.GIT_EMAIL }}"
           git clone $REMOTE_REPO_URL
-          mkdir -p sbom-files/spoon
-          cp build.provenance sbom-files/spoon/$(date +%s)-${{ github.sha }}.provenance
+          mkdir -p sbom-files/spoon/slsa
+          cp build.provenance sbom-files/spoon/slsa/$(date +%s)-${{ github.sha }}.provenance
           cd sbom-files
           git add .
           git commit -m "Spoon: ${{ github.sha }}"
--- a/.github/workflows/sbom.yml
+++ b/.github/workflows/sbom.yml
@@ -1,0 +1,74 @@
+# Workflow for generating a software bill of materials (SBOM).
+#
+# Currently, this uses the slsa-framework/github-actions-demo action to
+# generate an in-toto attestation. This is the pushed to a remote repository
+# for storage.
+
+name: sbom
+on:
+  pull_request:
+    branches:
+      - master
+  push:
+    branches:
+      - master
+
+env:
+  JAVA_DISTRIBUTION: temurin
+
+jobs:
+  sbom:
+    runs-on: ubuntu-latest
+    env:
+      MAVEN_OPTS: -Dhttp.keepAlive=false -Dmaven.wagon.http.pool=false
+      SSH_AUTH_SOCK: /tmp/ssh_agent.sock
+    name: Generate and store SBOM
+    steps:
+      - uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # tag=v3.1.0
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-java@a18c333f3f14249953dab3e186e5e21bf3390f1d # tag=v3.5.1
+        with:
+          java-version: 17
+          distribution: ${{ env.JAVA_DISTRIBUTION }}
+      - name: Get date for cache # see https://github.com/actions/cache README
+        id: get-date
+        run: echo "::set-output name=date::$(/bin/date -u "+%Y%m%d")"
+        shell: bash
+      - name: Use Maven dependency cache
+        uses: actions/cache@56461b9eb0f8438fd15c7a9968e3c9ebb18ceff1 # tag=v3.0.10
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-${{ steps.get-date.outputs.date }}-${{ hashFiles('**/pom.xml') }}
+          restore-keys: ${{ runner.os }}-maven-
+      - name: Build project
+        run: |
+          mvn -B test-compile
+      - name: Generate provenance
+        # Note that this action is a demo, meaning that it works fine now, but
+        # there are no guarantees of future support and development.
+        uses: slsa-framework/github-actions-demo@9474e92bbf825d5b4b46810fc9367dfc73429a2a # tag=v0.1
+        with:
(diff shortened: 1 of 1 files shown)


monperrus:
FYI, this work is part of the Chains research project, where we push the state of the art of software supply chain engineering, see https://chains.proj.kth.se/

.github/workflows/sbom.yml
```diff
@@ -0,0 +1,74 @@
...
+        env:
+          # Target repo to upload provenance file 
+          REMOTE_REPO_URL: git@github.com:chains-project/sbom-files.git
+          GIT_USER: provenance-bot
+          GIT_EMAIL: spoon+sbom@kth.se
+        run: |
+          git config --global user.name "${{ env.GIT_USER }}"
+          git config --global user.email "${{ env.GIT_EMAIL }}"
+          git clone $REMOTE_REPO_URL
+          mkdir -p sbom-files/spoon
```
algomaster99, line 69:
```suggestion
          mkdir -p sbom-files/spoon/slsa
```

You can nest it another level so that we can store SBOM/provenance files generated for Spoon by other tools as well. Proposed structure:
```
sbom-files/
    <project>
        <sbom-generator-tool>
            *sbom.json
            *.provenance
```

.github/workflows/sbom.yml
```diff
@@ -0,0 +1,74 @@
...
+          cp build.provenance sbom-files/spoon/$(date +%s)-${{ github.sha }}.provenance
```
algomaster99, line 70:
```suggestion
          cp build.provenance sbom-files/spoon/slsa/$(date +%s)-${{ github.sha }}.provenance
```

algomaster99 (changes requested):

.github/workflows/sbom.yml
siberov: Good suggestion. Regarding this though, if we want to have a common naming convention: I'm naming the files <timestamp in seconds>-<commit sha>.provenance. The main idea being that it would be easy to find the relevant file, and having them sorted by time by default. Is this a good idea, or do we have another naming convention?

siberov: I have not seen it recommended anywhere, however, it is the name given in the example of the bottom of the README here: https://github.com/slsa-framework/github-actions-demo. I copied it since it made enough sense to me. :)

siberov: Sounds good to me, it is probably also clearer for humans when it is explicitly said to be a json file.

algomaster99: > clearer for humans when it is explicitly said to be a JSON file.

Fully agree.

siberov: > mistake `.provenance` file to be an executable file

BTW, I think that that is caused by the file being too large, not the extension. But I still think that `.json` is a more appropriate extension.

algomaster99: > I think that that is caused by the file being too large, not the extension

You may be right. Let me try it. Thanks for bringing this to my notice. :)

MartinWitt:
I know this is scientific important, but I remember last time we had source graph builds in our CI which often failed and were broken for a longer time. How stable do you think is this CI step? I may have missed something significant, but would it be easier to have a GitHub bot crawling the repos and create the SBOMs for commit hashes, which are currently missing?

siberov:
While the [SBOM-generating action](https://github.com/slsa-framework/github-actions-demo) that this new action depends on is explicitly not production grade, I don't _think_ that this would be unstable, at least as long as we are pinning the version.

Essentially, it builds the project, and hashes the resulting binaries and the source files, storing the result in a [SLSA provenance](https://slsa.dev/provenance/v0.2) file. An example of how the result looks can be seen [here](https://raw.githubusercontent.com/siberov/sbom-files/main/spoon/1665565145-42317ef89a1a1c72987aea5443a4ac0271654d6c.provenance) (warning, large file). Since we want to hash the binaries, just crawling the repo would not be enough.

.github/workflows/sbom.yml
```diff
@@ -0,0 +1,74 @@
...
+#
+# Currently, this uses the slsa-framework/github-actions-demo action to
+# generate an in-toto attestation. This is the pushed to a remote repository
+# for storage.
+
+name: sbom
+on:
+  pull_request:
+    branches:
+      - master
```
algomaster99, line 11:
```suggestion
```

Let us make this action trigger only when a commit is pushed to `master`.

slarse:
> To work, this requires that a secret called SBOM_DEPLOY_SSH is added to the Spoon repo. It must contain the private part of an SSH key, where the public key is added as a deploy key to the [chains-project/sbom-files](https://github.com/chains-project/sbom-files) repository.

You should use an environment for this, add that environment to the workflow and also scope that environment only to the master branch. This will limit the risk of having a private SSH key live in the repository workflows.

See the [docs on GitHub Environments](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment), it's very easy to accomplish and requires only one additional line in the workflow.

siberov:
@slarse , that's a great suggestion, thank you. I was not aware of this feature. If I understand it correctly, it should be addressed by my [latest commit](https://github.com/INRIA/spoon/pull/4955/commits/64419dda4e109773b4283a841ed2a2c970772bd1).

slarse:
> @slarse , that's a great suggestion, thank you. I was not aware of this feature. If I understand it correctly, it should be addressed by my [latest commit](https://github.com/INRIA/spoon/pull/4955/commits/64419dda4e109773b4283a841ed2a2c970772bd1).

Assuming that environment exists and has the desired secret (has to be created through the Settings tab of the repository), then yes that is all you need to do on the workflow side of things :)

siberov:
Hi all,
As @monperrus mentions [here](https://github.com/INRIA/spoon/issues/3879#issuecomment-1280487098), the requisite secrets for this workflow have now been added to the repository.

Does any maintainer have an opinion on where I should go from here to hopefully get this PR merged? If you have any questions at all about the functionality or purpose of this, please don't hesitate to reach out.

MartinWitt (approved):

slarse (approved):
@MartinWitt All good by me

MartinWitt:
Okay, before I merge @algomaster99 is the mail `mannu.poski10@gmail.com` as coauthor correct?

algomaster99:
Yes. It's correct.

slarse:
@siberov This CI job is now consistently failing, could you have a look at what's wrong?

siberov:
Yes, I'll look into it.

slarse:
@siberov It's passing again so there's no immediate need to do anything. The error message just made it look like the action wasn't getting the right arguments, but it seems to have just resolved itself.

siberov:
Great to know. I was just about to have a look. Love it when things fix themselves. :)

PR #4957, opened 2022-10-12
fix: handle empty modifers in CodeFactory#createCatchVariable correctly

MartinWitt:
This should fix the regression produced in #4940. I also applied at least some formatting to the long call chain and removed a random unused field in the test case.

Edit: I created this PR inside VSC, and the developer thought it would be nice to always create the branch in the main repo if possible. We should delete the branch afterwards.


SirYwell (approved):
LGTM

PR #4972, opened 2022-10-19
feat: read parameter names from bytecode if possible

danielbobbert:
This feature allows Spoon to read the parameter name from Java 8 and later byte code that has been built with the "-parameters" flag such that CtParameter objects representing code from the class path contain the "real" parameter name instead of "argX".

### DIFF
--- a/src/main/java/spoon/support/visitor/java/reflect/RtParameter.java
+++ b/src/main/java/spoon/support/visitor/java/reflect/RtParameter.java
@@ -11,6 +11,7 @@ import spoon.SpoonException;
 
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Constructor;
+import java.lang.reflect.Parameter;
 import java.lang.reflect.Type;
 
 /**
@@ -143,7 +144,12 @@ public class RtParameter {
 	public static RtParameter[] parametersOf(RtMethod method) {
 		RtParameter[] parameters = new RtParameter[method.getParameterTypes().length];
 		for (int index = 0; index < method.getParameterTypes().length; index++) {
-			parameters[index] = new RtParameter(null, method.getParameterTypes()[index], method.getGenericParameterTypes()[index], method, null, index);
+			String name = null;
+			Parameter parameter = method.getMethod().getParameters()[index];
+			if (parameter.isNamePresent()) {
+				name = parameter.getName();
+			}
+			parameters[index] = new RtParameter(name, method.getParameterTypes()[index], method.getGenericParameterTypes()[index], method, null, index);
 		}
 		return parameters;
 	}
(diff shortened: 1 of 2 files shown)


MartinWitt:
Hi,
First, thank you for the feature. In spoon we try to have a test for every feature. Could you try to add a test case for this feature? Please don't hesitate to ask for help if you need any.

SirYwell:
I'm not sure how a test for this would look like. We either need .class files as resources (which is bad because we can't properly read them here) or compile a java file and read it then. For that, a proper test infrastructure that can be reused would be needed, otherwise it'll get ugly really quick.

danielbobbert:
I could potentially add a mvn sub-module to the project, which compiles a test class with the appropriate compiler settings for the sake of testing.
I feels like quite a bit of overhead for this relatively small and straightforward addition.
Please let me know how to proceed.

Best regards
Daniel

Am 26.10.2022 um 15:22 schrieb Hannes Greule ***@***.***>:

﻿ This email originated from outside of CGM. Please do not click links or open attachments unless you know the sender and know the content is safe.


I'm not sure how a test for this would look like. We either need .class files as resources (which is bad because we can't properly read them here) or compile a java file and read it then. For that, a proper test infrastructure that can be reused would be needed, otherwise it'll get ugly really quick.

I-Al-Istannen:
> I could potentially add a mvn sub-module to the project, which compiles a test class with the appropriate compiler settings for the sake of testing. I feels like quite a bit of overhead for this relatively small and straightforward addition.

I agree, that sounds a bit too heavy handed. What about an API like
```java
	@Test
	void parameterNamesAreParsedWhenCompilingWithParametersFlag() throws ClassNotFoundException {
		ClassLoader loader = JavacFacade.compileFiles(
			Map.of(
				"Test.java",
				"class Test {\n"
					+ "  public void foo(String bar) {}\n" +
					"}\n"
			),
			List.of("-parameters")
		);
		CtType<?> test = new JavaReflectionTreeBuilder(createFactory()).scan(loader.loadClass("Test"));
		CtMethod<?> method = test.getMethodsByName("foo").get(0);
		CtParameter<?> parameter = method.getParameters().get(0);

		assertThat(parameter.getSimpleName(), is("bar"));
	}
```

Depending on what @MartinWitt and @SirYwell think, I could PR such a facade and then you can use this test (or one you come up with)? Alternatively we could leverage JDT to compile this for us with the parameter flag instead of javac. I'd assume javac is a bit more common though, but this shouldn't really matter.

I-Al-Istannen:
@danielbobbert  The facade was merged in master. If you rebase or merge this PR onto master, you can adapt the test case it brought with it and/or write an additional one :)

I-Al-Istannen:
Hey @danielbobbert, is there anything we can help you with to finish this PR? :)

SirYwell (approved):
Looks good, will merge.

PR #4983, opened 2022-10-26
review: refactor: use putIfAbsent in TypeNameScope

SirYwell:
Simple clean up, avoiding unnecessary `containsKey` calls


slarse (approved):
Thanks @SirYwell, LGTM!

PR #4986, opened 2022-10-28
review: chore: replace deprecated set-output command in GHA

SirYwell:
We currently have warnings in our GH CI, caused by the changes described in https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/.

Example: https://github.com/INRIA/spoon/actions/runs/3341083524 and https://github.com/INRIA/spoon/actions/runs/3341083525

I updated our scripts to use the replacement.


slarse (approved):
Good catch @SirYwell. I keep telling myself I should keep track of the GitHub blog but I just never get around to it...

Merging tomorrow.

PR #4987, opened 2022-10-28
refactor: Refactor bad smell InnerClassMayBeStatic

MartinWitt:
# Repairing Code Style Issues
## InnerClassMayBeStatic
Inner classes that do not reference their enclosing instances can be made static.
This prevents a common cause of memory leaks and uses less memory per instance of the class.

# Repairing Code Style Issues
* InnerClassMayBeStatic (2)


I-Al-Istannen (approved):
LGTM. Though a classloader keeping the outer class alive sounds kinda cute :P

PR #4988, opened 2022-10-29
test: Add javac facade to test reflection tree builder

I-Al-Istannen:
This PR adds a — deliberately simple(istic) — facade to the system Javac compiler. This allows us to compile certain classes using Javac and ensure our reflection tree builder handles it correctly.

### Considerations
We could also use JDT (probably) to compile the files instead. I opted to use Javac as it is a lot more widely used and should also use different java versions in our CI.


MartinWitt (approved):

MartinWitt:
Awesome testing tool thanks @I-Al-Istannen

Issue #4881, opened 2022-09-05
Getting extra parenthesis on variable assignments with concatenation of other variables.


Vardhamansankar:
For example in my java code, I am having a **String query = query1 + query2 + query 3;**
 When I am using **CtLocalVariable.getAssignment()**, its giving **((query1 + query2) + query 3)** as **String query** variable assignment. It is adding extra parenthesis in that output.

I want output as query1 + query2 + query 3 as it is without any extra parenthesis added.

I-Al-Istannen:
Have you tried setting `setMinimizeRoundBrackets(true)` on your printer?

You sadly did not post any example code that triggers the bug for you, but the following might still help you:
```java
        launcher.getEnvironment().setPrettyPrinterCreator(() -> {
            DefaultJavaPrettyPrinter printer = new DefaultJavaPrettyPrinter(launcher.getEnvironment());
            printer.setMinimizeRoundBrackets(true);
            return printer;
        });
```

Alternatively you could try using the *SniperPrinter*, which tries its best to only modify *needed* elements instead of reformatting the whole file.

algomaster99:
@I-Al-Istannen, `setMinimizeRoundBrackets` is `protected` and is not open to public. However, we had a similar issue raised before and this was a working solution for it - https://github.com/INRIA/spoon/issues/4807#issuecomment-1202263121.

I-Al-Istannen:
Ah, neat. Missed that discussion, thanks :)

Vardhamansankar:
Heya @I-Al-Istannen @algomaster99 , I tried this way while creating launcher as follows,
```
Launcher launcher = new Launcher();
		launcher.addInputResource("./src/test/java/spoon/test/ctLocalVariables/Foo.java");
		Environment e = launcher.getEnvironment();
		RoundBracketPrinter printer = new RoundBracketPrinter(launcher.getEnvironment());
		e.setPrettyPrinterCreator(() -> printer);
		launcher.buildModel();
		CtModel model = launcher.getModel();
List<CtLocalVariable> variables =   model.getElements(new TypeFilter<>(CtLocalVariable.class));
		for(CtLocalVariable variable: variables) {
			System.out.println(variable.getSimpleName().toString()+" --------- "+variable.getAssignment().toString());
		}

```

And having an explicit class as **RoundBracketPrinter**
`class RoundBracketPrinter extends DefaultJavaPrettyPrinter {
public RoundBracketPrinter(Environment env) {
super(env);
[...]

PR #4995, opened 2022-11-01
fix: minimise printing parenthesis for type casts in expression

algomaster99:
Reference: https://github.com/INRIA/spoon/issues/4881#issuecomment-1296181539

The changes fix bracket printing for expressions with type casts. It actually does two things:
1. Remove outer brackets. Eg. outermost brackets in `((int) myInt)` are not printed if minimization is requested.
2. Initially, brackets were enforced around the variable as well. For example, `((int) (myInt))`. Maybe, I should do this in a separate PR because it may entail a discussion.


algomaster99:
I also noticed in [`edea720` (#4995)](https://github.com/INRIA/spoon/pull/4995/commits/edea7200fee86094753055ed38a7df89368fcbad) that the suffix `D` is not printed.

algomaster99:
@I-Al-Istannen @MartinWitt can you drop your reviews? I think we need better tests. Maybe you could help me with that?

SirYwell:
Minimising parentheses for casts is conservative at the moment, but mainly because it's not trivial to do right.
As an example, consider following line as second statement in your test case:
`int myInt2 = ((Double) myDouble).intValue();`
With your current changes, this will be printed as `int myInt2 = (java.lang.Double) myDouble.intValue()` which does not compile anymore.

We can surely think about cases where the parentheses are not required (for example, a direct assignment like your test might be such case), but I'd prefer to still be conservative and only drop the brackets if we are really sure that it will still compile.

algomaster99:
I am sorry I am seeing this so late. It got lost somewhere in my notifications.

Thanks for this test case. A patch like checking if the default expression is an instance of invocation and then checking if it has a target and executable ref can fix it. 

> We can surely think about cases where the parentheses are not required

We should. As of now, I do not have any more test cases in mind. I usually rely on clients reporting bugs in this area to improve the printer.

> if we are really sure that it will still compile.

That is true. Maybe we can add an assertion in our test cases that compiles the pretty-printed string?

SirYwell:
> We should. As of now, I do not have any more test cases in mind. I usually rely on clients reporting bugs in this area to improve the printer.

The important point is that we go the safe route, so instead of "don't print parentheses unless ..." the code has to be "always print parentheses unless ...". Too many parentheses are okay, code that doesn't compile is not okay.

> Maybe we can add an assertion in our test cases that compiles the pretty-printed string?

I think there are some larger tests doing that. Afaik there are some test cases that take the spoon source code, print it and compile it to make sure the printer works. Maybe we can reuse something like that with the minimize brackets setting.

algomaster99:
>  The important point is that we go the safe route, so instead of "don't print parentheses unless ..." the code has to be "always print parentheses unless ...". Too many parentheses are okay, code that doesn't compile is not okay.

This is what is happening right now, don't you think? Initially, we used to print redundant brackets, but then clients requested minimisation so we started adding patches like https://github.com/INRIA/spoon/pull/3823, https://github.com/INRIA/spoon/pull/4809, and a few more to adhere to specific cases.

> Maybe we can reuse something like that with the minimize brackets setting.

Yes, I plan to do that.

algomaster99:
@SirYwell I tried to compile the pretty-printed string in [`24385dc` (#4995)](https://github.com/INRIA/spoon/pull/4995/commits/24385dc7b4df9993732806e08fbbf1af0105e7af). I have an intuition there may already be a utility for this so let me know if you are aware. Also, maybe we can add the `afterEach` method in `DJPPTest` and `SJPPTest` that checks for compilation afterward in a separate PR?

I feel the changes in [`4c68134` (#4995)](https://github.com/INRIA/spoon/pull/4995/commits/4c681343f10f6f755ecc77878b6751e59a14e1fc) belong to [RoundBracketAnalyzer#requiresRoundBrackets](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/reflect/visitor/RoundBracketAnalyzer.java#L32). However, in its current state, it only checks for unary and binary expression so it would take some refactoring to make it generic.

algomaster99:
There seems to be some flakiness in the CI. Could anyone please run the workflow again?

I-Al-Istannen:
You should be able to just create a new Launcher and build the printed model. Set `Environment#shouldCompile` to true. If `buildModel` exists cleanly, it compiled.

The failing workflow was https://github.com/INRIA/spoon/actions/runs/3438229451/jobs/5734018883, I will need to have a look at that later. A flakey CI is really annoying.

algomaster99:
> The failing workflow was [INRIA/spoon/actions/runs/3438229451/jobs/5734018883](https://github.com/INRIA/spoon/actions/runs/3438229451/jobs/5734018883), I will need to have a look at that later. A flaky CI is really annoying.

This is not flaky. I think I introduced some regression. Let me check again.

SirYwell:
> This is what is happening right now, don't you think? Initially, we used to print redundant brackets, but then clients requested minimisation so we started adding patches like #3823, #4809, and a few more to adhere to specific cases.

This is true for the linked PRs, but with the current state of this PR, we say `should set bracket` for a specific case (we don't know if we cover all other cases) instead of `should not set bracket´ for a specific case (we know that it is save for this case).

The `isTargetOfInvocation` check is also somewhat duplicated below (https://github.com/INRIA/spoon/pull/4995/files#diff-69f595720b242e1a1322a888005148a4400e96e04f3fe5a2e898e44e6471ebf4R423-R425) but with more special casing. That's also the source of the current test failures, you're setting brackets now *always* if minimizeRoundBrackets is enabled.

> I have an intuition there may already be a utility for this so let me know if you are aware.

I don't have anything in my mind right now, I would need to skim through existing tests first.

> Also, maybe we can add the `afterEach` method in `DJPPTest` and `SJPPTest` that checks for compilation afterward in a separate PR?

I'm not sure if that's doable, the tests are likely not consistent enough and if it does fail *during* the test, tooling might break.

> I feel the changes in [`4c68134` (#4995)](https://github.com/INRIA/spoon/pull/4995/commits/4c681343f10f6f755ecc77878b6751e59a14e1fc) belong to [RoundBracketAnalyzer#requiresRoundBrackets](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/reflect/visitor/RoundBracketAnalyzer.java?rgh-link-date=2022-11-10T15%3A48%3A21Z#L32). However, in its current state, it only checks for unary and binary expression so it would take some refactoring to make it generic.

Yes, that sounds like a good place for anything related to minimizeRoundBrackets. Generally, I think that should touch the DJPP as little as possible.

algomaster99:
> The isTargetOfInvocation check is also somewhat duplicated below ([#4995 (files)](https://github.com/INRIA/spoon/pull/4995/files#diff-69f595720b242e1a1322a888005148a4400e96e04f3fe5a2e898e44e6471ebf4R423-R425)) but with more special casing. That's also the source of the current test failures, you're setting brackets now always if minimizeRoundBrackets is enabled.

Indeed that was the source. Thank you! I fixed it in recent commits.

algomaster99:
@SirYwell it is ready for review now.

src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
```diff
@@ -400,7 +402,7 @@ public DefaultJavaPrettyPrinter scan(CtElement e) {
 	}
 
 	private boolean shouldSetBracket(CtExpression<?> e) {
-		if (!e.getTypeCasts().isEmpty()) {
+		if (!e.getTypeCasts().isEmpty() && !isMinimizeRoundBrackets()) {
```
SirYwell, line 405: This additional check can still cause issues, because now we can go through the whole method and return `false` as a default value, even if there *are* type casts. It might work by returning `!e.getTypeCasts().isEmpty()` at the bottom,

Consider following silly example:
```java
class A {
    A a;
    int i;
}
class C extends A {
    Object a;
}
```
(the subclass shadows a field of its superclass)
and following statement:
```java
C c = null;
int i = ((A) c).a.i;
```
This currently does not enter the if body but instead return `false` at the end of the method. Obviously, `(A) c.a.i` does not compile.

Even worse, we can construct code that still compiles but has different semantics by adjusting the classes as follows:

```java
class A {
    A a;
    int i;
}
class C extends A {
    C a;
		A i;
}
```
and the statements:
```java
C c = null;
var i = ((A) c).a.i;
```

SirYwell (changes requested):

algomaster99:
> It might work by returning !e.getTypeCasts().isEmpty()

I think we should return `false` by default as it aligns with your statement - "always print parentheses unless ...".

> `int i = ((A) c).a.i;`

For this case, we need to enclose `c` which is `VariableRead` so if we just add a case for it, the test would pass.

However, I think it is becoming very confusing to read through `shouldSetBracket`. We should really consider refactoring it inside `RoundBracketAnalyzer`. Basically, refactor https://github.com/INRIA/spoon/blob/8e23bae972aa65c6d1580e728fa0b1a9bc7544d9/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java#L414-L419 and possibly minimise the conditions. In theory, the conditions should be built upon from these three conditions:
1. We need to put round brackets wherever we type cast _before_ invoking some executable on it.
2. We should not put round brackets if we type cast, but we do not invoke anything on it.
3. By default, we always set round brackets.

SirYwell:
> I think we should return `false` by default as it aligns with your statement - "always print parentheses unless ...".

From my understanding, the method should return `true` whenever the printer should print parentheses.

> However, I think it is becoming very confusing to read through `shouldSetBracket`. [...]

I think we're at a stage where we need to refactor something, but I'm not sure about the *where* and the *how*.
One problem is that this method currently does two things:

1. It avoids the obvious cases where no parentheses are needed, e.g., `System.out.println("Hello World")` instead of `(System.out).println("Hello World")`
2. It *conditionally* (depending on the `minimizeRoundBrackets`) avoids additional, not as obvious cases in which parentheses can be omitted. Before this PR, this only happened in the `RoundBracketAnalyzer`, and never if casts are present.

With the new additional check at the very top of the method, we do not longer return `true` unconditionally if casts are present, but in worst case we fall back to returning `false`. This heavily changes how the method behaves in a lot of cases which were only covered implicitly (i.e. "we always print parentheses if casts are present").

One idea would be to modify the first `if` as follows:
```java
if (isMinimizeRoundBrackets()) {
	// Do RoundBracketAnalyzer stuff here, make RoundBracketAnalyzer work with type casts too
} else if (!e.getTypeCasts().isEmpty()) {
	return true; // keep behavior of non-minimizing printer
}
// drop RoundBracketAnalyzer stuff below, keep everything else as before this PR
```

This way, we can make sure the behavior of non-minimizing printing just remains the same.
WDYT?

algomaster99:
@SirYwell do you think the refactoring makes it more readable?

SirYwell:
Yes! This definitely looks better. I'm still somewhat concerned as the RoundBracketAnalyzer is now called in cases with casts present. This wasn't the case before, so any behavior of the RoundBracketAnalyzer probably isn't properly tested for such cases. The trivial fix for now would be to also have an addtional isEmpty check in the return statement here https://github.com/INRIA/spoon/pull/4995/files#diff-69f595720b242e1a1322a888005148a4400e96e04f3fe5a2e898e44e6471ebf4R409 too.

algomaster99:
@SirYwell can you check now?

SirYwell (dismissed):
This is good from my side now, but someone else should probably take a look too (@I-Al-Istannen maybe?).

I-Al-Istannen (changes requested):
```patch
From c2a903a046d4781015a5f523bc250fbabbd8e619 Mon Sep 17 00:00:00 2001
From: I-Al-Istannen <i-al-istannen@users.noreply.github.com>
Date: Wed, 16 Nov 2022 18:45:34 +0100
Subject: [PATCH] Some comments

Co-authored-by: Hannes Greule <SirYwell@users.noreply.github.com>
---
 .../visitor/DefaultJavaPrettyPrinter.java     | 32 +++++++++++++++----
 .../visitor/DefaultJavaPrettyPrinterTest.java |  4 ++-
 src/test/java/spoon/test/eval/EvalTest.java   |  2 +-
 3 files changed, 29 insertions(+), 9 deletions(-)

diff --git a/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java b/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
index 9a8b4a935..ede37c500 100644
--- a/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
+++ b/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
@@ -68,6 +68,7 @@ import spoon.reflect.code.CtTryWithResource;
 import spoon.reflect.code.CtTypeAccess;
 import spoon.reflect.code.CtTypePattern;
 import spoon.reflect.code.CtUnaryOperator;
+import spoon.reflect.code.CtVariableAccess;
 import spoon.reflect.code.CtVariableRead;
 import spoon.reflect.code.CtVariableWrite;
 import spoon.reflect.code.CtWhile;
@@ -296,7 +297,7 @@ public class DefaultJavaPrettyPrinter implements CtVisitor, PrettyPrinter {
 			elementPrinterHelper.writeComment(e, CommentOffset.BEFORE);
 		}
 		getPrinterHelper().mapLine(e, sourceCompilationUnit);
-		if (shouldSetBracket(e)) {
+		if (shouldSetBracketAroundExpressionAndCast(e)) {
 			context.parenthesedExpression.push(e);
 			printer.writeSeparator("(");
 		}
@@ -305,12 +306,29 @@ public class DefaultJavaPrettyPrinter implements CtVisitor, PrettyPrinter {
 				printer.writeSeparator("(");
 				scan(r);
 				printer.writeSeparator(")").writeSpace();
-				if (!isMinimizeRoundBrackets()) {
-					printer.writeSeparator("(");
-					context.parenthesedExpression.push(e);
-				}
 			}
+			if (shouldSetBracketAroundCastTarget(e)) {
+				printer.writeSeparator("(");
+				context.parenthesedExpression.push(e);
+			}
+		}
+	}
+
+	private boolean shouldSetBracketAroundCastTarget(CtExpression<?> expr) {
+		if (!isMinimizeRoundBrackets()) {
+			return true;
 		}
+
+		if (expr instanceof CtTargetedExpression) {
+			return false;
+		}
+		if (expr instanceof CtLiteral) {
+			return false;
+		}
+		if (expr instanceof CtVariableAccess) {
+			return false;
+		}
+		return true;
 	}
 
 	/**
@@ -401,12 +419,12 @@ public class DefaultJavaPrettyPrinter implements CtVisitor, PrettyPrinter {
 		return this;
 	}
 
-	private boolean shouldSetBracket(CtExpression<?> e) {
+	private boolean shouldSetBracketAroundExpressionAndCast(CtExpression<?> e) {
 		if (isMinimizeRoundBrackets()) {
 			RoundBracketAnalyzer.EncloseInRoundBrackets requiresBrackets =
 					RoundBracketAnalyzer.requiresRoundBrackets(e);
 			if (requiresBrackets != RoundBracketAnalyzer.EncloseInRoundBrackets.UNKNOWN) {
-				return requiresBrackets == RoundBracketAnalyzer.EncloseInRoundBrackets.YES && e.getTypeCasts().isEmpty();
+				return requiresBrackets == RoundBracketAnalyzer.EncloseInRoundBrackets.YES || !e.getTypeCasts().isEmpty();
 			}
 			if (e.isParentInitialized() && e.getParent() instanceof CtTargetedExpression && ((CtTargetedExpression) e.getParent()).getTarget() == e) {
 				return e instanceof CtVariableRead<?> && !e.getTypeCasts().isEmpty();
diff --git a/src/test/java/spoon/reflect/visitor/DefaultJavaPrettyPrinterTest.java b/src/test/java/spoon/reflect/visitor/DefaultJavaPrettyPrinterTest.java
index 6546fa2e7..84ff94fe4 100644
--- a/src/test/java/spoon/reflect/visitor/DefaultJavaPrettyPrinterTest.java
+++ b/src/test/java/spoon/reflect/visitor/DefaultJavaPrettyPrinterTest.java
@@ -67,7 +67,9 @@ public class DefaultJavaPrettyPrinterTest {
             "1 | 2 & 3",
             "(1 | 2) & 3",
             "1 | 2 ^ 3",
-            "(1 | 2) ^ 3"
+            "(1 | 2) ^ 3",
+            "((int) (1 + 2)) * 3",
+            "(int) (int) (1 + 1)",
     })
     public void testParenOptimizationCorrectlyPrintsParenthesesForExpressions(String rawExpression) {
         // contract: When input expressions are minimally parenthesized, pretty-printed output
diff --git a/src/test/java/spoon/test/eval/EvalTest.java b/src/test/java/spoon/test/eval/EvalTest.java
index 8a565bde2..c6b55b009 100644
--- a/src/test/java/spoon/test/eval/EvalTest.java
+++ b/src/test/java/spoon/test/eval/EvalTest.java
@@ -96,7 +96,7 @@ public class EvalTest {
 		CtBlock<?> b = type.getMethodsByName("testDoNotSimplifyCasts").get(0).getBody();
 		assertEquals(1, b.getStatements().size());
 		b = b.partiallyEvaluate();
-		assertEquals("return ((U) ((java.lang.Object) (spoon.test.eval.testclasses.ToEvaluate.castTarget(element).getClass())))", b.getStatements().get(0).toString());
+		assertEquals("return ((U) (java.lang.Object) (spoon.test.eval.testclasses.ToEvaluate.castTarget(element).getClass()))", b.getStatements().get(0).toString());
 	}
 
 	@Test
-- 
2.38.1
```

algomaster99:
> return requiresBrackets == RoundBracketAnalyzer.EncloseInRoundBrackets.YES || !e.getTypeCasts().isEmpty();

It should have been this way to preserve the old behaviour. Thank you for noticing. The commit looks good to me and I have pushed it.

MartinWitt (approved):

I-Al-Istannen (approved):

MartinWitt:
@algomaster99 we had some changes to Qodana on the master branch. Could you rebase/merge with the master to run the checks with the current Qodana version?

algomaster99:
Done :)

MartinWitt:
Thanks @algomaster99

PR #5014, opened 2022-11-17
review: fix: Properly accept urls with plus signs (+)

I-Al-Istannen:
This fixes spoon's input validation rejection classloader with URLs containing plus signs (+).

This was accidentally broken in https://github.com/INRIA/spoon/pull/4358 without any of us noticing. I applied the suggestion I had back then (https://github.com/INRIA/spoon/pull/4358#discussion_r771256255) and now use the path API. If this fails, we fall back to the URL encoding added in #4358.

This change should therefore be mostly harmless but maybe @slarse wants to weigh in.


slarse (approved):
Can't we just ask people to not put odd characters in their filepaths? :/

slarse:
@I-Al-Istannen LGTM, ready to merge?

I-Al-Istannen:
> Can't we just ask people to not put odd characters in their filepaths? :/

I wish :D Figuring our the root cause for this issue was really annoying :P Especially because this issue prevented a full classpath from being built which triggered another bug which I initially investigated…

> @I-Al-Istannen LGTM, ready to merge?

Yes, should be fine from my end. I tried to be as conservative as possible.

slarse:
Thanks @I-Al-Istannen !

PR #5016, opened 2022-11-17
fix: model building when annotation on record has no @Target

I-Al-Istannen:
JLS [9.6.4.1](https://docs.oracle.com/javase/specs/jls/se19/html/jls-9.html#jls-9.6.4.1) states:
>  If an annotation of type java.lang.annotation.Target is not present on the declaration of an annotation interface A, then A is applicable in all declaration contexts and in no type contexts. 

Previously the code crashed with a NPE.


MartinWitt (approved):

MartinWitt:
LGTM will merge. One day, we should really use Nullaway to fix all NPEs.

PR #5025, opened 2022-11-26
chore: Update CI python version from 3.6 to 3.11

I-Al-Istannen:
Github recently updated their `ubuntu-latest` image to 22.04. This broke setup-python as it no python 3.6 tarball is provided (https://github.com/actions/setup-python/issues/544). 

This PR fixes the issue by just updating the CI to use python 3.9. It does not adjust the ci scripts to take advantage of possible new python features introduced since then.

This breakage is responsible for the current red master (as the PR run was executed *before* the Github runner change but the master action ran *after* it) and unblocks some PRs.

----

I also needed to update commonmark. Note that commonmark is **deprecated** for **over a year** now: https://github.com/readthedocs/commonmark.py/issues/308. We might to want migrate away from it in a future PR, though we don't really *need* to. The source code we analyze is benign so security isn't really an issue and the featureset is enough for us.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -122,7 +122,7 @@ jobs:
           distribution: ${{ env.JAVA_DISTRIBUTION }}
       - uses: actions/setup-python@13ae5bb136fac2878aff31522b9efb785519f984 # tag=v4.3.0
         with:
-          python-version: 3.6
+          python-version: 3.11
 
       - name: Get date for cache # see https://github.com/actions/cache README
         id: get-date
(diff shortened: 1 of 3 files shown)


MartinWitt:
Is there a reason we don't upgrade to the newest python version? I'm kind of out of the loop for python and if the break APIs we use.

I-Al-Istannen:
I consider the python version in the CI some kind of "we support at least Version X" guarantee (as we see breakages). According to repology, a few are still on 3.9: https://repology.org/project/python/versions. Our python files are small and simple, upgrading for us should be basically free.

I'm fine with 3.11 as well.

MartinWitt:
Thanks @I-Al-Istannen

slarse:
We only use Python for scripts so we hardly need to care about... Anything. If it runs, it runs. There are a lot of breaking changes between 3.6 and 3.11 (especially when it comes to third party packages), but again, if our CI runs we're good :)

PR #5035, opened 2022-12-10
fix: Do not return null in fluent API setters

I-Al-Istannen:
Returning null in a fluent setter kind of defeats the purpose: Using the fluent API crashes with a NPE.

The first location was found because it crashed on me :'(. The other location was found by:
```java
	@ModelTest("src/main/java")
	void foo(CtModel model, Factory factory) {
		CtTypeReference<?> element = factory.Type().get(CtElement.class).getReference();
		for (CtModule module : model.getAllModules()) {
			module.accept(new CtScanner() {
				@Override
				public <T> void visitCtMethod(CtMethod<T> m) {
					if (!TypeAdaptor.isSubtype(m.getDeclaringType(), element)) {
						return;
					}
					if (m.getSignature().equals("S()")) {
						return;
					}
					if (!m.getSimpleName().startsWith("set")) {
						return;
					}
					m.accept(new CtScanner() {
						@Override
						public <R> void visitCtReturn(CtReturn<R> returnStatement) {
							if (!(returnStatement.getReturnedExpression() instanceof CtLiteral)) {
								return;
							}
							if (((CtLiteral<R>) returnStatement.getReturnedExpression()).getValue() == null) {
								System.out.println("OH NO " + m.getDeclaringType().getQualifiedName() + "#" + m.getSignature());
							}
						}
					});
				}
			});
		}
	}
```

### DIFF
--- a/src/main/java/spoon/support/reflect/code/CtFieldAccessImpl.java
+++ b/src/main/java/spoon/support/reflect/code/CtFieldAccessImpl.java
@@ -33,7 +33,7 @@ public abstract class CtFieldAccessImpl<T> extends CtVariableAccessImpl<T> imple
 		}
 		getFactory().getEnvironment().getModelChangeListener().onObjectUpdate(this, TARGET, target, this.target);
 		this.target = target;
-		return null;
+		return (C) this;
 	}
 
 	@Override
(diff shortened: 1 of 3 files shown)


MartinWitt:
Could you add your search as an architecture check? Returning null in a fluent API should never happen.

MartinWitt (dismissed):

MartinWitt:
LGTM will merge after the 24 hours.

MartinWitt (approved):

MartinWitt:
Thanks @I-Al-Istannen

PR #5060, opened 2022-12-30
chore: Use Java 11 for extra CI checks

slarse:
We should run our extra checks on the least supported version. It's significantly more likely to have breakage because of backwards-incompatible changes (e.g. Depclean moving to Java 17) than forwards-incompatible changes.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -118,7 +118,7 @@ jobs:
           fetch-depth: 0
       - uses: actions/setup-java@1df8dbefe2a8cbc99770194893dd902763bee34b # v3.9.0
         with:
-          java-version: 17
+          java-version: 11
           distribution: ${{ env.JAVA_DISTRIBUTION }}
       - uses: actions/setup-python@5ccb29d8773c3f3f653e1705f474dfaa8a06a912 # v4.4.0
         with:


SirYwell:
What happens if we then update the depclean plugin?

slarse:
> What happens if we then update the depclean plugin?

It breaks, which IMO is desirable. The extra CI checks verify the submodules, which should be as compatible with Java 11 as spoon-core. If a single thing requires a later version of Java, we need to run that in a separate CI job.

It also just occurred to me that I don't think it's intentional for Depclean to require Java 17 going from 2.0.3 to 2.0.4.

SirYwell:
Would it make sense to wait with merging this PR until the situation of depclean is clarified? If it stays on Java 17, we'll need to run the task with Java 17 too, and e.g., rather use the `--release` option. WDYT?

slarse:
> If it stays on Java 17, we'll need to run the task with Java 17 too, and e.g., rather use the --release option.

That would be true for the deploy job, but that's not what this is. The extra checks is a test job, it publishes nothing.

For tests, which a lot of the extra checks are when it comes to the submodules, we should run either with all supported versions or the least supported version. That's the safest bet. If some particular tool requires a later version of Java, it should then be executed in a dedicated job for that particular tool.

Regardless, see castor-software/depclean#150. I think we can merge this now.

SirYwell (approved):
Okay, makes sense. Let's merge it then.

PR #5065, opened 2023-01-05
doc: update supported java version in readme.md

MartinWitt:
To my best knowledge, we support up to java 20, so we should update the doc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 # Spoon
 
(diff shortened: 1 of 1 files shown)


slarse:
Do we really support all features up until Java 20? Or do you mean that Spoon _runs_ on Java 20?

MartinWitt:
The latest non preview features were: records and type patterns, both are implemented (see CtRecord and CtTypePattern). Java 20 has no new non preview feature relevant for spoon(see https://openjdk.org/projects/jdk/20/). So yes, we support all non preview features up to java 20 and spoon runs on java 20.

slarse:
Hmm, isn't it a bit misleading to claim that Spoon _fully_ supports version X when it does not implement any preview features? Minor semantics here, but perhaps drop the "fully"?

MartinWitt:
Agree, I removed the `fully`.

slarse (approved):
LGTM, thanks @MartinWitt :)

PR #5070, opened 2023-01-10
fix: Remove escaping of ' in a CtLiteral<String>

MartinWitt:
Currently, spoon escapes every `'` in a `CtLiteral<String>` without a source position. In a `string `escaping `'` is not necessary.

### DIFF
--- a/src/main/java/spoon/reflect/visitor/LiteralHelper.java
+++ b/src/main/java/spoon/reflect/visitor/LiteralHelper.java
@@ -118,7 +118,7 @@ abstract class LiteralHelper {
 	 * This exists for backwards compatibility.
 	 * @param literal to be converted literal
 	 * @param sb the string builder to append the literal
-	 * @param c  the character to append 
+	 * @param c  the character to append
 	 * @param mayContainsSpecialCharacter  true if the string may contains special characters.
 	 */
 	static void appendCharLiteral(StringBuilder sb, Character c, boolean mayContainsSpecialCharacter) {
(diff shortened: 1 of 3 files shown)


MartinWitt:
I could create another method overload* to fix the constant value warning. But this would increase the size of the class unnecessary. Any opinions? 

* Actually this is not a method overload because the methods are static

I-Al-Istannen:
> `	 * This exists for backwards compatibility`

The method seems  to be package private, why do you need to keep it if it is never called?

MartinWitt:
I thought about the one guy creating his code inside the same package in his project. I can remove this method if we ignore that one guy.

I-Al-Istannen:
Yes, I would ignore them.

src/main/java/spoon/reflect/visitor/LiteralHelper.java
```diff
@@ -155,7 +159,7 @@ static String getStringLiteral(String value, boolean mayContainsSpecialCharacter
 		} else {
 			StringBuilder sb = new StringBuilder(value.length() * 2);
 			for (int i = 0; i < value.length(); i++) {
-				appendCharLiteral(sb, value.charAt(i), mayContainsSpecialCharacter);
+				appendCharLiteral(sb, value.charAt(i), true, true);
```
MartinWitt, line 162: This fixes the qodana warning, I have no clue if this is a better code style.

I-Al-Istannen (changes requested):

I-Al-Istannen (approved):

PR #5094, opened 2023-02-01
fix: creation of CtTypeReferences from java classes

I-Al-Istannen:
Creating `CtTypeReference`s from java classes was a bit too ad-hoc for my usecase. This PR adds support for array types and anonymous classes. Anonymous classes do not appear in the shadow model but can still be useful when dealing with arbitrary objects. They are also supported by the source model.

Notable changes: This PR switches from `CLass#getDeclaringClass` to `Class#getEnclosingClass` to properly work with anonymous classes. This seems to be closer to what the source model uses as well.

### DIFF
--- a/src/test/java/spoon/test/factory/CodeFactoryTest.java
+++ b/src/test/java/spoon/test/factory/CodeFactoryTest.java
@@ -17,6 +17,8 @@
 package spoon.test.factory;
 
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
 import spoon.reflect.code.CtAssignment;
 import spoon.reflect.code.CtThisAccess;
 import spoon.reflect.code.CtTypeAccess;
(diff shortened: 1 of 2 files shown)


I-Al-Istannen:
@MartinWitt ping!

src/test/java/spoon/test/factory/CodeFactoryTest.java
```diff
@@ -66,4 +68,54 @@ void createCtCatcVariablehWithoutModifiers() {
...
+		Class<?> clazz = (new Object() {
+		}.getClass());
+		Factory factory = createFactory();
+		assertEquals(
+			clazz.getName(),
+			factory.createCtTypeReference(clazz).getQualifiedName()
+		);
+	}
+
+	private static class TestClassInnerClass {}
```
MartinWitt, line 120: Maybe move this test object closer to the usage?

I-Al-Istannen: I inlined it.

MartinWitt (approved):

MartinWitt:
Thanks @I-Al-Istannen

PR #5096, opened 2023-02-02
fix: enclosing class parameters being ignored for static inner classes

I-Al-Istannen:
While constructors of non-static inner classes have an *implicit* parameter of the enclosing class's type, static classes *do not*. This should be reflected in the model.


SirYwell (approved):
Fix looks good and the JavacFacade really makes the tests easy to grasp. Will merge later.

PR #5113, opened 2023-02-16
fix: Do not visit package for nested classes

I-Al-Istannen:
If we did, we would overwrite potential top-level classes in the package. This happened with `java.util.ArrayList` and `java.util.Arrays$ArrayList` in a project of mine. As a consequence, lookups for `java.util.ArrayList$ListIter` fail: The now-overwritten `ArrayList` has no member classes and the lookup returns null.

The problem cost me a few hours to find and fix, but I hope the fix seems reasonable enough.

### DIFF
--- a/src/test/java/spoon/support/visitor/java/JavaReflectionTreeBuilderTest.java
+++ b/src/test/java/spoon/support/visitor/java/JavaReflectionTreeBuilderTest.java
@@ -55,6 +55,8 @@ import com.mysema.query.support.ProjectableQuery;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.condition.EnabledForJreRange;
 import org.junit.jupiter.api.condition.JRE;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
 import spoon.Launcher;
 import spoon.SpoonException;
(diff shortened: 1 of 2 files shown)


src/test/java/spoon/support/visitor/java/JavaReflectionTreeBuilderTest.java
```diff
@@ -916,4 +918,43 @@ void testNonStaticInnerClassConstructorWithEnclosingClassArgument() throws Class
...
+			List.of()
+		);
+		Factory factory = createFactory();
+		// Load the victim
+		factory.Type().get(loader.loadClass("Victim"));
+		// Let it get replaced by First$Collider
+		factory.Type().get(loader.loadClass("First"));
+
+		// This will throw if the replacement was successful
+		CtType<?> victim = factory.Type().get(loader.loadClass("Victim$Inner"));
```
SirYwell, line 952: This could be wrapped into `assertDoesNotThrow` to clarify what went wrong. (also the "successful" sounds very positive, while it really isn't, but I don't have any better idea for that comment right now)

SirYwell (changes requested):

SirYwell (approved):
Looks good! Will merge tomorrow.

PR #5119, opened 2023-02-21
fix: Do not import two types with the same simple name

I-Al-Istannen:
Previously, spoon would turn
```java
package bar;
public class User {
  public static void foo() {
    System.out.println(foo.bar.baz.Locale.HELLO);
    System.out.println(java.util.Locale.GERMANY);
  }
}
```
into
```java
package bar;
import foo.bar.baz.Locale;
import java.util.Locale;
public class User {
    public static void foo() {
        System.out.println(Locale.HELLO);
        System.out.println(Locale.GERMANY);
    }
}
```

which is *not ideal*. There was a test case verifying this does not happen, but as far as I can see the assertions in the test are wrong. I also have no idea what code was supposed to implement it.

### DIFF
--- a/src/main/java/spoon/reflect/visitor/ImportConflictDetector.java
+++ b/src/main/java/spoon/reflect/visitor/ImportConflictDetector.java
@@ -26,6 +26,7 @@ import spoon.reflect.reference.CtPackageReference;
 import spoon.reflect.reference.CtTypeReference;
 import spoon.support.Experimental;
 
+import java.util.Map;
 
 /**
  * Detects conflicts needed to be required be a fully-qualified name.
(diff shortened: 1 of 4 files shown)


src/main/java/spoon/reflect/visitor/ImportConflictDetector.java
```diff
@@ -26,6 +26,8 @@
 import spoon.reflect.reference.CtTypeReference;
 import spoon.support.Experimental;
 
+import java.util.Map;
+
```
MartinWitt, line 31: keep the diff small

src/main/java/spoon/reflect/visitor/ImportConflictDetector.java
```diff
@@ -196,6 +198,18 @@ protected void handleTypeReference(CtTypeReference<?> ref, LexicalScope nameScop
 				ref.setSimplyQualified(false);
 				return false;
 			});
+
+			if (!ref.isImplicit() && ref.isSimplyQualified()) {
+				Map<String, String> encounteredNames = ((LexicalScopeScanner) scanner).getEncounteredImportedQualifiedNames();
+				if (!encounteredNames.containsKey(ref.getSimpleName())) {
+					encounteredNames.putIfAbsent(ref.getSimpleName(), ref.getQualifiedName());
+				}
```
MartinWitt, line 206: Shouldn't the `putIfAbsent` replace the check in the if?

I-Al-Istannen: That is true, I refactored the check a few times and it became redundant :)

MartinWitt:
I will have a look after you passed the CI :)

I-Al-Istannen:
> I will have a look after you passed the CI :)

grafik

MartinWitt:
I belive you need to refresh the base branch because of #5120

I-Al-Istannen:
> I belive you need to refresh the base branch because of https://github.com/INRIA/spoon/pull/5120

Yes, this would have been be a silent bad merge :) One of the reasons some projects require a CI run on an up-to-date PR…

MartinWitt (approved):

MartinWitt:
Thanks @I-Al-Istannen

PR #5139, opened 2023-03-13
chore: prepare next version

monperrus:


pom.xml
```diff
@@ -18,7 +18,7 @@
 
   <artifactId>spoon-core</artifactId>
   <packaging>jar</packaging>
-  <version>10.3.0-SNAPSHOT</version>
+  <version>10.4.0-SNAPSHOT</version>
```
hboutemy, line 21: and before switching to next SNAPSHOT, it would be nice:
1. to have 10.3.0
2. that the Git tag points to that commit

and it would also be nice to have each beta marked: nothing is currently reproducible because Git has no pointer to correct commits
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/fr/inria/gforge/spoon/spoon-core/README.md

monperrus: @hboutemy fully agree, this will be part of our upcoming CD release pipeline, see https://github.com/INRIA/spoon/issues/3890

PR #5143, opened 2023-03-19
review: chore(CI): bump java ea version to 21

MartinWitt:
As java 20 is released at 21.03 we can start using the 21-ea builds in our CI.


MartinWitt:
I believe they will add 21-ea builds after 20 is published.

SirYwell:
Maybe we can go from 19 to 20 in this PR too?

MartinWitt:
Sure I will update this pr

Hannes Greule ***@***.***> schrieb am Mo., 20. März 2023,
08:13:

MartinWitt:
I am not 100% happy with changing to the oracle setup-java action, but the current setup-java action simply does not have early access builds for jdk-20. The oracle actions uses setup-java by itself, so there shouldn't be any large changes for us.

SirYwell (approved):
Looks like the oracle action is just a wrapper that downloads the JDK from the given website and uses the actions/setup-java action then normally.

Let's try that out then!

PR #5147, opened 2023-03-27
Case position hack

raghav-deepsource:
This PR fixes a particular exception during model build by adding an exception guard with null return in certain edge cases.


raghav-deepsource:
oops.

PR #5164, opened 2023-04-06
review: chore(actions): add pom quality checker to github actions

MartinWitt:
This PR adds 2 checks to our CI. First we check if the doc has no error and second if the pom follows all the rules for maven central. We should be already fulfilling this all, but checking it on every commit seems useful.


MartinWitt:
Hmm I am unsure if only jreleaser/pomchecker want the developer tag non-empty. Sonatype itself also isn't clear there: https://central.sonatype.org/publish/requirements/#developer-information

MartinWitt:
Okay I have 2 solutions: Either we mention the original spoon authors or simply add all current integrators in it. Any comments or other solutions?

I-Al-Istannen:
> Developers are presumably members of the project's core development. Note that, although an organization may have many developers (programmers) as members, it is not good form to list them all as developers, but only those who are immediately responsible for the code. A good rule of thumb is, if the person should not be contacted about the project, they do not need to be listed here.

From https://maven.apache.org/pom.html#Developers would suggest listing the integrators, or at least some of them.

MartinWitt:
I have added all integrators with minimal information. If someone wants to include more information, feel free to PR it or comment here.

MartinWitt:
The hashes for the actions are missing but renovate should instantly add them after a merge.

monperrus:
thanks @MartinWitt

PR #5182, opened 2023-04-19
chore: Enable renovate automerge for non-major updates

MartinWitt:
After discussion with @monperrus this renovate config enables automerge of green PR updates from renovate. I have changed the master protection rules to let renovate bypass them. I removed some old mockito rule which seems useless.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -9,8 +9,7 @@
             "minor",
             "patch",
             "pin",
-            "digest",
-            "fix"
+            "digest"
          ],
          "automerge":true,
          "semanticCommitType":":arrow_up:"
--- a/renovate.json
+++ b/renovate.json
@@ -12,7 +12,6 @@
             "digest"
          ],
          "automerge":true,
-         "semanticCommitType":":arrow_up:"
       }
    ]
 }
--- a/renovate.json
+++ b/renovate.json
@@ -1,12 +1,19 @@
 {
-  "extends": [
-    "config:base",
-    "helpers:pinGitHubActionDigests"
-  ],
-  "packageRules": [
-    {
-      "packageNames": ["org.mockito:mockito-core"],
-      "schedule": ["on monday"]
-    }
-  ]
+   "extends":[
+      "config:base",
+      "helpers:pinGitHubActionDigests"
+   ],
+   "packageRules":[
+      {
+         "matchUpdateTypes":[
+            "minor",
+            "patch",
+            "pin",
+            "digest",
+            "fix"
+         ],
+         "automerge":true,
+         "semanticCommitType":":arrow_up:"
+      }
+   ]
 }


renovate.json
```diff
@@ -1,12 +1,19 @@
...
+      "helpers:pinGitHubActionDigests"
+   ],
+   "packageRules":[
+      {
+         "matchUpdateTypes":[
+            "minor",
+            "patch",
+            "pin",
+            "digest",
+            "fix"
```
algomaster99, line 13: `fix` is not an allowed value based on the documentation over [here](https://docs.renovatebot.com/configuration-options/#matchupdatetypes).

slarse:
Just for clarity, could we include in the commit message on master that this does not apply to major version updates? Something like "Enable renovate automerge for non-major updates".

MartinWitt:
> Just for clarity, could we include in the commit message on master that this does not apply to major version updates? Something like "Enable renovate automerge for non-major updates".

Yes I can do this, but this is also an open point for discussion. Shall we automerge major version changes?

slarse:
> > Just for clarity, could we include in the commit message on master that this does not apply to major version updates? Something like "Enable renovate automerge for non-major updates".
> 
> Yes I can do this, but this is also an open point for discussion. Shall we automerge major version changes?

I would spontaneously say no. Even if our tests pass, there may be problems for consumers when major library versions change. I think a major version upgrade always warrants at least a cursory inspection of the consequences of adopting it.

algomaster99:
> warrants at least a cursory inspection of the consequences of adopting it

It is hard to narrow down which part of the code would be affected by the dependency update, especially on such a big codebase. So I am not sure where I would look for possible failures.

Spoon heavily relies on eclipse JDT so if there is an incompatible change, our tests would most likely detect it. But for other dependencies, our tests may not detect it, but I imagine it would be rare.

MartinWitt:
Okay, let's start without major versions and see how it goes? We can change this any moment if we believe merging majors is useful too.

renovate.json
```diff
@@ -1,12 +1,18 @@
...
+   "packageRules":[
+      {
+         "matchUpdateTypes":[
+            "minor",
+            "patch",
+            "pin",
+            "digest"
+         ],
+         "automerge":true,
+         "semanticCommitType":":arrow_up:"
```
SirYwell, line 15: This needs to be adapted to spoon's commit conventions I guess?

MartinWitt: https://docs.renovatebot.com/semantic-commits/#semantic-commit-messages Given this doc I should remove this line completely tbh

slarse (approved):
Let's go? Any objections @SirYwell @I-Al-Istannen @algomaster99 ?

algomaster99 (approved):

SirYwell (approved):

I-Al-Istannen (approved):

monperrus:
Really nice, thanks a lot @MartinWitt and all for your inputs

MartinWitt:
Currently, renovate is not doing any auto-merge. I believe a config change only affects new pull requests. In the pull requests, the lights for the auto-merge are green, so our config is correct. Lets merge them and see if it works for new pull-requests?

slarse:
@MartinWitt It feels like something is wrong, because if you have a look at the recent PRs they say "Automerge: Enabled". Maybe some permission is missing for the bot s.t. it cannot automerge?

slarse:
Don't we have "Pull request required" enabled? I'm not boss enough to check the branch protection rules but seems like that would be a show-stopper (for some reason??).

https://docs.renovatebot.com/key-concepts/automerge/#pull-requests-required

PR #5189, opened 2023-04-25
feat: Add native transport configuration to Maven 

MartinWitt:
Add system properties to Maven options to set the native dependency transport for Aether connector. This will improve the speed of dependency resolution by using the native transport layer over HTTP. Also, the request timeout and connection timeout are increased to 300,000 milliseconds in order to allow for slower networks.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -17,6 +17,10 @@ on:
 
 env:
   JAVA_DISTRIBUTION: temurin
+  MAVEN_OPTS: >-
+    -Dmaven.resolver.transport=native
+    -Daether.connector.connectTimeout=300000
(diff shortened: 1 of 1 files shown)


MartinWitt:
Had a lengthy discussion with @I-Al-Istannen and @SirYwell about our failing master and how to fix this. We concluded to switch to mavens new http transport and increase the timeout. This should reduce/remove the failing of mavens dependency resolution on master. The timeout is still set to a reasonable number, but we will monitor it closely and increase if we see we still need a large.

algomaster99:
> The / in front of the branch name simple killed GitHub, don't do it.

What happened? :thinking: Were you having some problems merging the branch?

PR #5190, opened 2023-04-25
chore: Use native transport configuration for Maven in CI

MartinWitt:
replaces #5189 Somehow the GitHub API allows creating branches which destroy GitHub. Having a `/` leading your branch name is not the best idea.

Had a lengthy discussion with @I-Al-Istannen and @SirYwell about our failing master and how to fix this. We concluded to switch to mavens new http transport and increase the timeout. This should reduce/remove the failing of mavens dependency resolution on master. The timeout is still set to a reasonable number, but we will monitor it closely and increase if we see we still need a large.

Add system properties to Maven options to set the native dependency transport for Aether connector. This will improve the speed of dependency resolution by using the native transport layer over HTTP. Also, the request timeout and connection timeout are increased to 300,000 milliseconds in order to allow for slower networks.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -40,7 +40,7 @@ jobs:
 
 
     env:
-      MAVEN_OPTS: -Djava.src.version=${{ matrix.java }} -Dhttp.keepAlive=false -Dmaven.wagon.http.pool=false
+      MAVEN_OPTS: -Djava.src.version=${{ matrix.java }}
 
     name: Tests with Java ${{ matrix.java }} on ${{ matrix.os }}
     steps:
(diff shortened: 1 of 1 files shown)


MartinWitt:
The removed env var `java.src.version` was only used in the spoon pom if there exists no `java.version` property. This is never the case in test cases and if the property is null, we fall back to the way better maven property https://github.com/INRIA/spoon/blob/18b058ade2889330d42dd09e25f24146bb1a8f96/src/main/java/spoon/support/compiler/SpoonPom.java#L393

I-Al-Istannen (approved):
I ran it twice without caches and once with them and it succeeded every time. That's a lot better than any other pipeline I've seen in the last ~two weeks.

MartinWitt:
Was a nice ride with wagon. Let's hope the new HTTP native transport is better.

I-Al-Istannen:
Let's cross our fingers! Thanks :)

PR #5198, opened 2023-05-01
deps: remove outdated dependency that cannot be resolved anymore

monperrus:
Fixes one problem found by Renovate in https://github.com/INRIA/spoon/issues/4143

Could be considered as debloat @cesarsotovalero

### DIFF
--- a/spoon-pom/pom.xml
+++ b/spoon-pom/pom.xml
@@ -279,40 +279,6 @@
                     <version>3.0.0</version>
                 </plugin>
 
-                <!--This plugin's configuration is used to store Eclipse m2e settings
-                  only. It has no influence on the Maven build itself. -->
-                <plugin>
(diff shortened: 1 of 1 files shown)


monperrus:
@I-Al-Istannen any idea why this breaks the reproducible build pipeline?

I-Al-Istannen:
No. It also failed in 4 seconds, which isn't enough to do basically anything in the script. It works after a rerun. I would guess it died during the maven build (where it throws away the log), probably due to the github connectivity issues we have seen before? Sounds like a fluke for now, we could migrate it to native http if it occurs more often.

I-Al-Istannen (approved):

monperrus:
thanks @I-Al-Istannen 

renovate is happy and has updated the dashboard

Issue #5199, opened 2023-05-02
[Bug]: Crash when calling `CtFieldReference#getActualField` on a reference to the array's length: The field `int#length` not found


Luro02:
### Describe the bug

A crash occurs when you try to execute the code below.

The `getActualField` method is not used by our code, but indirectly called through `getModifiers`.
This is the code in spoon that causes the crash:
https://github.com/INRIA/spoon/blob/194f91a7e963901a7034fbe1bd18c682a461e766/src/main/java/spoon/support/reflect/reference/CtFieldReferenceImpl.java#L56-L80

The `declaringType` for the field reference to the array's `length` attribute is set to `int`, which seems to be the root cause for this issue?

As a side-note: The inlined instantiation of the array is not the reason for the crash. This code will produce the same exception:
```java
String[] array = new String[1];
int value = array.length;
```

### Source code you are trying to analyze/transform

```Java
public class Example {
    public static void main(String[] args) {
        int value = ((new String[1])).length;
    }
}
```

### Source code for your Spoon processing

```Java
Launcher launcher = new Launcher();

launcher.getEnvironment().setComplianceLevel(17);
launcher.addInputResource(new VirtualFile("""
    public class Example {
        public static void main(String[] args) {
            int value = ((new String[1])).length;
        }
    }
    """));

    CtModel ctModel = launcher.buildModel();

    ctModel.processWith(new AbstractProcessor<CtVariableAccess<?>>() {
        @Override
        public void process(CtVariableAccess<?> element) {
            if (!(element.getVariable() instanceof CtFieldReference<?> ctFieldReference)) {
                return;
            }

            ctFieldReference.getActualField(); // <- this crashes
        }
    });
```

### Actual output

```Java
Exception in thread "main" spoon.SpoonException: The field int#length not found
	at spoon.support.reflect.reference.CtFieldReferenceImpl.getActualField(CtFieldReferenceImpl.java:78)
	at org.example.playground.App$1.process(App.java:78)
	at org.example.playground.App$1.process(App.java:70)
	at spoon.support.visitor.ProcessingVisitor.scan(ProcessingVisitor.java:72)
	at spoon.reflect.visitor.CtScanner.scan(CtScanner.java:184)
	at spoon.reflect.visitor.CtScanner.visitCtLocalVariable(CtScanner.java:555)
	at spoon.support.reflect.code.CtLocalVariableImpl.accept(CtLocalVariableImpl.java:51)
[...]

PR #5206, opened 2023-05-10
review: fix: set declaring type for array field accesses

SirYwell:
Step 1 addressing #5199.

Previously, spoon set the declaring type to the type of the field if JDT didn't provide a declaring type. From running all tests, this seems to be only the case for access to the `length` field of arrays. Instead of setting the declaring type to `int` then, we now use the type binding provided by JDT that holds the actual declaring type.

Note that this does not fully fix the bug, but we get `The field java.lang.String[]#length not found` instead of `The field int#length not found`.


monperrus (approved):
Thanks a lot, LGTM

MartinWitt:
Thanks @SirYwell

PR #5236, opened 2023-05-27
review: fix: special-case getModifiers for array.length accesses

SirYwell:
Fixes the use case described in #5199, `getModifiers()` now works for the `length` field in arrays.

`getActualField()` still throws the exception, but this modelling is just similar to the reflection api.

I also added a contract and renamed the test method from my previous work on this...


MartinWitt (approved):
LGTM, 👍 @SirYwell

MartinWitt:
Thanks @SirYwell

PR #5238, opened 2023-05-28
wip: fix: make getActualField respect static fields in annotations

SirYwell:
Annotations can have static fields, so looking up a method for them fails. The fix is rather simple: If the referenced field is static, it must be an actual field in the annotation.


MartinWitt (approved):
LGTM will merge

MartinWitt:
Thanks @SirYwell

PR #5239, opened 2023-05-28
refactor: remove cast by narrowing type

MartinWitt:
Simplify file iteration loop in `Launcher.java` by changing the `resources` collection to be a `Collection<File>`, and using `File` instead of `Object` in the loop, which removes the requirement of casting it before converting it to a `Path`.


SirYwell (approved):
Good change

MartinWitt:
We could in a further PR replace all this code with java 11 NIO.

PR #5242, opened 2023-05-30
review: refactor: reduce complexity of setInputClassLoader

MartinWitt:
I removed the unneeded variable and converted it to a directly thrown Exception.


SirYwell (approved):
Nice simplification!

PR #5244, opened 2023-05-31
review: fix: refactor visitCtForEach method

MartinWitt:
Narrow instanceof from CtVariable to CtLocalVariable aligning it with the cast afterwards.
Improve if, else construct with else at the end.


SirYwell (approved):
Looks good

PR #5283, opened 2023-06-16
Fix pom.xml for JReleaser Maven release.

MartinWitt:


spoon-pom/pom.xml
```diff
@@ -427,8 +424,57 @@
...
+                    </plugin>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-deploy-plugin</artifactId>
+                        <version>3.1.1</version>
+                    </plugin>
+                    <plugin>
+                        <groupId>org.jacoco</groupId>
+                        <artifactId>jacoco-maven-plugin</artifactId>
+                        <version>0.8.10</version>
```
I-Al-Istannen, line 468: If you just change the config, do you need the version here?

MartinWitt: I wanted to be safe because I am completely unsure what maven is doing and already debugged a maven version disaster a few days ago.

I-Al-Istannen: Then we hope renovate also finds that particular instantiation :^)

monperrus:
thanks for the progress.

I suggest that we touch the least the parent pom, this could have unanticipated effects on other modules, and on the same line, that we don't touch the existing `release` profile (it should continue to work for manually released submodules)

what about putting the jreleaser profile in the spoon- pom.xml (and not in spoon-parent)?

I-Al-Istannen:
Well, we do not release spoon-pom currently, so this will have no effect on downstream consumers :^) [We should really fix that]

We also don't release any other modules as far as I am aware? If we want to release the javadoc submodule automatically as well, it would be nice if the different submodules could share such code.

The `jreleaser` profile should not be active by default, so I don't think this should affect other submodules? The only real change seems to be moving the `nexus-staging-maven-plugin` to only be active in the `release` profile?

MartinWitt:
> Well, we do not release spoon-pom currently, so this will have no effect on downstream consumers :^) [We should really fix that]

On my agenda after publishing spoon-core. Also, the Javadoc parser should be released in the same way as you said.

> what about putting the jreleaser profile in the spoon- pom.xml (and not in spoon-parent)?

As @I-Al-Istannen said, the profile is only active if you use the profile flag and we really need this reusable profile for the other modules/parent-pom.

monperrus:
ack, merged and started.

MartinWitt:
The first run failed because the branch was still existent. I deleted it and restarted the run.

PR #5287, opened 2023-06-18
review: doc: highlight for CtRecordComponent that return values are views 

### LINKED ISSUE #5265: [Bug]: `CtRecordComponent::toField` adds new fields to the model with broken parents
### Describe the bug

My code queries the `CtModel` for certain references and then tries to find the first common parent between all references. For this to work, it requires that all `CtReference`s have at least the same root (should be the unnamed module). I assumed this is always the case, but the code below has two references for which this is not the case.

The call to `CtRecordComponent::toField` seems to be the main problem, without this call or by setting the parent explicitly (`ctField.setParent(ctRecord)`) the code works as expected. It is surprising to me that the `toField` call has side-effects on the model.

```Java
public class Main {
    protected static final String PLACEHOLDER = "";

    record Name(String name) {}
}
```

```Java
public final class Main {
    private Main() {
    }

    // returns all parents of a CtElement
    private static Iterable<CtElement> parents(CtElement ctElement) {
        return () -> new Iterator<>() {
            private CtElement current = ctElement;

            @Override
            public boolean hasNext() {
                return this.current.isParentInitialized();
            }

            @Override
            public CtElement next() throws NoSuchElementException {
                if (!this.hasNext()) {
                    throw new NoSuchElementException("No more parents");
                }

                CtElement result = this.current.getParent();
                this.current = result;
                return result;
            }
        };
    }

    private static void run() {

        launcher.addInputResource(new VirtualFile("""
            public class Main {
                protected static final String PLACEHOLDER = "";

                record Name(String name) {}
            }
            """));

        CtRecord ctRecord = ctModel.getElements(new TypeFilter<>(CtRecord.class)).get(0);
        for (CtRecordComponent component : ctRecord.getRecordComponents()) {
            CtField<?> ctField = component.toField();
            System.out.println(ctField);
        }

        CtClass<?> exampleClass = ctModel.getElements(new TypeFilter<>(CtClass.class)).get(0);

        List<CtReference> references = ctModel.getElements(new DirectReferenceFilter<>(exampleClass.getField("PLACEHOLDER").getType()));

        for (CtReference ctReference : references) {
            List<CtElement> path = new ArrayList<>();
            parents(ctReference).forEach(path::add);
            if (path.size() == 1) {
                // this prints the references with broken parents
                System.out.println(
                    "parents('%s'): [%s]".formatted(
                        ctReference,
                        path.stream()
                            .map(CtElement::prettyprint)
                            .collect(Collectors.joining(", "))
                    )
                );
            }
        }
    }

        run();
    }
}
```

### Actual output

```Java
java.lang.String name;
parents('java.lang.String'): [String name;]
parents('java.lang.String'): [String name;]
```

### Expected output

```Java
java.lang.String name;
```

### Spoon Version

10.3.0

### JVM Version

Windows 10

MartinWitt:
Update the comments of the `toMethod()` and `toField()` functions in the `CtRecordComponent.java` file to improve the view of the corresponding methods and fields. This change improves the readability of previously confusing aspects of the code while highlighting ways to view specific fields and methods.

Fixes #5265


src/main/java/spoon/reflect/declaration/CtRecordComponent.java
```diff
@@ -19,14 +19,22 @@
 public interface CtRecordComponent extends CtTypedElement<Object>, CtNamedElement, CtShadowable {
 
 	/**
-	 * Converts the component to an implicit method.
-	 * @return the method
+	 * Converts the component to an implicit method. The returned method is a view and has no parent.
+	 * This means that any modification on the returned method will not be reflected on the component.
+	 * Also this element <b> is not </b> part of the model. A record already has the methods corresponding to its components.
```
I-Al-Istannen, line 24:
```suggestion
	 * Also this element <strong>is not</strong> part of the model. A record already has the methods corresponding to its components.
```
Semantic HTML

src/main/java/spoon/reflect/declaration/CtRecordComponent.java
```diff
@@ -19,14 +19,22 @@
...
+	 * @return the method corresponding to the component (a getter) as a view.
 	 */
 	CtMethod<?> toMethod();
 
 	/**
-	 * Converts the component to an implicit field.
-	 * @return  the field
+	 * Converts the component to an implicit field.The returned field is a view and has <b>no</b> parent.
+	 * This means that any modification on the returned field will not be reflected on the component.
+	 * Also this element <b> is not </b> part of the model. A record already has the field corresponding to its components.
```
I-Al-Istannen, line 34:
```suggestion
	 * Also this element <strong>is not</strong> part of the model. A record already has the field corresponding to its components.
```

src/main/java/spoon/reflect/declaration/CtRecordComponent.java
```diff
@@ -19,14 +19,22 @@
...
+	 * Use {@link CtRecord#getFields()} to get the field of a record.
```
I-Al-Istannen, line 35:
```suggestion
	 * Use {@link CtRecord#getFields()} to get the fields of a record.
```

src/main/java/spoon/reflect/declaration/CtRecordComponent.java
```diff
@@ -19,14 +19,22 @@
...
+	 *
+	 * @return  the field corresponding to the component as a view.
```
I-Al-Istannen, line 37:
```suggestion
	 * @return the field corresponding to the component as a view.
```

I-Al-Istannen (changes requested):

I-Al-Istannen (approved):

PR #5300, opened 2023-06-21
avoid version range: makes build is not stable

hboutemy:
unstable build is even worse than non-reproducible build

and FYI, Spoon 10.3.0 ingested Maven 4.0.0-alpha-4 components: did you really mean to inject alpha components?

see https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/fr/inria/gforge/spoon/spoon-core/README.md

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -73,7 +73,7 @@
     <dependency>
       <groupId>org.apache.maven</groupId>
       <artifactId>maven-model</artifactId>
-      <version>[3.6.0,)</version>
+      <version>3.9.2</version>
     </dependency>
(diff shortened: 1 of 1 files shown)


MartinWitt:
Hey, thank you for the PR. Could you change the version to the newest stable released, non maven 4, maven version? This should be `3.9.2`

hboutemy:
done

MartinWitt:
Thanks, @hboutemy. Yeah, spoon looks terrible in this list because we never had releases with a commit on master. We are currently fixing this.

PR #5306, opened 2023-06-22
fix: use old sonatype url

MartinWitt:
Spoon has a pre2021 URL. Also the project name should be `spoon` instead of ´spoon-core´

### DIFF
--- a/jreleaser.yml
+++ b/jreleaser.yml
@@ -40,6 +40,8 @@ deploy:
     nexus2:
       maven-central:
         active: ALWAYS
+        # Spoon is hosted on the legacy sonatype instance, see
+        # https://central.sonatype.org/publish/publish-guide/#releasing-to-central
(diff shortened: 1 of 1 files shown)


jreleaser.yml
```diff
@@ -40,7 +40,7 @@ deploy:
     nexus2:
       maven-central:
         active: ALWAYS
-        url: https://s01.oss.sonatype.org/service/local
+        url: https://oss.sonatype.org/service/local
```
I-Al-Istannen, line 43:
```suggestion
        # Spoon is hosted on the legacy sonatype instance, see
        # https://central.sonatype.org/publish/publish-guide/#releasing-to-central
        url: https://oss.sonatype.org/service/local
```

I-Al-Istannen (changes requested):

I-Al-Istannen (approved):

PR #5315, opened 2023-06-27
chore: use correct groupID for staging repo

MartinWitt:
Looks like spoon has `fr.inria.gforge` as groupdId for our staging repository instead of `fr.inria.gforge.spoon`. This looks like a mistake from the old days, but we cannot change this (and shouldnot!)


I-Al-Istannen (approved):
Seems to be working based on a snapshot test release :)

Issue #5324, opened 2023-07-04
[Bug]: Parsing Java sources with module-info.java fails


kaklakariada:
### Describe the bug

I am using the Spoon java lib to parse Java sources. When I add a `module-info.java` file with content `module dummy.module {requires error.reporting.java;}` parsing fails with the following exception:

```
Caused by: spoon.compiler.ModelBuildingException: error.reporting.java cannot be resolved to a module at /path/to/project/src/main/java/module-info.java:1
```

The library jar [com.exasol:error-reporting-java](https://search.maven.org/artifact/com.exasol/error-reporting-java/1.0.1/jar) I am referencing contains a `module-info.class` (see [source](https://github.com/exasol/error-reporting-java/blob/main/src/main/java/module-info.java)). I configure the library JAR using Spoon's `Environment.setSourceClasspath()`.

Using an empty `module-info.java` file without `requires` works, Spoon parses the sources without a problem.

Does Spoon support the module path at all? I didn't find any relevant configuration option in `Environment`.

### Source code you are trying to analyze/transform

```Java
// src/main/java/module-info.java:
module dummy.module {
    requires error.reporting.java;
}

// src/main/java/mypkg/Main.java:
class Dummy {
}
```

### Source code for your Spoon processing

```Java
final SpoonAPI spoon = new Launcher();
final var environment = spoon.getEnvironment();
environment.setSourceClasspath(this.classPath);
environment.setNoClasspath(false);
environment.setComplianceLevel(11);
for (final Path path : pathsToCrawl) {
    spoon.addInputResource(path.toString());
}
spoon.buildModel();
```

### Actual output

```Java
Caused by: spoon.compiler.ModelBuildingException: error.reporting.java cannot be resolved to a module at /path/to/project/src/main/java/module-info.java:1
    at spoon.support.compiler.jdt.JDTBasedSpoonCompiler.report (JDTBasedSpoonCompiler.java:640)
    at spoon.support.compiler.jdt.JDTBasedSpoonCompiler.reportProblems (JDTBasedSpoonCompiler.java:622)
    at spoon.support.compiler.jdt.JDTBasedSpoonCompiler.build (JDTBasedSpoonCompiler.java:119)
    at spoon.support.compiler.jdt.JDTBasedSpoonCompiler.build (JDTBasedSpoonCompiler.java:100)
    at spoon.Launcher.buildModel (Launcher.java:782)
```

### Expected output

```Java
Successful parsing of Java sources
```

### Spoon Version

[...]

PR #5325, opened 2023-07-04
experiment: use module-path and see what breaks

SirYwell:
todo


MartinWitt:
Can we add every classpath file(?) that is a jar to the modulepath given source version > 9?

SirYwell:
> Can we add every classpath file(?) that is a jar to the modulepath given source version > 9?

According to https://github.com/INRIA/spoon/issues/5324#issuecomment-1623328261, I'd assume that's not possible. I also think it's better to have a clear distinction in the api without magic happening in the background.

src/main/java/spoon/support/compiler/jdt/JDTBasedSpoonCompiler.java
```diff
@@ -366,8 +376,13 @@ protected boolean buildTemplates(JDTBuilder jdtBuilder) {
...
 	 */
-	protected boolean buildUnitsAndModel(JDTBuilder jdtBuilder, SpoonFolder sourcesFolder, String[] classpath, String debugMessagePrefix) {
-		CompilationUnitDeclaration[] units = buildUnits(jdtBuilder, sourcesFolder, classpath, debugMessagePrefix);
+	protected boolean buildUnitsAndModel(
+					JDTBuilder jdtBuilder,
+					SpoonFolder sourcesFolder,
+					String[] classpath,
+					List<String> modulePath,
+					String debugMessagePrefix) {
+		CompilationUnitDeclaration[] units = buildUnits(jdtBuilder, sourcesFolder, classpath, modulePath, debugMessagePrefix);
```
MartinWitt, line 385: Isn't this breaking if someone used this class?

src/main/java/spoon/compiler/Environment.java
```diff
@@ -306,6 +307,24 @@ public interface Environment {
...
+	List<String> getSourceModulePath();
+
+	/**
+	 * Sets the module path that is used to build/compile the input sources.
+	 *
+	 * @param sourceModulePath The new module path to be set. Each string element
+	 *                         should be the path to a directory or a module jar file.
+	 * @throws NullPointerException if the argument is null or an element of the list is null.
+	 */
+	void setSourceModulePath(List<String> sourceModulePath);
```
MartinWitt, line 326: Clarify that a module in our context is a jar with a `module.info`.

MartinWitt (changes requested):
Thank you openai for this nice Javadoc. Could yo fix the rest @SirYwell

src/main/java/spoon/support/compiler/jdt/JDTBasedSpoonCompiler.java
MartinWitt: The question is, is it valuable to add a method without this parameter calling `buildUnitsAndModel` with an empty list. Do you believe anyone directly calls this method?

SirYwell:
@MartinWitt reminder

MartinWitt (approved):

PR #5333, opened 2023-07-07
review: chore: convert extra checks to github action

MartinWitt:


MartinWitt:
We should really convert this to a github action.

.github/workflows/tests.yml
```diff
@@ -132,11 +131,80 @@ jobs:
...
+
+      - name: Check documentation links
+        run: python3 ./chore/check-links-in-doc.py
+
+      - name: Analyze dependencies through DepClean in spoon-core
+        run: mvn -q depclean:depclean
+
+      - name: Spoon-decompiler
+        run: |
+          cd spoon-decompiler
```
I-Al-Istannen, line 161: How about using the `working-directory` step option for such things: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsrun?

I-Al-Istannen (changes requested):

.github/workflows/tests.yml
MartinWitt: I convert this. Also I remove the bash script

.github/workflows/tests.yml
```diff
@@ -48,7 +48,7 @@ jobs:
       - uses: oracle-actions/setup-java@2c4df2930e35870536667f383d87f1246fbe613f # v1
         with:
           website: jdk.java.net
-          java-version: ${{ matrix.java }}
+          version: ${{ matrix.java }}
```
MartinWitt, line 51: The oracle action uses version and not java-version.

SirYwell (commented):
I think this is a great improvement to the CI. I have a few things that could be considered to further improve it (probably here, probably in separate PRs):

1. Most steps don't depend on each other. What do you think about having separate jobs for them?
2. We download some files (the z3 build, gradle). Can we make use of caching?
3.  It seems like most sub projects run similar maven commands, however, it's somewhat inconsistent (some run depclean, some run checkstyle, some run both). Is this something we want to improve?

.github/workflows/tests.yml
MartinWitt: Okay lets fix this in another PR.

MartinWitt:
> I think this is a great improvement to the CI. I have a few things that could be considered to further improve it (probably here, probably in separate PRs):
> 
> 1. Most steps don't depend on each other. What do you think about having separate jobs for them?

Actually, most jobs need to install spoon-core/spoon-pom before. We dont want to run `mvn install` to often, as this takes time.

> 2. We download some files (the z3 build, gradle). Can we make use of caching?

Sure
> 3. It seems like most sub projects run similar maven commands, however, it's somewhat inconsistent (some run depclean, some run checkstyle, some run both). Is this something we want to improve?

I hope there is some reasoning behind this

MartinWitt:
The z3 download is cached now.

I-Al-Istannen (approved):

Issue #5343, opened 2023-07-12
[Question]: Is there a way to set the character encoding of processed source code?


tenax66:
When processing source code as follows, the character encoding of output files depends on the environment.

```java
Launcher l = new Launcher();
l.addInputResource("src/main/java");
l.run();
```

For example, when working in Windows (Japanese setting), the encoding of processed source code is [CP932](https://en.wikipedia.org/wiki/Code_page_932_(Microsoft_Windows)), I wanted UTF-8 code though.
Is there any simple way to set the encoding of processed source code?

In the case above, I created my own FileGenerator (which outputs files in UTF-8)  from `JavaOutputProcessor` and set it to `Environment` as a work-around. like the following:

```java
l.getEnvironment().setDefaultFileGenerator(new SomeUTF8OutputProcessor());
```

But I don't think this is a good approach.

MartinWitt:
Hey, 

have you tried using https://spoon.gforge.inria.fr/mvnsites/spoon-core/apidocs/spoon/compiler/Environment.html#setEncoding(java.nio.charset.Charset) to set the charset?

tenax66:
Yes, I tried `Environment.setEncoding(StandardCharsets.UTF_8)`.
But it only set the encoding of source code that spoon parses, not the output.

MartinWitt:
Oh, you are correct. It looks like we have no API, or I am not aware of any method for this use case. Your solution might be the correct one.

PR #5350, opened 2023-07-18
review: feat: configurable encoding of processed source code

tenax66:
This pull request is related to #5343.

With this commit, `JavaOutputProcessor` now uses `PrintStream​(File file, Charset charset)` instead of `PrintStream​(File file)`.
By default, `Charset.defaultCharset()`  is assigned to `charset`. So the default behavior remains the same.


MartinWitt (approved):

MartinWitt:
Thanks @tenax66

PR #5365, opened 2023-07-26
doc: fix typos "is is..."

tenax66:
There are typos "when *is* is hard..." in several classes, fixed them to "it is".


MartinWitt:
Thanks @tenax66

PR #5370, opened 2023-07-27
review: build: Add nullaway from uber

MartinWitt:
Adds nullway from uber and set the logging leveling to warning. This currently assumes our packages are correct annotated, which they are not. I disabled all errorprone checks as we don't need two tools checking the same, we only want nullaway.


SirYwell:
example
Example from the CI how the warnings look like.

Is there additional tooling support, e.g. settings to enable in IntelliJ to directly see such warnings?

MartinWitt:
Sadly the errorprone team is strictly against creating a tool with good useability. The best we have is some community tooling see https://github.com/google/error-prone/issues/444.

MartinWitt:
As I look more into it, this view is more or less never needed and only required if a person wants to make a PR for nullablility. Qodana will report nullable issues if we add more annotations and has a goood ui.

PR #5374, opened 2023-07-31
review: refactor(SpoonifierVisitor): `Deque` should be used instead of `Stack`

tenax66:
Unless there's particular reason, `Deque` should be used instead of `Stack` (like other part of spoon code).


SirYwell (approved):
Looks good, will merge.

SirYwell:
Thanks!

Issue #5320, opened 2023-07-01
Spoon and Nullable


MartinWitt:
Currently, spoon has non `@Nullable` annotation and we only document nullablility in the Javadoc. Why don't we adopt https://jspecify.dev/docs/start-here and migrate over the time to an annotated library? An alternative is `@Contract` from JB, but these are only supported in IJ and are way harder to write.

MartinWitt:
We start with 0.3 and can update it later. The goal is to start annotating spoon and waiting for the 1.0 release can take years.

MartinWitt:
A noteworthy alternative is using https://www.jetbrains.com/help/idea/annotating-source-code.html. These annotations are at least stable and do the same. But there could be some future where everyone uses jspecify.

tenax66:
> A good starting point are methods with null documented in their javadoc for return values.

I will see what I can do about this.

tenax66:
I looked for the word "null" in `@return` tags in Javadoc of the `spoon-core` code.
Here are the counts for each package.
We may be able to start from these.

|package|count|
|----|----|
|spoon|1|
|spoon.compiler|1|
|spoon.metamodel|3|
|spoon.pattern.internal|2|
|spoon.pattern.internal.matcher|3|
|spoon.pattern.internal.node|1|
|spoon.pattern.internal.parameter|2|
|spoon.reflect.code|3|
|spoon.reflect.declaration|12|
|spoon.reflect.factory|7|
|spoon.reflect.path|1|
|spoon.reflect.reference|8|
|spoon.reflect.visitor|5|
|spoon.support.compiler|1|
|spoon.support.compiler.jdt|4|
|spoon.support.modelobs|1|
|spoon.support.reflect.eval|1|
|spoon.support.reflect.reference|1|
|spoon.support.sniper.internal|3|
|spoon.support.util|2|
|spoon.support.util.internal|1|
|spoon.support.visitor|3|
|spoon.support.visitor.java|2|

tenax66:
Create a task list to add `@Nullable` annotations to the classes under each package.

- [ ] spoon
- [ ] spoon.compiler
- [ ] spoon.metamodel
- [ ] spoon.pattern.internal
- [ ] spoon.pattern.internal.matcher
- [ ] spoon.pattern.internal.node
- [ ] spoon.pattern.internal.parameter
- [ ] spoon.reflect.code
- [ ] spoon.reflect.declaration
- [ ] spoon.reflect.factory
[...]

PR #5384, opened 2023-08-13
refactor: adopt jspecify

tenax66:
This is a pull request for issue #5320 .

- add the dependency for `jspecify`.
- applied the `@Nullable` annotation to several packages as a first step:`spoon`, `spoon.compiler`, `spoon.metamodel` and `spoon.pattern.internal`.

Note: This pull request does not include checks for annotations. We may use [jspecify-reference-checker](https://github.com/jspecify/jspecify-reference-checker) in the future.


tenax66:
When annotating a return type of a method by jspecify's `@Nullable`, it should look like this.
```java
public @Nullable String foo() {
  return ""
}
```
That is to say, the annotation is applied directly to the return type.

But this does not pass Qodana's checks (MissortedModifiers).

SirYwell:
> When annotating a return type of a method by jspecify's `@Nullable`, it should look like this.
> 
> ```java
> public @Nullable String foo() {
>   return ""
> }
> ```
> 
> That is to say, the annotation is applied directly to the return type.
> 
> But annotations in that position don't pass Qodana's checks (MissortedModifiers).

This is configurable in IntelliJ, but I don't know if/how it can be configured using the `qodana.yaml`. @MartinWitt maybe you know more about that?

MartinWitt:
Hey,

The current qodana config does not allow this directly, but you could try
adding the .idea config file to the repo after changing the seting in IJ.
If this does not work disabling the rule with a comment why ist also fine
for me.
Otherwise great progress.

Hannes Greule ***@***.***> schrieb am So., 13. Aug. 2023,
09:49:

tenax66:
> you could try adding the   `.idea` config file to the repo after changing the setting in IJ.
> If this does not work, disabling the rule with a comment why is also fine for me. 

I tried the latter, as I could not find any information that the `.idea` folder can modify qodana's CI settings.

MartinWitt:
@SirYwell can you provide the config file with the changed IJ setting?

I-Al-Istannen:
Okay, I give up. It seems to work locally, but what do I know. Let's disable that modifier.

MartinWitt:
@I-Al-Istannen so we merge and fix this at a later stage?

MartinWitt:
Thanks @tenax66 for your work.

PR #5395, opened 2023-08-25
chore: ensure in CI that there is no dependency to snapshots

### LINKED ISSUE #5340: Failed to read artifact descriptor for Spoon 10.4.1-SNAPSHOT
Dear all,

Since 10.4.1-SNAHPSHOT my projects that use Spoon fail with 

Failed to read artifact descriptor for fr.inria.gforge.spoon:spoon-core:jar:10.4.1-SNAPSHOT: Failure to find fr.inria.gforge.spoon:spoon-pom:pom:10.4.1-SNAPSHOT in https://repository.ow2.org/nexus/content/repositories/snapshots/ was cached in the local repository, resolution will not be reattempted until the update interval of [ow2.org](http://ow2.org/)-snapshot has elapsed or updates are forced

For those with access to the Spoon Inria CI server, basically all projects are down with this issue.

Are we missing something? Or is it related to the (new?) way the snapshot is deployed?

Thanks,
Lionel.

monperrus:
fix #5340

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -162,6 +162,27 @@
   </dependencies>
   <build>
     <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-enforcer-plugin</artifactId>
+        <version>1.4.1</version>
+        <executions>
+          <execution>
+            <id>enforce-no-snapshots</id>
+            <goals>
+              <goal>enforce</goal>
+            </goals>
+            <configuration>
+              <rules>
+                <requireReleaseDeps>
+                  <message>No Snapshots Allowed!</message>
+                </requireReleaseDeps>
+              </rules>
+              <fail>true</fail>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
         <plugin>
             <groupId>org.jacoco</groupId>
             <artifactId>jacoco-maven-plugin</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -12,7 +12,7 @@
   <parent>
     <groupId>fr.inria.gforge.spoon</groupId>
     <artifactId>spoon-pom</artifactId>
-    <version>10.4.1</version>
+    <version>10.4.0</version>
     <relativePath>spoon-pom</relativePath>
   </parent>
 
--- a/pom.xml
+++ b/pom.xml
@@ -12,7 +12,7 @@
   <parent>
     <groupId>fr.inria.gforge.spoon</groupId>
     <artifactId>spoon-pom</artifactId>
-    <version>10.4.1-SNAPSHOT</version>
+    <version>10.4.1</version>
     <relativePath>spoon-pom</relativePath>
   </parent>
 


I-Al-Istannen:
Changing the spoon-pom version is done by the release workflow now. If we do not bump that to a new snapshot version, we are unable to change anything in there during development, as spoon-core will use the old, released version.
The release commit (e.g. https://github.com/INRIA/spoon/commit/c41c4076a1c41c6db3fe7f173a20702014913a97) should use a proper non-snapshot version, but I do not see a nice way to maintain that property during development.

What is the reason for this change? Is it necessary, that spoon snapshot versions (i.e. currently being worked on) never refer to spoon in a snapshot state? I would think that only the final release version must point to fixed, stable version numbers.

monperrus:
> What is the reason for this change? I

The primary reason is that this is the cause for which many jobs our integration CI (checking that client code still works on SNAPSHOT) are broken: https://ci.inria.fr/sos/. The underlying root cause is that `spoon-parent` is not snapshot CDed.

To fix the integration CI, I see two options:
- this one
- set up snapshot CD for `spoon-parent`

I prefer this one because 1) it minimizes coupling, 2) this check is very meaningful and useful in general, beyond this current breakage.

I-Al-Istannen:
The problem with your solution is that changes in spoon-pom (e.g. to update dependencies) will be built **in the release commit** for the first time.

As an example, consider this sequence of events:
1. Renovate changes a version in spoon-pom to a new, **breaking** release
2. Our CI builds spoon-core, but it references the released and published older spoon-pom *without the change*
3. The CI subsequently passes
4. When we make a new spoon release, the release commit breaks and we have to untangle the dependency mess, potentially downgrading things again


To ensure changes in spoon-pom are directly applied, we would need to leave the version of the spoon-pom in this repository the same after a release. This is not a good idea though, IMHO, because then we have multiple spoon-poms with the same version tag but different contents.

---

Btw, https://ci.inria.fr/sos/ seems to only show builds if you are logged in — and I don't believe I have an account :)

monperrus:
> As an example, consider this sequence of events:

This would work, the CI would build what's written in the code for a given commit. And then Renovate would update spoon-parent one's newly released.

> To ensure changes in spoon-pom are directly applied

Yes, we would exclude this workflow on purpose. We would change spoon-pom, release it and then update in another commit. On this topic, I've created long time overdue #5397

monperrus:
any idea why and how github-actions[bot] reacted with thumbs up emoji?

I-Al-Istannen:
> This would work, the CI would build what's written in the code for a given commit. And then Renovate would update spoon-parent one's newly released.

I am not sure I follow. How would the changes in spoon-pom be picked up by spoon-core, if spoon-core refers to an old version of the parent? I think spoon-core would either use the *previous* versions spoon-pom (if we only update the parent after the release) or might fail in the release commit for the first time.

I am not really keen on the version of the parent and core going out of sync in a way that delays dependency errors until a release happens. Maybe I am just missing something?

> any idea why and how github-actions[bot] reacted with thumbs up emoji?

Qodana found no issues (and used the github actions token to create the reaction)

monperrus:
> How would the changes in spoon-pom be picked up by spoon-core, if spoon-core refers to an old version of the parent?

It would not, and that's exactly the point. If we want to change the parent, the workflow is:
- change parent pom
- review and merge change in parent
- release parent
- change parent relationship info clients, incl. spoon-core
- review and merge change in clients

Fully reproducible dependency.  The parent always point to an addressable, immutable, reproducible dependency (think NIX :).

On a related note, #5397 is the best option for making everything explicit with the same properties.

PR #5399, opened 2023-08-27
fix: Use wildcard type in CtUnaryOperator get/setOperand

### LINKED ISSUE #5349: `CtUnaryOperator`s `getOperand` and `setOperand` should return/set `CtExpression<?>`
The `setOperand` method of a `CtUnaryOperator<T>` requires that a `CtExpression<T>` is passed.

This is confusing, because one can for example write this:
```java
short s = (short) 1;
int i = ~s;
```
Here `~s` is `CtUnaryOperator<Integer>`, but the operand `s` is a `CtExpression<Short>`.

Except for being confusing, this does not result in code crashing, because generics are erased at runtime.

For example this test-case shows that the types are really `Integer` for the `CtUnaryOperator` and `Short` for the operand:
```java
class Example {
    @Test
    void testsCtUnaryOperatorExpressionType() {
        CtUnaryOperator<Integer> ctUnaryOperator = Launcher.parseClass(
            "public class Test { public void f() { short s = (short) 1; int i = ~s; } }"
        ).getElements(new TypeFilter<CtUnaryOperator<Integer>>(CtUnaryOperator.class)).get(0);

        assertEquals(
            ctUnaryOperator.getFactory().Type().INTEGER_PRIMITIVE,
            ctUnaryOperator.getType()
        );

        assertEquals(
            ctUnaryOperator.getFactory().Type().SHORT_PRIMITIVE,
            ctUnaryOperator.getOperand().getType()
        );
    }
}
```
and IntelliJ marks the `assertEquals` as problematic, because the generics are differing:
image

One should note that for `CtBinaryOperator` a `CtExpression<?>` is returned and set (which is correct).
[See here](https://spoon.gforge.inria.fr/mvnsites/spoon-core/apidocs/spoon/reflect/code/CtBinaryOperator.html#setLeftHandOperand(spoon.reflect.code.CtExpression))

I-Al-Istannen:
Closes #5349

This breaks source compatibility for `getOperand`, if a user assigned it to anything but `CtExpression<?>` or a rawtype. Currently we just lie about the type which easily leads to heap pollution.


monperrus (approved):
Good, thanks @I-Al-Istannen

PR #5402, opened 2023-08-28
chore: convert qodana config to new format

MartinWitt:
This PR enables more or less all java rules. We could cut them down a bit if we see some we really hate. But as they are adding more and more nice rules, it is real hard to keep the file up-to-date otherwise.


MartinWitt:
I have added some code to show that Qodana still works

PR #5403, opened 2023-08-28
fix: Synchronize access to anonymous type cache

I-Al-Istannen:
This fixes a concurrent modification exception when parallel queries try to look up the same anonymous types.

CC: @Luro02


monperrus:
:+1:

PR #5404, opened 2023-08-28
chore: convert qodana config to new format

MartinWitt:
Need this for access to repo secrets.

### DIFF
--- a/.github/workflows/qodana-cloud.yaml
+++ b/.github/workflows/qodana-cloud.yaml
@@ -4,7 +4,7 @@
 # risk of someone injecting malicious code into a release and then simply
 # changing a tag.
 
-name: Qodana
+name: Qodana-Cloud-Scan
 on:
   workflow_dispatch:
   pull_request:
--- a/.github/workflows/qodana-cloud.yaml
+++ b/.github/workflows/qodana-cloud.yaml
@@ -1,15 +1,9 @@
-# Workflow for testing spoon code quality.
-#
-# Note that actions are specified by commit hash. This is to avoid the security
-# risk of someone injecting malicious code into a release and then simply
-# changing a tag.
-
 name: Qodana-Cloud-Scan
 on:
   workflow_dispatch:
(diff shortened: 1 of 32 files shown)


MartinWitt:
Still need to disable the emote and the message. A `fixall` PR we can get daily would look like this https://github.com/INRIA/spoon/pull/5405#. We can also first start with a smaller clean PR:

MartinWitt:
The instance is now public you can have a look: `https://qodana.cloud/projects/3KNVy/reports/RLk7X`

.github/workflows/qodana-cloud.yaml
```diff
@@ -0,0 +1,30 @@
...
+    steps:
+      - uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3
+      - name: 'Qodana Scan'
+        uses: JetBrains/qodana-action@77f0ff0c702065648df9fd0340a48919dca5a1ff # v2023.2.1
+        env:
+          QODANA_TOKEN: ${{ secrets.QODANA_TOKEN }}
+        with:
+          pr-mode: false
+          post-pr-comment: false
+          args: --cleanup, -l, jetbrains/qodana-jvm:2023.2
```
MartinWitt, line 28: We use the enterprise scanner here because it needs a Qodana token, which is a secret. We can't use this for the PR workflow as PRs don't have access to the secret.

.github/workflows/qodana-cloud.yaml
```diff
@@ -0,0 +1,33 @@
...
+    branches:
+      - master
+  push:
+    branches:
+      - master
+
+permissions:
+  contents: read
+jobs:
+  qodana:
```
MartinWitt, line 16: Rename to Qodana-Cloud-Scan

I-Al-Istannen (changes requested):
Could you please disable spell checking, as I am not sure the 20k problems add much :)

PR #5406, opened 2023-08-28
fix: Use SuperAccess when implicitly accessing superclass field

### LINKED ISSUE #5221: [Bug]: `CtFieldAccess` with an implicit `super` has a target of `CtThisAccess` instead of `CtSuperAccess`
### Describe the bug

For the given code, spoon sets the target to a `CtThisAccessImpl` instead of a `CtSuperAccessImpl`. The compiler accepts both `this.a` and `super.a` if there is no attribute with the same name in this class.

Another problem is that `ctFieldRead.target.type` is set to type `B` instead of `A`.

image

I think the same problem will be present for `CtFieldWrite`s with an implicit `super`.

As a workaround I wrote the following code, which adjusts the target to be a `CtSuperAccess`:
```java
    private static CtTypeReference<?> findDeclaredType(CtFieldReference<?> ctFieldReference, CtTypeReference<?> ctTypeReference) {
        CtTypeReference<?> current = ctTypeReference;

        while (current != null) {
            if (current.getDeclaredFields().contains(ctFieldReference)) {
                return current;
            }

            current = current.getSuperclass();
        }

        return current;
    }

    private static <T> void adjustTarget(CtFieldAccess<T> ctFieldAccess) {
        if (!(ctFieldAccess.getTarget() instanceof CtThisAccess<?> ctThisAccess)) {
            return;
        }

        // check if the field is declared in the class
        if (ctThisAccess.getType().getDeclaredFields().contains(ctFieldAccess.getVariable())) {
            return;
        }

        // if not it is declared in a parent class
        // => change the target to super
        CtSuperAccess<?> ctSuperAccess = ctFieldAccess.getFactory().createSuperAccess();
        ctSuperAccess.setImplicit(true);
        ctSuperAccess.setTarget(null);
        CtVariableReference ctVariableReference = ctFieldAccess.getFactory().createLocalVariableReference();
        ctVariableReference.setSimpleName("");
        ctVariableReference.setParent(ctSuperAccess);
        ctVariableReference.setImplicit(false);
        ctVariableReference.setType(findDeclaredType(ctFieldAccess.getVariable(), ctThisAccess.getType()));

        ctSuperAccess.setVariable(ctVariableReference);
        ctSuperAccess.setType(null);
        ctSuperAccess.setParent(ctFieldAccess);

        ctFieldAccess.setTarget(ctSuperAccess);
    }
```

```Java
class A {
    int a;
}

class B extends A {
    void foo() {
        super.a = a;
    }
}
```

```Java
import spoon.Launcher;
import spoon.processing.AbstractProcessor;
import spoon.reflect.CtModel;
import spoon.reflect.code.CtFieldRead;
import spoon.support.compiler.VirtualFile;

public final class Main2 {
    private Main2() {
    }

        launcher.addInputResource(new VirtualFile("""
            class A {
                int a;
            }
                            
            class B extends A {
                void foo() {
                    super.a = a;
                }
            }
                """));

        ctModel.processWith(new AbstractProcessor<CtFieldRead<?>>() {
            @Override
            public void process(CtFieldRead<?> ctFieldRead) {
                System.out.println(ctFieldRead);
            }
        });
    }
}
```

### Actual output

_No response_

### Expected output

_No response_

### Spoon Version

10.3.0

### JVM Version

Windows 10

I-Al-Istannen:
Closes #5221


src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java
```diff
@@ -364,12 +365,32 @@ <T> CtFieldAccess<T> createFieldAccess(SingleNameReference singleNameReference)
 			if (ref.isStatic() && !ref.getDeclaringType().isAnonymous()) {
 				va.setTarget(jdtTreeBuilder.getFactory().Code().createTypeAccess(ref.getDeclaringType(), true));
 			} else if (!ref.isStatic()) {
-				va.setTarget(jdtTreeBuilder.getFactory().Code().createThisAccess(jdtTreeBuilder.getReferencesBuilder().getTypeReference(singleNameReference.actualReceiverType), true));
+				CtTypeReference<?> owningType = jdtTreeBuilder.getReferencesBuilder().getTypeReference(singleNameReference.actualReceiverType);
+				if (enclosingClassHasFieldWithName(singleNameReference)) {
```
SirYwell, line 369: It seems like `singleNameReference.actualReceiverType == singleNameReference.fieldBinding().declaringClass` does the job too, however I'm wondering if actualReceiverType is what we want here for the owning type?

MartinWitt: @I-Al-Istannen can you resolve this?

MartinWitt: @I-Al-Istannen ping

src/test/java/spoon/reflect/visitor/CtScannerTest.java
```diff
@@ -39,6 +39,7 @@
 import spoon.metamodel.MetamodelConcept;
 import spoon.reflect.code.CtFieldRead;
 import spoon.reflect.code.CtInvocation;
+import spoon.reflect.code.CtSuperAccess;
```
SirYwell, line 42: Unused import?

SirYwell (approved):

PR #5416, opened 2023-08-30
chore: Fix release scripts to at least execute jreleaser in dry run

I-Al-Istannen:
The release scripts are quite hard to test locally (as jreleaser requires credentials), but this fixes a few errors in the scripts ported from @MartinWitt's PR and my bash prowess.


MartinWitt:
I tested the snapshot deploy locally and it works. Lets merge and see what happens in the next try.

PR #5432, opened 2023-09-03
chore: Use workflow token with permissions for automerge

I-Al-Istannen:
Use the workflow token with appropriate permissions to ensure automerge can actually be enabled.


MartinWitt:
Thanks @I-Al-Istannen

PR #5444, opened 2023-09-10
chore: enable renovate nix

MartinWitt:
If the doc is correct, this should update our flake automatically <https://docs.renovatebot.com/modules/manager/nix/>. We do this to keep our dependency update process in a single place.


I-Al-Istannen:
Please also delete the update-lockfile workflow

MartinWitt:
Done

I-Al-Istannen (approved):

PR #5447, opened 2023-09-13
chore: Replace version range with specific version

I-Al-Istannen:
Maven 3 does not allow version ranges in plugin versions. This currently causes lots of failures in the SOS CI.


MartinWitt:
Just for my future self, we need to replace this 3.7 with an expression updating the plugin version otherwise LGTM.

MartinWitt (approved):

MartinWitt:
Thanks @I-Al-Istannen

PR #5456, opened 2023-09-18
refactor: Apply the `@Nullable` annotation to `spoon.support`

tenax66:
See #5320.
Apply the `@Nullable` annotation to classes in `spoon.support`.
Specifically, for the following packages:

- `spoon.support.modelobs`
- `spoon.support.sniper.internal`


MartinWitt:
Thanks @tenax66. Btw this PR is incomplete for the package but all annotations are correct. So I will merge it anyway. IntelliJ has a built-in feature to add these annotation. Maybe worth a look.

PR #5466, opened 2023-09-25
review: fix: Subtyping relationships for arrays

### LINKED ISSUE #5462: [Bug]: `ctArrayTypeReference.isSubtypeOf(...)` returns `true` if the array type is a subtype
### Describe the bug

Obviously `Integer[]` is not a `Number`.

```Java
public class Test { Integer[] field; }
```

```Java
import org.junit.jupiter.api.Test;
import spoon.Launcher;
import spoon.reflect.declaration.CtClass;
import spoon.reflect.declaration.CtField;
import spoon.reflect.reference.CtArrayTypeReference;

import static org.junit.jupiter.api.Assertions.*;

class TestSpoon {
    @Test
    void testArraySubtype() {
        CtField<?> ctField = Launcher.parseClass("public class Test { Integer[] field; }")
            .getField("field");

        CtTypeReference<?> ctArrayType = ctField.getType();

        assertTrue(ctArrayType instanceof CtArrayTypeReference<?>, "ctArrayType should be an instance of CtArrayTypeReference");
        assertFalse(
            ctArrayType.isSubtypeOf(ctField.getFactory().Type().get(java.lang.Number.class).getReference()),
            "an array of integers should not be a subtype of Number"
        );
    }
}
```

### Actual output

```Java
org.opentest4j.AssertionFailedError: an array of integers should not be a subtype of Number ==> 
Expected :false
Actual   :true
```

### Expected output

_No response_

### Spoon Version

10.4.1-SNAPSHOT

### JVM Version

Windows 10

I-Al-Istannen:
This one was a bit of a doozy.

Array subtyping needs to be special cased, as Spoon is not able to build a `CtType` for source-level arrays. Similarly, once we are given a shadow `CtType` that refers to an array type, we need to unpack to find the real type to build our hierarchy for. We can not just fall back to the java `Class#isAssignableFrom`, as the super type reference might not be shadowable.

Closes #5462.

### DIFF
--- a/src/main/java/spoon/support/adaption/TypeAdaptor.java
+++ b/src/main/java/spoon/support/adaption/TypeAdaptor.java
@@ -134,28 +134,10 @@ public class TypeAdaptor {
 	 * @return true if start is a subtype of superRef
 	 */
 	private static boolean handleArraySubtyping(CtArrayTypeReference<?> start, CtTypeReference<?> superRef) {
-		CtTypeReference<?> startInner = start.getArrayType();
-
 		// array-array subtyping
 		if (superRef instanceof CtArrayTypeReference) {
-			int startArrayDim = getArrayDimension(start.getSimpleName());
-			int superArrayDim = getArrayDimension(superRef.getSimpleName());
-			// Arrays of different shapes are not subtypes
-			if (startArrayDim != superArrayDim) {
-				return false;
-			}
-			CtTypeReference<?> superInner = ((CtArrayTypeReference<?>) superRef).getArrayType();
-
-			if (!isSubtype(startInner.getTypeDeclaration(), superInner)) {
-				return false;
-			}
-			// No generics -> All good, no further analysis needed, we can say they are subtypes
-			if (startInner.getActualTypeArguments().isEmpty() && superInner.getActualTypeArguments().isEmpty()) {
-				return true;
-			}
-			// Generics? We need to check for subtyping relationships between wildcard and other parameters
-			// (co/contra variance). Delegate.
-			return new ClassTypingContext(start).isSubtypeOf(superRef);
+			CtTypeReference<?> superInner = ((CtArrayTypeReference<?>) superRef).getComponentType();
+			return new TypeAdaptor(start.getComponentType()).isSubtypeOf(superInner);
 		}
 		// array-normal subtyping
 		// https://docs.oracle.com/javase/specs/jls/se21/html/jls-4.html#jls-4.10.3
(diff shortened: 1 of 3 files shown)


src/main/java/spoon/support/adaption/TypeAdaptor.java
```diff
@@ -150,11 +195,51 @@ public static boolean isSubtype(CtType<?> base, CtTypeReference<?> superRef) {
...
 		}
+
+		// Handle shadow array types, as we can build a CtType for any Class, which includes arrays. We need to lower
+		// them to their innermost component type and then decide subtyping based on their relationship.
+		if (base.isArray() && superRef instanceof CtArrayTypeReference<?>) {
+			if (!base.isShadow()) {
+				throw new SpoonException("There are no source level array type declarations");
+			}
+			Class<?> actualClass = base.getActualClass();
+			// Arrays of different shapes are not subtypes
```
intrigus-lgtm, line 206: This isn't true?
> new Object[0][0] instanceof Object[]
$3 ==> true

This is a consequence of 
- If S and T are both reference types, then S[] >1 T[] iff S >1 T.
- Object >1 Object[]

unless I'm missing some context here.

I-Al-Istannen: Addressed (hopefully :P). Thanks :)

src/main/java/spoon/support/visitor/java/JavaReflectionTreeBuilder.java
```diff
@@ -141,6 +136,24 @@ public <T, R extends CtType<T>> R scan(Class<T> clazz) {
...
+		if (javaPackage == null && clazz.isArray()) {
+			javaPackage = getArrayType(clazz).getPackage();
+		}
+		if (javaPackage == null) {
+			return factory.Package().getRootPackage();
+		}
+		return factory.Package().getOrCreate(javaPackage.getName());
+	}
+
+	private static Class<?> getArrayType(Class<?> array) {
```
SirYwell, line 150: The naming here is confusing, maybe something like `getInnermostComponentType` would be better?

SirYwell (changes requested):

src/main/java/spoon/support/visitor/java/JavaReflectionTreeBuilder.java
I-Al-Istannen: The rest of the code `CtArrayReference` uses the same terminology, so I opted to keep it the same. But I can adjust it, sure :)

SirYwell: I see... well, then it's fine I guess.

SirYwell (approved):

SirYwell:
Thanks!

PR #5482, opened 2023-10-04
Use prebuilt z3 package/binary

intrigus-lgtm:
Use z3-turnkey which extracts the native z3 libraries and loads them. This releases the user from compiling and installing the z3 binaries. We use 4.11.2 because it still supports glibc-2.31.

This should allow @I-Al-Istannen to remove z3 specific config from the nix packages which in turn should shorten CI duration on a cache miss.

### DIFF
--- a/spoon-dataflow/README.md
+++ b/spoon-dataflow/README.md
@@ -32,7 +32,8 @@ Check out test directory for more examples.
 In general, spoon-dataflow is capable to evaluate expressions statically, perform symbolic execution, handle control flow of a program, and so on. It also features a proper memory model, so it nicely deals with reference aliasing.
 
 ## Build and run
-In order to build spoon-dataflow you need JDK 8 or newer.
+In order to build spoon-dataflow, you will need to have JDK 11 or later, as
+this is required by the latest versions of spoon-core.
 
 Now you can go to the build directory and run the resulting jar:    
 `java -jar spoon-dataflow.jar -sources <arg...> [-classpath <arg...>]`
--- a/spoon-dataflow/README.md
+++ b/spoon-dataflow/README.md
@@ -32,26 +32,7 @@ Check out test directory for more examples.
 In general, spoon-dataflow is capable to evaluate expressions statically, perform symbolic execution, handle control flow of a program, and so on. It also features a proper memory model, so it nicely deals with reference aliasing.
 
 ## Build and run
-In order to build spoon-dataflow you need JDK 8 or newer. Also, you have to download and install Z3.
-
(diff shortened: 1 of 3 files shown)


spoon-dataflow/README.md
```diff
@@ -32,26 +32,7 @@ Check out test directory for more examples.
...
-2. Add bin directory to your LD_LIBRARY_PATH: `export LD_LIBRARY_PATH=/path/to/z3/bin`
-3. Run `./gradlew build`
-
-### macOS:
-1. Download z3-4.8.4 here https://github.com/Z3Prover/z3/releases
-2. Add bin directory to your DYLD_LIBRARY_PATH: `export DYLD_LIBRARY_PATH=/path/to/z3/bin`
-3. Run `./gradlew build`
-
-Note that you may still need to set up environment variables or java.lang.path in your IDE.
+In order to build spoon-dataflow you need JDK 8 or newer.
```
MartinWitt, line 35: Spoon in current versions needs java 11. Are you sure with this statement?

spoon-dataflow/build.gradle
```diff
@@ -23,7 +23,7 @@ repositories {
 dependencies {
     implementation group: 'fr.inria.gforge.spoon', name: 'spoon-core', version: '+'
     implementation group: 'commons-cli', name: 'commons-cli', version: '1.5.0'
-    implementation group: 'com.microsoft', name: 'z3', version: '4.8.4'
+    implementation group: 'tools.aqua', name: 'z3-turnkey', version: '4.11.2'
```
MartinWitt, line 26: Could we use the newest version? According to this there are newer ones https://mvnrepository.com/artifact/io.github.tudo-aqua/z3-turnkey/4.8.14.

spoon-dataflow/README.md
I-Al-Istannen: That was there before and works for sufficiently old versions of spoon-core. So maybe we should leave it as is?

MartinWitt: We should replace the `+` with a real version. As currently, spoon data flow isn't published, at least I am not aware of, any user needs to build it locally. The `+` gets translated to the newest resolvable version, which is hopefully a current spoon version. And this version needs 11.

I-Al-Istannen: I updated the message a while ago to state that you need java 11 due to spoon-core :)

MartinWitt: Oh sorry github was showing the old content. What a nice UI. LGTM we can merge

PR #5514, opened 2023-10-22
Update CI script inject_spoon_snapshot.py

monperrus:
Update CI script inject_spoon_snapshot.py for using new snapshot repo.


I-Al-Istannen:
Let's see if this makes SOS happy :)

Issue #5475, opened 2023-09-27
A suggestion regarding fixing the ambiguous package problem


WarpspeedSCP:
So at work, I got quite tired of seeing exceptions caused by ambiguous package names in user code. So I made a fix for the issue by merging such packages together, and deleting the smaller package.

Here's the change:
https://github.com/WarpspeedSCP/spoon/commit/1e62e4b99b286d6152ec786efcca4edbc2437501

What I'd like to explore is if it makes sense to take it forward, as this had actually helped in improving support for packages that are split across multiple source roots.

WarpspeedSCP:
there are unfortunately no tests in spoon itself but the changes have been tested within DeepSource's Java analyser.

PR #5520, opened 2023-10-25
[WIP] fix: Merge ambiguous packages instead of throwing an error 

WarpspeedSCP:
(cherry picked from commit 1e62e4b99b286d6152ec786efcca4edbc2437501)

Spoon currently will raise an exception if there are multiple CtPackages with the same name as a CtPackageReference. This can occur when code in two different directories contains the same package name. 

Consider large projects such as maven which have a lot of such code; maven has several libraries which all contribute classes to the same package in many cases.

This change attempts to fix the ambiguous package problem by merging all such packages and deleting any leftovers. The first non-empty package is selected as the merge target. 

There are no tests; I'd like to be pointed in the right direction regarding this.

fixes: #5475

### DIFF
--- a/src/main/java/spoon/reflect/factory/PackageFactory.java
+++ b/src/main/java/spoon/reflect/factory/PackageFactory.java
@@ -8,15 +8,17 @@
 package spoon.reflect.factory;
 
 
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.List;
-import java.util.StringTokenizer;
-import spoon.SpoonException;
 import spoon.reflect.declaration.CtModule;
 import spoon.reflect.declaration.CtPackage;
 import spoon.reflect.declaration.CtPackageDeclaration;
 import spoon.reflect.reference.CtPackageReference;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.StringTokenizer;
 
 
 /**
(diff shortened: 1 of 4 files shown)


I-Al-Istannen:
For testing I would say something along the lines of https://github.com/INRIA/spoon/pull/4139/files#diff-703cba552995d848d774527198c8562dea54acdc8c3657aa49e7a4b4cd39defa.

The problem is, that this is not spec-conform. Junit uses patch-module to tell the JVM to please shut up and accept the beating. Maybe we are at the point where majority rule decided that merging them is a suitable approach though.

I think I would like this to be an environment flag. What do you think @SirYwell @MartinWitt?

WarpspeedSCP:
this crops up when one performs analysis by including sources from multiple paths. for example, consider a multimodule project. it has sources for package com.ex.pkg1 at `module1/module2` as well as `module1/module3`.  

But when the packages are all compiled and converted to jars, the jvm will happily consider them as if they are in the same jigsaw module if no module info is specified, or if it is all compiled into a fat jar. 

Conceptually, at that point, all the classes are in the same package despite the source files having completely different file hierarchies.

So for example, if you intend to primarily process files of `module1/module2`, but you also want files of `module1/module3` in the classpath so types are resolved, one way would be to add the relevant files from module2 and module3 (with launcher.addInputResource). Then, they will run into this bug when they try to resolve a type that is in a different source root due to the module split. conceptually, they are in the same jigsaw module as far as the developer may be concerned. 

We can see this crop up in many big projects, where different classes for the same package are declared in completely different source roots. So I believe this would be useful for such contexts. The test usecase is definitely also a problem, but I personally think that is a scenario that is worth special casing for the sake of functionality. I was very surprised when I first ran into this spoon exception.

WarpspeedSCP:
In my opinion, spoon does not always need to conform to the java spec; if you go by that standard, projects like maven should not work because the spec may not explicitly mention that it is okay to split classes of a single package across multiple jars (I don't think the spec mentions this, but I may have missed that info). 

I think it is okay to deviate towards behaviour which is generally observed in real world environments if need be.

WarpspeedSCP:
I'm closing this since I haven't worked on it since my last comment. I'll try to revisit it at some point.

PR #5530, opened 2023-11-03
chore: associate a commit with beta release

algomaster99:
I am proposing changes in the beta release script to associate commit with a release so that the reproducibility of **all releases** on maven central can be checked by [reproducible central](https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/fr/inria/gforge/spoon/spoon-core/README.md). They are not able to do so right now because they cannot find the source code corresponding to the release.

I simply just copied the content from `release.sh` to `release-beta.sh` to push the commits. However, I am not sure how `jreleaser` will know which commit to tag as it is on a different branch, but I assume it would work as it works for normal releases.


MartinWitt:
Hey,

From my perspective, the beta releases are only a relict of the past, and I am against promoting them as real releases. Beta releases are a random snapshot of the master, and we have no ability to express what this release contains. Our SemVer releases are way better for end user and should be the preferred usage. Since our CD upgrade we do them more often (we should still push more fix releases, but that's another topic). I would prefer to stop releasing them before we add many commits to the master for beta releases.

algomaster99:
I understand your point. I think the beta releases exist because our main release cycles were slow to incorporate a features/fixes on the `master` branch because Martin M did them manually. Since automating them, I agree that we could completely remove beta releases and just perform normal (SemVer) releases.

algomaster99:
So should I close this PR and submit a PR for removing/disabling beta releases? The goal is that all packages that are pushed to maven central should be reproducible.

MartinWitt:
I discussed this topic with @monperrus. We decided to fully commit to beta releases. Every beta release should get a tag *and* a commit on master. Sorry for the confusion.

algomaster99:
@I-Al-Istannen if you are also in favour of " Every beta release should get a tag and a commit on master", should I submit a PR to merge the two release processes? I would change https://github.com/INRIA/spoon/blob/master/.github/workflows/release-beta.yml#L11 to use `release.sh`.

I-Al-Istannen:
I am not quite how you would merge them, as they use a different naming scheme (but both deploy to maven central, and not sonatype snapshots, right?).

Committing the `beta` version to master would entail making a commit changing the version to include `-beta`, which likely breaks the version bumping in the other release scripts, right?

algomaster99:
> deploy to maven central, and not sonatype snapshots, right?

Yes. This is why I thought I could merge them. However naming scheme, commit message, and tags would differ like you said. Thanks for pointing it out.

Then I believe reviewing this PR itself would be better.

algomaster99:
Hi! Can we merge this? We want to do it for the sake of reproducibility checks on our beta releases.

chore/release-beta.sh
```diff
@@ -19,17 +19,45 @@ echo "NEW_BETA_NUMBER $NEW_BETA_NUMBER"
...
+echo "::endgroup::"
+
+# Set next version (patch of release version) with -SNAPSHOT suffix
+NEXT_RELEASE_VERSION=$CURRENT_VERSION_WITH_SNAPSHOT
+
+echo "::group::Updating poms to next target version"
+mvn -f spoon-pom --no-transfer-progress --batch-mode versions:set -DnewVersion="$NEXT_RELEASE_VERSION" -DprocessAllModules
+mvn --no-transfer-progress --batch-mode versions:set -DnewVersion="$NEXT_RELEASE_VERSION" -DprocessAllModules
+mvn -f spoon-javadoc --no-transfer-progress --batch-mode versions:set -DnewVersion="$NEXT_RELEASE_VERSION" -DprocessAllModules
+echo "::endgroup::"
```
I-Al-Istannen, line 52: As far as I can see you do not actually update anything as `NEXT_RELEASE_VERSION` is already identical to whatever was there originally?

chore/release-beta.sh
```diff
@@ -19,17 +19,45 @@ echo "NEW_BETA_NUMBER $NEW_BETA_NUMBER"
...
+
+echo "::group::Committing changes"
+git commit -am "release: Setting SNAPSHOT version back to $NEXT_RELEASE_VERSION"
+git push --set-upstream origin "$BRANCH_NAME"
+echo "::endgroup::"
+
+echo "::group::Merging into master (fast-forward)"
+git checkout master
+git merge --ff-only "$BRANCH_NAME"
+git push origin master
```
I-Al-Istannen, line 62: Which means I don't quite get what change you actually committed and are pushing back here.

I-Al-Istannen (changes requested):

chore/release-beta.sh
algomaster99: `NEXT_VERSION` is the beta version for which the commit needs to be pushed. So I push the commit, after rewritting POMs, in R30-R42.

Next is `NEXT_RELEASE_VERSION` which is the SNAPSHOT. It will be the same version before the beta version as we don't have SNAPSHOT versions for beta as well. The name of this variable is confusing, a better name would be `SNAPSHOT_VERSION_TO_RESET_TO`.

P.S. Sorry I did not notice your reviews earlier.

I-Al-Istannen:
@algomaster99 Could you change the names? :)

algomaster99:
@I-Al-Istannen I am sorry. Completely lost context, but I changed what you asked.

I-Al-Istannen (approved):

algomaster99:
@I-Al-Istannen done!

I-Al-Istannen:
Actually, sorry: Could you maybe just revert the commit (with a custom message ideally) and then push that? Instead of invoking all kinds of maven stuff that might fail somehow?

I-Al-Istannen:
If that's annoying, I am also fine enough with this solution

I-Al-Istannen:
The commit made by the workflow, bumping the version for the release

algomaster99:
Good idea. Can do that.

algomaster99:
@I-Al-Istannen done! I removed maven stuff and I also added a command at the end to delete the remote branch from repository.

I-Al-Istannen (approved):

I-Al-Istannen:
Then let's see what breaks... :^)

algomaster99:
Hoping my Sunday is not ruined :crossed_fingers:

algomaster99:
Nothing broke it seems :D

algomaster99:
@I-Al-Istannen should we mark the release as pre-release instead of latest?

PR #5550, opened 2023-11-20
 fix: fix updating of spoon version in smoke test

MartinWitt:
Given the results of shell check, we should make a second PR trying to fix them afterward.


I-Al-Istannen (approved):
Oh lord.

I-Al-Istannen:
Let's see if this bold move will pay off. Thank you @MartinWitt.

PR #5556, opened 2023-11-26
fix: compare package names for equality between CtType

### LINKED ISSUE #5536: `equals()` semantics on `CtType`?
I'm trying to understand the semantics of `equals()` on Spoon's `CtType`s.

---
Consider two packages, `a` and `b`, which both contain an _empty_ class `C`:
```
package a
  class C
package b
  class C
```

In this case, the `equals()` method on `a.C` and `b.C` returns `true`, even though their qualified names differ and they are arguably two different types.

---
In the following case, the `equals()` method on `a.C` and `b.D` unsurprisingly returns `false`:
```
package a
  class C
package b
  class D
```

---
Now, if we insert two identical methods in `a.C` and `b.C`, `equals()` returns `true` again:
```
package a
  class C
    void m()
package b
  class C
    void m()
```

---
But if their names differ, `equals()` returns `false` again:
```
package a
  class C
    void m1()
package b
  class C
    void m2()
```

---
It seems that `equals()` is checking for structural equivalence rather than nominal equivalence (but not strictly because in the second case `C` and `D` are not considered equal). Is this the intended semantics? The main issue is that it makes some analyses awkward to write. If I analyze a project with Spoon and insert all its types in a `Set`, many of them disappear, though they are indeed distinct types according to Java's semantics. Generally, manipulating Spoon types in collections is often awkward.

---
For reference, the minimal test. This is with Spoon 10.4.2.

```
@Test
void equals_semantics() {
  launcher.addInputResource("spoon-equals");

  CtType<?> ac = model.getRootPackage().getPackage("a").getType("C");
  CtType<?> bc = model.getRootPackage().getPackage("b").getType("C");

  assertEquals(ac, bc);
}
```

algomaster99:
Fixes #5536 

The types are unequal because they are in different packages, which affects their fully qualified names. Since `scanCtNamedElement` compares based upon `simpleName`, the comparison would pass because it would not account for package names in names of types. 

I am proposing changes that would compare types based on qualified names and would take any implicit packages into account. Another thing that I noticed was `getValueByRole(CtRole.NAME)` for a class returns its simple name. So, it is semantically wrong to set the role unequal based on comparing qualified names.

### DIFF
--- a/src/main/java/spoon/experimental/SpoonifierVisitor.java
+++ b/src/main/java/spoon/experimental/SpoonifierVisitor.java
@@ -146,7 +146,7 @@ public class SpoonifierVisitor extends CtScanner {
 
 		if (element instanceof CtType<?> && !((CtType<?>) element).isAnonymous()) {
 			// anonymous classes don't have a package
-			result.append(printTabs() + variableName + ".setParent(factory.Package().getOrCreate(\"" + ((CtType<?>) element).getPackage().getQualifiedName() + "\"));\n");
+			result.append(printTabs()).append(variableName).append(".setParent(factory.Package().getOrCreate(\"").append(((CtType<?>) element).getPackage().getQualifiedName()).append("\"));\n");
 		}
 
 		if (element.isParentInitialized() && !parentName.isEmpty()) {
--- a/src/main/java/spoon/experimental/SpoonifierVisitor.java
+++ b/src/main/java/spoon/experimental/SpoonifierVisitor.java
@@ -20,6 +20,7 @@ import spoon.reflect.declaration.CtMethod;
 import spoon.reflect.declaration.CtModifiable;
 import spoon.reflect.declaration.CtNamedElement;
 import spoon.reflect.declaration.CtParameter;
+import spoon.reflect.declaration.CtType;
 import spoon.reflect.declaration.ModifierKind;
 import spoon.reflect.path.CtRole;
 import spoon.reflect.reference.CtArrayTypeReference;
@@ -141,6 +142,11 @@ public class SpoonifierVisitor extends CtScanner {
 			}
 			propertyScanner.variableName = variableName;
 			element.accept(propertyScanner);
+		}
+
+		if (element instanceof CtType<?> && !((CtType<?>) element).isAnonymous()) {
+			// anonymous classes don't have a package
+			result.append(printTabs() + variableName + ".setParent(factory.Package().getOrCreate(\"" + ((CtType<?>) element).getPackage().getQualifiedName() + "\"));\n");
 		}
 
 		if (element.isParentInitialized() && !parentName.isEmpty()) {
--- a/src/test/java/spoon/test/eval/EvalTest.java
+++ b/src/test/java/spoon/test/eval/EvalTest.java
@@ -465,13 +465,12 @@ public class EvalTest {
 			.getElements(new TypeFilter<>(CtBinaryOperator.class))
 			.get(0);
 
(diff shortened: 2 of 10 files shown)


src/test/java/spoon/test/eval/EvalTest.java
```diff
@@ -465,13 +465,12 @@ void testVisitCtBinaryOperatorLiteralType(
...
 			.get(0);
 
-		CtType<?> currentType = ctBinaryOperator.getType().getTypeDeclaration().clone();
 		CtExpression<?> evaluated = ctBinaryOperator.partiallyEvaluate();
 		assertNotNull(
 			evaluated.getType(),
 			String.format("type of '%s' is null after evaluation", ctBinaryOperator)
 		);
-		assertEquals(currentType, evaluated.getType().getTypeDeclaration());
+		assertEquals(ctBinaryOperator.getType(), evaluated.getType());
```
algomaster99, line 473: The test should only check for the type of the binary operator and the evaluated result. There is no need to check the equality between type declarations.

src/test/java/spoon/test/spoonifier/SpoonifierTest.java
```diff
@@ -91,6 +91,7 @@ public void testSpoonifier() throws ClassNotFoundException, MalformedURLExceptio
 		testSpoonifierWith("src/test/java/spoon/test/spoonifier/testclasses/ArrayRealVector.java", i++);
 		testSpoonifierWith("src/test/java/spoon/test/prettyprinter/testclasses/FooCasper.java", i++);
 		testSpoonifierWith("src/test/java/spoon/test/prettyprinter/testclasses/Rule.java", i++);
+		testSpoonifierWith("src/test/resources/UnnamedPackage.java", i++);
```
algomaster99, line 94: I added this test just to ensure that classes with `unnamed packages` can also be spoonified.

src/main/java/spoon/experimental/SpoonifierVisitor.java
```diff
@@ -143,6 +144,11 @@ && isLeafTypeReference(element.getParent())) {
 			element.accept(propertyScanner);
 		}
 
+		if (element instanceof CtType<?> && !((CtType<?>) element).isAnonymous()) {
+			// anonymous classes don't have a package
+			result.append(printTabs() + variableName + ".setParent(factory.Package().getOrCreate(\"" + ((CtType<?>) element).getPackage().getQualifiedName() + "\"));\n");
```
algomaster99, line 149: Qodana complains below. Would it go away if I fix this line or I will have to fix the entire file?

algomaster99:
I am not sure why this one test is failing. :(
Could use some help, @I-Al-Istannen

algomaster99:
https://github.com/INRIA/spoon/pull/5556/commits/07550abdccc21f5d6f4394646a082409d3cc3500 is a bad solution to the problem above. But it passes the test if I explicitly set the package reference after clone. I will look for a better solution.

I-Al-Istannen:
> What is the contract around clone? What all should be cloned and what should be skipped? For example, I know that metadata is not cloned.

I think so far the parent relationships are not cloned, as that can lead to all kinds of trouble. If you set the parent, it might also attach itself in other ways and vice versa. Changing the parent or setting children potentially modifies quite a bit of the model.

I am not quite sure you always want to compare the package name as well. If you do that, there is no real way to do structural comparisons anymore, it is a lot easier to check the package name as well.

algomaster99:
The solution is discussed [here](https://github.com/INRIA/spoon/issues/5536#issuecomment-1827779594).

PR #5637, opened 2024-01-24
review: feature: Record Patterns (JEP 440)

SirYwell:
This PR adds support for record patterns as described by [JEP 440](https://openjdk.org/jeps/440).

Extra checks Javadoc issues and Qodana can be ignored, again.


src/main/java/spoon/reflect/code/CtRecordPattern.java
```diff
@@ -0,0 +1,84 @@
...
+ *     Object obj = null;
+ *     boolean longerThanTwo = false;
+ *     record MyRecord(String value) {}
+ *     // MyRecord(var string) is the record pattern
+ *     if (obj instanceof MyRecord(var string)) {
+ *         longerThanTwo = string.length() > 2;
+ *     }
+ * </pre>
+ */
+public interface CtRecordPattern extends CtPattern, CtExpression<Void> {
```
I-Al-Istannen, line 35: Why does this extend `CtExpression<Void>`?

SirYwell: This decision is basically copied from `CtTypePattern`. Iirc, we decided to go this route because we needed a `CtExpression` for the `CtBinaryOperator`, but the pattern doesn't really have a type.

I-Al-Istannen (approved):

PR #5642, opened 2024-01-26
review: chore: Rewrite javadoc checkstyle in spoon 

I-Al-Istannen:
Checkstyle still does not support `{@return}` inline tags :/

### DIFF
--- a/chore/CheckJavadoc.java
+++ b/chore/CheckJavadoc.java
@@ -45,6 +45,18 @@ public class CheckJavadoc {
 
   private static final Path CWD = Path.of("").toAbsolutePath();
 
+  private static final Set<String> EXCLUDED_TYPES = Set.of(
+      "spoon.reflect.visitor.CtAbstractVisitor",
+      "spoon.reflect.visitor.CtBiScannerDefault",
+      "spoon.reflect.visitor.CtInheritanceScanner",
+      "spoon.reflect.visitor.CtScanner",
+      "spoon.reflect.visitor.CtVisitor",
+      "spoon.support.compiler.jdt.JDTTreeBuilder",
+      "spoon.support.compiler.jdt.ParentExiter",
+      "spoon.support.visitor.clone.CloneVisitor",
+      "spoon.support.visitor.replace.ReplacementVisitor"
+  );
+
   private final Map<ViolationType, Integer> violationCounts = new HashMap<>();
 
   private void check(CtModel model, Factory factory) {
@@ -122,6 +134,9 @@ public class CheckJavadoc {
         if (!member.isPublic()) {
           return false;
         }
+      }
+      if (current instanceof CtType<?> type && EXCLUDED_TYPES.contains(type.getQualifiedName())) {
+        return false;
       }
       current = current.getParent();
     }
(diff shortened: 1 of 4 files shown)


chore/check-javadoc-regressions.py
```diff
@@ -191,17 +172,18 @@ def command_compare_with_branch(target_branch: str) -> None:
     Compares the current branch with a given reference branch and prints a change summary.
     Exits with an error if the current branch has more violations than the passed target branch.
     """
-    # Switch to the root so checkstyle will run against the whole project
+    # Switch to the root soquality check will run against the whole project
```
SirYwell, line 175:
```suggestion
    # Switch to the root so quality check will run against the whole project
```

chore/CheckJavadoc.java
```diff
@@ -0,0 +1,285 @@
...
+
+    CheckJavadoc checker = new CheckJavadoc();
+    checker.check(model, launcher.getFactory());
+
+    System.out.println("Violations per category");
+    for (var entry : checker.violationCounts.entrySet().stream().sorted(Entry.comparingByValue()).toList()) {
+      System.out.println(entry);
+    }
+
+    int errorCount = checker.violationCounts.values().stream().mapToInt(Integer::intValue).sum();
```
SirYwell, line 226: I hope we don't encounter an overflow here \:)

SirYwell (commented):
Probably rather a follow-up: Would it be possible to extract this from the extra checks into its own job?

chore/CheckJavadoc.java
MartinWitt: I hope we just delete the check if we have an overflow.

MartinWitt:
The extra checks are failing because we have a better analysis now?

chore/CheckJavadoc.java
```diff
@@ -0,0 +1,285 @@
+///usr/bin/env jbang "$0" "$@" ; exit $?
+//DEPS fr.inria.gforge.spoon:spoon-core:10.4.2
+//DEPS fr.inria.gforge.spoon:spoon-javadoc:10.4.2
```
MartinWitt, line 3: Could we use the newest snapshot here? Otherwise, fixing bugs in spoon javadoc is not directly reflected in the check.

chore/CheckJavadoc.java
I-Al-Istannen: I would delete spoon tbh. At that point we had a good run, but you sometimes have to admit defeat.

chore/CheckJavadoc.java
I-Al-Istannen: How exactly?

I-Al-Istannen:
> Probably rather a follow-up: Would it be possible to extract this from the extra checks into its own job?

That would be trivial, yes @SirYwell. I can do that here if you want.


> The extra checks are failing because we have a better analysis now?
@MartinWitt The java17 branch added support for patterns but no javadoc, yea. We could ignore the jdt tree builder, if you'd like.

SirYwell:
> That would be trivial, yes @SirYwell. I can do that here if you want.

That would be nice, the extra checks typically run for a long time and seeing javadoc issues only at the end is somewhat annoying.

I-Al-Istannen:
Done @SirYwell.

SirYwell (approved):
Nice, finally I don't need to wait 10 minutes for javadoc issues to show up :)

PR #5644, opened 2024-01-29
review: test: introduce generated AssertJ asserts

SirYwell:
This pull request introduces enhanced support for AssertJ for our test infrastructure. Specialized Assert classes allow directly testing for specific metamodel properties, reducing the repetitive code we currently write in tests.

The classes and interfaces are automatically generated, methods added to interfaces manually will persist when regenerating. To ensure consistency, we have a CI check that fails if regeneration is needed.


MartinWitt:
The java 21 failures are related to:
> Caused by: java.lang.IllegalArgumentException: Java 21 (65) is not supported by the current version of Byte Buddy which officially supports Java 20 (64) - update Byte Buddy or set net.bytebuddy.experimental as a VM property

MartinWitt:
Okay @SirYwell's AssertJ version was built a long time ago in a galaxy far away. Updating AssertJ will fix this.

PR #5645, opened 2024-01-29
review: fix: Do not inherit the return tag if an inline return is present

I-Al-Istannen:


SirYwell (approved):

SirYwell (changes requested):
Actually, could you add contract comments to the tests?

SirYwell (approved):

SirYwell (approved):

PR #5646, opened 2024-01-30
review: chore: deprecate public CtTypeReference fields in TypeFactory

SirYwell:
Direct access allows mutating the base object, which should be prevented.


I-Al-Istannen (changes requested):
Could you replace the factory methods with the actual value instead of cloning the deprecated constants?

I-Al-Istannen (approved):

I-Al-Istannen:
If you merge https://github.com/INRIA/spoon/pull/5662 and rebase ontop of the new java17 branch, the CI should be green :)

Issue #4129, opened 2021-08-25
bug: DefaultJavaPrettyPrinter misses line breaks with enums


SirYwell:
I found some issues when writing enums:

1. If an enum only contains enum values but no other type members, code like
```java
enum SimpleEnum {
    CONSTANT
}
```
is transformed into
```java
enum SimpleEnum {

    CONSTANT;}
```
This comes from the printList call here: https://github.com/INRIA/spoon/blob/e8c0d1b75e694e6dea8c70ef82bc3b6e8cb7cb0b/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java#L753
It writes a line break before each enum value but does not add one after the `;`. This isn't a problem if there are any other type members to write because `ElementPrinterHelper#writeElementList` will add the line break (although there is no empty line then, but that breaks a test here due to the expected output: https://github.com/INRIA/spoon/blob/cd339e2c5f0e5c1e42c66b890f02bc282c3a0ea1/src/test/java/spoon/test/prettyprinter/DefaultPrettyPrinterTest.java#L303-L311)

Only writing a new line if enum values were written but no type members makes it work.

2. Using `preserveLineNumbers` writes an initial code
```java
enum AlmostEmptyEnum {
    ;
    @Override
    public void toString() {
        return name().toLowerCase();
    }
}
```
as
```java
enum AlmostEmptyEnum {
    ;
    @Override
    public void toString() {
        return name().toLowerCase();
    }}
```
This only happens with that option, without the option it looks like
```java
enum AlmostEmptyEnum {
    ;

    @Override
    public void toString() {
        return name().toLowerCase();
    }
}
```
Note that without preserving line numbers, a new line is added before the method. This line is not added with that option but  due to https://github.com/INRIA/spoon/blob/2fff4c50266f79bc06b961e3b01ca6c6e0fcc1ab/src/main/java/spoon/reflect/visitor/ElementPrinterHelper.java#L185-L187 the last line break is still skipped (this might be an issue with normal classes too(?)).

Another - rather unrelated - issue I encountered with printing of enum values is that an enum
```java
enum SimpleEnum {
    CONSTANT {

    }
}
```
is written as
```java
enum SimpleEnum {

    CONSTANT() {};}
```
The added `()` and the `;` also appear with the Sniper printer.

----

### Edit

[...]

PR #5649, opened 2024-01-31
fix: Add extra newline after printing enum values

sanderploegsma:
This partly solves #4129 by adding an extra newline after printing the enum values.

So instead of printing this:

```java
public enum SimpleEnum {

    CONSTANT;}
```

It now prints this:

```java
public enum SimpleEnum {

    CONSTANT;
}
```

This does have a small impact on enums with members, as previously they would be printed like this:

```java
public enum EnumWithMembers {

    ONE,
    TWO,
    THREE;
    private static int len = -1;

    public static void f() {
        len = 44;
    }
}
```

And now they will be printed like this:

```java
public enum EnumWithMembers {

    ONE,
    TWO,
    THREE;

    private static int len = -1;

    public static void f() {
        len = 44;
    }
}
```

To be honest I prefer the second anyway.


SirYwell (approved):
Looks good! Thanks.

PR #5673, opened 2024-02-17
review: test: fix AssertJ code generation

SirYwell:
The check to ensure the assertions are all up-to-date didn't run.
This PR fixes that, and also fixes the missing types. We also need concrete classes for interfaces that don't have one specific implementation, otherwise newly introduced code wouldn't compile anymore.

Additionally, interfaces are public now, otherwise methods returning those interfaces are basically useless.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -102,9 +102,9 @@
           '';
           codegen = pkgs.writeScriptBin "codegen" ''
            set -eu
-           mvn test -Dgroups=codegen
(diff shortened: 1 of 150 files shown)


MartinWitt:
@SirYwell there are a lot of changed files. Can you summarize the change?

SirYwell:
> @SirYwell there are a lot of changed files. Can you summarize the change?

I edited the description, I hope that helps.

PR #5674, opened 2024-02-17
feat: Add support for CtReceiverParameter in spoon model. 

### LINKED ISSUE #4069: Handling of receiver parameters
This issue covers the general handling of receiver parameters in spoon.

Receiver parameters were introduced with Java 8. They are specified in [JLS 8.4](https://docs.oracle.com/javase/specs/jls/se16/html/jls-8.html#jls-8.4) with example code in [JLS 8.4.1](https://docs.oracle.com/javase/specs/jls/se16/html/jls-8.html#jls-8.4.1).

Spoon got support for this feature with this [Pull request](https://github.com/INRIA/spoon/pull/1298) ([related issue](https://github.com/INRIA/spoon/issues/1283)).
Currently, spoon handles receiver parameters in the same way other parameters are handled too. This brings us to the first issue:
When creating a `CtExecutableReference` (using `CtExecutable#getReference()` or `ExecutableFactory#createReference(CtExecutable<?>)`), the receiver parameter is included. For me, this is unexpected or even wrong, and a call to `CtExecutableReference#getActualMethod()` will always return `null`.

Having the receiver parameter in the parameter list also makes the creation of invocations harder, as you manually need to ignore receiver parameters on your own.

A second issue becomes visible when looking at the example code in the JLS: For now, spoon only handles *methods*. A receiver parameter in a constructor of an inner class (referencing the enclosing class) will simply be ignored and does not appear in the model created by spoon.

Adding support for constructors in the same way as with methods should be pretty simple, but I'd like to discuss alternatives of how to represent receiver parameters in the metamodel altogether.

I thought about several solutions with different pros and cons:

## 1) Add a `isReceiver(): boolean` method to `CtParameter`

Pros:

- Backwards compatible
- Allows to check if a parameter is a receiver parameter without the need to check `CtParameter#getSimpleName().equals("this")`
- Can be implemented by either setting a boolean or just checking `CtParameter#getSimpleName().equals("this")` internally
- Works for both constructors and methods directly

Cons:

- Only one parameter (the first) can be a receiver parameter - we have very specific code in a less specific class
- Static methods and non-inner constructors do not allow receiver parameters - this is not modeled with this solution
- `CtExecutableReference#getParameters()` still returns a list of `CtTypeReference`s including the receiver parameters
  - Duplicating the `isReceiver()` method in `CtTypeReference` would increase code that does not really belong there
  - In best case, you should not need to check the `CtParameter`s of the `CtExecutable` to see if a `CtTypeReference` is a receiver
- The receiver parameter is not part of the method signature but in the list of method parameters - this might be unexpected for any work with the API

### a) but keep the parameter in the parameter list

Pros:

- Backwards compatible
- Easy to implement, basically the same way as mentioned in 1) by checking the first parameter (if present) or a boolean
- Works for both constructors and methods directly
- The receiver parameter is unique in its name (besides qualified names like `Outer.Inner.this`) and position (0) and is represented in the model this way

Cons:

- Should be reflected in `CtExecutableReference` too
- The receiver parameter is still in the list of parameters

### b) and remove it from the parameter list

Pros:

- Easy to implement
- Works for both constructors and methods directly
- Does not infer with the parameter list for invocations and references

Cons:

- Not backwards compatible if the current way is considered to be correct, although I'm not sure what to expect when reading the current documentation of `CtExecutable#getParameters()` or `CtParameter`

## 3) Add a `CtReceiver`

Do not add it to the parameter list but add `get/setReceiver()` methods. Add the respective properties to `CtExecutable`. I'm not sure if it should be part of `CtExecutableReference` with a `CtReceiverReference` or similar.

(super types of `CtReceiver` might be discussed)

Pros:

- Makes it easy to deal with other parameters
- Makes dealing with receiver parameters easy as well
- Allows to keep track of additional information (as e.g. qualified names)
- Does not infer with the parameter list for invocations and references
- Works for both constructors and methods directly

Cons:

- Not backwards compatible in the way described in 2b)

#### 4) Only add support for constructors the same way as for methods

Pros:

- Easy to implement

Cons:

- None of the points I considered above are applied
- Handling receiver parameters is more difficult
- Handling constructors without thinking about receivers will be error prone (could even break existing code)

----

Some general questions:

- How to deal with a `CtExecutable#addParameterAt(int, CtParameter)` with `0` as `position`? (e.g. refactoring a method and its invocations by adding a parameter at 0 would fail currently)
- Would anyone expect receiver parameters in the parameter list? Would it makes sense to clarify the docs in such places?

**Side note**: When looking at the code, I was super confused because the JDT `Receiver` is never traversed by JDT itself - in contrast to the other arguments, annotations, type parameters, etc. I don't know if this is a bug in JDT, but calling the traverse method ourselves is fine I guess.

----

I'm happy to implement any of this, or modified variants depending on your opinions and ideas. Feedback is very welcome.

MartinWitt:
Fixes #4069 by implementing a variant of approach 3. *Theoretically*, this is a breaking change, but practically this fixes bugs in Spoon (e.g. `CtExecutableReference#getExecutableDeclaration` returns null if the declaration has an explicit receiver parameter; or type adaption). Other uses of Spoon will similarly miss the special handling for receiver parameters. By making it its own element in the spoon model, it's easier to observe and deal with it correctly.


src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java
```diff
@@ -2359,4 +2360,22 @@ public void visitCtRecordPattern(CtRecordPattern recordPattern) {
...
 	}
+
+	@Override
+	public void visitCtReceiverParameter(CtReceiverParameter receiverParameter) {
+		elementPrinterHelper.writeComment(receiverParameter);
+		elementPrinterHelper.writeAnnotations(receiverParameter);
+
+		printer.writeIdentifier(receiverParameter.getType().getSimpleName());
+		printer.writeSpace();
+		boolean isInnerClass = receiverParameter.getType().getTopLevelType().equals(receiverParameter.getParent(CtType.class));
```
SirYwell, line 2371: According to Qodana, this is wrong, and I think it's also wrong when having nested classes in nested classes

src/main/java/spoon/support/compiler/jdt/ParentExiter.java
```diff
@@ -207,6 +208,8 @@ public <R> void scanCtExecutable(CtExecutable<R> e) {
 		} else if (child instanceof CtBlock && !(e instanceof CtMethod || e instanceof CtConstructor)) {
 			e.setBody((CtBlock<R>) child);
 			return;
+		} else if (child instanceof CtReceiverParameter) {
+				e.setReceiverParameter((CtReceiverParameter) child);
```
SirYwell, line 212: We have pattern matching now!

src/main/java/spoon/support/reflect/declaration/CtReceiverParameterImpl.java
```diff
@@ -0,0 +1,60 @@
...
+import spoon.reflect.visitor.CtVisitor;
+
+public class CtReceiverParameterImpl extends CtElementImpl implements CtReceiverParameter   {
+
+	@MetamodelPropertyField(role = CtRole.TYPE)
+	private CtTypeReference<Object> type;
+	@MetamodelPropertyField(role = CtRole.IS_SHADOW)
+	boolean isShadow;
+
+	public boolean isShadow() {
```
SirYwell, line 25: Missing `@Override`?

src/main/java/spoon/support/visitor/clone/CloneBuilder.java
```diff
@@ -5,6 +5,12 @@
  *
  * Spoon is available either under the terms of the MIT License (see LICENSE-MIT.txt) or the Cecill-C License (see LICENSE-CECILL-C.txt). You as the user are entitled to choose the terms under which to adopt Spoon.
  */
+/* SPDX-License-Identifier: (MIT OR CECILL-C)
+
+Copyright (C) 2006-2023 INRIA and contributors
+
+Spoon is available either under the terms of the MIT License (see LICENSE-MIT.txt) or the Cecill-C License (see LICENSE-CECILL-C.txt). You as the user are entitled to choose the terms under which to adopt Spoon.
+ */
```
SirYwell, line 13: This should be reverted in the generated files

src/test/java/spoon/test/receiverparameter/ReceiverParameterTest.java
```diff
@@ -0,0 +1,46 @@
...
+        CtType<?> targetType = model.getAllTypes().iterator().next();
+        List<CtReceiverParameter> receiverParams = targetType.getElements(new TypeFilter<>(CtReceiverParameter.class));
+        assertEquals(1, receiverParams.size());
+        CtReceiverParameter next1 = receiverParams.iterator().next();
+        assertEquals("receiver.Outer", next1.getType().getQualifiedName());
+        CtConstructor<?> ctConstructor = targetType.getElements(new TypeFilter<>(CtConstructor.class)).stream().filter(v -> v.getReceiverParameter() != null).findFirst().get();
+        SpoonAssertions.assertThat(ctConstructor).isNotNull();
+        SpoonAssertions.assertThat(ctConstructor.getReceiverParameter()).isNotNull();
+    }
+}
```
SirYwell, line 46: Could we have some printer tests here too?

SirYwell (changes requested):
There's a lot of noise due to imports and formatting changes, can you revert them?

src/test/java/spoon/test/receiverparameter/ReceiverParameterTest.java
MartinWitt: I added tests for printing ReceiverParameters

src/test/java/spoon/test/receiverparameter/ReceiverParameterTest.java
```diff
@@ -0,0 +1,61 @@
...
+		CtReceiverParameter receiverParam = receiverParams.iterator().next();
+		SpoonAssertions.assertThat(receiverParam.getType()).extracting(CtTypeInformation::getQualifiedName).isEqualTo("receiver.SimpleReceiverParameter");
+		SpoonAssertions.assertThat(receiverParam).extracting(CtElement::toString).isEqualTo("SimpleReceiverParameter this");
+
+	}
+
+
+	@ModelTest(
+			value = "src/test/resources/receiver/InnerClassCtor.java"
+	)
```
I-Al-Istannen, line 34:
```suggestion
	@ModelTest("src/test/resources/receiver/InnerClassCtor.java")
```

I-Al-Istannen (changes requested):

MartinWitt:
@SirYwell

SirYwell (approved):
Looks good!

I-Al-Istannen (approved):

PR #5718, opened 2024-03-27
review: refactor: make ASTPair a record

SirYwell:
`ASTPair` is a perfect example of a class that should be a record. In theory it's breaking, but it's supposed to be used internally only anyways, and it's targeting the java17 branch.


MartinWitt:
Can we mark this class as internal? Normally, the class should never change, but lets keep at least the possibility?

I-Al-Istannen (approved):

PR #5726, opened 2024-04-03
Update spoon pom to current version in spoon-control-flow

Mr-Pine:
The `spoon-pom` parent in `spoon-control-flow` was still on version 1.0. This caused `mvn test` to fail with a compiler error because `spoon-pom` only includes junit4. I updated the version to the latest spoon version.


Mr-Pine:
This is now included in #5733

PR #5739, opened 2024-04-05
review: fix(control-flow): Proper generation of CFG for constructor

Mr-Pine:
Currently, generating a control flow graph for a `CtConstructor` generates an "empty" CFG, consisting just of the `BEGIN` and `EXIT` node and an edge between them. With this PR they are processed the same as a `CtMethod`


MartinWitt:
@Mr-Pine `getFirst` is not part of java 17

SirYwell (approved):

PR #5740, opened 2024-04-05
review: fix(control-flow): Handle synchronized statement in CFG

Mr-Pine:
Currently, a `synchronized` statement is completely ignored in the CFG. This PR adds it to the CFG properly.

### DIFF
--- a/spoon-control-flow/src/main/java/fr/inria/controlflow/ControlFlowBuilder.java
+++ b/spoon-control-flow/src/main/java/fr/inria/controlflow/ControlFlowBuilder.java
@@ -798,7 +798,11 @@ public class ControlFlowBuilder extends CtAbstractVisitor {
 
 	@Override
 	public void visitCtSynchronized(CtSynchronized synchro) {
+		ControlFlowNode expressionNode = new ControlFlowNode(synchro.getExpression(), result, BranchKind.STATEMENT);
+		tryAddEdge(lastNode, expressionNode);
+		lastNode = expressionNode;
 
+		synchro.getBlock().accept(this);
 	}
 
 	@Override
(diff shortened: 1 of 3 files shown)


MartinWitt (approved):

MartinWitt:
LGTM will merge

spoon-control-flow/src/main/java/fr/inria/controlflow/ControlFlowBuilder.java
```diff
@@ -798,7 +798,11 @@ public <T, S> void visitCtSwitchExpression(CtSwitchExpression<T, S> switchExpres
 
 	@Override
 	public void visitCtSynchronized(CtSynchronized synchro) {
+		ControlFlowNode expressionNode = new ControlFlowNode(synchro.getExpression(), result, BranchKind.STATEMENT);
```
I-Al-Istannen, line 801: If the synchronization has no object (e.g. `synchronized {}`) this will likely be null and create a node with `null`. That sounds a bit weird? Could you add a test for that?

spoon-control-flow/src/main/java/fr/inria/controlflow/ControlFlowBuilder.java
```diff
@@ -798,7 +798,11 @@ public <T, S> void visitCtSwitchExpression(CtSwitchExpression<T, S> switchExpres
...
+		tryAddEdge(lastNode, expressionNode);
+		lastNode = expressionNode;
```
I-Al-Istannen, line 803: Do we want to preserve the synchronized statement itself in the graph? I.e. call `defaultAction(synchro, BranchKind.STATEMENT)` or w/e somewhere? Currently it looks like the synchronized is unrolled and no trace of it is left?

I-Al-Istannen (changes requested):

spoon-control-flow/src/main/java/fr/inria/controlflow/ControlFlowBuilder.java
Mr-Pine: Ideally yes, but in the current CFG model this is not possible in any way I like. This is one of the things I would like to improve with the redesign started in #5742

PR #5746, opened 2024-04-07
review: fix: skip synthetic break from JDT

### LINKED ISSUE #5743: [Bug]: Spoon inserts incorrect break in enhanced switch block
### Describe the bug

As seen in the outputs below, spoon inserts a break statement at the end of blocks in an enhanced (arrow) switch that don't exist in the original source code
Note: With complianceVersion < 14 the bug doesn't appear. But those versions should not support enhanced switches at all, so...

```Java
class Main {
    void main() {
        switch(0) {
            case 1 -> {
            }
        }
    }
}
```

```Java
import spoon.Launcher;
import spoon.SpoonException;
import spoon.reflect.declaration.CtClass;
import spoon.support.compiler.VirtualFile;

import java.util.Collection;

public class Main {

        CtClass<?> clazz = parseClass("""
                class Main {
                    void main() {
                        switch(0) {
                            case 1 -> {
                            }
                        }
                    }
                }
                """);
        System.out.println(clazz);
    }

    public static CtClass<?> parseClass(String code) {
        launcher.getEnvironment().setComplianceLevel(21);
        Collection<CtType<?>> allTypes = launcher.buildModel().getAllTypes();
        try {
            return (CtClass<?>) allTypes.stream().findFirst().get();
        } catch (ClassCastException e) {
            throw new SpoonException("parseClass only considers classes (and not interfaces and enums). Please consider using a Launcher object for more advanced usage.");
        }
    }
}
```

### Actual output

```Java
class Main {
    void main() {
        switch (0) {
            case 1 ->
                {
                    break;
                }
        }
    }
}
```

### Expected output

```Java
class Main {
    void main() {
        switch (0) {
            case 1 ->
                {
                }
        }
    }
}
```

### Spoon Version

11.0.0

### JVM Version

21

Linux

SirYwell:
Fixes #5743

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilder.java
+++ b/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilder.java
@@ -342,6 +342,9 @@ public class JDTTreeBuilder extends ASTVisitor {
 
 	@Override
 	public void endVisit(BreakStatement breakStatement, BlockScope scope) {
+		if (breakStatement.isSynthetic) {
+			return; // we never entered
+		}
 		context.exit(breakStatement);
 	}
 
@@ -1077,6 +1080,9 @@ public class JDTTreeBuilder extends ASTVisitor {
 
 	@Override
 	public boolean visit(BreakStatement breakStatement, BlockScope scope) {
+		if (breakStatement.isSynthetic) {
+			return false;
+		}
 		CtBreak b = factory.Core().createBreak();
 		if (breakStatement.label != null) {
 			b.setTargetLabel(new String(breakStatement.label));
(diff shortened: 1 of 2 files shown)


src/main/java/spoon/support/compiler/jdt/JDTTreeBuilder.java
```diff
@@ -342,6 +342,9 @@ public void endVisit(Block block, BlockScope scope) {
 
 	@Override
 	public void endVisit(BreakStatement breakStatement, BlockScope scope) {
+		if (breakStatement.isSynthetic) {
+			return; // we never entered
```
I-Al-Istannen, line 346: Why handle this then instead of throwing an exception?

SirYwell: This method is called by JDT, whether we want or not (i.e. even if we return false in visit(BreakStatement, ...))

I-Al-Istannen: ah right we would need to return false in the *parent* construct. That makes sense

MartinWitt:
Thanks @SirYwell

PR #5754, opened 2024-04-08
chore: Remove jbang. It falls apart randomly during dependency resolution

I-Al-Istannen:
The downside is that we commit a lot of sins. This is a giant clusterfuck, but the nice way to run a java file with dependencies via jbang sporadically runs into dependency resolution failures...

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -119,8 +119,10 @@ jobs:
 
       - name: Time nix setup
         run: nix develop ${{ env.NIX_OPTIONS }} .#extraChecks --command true
+      - name: Time nix setup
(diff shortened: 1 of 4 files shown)


.github/workflows/tests.yml
```diff
@@ -119,8 +119,10 @@ jobs:
 
       - name: Time nix setup
         run: nix develop ${{ env.NIX_OPTIONS }} .#extraChecks --command true
+      - name: Time nix setup
```
SirYwell, line 122: This has the same name as the step above?

SirYwell (approved):

MartinWitt:
Thanks @I-Al-Istannen lets make a new patch release soonish after the CFG is done to test this.

PR #5801, opened 2024-05-09
review: fix `toMethod()` implementation of `CtRecordComponent` to return useful spoon classes

Luro02:


MartinWitt:
Can you fix the CI issues?

src/main/java/spoon/support/reflect/declaration/CtRecordComponentImpl.java
```diff
@@ -36,18 +39,45 @@ public class CtRecordComponentImpl extends CtNamedElementImpl implements CtRecor
 	public CtMethod<?> toMethod() {
 		CtMethod<?> method = this.getFactory().createMethod();
 		method.setSimpleName(getSimpleName());
-		method.setType((CtTypeReference) getType());
+		method.setType(getType());
```
Luro02, line 42:
```suggestion
		method.setType(getType().clone());
```
Otherwise the method will have an invalid parent, see #5870

src/main/java/spoon/support/reflect/declaration/CtRecordComponentImpl.java
```diff
@@ -36,18 +39,45 @@ public class CtRecordComponentImpl extends CtNamedElementImpl implements CtRecor
...
 		method.setSimpleName(getSimpleName());
-		method.setType((CtTypeReference) getType());
+		method.setType(getType() != null ? getType().clone() : null);
 		method.setExtendedModifiers(Collections.singleton(new CtExtendedModifier(ModifierKind.PUBLIC, true)));
 		method.setImplicit(true);
-		method.setBody(getFactory().createCodeSnippetStatement("return " + getSimpleName()));
+
+		CtFieldAccess<?> ctVariableAccess = (CtFieldAccess<?>) getFactory().Code()
+			.createVariableRead(getRecordFieldReference(), false);
+		// ensure the `this.` is implicit (important as we are also implicit)
```
SirYwell, line 48: why is this important? the variable access and the return are not implicit

src/main/java/spoon/support/reflect/declaration/CtRecordImpl.java
```diff
@@ -100,12 +104,20 @@ public void accept(CtVisitor v) {
 	public <C extends CtType<Object>> C addTypeMemberAt(int position, CtTypeMember member) {
 		// a record can have only implicit instance fields and this is the best point to preserve the invariant
 		// because there are multiple ways to add a field to a record
+		String memberName = member.getSimpleName();
+
 		if (member instanceof CtField && !member.isStatic()) {
+			// We add the field in addRecordComponent. Afterward, however, JDT visits the Field itself -> Duplication.
```
SirYwell, line 110: This feels weird here, as we drag JDT specific code into unrelated places. Can we avoid the duplication somewhere else?

src/main/java/spoon/support/reflect/declaration/CtRecordComponentImpl.java
```diff
@@ -36,18 +39,45 @@ public class CtRecordComponentImpl extends CtNamedElementImpl implements CtRecor
 	public CtMethod<?> toMethod() {
 		CtMethod<?> method = this.getFactory().createMethod();
 		method.setSimpleName(getSimpleName());
-		method.setType((CtTypeReference) getType());
+		method.setType(getType() != null ? getType().clone() : null);
```
SirYwell, line 42: might make sense to have a `cloneType` method considering how often this code is used here

SirYwell (changes requested):

src/main/java/spoon/support/reflect/declaration/CtRecordComponentImpl.java
Luro02: I can't remember what the reason for this comment was, the PR has been open for months. I guess this could be removed

src/main/java/spoon/support/reflect/declaration/CtRecordComponentImpl.java
I-Al-Istannen: That is something that is generally useful, but the getters require returning the original reference

src/main/java/spoon/support/reflect/declaration/CtRecordImpl.java
I-Al-Istannen: If you have an idea, feel free. I didn't find anything nice.

src/main/java/spoon/support/reflect/declaration/CtRecordComponentImpl.java
I-Al-Istannen: I would imagine you wanted `return foo` instead of `return this.foo`, as the whole synthetic method should arguably be implicit (it does not exist in the source). WDYT @SirYwell? Do we want the whole thing to be implicit (method and all) or explicit?

src/main/java/spoon/support/reflect/declaration/CtRecordImpl.java
I-Al-Istannen: I moved it to the `ParentExiter`.

I-Al-Istannen:
CI failure is unrelated and will fix itself.

src/main/java/spoon/support/reflect/declaration/CtRecordComponentImpl.java
```diff
@@ -36,31 +41,62 @@ public class CtRecordComponentImpl extends CtNamedElementImpl implements CtRecor
...
+		return makeTreeImplicit(field);
 	}
 
 	@Override
 	public boolean isImplicit() {
 		return true;
 	}
 
+	private CtTypeReference<?> getClonedType() {
+		//noinspection ReturnOfNull
```
SirYwell, line 96: we also have `@Nullable`

src/main/java/spoon/support/reflect/declaration/CtRecordComponentImpl.java
```diff
@@ -36,31 +41,62 @@ public class CtRecordComponentImpl extends CtNamedElementImpl implements CtRecor
...
-		return method;
+
+		CtFieldAccess<?> ctVariableAccess = (CtFieldAccess<?>) getFactory().Code()
+			.createVariableRead(getRecordFieldReference(), false);
+		// ensure the `this.` is implicit (important as we are also implicit)
+		ctVariableAccess.getTarget().setImplicit(true);
+
+		method.setBody(getFactory().Code().createCtReturn(ctVariableAccess));
+
+		return makeTreeImplicit(method);
```
SirYwell, line 55: I guess we can remove the other `setImplicit(true)` calls now?

SirYwell (changes requested):

SirYwell (approved):
Looks good.

PR #5860, opened 2024-06-18
review: fix: Make CtFieldRead inherit from CtResource

I-Al-Istannen:
Fields are allowed in try-with-resources statements now, so spoon should model this case. Before this patch, only `CtLocalVariable` implemented `CtResource`.

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/ParentExiter.java
+++ b/src/main/java/spoon/support/compiler/jdt/ParentExiter.java
@@ -62,7 +62,6 @@ import spoon.reflect.code.CtNewArray;
 import spoon.reflect.code.CtNewClass;
 import spoon.reflect.code.CtPattern;
 import spoon.reflect.code.CtRecordPattern;
-import spoon.reflect.code.CtResource;
 import spoon.reflect.code.CtReturn;
 import spoon.reflect.code.CtStatement;
 import spoon.reflect.code.CtSuperAccess;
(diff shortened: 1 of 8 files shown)


src/main/java/spoon/support/compiler/jdt/ParentExiter.java
```diff
@@ -1103,31 +1100,11 @@ public void visitCtTry(CtTry tryBlock) {
...
-							// we found the resource
-							// we clone it in order to comply with the contract of being a tree
-							final CtLocalVariable clone = v.clone();
-							clone.setImplicit(true);
-							tryWithResource.addResource(clone);
-							break;
-						}
-					}
-				}
-			}
```
MartinWitt, line 1130: I will miss this easy to understand and maintainable code.

src/test/java/spoon/test/trycatch/TryCatchTest.java
```diff
@@ -492,4 +498,80 @@ private <T extends Throwable> CtCatch createCatch(Factory factory, Class<T> type
...
+		return read;
+	}
+
+	private static CtModel createModelFromString(String code) {
+		Launcher launcher = new Launcher();
+		launcher.getEnvironment().setComplianceLevel(21);
+		launcher.getEnvironment().setShouldCompile(true);
+		launcher.addInputResource(new VirtualFile(code));
+		return launcher.buildModel();
+	}
```
MartinWitt, line 575: One day, we will have a single method for this :(

MartinWitt (approved):

src/test/java/spoon/test/trycatch/TryCatchTest.java
I-Al-Istannen: And other lies I tell myself? *despair*

I-Al-Istannen:
This is technically a breaking change (though I would hope the breakage isn't too much). The old model just isn't sustainable anymore with modern code and can not be salvaged.

MartinWitt:
@SirYwell ping

MartinWitt:
@SirYwell ping

monperrus:
Thanks for the modeling PR @I-Al-Istannen 

For, the record the corresponding update to handle the breaking change: https://github.com/SpoonLabs/npefix/commit/9512a41942926b255f1d441de7867c50b159c035

PR #5912, opened 2024-08-06
fix: fix modeling bug

monperrus:
expected behavior: buf is a field access

actual behavior: buf is a typeaccess

cause: isResolvedField assumed that everything is source-available and returned false if a library class (binary only, ByteArrayOutputStream here)

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java
+++ b/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java
@@ -723,7 +723,7 @@ public class JDTTreeBuilderHelper {
 		} else if (ref.isStatic()) {
 			target = createTypeAccess(qualifiedNameReference, ref);
 		} else if (!ref.isStatic() && !ref.getDeclaringType().isAnonymous()) {
-			if (!JDTTreeBuilderQuery.isResolvedField(qualifiedNameReference)) {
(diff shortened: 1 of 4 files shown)


SirYwell (commented):
From my understanding, this was only a problem in no classpath mode?

monperrus:
hi @SirYwell thanks for your review.

no, the serious bug is in fullclasspath, see https://github.com/INRIA/spoon/pull/5912/files#diff-be3ade2426b2d0edf889e62dc3e857c91c4c5935d502b792d7a751800fd327eaR181

so we want fully correct model in fullclasspath, and OKish behavior in noclasspath (this is the change in the other test)

SirYwell:
Thanks for clarifying.

Now, the model represents an implicit `this.<field>` access. In the mentioned code however, I wonder if it would make more sense to model it as `super.<field>` access, as the field is actually declared in a super class. I'm not sure how relevant that is for analysis, but I could imagine that it is relevant for transformations if they shadow a field (although that might break in even more places, e.g. printing).

I guess that's out of scope for this PR, but it's something that came to my mind seeing the code.

monperrus:
great, thanks @SirYwell . let's merge then to have it in the next beta

SirYwell (approved):

monperrus:
And now Guava can be spooned again: https://ci.inria.fr/sos/job/Guava/

Thanks @SirYwell

Issue #5934, opened 2024-08-17
Spelling mistake in AnnotationProcessor API


PinkMoustacheMan:
In the [AnnotationProcessor](https://github.com/INRIA/spoon/blob/15c8bb35e034a4d553cb1a87f413e686a06b2544/src/main/java/spoon/processing/AnnotationProcessor.java#L66) class, the "shoudBeConsumed" method is misspelled and should be called "shouldBeConsumed".

This change has an effect on the [AbstractAnnotationProcessor](https://github.com/INRIA/spoon/blob/15c8bb35e034a4d553cb1a87f413e686a06b2544/src/main/java/spoon/processing/AbstractAnnotationProcessor.java#L159) as well as all projects using this API.

The spelling is also incorrect for the "shoudBeProcessed" method in the [AbstractAnnotationProcessor](https://github.com/INRIA/spoon/blob/15c8bb35e034a4d553cb1a87f413e686a06b2544/src/main/java/spoon/processing/AbstractAnnotationProcessor.java#L163) but due to it being private, this should have a smaller impact.

monperrus:
Thanks for the bug report @PinkMoustacheMan .

I suggest to add a new method with correct spelling, and keep the previous one for backward compatibility `@Deprecated`

A pull-request is welcome!

PR #5943, opened 2024-08-20
fix: correct spelling mistake in API, doc and tests (#5934)

PinkMoustacheMan:
- mark misspelled method as deprecated
- add new method with correct spelling
- correct spelling of method in docs and test classes

### DIFF
--- a/doc/processor_annotations.md
+++ b/doc/processor_annotations.md
@@ -73,7 +73,7 @@ public interface AnnotationProcessor<A extends Annotation, E extends CtElement> 
 	boolean inferConsumedAnnotationType();
 	Set<Class<? extends A>> getProcessedAnnotationTypes();
 	Set<Class<? extends A>> getConsumedAnnotationTypes();
-	boolean shoudBeConsumed(CtAnnotation<? extends Annotation> annotation);
+	boolean shouldBeConsumed(CtAnnotation<? extends Annotation> annotation);
 }
 ```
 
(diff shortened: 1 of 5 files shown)


monperrus:
thanks for the PR. would you have a look at the failing CI job?

SirYwell (commented):
Should the method added to the interface be `default` and delegate to the deprecated method to avoid breaking changes?

monperrus:
Per the theory of software engineering, there should be no delegation:

- the deprecated method is frozen, kept as is for backward compatibility, and will be removed in the future after the deprecation period (do we have an official one in Spoon?)
- the new method is the one actively maintained.

Here, the code is so simple, so I'm happy to merge anyway when CI is green.

PinkMoustacheMan:
I do agree with the notion that deprecated methods should not change anymore, so I reverted that change.
I also fixed the whitespace issue, I thought this was not part of the checks (or that it would be fixed as part of a commit hook or CI step).

monperrus:
Thanks.

Why do we have a public and a private version of the same method AFAIU? see `private boolean shouldBeProcessed`

PinkMoustacheMan:
There is a public `shouldBeConsumed` method as well as a private `shouldBeProcessed` method.
They are doing the same thing but referencing different fields.

monperrus:
thanks a lot @PinkMoustacheMan

PR #5994, opened 2024-10-08
review: test: only run tests with compliance level >=8

SirYwell:
In preparation for #5977, we need to make sure that tests only run with a compliance level >= 8.

I went through the affected tests trying to figure out if there is a specific reason why they used a lower language level. Those where it doesn't seem to make sense to run on higher language levels are disabled now with a comment. I'm not sure if we want to remove them directly.

Note: There are more failing tests in #5977, but the others have different reasons that we need to analyze more closely I guess. That's why I'd like to address them separately.

### DIFF
--- a/src/main/java/spoon/Launcher.java
+++ b/src/main/java/spoon/Launcher.java
@@ -251,7 +251,7 @@ public class Launcher implements SpoonAPI {
 			// java compliance
 			opt2 = new FlaggedOption("compliance");
 			opt2.setLongFlag("compliance");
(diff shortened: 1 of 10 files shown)


I-Al-Istannen (approved):

I-Al-Istannen:
LGTM. We can delete them (they are in git) or leave them there as relics/documentation of old bugs. Deleting is probably fine.

I-Al-Istannen:
Thank you @SirYwell :)

PR #6043, opened 2024-10-27
review: feat: read jvm version for default compliance level

MartinWitt:
Maybe the current JVM version is a better default compared to 8 or?

### DIFF
--- a/src/main/java/spoon/support/StandardEnvironment.java
+++ b/src/main/java/spoon/support/StandardEnvironment.java
@@ -68,9 +68,9 @@ public class StandardEnvironment implements Serializable, Environment {
 	/**
 	 *
 	 * Only features available in the compliance level are correctly parsed by spoon.
-	 * So if you use 11, spoon cant parse records. By default, we use the language level of the executing jvm.
-	 * If you want to parse java 21 code with a java 17 jvm you need set the compliance level with {@link  #setComplianceLevel}
-	 * to least 21.
+	 * By default, spoon uses the language level of the executing JVM. So if you use Java 11, spoon can't parse records.
+	 * If you want to parse Java 21 code with a Java 17 JVM you need set the compliance level with {@link  #setComplianceLevel}
+	 * to at least 21.
 	 *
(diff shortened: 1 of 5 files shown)


SirYwell:
I think that's indeed a better default, but it's probably something we want to at least highlight in the next release? I also don't think we need to catch any exception around the version code, as it should work on Java 10 onwards and we're targeting Java 17.

MartinWitt:
> I think that's indeed a better default, but it's probably something we want to at least highlight in the next release?

Hmm, we could start collecting common answers and important behaviors on the website.

SirYwell:
> Hmm, we could start collecting common answers and important behaviors on the website.

Good idea. But maybe we can also improve the in-code documentation (e.g. add a comment to the `DEFAULT_CODE_COMPLIANCE_LEVEL` field). WDYT?

src/main/java/spoon/support/StandardEnvironment.java
```diff
@@ -65,7 +65,15 @@ public class StandardEnvironment implements Serializable, Environment {
 
 	private static final long serialVersionUID = 1L;
 
-	public static final int DEFAULT_CODE_COMPLIANCE_LEVEL = 8;
+	/**
+	 *
+	 * Only features available in the compliance level are correctly parsed by spoon.
+	 * So if you use 11, spoon cant parse records. By default, we use the language level of the executing jvm.
+	 * If you want to parse java 21 code with a java 17 jvm you need set the compliance level with {@link  #setComplianceLevel}
+	 * to least 21.
```
SirYwell, line 73:
```suggestion
	 * Only features available in the compliance level are correctly parsed by spoon.
	 * By default, spoon uses the language level of the executing JVM. So if you use Java 11, spoon can't parse records.
	 * If you want to parse Java 21 code with a Java 17 JVM you need set the compliance level with {@link  #setComplianceLevel}
	 * to at least 21.
```

SirYwell (approved):
:+1:

PR #6049, opened 2024-10-28
fix: Make SortedList#add return value adhere to List contract

### LINKED ISSUE #6034: [Bug]: spoon.support.util.SortedList.addAll(...) return value is not JLS compliant
### Describe the bug

According to Java JLS/API docs we see that:
https://docs.oracle.com/javase/8/docs/api/?java/util/Collections.html
addAll(...) Returns:
    true if the collection changed as a result of the call

However in spoon.support.util.SortedList.addAll(...) calling addAll(...) on an empty collection returns true, even though no actual change was made in the collections and secondly, if some element fails to be added, despite some of them are added will return false, when at least one object was inserted in the collection, and thus the collection was modified.

It should also be noted that LinkedList which SortedList extends always inserts all elements and returns true. It only returns false if no element could be added and that if some problem occurs at the middle of an insertion of multiple objects, an exception will be thrown.

We propose the following patch:
```
--- a/src/main/java/spoon/support/util/SortedList.java	2024-10-23 02:02:48.000000000 +0100
+++ b/src/main/java/spoon/support/util/SortedList.java	2024-10-23 17:25:47.531343136 +0100
@@ -42,9 +42,9 @@
 
 	@Override
 	public boolean addAll(Collection<? extends E> c) {
-		boolean ret = true;
+		boolean ret = false;
 		for (E e : c) {
-			ret &= add(e);
+			ret |= add(e);
 		}
 		return ret;
 	}
```

_No response_

_No response_

### Actual output

_No response_

### Expected output

_No response_

### Spoon Version

latest from master

### JVM Version

Java 17, but not relevant for this problem

NixOS Linux, but again not relevant for this issue

CoreRasurae:
- Fixes #6034 
- SortedList.addAll(...) should return a boolean value indicating if the Collection was modified is any way as a result of the call
- Previously addAll(...) would return false if only partial changes occurred and would only return true if all elements had been inserted, or if adding an empty list which in not in accordance to the Java JLS/API as documented in https://docs.oracle.com/javase/8/docs/api/?java/util/Collections.html.

### DIFF
--- a/src/main/java/spoon/support/util/SortedList.java
+++ b/src/main/java/spoon/support/util/SortedList.java
@@ -42,9 +42,9 @@ public class SortedList<E> extends LinkedList<E> {
 
 	@Override
 	public boolean addAll(Collection<? extends E> c) {
-		boolean ret = true;
+		boolean ret = false;
 		for (E e : c) {
-			ret &= add(e);
+			ret |= add(e);
 		}
 		return ret;
 	}
--- a/src/test/java/spoon/support/util/SortedListTest.java
+++ b/src/test/java/spoon/support/util/SortedListTest.java
@@ -1,0 +1,154 @@
+package spoon.support.util;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.LinkedList;
+import java.util.List;
+
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.function.Executable;
+
+import spoon.reflect.declaration.CtClass;
+import spoon.reflect.declaration.CtElement;
+
+public class SortedListTest {
+    public class BasicComparator implements Comparator<CtElement> {
+
+        private final List<CtElement> myOrdering;
+        
+        public BasicComparator(List<CtElement> orderList) {
+            this.myOrdering = orderList;
+        }
+        
+        @Override
+        public int compare(CtElement o1, CtElement o2) {
+            int idxFirst = - 1;
+            int idxSecond = -1;
+            int idx = 1;
+            for(CtElement clazz : myOrdering) {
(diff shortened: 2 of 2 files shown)


src/main/java/spoon/support/util/SortedList.java
```diff
@@ -42,9 +42,9 @@ public void add(int index, E element) {
 
 	@Override
 	public boolean addAll(Collection<? extends E> c) {
-		boolean ret = true;
+		boolean ret = false;
 		for (E e : c) {
-			ret &= add(e);
+			ret |= add(e);
```
SirYwell, line 47: As we'll always return `true` this way (because `add` always returns true), we can probably also just return `!c.isEmpty()` (assuming `c` doesn't get modified in between) or just do `ret = true`, similar to the default implementation in `AbstractList`. What do you think? (Seems like the Qodana inspection is a bit aggressive here...)

SirYwell (commented):
One minor thing, other than that, please make sure your PR title follows our guidelines (<https://github.com/INRIA/spoon/blob/master/CONTRIBUTING.md#guidelines-for-pull-requests>). Thanks!

src/main/java/spoon/support/util/SortedList.java
CoreRasurae: Qodana is right i didn't notice it was using bit-wise operations instead of logical operations i have fixed that now

CoreRasurae: I hope the PR title is okay now

CoreRasurae: i hope that the title is okay now... trying to follow the procedures that indicate that fic and the number of the issue needs to be present too

CoreRasurae: Ping?

SirYwell: Please use one of my suggested variants or your initial version instead of the `||`. The title is fine now.

CoreRasurae: Okay, i've reverted the change to use bit-wise logic again.

SirYwell (approved):

CoreRasurae:
I am unable to merge... The base branch restricts merging to authorized users.

SirYwell:
> @CoreRasurae Qodana still fails. Could you take a look? @SirYwell would merge afterwards.

The Qodana problem is basically a false positive. A short-circuit operation isn't intuitive there at all imho.

PR #6051, opened 2024-10-30
review: test: easy type access for ModelTests

SirYwell:
This is somewhat an experiment based on our `ModelTest` annotation. We can reduce *arrange* code by extracting required `CtType`s centrally and inject them via parameters. WDYT?


I-Al-Istannen (approved):
I like it. Seems simple and useful

MartinWitt:
Nice work, getting a single type always felt like the missing piece.

PR #6053, opened 2024-11-02
review: chore: remove depclean

MartinWitt:
It is broken for a year, maybe we should remove it instead of skipping it.


I-Al-Istannen (approved):
The master branch seems broken and there hasn't been a release since January 2023 from what I can tell. So this seems unlikely to change.

SirYwell (approved):

PR #6060, opened 2024-11-07
Remove junit dependency in spoon-control-flow

bolu61:
The `junit` dependency is causing test compilation to fail.


monperrus:
thanks @bolu61

PR #6084, opened 2024-11-25
review: test: Fixed 4 non-deterministic flaky tests 

nikunjagarwal321:
### What changes were proposed in this pull request?
Sorting the order of _imports_ list variable  and _enums_ list variable in 4 flaky tests in order to make them deterministic. The assertions are written with an assumption that the list will always be in a particular predefined order, but that may not be the case. The PR contains fix for the below test cases to make them deterministic:

- spoon.test.imports.ImportTest#staticImports_ofNestedTypes_shouldBeRecorded
- spoon.test.imports.ImportTest#staticTypeAndMethodImport_importShouldAppearOnlyOnceIfTheirSimpleNamesAreEqual
- spoon.test.position.PositionTest#testPositionOfCtImport
- spoon.test.template.SubstitutionTest#testCreateTypeFromTemplate

### Why are the changes needed?
Multiple flaky tests were detected while trying to run the tests using the nondex tool. [NonDex](https://github.com/TestingResearchIllinois/NonDex) is a tool for detecting and debugging wrong assumptions on under-determined Java APIs. 

#### Sample ERROR logs: 
 
Screenshot 2024-11-24 at 8 40 54 PM

Screenshot 2024-11-24 at 8 43 33 PM

### Reason for Failure 
As mentioned above, the assertions assume that the list will always be in a particular predefined order, but that may not be the case. For eg, in test case : staticImports_ofNestedTypes_shouldBeRecorded, we have the following test case

```
List<CtImport> imports = mainType.getPosition().getCompilationUnit().getImports();
assertThat(imports, hasSize(2));

CtImport import0 = imports.get(0);
assertThat(import0.getReference().getSimpleName(), is("InnerClass"));
```
The above code assumes that the first element of the list will be `InnerClass` which may not be true. This is because the imports is being set by `JDTImportBuilder.java` class where the [imports](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/compiler/jdt/JDTImportBuilder.java#L47) variable is implemented as HashSet. As per the javadocs, the order of elements is not maintained in HashSet.
Screenshot 2024-11-24 at 8 55 01 PM

Hence, we sort the fields before asserting based on certain fields to make the order deterministic.

#### Steps to reproduce flakiness using nondex -
```
mvn install -DskipTests

mvn edu.illinois:nondex-maven-plugin:2.1.7:nondex -Dtest=spoon.test.position.PositionTest#testPositionOfCtImport
mvn edu.illinois:nondex-maven-plugin:2.1.7:nondex -Dtest=spoon.test.template.SubstitutionTest#testCreateTypeFromTemplate
mvn edu.illinois:nondex-maven-plugin:2.1.7:nondex -Dtest=spoon.test.imports.ImportTest#staticTypeAndMethodImport_importShouldAppearOnlyOnceIfTheirSimpleNamesAreEqual
mvn edu.illinois:nondex-maven-plugin:2.1.7:nondex -Dtest=spoon.test.imports.ImportTest#staticImports_ofNestedTypes_shouldBeRecorded
```

Please let me know if you have any questions or would like to discuss this further for any clarificaitions.

### DIFF
--- a/src/main/java/spoon/support/util/RtHelper.java
+++ b/src/main/java/spoon/support/util/RtHelper.java
@@ -22,12 +22,7 @@ import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
 
 /**
  * This class is a helper for runtime reflection.
@@ -50,7 +45,9 @@ public abstract class RtHelper {
 
 	private static void addAllFields(Class<?> c, List<Field> fields) {
 		if (c != null && c != Object.class) {
-			Collections.addAll(fields, c.getDeclaredFields());
+			Field[] declaredFields = c.getDeclaredFields();
+			Arrays.sort(declaredFields, Comparator.comparing(Field::getName));
+			Collections.addAll(fields, declaredFields);
 			addAllFields(c.getSuperclass(), fields);
 			for (Class<?> iface : c.getInterfaces()) {
 				addAllFields(iface, fields);
--- a/src/test/java/spoon/test/imports/ImportTest.java
+++ b/src/test/java/spoon/test/imports/ImportTest.java
@@ -89,6 +89,7 @@ import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -1877,6 +1878,7 @@ public class ImportTest {
 
 		// assert
 		List<CtImport> imports = mainType.getPosition().getCompilationUnit().getImports();
+		imports.sort(Comparator.comparing(importElement -> importElement.getReference().getSimpleName()));
 		assertThat(imports, hasSize(2));
 
 		CtImport import0 = imports.get(0);
@@ -1894,6 +1896,7 @@ public class ImportTest {
 		// assert
 		List<CtImport> imports = mainType.getPosition().getCompilationUnit().getImports();
 		assertThat(imports, hasSize(2));
+		imports.sort(Comparator.comparing(importElement -> importElement.getImportKind().toString()));
 
 		CtImport import0 = imports.get(0);
 		assertThat(import0.getImportKind(), is(CtImportKind.METHOD));
--- a/src/test/java/spoon/test/position/PositionTest.java
+++ b/src/test/java/spoon/test/position/PositionTest.java
@@ -19,6 +19,7 @@ package spoon.test.position;
 
 import java.io.File;
 import java.io.IOException;
+import java.util.Comparator;
 import java.util.Iterator;
 import java.util.List;
 
@@ -806,6 +807,8 @@ public class PositionTest {
 			AnnonymousClassNewIface.class);
 		String originSources = foo.getPosition().getCompilationUnit().getOriginalSourceCode();
 		List<CtImport> imports = foo.getPosition().getCompilationUnit().getImports();
+		// Sorting using position of import element so that they are sorted in the order they are imported
+		imports.sort(Comparator.comparing(importElement -> importElement.getPosition().getSourceStart()));
 		assertEquals(2, imports.size());
 		Iterator<CtImport> iter = imports.iterator();
 		{
--- a/src/test/java/spoon/test/template/SubstitutionTest.java
+++ b/src/test/java/spoon/test/template/SubstitutionTest.java
@@ -176,8 +176,12 @@ public class SubstitutionTest {
         assertNotNull(genEnum);
         assertSame(genEnum, factory.Type().get("generated.GenEnum"));
         assertEquals(2, genEnum.getEnumValues().size());
-        assertEquals("GOOD", genEnum.getEnumValues().get(0).getSimpleName());
-        assertEquals("BETTER", genEnum.getEnumValues().get(1).getSimpleName());
+        List<String> enumValueNames = genEnum.getEnumValues().stream()
+                .map(CtEnumValue::getSimpleName)
+                .sorted()
+                .toList();
+        assertEquals("GOOD", enumValueNames.get(1));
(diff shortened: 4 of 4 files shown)


nikunjagarwal321:
Added fix for another flaky test : 
- spoon.test.processing.ProcessingTest#testInitPropertiesWithWrongType

The change is to sort the results of getDeclaredFields() in `RTHelper.java` to make it deterministic. As per the java docs, the given method does not return the results in a deterministic order, which leads to the nondeterministic behaviour of the above test case.

Screenshot 2024-12-04 at 8 53 16 PM


Hence, we are sorting the fields in the function `addAllFields()` so as to make it deterministic.

SirYwell:
Hey, thank you for your PR.

I'm not sure if the tests are at fault or if the implementations are broken.

For example the enum constants situation: the test asserts that substituting `case1` with `GOOD` and `case2` with `BETTER` in the enum 
```java
public enum AnEnumModel {
	case1, case2;
}
```
Now, the resulting enum must have `GOOD, BETTER` in that exact order, otherwise the substitution misbehaved (and the test should fail).

Can you point out where exactly the order is lost in the underlying code?

Similarly for the imports. There, the order isn't as important, but I think it is reasonable to expect (and we could also specify that in the Javadocs) that the order will match the order in the source code.

In that case, I'd prefer having one separate Pull Request per problem. 

Regarding the `getDeclaredFields()` method, while it doesn't guarantee any order, I think we should not sort it in the model but be more flexible in the tests. But I'd like to know where exactly the problem occurs to understand it better.

nikunjagarwal321:
Hi @SirYwell 
Thanks for your comment and I totally agree with it. 

1. ### Imports Ordering Issue: 

> Can you point out where exactly the order is lost in the underlying code?
> 
> Similarly for the imports. There, the order isn't as important, but I think it is reasonable to expect (and we could also specify that in the Javadocs) that the order will match the order in the source code.

I debugged further and found that for imports, the order is changing from [JDTImportBuilder.java](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/compiler/jdt/JDTImportBuilder.java#L47)

**Root Cause** : In `JDTImportBuilder`, we are using `HashMap` which does not guarantee the order of elements.

If the order of imports is expected to be in a particular order, we may need to change the code in the JDTImportBuilder here. Otherwise, if we are okay with a nondeterministic order of the function, we can improve the test case as done in the PR. 

Please let me know what is the expected behaviour here so that I can proceed with that fix.

---- 

2. ### getDeclaredFields() Ordering Issue: 

> Regarding the `getDeclaredFields()` method, while it doesn't guarantee any order, I think we should not sort it in the model but be more flexible in the tests. But I'd like to know where exactly the problem occurs to understand it better.

The given fix is for spoon.test.processing.ProcessingTest#testInitPropertiesWithWrongType test, where we are setting the fields through ProcessorProperties and then [asserting](https://github.com/INRIA/spoon/blob/master/src/test/java/spoon/test/processing/ProcessingTest.java#L275) on the fields.
 
```
		props.set("aString", "foo");
		props.set("anObject", "foo");
		props.set("anInt", "foo");

		try {
			ProcessorUtils.initProperties(p, props);
			fail();
		} catch (SpoonException e) {
			assertTrue(e.getMessage().contains("anInt"));
		}


		assertEquals("foo", p.aString);
		assertEquals(0, p.anInt);
		assertNull(p.anObject);
```
In the given test, the assertions are expecting that the value of attribute aString will be set first, then the value of anInt will be set, which will throw SpoonException. 
But this may not be the case. It will fail in the following scenarios:

- If `initProperties` tries to set the value of `anInt` first, then the value of `aString` will still remain NULL and the assertion will fail. 
- Similarly, if the value of `anInt` is set after setting `anObject`, that will also fail, since the value of anObject will be set as "foo". 

Hence, to fix this test case, we may have to remove the assertion on aString and anObject. Please let me know your views on it or if I can provide any more details. Post that, I'll proceed with the change in a separate PR as you suggested.

---- 

3. ### Enums Ordering Issue: 

> Now, the resulting enum must have `GOOD, BETTER` in that exact order, otherwise the substitution misbehaved (and the test should fail).

For the enumValues, I am still investigating. I have narrowed it down to function [generateTargets](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/pattern/internal/DefaultGenerator.java#L88) in `DefaultGenerator.java` and am looking into it. Will update you once I find the exact implementation which affects this.

nikunjagarwal321:
Hi @SirYwell , 
Please let me know if the above changes make sense. I can open separate PRs for the changes or make change in the same one as you say.

SirYwell:
> Hi @SirYwell , Please let me know if the above changes make sense. I can open separate PRs for the changes or make change in the same one as you say.

Hi, if you found the root cause for the enum value problem, this seems to be the most relevant one. 
For the import order, a separate PR would be appreciated. 
For the reflective getDeclaredField call, I'm not sure what's sensible.

nikunjagarwal321:
> Hi, if you found the root cause for the enum value problem, this seems to be the most relevant one. For the import order, a separate PR would be appreciated. For the reflective getDeclaredField call, I'm not sure what's sensible.

Hi,
Sure.

1.  For enum value problem, I'll check this again. Earlier, I was able to narrow it down to [generateTargets](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/pattern/internal/DefaultGenerator.java#L88) in DefaultGenerator.java.

2. For import order flaky tests, I have raised a new PR: https://github.com/INRIA/spoon/pull/6123

3. For the reflective getDeclaredField call, may be we can remove the following 2 assertions  from [ProcessingTest.java](https://github.com/INRIA/spoon/blob/master/src/test/java/spoon/test/processing/ProcessingTest.java#L275) as they depend on the order of getDeclaredFields()
```
		assertEquals("foo", p.aString);
		assertNull(p.anObject);
```

Or maybe we can change the code to: 
```
		props.set("aString", "foo");
		props.set("anObject", "foo");
		ProcessorUtils.initProperties(p, props);
		try {
			props.set("anInt", "foo");
			ProcessorUtils.initProperties(p, props);
			fail();
		} catch (SpoonException e) {
			assertTrue(e.getMessage().contains("anInt"));
		}


		assertEquals("foo", p.aString);
		assertEquals(0, p.anInt);
		assertEquals("foo", p.anObject);
```
This will test the behaviour of wrongDataType correctly irrespective of the order the fields are returned. Let me know your thoughts on this.

PR #6094, opened 2024-12-09
review: fix: Fix missing parens for `(string + string).contains(...)`

Luro02:
For the code `(string + string).contains(string)` will remove the parenthesis which results in `string + string.contains(string)`


I-Al-Istannen (approved):

SirYwell (approved):
The change looks good, but I think this is something where we need to go through all expressions and check whether they need parentheses. But I can do that as a follow-up PR.

PR #6095, opened 2024-12-09
fix: Crashes when records contained static fields

I-Al-Istannen:
Spoon tried to remove the field to prevent record field duplication. This fails for static fields, which do not exist (yet). For these, spoon passed null to the remove method, which failed with an exception. We now check for the modifier and null explicitly.

Reported by @Luro02.


SirYwell (approved):
Looks good.

PR #6107, opened 2024-12-17
review: Use java 17 for tests

I-Al-Istannen:
Previously spoon-pom used not-existent properties to configure the maven compiler, which tripped up intellij. This patch consistently uses maven.compiler.(test)Release.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -136,7 +136,7 @@
             mvn -q versions:use-latest-versions -DallowSnapshots=true -Dincludes=fr.inria.gforge.spoon
             mvn -q versions:update-parent -DallowSnapshots=true
             git diff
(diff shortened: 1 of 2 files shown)


SirYwell:
We set `java.src.version` for one mvn call in flake.nix, I assume this doesn't affect anything anymore this way? And it also seems like SpoonPom is using this option but doesn't consider the release flag...

MartinWitt:
@SirYwell How is the evaluation progressing?

I-Al-Istannen:
@SirYwell Shall I also have a look? :P

SirYwell (approved):
A very welcome change, thanks

monperrus:
thanks @I-Al-Istannen

PR #6162, opened 2025-01-28
fix: avoid unnecessary downcast in ReferenceBuilder

tdegueul:
TL;DR: The cast to `(TreeBulderCompiler)` is overly restrictive here.

### Context

For a specific use case, I would like Spoon to use a custom JDT parser/compiler. Specifically, I do not care about method bodies in the parsed code as I only focus on the extracted API, so passing `ignoreMethodBodies` to JDT is a _huge_ performance boost in terms of time and memory (30 to 70% faster `buildModel()` on my projects).

My understanding is that Spoon does not and will not provide an "easy" way for clients to provide their own JDT parser/compiler or their own parsing/compiling options to JDT, which is perfectly understandable.

My current solution involves injecting a custom `JDTBatchCompiler` into the `Launcher`. Because `JDTBatchCompiler` depends on the package-private `TreeBuilderCompiler`, one must implement its own `Compiler` implementation instead, which eventually propagates to this unnecessary downcast and crashes. This PR would allow anyone to pass their own compiler without disrupting Spoon's policy regarding JDT customizations.

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
+++ b/src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
@@ -8,6 +8,7 @@
 package spoon.support.compiler.jdt;
 
 import org.eclipse.jdt.core.compiler.CharOperation;
+import org.eclipse.jdt.internal.compiler.Compiler;
 import org.eclipse.jdt.internal.compiler.ast.ASTNode;
 import org.eclipse.jdt.internal.compiler.ast.AllocationExpression;
 import org.eclipse.jdt.internal.compiler.ast.Annotation;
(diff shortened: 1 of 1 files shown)


monperrus (approved):
LGTM, approving, thanks @tdegueul

I-Al-Istannen:
@SirYwell did something similar for me here: https://github.com/SirYwell/JavadocApi/commit/e0f45b448b221d9c392daffe46a925d104d79a52 (now subsumed by javadoc support in the spoon.javadoc module and <https://github.com/I-Al-Istannen/JavadocIndexer>)

This does not pass the flag to JDT but prevents spoon from converting the JDT AST instead. This was still quite nice for performance, IIRC. You aren't the first one with this goal, apparently, but probably more determined :P

+1 from me for making this easier.

tdegueul:
> @SirYwell did something similar for me here: [SirYwell/JavadocApi@e0f45b4](https://github.com/SirYwell/JavadocApi/commit/e0f45b448b221d9c392daffe46a925d104d79a52)

That's a nice trick, too! Specializing the `JDTBasedSpoonCompiler` is currently doable, but the issues arise when specializing the `TreeBulderCompiler`, hence the PR.

Thanks, for approval :)

Issue #2228, opened 2018-07-15
refactor: fix all true positives on SonarQube


zielint0:
I am using IDEA checkers to improve our code base but, at the same time some Sonar defects are solved as well (some checkers overlap).

I am going to fix all true positives on SonarQube on regular basis.

Today (2018-07-16) I am starting with  2076 issues to fix.

One more time I would like to get labels:
- code inspection (yellow)
- sonarqube (orange)

to manage PRs with fixes. Also it would be change color of "bug" to RED :)

"refactor" label is better suited for refactoring of pure functionality (for instance: rework parser to support syntax X).

zielint0:
Thanks to fixing defects using IDEA Code Inspection #2105, our SonarQube Quality Gate changed status from Red to Greem :-)

https://sonarqube.ow2.org/project/activity?id=fr.inria.gforge.spoon%3Aspoon-core

```
July 16, 2018
4:46 AM
7.1.0-SNAPSHOT
Quality Gate: Green (was Red)
```

zielint0:
How often SonarQube results of analysis are refreshed?

TheMarvelFan:
Hi can someone guide me how to start working on this issue?

MartinWitt:
We no longer use SonarQube anymore, so this issue is finished. But IntelliJ still highlights a few warnings which you could fix.

PR #6172, opened 2025-02-03
refactor: fixing sonarqube issues in Pattern.java

rishivijayv:
https://github.com/INRIA/spoon/issues/2228

This fixes a few sonarqube warnings in the file `Pattern.java` that were pointed out by the [SonarQube for IDE Plugin](https://plugins.jetbrains.com/plugin/7973-sonarqube-for-ide). Namely, the issues were:

- **java:S1905** (Redundant casts should not be used)
- **java:S1612** (Lambdas should be replaced with method references)
- **java:S1602** (when a lambda expression uses block notation while expression notation could be used)

I believe removing the explicit cast to `Map<...,...>` also seems to get rid of one of the warnings during compile time, so..._yay!_ ?


monperrus:
thanks @rishivijayv

Issue #5216, opened 2023-05-17
OpenSSF scorecard of Spoon


monperrus:
Google's Deps.Dev supports [Open Source Security Foundation](https://openssf.org/) scorecard

https://deps.dev/maven/fr.inria.gforge.spoon%3Aspoon-core/10.3.0

Good news:
* Maintained 10/10
* License 9/10
* Dangerous-Workflow 10/10
* Security-Policy 10/10
* Binary-Artifacts 9/10

Bad news:
* CII-Best-Practices 0/10
* Signed-Releases 0/10
* Token-Permissions 0/10
* Fuzzing 0/10
* Pinned-Dependencies 0/10
* Vulnerabilities 0/10

How to fix that?

monperrus:
In particular, we should already have Signed-Releases and  Pinned-Dependencies?

What's wrong here?

monperrus:
"Signed-Releases": we have old releases with unsigned commits. 

Possible solutions:
* we remove them :-1: 
* we sign commits a posteriori (doable? would require to contact past contributors who released)
* we upload a GPG signature for them: is that supported by Deps.dev? what's the expected format? naming convention?

MartinWitt:
On the list, an interesting point is https://google.github.io/clusterfuzzlite/. I always here good feedback about fuzzing, maybe it is time to try it for spoon?

MartinWitt:
@monperrus you as an owner has to apply here: https://bestpractices.coreinfrastructure.org/en. Then we should get 10 instead of 2 points in the category.

monperrus:
> you as the owner have to apply here: [bestpractices.coreinfrastructure.org/en](https://bestpractices.coreinfrastructure.org/en).

done.

should we add the badge to the README?

MartinWitt:
Hey, I updated our scorecode, and now we have 94% see https://bestpractices.coreinfrastructure.org/en/projects/7377/edit. If you have a different opinion on some fields, feel free to change them.

> should we add the badge to the README?

We can include it yes.

monperrus:
Trusty score computed by trustypkg.dev, from stacklok
https://www.trustypkg.dev/maven/fr.inria.gforge.spoon%3Aspoon-core

Score 6.8, green, seems good

monperrus:
OpenSSF Best Practices: https://www.bestpractices.dev/en/projects/7377

We are almost perfect.

There is a still a bug with https, and we would need some accepted dynamic analysis.

[...]

PR #6182, opened 2025-02-17
feat: add dirty-waters workflow to CI

randomicecube:
cc @monperrus @Stamp9

Relates to https://github.com/INRIA/spoon/issues/5216, https://github.com/chains-project/dirty-waters/issues/37, https://github.com/chains-project/dirty-waters/issues/58

Key notes:

- [ ] `x_to_fail` parameter: the percentage of a single non-high severity issue present among the dependencies for the CI to break. It defaults to 5%, so what should it be here?
- [ ] `comment_on_commit`: whether the reports are allowed to be pasted as comments in the commits, in the case of high-severity issues breaking CI. Defaults to false, what do we want here?
- [ ] `allow_pr_comments`: whether the reports are allowed to be pasted as comments in pull requests. Defaults to true, what do we want here?
- [ ] By default, only static analysis happens. Do we want differential to be performed in certain scenarios too? If so, which ones?
- [ ] Gradual reporting is enabled by default -- only one of the reported smells has its table displayed per time, with the higher severity issues showing up first.
- [ ] In https://github.com/chains-project/sbom.exe/pull/317, @algomaster99 included a step before the analysis which builds the project and skips tests; should the same be done here?

For more information on the action, see the [wiki](https://github.com/chains-project/dirty-waters-action/wiki).

### DIFF
--- a/.github/workflows/dirty-waters.yml
+++ b/.github/workflows/dirty-waters.yml
@@ -1,0 +1,28 @@
+name: dirty-waters
+
+on:
+  pull_request:
+  push:
+    branches: [ master ]
+
+jobs:
+  dirty-waters:
+    runs-on:
+        ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: true
+      - name: Setup JDK17
+        uses: actions/setup-java@v4
+        with:
+          distribution: 'temurin'
+          java-version: '17'
+      - name: Dirty Waters Analysis
+        uses: chains-project/dirty-waters-action@v1.6
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          package_manager: maven
+          project_repo: ${{ github.repository }}
+          latest_commit_sha: ${{ github.sha }}
+          github_event_before: ${{github.event.before}}


algomaster99:
>     included a step before the analysis which builds the project and skips tests; should the same be done here?

Skipping tests first time is necessary for me because I want to build the agent to run the tests. But if run the tests straightaway, the agent does not exist so tests fail and I also don't get the built agent.
But I would recommend doing so. I don't think so any of the software supply chain smells in dirty-waters are test dependent.

monperrus:
> x_to_fail parameter: the percentage of a single non-high severity issue present among the dependencies for the CI to break. It defaults to 5%, so what should it be here?

the current value, we should never go higher than this.

> comment_on_commit: whether the reports are allowed to be pasted as comments in the commits, in the case of high-severity issues breaking CI. Defaults to false, what do we want here?

false

> allow_pr_comments: whether the reports are allowed to be pasted as comments in pull requests. Defaults to true, what do we want here?

If the comment it only in case of CI failure yes, else no.  

> By default, only static analysis happens. Do we want differential to be performed in certain scenarios too? If so, which ones?

no, not needed for this MVP

> Gradual reporting is enabled by default -- only one of the reported smells has its table displayed per time, with the higher severity issues showing up first.

full reporting.


thanks @randomicecube

randomicecube:
> If the comment it only in case of CI failure yes, else no.

It currently isn't, but I'll make it so

Thanks for the feedback, I'll ping again when everything is done

I-Al-Istannen:
I also want to note that I will not merge any per-commit/PR workflow that takes longer than ~10 Minutes (ideally much shorter). The CI already takes ages to run, I don't want to wait 40 minutes for a result.

randomicecube:
@I-Al-Istannen  Absolutely, there's a bug which I'm working on fixing right now which is making it so that the cache isn't being stored with the correct key -- I'll fix it, test it, link the test results (and run them here too)
With caching, the action is fast!

randomicecube:
From my tests in the fork I have, this doesn't work yet either with v1.8, so no need to accept the workflow just yet

monperrus:
@randomicecube can we get that one merged?

randomicecube:
@monperrus haven't touched on the project in a bit but Maven should work fine (bear in mind, as previously discussed, cold runs may take a while, subsequent ones should be fast)

Issue #5357, opened 2023-07-20
[Bug]: package-info incorrect processing


embraceddd:
### Describe the bug

Attaching demo-project https://github.com/embraceddd/spoonpckinfobug
Quickstart for it:
[spoonpckinfobug](https://github.com/embraceddd/spoonpckinfobug/tree/master)/[src](https://github.com/embraceddd/spoonpckinfobug/tree/master/src)/[main](https://github.com/embraceddd/spoonpckinfobug/tree/master/src/main)/[resources](https://github.com/embraceddd/spoonpckinfobug/tree/master/src/main/resources)/[buggedProject](https://github.com/embraceddd/spoonpckinfobug/tree/master/src/main/resources/buggedProject)/[src](https://github.com/embraceddd/spoonpckinfobug/tree/master/src/main/resources/buggedProject/src)/[main](https://github.com/embraceddd/spoonpckinfobug/tree/master/src/main/resources/buggedProject/src/main)/[java](https://github.com/embraceddd/spoonpckinfobug/tree/master/src/main/resources/buggedProject/src/main/java)/[demo](https://github.com/embraceddd/spoonpckinfobug/tree/master/src/main/resources/buggedProject/src/main/java/demo)

There are two packages [notwork](https://github.com/embraceddd/spoonpckinfobug/tree/master/src/main/resources/buggedProject/src/main/java/demo/notwork) and [work](https://github.com/embraceddd/spoonpckinfobug/tree/master/src/main/resources/buggedProject/src/main/java/demo/work) that contains package-info files. Package-info file in package [work](https://github.com/embraceddd/spoonpckinfobug/tree/master/src/main/resources/buggedProject/src/main/java/demo/work) proccesing fine in all ways. 
For package-info file in package [notwork](https://github.com/embraceddd/spoonpckinfobug/tree/master/src/main/resources/buggedProject/src/main/java/demo/notwork) we can get different result depends on app dependencies and application startup method

1) If we **HAVE** jakarta-xml-bind dependency in our project and we will start app with shorten-command-line **_JARManifest_** method, app will fall down with JLSViolation.

2) If we **DON'T** have jakarta-xml-bind dependency in our project and we will start app with shorten-command-line **_JARManifest_** method, app will fall down with JLSViolation.

3) If we **HAVE** jakarta-xml-bind dependency in our project and we will start app with shorten-command-line **_ARGFILE_** method, app will successflyy process package-info file.

[...]

PR #6183, opened 2025-02-17
test: Separate imports work for annotations in package-info.java

rishivijayv:
https://github.com/INRIA/spoon/issues/5357

In the issue I've documented my attempts to try and reproduce the issue highlighted in the original comment with the help of a unit test. However, I have been unable to do so.

Spoke with @algomaster99 and we felt adding an explicit test to demonstrate that the issue _seems_ to be fixed might be a good way to state this in the code base. 

So, this PR adds a test where I try to imitate the code posted by the original poster in https://github.com/INRIA/spoon/issues/5357, but no Exceptions seem to be thrown here. 

Please let me know if I've overlooked something and my attempts to recreate the issue (documented in the issue starting [here](https://github.com/INRIA/spoon/issues/5357#issuecomment-2647585111)) were misguided and there is indeed still a bug that might need fixing. 

cc @algomaster99 @monperrus @I-Al-Istannen would appreciate a look whenever you get a chance! :) 🙏🏽


src/test/java/spoon/test/pkg/PackageTest.java
```diff
@@ -191,6 +186,19 @@ public void testAnnotationInPackageInfoWhenTemplatesCompiled() throws Exception
 		canBeBuilt("./target/spooned/packageAndTemplate/spoon/test/pkg/package-info.java", 8);
 	}
 
+	@Test
+	@GitHubIssue(issueNumber = 5357, fixed = true)
+	public void testSeparateImportForPackageInfoAnnotation() {
+		assertDoesNotThrow(() -> {
```
algomaster99, line 192: Can you also specify the specific class of exception?

src/test/java/spoon/test/pkg/PackageTest.java
```diff
@@ -191,6 +186,19 @@ public void testAnnotationInPackageInfoWhenTemplatesCompiled() throws Exception
 		canBeBuilt("./target/spooned/packageAndTemplate/spoon/test/pkg/package-info.java", 8);
 	}
 
+	@Test
+	@GitHubIssue(issueNumber = 5357, fixed = true)
+	public void testSeparateImportForPackageInfoAnnotation() {
```
algomaster99, line 191: You can further refactor this test using [`ModelTest`](https://github.com/INRIA/spoon/blob/361bc03ba9b6a3f488e8572d553a7f419ab807e7/src/test/java/spoon/testing/utils/ModelTest.java#L18).

algomaster99 (changes requested):

src/test/java/spoon/test/pkg/PackageTest.java
rishivijayv: Hmm I might be looking in the wrong place, but I wasn't able to find a method that asserts that a specific exception _is not_ thrown. [This](https://stackoverflow.com/questions/17731234/how-to-test-that-no-exception-is-thrown) is the SO post I referred to for getting the method, and [this](https://stackoverflow.com/questions/70189729/is-there-a-way-to-test-that-a-java-method-does-not-throw-a-specific-type-of-exce) is a SO post I found about asserting a specific exception not being thrown -- both of which suggest that there's not a method for this 🤔 

There is an `assertThrows()` which accepts a specific exception as the first argument, but I wasn't able to find something similar for `assertDoesNotThrow()`

src/test/java/spoon/test/pkg/PackageTest.java
rishivijayv: Oooh that's a pretty handy decorator! If we use that though, won't that pretty much cover the entire body of this test (since we just create a model here)? 

Might be missing something but my understanding was that we wanted to test that this "setting-up" code explicitly does not throw an exception. So if we use the `@ModelTest` annotation, what would we wrap inside the `assertDoesNotThrow`'s lambda? 🤔

src/test/java/spoon/test/pkg/PackageTest.java
algomaster99: Oh right. That makes sense. The behaviour of `assertDoesNotThrow` would not be clear otherwise. :sweat_smile:

src/test/java/spoon/test/pkg/PackageTest.java
algomaster99: >  what would we wrap inside the assertDoesNotThrow's lambda?

also makes sense! Then you can let this be.

src/test/java/spoon/test/pkg/PackageTest.java
```diff
@@ -55,18 +55,13 @@
...
 
 import static org.hamcrest.CoreMatchers.equalTo;
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertFalse;
-import static org.junit.jupiter.api.Assertions.assertNotNull;
-import static org.junit.jupiter.api.Assertions.assertNull;
-import static org.junit.jupiter.api.Assertions.assertSame;
-import static org.junit.jupiter.api.Assertions.assertTrue;
-import static org.junit.jupiter.api.Assertions.fail;
+import static org.junit.jupiter.api.Assertions.*;
```
I-Al-Istannen, line 64:
```suggestion
import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
import static org.junit.jupiter.api.Assertions.assertEquals;
import static org.junit.jupiter.api.Assertions.assertFalse;
import static org.junit.jupiter.api.Assertions.assertNotNull;
import static org.junit.jupiter.api.Assertions.assertNull;
import static org.junit.jupiter.api.Assertions.assertSame;
import static org.junit.jupiter.api.Assertions.assertTrue;
import static org.junit.jupiter.api.Assertions.fail;
```
If I ordered that correctly in my head :P

src/test/java/spoon/test/pkg/PackageTest.java
```diff
@@ -191,6 +186,19 @@ public void testAnnotationInPackageInfoWhenTemplatesCompiled() throws Exception
...
+	@Test
+	@GitHubIssue(issueNumber = 5357, fixed = true)
+	public void testSeparateImportForPackageInfoAnnotation() {
+		assertDoesNotThrow(() -> {
+			Launcher launcher = new Launcher();
+			launcher.addInputResource("src/test/java/spoon/test/pkg/annotate/package-info.java");
+			launcher.getEnvironment().setNoClasspath(true);
+			launcher.getEnvironment().setCopyResources(false);
+			launcher.run();
+			CtModel model = launcher.getModel();
```
I-Al-Istannen, line 198:
```suggestion
```
I think getting the model here doesn't do anything

I-Al-Istannen (changes requested):

I-Al-Istannen (approved):

monperrus:
thanks all!

PR #6186, opened 2025-02-19
chore: Update extraChecks JDK to v23 in flake.nix

rishivijayv:
See discussion in https://github.com/INRIA/spoon/pull/6158 for context. 

Based on the [discussion](https://github.com/INRIA/spoon/pull/6158#issuecomment-2667121734) it looks like only the `extraChecks` runtime needs this bump. I've just manually changed that version to 22. Are there any other chores that need to be performed associated with this bump ? 

cc @monperrus @I-Al-Istannen

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -188,7 +188,7 @@
                 jdk17 = mkShell system { javaVersion = 17; };
                 jdk21 = mkShell system { javaVersion = 21; };
                 jdk22 = mkShell system { javaVersion = 22; };
-                extraChecks = mkShell system { extraChecks = true; javaVersion = 21; };
+                extraChecks = mkShell system { extraChecks = true; javaVersion = 22; };
                 jReleaser = mkShell system { release = true; javaVersion = 21; };
               });
         in
--- a/flake.nix
+++ b/flake.nix
@@ -187,7 +187,6 @@
                 default = jdk17;
                 jdk17 = mkShell system { javaVersion = 17; };
                 jdk21 = mkShell system { javaVersion = 21; };
-                jdk22 = mkShell system { javaVersion = 22; };
                 extraChecks = mkShell system { extraChecks = true; javaVersion = 23; };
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -188,7 +188,7 @@
                 jdk17 = mkShell system { javaVersion = 17; };
                 jdk21 = mkShell system { javaVersion = 21; };
                 jdk22 = mkShell system { javaVersion = 22; };
-                extraChecks = mkShell system { extraChecks = true; javaVersion = 21; };
+                extraChecks = mkShell system { extraChecks = true; javaVersion = 22; };
```
I-Al-Istannen, line 191:
```suggestion
                extraChecks = mkShell system { extraChecks = true; javaVersion = 23; };
```

22 is EOL, I would just go to 23

I-Al-Istannen (approved):

I-Al-Istannen (changes requested):
Whoops, clicked the wrong button

flake.nix
rishivijayv: Done! A couple of lingering questions I had about JDK versions and `flake.nix` in general:

**Q1**: Should we remove the `jdk22` version from this file now, since it has reached EOL and looks like should not be used here anymore? 

**Q2**: Is there a way to test these changes to `flake.nix` locally? Or is running them in the context of Github actions the only way to verify this works?

I-Al-Istannen (changes requested):

flake.nix
I-Al-Istannen: **Q1**: I did that in a later commit in this PR

**Q2**: You can just copy-paste the command from Github Actions and it should reproduce exactly:
`nix develop --ignore-environment --keep MAVEN_OPTS --keep PATH --keep LANG .#extraChecks --command extra`

I-Al-Istannen (approved):

monperrus:
thanks  @rishivijayv and @I-Al-Istannen

PR #6187, opened 2025-02-19
test 1.8

randomicecube:
goal: cache being used (not being for some reason)

### DIFF
--- a/.github/workflows/dirty-waters.yml
+++ b/.github/workflows/dirty-waters.yml
@@ -1,0 +1,28 @@
+name: dirty-waters
+
+on:
+  pull_request:
+  push:
+    branches: [ master ]
+
+jobs:
+  dirty-waters:
+    runs-on:
+        ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


I-Al-Istannen:
I have no idea what your problem is, but I assume you have looked at [cache inheritance for GitHub actions](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache)?

randomicecube:
I have, I'm not exactly sure what the issue is/was with the other PR not using cache, because as you can see, https://github.com/randomicecube/spoon/pull/2 uses cache without issues.
From what I could see in the logs, the key used for cache retrieval was being the correct one (based on the previous commit's SHA), but it wasn't storing it with the current commit's SHA for some odd reason -- it is really odd, because in other projects it seems to be going fine. I'll update it to v1.8 in the original PR, to check if it works

PR #6188, opened 2025-02-20
review: fix: Ensure CU-level elements are included when renaming a top-level type

### LINKED ISSUE #5086: [Bug]: Imports are removed
### Describe the bug

Imports are removed when using Sniper Printer and `autoImports` set to `true`.

```Java
https://github.com/DaveJarvis/nts
```

```Java
import spoon.Launcher;
import spoon.refactoring.Refactoring;
import spoon.support.sniper.SniperJavaPrettyPrinter;

import java.io.File;
import java.util.regex.Pattern;

public class App {
  public static void main( final String[] args ) {
    final var outputDir = new File("/tmp/nts-new");
    final var spoon = new Launcher();

    final var env = spoon.getEnvironment();
    env.setPrettyPrinterCreator(
      () -> new SniperJavaPrettyPrinter( env )
    );
    env.setAutoImports( true );

    spoon.setSourceOutputDirectory( outputDir );

    spoon.addInputResource( "/tmp/nts/app/src/main/java/nts/app" );
    spoon.addInputResource( "/tmp/nts/dvi/src/main/java/nts/dvi" );
    spoon.addInputResource( "/tmp/nts/png/src/main/java/nts/png" );
    spoon.addInputResource( "/tmp/nts/svg/src/main/java/nts/svg" );
    spoon.addInputResource( "/tmp/nts/tex/src/main/java/nts/align" );
    spoon.addInputResource( "/tmp/nts/tex/src/main/java/nts/base" );
    spoon.addInputResource( "/tmp/nts/tex/src/main/java/nts/builder" );
    spoon.addInputResource( "/tmp/nts/tex/src/main/java/nts/command" );
    spoon.addInputResource( "/tmp/nts/tex/src/main/java/nts/hyph" );
    spoon.addInputResource( "/tmp/nts/tex/src/main/java/nts/io" );
    spoon.addInputResource( "/tmp/nts/tex/src/main/java/nts/math" );
    spoon.addInputResource( "/tmp/nts/tex/src/main/java/nts/noad" );
    spoon.addInputResource( "/tmp/nts/tex/src/main/java/nts/node" );
    spoon.addInputResource( "/tmp/nts/tex/src/main/java/nts/render" );
    spoon.addInputResource( "/tmp/nts/tex/src/main/java/nts/tex" );
    spoon.addInputResource( "/tmp/nts/tex/src/main/java/nts/tfm" );
    spoon.addInputResource( "/tmp/nts/tex/src/main/java/nts/typo" );
    spoon.addInputResource( "/tmp/nts/tex/src/test/java/nts/io" );

    final var model = spoon.buildModel();
    final var p = Pattern.compile( "^(?!Kt).*" );

    for( final var clazz : model.getAllTypes() ) {
      final var name = clazz.getSimpleName();
      final var m = p.matcher( name );

      if( m.matches() ) {
        final var newName = "Kt" + name;

        System.out.println( "rename " + name + " to " + newName );
        Refactoring.changeTypeName( clazz, newName );
      }
    }

    if( outputDir.exists() || outputDir.mkdirs() ) {
      spoon.prettyprint();
    }
  }
}
```

### Actual output

```Java
package nts.tfm;/**
 * |IndexMultimap| can store and retrieve |int| values associated to particular |int| key. There can
 * be more values associated to the same key. This class can be replaced by any generic associative
 * container which provides one to many mapping.
 */
public class KtIndexMultimap {

  /** The internal representation of (key, value) pair. */
  protected static class Pair {

    /** The key */
    int key;

    /** The value */
    int val;

    /**
     * Makes new |Pair| with given key and value.
     *
     * @param k the key
     * @param v the value
     */
    Pair(int k, int v) {
      key = k;
      val = v;
    }
  }

  /*
   * The (key, value) pairs are kept in |Vector| sorted acording the key.
   */

  /** Internal storage of (key, value) pairs */
  private Vector<Pair> data = new Vector<Pair>();

  /**
   * The number of (key, value) pairs kept.
   *
   * @return the number of stored pairs.
   */
  protected final int size() {
    return data.size();
  }

  /**
   * (key, value) pair at given position.
   *
   * @param i the position of pair to be examined.
   * @return the pair at given position.
   */
  protected final Pair at(int i) {
    return data.elementAt(i);
  }

  /**
   * Insert a (key, value) pair at the given position.
   *
   * @param i the pair to be inserted.
   * @param i the position to be inserted to.
   */
  protected final void insert(Pair p, int i) {
    data.insertElementAt(p, i);
  }

  /*
   * Searching is implemented by binary search algorithm. If any
   * (key, value) pair withe given key is present in the sorted sequence, it
   * returns the position of one of such pairs. If not it returns the
   * position where a new pair with given key should be inserted to keep the
   * the sequence sorted.
   */

  /**
   * Gives the position where a (key, value) pair with given key is stored or where it should be
   * stored if there is no such pair.
   *
   * @param key the key searched for.
   * @return the position.
   */
  protected final int search(int key) {
    int beg = 0;
    int end = size();
    while (beg < end) {
      int med = (beg + end) / 2;
      Pair p = at(med);
      if (key < p.key) end = med;
      else if (key > p.key) beg = med + 1;
      else return med;
    }
    return beg;
  }

  /*
   * When adding a new (key, value) pair the position of some pair with
   * given key is found first, then it is inserted after all such pairs
   * (or to founded position if there is none such pair).
   */

  /**
   * Adds a new (key, value) pair.
   *
   * @param key the key of the new pair.
   * @param val the value of the new pair.
   */
  public void add(int key, int val) {
    synchronized (data) {
      int pos = search(key);
      while (pos < size() && at(pos).key == key) pos++;
      insert(new Pair(key, val), pos);
    }
  }

  /*
   * The class |Enum| is similar to interface |Enumeration| in the sense
   * that it provides methods for geting next value in a sequnce and for
   * testing the availability of the next value. But instead of |Object|s
   * it works with |ints|.
   */

  /** Class |Enum| provides the sequence of all values associated to particular key. */
  public class Enum {

    /** the current position in the sequence of pairs */
    private int pos;

    /** the key for which the values are required */
    private final int key;

    /**
     * Makes new |Enum| for given key.
     *
     * @param k the key for which the values are required.
     */
    /*
     * The constructor is private so only the enclosing class can
     * instantiate it.
     */
    private Enum(int k) {
      synchronized (data) {
        key = k;
        pos = search(key);
        while (pos > 0 && at(pos - 1).key == key) pos--;
      }
    }

    /**
     * Tests if there is another associated value.
     *
     * @return |true| if next value is available, |false| otherwise.
     */
    public final boolean hasMore() {
      return pos < size() && at(pos).key == key;
    }

    /**
     * Gives the next value from the sequence of associated values.
     *
     * @return the next value.
     */
    public final int next() {
      return at(pos++).val;
    }
  }

  /**
   * Gives the sequence of all keys associated to the given key.
   *
   * @param key the given key.
   * @return the object representing the sequence of associated values.
   */
  public Enum forKey(int key) {
    return new Enum(key);
  }
}
```

### Expected output

```Java
// Copyright 2001 by
// DANTE e.V. and any individual authors listed elsewhere in this file.
//
// This file is part of the NTS system.
// ------------------------------------
//
// It may be distributed and/or modified under the
// conditions of the NTS Public License (NTSPL), either version 1.0
// of this license or (at your option) any later version.
// The latest version of this license is in
//    http://www.dante.de/projects/nts/ntspl.txt
// and version 1.0 or later is part of all distributions of NTS
// version 1.0-beta or later.
//
// The list of all files belonging to the NTS distribution is given in
// the file `manifest.txt'.
//
// Filename: nts/tfm/IndexMultimap.java
// $Id: IndexMultimap.java,v 1.1.1.1 1999/06/05 04:18:58 ksk Exp $
package nts.tfm;

import java.util.Vector;

/**
 * |IndexMultimap| can store and retrieve |int| values associated to particular |int| key. There can
 * be more values associated to the same key. This class can be replaced by any generic associative
 * container which provides one to many mapping.
 */
public class KtIndexMultimap {

  /** The internal representation of (key, value) pair. */
  protected static class Pair {

    /** The key */
    int key;

    /** The value */
    int val;

    /**
     * Makes new |Pair| with given key and value.
     *
     * @param k the key
     * @param v the value
     */
    Pair(int k, int v) {
      key = k;
      val = v;
    }
  }

  /*
   * The (key, value) pairs are kept in |Vector| sorted acording the key.
   */

  /** Internal storage of (key, value) pairs */
  private Vector<Pair> data = new Vector<Pair>();

  /**
   * The number of (key, value) pairs kept.
   *
   * @return the number of stored pairs.
   */
  protected final int size() {
    return data.size();
  }

  /**
   * (key, value) pair at given position.
   *
   * @param i the position of pair to be examined.
   * @return the pair at given position.
   */
  protected final Pair at(int i) {
    return data.elementAt(i);
  }

  /**
   * Insert a (key, value) pair at the given position.
   *
   * @param i the pair to be inserted.
   * @param i the position to be inserted to.
   */
  protected final void insert(Pair p, int i) {
    data.insertElementAt(p, i);
  }

  /*
   * Searching is implemented by binary search algorithm. If any
   * (key, value) pair withe given key is present in the sorted sequence, it
   * returns the position of one of such pairs. If not it returns the
   * position where a new pair with given key should be inserted to keep the
   * the sequence sorted.
   */

  /**
   * Gives the position where a (key, value) pair with given key is stored or where it should be
   * stored if there is no such pair.
   *
   * @param key the key searched for.
   * @return the position.
   */
  protected final int search(int key) {
    int beg = 0;
    int end = size();
    while (beg < end) {
      int med = (beg + end) / 2;
      Pair p = at(med);
      if (key < p.key) end = med;
      else if (key > p.key) beg = med + 1;
      else return med;
    }
    return beg;
  }

  /*
   * When adding a new (key, value) pair the position of some pair with
   * given key is found first, then it is inserted after all such pairs
   * (or to founded position if there is none such pair).
   */

  /**
   * Adds a new (key, value) pair.
   *
   * @param key the key of the new pair.
   * @param val the value of the new pair.
   */
  public void add(int key, int val) {
    synchronized (data) {
      int pos = search(key);
      while (pos < size() && at(pos).key == key) pos++;
      insert(new Pair(key, val), pos);
    }
  }

  /*
   * The class |Enum| is similar to interface |Enumeration| in the sense
   * that it provides methods for geting next value in a sequnce and for
   * testing the availability of the next value. But instead of |Object|s
   * it works with |ints|.
   */

  /** Class |Enum| provides the sequence of all values associated to particular key. */
  public class Enum {

    /** the current position in the sequence of pairs */
    private int pos;

    /** the key for which the values are required */
    private final int key;

    /**
     * Makes new |Enum| for given key.
     *
     * @param k the key for which the values are required.
     */
    /*
     * The constructor is private so only the enclosing class can
     * instantiate it.
     */
    private Enum(int k) {
      synchronized (data) {
        key = k;
        pos = search(key);
        while (pos > 0 && at(pos - 1).key == key) pos--;
      }
    }

    /**
     * Tests if there is another associated value.
     *
     * @return |true| if next value is available, |false| otherwise.
     */
    public final boolean hasMore() {
      return (pos < size() && at(pos).key == key);
    }

    /**
     * Gives the next value from the sequence of associated values.
     *
     * @return the next value.
     */
    public final int next() {
      return at(pos++).val;
    }
  }

  /**
   * Gives the sequence of all keys associated to the given key.
   *
   * @param key the given key.
   * @return the object representing the sequence of associated values.
   */
  public Enum forKey(int key) {
    return new Enum(key);
  }
}
```

### Spoon Version

Latest

### JVM Version

18

Linux

jdmcmahan:
fix #5086 

This PR modifies the `Refactoring.changeTypeName` logic to prevent loss of information from the original class.

The current logic effectively creates a new compilation unit to house the renamed type. However, the type by itself only contains the source code elements for the class. All other `SourceFragment`s outside of the class block, such as imports, comments, and whitespace, are lost in this conversion. The most apparent issue is that `SniperJavaPrettyPrinter` does not have the source context to properly print the full class file after a rename.

This can easily be illustrated with the existing `TestSniperPrinter#testClassRename1` test (which currently does not have assertions for these external elements):

Original input:

```java
package spoon.test.prettyprinter.testclasses;

import java.util.ArrayList;
import java.util.List;

/**
 * The content of this file 
 * 

 * 		should not be changed
 * Because DJPP should print only modified content again 
 */
public
@Deprecated
abstract class /* even this comment stays here together with all SPACES and EOLs*/ ToBeChanged<T, K> /*before extends*/
	extends ArrayList<T /* let's confuse > it */ > implements List<T>,
	Cloneable
{
	
	
	/**/
	final
	//
	private String string = "a"
			+ "b" + "c"+"d";
	
	//and spaces here are wanted too
	
	
	public <T, K> void andSomeOtherMethod(
			int param1,
			String param2         , List<?>[][]... twoDArrayOfLists)
	{/**/
		System.out.println("aaa"
				+ "xyz");
	/*x*/}
	List<?>[][] twoDArrayOfLists = new List<?>[7][];
}

//and what about this comment? 
```

Current result of `TestSniperPrinter#testClassRename1`:

```java
package spoon.test.prettyprinter.testclasses;/**
 * The content of this file 
 * 

 * 		should not be changed
 * Because DJPP should print only modified content again 
 */
public
@Deprecated
abstract class /* even this comment stays here together with all SPACES and EOLs*/ Bar<T, K> /*before extends*/
	extends ArrayList<T /* let's confuse > it */ > implements List<T>,
	Cloneable
{
	
	
	/**/
	final
	//
	private String string = "a"
			+ "b" + "c"+"d";
	
	//and spaces here are wanted too
	
	
	public <T, K> void andSomeOtherMethod(
			int param1,
			String param2         , List<?>[][]... twoDArrayOfLists)
	{/**/
		System.out.println("aaa"
				+ "xyz");
	/*x*/}
	List<?>[][] twoDArrayOfLists = new List<?>[7][];
}
```

The change proposed here simply reuses the original compilation unit object to ensure top-level elements (and their source context) are retained. This comes with two consequences of note:
- Since compilation units are made up of type *references* instead of the types themselves, we must do some introspection to ensure the original type declaration is properly renamed. Note that the existing `Query` for type references is not sufficient since the reference in question is a child of the compilation unit (not the package) and therefore isn't captured with the default scope.
- Since the renamed type no longer matches the file name, we need to tweak `CtCompilationUnit#getMainType` to be more lenient with the top-level type. I believe it's safe to assume that a single declared type should always be considered the main type, especially given the rarity (and discouraged practice) of having multiple top-level types in a single file. We could further refine this logic for robustness (possibly adding a check on the type visibility) but this may be getting into some remote edge cases that even the existing logic doesn't cover.

Thanks in advance for the review! Don't hesitate to @ me if there are other considerations to discuss.

### Copying elements into the new compilation unit

This method involved manually copying the imports and comments into the compilation unit created by `CompilationUnitFactory#addType` in the current logic. It's also important to set the parent of these elements to avoid consistency errors downstream.

However, this devolved into directly manipulating the `SourceFragment`s of the compilation unit in an attempt to recreate whitespace around the new elements. This proved futile as (naturally) we can only add instances of `ElementSourceFragment` as children of `CtCompilationUnit#getOriginalSourceFragment`, and wrapping whitespace as an element felt hacky at best.

Setting the file to the same source as the original compilation unit helped to regenerate the source context, but at that point we've seemingly defeated the purpose of copying the compilation unit in the first place.

### Copying the source file along with the compilation unit

Another possibility is to copy the source file along with the compilation unit. The big question here is *where* to copy the source file to on the filesystem. We cannot copy the file to `Environment#getSourceOutputDirectory` prematurely because the contents will be cleared on `JDTBasedSpoonCompiler#generateProcessedSourceFilesUsingCUs` (unless this logic is refined, perhaps by creating the file if it doesn't exist to satisfy the comments here https://github.com/INRIA/spoon/blob/028a12396a2f61fc8b18c18807033225b9972278/src/main/java/spoon/support/compiler/jdt/JDTBasedSpoonCompiler.java#L567-L569). `Environment#getBinaryOutputDirectory` might be a better candidate, but it's unclear whether a Maven `generated-sources`-type approach is appropriate for its intended use case.

A more creative solution could involve a virtual/in-memory file, although we are limited to instances of `java.io.File` which `VirtualFile` and the like does not implement. Something like Apache Commons VFS may prove useful, but this would require a new dependency as well as extensive refactoring without a provider that adheres to `java.io.File`.

### Exposing new capabilities from `CompilationUnitFactory`

If the issue is that `CompilationUnitFactory` only accepts `CtType` for building a new compilation unit, perhaps another solution would be to add methods that include CU-level context. Unfortunately, this doesn't solve many of the other challenges discussed here but does provide the benefit of modularizing the solution for use cases outside of renaming a top-level type. This also gives us access to the `cachedCompilationUnits` internals to ensure consistency if needed.

### DIFF
--- a/src/main/java/spoon/refactoring/Refactoring.java
+++ b/src/main/java/spoon/refactoring/Refactoring.java
@@ -9,6 +9,7 @@ package spoon.refactoring;
 
 import spoon.SpoonException;
 import spoon.reflect.code.CtLocalVariable;
+import spoon.reflect.declaration.CtCompilationUnit;
 import spoon.reflect.declaration.CtExecutable;
 import spoon.reflect.declaration.CtField;
 import spoon.reflect.declaration.CtMethod;
@@ -37,13 +38,6 @@ public final class Refactoring {
 	 * 		New name of the element.
 	 */
 	public static void changeTypeName(final CtType<?> type, String name) {
-
-		// first we remove the type from the list of types
-		// to be pretty-printed
-		if (type.isTopLevel()) {
-			type.getFactory().CompilationUnit().removeType(type);
-		}
-
 		final String typeQFN = type.getQualifiedName();
 		final List<CtTypeReference<?>> references = Query.getElements(type.getFactory(), new TypeFilter<CtTypeReference<?>>(CtTypeReference.class) {
 			@Override
@@ -60,7 +54,19 @@ public final class Refactoring {
 
 		// adding the new type
 		if (type.isTopLevel()) {
-			type.getFactory().CompilationUnit().addType(type);
+			CtCompilationUnit compilationUnit = type.getFactory().CompilationUnit().getOrCreate(type);
+
+			new CtScanner() {
+
+				@Override
+				public <T> void visitCtTypeReference(CtTypeReference<T> reference) {
+					if (typeQFN.equals(reference.getQualifiedName())) {
+						reference.setSimpleName(name);
+					}
+
+					super.visitCtTypeReference(reference);
+				}
+			}.scan(compilationUnit);
 		}
 
 	}
--- a/src/main/java/spoon/support/reflect/declaration/CtCompilationUnitImpl.java
+++ b/src/main/java/spoon/support/reflect/declaration/CtCompilationUnitImpl.java
@@ -141,7 +141,7 @@ public class CtCompilationUnitImpl extends CtElementImpl implements CtCompilatio
 
 	@Override
 	public CtType<?> getMainType() {
-		if (getFile() == null) {
+		if (getFile() == null || getDeclaredTypes().size() == 1) {
 			return getDeclaredTypes().get(0);
 		}
 		for (CtType<?> t : getDeclaredTypes()) {
--- a/src/test/java/spoon/test/prettyprinter/TestSniperPrinter.java
+++ b/src/test/java/spoon/test/prettyprinter/TestSniperPrinter.java
@@ -11,8 +11,6 @@ import org.apache.commons.io.FileUtils;
 import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.io.TempDir;
-import org.junit.jupiter.params.ParameterizedTest;
-import org.junit.jupiter.params.provider.MethodSource;
 import spoon.Launcher;
 import spoon.SpoonException;
 import spoon.compiler.Environment;
@@ -70,7 +68,6 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
-import java.util.stream.Stream;
 
 import static org.hamcrest.CoreMatchers.allOf;
 import static org.hamcrest.CoreMatchers.containsString;
@@ -99,7 +96,7 @@ public class TestSniperPrinter {
 		// with the necessary tweaks
 		testClassRename(tempDir, type -> {
 			type.setSimpleName("Bar");
-			type.getFactory().CompilationUnit().addType(type);
+			type.getPosition().getCompilationUnit().getDeclaredTypeReferences().get(0).setSimpleName("Bar");
 		});
 
 	}
@@ -117,15 +114,17 @@ public class TestSniperPrinter {
 		Factory f = launcher.getFactory();
 
 		final CtClass<?> type = f.Class().get(testClass);
+		String original = type.getPosition().getCompilationUnit().getOriginalSourceCode();
 
 		// performing the type rename
 		renameTransfo.accept(type);
 		//print the changed model
 		launcher.prettyprint();
 
+		String expected = original.replaceAll("\\bToBeChanged\\b", "Bar");
 
 		String contentOfPrettyPrintedClassFromDisk = getContentOfPrettyPrintedClassFromDisk(type);
-		assertTrue(contentOfPrettyPrintedClassFromDisk.contains("EOLs*/ Bar<T, K>"), contentOfPrettyPrintedClassFromDisk);
+		assertEquals(expected, contentOfPrettyPrintedClassFromDisk);
 
 	}
 


monperrus:
thanks a lot @jdmcmahan

PR #6202, opened 2025-02-25
Fix spoon-control-flow pom version

bolu61:
It seems that `spoon-control-flow` was pointing to the wrong parent pom version. Sometimes builds will break. This should fix the versioning, consistent with the other modules.


monperrus:
thanks @bolu61

Issue #4335, opened 2021-12-06
[Bug] SniperJavaPrettyPrinter - Extra parenthesis around typecast


plcarmel:
When renaming "a", in this code:
```java
package org.example;

class Example {

    void example() {
        var a = 12345.0
        var x = ((Double) a).toString();
    }

}
```
it gets transformed into this code, that does not compile:

```java
package org.example;

class Example {

    void example() {
        var b = 12345.0;
        var x = (((Double)) (b)).toString();
    }

}
```

Extra parenthesis are added around the typecast.

Code to reproduce the issue is available on GitHub:
https://github.com/plcarmel/spoon-parenthesis-bug

```
> java -version
java version "11.0.13" 2021-10-19 LTS
Java(TM) SE Runtime Environment 18.9 (build 11.0.13+10-LTS-370)
Java HotSpot(TM) 64-Bit Server VM 18.9 (build 11.0.13+10-LTS-370, mixed mode)
```

slarse:
Thanks for the bug report.

This is related to #3809 and #3911

Issue #4221, opened 2021-10-12
SniperPrettyPrinter - Extra parenthesis are added while using CtTypeAccess for changing code


abhi22493:
When using SniperPrettyPrinter, for the below input the expected output is not generated. Spoon is adding extra parenthesis while using CtTypeAccess for changing code.

Input : 
`gconn= (OracleConnection) WSJdbcUtil.getNativeConnection((WSJdbcConnection)connection);`

Expected : 
`gconn= (OracleConnection) com.wildfly.ws.rsadapter.jdbc.WSWildflyJdbcUtil.getNativeConnection((WSJdbcConnection)connection);`

Actual : 
`gconn= ((OracleConnection) (com.wildfly.ws.rsadapter.jdbc.WSWildflyJdbcUtil.getNativeConnection(((WSJdbcConnection) (connection)))));`

Spoon print config :
launcher.getEnvironment().setPrettyPrinterCreator(() -> new SniperJavaPrettyPrinter(launcher.getEnvironment()));

Spoon version : 9.1.0-beta-10

Issue #3911, opened 2021-05-05
bug(sniper): Round parenthesis of type cast aren't matched to corresponding token


slarse:
When sinper printing an expression that contains a type cast inside of round brackets (i.e. round brackets enclosing the type cast's own round brackets), the sniper printer can insert weird spacing. So, for example, something that should be printed like so:

```java
double value = ((double) (3 / 2)) / 2;
```

Instead gets printed like so:

```java
double value = ( (double) (3 / 2)) / 2;
```

The reason for this is that the token corresponding to the type cast's opening parenthesis isn't actually matched to the type cast. Instead, it's matched to the parenthesis _following_, so the one preceding the `3` here. This causes the space that actually appears after the `(double)` in this case, to be printed before it.

I'm sure this kind of mismatching can cause issues elsewhere. I'm working on a fix for this.

algomaster99:
> Instead, it's matched to the parenthesis following, so the one preceding the `3` here.

Why is `(` (just before the `d` in `double`) getting matched with another opening parenthesis (just before the `3`)? It should anyway be programmed to match with a closing parenthesis, right?

Also, does this issue affect the behaviour [here](https://github.com/SpoonLabs/diffmin/pull/51#discussion_r627698088) as well?

slarse:
> > Instead, it's matched to the parenthesis following, so the one preceding the `3` here.
> 
> Why is `(` (just before the `d` in `double`) getting matched with another opening parenthesis (just before the `3`)?

Because the cast isn't correctly split into source fragments.

> It should anyway be programmed to match with a closing parenthesis, right?

Not sure what you mean by that, the opening parenthesis should match an opening parenthesis.

> Also, does this issue affect the behaviour [here](https://github.com/SpoonLabs/diffmin/pull/51#discussion_r627698088) as well?

That's very much possible!

algomaster99:
> Not sure what you mean by that, the opening parenthesis should match an opening parenthesis.

My intuition of parsing such things is that the program would try to find pairs of `(` and `)`. This will help in checking if the brackets are in order and there aren't any extra or missing tokens. So I expected that opening parenthesis would match with a closing one just to create that pair as defined above.

slarse:
> > Not sure what you mean by that, the opening parenthesis should match an opening parenthesis.
> 
[...]

PR #6215, opened 2025-03-04
fix: Correct parenthesis around type-casting for refactored code when using SniperJavaPrettyPrinter

rishivijayv:
Fix https://github.com/INRIA/spoon/issues/4335

# Context and Fix
When a piece of code involving a type-cast is refactored and then printed using `SniperJavaPrettyPrinter` (where the `cast` itself is to be printed from the source-code) , we seem to be printing extra parenthesis around the type-cast, leading to non-compilable java code. 

I believe this was occurring because when processing a `CAST` fragment specifically, we by default seem to add extra `(...)` pairs around it [here](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/reflect/visitor/DefaultJavaPrettyPrinter.java#L311).

Then, when the `scan` between those 2 additions visits the `SniperJavaPrettyPrinter` scan [here](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/sniper/SniperJavaPrettyPrinter.java#L299), we eventually get to a call to print the fragment's source-code as is -- which involves the `(...)` that already come with a cast. I believe this is will always be the case, since explicit casting (that I believe we are interested with here) will always be enclosed within its own `(...)` for it to be valid Java source code (according to references like this: https://www.w3schools.com/java/java_type_casting.asp). 

So, in `SniperJavaPrettyPrinter`, instead of just printing the fragment's source-code as is, I've added a helper function which can be used to modify the source-code being printed to account for such "default" behaviour being done by our pretty-printing code. So, all calls to `fragment.getSourceCode()` in `SniperJavaPrettyPrinter` should have been replaced with this helper now. For now it just involves special logic to handle `CAST` type-fragments, but I can see that being expanded in the future if there are other "special cases" like this one. 

# Testing
I've recreated the example found in the original github issue (changing `var` to valid `Java` types that seem to be required according to a list in this test file: [link](https://github.com/INRIA/spoon/blob/master/src/test/java/spoon/test/prettyprinter/PrinterTest.java#L313) -- with `var` the `testPrinterTokenListener` in `PrinterTest` was failing). Now I believe there are no longer extra parenthesis around `(Double)`. 

I _think_ this fix might also resolve https://github.com/INRIA/spoon/issues/4221? I'm not entirely sure how to go about testing that though, since from my understanding `SniperJavaPrettyPrinter` really comes in to play whenever there is a change/refactoring from the original source code to the new one, and in https://github.com/INRIA/spoon/issues/4221 I can't see what the change being made is, just the `Expected` and the `Original`. The use of non-standard Java library packages (OracleConnecteion, etc) also makes it a little hard to recreate exactly. If there is a way/need to recreate that test or something that is similar, let me know how and I'll add/verify that test here!

If there are any other tests that I should add, please let me know as well! All the existing tests pass, which I think is a good sign 😄 

# Additional Notes
I was also investigating another `SniperJavaPrettyPrinter` bug related to extra spaces being added around type-casts. The source code still seems compilable (which imo makes this fix a little more important), but I think the explicit addition of `(` brackets in the `enterCtExpression` function of `DefaultJavaPrettyPrinter` might be making things difficult. Here, the fix works in the confines of that explicit addition. More info can be found on my comment here: https://github.com/INRIA/spoon/issues/3911#issuecomment-2697135485

cc @algomaster99 @monperrus @I-Al-Istannen would appreciate your thoughts on this whenever you get the time 🙏🏽 !

### DIFF
--- a/src/main/java/spoon/support/sniper/SniperJavaPrettyPrinter.java
+++ b/src/main/java/spoon/support/sniper/SniperJavaPrettyPrinter.java
@@ -351,7 +351,8 @@ public class SniperJavaPrettyPrinter extends DefaultJavaPrettyPrinter implements
 	 *
 	 * This function handles some custom logic which might be needed for special types of code-fragments.
 	 * For instance, we by default add parenthesis around the type while type-casting. The source code for the cast
-	 * has parenthesis around it as well (which would be required for the original code to be valid Java code).
+	 * has parenthesis around it as well (which would be required for the original code to be valid Java code -- the
+	 * assumption here is that the fragment supplied comes from valid/compilable Java code).
 	 * Printing that source-code without any modification would then lead to non-compilable java code. This function
 	 * can handle such special cases.
 	 * */
--- a/src/main/java/spoon/support/sniper/SniperJavaPrettyPrinter.java
+++ b/src/main/java/spoon/support/sniper/SniperJavaPrettyPrinter.java
@@ -231,7 +231,7 @@ public class SniperJavaPrettyPrinter extends DefaultJavaPrettyPrinter implements
 						//push the context of this collection
 						pushContext(listContext);
 					}
-					mutableTokenWriter.directPrint(fragment.getSourceCode());
+					mutableTokenWriter.directPrint(getSourceCodeForSniperPrinting(fragment));
 				}
 			}
 		});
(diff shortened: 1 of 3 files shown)


monperrus:
thanks a lot @rishivijayv

PR #6226, opened 2025-03-17
chore: update jdt.core to v3.41.0

rishivijayv:
See https://github.com/INRIA/spoon/pull/6220 for context and discussion

cc @monperrus @I-Al-Istannen @algomaster99 @frankreyesgarcia

### DIFF
--- a/src/test/java/spoon/reflect/visitor/DefaultJavaPrettyPrinterTest.java
+++ b/src/test/java/spoon/reflect/visitor/DefaultJavaPrettyPrinterTest.java
@@ -80,11 +80,11 @@ public class DefaultJavaPrettyPrinterTest {
             "((int) (1 + 2)) * 3",
             "(int) (int) (1 + 1)",
             "(\"1\" + \"2\").contains(\"1\")",
-            "null instanceof java.lang.String s && (s = \"1\").contains(\"1\")",
-            "null instanceof java.lang.String s && (s += \"1\").contains(\"1\")",
-            "null instanceof java.lang.String[] arr && (arr[0] = \"1\").contains(\"1\")",
-            "null instanceof java.lang.Integer i && (++i).toString().isEmpty()",
-            "null instanceof java.lang.Integer i && (i--).toString().isEmpty()",
+            "new java.lang.Object() instanceof java.lang.String s && (s = \"1\").contains(\"1\")",
+            "new java.lang.Object() instanceof java.lang.String s && (s += \"1\").contains(\"1\")",
+            "new java.lang.Object() instanceof java.lang.String[] arr && (arr[0] = \"1\").contains(\"1\")",
+            "new java.lang.Object() instanceof java.lang.Integer i && (++i).toString().isEmpty()",
+            "new java.lang.Object() instanceof java.lang.Integer i && (i--).toString().isEmpty()",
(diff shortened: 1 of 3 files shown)


src/test/java/spoon/reflect/visitor/DefaultJavaPrettyPrinterTest.java
```diff
@@ -80,11 +80,11 @@ public class DefaultJavaPrettyPrinterTest {
...
-            "null instanceof java.lang.String s && (s = \"1\").contains(\"1\")",
-            "null instanceof java.lang.String s && (s += \"1\").contains(\"1\")",
-            "null instanceof java.lang.String[] arr && (arr[0] = \"1\").contains(\"1\")",
-            "null instanceof java.lang.Integer i && (++i).toString().isEmpty()",
-            "null instanceof java.lang.Integer i && (i--).toString().isEmpty()",
+            "new java.lang.Object() instanceof java.lang.String s && (s = \"1\").contains(\"1\")",
+            "new java.lang.Object() instanceof java.lang.String s && (s += \"1\").contains(\"1\")",
+            "new java.lang.Object() instanceof java.lang.String[] arr && (arr[0] = \"1\").contains(\"1\")",
+            "new java.lang.Object() instanceof java.lang.Integer i && (++i).toString().isEmpty()",
+            "new java.lang.Object() instanceof java.lang.Integer i && (i--).toString().isEmpty()",
```
SirYwell, line 87: What's the intention of this change?

rishivijayv: I've described the issue and my reasoning here in more detail: https://github.com/INRIA/spoon/pull/6220#issuecomment-2731055462 but as a very quick summary, without this, I believe we were getting a Java error being surfaced by the new JDT Core Compiler, which complained: `Expression type cannot be a subtype of the Pattern type`. I believe here it was referencing `null` being a subtype of the all the types following `instanceof` here. 

Using `new Object()`, we no longer get that complaint

monperrus:
Thanks a lot @rishivijayv

src/test/java/spoon/reflect/visitor/DefaultJavaPrettyPrinterTest.java
rishivijayv: @monperrus @SirYwell Ooh I was about to push a commit that sets the compliance level to `21`, after which `null instanceof ...` works for this test case 😅 

Is that something important enough to be done in a separate PR, or is it fine to leave this change in as is?

SirYwell: It's fine probably. The usage of `null` was mainly about brevity.

SirYwell: After thinking about it for a while, I think we should still set the compliance level to at least 17, otherwise we just rely on JDT being tolerant with us. But that just might break things again on the next update.

rishivijayv: Sure thing @SirYwell ! I'll put up a PR for this soon! 🫡

PR #6237, opened 2025-03-22
fix: Prevent printing spaces before a type-cast (SniperPrinter)

rishivijayv:
fix https://github.com/INRIA/spoon/issues/3911

I've highlighted what I believe the issue was in this comment made on the issue: https://github.com/INRIA/spoon/issues/3911#issuecomment-2697135485

The fix here prevents printing spaces before the typecast (ie, something like `...(double)`) if there are no spaces in the original code. If there is a space in the original code, then that should be retained (I've added a new test case for that). Lmk if there are any additional test cases that need to be added too! 

Now, instead of _always_ printing spaces, we *do not* print spaces if *all* of the following conditions are met: 
1. The fragment to be printed after the last printed fragment (which I assume is indicated by `prevIndex` below) is a type-cast. 
2. The current fragment about to be printed (which I assume is indicated by `index` below) is a token for `(`
3. There has been a "jump" between the last index that was printed and the next index about to be printed 
I believe (1) and (2) should automatically imply (3), but I've still kept that here for explicitness/good measure. 

So we basically guard the call to `printSpaces` and only execute it when (1) - (3) *do not* hold. 

There _might_ be other calls to `printSpaces`, such as [here](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/sniper/internal/AbstractSourceFragmentPrinter.java#L94) in the file, or [here](https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/support/sniper/internal/SourceFragmentContextNormal.java#L58) in a child class. The failing test case passes with just this change so I didn't add this guard anywhere else, since I wanted to keep the changes to pretty-printer limited unless really needed 🤷🏽 😅 

The code to check whether a fragment is that of a type-cast is _sort of_ extrapolated from the structure of the type-cast containing fragment I saw in the test case -- it being contained inside a singleton `CollectionSourceFragment`. I've still added an explicit case for `ElementSourceFragment` in case type-casts might be present inside simple `ElementSourceFragments` as well.

@monperrus @I-Al-Istannen @algomaster99 would love your thoughts on this whenever you get a chance! 🫡

### DIFF
--- a/src/main/java/spoon/support/sniper/internal/AbstractSourceFragmentPrinter.java
+++ b/src/main/java/spoon/support/sniper/internal/AbstractSourceFragmentPrinter.java
@@ -16,6 +16,7 @@ import spoon.reflect.code.CtComment;
 import spoon.reflect.cu.SourcePositionHolder;
 import spoon.reflect.declaration.CtModifiable;
 import spoon.reflect.path.CtRole;
+import spoon.reflect.reference.CtTypeReference;
 
 import static spoon.support.sniper.internal.ElementSourceFragment.findIndexOfNextFragment;
 import static spoon.support.sniper.internal.ElementSourceFragment.filter;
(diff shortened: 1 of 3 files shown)


algomaster99:
Thanks for the PR @rishivijayv ! It does fix a long standing bug :) The fix seems hacky but I don't have a better solution to offer. However, while ago, [the lambda printing was also broken](https://g.co/kgs/kx7xfNW). It could be because of the same issue. Could you also add this as test case?

rishivijayv:
> The fix seems hacky but I don't have a better solution to offer. However, while ago, [the lambda printing was also broken](https://g.co/kgs/kx7xfNW)

Sure thing! @algomaster99 I think the link for the lambda bug is pointing me to something else at the moment, however 😅 Could you re-link the issue and I'll have a look?

algomaster99:
Oops sorry to clickbait you. Here is the correct one - https://github.com/ASSERT-KTH/diffmin/pull/51#discussion_r627698088.

monperrus:
thanks a lot @rishivijayv

rishivijayv:
@algomaster99 sorry I was not able to test the lambda example you mentioned before this was merged 😅 I can give it a test soon though! Just to make sure I understand, in the linked comment, the original **file** (`NEW_Lambda.java`) is the _input_ to the pretty-printer, and the code in the comment is the _output_ produced by the pretty-printer? With the issue being the weird space before `x`? 

Just basing it off the code in this PR though, I _think_ that issue might not be fixed by this 🤔 Since here to prevent printing spaces I use the `isTypeCastFragment` to determine if we're processing a type-cast. So either we might be adding an explicit `(` _somewhere_ while processing a lambda's parameters (in which case, maybe a function which consolidates `isTypeCastFragment` and something like `isLambdaParams` can be used to check if we need to print a space or not), _or_ that might be caused by something else. 

I can give it a try though soon and let you know what I see! 🫡

algomaster99:
> the original file (NEW_Lambda.java) is the input to the pretty-printer, and the code in the comment is the output produced by the pretty-printer? With the issue being the weird space before x

The input was 
https://github.com/ASSERT-KTH/diffmin/pull/51/files#diff-d2d05a10155f91956fc21857efe7ee2692a364420104c92636d249aff64148aa

```java
// PREV_Lambda.java
import java.util.function.Function;

public class Lambda {
        Function<String, String> lambda = (y) -> { return y + " "; };
    }
}
```
and output (after some transformations) was https://github.com/ASSERT-KTH/diffmin/pull/51/files#diff-0690904eb2c6bb30b31747ffb533825c8eb6d202aa2a3850b24556dcec090c4e
```java
// NEW_Lambda.java
import java.util.function.Function;
public class Lambda {
        BiFunction<String, String, String> lambda = ( x, y) -> {
            return y + " ";
        };
    }
}
```

This project basically diff-ed the two files and then tried to patch the previous file using the diff and then compare it to the new one. I would have noticed a different in string values because of the bracket spacing.

Your test could use the same input file `PREV_Lambda.java` and then apply transformations and assert the string. The transformations would be:
1. Rename Function to BiFunction
2. Add type parameter `String`.
3. Add parameter to lambda.

And then you assert the string generated after transformations.

> I can give it a try though soon and let you know what I see!

:+1:

rishivijayv:
@algomaster99 wasn't able to test this unfortunately but I've opened a new issue to check and fix this lambda spacing issue if it still exists. I hope that's more helpful for future contributors so this does not get lost again in a comment of a closed PR 😅

PR #6241, opened 2025-03-25
chore: Set Launcher's compliance level to 17 for some default-pretty-printer tests

rishivijayv:
See discussion here: https://github.com/INRIA/spoon/pull/6226#discussion_r2009030723

I believe this should set the compliance level of Launcher to 17 **always** if not provided, based on the `JSAPResult` documentation [here](https://www.martiansoftware.com/jsap/doc/javadoc/com/martiansoftware/jsap/JSAPResult#getInt(java.lang.String,%20int)).

cc @SirYwell

### DIFF
--- a/src/main/java/spoon/Launcher.java
+++ b/src/main/java/spoon/Launcher.java
@@ -450,7 +450,7 @@ public class Launcher implements SpoonAPI {
 
 		Environment environment = factory.getEnvironment();
 		// environment initialization
-		environment.setComplianceLevel(jsapActualArgs.getInt("compliance", 17));
+		environment.setComplianceLevel(jsapActualArgs.getInt("compliance"));
 		environment.setLevel(jsapActualArgs.getString("level"));
 		if (jsapActualArgs.getBoolean("imports")) {
 			environment.setPrettyPrintingMode(Environment.PRETTY_PRINTING_MODE.AUTOIMPORT);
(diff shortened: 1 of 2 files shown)


SirYwell:
This is a bit more invasive than I envisioned. I think it is enough to change the test, i.e., setting the compliance level of the environment of the launcher here: https://github.com/INRIA/spoon/blob/6dd1612ed11d994236bc4e8a20fed2831e03daa0/src/test/java/spoon/reflect/visitor/DefaultJavaPrettyPrinterTest.java#L123

rishivijayv:
> I think it is enough to change the test

@SirYwell  Ahh gotcha! There are other `Launchers` created in that test file that don't use that utility function. I'm guessing we just want to change it here for this utility function and not for _every_ test in the file?

I-Al-Istannen:
There are about a dozen different ways to construct a launcher in tests currently, all doing something slightly different. One of these days we should consolidate them. I think using the @ModelTest annotation would be a good start in that direction, where applicable. Not here though.

In general it is a bit hard to tell currently if a test uses a given compliance level because it didn't care, because the default was too low or because the bug is version specific. I don't have a nice solution for that (apart from encoding that in ModelTest or whatever else we want to standardize on).

So the answer to your question is probably that nobody knows what compliance level is actually *needed* for each test, so a minimal change is likely the path of least resistance.

rishivijayv:
@I-Al-Istannen got it! The tests in that file that do use this utility function instead of creating Launchers in other ways seem to pass here, so I believe this should be good 🤔 ? But yeah a "global" way of setting up Launchers for the test would be a lot better 😅

SirYwell (approved):
Thanks, looks good.

Issue #5579, opened 2023-12-13
[Bug]: A null pointer exception occurred in build model


honghao12:
### Describe the bug

A null pointer exception occurred in build model

### Source code you are trying to analyze/transform

```Java
.
```

### Source code for your Spoon processing

```Java
Launcher launcher = new Launcher();
        launcher.getEnvironment().setComplianceLevel(11);
        launcher.getEnvironment().setAutoImports(true);
        launcher.getEnvironment().setIgnoreDuplicateDeclarations(true);
        launcher.addInputResource(filteringFolder);
        launcher.run();
```

### Actual output

```Java
null
Exception in thread "main" java.lang.reflect.InvocationTargetException
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.springframework.boot.loader.MainMethodRunner.run(MainMethodRunner.java:49)
	at org.springframework.boot.loader.Launcher.launch(Launcher.java:107)
	at org.springframework.boot.loader.Launcher.launch(Launcher.java:58)
	at org.springframework.boot.loader.JarLauncher.main(JarLauncher.java:88)
Caused by: java.lang.NullPointerException
	at spoon.support.compiler.jdt.ReferenceBuilder.isIncorrectlyBoundExecutableInNoClasspath(ReferenceBuilder.java:527)
	at spoon.support.compiler.jdt.ReferenceBuilder.getExecutableReference(ReferenceBuilder.java:495)
	at spoon.support.compiler.jdt.JDTTreeBuilder.visit(JDTTreeBuilder.java:880)
	at org.eclipse.jdt.internal.compiler.ast.AllocationExpression.traverse(AllocationExpression.java:716)
	at org.eclipse.jdt.internal.compiler.ast.Assignment.traverse(Assignment.java:274)
	at org.eclipse.jdt.internal.compiler.ast.MethodDeclaration.traverse(MethodDeclaration.java:437)
	at org.eclipse.jdt.internal.compiler.ast.TypeDeclaration.traverse(TypeDeclaration.java:1702)
[...]

PR #6268, opened 2025-04-18
fix(noclasspath): Add null check for executable type in ReferenceBuilder

ganeshs:
Prevent NullPointerException in isIncorrectlyBoundExecutableInNoClasspath when ref.getType() returns null, which can occur in no-classpath mode when analyzing complex enum initializers involving lambdas.

Fixes INRIA/spoon#5579


monperrus:
thanks @ganeshs

Issue #5001, opened 2022-11-07
Getting extra round parenthesis on some random places in variable assignments


Vardhamansankar:
Hello @I-Al-Istannen @MartinWitt @algomaster99 

I am trying to get the variable details of a String variable used in the java file as same as below,

```
String sql = "Select distinct t.NETWORK_IP, t.NETWORK_IP1, t.NETWORK_IP2, t.NETWORK_IP3, t.NETWORK_IP4 " +

1. 			 "from (SELECT DISTINCT t1.ipv4digit1 || '.' || t1.ipv4digit2 || '.' || t1.ipv4digit3 "+

                 " || '.0' network_ip, "+
				 " TO_NUMBER (t1.ipv4digit1) network_ip1, "+
				 " TO_NUMBER (t1.ipv4digit2) network_ip2, "+
				 " TO_NUMBER (t1.ipv4digit3) network_ip3, "+
				 " TO_NUMBER ('0') network_ip4, t1.t2_team_id, "+
				 " t1.system_owner_id, t1.system_owner_team_id "+
				 " FROM ip_info t1 "+

2. 				 " where t1.binary_ip >= '"+binaryIpStart+"' " +

				 " and t1.binary_ip <= '"+binaryIpEnd+"' " +
				 " ORDER BY network_ip1, network_ip2, network_ip3  "+
				 " ) t order by t.NETWORK_IP1,t.NETWORK_IP2,t.NETWORK_IP3,t.NETWORK_IP4 ";;

```

And when I used the  spoon code to get the variable 'sql' value, I am getting output as 
```

String sql = "Select distinct t.NETWORK_IP, t.NETWORK_IP1, t.NETWORK_IP2, t.NETWORK_IP3, t.NETWORK_IP4 " +

3. 				 "(from (SELECT DISTINCT t1.ipv4digit1 || '.' || t1.ipv4digit2 || '.' || t1.ipv4digit3 "+

                 " || '.0' network_ip, "+
				 " TO_NUMBER (t1.ipv4digit1) network_ip1, "+
				 " TO_NUMBER (t1.ipv4digit2) network_ip2, "+
				 " TO_NUMBER (t1.ipv4digit3) network_ip3, "+
				 " TO_NUMBER ('0') network_ip4, t1.t2_team_id, "+
				 " t1.system_owner_id, t1.system_owner_team_id "+
				 " FROM ip_info t1 "+

4. 				 " where t1.binary_ip >= '")+binaryIpStart+"' " +

[...]

PR #6285, opened 2025-05-03
test: adding some tests (pretty-printing and visitor) sql query in string literal issue

rishivijayv:
See https://github.com/INRIA/spoon/issues/5001

I'm adding some tests, all but one pass. Some of the sanity tests (which can be found in the `EvalTest.java` file) were the ones I adapted from @I-Al-Istannen 's comments on the issue,  [here](https://github.com/INRIA/spoon/issues/5001#issuecomment-1305598985) for example.

I also thought of adding some pretty-printing related tests, since I thought that might be another thing that @Vardhamansankar might be trying to do in the issue.

For `SniperPrettyPrinter`, they seem to pass. For `DefaultJavaPrettyPrinter` however, when the parenthesis optimization is _not_ used, they fail. I've made a _much_ more detailed comment [here](https://github.com/INRIA/spoon/issues/5001#issuecomment-2804967607).

So I've added the `fixed = false` flag with the `DefaultJavaPrettyPrinter` test. But I'm unsure if that test is needed and am open to removing the failing test if we want to. If we can tweak the pretty-printer's settings to _not_ print extra parenthesis (which is done with the other test for `DefaultJavaPrettyPrinter`), then I think the test passes -- and I'm not quite sure why we'd be in a scenario where we _need_ extra parenthesis 🤔 

But yeah, more info in the comment(s) I've linked. Let me know!

cc @algomaster99 @I-Al-Istannen @monperrus


rishivijayv:
Hmm looks like this is the issue with ExtraChecks above
Screenshot 2025-05-03 at 6 50 41 PM

Also present in these 2 renovate-bot PRs: [PR-1](https://github.com/INRIA/spoon/pull/6283), [PR-2](https://github.com/INRIA/spoon/pull/6284).

I'll try and look in to this soon!

I-Al-Istannen:
The link checker gets a 429, it should chill out a bit when receiving it: Too many requests

monperrus:
Let's remove `fixed = false`, it's dangerous. We can move them in a separate PR, to come with a fix.

Spun off the adorable cats in https://github.com/INRIA/spoon/issues/6288

rishivijayv:
> Let's remove `fixed = false`, it's dangerous.

@monperrus done! I've created a separate issue for that specific case here: https://github.com/INRIA/spoon/issues/6289 with a link back to the original issue https://github.com/INRIA/spoon/issues/5001. 
I'm not sure if that is an actual issue though -- since there is a workaround for that at the moment (using `DefaultJavaPrettyPrinter` with optimized parenthesis). So we might just close that new issue in the future. But regardless, just so the i's are dotted :) 

Given it wasn't explicitly stated in https://github.com/INRIA/spoon/issues/5001 that the OP was trying to pretty-print the entire file, would the merging of this PR with the current set of tests be enough to close that specific issue? 🤔

monperrus:
thanks @rishivijayv

PR #6286, opened 2025-05-03
chore: Sleep a bit when encountering a 429 in doc link checker

I-Al-Istannen:
wip


I-Al-Istannen:
Hm, secondary ratelimit and waiting minutes sounds a bit weird. https://github.com/orgs/community/discussions/158213 looks related :/

monperrus:
undraft?

I-Al-Istannen:
Sorry, I am drowning in work at the moment. GitHub reduced the (un)authenticated rate limits. We probably need to use an access token for these requests if we want to finish in a reasonable timespan. Otherwise, or in any case, we should parse the ratelimit header to find out when our quota resets.

monperrus:
moving on! thanks @I-Al-Istannen.

PR #6326, opened 2025-06-17
chore: merge similar switch cases

commandini:


src/test/resources/toStringBugTest/TestClass2.java
```diff
@@ -2108,37 +2108,7 @@ public Page<Redemption> searchRedemptionsForPay(Long rdmMerchantNo, Long userNo)
...
-            case CatalogueRedemptionStatus.CAT_RDM_REDEEM_FAILED:
-
-                apiErrorCode = APIErrorCode.ERR_CATALOGUE_REDEMPTION_FAILED;
-
-
-
-        }
-
-        return apiErrorCode;
-
```
commandini, line 2141: this method always returns `APIErrorCode.ERR_CATALOGUE_REDEMPTION_FAILED` no matter the given status code is. After this change, we can remove `int status` parameter too, but maybe some reviewer will notice what `getRedemptionErrorCode` should actually be doing (maybe with a proper switch block?).

src/test/resources/toStringBugTest/TestClass2.java
```diff
@@ -2108,37 +2108,7 @@ public Page<Redemption> searchRedemptionsForPay(Long rdmMerchantNo, Long userNo)
     }
 
     private APIErrorCode getRedemptionErrorCode(int status) {
-
```
SirYwell, line 2111: Please don't modify code that is used as test input. There is little benefit as it isn't actually run, but adds significant overhead to the review process as we need to verify that the involved test still does what it is supposed to do.

The other change looks clean

PR #6341, opened 2025-07-01
chore: port CD to new OSSRH pipeline

algomaster99:
Fixes #6313 

I follwed the documentation [here](https://jreleaser.org/guide/latest/examples/maven/maven-central.html) and it seems `jreleaser.json` has only some URL changes.

Now I don't know what type of authentication is set for `fr.inria.gforge.spoon` namespace. Based on that we need to set up the following env variables.
```
JRELEASER_MAVENCENTRAL_SONATYPE_USERNAME or JRELEASER_MAVENCENTRAL_USERNAME
JRELEASER_MAVENCENTRAL_SONATYPE_PASSWORD or JRELEASER_MAVENCENTRAL_PASSWORD
OR
JRELEASER_MAVENCENTRAL_SONATYPE_USERNAME or JRELEASER_MAVENCENTRAL_USERNAME
JRELEASER_MAVENCENTRAL_SONATYPE_TOKEN or JRELEASER_MAVENCENTRAL_TOKEN
```

and remove `NEXUS2` related env variables. Finally, we can keep GPG related env variables.

PR #6353, opened 2025-07-09
chore: port CD to new OSSRH pipeline (#6341)

algomaster99:
The `active` status for snapshot release should be `SNAPSHOT` as documented [here](https://jreleaser.org/guide/latest/examples/maven/maven-central.html).

@I-Al-Istannen please merge when tests pass.


algomaster99:
The tests are not being run :confused:

PR #6357, opened 2025-07-09
revert: remove manual trigger for nightly releases

algomaster99:
SNAPSHOT release is working so I am reverting the manual trigger.

We have tested the SNAPSHOT release, but not beta and others. There [does not seem to be an option to only stage beta releases](https://jreleaser.org/guide/latest/reference/deploy/maven/maven-central.html) so let's just wing it and see how it turns out to be on Sunday?

### DIFF
--- a/.github/workflows/release-nightly.yml
+++ b/.github/workflows/release-nightly.yml
@@ -1,7 +1,6 @@
 name: Release nightly version
 
 on:
-    workflow_dispatch:
     schedule:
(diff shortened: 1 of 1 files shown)


algomaster99:
@I-Al-Istannen please merge this PR and then we can merge into main

algomaster99:
Oops sorry. Need some sleep. I will fix it

I-Al-Istannen:
It's not harmful to keep it, I think? If so I would just leave it in?

algomaster99:
Indeed it is not. We can leave it.

PR #6358, opened 2025-07-09
chore: port CD to new OSSRH pipeline (#6341)

I-Al-Istannen:
This PR is the result of #6341 and updates to the new OSSRH pipeline.


algomaster99 (approved):
We have tested for SNAPSHOT releases, but not for beta and other releases. The reason for latter is that there is no [convenient option such as `releaseRepsository` on Maven central deployer](https://jreleaser.org/guide/latest/reference/deploy/maven/maven-central.html) to only stage the release and not deploy it directly. However, this option is available on the nexus deployer, yet it [does not work](https://github.com/jreleaser/jreleaser/issues/1925).

I say we merge this PR and then hope that the beta release is deployed on Sunday morning :)

I-Al-Istannen:
Thank you, @algomaster99 🐞

algomaster99:
Hoping for a peaceful Sunday :)

PR #6364, opened 2025-07-14
chore: allow for manual beta releases

algomaster99:
In the last automatic action for beta release, the workflow failed because renovate merged in between and now that workflow is unable to push more commits. We want to enable manual trigger for beta release.


algomaster99:
Already in #6363

PR #6425, opened 2025-09-05
enter AI coding age

monperrus:
add `.github/copilot-instructions.md`


I-Al-Istannen:
Has this demonstrated any net benefit for you? Also, an AGENTS.md might be the better format?

SirYwell:
One downside of additional files is that the information might become outdated.
In fact, it already is outdated, as Spoon currently targets Java 17 and JUnit 6.

monperrus:
not need to be specific to copilot.

PR #6444, opened 2025-09-16
WIP Resolve references to variables declared in type patterns

### LINKED ISSUE #6318: [Bug]: LocalVariableReference does not link to declaration within Type Pattern
### Describe the bug

I recently noticed that when using `Refactoring.changeLocalVariableName(...)` on a `LocalVariable` declared in a Type Pattern, references will not be renamed.
I suspect this is because the `LocalVariableReference`s are not linked to the declaration.

Image

```Java
public class SomeOtherClass extends SomeClass{ }

public class SomeClass {
    public void someMethod(SomeClass someParameter) {
        SomeOtherClass b;
        if (someParameter instanceof SomeOtherClass i) {
            b = i;
        }
    }
}
```

```Java

```

### Actual output

```Java
`getDeclaration()` returns `null`
```

### Expected output

```Java
The `CtLocalVariableReference` should link to the CtLocalVariable declaration under the TypePattern
```

### Spoon Version

a9bf1eb6bbecf579071bce82edfe7f7a41b44833 

### JVM Version

openjdk 24.0.1 2025-04-15

Arch Linux

MaxAtoms:
This adds support for variable references to local variables declared in type patterns (`spoon.reflect.reference.CtLocalVariableReference#getDeclaration`).

Closes #6318

### DIFF
--- a/src/main/java/spoon/reflect/visitor/filter/PotentialVariableDeclarationFunction.java
+++ b/src/main/java/spoon/reflect/visitor/filter/PotentialVariableDeclarationFunction.java
@@ -13,11 +13,15 @@ import spoon.reflect.code.CtBodyHolder;
 import spoon.reflect.code.CtCase;
 import spoon.reflect.code.CtCatch;
 import spoon.reflect.code.CtCatchVariable;
+import spoon.reflect.code.CtFor;
 import spoon.reflect.code.CtIf;
+import spoon.reflect.code.CtExpression;
 import spoon.reflect.code.CtLocalVariable;
 import spoon.reflect.code.CtStatement;
 import spoon.reflect.code.CtStatementList;
 import spoon.reflect.code.CtSwitch;
+import spoon.reflect.code.CtTypePattern;
+import spoon.reflect.code.CtWhile;
 import spoon.reflect.declaration.CtElement;
 import spoon.reflect.declaration.CtExecutable;
 import spoon.reflect.declaration.CtField;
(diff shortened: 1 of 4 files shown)


MaxAtoms:
@monperrus What do you think of this PR thus far?
It all feels rather improvised to me.
Note that there are some failing tests showing the peculiarities of flow scope.

monperrus:
I like the concept. You may document better the added code, esp in a big switch-style if/then/else. 

If the existing tests encode bad behavior, we should change them, with good comments explaining  the new behavior.

SirYwell:
The approach seems good enough, although we might want to rewrite the whole logic at some point, also to not longer depend on the `CtQuery` stuff there.

How are type patterns in record patterns handled?

MaxAtoms:
@SirYwell I'd be interested to hear your thoughts on how to better structure this logic.
It feels overly complicated for finding block scoped variables.
And for flow scope, the backwards search approach is kind of awkward.

SirYwell:
> @SirYwell I'd be interested to hear your thoughts on how to better structure this logic. It feels overly complicated for finding block scoped variables. And for flow scope, the backwards search approach is kind of awkward.

I *assume* that a "naive" implementation that doesn't try to re-use existing functionality (i.e., the CtQuery stuff here) would a) be easier to maintain/understand (if written in a clean way), and b) perform better, as we noticed CtQuery to have some overhead in other cases before.

It might make sense to only do a rewrite when we target Java 21, as pattern matching for switch sounds like a really good fit here. Making it more explicit which action is taken for which code element (instead of the abstract "search siblings before" for example) would make the semantics clearer, and for subsequent changes (e.g., when more patterns are added to the language) it would also be clearer which changes are needed there.

MaxAtoms:
> How are type patterns in record patterns handled?

They should work because this approach does a coarse search for any (nested) type pattern within an if condition, while/for loop expression or the left-hand side of a binary operator.
Thanks for the hint, though! I added a test case for record patterns: `spoon.test.reference.InstanceOfReferenceTest#testRecordPatterns`

MaxAtoms:
All tests are now green. Although I'm not too happy with the code structure, I would leave it like this for now. Let me know if anyone has ideas for improvements or potential edge cases I may have forgotten ✌🏽

monperrus:
Thanks @MaxAtoms LGTM. @SirYwell if you're OK with it, we can merge!

SirYwell (approved):
Thanks!

monperrus:
thanks @MaxAtoms

PR #6457, opened 2025-09-23
review: add renaming of annotation methods

MaxAtoms:
Before this change, if an annotation method was renamed, references in annotations would not be updated.


monperrus (approved):
LGTM, will merge.

monperrus:
thanks a lot @MaxAtoms

PR #6458, opened 2025-09-23
review: fix renaming of generics

MaxAtoms:
Previously, when two methods used the same identifier for a generic, the references to that generic were changed in both methods (i.e. the scoping was disregarded). See example below.

Original code before erroneous renaming:

```java
package spoon.test.refactoring.testclasses;

public class MethodGenericRenaming {
	public <T> T[] sort(T[] array) {
		return sort(array, false);
	}

	public <T> T[] sort(T[] array, boolean order) {
		return array;
	}
}
```

After renaming the `T` in the first sort method to `TNew`:

```java
public class MethodGenericRenaming {
    public <TNew> TNew[] sort(TNew[] array) {
        return sort(array, false);
    }

    public <T> TNew[] sort(TNew[] array, boolean order) {
        return array;
    }
}
```

### DIFF
--- a/src/test/java/spoon/test/refactoring/RefactoringTest.java
+++ b/src/test/java/spoon/test/refactoring/RefactoringTest.java
@@ -40,10 +40,14 @@ import spoon.reflect.CtModel;
 import spoon.reflect.code.BinaryOperatorKind;
 import spoon.reflect.code.CtBinaryOperator;
 import spoon.reflect.code.CtInvocation;
+import spoon.reflect.declaration.CtAnnotation;
+import spoon.reflect.declaration.CtAnnotationMethod;
 import spoon.reflect.declaration.CtClass;
 import spoon.reflect.declaration.CtMethod;
 import spoon.reflect.declaration.CtType;
 import spoon.reflect.factory.Factory;
+import spoon.reflect.path.CtRole;
+import spoon.reflect.reference.CtExecutableReference;
 import spoon.reflect.reference.CtTypeReference;
 import spoon.reflect.visitor.Query;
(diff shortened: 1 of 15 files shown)


src/test/java/spoon/test/refactoring/RefactoringTest.java
```diff
@@ -212,4 +217,66 @@ void renameClassInVirtualFile() {
...
+		}
+	}
+
+	@Test
+	void testRenameGenerics() throws Exception {
+		Factory factory = ModelUtils.build(GenericRenaming.class);
+		CtClass<?> clazz = factory.Class().get(GenericRenaming.class);
+		List<CtType<?>> types = clazz.getElements(new TypeFilter<>(CtType.class));
+
+		renameAndCheckClassGeneric(clazz, types, "SomeIdentifier", "TNew1", 15);
```
monperrus, line 250: how do we get 15 references, I only see 4 in GenericRenaming?

monperrus (commented):
Thanks for the PR @MaxAtoms LGTM, see my question about tests.

src/test/java/spoon/test/refactoring/RefactoringTest.java
MaxAtoms: Each access to the field `content` or the class (via `this`) seems to have a reference to the type parameter.

Here's one example:

image

monperrus:
conflicts here

monperrus:
thanks @MaxAtoms

MaxAtoms:
You're welcome 😊 I'll try to take another look at #6444 and get it ready for review as soon as possible, but it'll probably be a couple more days

PR #6467, opened 2025-09-28
Fix single-statement body replacement in visitCtMethod

mywar5:
Problem Description:
When using the CtElement#replace method to replace the sole statement (CtStatement) within a method body (CtBody), a ClassCastException is triggered. This issue is particularly pronounced in noClasspath mode, severely compromising Spoon's reliability during code refactoring and transformation scenarios.

Steps to Reproduce:
Construction Scenario: Create a method whose body contains only one statement (e.g., a method call).
Acquire Target: Retrieve this CtStatement from the method body.
Execute Replacement: Call the replace method on this CtStatement to attempt replacing it with another CtStatement.
Trigger Exception: At this point, Spoon internally throws a ClassCastException.
Minimal Reproducible Code Example:

// Assume the following method exists
// public void myMethod() {
//     someLegacyCall();
// }

CtMethod<?> method = ... // Obtain the CtMethod instance for myMethod
CtStatement singleStatement = method.getBody().getStatement(0); // Obtain “someLegacyCall();”

// Attempt replacement
singleStatement.replace(factory.createCodeSnippetStatement(“aNewCall()”));

// This will throw a ClassCastException

Root Cause Analysis:
Through deep debugging of Spoon's source code, we identified the root cause in the visitCtMethod of spoon.support.visitor.replace.ReplacementVisitor.

When the replacement target is the sole CtStatement within the body of CtMethod, the parent node of the original element is CtBlock, and the parent node of CtBlock is CtMethod.

In the ReplacementVisitor's replacement logic, it traverses the AST upward, ultimately invoking visitCtMethod. Within this method, the logic for handling the method body is:

replaceElementIfExist(m.getBody(), new spoon.support.visitor.replace.ReplacementVisitor.CtExecutableBodyReplaceListener(m));

Here, CtExecutableBodyReplaceListener expects to receive a replacement element of type CtBlock. However, when our replace call originates from CtStatement, the replacement element we pass is also a CtStatement.

Therefore, inside the replaceElementIfExist method, attempting to assign a CtStatement to a listener expecting a CtBlock triggers a ClassCastException. Essentially, this occurs because the ReplacementVisitor fails to properly handle the edge case where “a CtStatement replaces another CtStatement that happens to be the sole child node of a CtBlock.”

Solution:
We added a pre-condition check in the visitCtMethod of ReplacementVisitor.java specifically to handle this edge case during method body processing.

Core Fix Logic:

Before the original replaceElementIfExist(m.getBody(), ...) call, we inserted the following code block:

// FIX for single-statement body replacement
if (m.getBody() == original && original instanceof spoon.reflect.code.CtStatement) {
    if (replace.length > 1) {
        throw new spoon.support.visitor.replace.InvalidReplaceException(“Cannot replace a single statement body with multiple elements.”);
    }
    spoon.reflect.code.CtBlock<?> newBlock = m.getFactory().createBlock();
    if (replace.length == 1) {
        if (!(replace[0] instanceof spoon.reflect.code.CtStatement)) {
            throw new spoon.support.visitor.replace.InvalidReplaceException(“A single statement body can only be replaced by another statement.”);
        }
        newBlock.addStatement((spoon.reflect.code.CtStatement) replace[0]);
    }
    m.setBody(newBlock);
} else {
    replaceElementIfExist(m.getBody(), new spoon.support.visitor.replace.ReplacementVisitor.CtExecutableBodyReplaceListener(m));
}

Logic Explanation:

Conditional Check: if (m.getBody() == original && original instanceof spoon.reflect.code.CtStatement)
This condition precisely captures the edge case we face: the replacement target (original) is a CtStatement, and its parent node (indirectly determined via getParent()) is the body of a CtMethod.
Create new block: spoon.reflect.code.CtBlock<?> newBlock = m.getFactory().createBlock();
We create a new, empty CtBlock.
Populate new block: newBlock.addStatement((spoon.reflect.code.CtStatement) replace[0]);
We add the replacement statement (replace[0]) to this newly created CtBlock.
Set new method body: m.setBody(newBlock);
Finally, we set this CtBlock containing the new statement as the new method body for CtMethod.
Fallback to legacy logic: If the conditions for the edge case above are not met, the code executes the else block, following the original logic for handling CtBlock replacements. This ensures backward compatibility for the fix.
Through this fix, we intercepted the incorrect type conversion and manually constructed the correct AST structure, fundamentally resolving the ClassCastException issue.

Testing and Validation:
We constructed a dedicated unit test case that precisely reproduces the bug scenario described above. After applying this fix, both this test case and all relevant refactoring tests in our project passed successfully. This validates the effectiveness and stability of the fix.

### DIFF
--- a/src/main/java/spoon/support/visitor/replace/ReplacementVisitor.java
+++ b/src/main/java/spoon/support/visitor/replace/ReplacementVisitor.java
@@ -1934,17 +1934,35 @@ public class ReplacementVisitor extends spoon.reflect.visitor.CtScanner {
 	}
 
 	// auto-generated, see spoon.generating.ReplacementVisitorGenerator
-	@java.lang.Override
-	public <T> void visitCtMethod(final spoon.reflect.declaration.CtMethod<T> m) {
-		replaceInListIfExist(m.getAnnotations(), new spoon.support.visitor.replace.ReplacementVisitor.CtElementAnnotationsReplaceListener(m));
-		replaceInListIfExist(m.getFormalCtTypeParameters(), new spoon.support.visitor.replace.ReplacementVisitor.CtFormalTypeDeclarerFormalCtTypeParametersReplaceListener(m));
-		replaceElementIfExist(m.getType(), new spoon.support.visitor.replace.ReplacementVisitor.CtTypedElementTypeReplaceListener(m));
-		replaceElementIfExist(m.getReceiverParameter(), new spoon.support.visitor.replace.ReplacementVisitor.CtExecutableReceiverParameterReplaceListener(m));
-		replaceInListIfExist(m.getParameters(), new spoon.support.visitor.replace.ReplacementVisitor.CtExecutableParametersReplaceListener(m));
-		replaceInSetIfExist(m.getThrownTypes(), new spoon.support.visitor.replace.ReplacementVisitor.CtExecutableThrownTypesReplaceListener(m));
-		replaceElementIfExist(m.getBody(), new spoon.support.visitor.replace.ReplacementVisitor.CtExecutableBodyReplaceListener(m));
-		replaceInListIfExist(m.getComments(), new spoon.support.visitor.replace.ReplacementVisitor.CtElementCommentsReplaceListener(m));
-	}
+	   @java.lang.Override
+    public <T> void visitCtMethod(final spoon.reflect.declaration.CtMethod<T> m) {
+        replaceInListIfExist(m.getAnnotations(), new spoon.support.visitor.replace.ReplacementVisitor.CtElementAnnotationsReplaceListener(m));
+        replaceInListIfExist(m.getFormalCtTypeParameters(), new spoon.support.visitor.replace.ReplacementVisitor.CtFormalTypeDeclarerFormalCtTypeParametersReplaceListener(m));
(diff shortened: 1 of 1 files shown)


SirYwell:
Hi, please make sure to follow the [Contribution Guidelines](https://github.com/INRIA/spoon/blob/master/CONTRIBUTING.md#guidelines-for-pull-requests). In particular, you need to add a test case that fails without your change and succeeds with your change.

mywar5:
> Hi, please make sure to follow the [Contribution Guidelines](https://github.com/INRIA/spoon/blob/master/CONTRIBUTING.md#guidelines-for-pull-requests). In particular, you need to add a test case that fails without your change and succeeds with your change.

Okay, I'm sorry.

mywar5:
> Hi, please make sure to follow the [Contribution Guidelines](https://github.com/INRIA/spoon/blob/master/CONTRIBUTING.md#guidelines-for-pull-requests). In particular, you need to add a test case that fails without your change and succeeds with your change.

Sorry, I'm quite busy and might not have time to write another test case. I'll close this PR. This bug was encountered during my development work, and I've already explained the specifics and solution. You can review and fix it based on what I've outlined. Of course, you're free to ignore it entirely. Either way, consider this my small contribution.

PR #4619, opened 2022-02-21
test: delete junit 4 dependency

MartinWitt:

PR #6487, opened 2025-10-18
review: Remove references to JUnit 4

MaxAtoms:
As far as I can tell, JUnit 4 tests were all migrated in #3919 and JUnit 4 was removed with #4619. Therefore, the last remaining remnants should be removed.


monperrus:
:+1:

PR #6495, opened 2025-10-24
test: Fixed nondeterministic test failures in testShadowModelEqualsNormalModel()

yonghanlin:
## What does this PR do?
This PR fixes a nondeterministic test failure in ```JavaReflectionTreeBuilderTest.testShadowModelEqualsNormalModel()``` within ```.``` module when running with NonDex.

## Problem
The test compares ```CtType``` models built from source and from reflection (shadow models). However, both models internally use unordered collections, so the order of elements such as type members, enum values, and annotations vary between runs.

## Reproduce Test
```
mvn -pl . edu.illinois:nondex-maven-plugin:2.1.7:nondex -Dtest=spoon.support.visitor.java.JavaReflectionTreeBuilderTest#testShadowModelEqualsNormalModel
```

## The Fix
- Sorts and matches type members by signature (```memberSignature```).
- Compares enum constants by name instead of position.
- Sorts annotations by annotation type (qualified name).

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,16 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.java]
+indent_style = tab
+
+# IntelliJ IDEA specific settings
+ij_java_use_single_class_imports = true
+
+# https://stackoverflow.com/questions/3348816/intellij-idea-never-use-wildcard-imports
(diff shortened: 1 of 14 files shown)


monperrus:
This PR seems too long and too complex given the severity of the problem. Is there a shorter fix as in https://github.com/INRIA/spoon/pull/6500 and https://github.com/INRIA/spoon/pull/6501?

PR #6500, opened 2025-10-28
review: test: Fixed nondeterministic failures in ProcessingTest.testInitPropertiesWithWrongType()

yonghanlin:
## What does this PR do?
This PR draws inspiration from the earlier PR #6084, which has not seen recent activity. The fix has been re-implemented on the current codebase and updated accordingly. In particular, this PR fixes a nondeterministic test failure in ```ProcessingTest.testInitPropertiesWithWrongType()``` when running with NonDex.

## Problem
The failures occur when the test applies multiple processor properties within a single ```ProcessorProperties``` initialization. In the test, ```aString```, ```anObject```, and ```anInt``` were set within the same ```ProcessorProperties``` instance. Depending on iteration order, the ```anInt``` property could be processed before the valid ```aString``` and ```anObject```, causing ```initProperties()``` to throw before they were set.

## Reproduce Test
```
mvn -pl . edu.illinois:nondex-maven-plugin:2.1.7:nondex -Dtest=spoon.test.processing.ProcessingTest#testInitPropertiesWithWrongType
```

## The Fix
The test is restructured into two initialization phases:
1. Apply valid properties (```aString```, ```anObject```) and verify they are correctly set.
2. Apply the invalid property (```anInt="foo"```) separately and verify that it triggers ```SpoonException``` while leaving other properties unchanged.


monperrus:
Thanks a lot @yonghanlin 

We like short PRs to improve tests.

Tip: avoid opening many PRs at once, it's easily overwhelming for maintainers. Prefer opening them one by one, one after the other, once the previous one has been merged.

PR #6501, opened 2025-10-28
review: test: Fixed nondeterministic failures in SubstitutionTest.testCreateTypeFromTemplate()

yonghanlin:
## What does this PR do?
This PR draws inspiration from the earlier PR #6084, which has not seen recent activity. The fix has been re-implemented on the current codebase and updated accordingly. In particular, this PR fixes a nondeterministic test failure in ```SubstitutionTest.testCreateTypeFromTemplate()``` when running with NonDex.

## Problem
NonDex exposes that the generation of enum constants inside ```createTypeFromTemplate``` does not guarantee a fixed iteration order. During type construction, enum values are collected and inserted using data structures whose traversal order is unspecified (e.g., ```HashMap```). As a result, the produced enum may list its values as ```GOOD, BETTER``` in one run and ```BETTER, GOOD``` in another. The test assumed a strict positional order of enum constants, even though the underlying creation process does not enforce such ordering. This causes failures under certain NonDex seeds.

## Reproduce Test
```
mvn -pl . edu.illinois:nondex-maven-plugin:2.1.7:nondex -Dtest=spoon.test.template.SubstitutionTest#testCreateTypeFromTemplate
```

## The Fix
The assertion in ```testCreateTypeFromTemplate``` is updated to extract the enum value names, sort them, and then assert against the canonical list ```["BETTER", "GOOD"]```. This verifies correctness by content, not by insertion order.

### DIFF
--- a/.github/workflows/jreleaser.yml
+++ b/.github/workflows/jreleaser.yml
@@ -35,10 +35,15 @@ jobs:
           JRELEASER_DEPLOY_MAVEN_MAVENCENTRAL_SONATYPE_USERNAME: ${{ secrets.JRELEASER_DEPLOY_MAVEN_MAVENCENTRAL_SONATYPE_USERNAME }}
           JRELEASER_DEPLOY_MAVEN_MAVENCENTRAL_SONATYPE_TOKEN: ${{ secrets.JRELEASER_DEPLOY_MAVEN_MAVENCENTRAL_SONATYPE_TOKEN }}
 
+      - name: Prepare Signature Subject Checksums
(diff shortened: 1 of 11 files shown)


monperrus:
Thanks a lot @yonghanlin 

We like short PRs to improve tests.

Tip: avoid opening many PRs at once, it's easily overwhelming for maintainers. Prefer opening them one by one, one after the other, once the previous one has been merged.

yonghanlin:
@monperrus Thank you for the review and suggestion!

PR #6512, opened 2025-11-05
cd: Remove git merge into master for beta releases

LogFlames:
PR from discussion on CHAINS meeting.

Currently every week the beta-releases cause two commits to the master branch
```
release: Reverting to SNAPSHOT version 11.2.2-SNAPSHOT
release: Releasing version 11.2.2-beta-9
```

These make the commit history longer and more difficult to overview.

We can instead leave the release commits in separate branches without merging into master.
```
* master
|
| * beta-releases/11.2.2-beta-11 
|/
| * beta-releases/11.2.2-beta-10
|/
| * beta-releases/11.2.2-beta-9
|/
```

### DIFF
--- a/chore/release-beta.sh
+++ b/chore/release-beta.sh
@@ -41,3 +41,7 @@ echo "::endgroup::"
 echo "::group::Running jreleaser"
 JRELEASER_PROJECT_VERSION="$NEXT_BETA_VERSION" jreleaser-cli release
 echo "::endgroup::"
+
+echo "::group::Delete branch"
+git push origin --delete "$BRANCH_NAME"
+echo "::endgroup::"


I-Al-Istannen:
I am not a fan of leaving all those branches lying around to be honest. If you want to prevent GC, why not give them a tag instead? That sounds like the nicer model, unless we want to keep iterating on the beta branches?

LogFlames:
We discussed alternative solution, another proposal is to have a completely separate branch for beta releases, where we cherry-pick the release commits. What do you think of this setup?

```
* master
|
|    * beta-releases
|    | release: Releasing version 11.2.2-beta-11 (tag v11.2.2-beta-11)
|    | release: Releasing version 11.2.2-beta-10 (tag v11.2.2-beta-10)
|    | release: Releasing version 11.2.2-beta-9 (tag v11.2.2-beta-9)
```

Will update PR accordingly

I-Al-Istannen:
I think that sounds a bit annoying to view in retrospect, as you would need to squash all changes since the last beta into that one commit, no? That makes diffing and the lineage a lot less clear than a single commit based off the version it will release and kept alive by a branch or tag.

I mainly don't quite see why you want a branch instead of a tag, I am fine with the rest :) I.e. why not this:
image
Currently we use this model ^, but with branches instead of tags. Branches are just a bit more clutter in git (t)uis than tags, hence my question :) I don't quite see what the branch buys us, if we don't intend to add more fixup commits to the beta afterwards (which we probably don't)?

I am alright with merging this as-is, I just thought we might as well change the branch to a tag if we touch the workflow anyways.

LogFlames:
Ah now I see, thanks for the image!

Yes you're right, that is a way nicer solution. Have updated the PR to only remove the branch at the of the workflow. JReleaser will create and push the tag as part of the github release.

PR #6516, opened 2025-11-07
ignore sha256&512 in attestations

hboutemy:
will limit the number of useful subjects https://github.com/INRIA/spoon/attestations/13033254


monperrus:
thanks @hboutemy

PR #6556, opened 2025-12-07
Fix typo in argument name

seintur:
This PR fixes a typo in the argument name: suffix (not suffic).


I-Al-Istannen:
The CI failure is a bit unfortunate (we can't update the nix lockfile as one of the java versions we use is EOL and Hannes said the next one isn't yet supported by JDT).

I think I will move the update time checker out of the main setup path, see https://github.com/INRIA/spoon/pull/6557.

SirYwell (approved):

PR #6557, opened 2025-12-07
chore: Disable fail mode for flake-checker-action

I-Al-Istannen:
This came up again in https://github.com/INRIA/spoon/pull/6556 and we likely do not want to block contributions on our failure to keep dependencies up to date. At the same time, we really do want to update our dependencies sometime, to ensure we build against up-to-date software, hence marking this as non-fail and emitting a warning. Once JDT is updated we can adjust our CI to drop 23 and then renovate can bump the lockfile again.


SirYwell:
Merging directly to unblock other PRs.

PR #6560, opened 2025-12-10
Fix: ordering of implicit fields generated from record components

seintur:
This PR preserves the same ordering for implicit fields generated from record components as the one of record components.

Implicit type members are inserted at index 0. When adding several record components, the implicit fields generated from these record components end up being in the reverse order. This PR fixes that by inserting each field at the index of the corresponding record component in the list of record components.

### DIFF
--- a/src/test/java/spoon/test/record/CtRecordTest.java
+++ b/src/test/java/spoon/test/record/CtRecordTest.java
@@ -336,20 +336,22 @@ public class CtRecordTest {
 	void testRecordComponentOrder() {
 		// contract: implicit fields generated from record components are the same order as the record components
 		Factory factory = new Launcher().getFactory();
-		CtRecord record = factory.createRecord();
-		record.addModifier(ModifierKind.PUBLIC);
-		record.setSimpleName("RecordComponentOrder");
-		CtRecordComponent recordComponent1 = factory.createRecordComponent();
-		recordComponent1.setType(factory.createCtTypeReference(int.class));
-		recordComponent1.setSimpleName("first");
-		record.addRecordComponent(recordComponent1);
-		CtRecordComponent recordComponent2 = factory.createRecordComponent();
-		recordComponent2.setType(factory.createCtTypeReference(float.class));
-		recordComponent2.setSimpleName("second");
-		record.addRecordComponent(recordComponent2);
-
-		assertEquals("first", record.getFields().get(0).getSimpleName());
-		assertEquals("second", record.getFields().get(1).getSimpleName());
+		CtRecord record = factory.createRecord()
+			.<CtRecord>setSimpleName("RecordComponentOrder")
+			.<CtRecord>addModifier(ModifierKind.PUBLIC)
+			.addRecordComponent(
+				factory.createRecordComponent()
+					.<CtRecordComponent>setType(factory.Type().integerPrimitiveType())
+					.setSimpleName("first")
+			)
+			.addRecordComponent(
+				factory.createRecordComponent()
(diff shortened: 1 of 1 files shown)


src/test/java/spoon/test/record/CtRecordTest.java
```diff
@@ -331,6 +332,26 @@ public record User(int id, String name) {
...
+		record.addRecordComponent(recordComponent1);
+		CtRecordComponent recordComponent2 = factory.createRecordComponent();
+		recordComponent2.setType(factory.createCtTypeReference(float.class));
+		recordComponent2.setSimpleName("second");
+		record.addRecordComponent(recordComponent2);
+
+		assertEquals("first", record.getFields().get(0).getSimpleName());
+		assertEquals("second", record.getFields().get(1).getSimpleName());
+	}
+
```
I-Al-Istannen, line 356: I am not sure if this is nicer. Maybe? I think at least the assert is. It also has much nicer errors:
```
Expecting actual:
  ["first", "second"]
to contain exactly (and in same order):
  ["first", "second", "sa"]
but could not find the following elements:
  ["sa"]
```
or
```
Expecting actual:
  ["first", "second"]
to contain exactly (and in same order):
  ["second", "first"]
but there were differences at these indexes:
  - element at index 0: expected "second" but was "first"
  - element at index 1: expected "first" but was "second"
```

```suggestion
	@Test
	void testRecordComponentOrder() {
		// contract: implicit fields generated from record components are the same order as the record components
		Factory factory = new Launcher().getFactory();
		CtRecord record = factory.createRecord()
			.<CtRecord>setSimpleName("RecordComponentOrder")
			.<CtRecord>addModifier(ModifierKind.PUBLIC)
			.addRecordComponent(
				factory.createRecordComponent()
					.<CtRecordComponent>setType(factory.Type().integerPrimitiveType())
					.setSimpleName("first")
			)
			.addRecordComponent(
				factory.createRecordComponent()
					.<CtRecordComponent>setType(factory.Type().floatPrimitiveType())
					.setSimpleName("second")
			);

		assertThat(record.getFields())
			.map(CtField::getSimpleName)
			.containsExactly("first", "second");
	}
```

seintur:
Very good to me, thanks for the review and the suggestion

src/test/java/spoon/test/record/CtRecordTest.java
```diff
@@ -331,6 +332,26 @@ public record User(int id, String name) {
...
+		recordComponent1.setType(factory.createCtTypeReference(int.class));
+		recordComponent1.setSimpleName("first");
+		record.addRecordComponent(recordComponent1);
+		CtRecordComponent recordComponent2 = factory.createRecordComponent();
+		recordComponent2.setType(factory.createCtTypeReference(float.class));
+		recordComponent2.setSimpleName("second");
+		record.addRecordComponent(recordComponent2);
+
+		assertEquals("first", record.getFields().get(0).getSimpleName());
+		assertEquals("second", record.getFields().get(1).getSimpleName());
```
I-Al-Istannen: IMHO at least the assertion should be changed to what I suggested. I don't care as much about the rest, but the better error messages from the (customized) assertj-assert are worth it. It also looks a bit cleaner IMHO, but that's up for interpretation.

```suggestion
		assertThat(record.getFields())
			.map(CtField::getSimpleName)
			.containsExactly("first", "second");
```

I-Al-Istannen (changes requested):

src/test/java/spoon/test/record/CtRecordTest.java
seintur: Agree and took into account, thanks

PR #6572, opened 2025-12-21
Add the record canonical constructor if not present

seintur:
This PR adds the canonical constructor to a record if it is not already present.

### DIFF
--- a/src/test/java/spoon/test/record/CtRecordFactoryTest.java
+++ b/src/test/java/spoon/test/record/CtRecordFactoryTest.java
@@ -11,6 +11,7 @@ import spoon.reflect.declaration.CtRecord;
 import spoon.reflect.declaration.CtRecordComponent;
 import spoon.reflect.declaration.ModifierKind;
 import spoon.reflect.factory.Factory;
+import spoon.reflect.reference.CtTypeReference;
 import spoon.reflect.visitor.CtScanner;
 
 import java.util.Arrays;
@@ -193,6 +194,67 @@ public class CtRecordFactoryTest {
 		assertTrue(sortedConstructors[1].isImplicit());
 		assertEquals("i", sortedConstructors[1].getParameters().get(0).getSimpleName());
 		assertFalse(sortedConstructors[1].isCompactConstructor());
+	}
+
+	@Test
+	void testCreateCanonicalConstructorIfMissing() {
+		// contract: a canonical constructor is added when the record is missing one
+		Factory factory = new Launcher().getFactory();
+		CtRecord record = factory.createRecord()
+			.<CtRecord>setSimpleName("RecordWithNoCanonicalConstructor")
+			.<CtRecord>addModifier(ModifierKind.PUBLIC)
+			.addRecordComponent(
+				factory.createRecordComponent()
+					.<CtRecordComponent>setType(factory.Type().integerPrimitiveType())
+					.setSimpleName("value")
+			);
+
+		assertThat(record).getConstructors().hasSize(0);
+
+		record.createCanonicalConstructorIfMissing();
+
+		assertThat(record).getConstructors().hasSize(1);
+		CtConstructor<?> constructor = getFirst(record.getConstructors());
+		assertThat(constructor).isImplicit();
+		assertThat(constructor).getParameters().hasSize(1);
+		CtTypeReference<?> expectedType = factory.Type().integerPrimitiveType();
+		assertThat(expectedType).isEqualTo(constructor.getParameters().get(0).getType());
+	}
+
+	@Test
+	void testCreateCanonicalConstructorIfPresent() {
+		// contract: a canonical constructor is not added when the record defines one
+		Factory factory = new Launcher().getFactory();
+		CtRecord record = factory.createRecord()
+			.<CtRecord>setSimpleName("RecordWithCanonicalConstructor")
+			.<CtRecord>addModifier(ModifierKind.PUBLIC)
+			.addRecordComponent(
+				factory.createRecordComponent()
+					.<CtRecordComponent>setType(factory.Type().integerPrimitiveType())
+					.setSimpleName("value")
+			)
+			.addConstructor(
+				factory.createConstructor()
+					.addParameter(
+						factory.createParameter()
+							.<CtParameter<?>>setType(factory.Type().integerPrimitiveType())
+							.setSimpleName("s")
+					)
+			);
+
+		assertThat(record).getConstructors().hasSize(1);
+		CtConstructor<?> constructor = getFirst(record.getConstructors());
+		assertThat(constructor).getParameters().hasSize(1);
+		CtTypeReference<?> expectedType = factory.Type().integerPrimitiveType();
+		assertThat(expectedType).isEqualTo(constructor.getParameters().get(0).getType());
+
+		record.createCanonicalConstructorIfMissing();
+
+		assertThat(record).getConstructors().hasSize(1);
+		constructor = getFirst(record.getConstructors());
+		assertThat(constructor).isImplicit();
+		assertThat(constructor).getParameters().hasSize(1);
+		assertThat(expectedType).isEqualTo(constructor.getParameters().get(0).getType());
 	}
 
 	private <T> T getFirst(Collection<T> collection) {


src/main/java/spoon/reflect/declaration/CtRecord.java
```diff
@@ -30,6 +30,13 @@ public interface CtRecord extends CtClass<Object> {
 	@PropertySetter(role = CtRole.RECORD_COMPONENT)
 	CtRecord  removeRecordComponent(CtRecordComponent component);
 
+	/**
+	 * Adds the canonical constructor for this record if it is not already present.
+	 *
+	 * @return the current record declaration instance
+	 */
+	CtRecord canonicalConstructor();
```
SirYwell: Hm, I'm not sure if we want to expose it this way. The naming is a bit odd and it sticks out compared to other methods in how much it does. To not block this PR, maybe we can mark the method as `@Experimental`? Maybe others have more opinions on this :)

seintur: Thanks for the review. It's fine with me if one finds a better name or a better way to expose this feature. I am open to suggestions.

Two comments on why this PR.

First, so far, Spoon provides a way to add implicit fields and accessor methods associated to record component (see method addRecordComponent https://github.com/INRIA/spoon/blob/cd0daf82fe89eff630b05ef4256122661568f1b3/src/main/java/spoon/support/reflect/declaration/CtRecordImpl.java#L62) but not the implicit canonical constructor that tests assume is present in CtRecord instances. This PR aims at providing a way for that.

Second, up until the current version of JDT used by Spoon, JDT adds fields, methods, and constructor to the AST for records. As the currently failing PR https://github.com/INRIA/spoon/pull/6316 shows it, this is no longer the case starting with JDT 3.42.0. So if we want to keep enforcing the same contracts, one needs a way to add the canonical constructor to a CtRecord instance.

Have a happy end of year season :)

src/main/java/spoon/support/reflect/declaration/CtRecordImpl.java
```diff
@@ -92,6 +100,51 @@ && getField(component.getSimpleName()).isImplicit()) {
...
+				.filter(CtElement::isImplicit)
+				.map(CtTypedElement::getType)
+				.toList().toArray(new CtTypeReference[0]);
+
+		// If a canonical constructor is not already present, adds one
+		CtConstructor<?> constructor = getConstructor(typeReferences);
+		if (constructor == null) {
+			CtConstructor<?> canonical = getFactory().createConstructor();
+			canonical.setImplicit(true);
+			canonical.setExtendedModifiers(Collections.singleton(new CtExtendedModifier(ModifierKind.PUBLIC, true)));
```
MartinWitt: `List.of` and  `CtExtendedModifier#implict`

src/test/java/spoon/test/record/CtRecordFactoryTest.java
```diff
@@ -0,0 +1,200 @@
...
+			.<CtRecord>setSimpleName("SingleParameterRecord")
+			.<CtRecord>addModifier(ModifierKind.PUBLIC)
+			.addRecordComponent(
+				factory.createRecordComponent()
+					.<CtRecordComponent>setType(factory.Type().integerPrimitiveType())
+					.setSimpleName("value")
+			)
+			.canonicalConstructor();
+
+		assertEquals(1, record.getFields().size());
```
MartinWitt: I prefer assertJ (with our own asserts), could u convert it? assertTrue alone is very hard to understand in the future.

src/main/java/spoon/reflect/declaration/CtRecord.java
MartinWitt: How about something like `createCanonicalConstructorIfMissing`, and absolutly `@Experimental`?

src/test/java/spoon/test/record/CtRecordFactoryTest.java
```diff
@@ -0,0 +1,200 @@
...
+		assertFalse(sortedConstructors[0].isImplicit());
+		assertEquals("s", sortedConstructors[0].getParameters().get(0).getSimpleName());
+		assertFalse(sortedConstructors[0].isCompactConstructor());
+
+		assertTrue(sortedConstructors[1].isImplicit());
+		assertEquals("i", sortedConstructors[1].getParameters().get(0).getSimpleName());
+		assertFalse(sortedConstructors[1].isCompactConstructor());
+	}
+
+	private <T> T head(Collection<T> collection) {
```
MartinWitt: kinda funny how long this joke survives. Given that Java uses getFirst we should at least align the naming even if we can't use the new API.

MartinWitt (commented):
Just some opinions about code style.

src/main/java/spoon/reflect/declaration/CtRecord.java
seintur: Thanks for the review. Commit done.

src/main/java/spoon/support/reflect/declaration/CtRecordImpl.java
seintur: Thanks for the review. Commit done.

src/test/java/spoon/test/record/CtRecordFactoryTest.java
seintur: Thanks for the review. Commit done.

src/test/java/spoon/test/record/CtRecordFactoryTest.java
seintur: Thanks for the review. Commit done.

src/main/java/spoon/support/reflect/declaration/CtRecordImpl.java
```diff
@@ -92,6 +100,51 @@ && getField(component.getSimpleName()).isImplicit()) {
 		return this;
 	}
 
+	@Override
+	public CtRecord createCanonicalConstructorIfMissing() {
+
+		CtTypeReference<?>[] typeReferences =
+			getFields().stream()
+				.filter(CtElement::isImplicit)
```
SirYwell: Is there any reason for not using `CtRecord#getRecordComponents` here?

src/main/java/spoon/support/reflect/declaration/CtRecordImpl.java
```diff
@@ -92,6 +100,51 @@ && getField(component.getSimpleName()).isImplicit()) {
...
+				.map(CtTypedElement::getType)
+				.toList().toArray(new CtTypeReference[0]);
+
+		// If a canonical constructor is not already present, adds one
+		CtConstructor<?> constructor = getConstructor(typeReferences);
+		if (constructor == null) {
+			CtConstructor<?> canonical = getFactory().createConstructor();
+			canonical.setImplicit(true);
+			canonical.setExtendedModifiers(Set.of(CtExtendedModifier.implicit(ModifierKind.PUBLIC)));
```
SirYwell: According to <https://docs.oracle.com/javase/specs/jls/se25/html/jls-8.html#jls-8.10.4>, the modifier should correspond to the visibility of the record itself. I guess the easiest would be to use `ModifierTarget.CONSTRUCTOR.getAllowedKinds()`, check if the modifier is existent on the record, and apply it on the constructor if so.

SirYwell (changes requested):

src/main/java/spoon/support/reflect/declaration/CtRecordImpl.java
seintur: Thanks for the review. No reason indeed. Commit done.

src/main/java/spoon/support/reflect/declaration/CtRecordImpl.java
seintur: Thanks for pointing that.

ModifierTarget is package protected in spoon.support.compiler.jdt hence not visible from CtRecordImpl:

https://github.com/INRIA/spoon/blob/cad2b2a8d4fdc5993a65f960d0089185634d0959/src/main/java/spoon/support/compiler/jdt/ModifierTarget.java#L23

Any recommendation between: (1) making ModifierTarget public? or (2) implementing in CtRecordImpl the four cases (public, protected, package access, private) mentioned in the JLS?

SirYwell: Hm, I'm fine with (2), which should be simple enough. If you want to go with (1), it might make sense to move the class to a different place.

seintur: Commit done with (2).

src/test/java/spoon/test/record/CtRecordFactoryTest.java
```diff
@@ -0,0 +1,199 @@
...
+			record.getConstructors().stream()
+				.map(String::valueOf)
+				.map(s -> s.replaceAll("\\R", "\n")) // fix newlines on Windows
+				.collect(Collectors.toList())
+		);
+	}
+
+	@Test
+	void testMultipleConstructors() {
+		// contract: we can have an explicit constructor delegating to the implicit canonical constructor
```
SirYwell, line 160: This test isn't really clear to me. The parameter of the second constructor also doesn't have a type from what it looks like.

It might also be good to have a test that tests the *if missing* (or rather, the *if present*) part of `createCanonicalConstructorIfMissing` method, to ensure it doesn't change the model if a canonical constructor is already present.

seintur: As a preliminary statement, I can say that this test is adapted from CtRecordTest which was there before:

https://github.com/INRIA/spoon/blob/cad2b2a8d4fdc5993a65f960d0089185634d0959/src/test/java/spoon/test/record/CtRecordTest.java#L251

My intent with CtRecordFactoryTest was to stay as close as possible to the existing tests in CtRecordTest: CtRecordTest provides tests for models built from JDT, and CtRecordFactoryTest would provide the same tests for models built directly from the Factory API.

Having said that, I agree that the tests can be clearer:

- I am adding the type for the second constructor.
- CtRecordFactoryTest#testMultipleParameterRecord checks the _if missing_ (and the fields, and the accessor methods). Again I agree that this may be more explicit with a _if missing_ test for the canonical constructor. I am adding that.

seintur: Commits done.

SirYwell (approved):
Very cool, looks good now. Let's wait a bit for others to chime in, otherwise I'll merge tomorrow.

SirYwell:
Thanks a lot for addressing this @seintur.

seintur:
You are welcome

PR #6575, opened 2025-12-25
review: fix: fix equals bug in ElementNameMap

### LINKED ISSUE #6033: [Bug]: spoon.support.util.internal.NamedElementMap has a bug in equals() method
### Describe the bug

- ElementNameMap.equals() only returns equal if the InsertOrderWrapper internal instances are exactly the same, as they don't implement hashCode() and equals() so that unfortunately means that equal() can only return true if the other object is the same ElementNameMap instance.

- ElementNameMap.equals() does not consider the order of the wrapped elements. That is the wrapped elements in two separate ElementNameMap instances can have substantially different insertionNumbers, but have the same order and contain equivalent objects, so that for those cases the equal() test should return true. This can be solved by comparing the elements in equal() by their correct order using entrySet(...) to check. 

Proposed solution patch:
```
--- a/src/main/java/spoon/support/util/internal/ElementNameMap.java	2024-10-23 02:02:48.000000000 +0100
+++ b/src/main/java/spoon/support/util/internal/ElementNameMap.java	2024-10-23 16:34:58.573946074 +0100
@@ -71,6 +71,28 @@
 			this.value = value;
 			this.insertionNumber = insertionNumber;
 		}
+		
+		@Override
+		public int hashCode() {
+			return Objects.hash(this.value);
+		}
+
+		@Override
+		public boolean equals(Object o) {
+			if (this == o) {
+				return true;
+			}
+			if (o == null || getClass() != o.getClass()) {
+				return false;
+			}
+            
+			InsertOrderWrapper<?> that = (InsertOrderWrapper<?>) o;
+
+			//Insertion number is not relevant for comparing the wrapped elements as long as they
+			//are presented in the same ordering by entrySet().
+			return Objects.equals(this.value, that.value);
+        	}
+
 	}
 
 
@@ -211,7 +233,7 @@
 			return false;
 		}
 		ElementNameMap<?> that = (ElementNameMap<?>) o;
-		return Objects.equals(map, that.map);
+		return Objects.equals(entrySet(), that.entrySet());
 	}
 
 	@Override
```

_No response_

_No response_

### Actual output

_No response_

### Expected output

_No response_

### Spoon Version

Latest from master

### JVM Version

Not a JVM version issue

NixOS Linux, but again not relevant for the problem

MartinWitt:
fixes #6033

### DIFF
--- a/src/main/java/spoon/support/util/internal/ElementNameMap.java
+++ b/src/main/java/spoon/support/util/internal/ElementNameMap.java
@@ -62,9 +62,9 @@ public abstract class ElementNameMap<T extends CtElement> extends AbstractMap<St
 	/**
 	 * Wrapper class that allows us to return entries in the order they were inserted.
 	 */
-		private record InsertOrderWrapper<T extends Serializable>(T value, long insertionNumber) implements Serializable {
-			@Serial
-			private static final long serialVersionUID = 1L;
+	private record InsertOrderWrapper<T extends Serializable>(T value, long insertionNumber) implements Serializable {
+		@Serial
+		private static final long serialVersionUID = 1L;
 
 	}
 


src/main/java/spoon/support/util/internal/ElementNameMap.java
```diff
@@ -207,11 +202,11 @@ public boolean equals(Object o) {
 		if (this == o) {
 			return true;
 		}
-		if (o == null || getClass() != o.getClass()) {
+		if (!(o instanceof ElementNameMap<?> that)) {
```
I-Al-Istannen, line 205: This now has differing behaviour: It allows subclasses to be passed and compared as equals. I am not sure we want this, but our subclasses might be well-behaved?

Have you thought about whether that breaks stuff?

monperrus:
Thanks @MartinWitt

PR #6581, opened 2025-12-29
fix: Turning CtField elements in CtRecord explicit before asserting

seintur:
Before version 3.42.0, JDT was adding explicit fields corresponding to record components in the AST for records. Starting with 3.42.0 this is no longer the case. We have to rely on the way Spoon is handing this: addRecordComponent generate implicit fields. So this PR turns implicit fields into explicit ones before asserting.


SirYwell (approved):
Good, let's merge directly as it targets the JDT PR.

PR #6582, opened 2025-12-30
review: refactor: remove reflection needed on Java pre 17

SirYwell:
Before switching to Java 17, we already supported some features that require Java 16/17 methods for the reflection model. Since then, we can drop this workaround and just call the method directly.

`JavaReflectionVisitor`, `MethodHandleUtils`, and `JavaReflectionVisitorImpl` are package-private, so compatibility isn't an issue here. The only problem would be the changed method signature in `JavaReflectionTreeBuilder`, but I think this is acceptable.


MartinWitt:
looks good when CI is green.

MartinWitt:
thanks @SirYwell

PR #6586, opened 2026-01-03
review: feat: add support for module imports

MartinWitt:
needed for java25

### DIFF
--- a/src/test/java/spoon/reflect/visitor/DefaultJavaPrettyPrinterTest.java
+++ b/src/test/java/spoon/reflect/visitor/DefaultJavaPrettyPrinterTest.java
@@ -41,7 +41,6 @@ import spoon.reflect.reference.CtTypeReference;
 import spoon.reflect.visitor.filter.TypeFilter;
 import spoon.support.reflect.reference.CtArrayTypeReferenceImpl;
 import spoon.test.SpoonTestHelpers;
-import spoon.testing.assertions.SpoonAssertions;
 import spoon.testing.utils.BySimpleName;
 import spoon.testing.utils.GitHubIssue;
 import spoon.testing.utils.ModelTest;


src/main/java/spoon/reflect/declaration/CtImportKind.java
```diff
@@ -13,6 +13,7 @@ public enum CtImportKind {
 	ALL_STATIC_MEMBERS, // import static my.package.Type.*;
 	FIELD, // import static my.package.Type.f;
 	METHOD, // import static my.package.Type.m;
-	UNRESOLVED // Any of the above when in mode no classpath and the reference cannot be resolved.
+	UNRESOLVED, // Any of the above when in mode no classpath and the reference cannot be resolved.
 	// It is then stored as a pure String that will be printed as is when pretty printed.
+	MODULE
```
SirYwell: maybe it makes sense to put these comments (as javadoc?) above the constant. Otherwise it's looking a bit odd now, with the two line comment (and it also sounds like the comment doesn't really apply like this anymore?)

src/main/java/spoon/reflect/visitor/ImportCleaner.java
```diff
@@ -264,8 +264,11 @@ void onCompilationUnitProcessed(CtCompilationUnit compilationUnit) {
 			ModelList<CtImport> existingImports = compilationUnit.getImports();
 			Set<CtImport> computedImports = new HashSet<>(this.computedImports.values());
 			topfor: for (CtImport oldImport : new ArrayList<>(existingImports)) {
+				if (oldImport.getImportKind() == CtImportKind.MODULE) {
+					//never remove module imports
+					continue;
+				}
```
SirYwell, line 270: I assume this is a conservative solution, i.e., removing such an import would be legal/might make sense in some cases (e.g., if we're in the same module?).

In that case, I think the comment should reflect that a bit better, e.g., "be conservative: do not remove any module imports". This should help if anyone ever tries to do more here.

src/main/java/spoon/support/reflect/declaration/CtImportImpl.java
```diff
@@ -82,30 +85,14 @@ public void accept(CtVisitor visitor) {
...
-		default:
-			throw new SpoonException("Unexpected import kind: " + getImportKind());
+			case TYPE -> visitor.visitTypeImport((CtTypeReference<?>) localReference);
+			case METHOD -> visitor.visitMethodImport((CtExecutableReference<?>) localReference);
+			case FIELD -> visitor.visitFieldImport((CtFieldReference<?>) localReference);
+			case ALL_TYPES -> visitor.visitAllTypesImport((CtPackageReference) localReference);
+			case ALL_STATIC_MEMBERS -> visitor.visitAllStaticMembersImport((CtTypeMemberWildcardImportReference) localReference);
+			case MODULE -> visitor.visitModuleImport((CtModuleReference) localReference);
+			case UNRESOLVED -> visitor.visitUnresolvedImport((CtUnresolvedImport) localReference);
+			default -> throw new SpoonException("Unexpected import kind: " + getImportKind());
```
SirYwell, line 95: nice

src/test/java/spoon/reflect/visitor/DefaultJavaPrettyPrinterTest.java
```diff
@@ -474,4 +477,12 @@ void printAnnotationsInOrphanTypeReference(Factory factory) {
     private @interface TypeUseAnnotation {
 
     }
+
+	@ModelTest(value = {"src/test/resources/imports/ModuleImport.java"}, autoImport = true, complianceLevel = 25)
+	void moduleImportsArePrinted(@BySimpleName("ModuleImport") CtType<?> type, Launcher launcher) {
```
SirYwell, line 521: I already forgot about the `@BySimpleName` annotation again, but it definitely makes the test cleaner. Can you add a `// contract:` comment though?

src/test/java/spoon/test/imports/ImportTest.java
```diff
@@ -1904,4 +1915,15 @@ void staticTypeAndMethodImport_importShouldAppearOnlyOnceIfTheirSimpleNamesAreEq
 		assertThat(import1.getImportKind(), is(CtImportKind.TYPE));
 		assertThat(import1.getReference().getSimpleName(), is("foo"));
 	}
+
+	@ModelTest(value = {"src/test/resources/imports/ModuleImport.java"}, autoImport = true, complianceLevel = 25)
+	void moduleImportsAreCorrectAddedToTheModel(@BySimpleName("ModuleImport") CtType<?> type) {
+		// contract: we can handle module imports correctly
+		ModelList<CtImport> imports = type.getPosition().getCompilationUnit().getImports();
+		CtImport ctImport = imports.stream().filter(it -> it.getImportKind() == CtImportKind.MODULE).findFirst().orElseThrow();
```
SirYwell: Do we want to assert that no other imports are added somewhere accidentally?

MartinWitt: Actually, we are adding 3 more imports. I'm unsure if we should add a lot of code to compute what is included in an module and use it for the calculation. Imports were always scarry.

SirYwell: Hm, where are these coming from?

src/main/java/spoon/reflect/declaration/CtImportKind.java
```diff
@@ -7,12 +7,76 @@
  */
 package spoon.reflect.declaration;
 
+/**
+ * Enumeration of the different kinds of Java imports that can appear in a compilation unit.
+ *
+ * <p>This enum represents both traditional package imports (from Java source files) and module
+ * imports (from the Java Platform Module System, JPMS).
```
SirYwell:
```suggestion
 * imports.
```
JPMS is more of a working title, and not really relevant here.

src/main/java/spoon/reflect/declaration/CtImportKind.java
```diff
@@ -7,12 +7,76 @@
...
+	 *
+	 * <p>Example: {@code import static my.package.Type.f;}
+	 *
+	 * <p>This allows referencing the static field by its simple name within the importing
+	 * compilation unit.
+	 */
+	FIELD,
+
+	/**
+	 * A static import declaration for a single static method.
```
SirYwell: I don't think this is correct, as all static methods with that name are imported.

src/main/java/spoon/reflect/declaration/CtImportKind.java
```diff
@@ -7,12 +7,76 @@
...
+	 * An import that cannot be resolved to a specific kind.
+	 *
+	 * <p>This kind is used when Spoon operates in no-classpath mode and the reference cannot be
+	 * resolved. In such cases, the import is stored as a plain string and printed as-is during
+	 * pretty printing. The original import statement is preserved without semantic analysis.
+	 */
+	UNRESOLVED,
+
+	/**
+	 * A module import declaration for all public types of a Java Platform Module System module.
```
SirYwell:
```suggestion
	 * A module import declaration for all public types of a Java module.
```

src/test/java/spoon/test/imports/ImportTest.java
```diff
@@ -1904,4 +1915,18 @@ void staticTypeAndMethodImport_importShouldAppearOnlyOnceIfTheirSimpleNamesAreEq
...
 	}
+
+	@ModelTest(value = {"src/test/resources/imports/ModuleImport.java"}, autoImport = true, complianceLevel = 25)
+	void moduleImportsAreCorrectAddedToTheModel(@BySimpleName("ModuleImport") CtType<?> type) {
+		// contract: we can handle module imports correctly
+		ModelList<CtImport> imports = type.getPosition().getCompilationUnit().getImports();
+		//check that we only have one import
+		assertThat(imports, hasSize(1));
+		//check that it is module import
+		CtImport ctImport = imports.stream().filter(it -> it.getImportKind() == CtImportKind.MODULE).findFirst().orElseThrow();
```
SirYwell: We have an assert for that below, so we can probably just get the first element of the list?

SirYwell (changes requested):
A few more things

seintur:
Hi @MartinWitt

Great.

Just wondering: no `case MODULE` nor `setModuleImports` in ElementPrinterHelper#writeImports https://github.com/INRIA/spoon/blob/da794d3fcd1100d471574c2b46ca83e4d775c44d/src/main/java/spoon/reflect/visitor/ElementPrinterHelper.java#L301 ?

SirYwell:
@MartinWitt it looks like you addressed @seintur's comment, right?

MartinWitt:
> @MartinWitt it looks like you addressed @seintur's comment, right?

yes, should be ready for merge now

SirYwell (approved):

SirYwell:
Please resolve the conflicts @MartinWitt.

monperrus:
great!

sabroe:
How about making the actual Spoon implementation adhere to normal JPMS, hence add a module-info.java !?? It is very much missed.

MartinWitt:
> How about making the actual Spoon implementation adhere to normal JPMS, hence add a module-info.java !?? It is very much missed.

Do you mind moving this to an issue and providing some more context so we could discuss this?

PR #6588, opened 2026-01-05
fix: Implicit class for compact source files

seintur:
This PR:

-  declares the CtClass instance generated from a compact source file as implicit,
-  modifies the DefaultJavaPrettyPrinter to consider the case where a CtClass declaration can be implicit while the contained type members are explicit,
-  adds a printing test to CtClassTest#testCompactSourceFilesAndInstanceMainMethods to check that the printing corresponds to the compact source file.


SirYwell:
Would it make sense to make the test a bit more complex? AFAIK compact source files can have normal instance fields, static fields, additional methods, nested classes, imports...

seintur:
Thanks @SirYwell for the review.

> Would it make sense to make the test a bit more complex? AFAIK compact source files can have normal instance fields, static fields, additional methods, nested classes, imports...

I have added some.

SirYwell (approved):
Nice, looks good from my side!

SirYwell:
Thanks @seintur!

seintur:
You are welcome

PR #6594, opened 2026-01-11
doc: Update CONTRIBUTING.md with testing guidelines

MartinWitt:
This was missing, or at least I didn't find it. Any comments?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -124,12 +124,6 @@ Tests in Spoon must follow these guidelines:
 
 * **@ModelTest Annotation**: This annotation automatically builds and injects a Spoon model into your test. It is useful when your test needs to analyze or transform Java source code.
   * Usage: `@ModelTest(value = {"path/to/source"})` or `@ModelTest("./src/test/resources/...")`
-  * Supported parameters:
-    * `value`: Array of paths to Java source files or directories (required)
-    * `noClasspath`: Whether to build the model in no-classpath mode (default: true)
-    * `commentsEnabled`: Whether to preserve comments (default: true)
-    * `autoImport`: Whether to auto-import types (default: false)
-    * `complianceLevel`: Java compliance level (default: -1, which uses project default)
   * The annotation automatically injects `Factory`, `CtModel`, and/or `Launcher` parameters into your test method. You can request any or all of them.
   * Basic Example:
     ```java


CONTRIBUTING.md
```diff
@@ -95,3 +95,87 @@ OR
...
 Submodules are not published to Maven central, and so need to be built from source.
+
+Testing
+-------
+
+### Test Framework and Assertions
+
+Tests in Spoon must follow these guidelines:
+
+* **Test Framework**: Use JUnit 6 (Jupiter) with the `@Test` annotation or `@ModelTest` for tests that need a Spoon model.
```
SirYwell: `@ModelTest` doesn't work for all tests that need a spoon model, e.g., reflection tests.

I'd rather just reorder the list and also include `@ParameterizedTest` (is the spelling correct?)

CONTRIBUTING.md
```diff
@@ -95,3 +95,87 @@ OR
...
+    ```java
+    assertThat(element).isEqualTo(expected);
+    assertThat(parameter).hasSimpleName("paramName");
+    ```
+  * For standard assertions (objects, strings, numbers), use `org.assertj.core.api.Assertions.assertThat`.
+
+* **Contract Comments**: Every test must include a Javadoc contract comment explaining what behavior the test verifies. For example:
+  ```java
+  @Test
+  public void testSetType() {
```
SirYwell:
```suggestion
  void testSetType() {
```
let's drop that here and below

CONTRIBUTING.md
```diff
@@ -95,3 +95,87 @@ OR
...
+  * Basic Example:
+    ```java
+    @ModelTest("./src/test/resources/spoon/test/visitor/Foo.java")
+    public void testRecursiveDescent(Factory factory) {
+        final MyVisitor visitor = new MyVisitor(2);
+        visitor.scan(factory.Package().getRootPackage());
+        assertTrue(visitor.equals);
+    }
+    ```
+  * **@BySimpleName Annotation**: Use this to inject a specific type from the model by its simple name. This is useful when you want to directly work with a specific class.
```
SirYwell, line 137: There also is `@ByClass` or something like that, but it's mainly about older tests and probably not something we want in new tests.

monperrus:
@SirYwell  happy to merge?

PR #6628, opened 2026-02-09
Change publishDate position to 'none' in site.xml

monperrus:
Will save some commits in https://github.com/SpoonLabs/SpoonLabs.github.io/commits

### DIFF
--- a/src/site/site.xml
+++ b/src/site/site.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project>
-  <publishDate position="left" format="yyyy-MM-dd HH:mm" />
+  <publishDate position="none" format=""/>
 
   <!-- necessary for new site and checkstyle plugins -->
   <skin>


monperrus:
Interesting, it breaks the reproducible-build check.

```
The jars were not reproducible
  spoon-core-VERSION-with-dependencies.jar was not reproducible!
  spoon-core-VERSION.jar was not reproducible!
  spoon-javadoc-VERSION.jar was not reproducible!
```

@algomaster99 it would be great to have a diffoscope output in the job log, WDYT?

I-Al-Istannen:
It already runs diffoscope for comparisons and *should* print the results. In this case, docker pull failed.

> docker: Error response from daemon: Get "https://registry.salsa.debian.org/v2/": dial tcp 209.87.16.44:443: connect: connection refused

So I guess you need to restart the job. We can also switch to `nix run nixpkgs#diffoscopeMinimal` (~30 MiB) or `diffoscope` (1.5 GiB) to no longer depend on that docker pull.

monperrus:
thanks @I-Al-Istannen indeed it's green. would you merge then?

monperrus:
and happy to decrease our dependence to docker and increase it to Nix

I-Al-Istannen:
Sure, thank you! :)

PR #6633, opened 2026-02-14
feat: Add getImportKind to CtImportAssertInterface

MartinWitt:
To keep it simple I added the missing method and also a showcase. The testclass needs some refactor but this should ne done in another PR


monperrus:
thanks a lot @MartinWitt

PR #6641, opened 2026-02-20
Fix: Preserve the anonymity status of inner classes declared in methods

seintur:
This PR preserves the anonymity status of inner classes declared in methods.

Previously, all classes declared in methods were assumed to be anonymous. This PR fixes that.


seintur:
Wrong try, sorry.

PR #6642, opened 2026-02-21
fix: add access level validation in isOverriding method

vlade1k:
According to [JLS 8.4.8.3](https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.4.8.3), an overriding method must have an access modifier that is at least as accessible as the method it overrides (i.e., it cannot be more restrictive).
Added a check to enforce that the overriding method's access level is not weaker than that of the overridden method.
Additionally, private methods cannot be overridden at all, so a check was added to reject attempts to override a private method.

Before this fix, in the following situation:
```java
public class ClassA {
	public void notOverride() {
		// ....
	}
}

class ClassB extends ClassA {
	protected void notOverride() {
		// ....
	}
}
```

Method `TypeAdaptor.isOverriding` for `ClassA.notOverride` and `ClassB.notOverride` returns `true`

### DIFF
--- a/src/test/java/spoon/support/TypeAdaptorTest.java
+++ b/src/test/java/spoon/support/TypeAdaptorTest.java
@@ -628,6 +628,14 @@ class TypeAdaptorTest {
 		CtMethod<?> parentOverride1 = classA.getMethodsByName("override1").get(0);
 		CtMethod<?> childOverride1 = classB.getMethodsByName("override1").get(0);
 		assertTrue(typeAdaptor.isOverriding(childOverride1, parentOverride1));
+
+		CtMethod<?> parentOverride2 = classA.getMethodsByName("override2").get(0);
+		CtMethod<?> childOverride2 = classB.getMethodsByName("override2").get(0);
+		assertTrue(typeAdaptor.isOverriding(childOverride2, parentOverride2));
+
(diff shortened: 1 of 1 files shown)


monperrus:
thanks a lot @vlade1k

I-Al-Istannen:
Where did this come up by the way @vlade1k? The example you posted violates the JLS and does not compile in the first place. Did you run into a bug where this mattered or this some preventive help for fundamentally broken models? It's not wrong to check those invariants in `isOverriding`, but it also shouldn't come up in well-formed code?

vlade1k:
@I-Al-Istannen 
Yes, you're right, the example isn't entirely accurate. Initially, I encountered this issue while comparing private methods with matching signatures in a child and parent class. I noticed there was no check for private vs. non-private methods, so I decided it was worth adding. The remaining checks were added to complete the picture, although, of course, in other contexts this would only indicate uncompiled code.

I-Al-Istannen:
Ah, I see. Operating directly on a built model and trying to re-do the compiler work makes a lot of sense.

Thank you very much for fixing my oversights then! Enjoy your evening (if you roughly share my timezone, else you'll have to wait a bit) 🐞

Issue #6645, opened 2026-02-24
[Bug]: Spoon resolves wrong variable declaration with negated instanceof pattern and local variable shadowing


Luro02:
### Describe the bug

In the below code, in the last `System.out.println(i);` in the `main`, the variable reference `i` is resolved as a reference to the first `if (o instanceof String i) {`.

This is wrong, because `String i` is only valid inside the if block, and not outside of it.
The code worked on version `11.2.2-beta-10`, and started failing with `11.2.2-beta-11`.

I think the cause for this issue is a change introduced by https://github.com/INRIA/spoon/pull/6444

### Source code you are trying to analyze/transform

```Java
public class Main {
    public static void main(String[] args) {
        Object o = "Hello";
        if (o instanceof String i) {
            System.out.println(i);
        }
                            
        // negated if
        if (!(o instanceof String i)) {
        } else {
            System.out.println(i);
        }
                            
        // then branch cannot complete
        if (!(o instanceof String i)) {
            throw new IllegalArgumentException();
        }
        System.out.println(i);
    }
}
```

### Source code for your Spoon processing

```Java

```

### Actual output

```Java

```

### Expected output

```Java

```

### Spoon Version

11.3.1-beta-8

### JVM Version

openjdk version "21.0.2" 2024-01-16 LTS OpenJDK Runtime Environment Zulu21.32+17-CA (build 21.0.2+13-LTS) OpenJDK 64-Bit Server VM Zulu21.32+17-CA (build 21.0.2+13-LTS, mixed mode, sharing)

### What operating system are you using?

Windows 11

PR #6647, opened 2026-02-25
Fix: Correctly resolve in scope variables of instanceof patterns #6645

Luro02:
I have not looked at the JLS, so I might have missed something.

The code in the `SiblingsFunction` introduced in #6444 has been removed, because I think it does not belong there? Seems wrong to return the instanceof variables there.

### DIFF
--- a/src/main/java/spoon/reflect/visitor/filter/PotentialVariableDeclarationFunction.java
+++ b/src/main/java/spoon/reflect/visitor/filter/PotentialVariableDeclarationFunction.java
@@ -355,12 +355,6 @@ public class PotentialVariableDeclarationFunction implements CtConsumableFunctio
 	 * @return the scopes that apply to the parent
 	 */
 	private static List<Scope> updateChildScopesForParent(Collection<? extends Scope> childScopes, @NonNull CtElement child, @NonNull CtElement parent) {
-		List<Scope> filteredChildScopes = childScopes.stream()
-			// only keep the scopes that were valid for the child, the ones where the scope was less than the child
-			// can not escape the child, so they can be discarded
-			.filter(scope -> scope.element() == child)
-			.collect(Collectors.toCollection(ArrayList::new));
-
 		// This needs special handling, because of how the code is structured.
 		//
 		// The code related to type patterns operates on PatternScopes, and without this
@@ -369,6 +363,12 @@ public class PotentialVariableDeclarationFunction implements CtConsumableFunctio
 		if (child instanceof CtVariable<?> ctVariable && parent instanceof CtTypePattern) {
 			return List.of(new PatternScope(ctVariable, parent, true));
 		}
+
+		List<Scope> filteredChildScopes = childScopes.stream()
+			// only keep the scopes that were valid for the child, the ones where the scope was less than the child
+			// can not escape the child, so they can be discarded
+			.filter(scope -> scope.element() == child)
+			.collect(Collectors.toCollection(ArrayList::new));
 
 		if (child instanceof CtVariable<?> ctVariable) {
 			filteredChildScopes.add(new VariableScope(ctVariable, ctVariable));
--- a/src/main/java/spoon/reflect/visitor/filter/PotentialVariableDeclarationFunction.java
+++ b/src/main/java/spoon/reflect/visitor/filter/PotentialVariableDeclarationFunction.java
@@ -7,26 +7,35 @@
  */
 package spoon.reflect.visitor.filter;
 
-import spoon.reflect.code.CaseKind;
+import spoon.reflect.code.BinaryOperatorKind;
+import spoon.reflect.code.CtAbstractSwitch;
 import spoon.reflect.code.CtBinaryOperator;
+import spoon.reflect.code.CtBlock;
 import spoon.reflect.code.CtBodyHolder;
+import spoon.reflect.code.CtBreak;
+import spoon.reflect.code.CtCFlowBreak;
 import spoon.reflect.code.CtCase;
-import spoon.reflect.code.CtCatch;
+import spoon.reflect.code.CtCasePattern;
 import spoon.reflect.code.CtCatchVariable;
+import spoon.reflect.code.CtConditional;
+import spoon.reflect.code.CtDo;
 import spoon.reflect.code.CtFor;
 import spoon.reflect.code.CtIf;
 import spoon.reflect.code.CtExpression;
+import spoon.reflect.code.CtLabelledFlowBreak;
 import spoon.reflect.code.CtLocalVariable;
+import spoon.reflect.code.CtPattern;
+import spoon.reflect.code.CtRecordPattern;
 import spoon.reflect.code.CtStatement;
 import spoon.reflect.code.CtStatementList;
-import spoon.reflect.code.CtSwitch;
+import spoon.reflect.code.CtTryWithResource;
 import spoon.reflect.code.CtTypePattern;
+import spoon.reflect.code.CtUnaryOperator;
 import spoon.reflect.code.CtWhile;
+import spoon.reflect.code.UnaryOperatorKind;
 import spoon.reflect.declaration.CtElement;
-import spoon.reflect.declaration.CtExecutable;
 import spoon.reflect.declaration.CtField;
 import spoon.reflect.declaration.CtModifiable;
-import spoon.reflect.declaration.CtNamedElement;
 import spoon.reflect.declaration.CtPackage;
 import spoon.reflect.declaration.CtParameter;
 import spoon.reflect.declaration.CtType;
@@ -36,8 +45,13 @@ import spoon.reflect.visitor.chain.CtConsumableFunction;
 import spoon.reflect.visitor.chain.CtConsumer;
 import spoon.reflect.visitor.chain.CtQuery;
 import spoon.reflect.visitor.chain.CtQueryAware;
+import spoon.reflect.visitor.chain.CtQueryable;
 
+import java.util.ArrayList;
+import java.util.Collection;
 import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 /**
  * This mapping function searches for all {@link CtVariable} instances,
@@ -81,6 +95,7 @@ public class PotentialVariableDeclarationFunction implements CtConsumableFunctio
 
 	/**
 	 * Searches for a variable with exact name.
+	 *
 	 * @param variableName
 	 */
 	public PotentialVariableDeclarationFunction(String variableName) {
@@ -91,23 +106,15 @@ public class PotentialVariableDeclarationFunction implements CtConsumableFunctio
 	public void apply(CtElement input, CtConsumer<Object> outputConsumer) {
 		isTypeOnTheWay = false;
 		isInStaticScope = false;
-		//Search previous siblings for element which may represents the declaration of this local variable
-		CtQuery siblingsQuery = input.getFactory().createQuery()
-				.map(new SiblingsFunction().mode(SiblingsFunction.Mode.PREVIOUS))
-				//select only CtVariable nodes
-				.select(new TypeFilter<>(CtVariable.class));
-		if (variableName != null) {
-			//variable name is defined so we have to search only for variables with that name
-			siblingsQuery = siblingsQuery.select(new NamedElementFilter<>(CtNamedElement.class, variableName));
-		}
 
+		List<Scope> scopes = new ArrayList<>();
 		CtElement scopeElement = input;
-		//Search input and then all parents until first CtPackage for element which may represents the declaration of this local variable
+		// Search input and then all parents until first CtPackage for element which may represents the declaration of this local variable
 		while (scopeElement != null && !(scopeElement instanceof CtPackage) && scopeElement.isParentInitialized()) {
 			CtElement parent = scopeElement.getParent();
 			if (parent instanceof CtType<?>) {
 				isTypeOnTheWay = true;
-				//visit each CtField of `parent` CtType
+				// visit each CtField of `parent` CtType
 				CtQuery q = parent.map(new AllTypeMembersFunction(CtField.class));
 				q.forEach((CtField<?> field) -> {
 					if (isInStaticScope && !field.hasModifier(ModifierKind.STATIC)) {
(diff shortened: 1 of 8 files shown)


MartinWitt:
@Luro02 mind using our assertions and testing utils? Take a look at https://github.com/INRIA/spoon/blob/master/CONTRIBUTING.md or ask the next mistral instance.

Luro02:
> @Luro02 mind using our assertions and testing utils? Take a look at https://github.com/INRIA/spoon/blob/master/CONTRIBUTING.md or ask the next mistral instance.

Are you referring to that `assertNotNull(decl);` or do you have a problem with the `assertThat`? I am fine with changing the assertThat to a assertSame to conform with the rest of the test class.

src/main/java/spoon/reflect/visitor/filter/PotentialVariableDeclarationFunction.java
```diff
@@ -158,6 +161,13 @@ public void apply(CtElement input, CtConsumer<Object> outputConsumer) {
 				if (query.isTerminated()) {
 					return;
 				}
+
+				// Check previous sibling if statements for pattern variables that are in scope after the if
+				// Only do this when scopeElement is a statement (not when it's the body itself)
```
I-Al-Istannen: If you remember a _why_ would be useful. I can see from the if condition that it only triggers if the scope Element is a statement, but why does it matter?

src/main/java/spoon/reflect/visitor/filter/PotentialVariableDeclarationFunction.java
```diff
@@ -190,19 +200,95 @@ public void apply(CtElement input, CtConsumer<Object> outputConsumer) {
...
 
+	/**
+	 * Search for pattern variables from previous sibling if statements that may be in scope
+	 * after the if statement due to flow scoping rules.
+	 *
+	 * @param scopeElement the current scope element
+	 * @param outputConsumer the consumer to send matching variables to
+	 */
+	private void searchPatternVariablesFromPreviousSiblingIfs(CtElement scopeElement, CtConsumer<Object> outputConsumer) {
+		// Use query API to find all type patterns in previous sibling if statements
```
I-Al-Istannen: Not sure what this comment adds, but it also doesn't do much harm

src/main/java/spoon/reflect/visitor/filter/PotentialVariableDeclarationFunction.java
```diff
@@ -190,19 +200,95 @@ public void apply(CtElement input, CtConsumer<Object> outputConsumer) {
...
+		CtElement parent = typePattern.getParent();
+		return parent instanceof CtUnaryOperator<?> unaryOp
+				&& unaryOp.getKind() == UnaryOperatorKind.NOT
+				&& unaryOp.hasParent(condition);
+	}
+
+	/**
+	 * Checks if a statement completes abruptly (throws an exception or returns).
+	 */
+	private boolean isTerminalStatement(CtStatement statement) {
```
I-Al-Istannen: What is this used for? Why is the terminator check useful? And if it is useful, why is it enough to check the last node? What about `if("".isEmpty()) { throw } else { throw } `? i.e. what precisision does this check need to have?

SirYwell:
This looks somewhat incomplete, especially with more complex if conditions. For example:
- `if (!(!(a instanceof String s))) { ... }`
- `if (!(!(!(a instanceof String s)))) { ... }`
- `if (a instanceof String s || b) { ... }`
- `if (a instanceof String s && b) { ... }`
- `if (!(a instanceof String s1) || !(b instanceof String s2)) { ... }`

Could you expand the tests to cover these scenarios?

It feels like we need full DA/DU analysis here to properly support that feature...

Luro02:
This is likely going to be more involved than I expected, so I am converting this to a draft for now.

Luro02:
I went through the JLS https://docs.oracle.com/javase/specs/jls/se25/html/jls-6.html#jls-6.3.1 and implemented the pattern variable resolution accordingly.

The pushed implementation seems to be mostly working, except for switches, which I have not implemented yet. There are a lot of tests missing, many things are not covered, but I am unsure how these should look like.

The Contributing Guidelines suggest using `assertThat`, which is kind of useless here, but they should work for `assertThat(decl).isSame()` and `isNull`.

The main problem would be how to get the declarations and associated references from a piece of code. I am tempted to simply fetch a list of all declarations, and all references, then hardcode which indices of the references should resolve to which declaration, but maybe there is a cleaner way?

src/test/java/spoon/test/reference/InstanceOfReferenceTest.java
```diff
@@ -185,6 +193,122 @@ String typePattern(Object obj) {
 		assertEquals(variable, decl);
 	}
 
+	@Test
+	public void testFlowScope4() {
+		String code = """
```
I-Al-Istannen: We now have a new ModelTest variant for this pattern, just for you :P You can use `@ModelTest(source = """)` iirc.

Luro02:
Here is the current state of the PR:

The original code starts at the given element, checks if the desired variable is in scope, and if not, moves up to the parent of the element. Same repeats until it reaches the package declaration or runs out of parents.

For instanceof patterns, the scope changes through the parents, e.g. a pattern that was in scope will not be in scope anymore when negated (it would then match on false). The final scope of the pattern is defined by the statement containing the expression with the instanceof pattern. To figure out which instanceof pattern is in scope, one has to start at the pattern definition, then walk up the element tree.

### My implementation does the following:

The loop that moves up the tree keeps track of the current scopes. When it moves up to the parent, it updates the scopes with new variables that have become available (like one declared in a sibling) or removes ones that no longer apply. The code can then look at the scopes to find any variable that is in scope. 

When moving to the parent, it only knows the variables from the child, but there might be siblings that introduce variables too. The code currently hardcodes which siblings are explored, but I think always applying the siblings function might be more sensible here.

For siblings it will try to find one variable declaration (a leaf in the tree), then move up from there. Given that it uses the same code that is used for the child, it will automatically explore the other branches.

---

The code has become rather complicated, so the next steps would be to simplify where applicable. I found a few bugs in the old code as well, e.g. it used filterChildren to get the variables of a case, but then a variable declared in the block `case ..: { /* here */ }` would be discovered in the following blocks, which is wrong.

Luro02:
It has been a while since I last worked on the PR, because I was busy.

I think the implementation is far enough that it is ready for an initial review.
All tests are passing locally (let's see what the CI thinks), and most of the concerning pieces of code have been resolved/cleaned up.

It looks like I am not supposed to add custom test assertions? Will have to investigate where I should add them.

What remains to be done?
- [x] Go through the JLS and check that the code matches it, I think it does, but maybe I missed something
- [x] Add support for patterns in `a ? b : c`
- [x] Add support for switch expressions and their patterns, I skipped these JLS sections, but maybe I already covered them with the current code
- [x] Add more tests to ensure the code always behaves correctly (not sure how much time I will spend on this, maybe an AI agent can find some edge cases I might have forgotten).

What I need feedback on:
- [ ] There are two `FIXME`s, one of them is just the definition of "cannot complete normally" while the other requires some form of control flow analysis. The latter I would rather not implement, sounds like a ton of work.
- [x] I reused the old code that used the `SiblingsFunction` to resolve the variable declarations for things like executables, for loops, and so on. Here I would like to know whether I should keep it as-is or apply the mentioned fix (see TODO in the code)

src/main/java/spoon/reflect/visitor/filter/PotentialVariableDeclarationFunction.java
```diff
@@ -117,95 +124,573 @@ public void apply(CtElement input, CtConsumer<Object> outputConsumer) {
...
 				}
-				//visit parameters of CtCatch and CtExecutable (method, lambda)
-				if (parent instanceof CtCatch ctCatch) {
-					if (sendToOutput(ctCatch.getParameter(), outputConsumer)) {
-						return;
+			}
+
+			if (parent instanceof CtModifiable ctModifiable) {
+				isInStaticScope = isInStaticScope || ctModifiable.hasModifier(ModifierKind.STATIC);
+			}
```
I-Al-Istannen:
```suggestion
			if (parent instanceof CtModifiable ctModifiable && ctModifiable.hasModifier(ModifierKind.STATIC)) {
				isInStaticScope = true;
			}
```
?

src/main/java/spoon/reflect/visitor/filter/PotentialVariableDeclarationFunction.java
```diff
@@ -117,95 +124,573 @@ public void apply(CtElement input, CtConsumer<Object> outputConsumer) {
...
+
+	/**
+	 * Interface for all scopes
+	 */
+	private interface Scope {
+		/**
+		 * The variable this scope applies to.
+		 *
+		 * @return the variable
+		 */
```
I-Al-Istannen, line 169:
```suggestion
		/**
		 * {@return the variable this scope applies to}
		 */
```

src/main/java/spoon/reflect/visitor/filter/PotentialVariableDeclarationFunction.java
```diff
@@ -117,95 +124,573 @@ public void apply(CtElement input, CtConsumer<Object> outputConsumer) {
...
+		CtVariable<?> ctVariable();
+
+		/**
+		 * The element in which the variable can be referenced.
+		 *
+		 * @return the element
+		 */
```
I-Al-Istannen, line 180:
```suggestion
		/**
		 * {@return the element in which the variable can be referenced}
		 */
```

src/main/java/spoon/reflect/visitor/filter/PotentialVariableDeclarationFunction.java
```diff
@@ -117,95 +124,573 @@ public void apply(CtElement input, CtConsumer<Object> outputConsumer) {
...
+		CtElement ctElement();
+
+		/**
+		 * Checks whether the scope applies to the given element.
```
I-Al-Istannen, line 184: I can glance that from the method name, but what does "apply" mean?

src/main/java/spoon/reflect/visitor/filter/PotentialVariableDeclarationFunction.java
```diff
@@ -117,95 +124,573 @@ public void apply(CtElement input, CtConsumer<Object> outputConsumer) {
...
+	/**
+	 * Interface for all scopes
+	 */
+	private interface Scope {
+		/**
+		 * The variable this scope applies to.
+		 *
+		 * @return the variable
+		 */
+		CtVariable<?> ctVariable();
```
I-Al-Istannen: Is there any reason for prefixing everything with `ct`? I think the rest of the code doesn't do that.

src/main/java/spoon/reflect/visitor/filter/PotentialVariableDeclarationFunction.java
```diff
@@ -117,95 +124,573 @@ public void apply(CtElement input, CtConsumer<Object> outputConsumer) {
...
+	}
+
+	/**
+	 * Represents the scope of a local pattern variable.
+	 *
+	 * @param ctVariable the local variable
+	 * @param ctElement  the element in which the variable can be referenced
+	 * @param matches    for pattern variables whether the variable matches when true or false, for other variables this should be empty
+	 */
+	private record PatternScope(CtVariable<?> ctVariable, CtElement ctElement, boolean matches) implements Scope {
```
I-Al-Istannen: I think spoon's minimum Java version is high enough to make `Scope` sealed and put these into the definition of `Scope`? That might make things nicer down the line for exhaustiveness and pattern matching etc.

src/test/java/spoon/test/model/SwitchCaseTest.java
```diff
@@ -536,16 +543,15 @@ public void test_setCaseExpression_removeExistingCaseExpressionsAndInsertTheSpec
...
 			// argument
-			String code = "class A { public void f(int i) { switch(i) { case 1,2,3: break; } } }";
-			CtModel model = createModelFromString(code);
-			CtCase<?> ctCase = model.getElements(new TypeFilter<>(CtCase.class)).get(0);
+			CtCase<?> ctCase = ctClass.getElements(new TypeFilter<>(CtCase.class)).get(0);
 
 			ctCase.setCaseExpression(null);
 
-			assertThat(ctCase.getCaseExpressions().size(), equalTo(0));
+			assertThat(ctCase.getCaseExpressions()).hasSize(0);
```
I-Al-Istannen: `isEmpty()`?

src/test/java/spoon/test/query_function/VariableReferencesTest.java
```diff
@@ -395,4 +385,155 @@ public void testVariableReferenceFunctionWithEnum() throws Exception {
...
+			}
+		}
+		""")
+	public void testTryWithReferenceTest(@BySimpleName("Test") CtClass<?> ctClass) {
+		// contract: a reference to a try-with-resource variable is only accessible in the try block
+		List<CtLocalVariable<?>> variables = ctClass.getElements(new TypeFilter<>(CtLocalVariable.class));
+		List<CtVariableReference<?>> references = ctClass.getElements(new TypeFilter<>(CtVariableReference.class));
+
+		assertThat(variables)
+			.hasSize(2);
```
I-Al-Istannen:
```suggestion
		assertThat(variables).hasSize(2);
```

src/test/java/spoon/test/query_function/VariableReferencesTest.java
```diff
@@ -395,4 +385,155 @@ public void testVariableReferenceFunctionWithEnum() throws Exception {
...
+		assertNotNull(eField);
+
+
+		assertThat(references).hasSize(11);
+
+		assertThat(references.get(3)).hasExactlyPotentialDeclarations(tryWithVariable, scannerField);
+		assertThat(references.get(5)).hasExactlyPotentialDeclarations(errTryWithVariable, eField);
+		assertThat(references.get(7)).hasExactlyPotentialDeclarations(scannerField);
+		assertThat(references.get(8)).hasExactlyPotentialDeclarations(catchVariable, eField);
+		assertThat(references.get(10)).hasExactlyPotentialDeclarations(scannerField);
```
I-Al-Istannen, line 461: This assumes the order of `references`, but I guess that is fine? Hyrum is a bit scary, but I guess if somebody changes it they can change the test... It is clear enough what variable each line should be

src/test/java/spoon/testing/assertions/CtReferenceAssertInterface.java
```diff
@@ -1,10 +1,44 @@
...
+	 * For testing the variable resolution ({@link PotentialVariableDeclarationFunction}) and noticing potential mistakes
+	 * like duplicate variable declarations, it is useful to be able to assert what all potential declarations, including the ones
+	 * that are hidden by other declarations resolve to.
+	 *
+	 * @param potentialDeclarations
+	 * 		the potential declarations that the reference could resolve to, in the order they would be resolved
+	 */
+	default CtReferenceAssertInterface<? extends A, ? extends W> hasExactlyPotentialDeclarations(CtVariable<?>... potentialDeclarations) {
+		// TODO: Is this method okay here? It is technically only necessary to test the getDeclaration() resolution of a LocalVariableReference,
+		// but given that this is done in multiple test classes, I don't know where this should be placed?
```
I-Al-Istannen: Everything that helps make tests cleaner and can be nicely modeled as an AssertJ extension on a type sounds good to me :)

src/test/java/spoon/testing/assertions/CtReferenceAssertInterface.java
```diff
@@ -1,10 +1,44 @@
...
+		//
+		// TODO: Open for name suggestions
+		List<CtVariable<?>> declarations = actual().map(new PotentialVariableDeclarationFunction(actual().getSimpleName())).list();
+		// By default, the containsExactly assertion checks with equals or compareTo.
+		// A declaration `String i` and an unrelated `String i` are considered equal.
+		// Given that the potential declarations will all have the same name it is important to check
+		// that they are not only equal, but the same object through ==
+		//
+		// We are not sorting, so the else should be fine.
```
I-Al-Istannen: The else?

src/test/java/spoon/testing/assertions/CtReferenceAssertInterface.java
```diff
@@ -1,10 +1,44 @@
...
+	default CtReferenceAssertInterface<? extends A, ? extends W> hasExactlyPotentialDeclarations(CtVariable<?>... potentialDeclarations) {
+		// TODO: Is this method okay here? It is technically only necessary to test the getDeclaration() resolution of a LocalVariableReference,
+		// but given that this is done in multiple test classes, I don't know where this should be placed?
+		//
+		// TODO: Open for name suggestions
+		List<CtVariable<?>> declarations = actual().map(new PotentialVariableDeclarationFunction(actual().getSimpleName())).list();
+		// By default, the containsExactly assertion checks with equals or compareTo.
+		// A declaration `String i` and an unrelated `String i` are considered equal.
+		// Given that the potential declarations will all have the same name it is important to check
+		// that they are not only equal, but the same object through ==
```
I-Al-Istannen: I like the comment :)

src/test/java/spoon/testing/assertions/CtReferenceAssertInterface.java
```diff
@@ -1,10 +1,44 @@
...
+	 * that are hidden by other declarations resolve to.
+	 *
+	 * @param potentialDeclarations
+	 * 		the potential declarations that the reference could resolve to, in the order they would be resolved
+	 */
+	default CtReferenceAssertInterface<? extends A, ? extends W> hasExactlyPotentialDeclarations(CtVariable<?>... potentialDeclarations) {
+		// TODO: Is this method okay here? It is technically only necessary to test the getDeclaration() resolution of a LocalVariableReference,
+		// but given that this is done in multiple test classes, I don't know where this should be placed?
+		//
+		// TODO: Open for name suggestions
```
I-Al-Istannen: Its a local variable, that's fine. And for the method as a whole: I understood what it was trying to do from its name, so I think that's alright

src/test/java/spoon/test/reference/InstanceOfReferenceTest.java
```diff
@@ -185,6 +198,398 @@ String typePattern(Object obj) {
...
+			}
+		}
+		""", complianceLevel = 21)
+	public void testFlowScope4(@BySimpleName("Test") CtClass<?> ctClass) {
+		// contract: references to pattern variables hiding other variables with the same name are resolved correctly
+		List<CtLocalVariable<?>> variables = ctClass.getElements(new TypeFilter<>(CtLocalVariable.class));
+		List<CtLocalVariableReference<?>> references = ctClass.getElements(new TypeFilter<>(CtLocalVariableReference.class));
+
+		assertThat(variables)
+			.hasSize(3);
```
I-Al-Istannen:
```suggestion
		assertThat(variables).hasSize(3);
```

src/test/java/spoon/test/reference/InstanceOfReferenceTest.java
```diff
@@ -185,6 +198,398 @@ String typePattern(Object obj) {
...
+		assertThat(variables).extracting(CtLocalVariable::getSimpleName).allMatch("i"::equals);
```
I-Al-Istannen, line 226: Fancy :P

src/test/java/spoon/test/reference/InstanceOfReferenceTest.java
```diff
@@ -185,6 +198,398 @@ String typePattern(Object obj) {
...
+			}
+		}
+		""", complianceLevel = 21)
+	public void testDoWhilePattern(@BySimpleName("Test") CtClass<?> ctClass) {
+		// contract: references to pattern variables introduced in a do-while are resolved correctly
+		List<CtLocalVariable<?>> variables = ctClass.getElements(new TypeFilter<>(CtLocalVariable.class));
+		List<CtLocalVariableReference<?>> references = ctClass.getElements(new TypeFilter<>(CtLocalVariableReference.class));
+
+		assertThat(variables)
+			.hasSize(2);
```
I-Al-Istannen:
```suggestion
		assertThat(variables).hasSize(2);
```

src/test/java/spoon/test/reference/InstanceOfReferenceTest.java
```diff
@@ -185,6 +198,398 @@ String typePattern(Object obj) {
...
+			}
+		}
+		""", complianceLevel = 21)
+	public void testNegatedWhilePattern(@BySimpleName("Test") CtClass<?> ctClass) {
+		// contract: a pattern variable is introduced by while (e) S iff it is introduced by e when false
+		List<CtLocalVariable<?>> variables = ctClass.getElements(new TypeFilter<>(CtLocalVariable.class));
+		List<CtLocalVariableReference<?>> references = ctClass.getElements(new TypeFilter<>(CtLocalVariableReference.class));
+
+		assertThat(variables)
+			.hasSize(2);
```
I-Al-Istannen:
```suggestion
		assertThat(variables).hasSize(2);
```

src/test/java/spoon/test/reference/InstanceOfReferenceTest.java
```diff
@@ -185,6 +198,398 @@ String typePattern(Object obj) {
...
+				while (o instanceof String i) {
+					System.out.println(i);
+				}
+
+				System.out.println(i);
+			}
+		}
+		""", complianceLevel = 21)
+	public void testMatchingWhilePattern(@BySimpleName("Test") CtClass<?> ctClass) {
+		// contract: a pattern variable introduced by e when true is definitely matched at S.
```
I-Al-Istannen: I am not sure what this means. "a pattern variable introduced by e in while(e) inside the while body"?

src/test/java/spoon/test/reference/InstanceOfReferenceTest.java
```diff
@@ -185,6 +198,398 @@ String typePattern(Object obj) {
...
+		List<CtLocalVariable<?>> variables = ctClass.getElements(new TypeFilter<>(CtLocalVariable.class));
+		List<CtLocalVariableReference<?>> references = ctClass.getElements(new TypeFilter<>(CtLocalVariableReference.class));
+
+		assertThat(variables)
+			.hasSize(1);
```
I-Al-Istannen:
```suggestion
		assertThat(variables).hasSize(1);
```

src/test/java/spoon/test/reference/InstanceOfReferenceTest.java
```diff
@@ -185,6 +198,398 @@ String typePattern(Object obj) {
...
+			}
+		}
+		""", complianceLevel = 21)
+	public void testBinaryOperatorAnd(@BySimpleName("Test") CtClass<?> ctClass) {
+		// contract: references to pattern variables introduced by a && b are resolved correctly
+		List<CtVariable<?>> variables = ctClass.getElements(new TypeFilter<>(CtVariable.class));
+		List<CtVariableReference<?>> references = ctClass.getElements(new TypeFilter<>(CtVariableReference.class));
+
+		assertThat(variables)
+			.hasSize(12);
```
I-Al-Istannen:
```suggestion
		assertThat(variables).hasSize(12);
```

src/test/java/spoon/test/query_function/VariableReferencesTest.java
```diff
@@ -395,4 +385,155 @@ public void testVariableReferenceFunctionWithEnum() throws Exception {
...
+			}
+		}
+		""")
+	public void testForEachVariableResolution(@BySimpleName("Test") CtClass<?> ctClass) {
+		// contract: the variable referenced in a for-each resolves to the loop variable, then the field
+		List<CtVariable<?>> variables = ctClass.getElements(new TypeFilter<>(CtVariable.class));
+		List<CtVariableReference<?>> references = ctClass.getElements(new TypeFilter<>(CtVariableReference.class));
+
+		assertThat(variables)
+			.hasSize(3);
```
I-Al-Istannen:
```suggestion
		assertThat(variables).hasSize(3);
```

src/test/java/spoon/test/query_function/VariableReferencesTest.java
```diff
@@ -395,4 +385,155 @@ public void testVariableReferenceFunctionWithEnum() throws Exception {
...
+		assertThat(outerLocalVar1).isInstanceOf(CtLocalVariable.class).getSimpleName().isEqualTo("var1");
+		var runnableFieldVar1 = variables.get(1);
+		assertThat(runnableFieldVar1).isInstanceOf(CtField.class).getSimpleName().isEqualTo("var1");
+		var innerLocalVar1 = variables.get(2);
+		assertThat(innerLocalVar1).isInstanceOf(CtLocalVariable.class).getSimpleName().isEqualTo("var1");
+
+		assertThat(references.get(0)).hasExactlyPotentialDeclarations(runnableFieldVar1, outerLocalVar1);
+		assertThat(references.get(1)).hasExactlyPotentialDeclarations(innerLocalVar1, runnableFieldVar1, outerLocalVar1);
+		// The code does not filter out local variables on the way for a field reference
+		assertThat(references.get(2)).hasExactlyPotentialDeclarations(innerLocalVar1, runnableFieldVar1, outerLocalVar1);
```
I-Al-Istannen: i.e. `this` is ignored as each entry in the chain counts on its own?

src/test/java/spoon/test/query_function/VariableReferencesTest.java
```diff
@@ -395,4 +385,155 @@ public void testVariableReferenceFunctionWithEnum() throws Exception {
...
+			}
+		}
+		""")
+	public void testForInitReference(@BySimpleName("Test") CtClass<?> ctClass) {
+		// contract: a reference to a variable declared in a for loop must be resolved to the correct declaration
+		List<CtLocalVariable<?>> variables = ctClass.getElements(new TypeFilter<>(CtLocalVariable.class));
+		List<CtLocalVariableReference<?>> references = ctClass.getElements(new TypeFilter<>(CtLocalVariableReference.class));
+
+		assertThat(variables)
+			.hasSize(1);
```
I-Al-Istannen:
```suggestion
		assertThat(variables).hasSize(1);
```

src/main/java/spoon/reflect/visitor/filter/PotentialVariableDeclarationFunction.java
```diff
@@ -117,95 +124,573 @@ public void apply(CtElement input, CtConsumer<Object> outputConsumer) {
...
+				matches
+			);
+		}
+	}
+
+	private record VariableScope(CtVariable<?> ctVariable, CtElement ctElement) implements Scope {
+	}
+
+	// This searches for new scopes introduced by the branch and returns the ones that apply to the branch
+	private static List<Scope> exploreBranchForNewScopes(CtQueryable branch) {
```
I-Al-Istannen: Branch is a bit of a weird name, this can be anything, right? I.e. `int a = 20;` would also be a "branch". I think you mean that this is a "branch" in the AST in the sense that it might have children (or might be a leaf)?

src/main/java/spoon/reflect/visitor/filter/PotentialVariableDeclarationFunction.java
```diff
@@ -117,95 +124,573 @@ public void apply(CtElement input, CtConsumer<Object> outputConsumer) {
...
+			for (PatternScope scope : ctIf.getCondition() == child ? scopes : exploreBranchForNewPatternScopes(ctIf.getCondition())) {
+				// For an `if (e) S` (with maybe an else), a pattern variable introduced by `e` when `true` is definitely matched at `S`.
+				if (scope.matches()) {
+					result.add(scope.with(ctIf.getThenStatement(), true));
+				}
+
+				// A pattern variable is introduced by `if (e) S` iff
+				// -  (i) it is introduced by e when false and
+				// - (ii) S cannot complete normally.
+				if (ctIf.getElseStatement() == null && !scope.matches() && !canThenComplete) {
```
I-Al-Istannen:
```suggestion
				if ((ctIf.getElseStatement() == null && !scope.matches()) && !canThenComplete) {
```
To make part 1 and 2 clearer?

src/main/java/spoon/reflect/visitor/filter/PotentialVariableDeclarationFunction.java
```diff
@@ -117,95 +124,573 @@ public void apply(CtElement input, CtConsumer<Object> outputConsumer) {
...
+				.list(CtStatement.class)
+				.stream()
+				.flatMap(sibling -> exploreBranchForNewScopes(sibling).stream())
+				.map(scope -> (Scope) new VariableScope(scope.ctVariable(), child))
+				.toList();
+		}
+
+		// An if without a then statement `if (a);` can introduce pattern variables, but they can only be referenced from the
+		// then, which does not exist -> these can be ignored.
+		if (parent instanceof CtIf ctIf && ctIf.getThenStatement() != null) {
```
I-Al-Istannen: Doesn't this forget to handle
```java
if(in instanceof String s) {
} else {
  return;
}
System.out.println(s);
```

src/main/java/spoon/reflect/visitor/filter/PotentialVariableDeclarationFunction.java
```diff
@@ -117,95 +124,573 @@ public void apply(CtElement input, CtConsumer<Object> outputConsumer) {
...
+			// The following rule applies to a switch expression with a switch block consisting of switch rules:
+			//
+			// A pattern variable introduced by a switch label is definitely matched in the associated switch rule
+			// expression, switch rule block, or switch rule throw statement.
+			//
+			// -> All variables defined in the case expressions will be in scope for the case
+			List<Scope> result = new ArrayList<>();
+			for (var expr : ctCase.getCaseExpressions()) {
+				for (var scope : expr == child ? scopes : exploreBranchForNewPatternScopes(expr)) {
+					// NOTE: This assumes negated patterns can not appear in the case expression
```
I-Al-Istannen: Because they can not or is this a FIXME?

src/main/java/spoon/reflect/visitor/filter/PotentialVariableDeclarationFunction.java
```diff
@@ -117,95 +124,573 @@ public void apply(CtElement input, CtConsumer<Object> outputConsumer) {
...
+			//
+			// A pattern variable is introduced by a || b when false iff either
+			// -  (i) it is introduced by a when false or
+			// - (ii) it is introduced by b when false.
+			//
+			// The difference between && and || is that for && it must match true,
+			// and for || it must match false. The operator.getKind() == BinaryOperatorKind.AND is true
+			// when it is an &&, and false when it is an ||, which matches the above rules.
+			//
+			// The **either** part is only relevant for code that does not compile, which is assumed to not
```
I-Al-Istannen, line 717: Probably fine, yea

src/main/java/spoon/reflect/visitor/filter/PotentialVariableDeclarationFunction.java
```diff
@@ -117,95 +124,573 @@ public void apply(CtElement input, CtConsumer<Object> outputConsumer) {
...
+	private static boolean hasNoReachableBreakWithTarget(CtStatement ctStatement) {
+		// FIXME: This does not check whether the break statement is actually reachable
+		return ctStatement.getLabel() == null || (ctStatement instanceof CtBodyHolder bodyHolder ? bodyHolder.getBody() : ctStatement)
+			.filterChildren(new TypeFilter<>(CtBreak.class))
+			.filterChildren((CtLabelledFlowBreak ctBreak) -> ctBreak.getLabelledStatement() == ctStatement)
+			.first() == null;
+	}
+
+	/**
+	 * Updates the child scopes for the parent.
```
I-Al-Istannen: What does this mean?

src/main/java/spoon/reflect/visitor/filter/PotentialVariableDeclarationFunction.java
```diff
@@ -117,95 +124,573 @@ public void apply(CtElement input, CtConsumer<Object> outputConsumer) {
...
+		// This covers (CtBodyHolder):
+		// - CtCatch,
+		// - CtExecutable<R>: CtAnnotationMethod<T>, CtAnonymousExecutable, CtConstructor<T>, CtLambda<T>, CtMethod<T>
+		// - CtLoop: CtDo, CtFor, CtForEach, CtWhile
+		// - CtTry: CtTryWithResource,
+
+		// TODO: With the current implementation any new CtBodyHolders will be resolved by the below code, potentially causing
+		//       wrong variable resolution. Instead one could constrain the below if to
+		//       if (parent instanceof CtCatch || parent instanceof CtExecutable || ...)
+		//       Then in any new implementation the variables would resolve to null instead of a wrong variable (trickier to discover)
```
I-Al-Istannen: I am unsure. WDYT @SirYwell?

src/main/java/spoon/reflect/visitor/filter/PotentialVariableDeclarationFunction.java
```diff
@@ -117,95 +124,573 @@ public void apply(CtElement input, CtConsumer<Object> outputConsumer) {
...
+			result.add(new VariableScope(ctVariable, ctVariable));
+		}
+
+		return result.stream().filter(scope -> scope.ctElement() == branch).toList();
+	}
+
+	@SuppressWarnings("unchecked")
+	private static <T> List<T> castList(List<?> sourceList) {
+		return (List<T>) sourceList;
+	}
```
I-Al-Istannen: This feels... risky? I would inline that logic and add an explanatory comment ("Will only create VariableScope if branch is a CtVariable, which it isn't"), if you need that property

src/main/java/spoon/reflect/visitor/filter/PotentialVariableDeclarationFunction.java
```diff
@@ -117,95 +124,573 @@ public void apply(CtElement input, CtConsumer<Object> outputConsumer) {
...
+		CtElement child = branch.filterChildren(new TypeFilter<>(CtVariable.class)).first();
+		while (child != null && child != branch) {
+			// The children should always have a parent, given that they are children of the branch.
+			CtElement parent = child.getParent();
+			result = updateChildScopesForParent(result, child, parent);
+			child = parent;
+		}
+
+		// If the branch itself is a variable declaration, then this variable is in scope for the branch.
+		// The updateChildScopesForParent would have introduced it, but it was not called, because of the loop conidition
```
I-Al-Istannen:
```suggestion
		// The updateChildScopesForParent would have introduced it, but it was not called, because of the loop condition
```

src/main/java/spoon/reflect/visitor/filter/PotentialVariableDeclarationFunction.java
```diff
@@ -117,95 +124,573 @@ public void apply(CtElement input, CtConsumer<Object> outputConsumer) {
...
+	private static List<PatternScope> exploreBranchForNewPatternScopes(CtExpression<?> branch) {
+		return castList(exploreBranchForNewScopes(branch));
+	}
+
+	private static List<PatternScope> exploreBranchForNewPatternScopes(CtPattern branch) {
+		return castList(exploreBranchForNewScopes(branch));
+	}
+
+	private static boolean completesNormally(CtStatement statement) {
+		// FIXME: The JLS has a definition for what "cannot complete normally" is
```
I-Al-Istannen, line 323: We should under-approximate this, if the JLS implementation requires too much. I asked some AI and it spit out 60 lines of fun code, so I guess technically it is doable, but if you have a save under-approximation that's fine by me. Do keep a FIXME there then, though :)

I-Al-Istannen (changes requested):
Thank you very much! I added some nitpicks and some thoughts, but I don't think I have completely grasped it.

I think I found one omission (if without a then but a non-terminating else) and some other small comments.

src/main/java/spoon/reflect/visitor/filter/PotentialVariableDeclarationFunction.java
Luro02: That part is from the original code, but will update this. I think my change was just some whitespace adjustments

src/main/java/spoon/reflect/visitor/filter/PotentialVariableDeclarationFunction.java
Luro02: No, just my personal preference naming things after their types

src/main/java/spoon/reflect/visitor/filter/PotentialVariableDeclarationFunction.java
Luro02: My initial draft even had a nice switch case with pattern matching in it, but that is only java 21. I will try to add the sealed/permits

src/test/java/spoon/testing/assertions/CtReferenceAssertInterface.java
Luro02: While writing this, I was wondering why I have to specify a `Comparable` instead of for example a `BiPredicate` or something like that.

The else I am referring to here is the `Integer.compare(System.identityHashCode(a), System.identityHashCode(b))` which would define an order in which elements can be sorted.

I don't think it is smart to sort after "pointer values", but like the comment says it should only check for equality...

I will try to adjust the comment to make this easier to understand.

src/main/java/spoon/reflect/visitor/filter/PotentialVariableDeclarationFunction.java
Luro02: Yes, I was thinking of it as a tree, where each CtElement has child branches. So we are looking at another branch. I could rename it  to `exploreForNewScopes` or maybe just `explore`?

src/main/java/spoon/reflect/visitor/filter/PotentialVariableDeclarationFunction.java
Luro02: If I remember correctly, that would be represented as a `CtIf` where the then is a `CtBlock` with no statements, but I haven't checked it.

Will add a test case for this, and if you want I can remove the extra condition. It isn't necessary anyway, but it saves a few null checks

src/main/java/spoon/reflect/visitor/filter/PotentialVariableDeclarationFunction.java
Luro02: I wasn't sure... I have tried `case !(String string)`, but that is not valid.

Initially this was an assertion:
```java
if (!scope.matches) {
    throw new IllegalStateException("some message");
}
```
but I later replaced it with a comment, because I never encountered it.

src/main/java/spoon/reflect/visitor/filter/PotentialVariableDeclarationFunction.java
I-Al-Istannen: Ah, that's unfortunate :( You don't need sealed if you declare PatternScope inside the interface, but I don't mind either style. Probably doesn't buy us too much in the near future then, though. If it doesn't make the code harder to read it is still a win, I think.

src/main/java/spoon/reflect/visitor/filter/PotentialVariableDeclarationFunction.java
I-Al-Istannen: Maybe `collectScopes`? But I am a bit unsure as well. I think branch is also fine, if you add a few words explaining what I did above.

src/main/java/spoon/reflect/visitor/filter/PotentialVariableDeclarationFunction.java
I-Al-Istannen: I tried this
```java
	@ModelTest(code = """
		class Test {
			void method(Object in) {
				if(in instanceof String s) {
				} else {
				  return;
				}
				System.out.println(s);
			}
		}
		""")
	public void testFoobar(@BySimpleName("Test") CtClass<?> ctClass) {
		// contract: the variable referenced in a for-each resolves to the loop variable, then the field
		List<CtVariable<?>> variables = ctClass.getElements(new TypeFilter<>(CtVariable.class));
		List<CtVariableReference<?>> references = ctClass.getElements(new TypeFilter<>(CtVariableReference.class));

		assertThat(variables).hasSize(2);
		CtVariable<?> paramVar = variables.get(0);
		CtVariable<?> patternVar = variables.get(1);

		// System.out
		System.out.println(references);
		assertThat(references.get(1)).hasExactlyPotentialDeclarations(patternVar);
	}
```
Which failed, but maybe I was holding it wrong

src/main/java/spoon/reflect/visitor/filter/PotentialVariableDeclarationFunction.java
SirYwell: Something like `case !(String string)` doesn't really make sense, but `case CharSequence cs when !(cs instanceof String string)` is legal, although `string` is not available in the case body then.

src/main/java/spoon/reflect/visitor/filter/PotentialVariableDeclarationFunction.java
SirYwell: `CtBodyHolder` is a somewhat odd interface that doesn't really correspond to a JLS concept, so I'd rather cover the specific cases.

That also brings me to `synchronized` blocks, which is not a `CtBodyHolder` but has a similar shape. Is that handled correctly here?

Luro02: > `CtBodyHolder` is a somewhat odd interface that doesn't really correspond to a JLS concept, so I'd rather cover the specific cases.
> 
> That also brings me to `synchronized` blocks, which is not a `CtBodyHolder` but has a similar shape. Is that handled correctly here?

`CtSynchronized` isn't mentioned in the JLS for instanceof patterns, so I assume `synchronized (a instanceof B b) { ... }` is not a thing.

Other than that the CtSynchronized always has a `CtBlock`, and the variable can not escape the block. The CtBlock is handled correctly 


I will add some tests for this

SirYwell: > `CtSynchronized` isn't mentioned in the JLS for instanceof patterns, so I assume `synchronized (a instanceof B b) { ... }` is not a thing.
> 
> Other than that the CtSynchronized always has a `CtBlock`, and the variable can not escape the block. The CtBlock is handled correctly

Oh so e.g., `CtCatch` is relevant here because it introduces a variable?

Luro02: Yes, it only cares about elements that introduce variables. Then defines where in their element the variables are valid.

When moving to the parent, these will no longer apply, so they are discarded. The instanceof patterns are the special case for which we need the scopes, because they can still be valid in their parent element.

src/test/java/spoon/test/query_function/VariableReferencesTest.java
Luro02: The `PotentialVariableDeclaration` function is designed for resolving local variables. `CtFieldReferenceImpl` uses a different strategy based on the declaring type.

Technically the resolution is wrong, because `this.var1` is treated as if it were are local variable. I will adjust the test code to use the correct resolver for field references

src/main/java/spoon/reflect/visitor/filter/PotentialVariableDeclarationFunction.java
Luro02: You used the wrong reference, there are three references in the code:
1. a reference to the parameter in
2. a reference to System.out field
3. a reference to the pattern variable

I added your example as a test case

Image

Here is how the `CtIf` with an empty block is represented:

Image

The `then` is not null, so the condition holds, and it will execute the code there. It is only `null` with code like `if (a);`, but now that I am thinking about it... yeah, I missed an edge-case, which I did not expect to compile:
Image

src/main/java/spoon/reflect/visitor/filter/PotentialVariableDeclarationFunction.java
Luro02: That would be the guard of the case, which is not explored here, because it can not introduce variables that are valid in the body.

monperrus:
we're almost there to merge, @I-Al-Istannen WDYT? Thanks!

I-Al-Istannen (approved):

I-Al-Istannen:
Might be nice if @SirYwell would take another cursory glance, but I think it looks fine otherwise

src/test/java/spoon/testing/assertions/CtReferenceAssertInterface.java
```diff
@@ -1,10 +1,49 @@
...
+		} else {
+			declarations = actual().map(new PotentialVariableDeclarationFunction(actual().getSimpleName())).list();
+		}
+		// By default, the containsExactly assertion checks with equals or compareTo.
+		// A declaration `String i` and an unrelated `String i` are considered equal.
+		// Given that the potential declarations will all have the same name it is important to check
+		// that they are not only equal, but the same object through ==.
+		//
+		// The code expects a comparator, so we are using the identity hash code (= pointer value) for the comparison.
+		Assertions.assertThat(declarations).as("Potential declarations of variable <%s>", actual()).usingElementComparator(Comparator.comparing(System::identityHashCode)).containsExactly(potentialDeclarations);
```
SirYwell, line 46: That sounds rather odd, why do we need a comparator? The identity hash code also doesn't have many guarantees, so depending on what it is used for, the results might be wrong (and the comment definitely is).

src/test/java/spoon/test/query_function/VariableReferencesTest.java
```diff
@@ -395,4 +385,227 @@ public void testVariableReferenceFunctionWithEnum() throws Exception {
...
+			void method() {
+				int i = 0;
+				for (; ;) {
+					System.out.println(i);
+				}
+			}
+		}
+		""")
+	public void testForConditionNull(@BySimpleName("Test") CtClass<?> ctClass) {
+		// contract: the code must not crash if the condition in the for is null
```
SirYwell: The contract mentions a crash but the assertions do not include anything regarding that, right?

src/test/java/spoon/test/query_function/VariableReferencesTest.java
```diff
@@ -395,4 +385,227 @@ public void testVariableReferenceFunctionWithEnum() throws Exception {
...
+				if(obj instanceof String s); else {
+				  return;
+				}
+
+				System.out.println(s);
+			}
+		}
+		""")
+	public void testIfWithNullThen(@BySimpleName("Test") CtClass<?> ctClass) {
+		// contract: the code does not crash when then is null, and correctly resolves the reference to the pattern
```
SirYwell: Same as below

src/main/java/spoon/reflect/visitor/filter/PotentialVariableDeclarationFunction.java
```diff
@@ -117,95 +130,642 @@ public void apply(CtElement input, CtConsumer<Object> outputConsumer) {
...
+	 * @param child       the child for which the scopes have been found
+	 * @param parent      the parent of the passed child, this is passed separately in case the child is not initialized with the parent
+	 * @return the scopes that apply to the parent
+	 */
+	private static List<Scope> updateChildScopesForParent(Collection<? extends Scope> childScopes, @NonNull CtElement child, @NonNull CtElement parent) {
+		List<Scope> filteredChildScopes = childScopes.stream()
+			// only keep the scopes that were valid for the child, the ones where the scope was less than the child
+			// can not escape the child, so they can be discarded
+			.filter(scope -> scope.element() == child)
+			.collect(Collectors.toCollection(ArrayList::new));
```
SirYwell: Is there a reason why this is happening before the early return?

SirYwell (commented):
I wasn't able to take a closer look, but I noticed a few things.

src/test/java/spoon/testing/assertions/CtReferenceAssertInterface.java
Luro02: Do you have a suggestion on how to improve this? Like I said, assertj wants a comparator to change the equality behaviour, maybe there is a function I have missed?

src/main/java/spoon/reflect/visitor/filter/PotentialVariableDeclarationFunction.java
Luro02: This code is from an earlier revision where the special handling code wasn't present

src/test/java/spoon/testing/assertions/CtReferenceAssertInterface.java
SirYwell: I didn't find anything straightforward, but [zipSatisfy](https://www.javadoc.io/doc/org.assertj/assertj-core/3.27.7/org/assertj/core/api/AbstractIterableAssert.html#zipSatisfy(java.lang.Iterable,java.util.function.BiConsumer)) combined with an `assertThat(a).isSameAs(b)` might work? From my understanding this would require an additional assertion that the sizes match.

Luro02: I don't see how this would be better than my current implementation which does exactly what is desired.

SirYwell: Comparing the identity hash code has different semantics than comparing by `==`, that is `System.identityHashCode(a) == System.identityHashCode(b)` does not imply `a == b`. Using the identity hash code is not reliable.

Luro02: Hmm, now I understand what you mean. I think this would need something like guavas [`Ordering::arbitrary`](https://chromium.googlesource.com/external/guava-libraries/+/master/guava/src/com/google/common/collect/Ordering.java#256)

Luro02: Fortunately guava seems to already be a dependency in the spoon tests, making this very easy to fix.

Internally their code works by keeping an identity map of objects and an extra number that counts up for each entry, providing a unique identifier for comparison. (shouldn't be too difficult to implement by hand)

SirYwell (approved):
Looks good now, thanks for the work.

MartinWitt:
@SirYwell can we merge this?

MartinWitt:
@Luro02, thanks, and also thanks for all the reviewers

PR #6648, opened 2026-02-25
Fix: Anonymous class names

### LINKED ISSUE #6614: [Bug]: Can't parse inner classes located in methods
### Describe the bug

Java allows inner classes inside of methods now.  It appears that Spoon won't parse this construct correctly yet.  Is this supported yet?  Is this on the roadmap?

```Java
public class Foo {
   private static final Stats$1 extractStats$1(final String line, final Regex apacheLogRegex$1) {
      Option var3 = apacheLogRegex$1.findFirstIn(line);

      class Stats$1 implements Serializable {
         private final int count;
         private final int numBytes;

         public int count() {
            return this.count;
         }

         public int numBytes() {
            return this.numBytes;
         }

         public Stats$1 merge(final Stats$1 other) {
            return new Stats$1(this.count() + other.count(), this.numBytes() + other.numBytes());
         }

         public String toString() {
            return StringOps$.MODULE$.format$extension(Predef$.MODULE$.augmentString("bytes=%s\tn=%s"), ScalaRunTime$.MODULE$.genericWrapArray(new Object[]{BoxesRunTime.boxToInteger(this.numBytes()), BoxesRunTime.boxToInteger(this.count())}));
         }

         public Stats$1(final int count, final int numBytes) {
            this.count = count;
            this.numBytes = numBytes;
         }
      }

      if (var3 instanceof Some) {
         Some var4 = (Some)var3;
         String var5 = (String)var4.value();
         if (var5 != null) {
            Option var6 = apacheLogRegex$1.unapplySeq(var5);
            if (!var6.isEmpty() && var6.get() != null && ((List)var6.get()).lengthCompare(9) == 0) {
               String bytes = (String)((LinearSeqOps)var6.get()).apply(6);
               return new Stats$1(1, StringOps$.MODULE$.toInt$extension(Predef$.MODULE$.augmentString(bytes)));
            }
         }
      }

      return new Stats$1(1, 0);
   }
}
```

```Java
// load the provided LogQuery$.java file
String logQuerySourceText = ...
CtClass cls = Launcher.parseClass(logQuerySourceText);
System.out.println(cls.getMethodsByName("extractStats$1").get(0).toString);
```

### Actual output

```Java
    private static final org.apache.spark.examples.Stats$1 extractStats$1(final java.lang.String line, final scala.util.matching.Regex apacheLogRegex$1) {
        scala.Option var3 = apacheLogRegex$1.findFirstIn(line);
         {
            private final int count;

            private final int numBytes;

            public int count() {
                return this.count;
            }

            public int numBytes() {
                return this.numBytes;
            }

            public org.apache.spark.examples.LogQuery$$1 merge(final org.apache.spark.examples.LogQuery$$1 other) {
                return new org.apache.spark.examples.LogQuery$$1(this.count() + other.count(), this.numBytes() + other.numBytes());
            }

            public java.lang.String toString() {
                return scala.collection.StringOps$.MODULE$.format$extension(scala.Predef$.MODULE$.augmentString("bytes=%s\tn=%s"), scala.runtime.ScalaRunTime$.MODULE$.genericWrapArray(new java.lang.Object[]{ scala.runtime.BoxesRunTime.boxToInteger(this.numBytes()), scala.runtime.BoxesRunTime.boxToInteger(this.count()) }));
            }

            public 1(final int count, final int numBytes) {
                this.count = count;
                this.numBytes = numBytes;
            }
        }
        if (var3 instanceof scala.Some) {
            scala.Some var4 = ((scala.Some) (var3));
            java.lang.String var5 = ((java.lang.String) (var4.value()));
            if (var5 != null) {
                scala.Option var6 = apacheLogRegex$1.unapplySeq(var5);
                if (((!var6.isEmpty()) && (var6.get() != null)) && (((scala.collection.immutable.List) (var6.get())).lengthCompare(9) == 0)) {
                    java.lang.String bytes = ((java.lang.String) (((scala.collection.LinearSeqOps) (var6.get())).apply(6)));
                    return new org.apache.spark.examples.LogQuery$$1(1, scala.collection.StringOps$.MODULE$.toInt$extension(scala.Predef$.MODULE$.augmentString(bytes)));
                }
            }
        }
        return new org.apache.spark.examples.LogQuery$$1(1, 0);
    }

```

### Expected output

```Java

   private static final Stats$1 extractStats$1(final String line, final Regex apacheLogRegex$1) {
      Option var3 = apacheLogRegex$1.findFirstIn(line);

      class Stats$1 implements Serializable {
         private final int count;
         private final int numBytes;

         public int count() {
            return this.count;
         }

         public int numBytes() {
            return this.numBytes;
         }

         public Stats$1 merge(final Stats$1 other) {
            return new Stats$1(this.count() + other.count(), this.numBytes() + other.numBytes());
         }

         public String toString() {
            return StringOps$.MODULE$.format$extension(Predef$.MODULE$.augmentString("bytes=%s\tn=%s"), ScalaRunTime$.MODULE$.genericWrapArray(new Object[]{BoxesRunTime.boxToInteger(this.numBytes()), BoxesRunTime.boxToInteger(this.count())}));
         }

         public Stats$1(final int count, final int numBytes) {
            this.count = count;
            this.numBytes = numBytes;
         }
      }

      if (var3 instanceof Some) {
         Some var4 = (Some)var3;
         String var5 = (String)var4.value();
         if (var5 != null) {
            Option var6 = apacheLogRegex$1.unapplySeq(var5);
            if (!var6.isEmpty() && var6.get() != null && ((List)var6.get()).lengthCompare(9) == 0) {
               String bytes = (String)((LinearSeqOps)var6.get()).apply(6);
               return new Stats$1(1, StringOps$.MODULE$.toInt$extension(Predef$.MODULE$.augmentString(bytes)));
            }
         }
      }

      return new Stats$1(1, 0);
   }

```

### Spoon Version

11.2.1

### JVM Version

JDK 17

Ubuntu

seintur:
This PR provides a fix for:

- inner class names that contains the $ sign,
- classes with 10 or more inner classes.

Both cases are dealt with by fixing method https://github.com/INRIA/spoon/blob/2ecfe9d52519ad52263ea0d2534fa07134fe1966/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java#L106

An example of an inner class whose name contains the $ sign follows:

`class A { void m() { class B$1 {} }}
`

The qualified name of the inner class is:
- the source name of the outer class,
- followed by the $ sign,
- followed by the index (starting at value 1) of the inner class in the outer class,
- the source name of the inner class.

In the example, the qualified name of class B$1 is A$1B$1, and the contract of the computeAnonymousName method leads to return 1B$1.

~~For that, the computeAnonymousName method reversely iterates over anonymousQualifiedName:~~
~~- starting by skipping the trailing sourceName (in the example B$1),~~
~~- as long as digits are encountered in anonymousQualifiedName,~~
~~and returns the substring starting at the index of the first digit.~~

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java
+++ b/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java
@@ -101,11 +101,26 @@ public class JDTTreeBuilderHelper {
 	 *
 	 * @param anonymousQualifiedName
 	 * 		Qualified name which contains the anonymous name.
+	 * @param sourceName
+	 * 		The source name.
 	 * @return Anonymous simple name.
 	 */
-	static String computeAnonymousName(char[] anonymousQualifiedName) {
+	static String computeAnonymousName(char[] anonymousQualifiedName, char[] sourceName) {
+		/*
+		 * Case 1: Anonymous inner class such as
+		 * class A { void m() { new Runnable() { public void run() {} } } }
+		 * For the anonymous class that implements Runnable, anonymousQualifiedName is "A$1", and sourceName is "".
+		 *
+		 * Case 2: Non-anonymous inner class such as
+		 * class A { void m() { class B$1 {} } }
+		 * For class B$1, anonymousQualifiedName is "A$1B$1", and sourceName is "B$1".
+		 */
+		int idx = anonymousQualifiedName.length - sourceName.length - 1;
+		while (idx >= 0 && Character.isDigit(anonymousQualifiedName[idx])) {
+			idx--;
+		}
 		final String poolName = CharOperation.charToString(anonymousQualifiedName);
-		return poolName.substring(poolName.lastIndexOf(CtType.INNERTTYPE_SEPARATOR) + 1);
+		return poolName.substring(idx + 1);
 	}
 
 	/**
@@ -863,7 +878,7 @@ public class JDTTreeBuilderHelper {
 		if ((type instanceof CtClass || type instanceof CtInterface)
 				&& typeDeclaration.binding != null
 				&& (typeDeclaration.binding.isAnonymousType() || typeDeclaration.binding instanceof LocalTypeBinding && typeDeclaration.binding.enclosingMethod() != null)) {
-			type.setSimpleName(computeAnonymousName(typeDeclaration.binding.constantPoolName()));
+			type.setSimpleName(computeAnonymousName(typeDeclaration.binding.constantPoolName(), typeDeclaration.binding.sourceName));
 		} else {
 			type.setSimpleName(new String(typeDeclaration.name));
 		}
--- a/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java
+++ b/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java
@@ -98,7 +98,7 @@ public class JDTTreeBuilderHelper {
 	}
 
 	/**
-	 * Computes the anonymous simple name of the specified type..
+	 * Computes the anonymous simple name of the specified type.
 	 *
 	 * @param binding the binding for the type.
 	 * @return Anonymous simple name.
--- a/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java
+++ b/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java
@@ -39,6 +39,7 @@ import org.eclipse.jdt.internal.compiler.lookup.ProblemBinding;
 import org.eclipse.jdt.internal.compiler.lookup.ProblemReferenceBinding;
 import org.eclipse.jdt.internal.compiler.lookup.ReferenceBinding;
 import org.eclipse.jdt.internal.compiler.lookup.Scope;
+import org.eclipse.jdt.internal.compiler.lookup.SourceTypeBinding;
 import org.eclipse.jdt.internal.compiler.lookup.TagBits;
 import org.eclipse.jdt.internal.compiler.lookup.TypeBinding;
 import org.eclipse.jdt.internal.compiler.lookup.VariableBinding;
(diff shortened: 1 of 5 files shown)


seintur:
This PR should fix issue https://github.com/INRIA/spoon/issues/6614

src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java
```diff
@@ -101,11 +101,26 @@ public class JDTTreeBuilderHelper {
 	 *
 	 * @param anonymousQualifiedName
 	 * 		Qualified name which contains the anonymous name.
+	 * @param sourceName
+	 * 		The source name.
 	 * @return Anonymous simple name.
 	 */
-	static String computeAnonymousName(char[] anonymousQualifiedName) {
+	static String computeAnonymousName(char[] anonymousQualifiedName, char[] sourceName) {
```
SirYwell: Wouldn't it be more straightforward to pass the name of the enclosing class (or just the length)?

seintur: Hi SirYwell,

Thanks for suggesting.

About "to pass the name of the enclosing class". Unless I am missing something (in such a case, do not hesitate to tell), I cannot see a so straightforward way of having that in the general case.

For example, in the case of 

```
class DollarSignInInnerClassName {
  void m() {
    class B$1 {}
    class C$42 {
       void n() {
          class D$1 {}
       }
    }
  }
}
```

for class D$1, we would need to traverse typeDeclaration.binding.enclosingType(s). The enclosing type of class D$1 is class C$42 and the enclosing type of C$42 is DollarSignInInnerClassName. So in the end, I am afraid the resulting code would be trickier than reversely iterating from the end of anonymousQualifiedName.

About "or just the length". The length of the enclosing class name?

Do not hesitate to tell if I did not understand your suggestion.

I-Al-Istannen: I think he meant something like
```java
	static String computeAnonymousName(char[] anonymousQualifiedName, char[] enclosingTypeName) {
		String poolName = CharOperation.charToString(anonymousQualifiedName);
		String enclosingName = CharOperation.charToString(enclosingTypeName);
		int enclosingStartIndex = poolName.lastIndexOf(enclosingName) + "$".length();
		return poolName.substring(enclosingStartIndex + enclosingName.length());
	}
```
and then call with `computeAnonymousName(binding.constantPoolName(), binding.enclosingType().constantPoolName())`. At which point you can just take the `binding` directly and do the `.constantPoolName()` and `.enclosingType().constantPoolname()` internally.

That might be a bit more compact / clear?

I quickly prototyped it and it also passed all tests.

SirYwell: Pretty much like @I-Al-Istannen said, although I think it can be simplified further. We known that the constant pool name of the anonymous/local class consists of the constant pool name of the enclosing class, plus a `$`, plus a nonempty sequence of digits (, plus the simple name for local classes). That means the name we want to compute is the constant pool name minus the leading name of the enclosing class and minus the `$`. Something like 
```java
char[] name = binding.constantPoolName();
int n = binding.enclosingType().constantPoolName().length + 1;
return new String(name, n, name.length - n);
```
therefore should be equivalent. From a quick test, the new tests here still pass with this change.

Passing the binding instead of the char arrays is indeed a good improvement too.

seintur: Ok very good, thanks the detailed suggestion. This is commited.

src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java
```diff
@@ -97,16 +98,22 @@ public class JDTTreeBuilderHelper {
...
+		/*
+		 * Case 1: Anonymous inner class such as
+		 * class A { void m() { new Runnable() { public void run() {} } } }
+		 *
+		 * Case 2: Non-anonymous inner class such as
+		 * class A { void m() { class B$1 {} } }
+		 */
+		char[] name = binding.constantPoolName();
+		int n = binding.enclosingType().constantPoolName().length + 1;
+		return new String(name, n, name.length - n);	}
```
SirYwell: seems like the `}` slipped on the wrong line

SirYwell (changes requested):

src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java
seintur: Thanks for reporting. Last commit fixes it.

SirYwell (approved):
Thanks, looks good now!

monperrus:
thanks a lot @seintur and @SirYwell !!

PR #6669, opened 2026-03-09
Fix: When adding an inline comment in sniper mode, a newline is needed at the end of the line

seintur:
This PR fixes a bug in the sniper mode when adding a inline comment. A newline is needed at the end of the line.


seintur:
Failing test case for issue https://github.com/INRIA/spoon/issues/6665

PR #6704, opened 2026-04-17
Fix #6702: NPE in getEndColumn for VirtualFile with newlines

### LINKED ISSUE #6702: [Bug]: NPE on getEndColumn
### Describe the bug

Even though the source position of a CtCompilationUnit is valid, calling getEndColumn() throws an exception.

I'm running this in a JUnit test with a VirtualFile. If the file contains newlines I get the error, otherwise I don't.

Note that I extracted this from another unit test, so the bottom of the stacktrace does not match what you would get.

FWIW, the end position is completely bogus. I get 2147483646 for an end position.

```Java
class Source {
 // newline is essential; seems to work fine without newline
}
```

```Java
void test(String sourceCode) {
		// Load from test resources
		launcher.addInputResource(new VirtualFile(sourceCode, "x/y/z/Source.java"));
		launcher.buildModel();

		assert(1== launcher.getFactory().CompilationUnit().getMap().size());
		CtCompilationUnit ct = launcher.getFactory().CompilationUnit().getMap().values().iterator().next();
		assert(ct.getPosition().isValidPosition());
		ct.getPosition().getEndColumn();
}
```

### Actual output

```Java
java.lang.NullPointerException: Cannot invoke "String.charAt(int)" because "source" is null
	at spoon.support.reflect.cu.position.SourcePositionImpl.searchColumnNumber(SourcePositionImpl.java:87)
	at spoon.support.reflect.cu.position.SourcePositionImpl.getEndColumn(SourcePositionImpl.java:132)
	at us.str.axle.spoon.callbacks.Standard_JUnit.execCallback(Standard_JUnit.java:105)
	at us.str.axle.spoon.callbacks.Standard_JUnit.testSimple(Standard_JUnit.java:130)
```

### Expected output

```Java

```

### Spoon Version

11.3.0

### JVM Version

Java17

Alma Linux 9

caglareker:
Fixes #6702

`getOriginalSourceCode()` returns null for `VirtualFile` that contains newlines, so the loop in `searchColumnNumber()` hits NPE. Added a null check around the source access and a small regression test using `VirtualFile` with newlines. Tab counting just gets skipped when source isn't available, columns will be slightly off in that case but no crash.

### DIFF
--- a/src/test/java/spoon/test/position/PositionTest.java
+++ b/src/test/java/spoon/test/position/PositionTest.java
@@ -1445,8 +1445,8 @@ public class PositionTest {
 
 		CtCompilationUnit cu = launcher.getFactory().CompilationUnit().getMap().values().iterator().next();
 		assertTrue(cu.getPosition().isValidPosition());
-		// should not throw NPE
-		cu.getPosition().getEndColumn();
+		int endColumn = assertDoesNotThrow(() -> cu.getPosition().getEndColumn());
+		assertTrue(endColumn > 0, "getEndColumn should return a positive column");
 	}
 
 	@Test
--- a/src/test/java/spoon/test/position/PositionTest.java
+++ b/src/test/java/spoon/test/position/PositionTest.java
@@ -1436,6 +1436,20 @@ public class PositionTest {
 	}
 
 	@Test
+	public void testGetEndColumnWithVirtualFileContainingNewlines() {
+		// contract: getEndColumn() should not throw NPE when using VirtualFile with newlines
+		final Launcher launcher = new Launcher();
+		launcher.addInputResource(new VirtualFile("class Source {\n// comment\n}", "x/y/z/Source.java"));
+		launcher.getEnvironment().setNoClasspath(true);
+		launcher.buildModel();
+
+		CtCompilationUnit cu = launcher.getFactory().CompilationUnit().getMap().values().iterator().next();
+		assertTrue(cu.getPosition().isValidPosition());
+		// should not throw NPE
+		cu.getPosition().getEndColumn();
+	}
+
+	@Test
 	public void testLambdaParameterPosition() {
 		// contract: position of lambda parameter is correct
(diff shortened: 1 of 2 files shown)


src/test/java/spoon/test/position/PositionTest.java
```diff
@@ -1435,6 +1435,20 @@ public void testLinePositionOkWithOneLineClassCode() {
...
+		// contract: getEndColumn() should not throw NPE when using VirtualFile with newlines
+		final Launcher launcher = new Launcher();
+		launcher.addInputResource(new VirtualFile("class Source {\n// comment\n}", "x/y/z/Source.java"));
+		launcher.getEnvironment().setNoClasspath(true);
+		launcher.buildModel();
+
+		CtCompilationUnit cu = launcher.getFactory().CompilationUnit().getMap().values().iterator().next();
+		assertTrue(cu.getPosition().isValidPosition());
+		// should not throw NPE
+		cu.getPosition().getEndColumn();
```
SirYwell: If you want to test for the absence of an exception, use e.g., `assertDoesNotThrow`. But also please test the expected result; if it makes sense to not throw an exception, then the result should be well-specified.

caglareker: good point, switched to assertDoesNotThrow and added an assert on the returned column

SirYwell:
Given the current result of the call is `2147483621`, I wonder how much sense this fix makes. That surely is not the expected result, so we might as well fail with an exception.

One underlying problem is how virtual files and source positions interact with each other. To me, it looks like we want to improve that instead of having band-aid fixes that prevent an exception but hide the actual problem.

Would making `CtCompilationUnit#getOriginalSourceCode()` work with virtual files be possible? What are the problems there?

caglareker:
Yeah fair point, returning 2147483621 as column doesn't really help anyone. Better to fix the root cause with virtual files and source positions. Closing this.

PR #6727, opened 2026-05-16
review: test: add regression test for toString() with complex comments

MartinWitt:
Cleanup and package move


MartinWitt:
Had some claude tokens left and the test package seems really off before

monperrus:
thanks @MartinWitt

PR #6728, opened 2026-05-17
build(deps): upgrade JavaFX from 26-ea+6 to 27-ea+16 in spoon-visualisation

MartinWitt:
…


monperrus:
@MartinWitt that would be so cool if you get this one green!

PR #6741, opened 2026-05-31
refactor(testing): migrate to SpoonAssertions and deprecate old assertion package

MartinWitt:
Removes the usage of old Assertions in our codebase and marks them for deprecation. Given we have the new ones we should only use them. Removing them is downstream breaking but tbf u shouldn't use them.

### DIFF
--- a/src/test/java/spoon/testing/AbstractAssertTest.java
+++ b/src/test/java/spoon/testing/AbstractAssertTest.java
@@ -16,42 +16,68 @@
  */
 package spoon.testing;
 
-import org.junit.jupiter.api.Test;
+import org.assertj.core.api.Assertions;
 import spoon.processing.AbstractProcessor;
 import spoon.reflect.declaration.CtField;
 import spoon.reflect.declaration.CtType;
+import spoon.reflect.factory.Factory;
 import spoon.testing.processors.FooToBarProcessor;
+import spoon.testing.utils.ByClass;
+import spoon.testing.utils.ModelTest;
+import spoon.testing.utils.ProcessorUtils;
 
-import static spoon.testing.Assert.assertThat;
-import static spoon.testing.utils.ModelUtils.buildNoClasspath;
+import java.io.File;
+import java.util.List;
+
+import static spoon.testing.assertions.SpoonAssertions.assertThat;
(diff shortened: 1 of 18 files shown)


MartinWitt:
@I-Al-Istannen @SirYwell any comments?

src/test/java/spoon/testing/AbstractAssertTest.java
```diff
@@ -16,42 +16,68 @@
...
+import static spoon.testing.assertions.SpoonAssertions.assertThat;
+import static spoon.testing.utils.ModelUtils.build;
 
 public class AbstractAssertTest {
-	public static final String PATH = "./src/test/java/spoon/testing/testclasses/";
 
-	@Test
-	public void testTransformationWithProcessorInstantiated() {
-		assertThat(PATH + "Foo.java").withProcessor(new FooToBarProcessor()).isEqualTo(PATH + "Bar.java");
+	@ModelTest("./src/test/java/spoon/testing/testclasses/" + "Foo.java")
```
I-Al-Istannen: Why not one string?

src/test/java/spoon/testing/AbstractAssertTest.java
```diff
@@ -16,42 +16,68 @@
...
+	public void testTransformationWithProcessorInstantiated(Factory actual) {
+		ProcessorUtils.process(actual, List.of(new FooToBarProcessor()));
+		List<CtType<?>> actualTypes = actual.Type().getAll();
+		List<CtType<?>> expectedTypes = build(new File("./src/test/java/spoon/testing/testclasses/" + "Bar.java")).Type().getAll();
```
I-Al-Istannen: Same here

src/test/java/spoon/testing/AbstractAssertTest.java
```diff
@@ -16,42 +16,68 @@
...
+		Assertions.assertThat(actualTypes).hasSameSizeAs(expectedTypes);
```
I-Al-Istannen: We can probably use the list assert here?

I-Al-Istannen:
I think it also changes some test a bit but I like the idea :)

MartinWitt:
@I-Al-Istannen fixed ur comments

PR #6786, opened 2026-07-12
fix: do not crash on qualified generic types with unresolved arguments in noclasspath mode

### LINKED ISSUE #4077: bug: Qualified generic classes with incomplete classpath crash spoon
## Description
Explicitly qualified references for types Spoon doesn't know crashes it. If the reference uses an import, all is fine though.

## Crashing example
```java
import foo.A;
import foo.B;

class Test {
    public static void main(String args[]) {
        java.util.AbstractMap.SimpleEntry<A, B> entry = new java.util.AbstractMap.SimpleEntry<>(A.a, B.b);
    }
}
```

## Working example
```java
import foo.A;
import foo.B;
import java.util.AbstractMap.SimpleEntry;

class Test {
    public static void main(String args[]) {
        SimpleEntry<A, B> entry = new SimpleEntry<>(A.a, B.b);
    }
}
```

## Stacktrace
```
==== Building spoon model ====
Starting phase COMPILE
Phase COMPILE done! Discovered Classes: 1
Starting phase COMMENT
Phase COMMENT done! Discovered Classes: 1
Starting phase MODEL
Exception in thread "main" spoon.SpoonException: Unexpected QualifiedNameReference tokens java.util.AbstractMap.SimpleEntry<> for typeRef: java.util.AbstractMap.SimpleEntry
	at spoon.support.compiler.jdt.JDTTreeBuilderHelper.handleImplicit(JDTTreeBuilderHelper.java:557)
	at spoon.support.compiler.jdt.JDTTreeBuilderHelper.handleImplicit(JDTTreeBuilderHelper.java:488)
	at spoon.support.compiler.jdt.ReferenceBuilder.getTypeReference(ReferenceBuilder.java:518)
	at spoon.support.compiler.jdt.ReferenceBuilder.getExecutableReference(ReferenceBuilder.java:451)
	at spoon.support.compiler.jdt.JDTTreeBuilder.visit(JDTTreeBuilder.java:874)
	at org.eclipse.jdt.internal.compiler.ast.AllocationExpression.traverse(AllocationExpression.java:716)
	at org.eclipse.jdt.internal.compiler.ast.LocalDeclaration.traverse(LocalDeclaration.java:484)
	at org.eclipse.jdt.internal.compiler.ast.MethodDeclaration.traverse(MethodDeclaration.java:411)
	at org.eclipse.jdt.internal.compiler.ast.TypeDeclaration.traverse(TypeDeclaration.java:1708)
	at org.eclipse.jdt.internal.compiler.ast.CompilationUnitDeclaration.traverse(CompilationUnitDeclaration.java:827)
	at org.eclipse.jdt.internal.compiler.ast.CompilationUnitDeclaration.traverse(CompilationUnitDeclaration.java:788)
	at spoon.support.compiler.jdt.JDTBasedSpoonCompiler.traverseUnitDeclaration(JDTBasedSpoonCompiler.java:480)
	at spoon.support.compiler.jdt.JDTBasedSpoonCompiler.lambda$buildModel$0(JDTBasedSpoonCompiler.java:437)
	at spoon.support.compiler.jdt.JDTBasedSpoonCompiler.forEachCompilationUnit(JDTBasedSpoonCompiler.java:464)
	at spoon.support.compiler.jdt.JDTBasedSpoonCompiler.buildModel(JDTBasedSpoonCompiler.java:435)
	at spoon.support.compiler.jdt.JDTBasedSpoonCompiler.buildUnitsAndModel(JDTBasedSpoonCompiler.java:372)
	at spoon.support.compiler.jdt.JDTBasedSpoonCompiler.buildSources(JDTBasedSpoonCompiler.java:335)
	at spoon.support.compiler.jdt.JDTBasedSpoonCompiler.build(JDTBasedSpoonCompiler.java:116)
	at spoon.support.compiler.jdt.JDTBasedSpoonCompiler.build(JDTBasedSpoonCompiler.java:99)
	at spoon.Launcher.buildModel(Launcher.java:772)
	at de.ialistannen.javadocapi.indexing.Indexer.main(Indexer.java:62)

Process finished with exit code 1
```

mvanhorn:
## Summary

Building a model in no-classpath mode no longer throws `SpoonException: Unexpected QualifiedNameReference tokens` when source code uses a qualified nested generic type whose type arguments cannot be resolved (for example `java.util.AbstractMap.SimpleEntry<A, B>` where `A` and `B` come from a missing dependency). `handleImplicit` in `src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java` now keeps the package reference explicit and returns when the package tokens cannot be aligned with the JDT binding, instead of falling through to the throw.

## Why this matters

No-classpath mode exists precisely for analyzing code whose dependencies are absent, and this crash aborts the whole model build on a single unresolvable reference. The root cause is that JDT, lacking the type arguments, treats the declaring type as part of the package, so the Spoon-side package reference and the JDT package binding token arrays disagree in length. That mismatch is expected in this mode and not an internal error, which is what the remaining throw treated it as.

## Testing

Added `testQualifiedGenericTypeWithIncompleteClasspath` to `src/test/java/spoon/test/api/NoClasspathTest.java` using the `@ModelTest` harness with unresolvable imports. It asserts the model builds, the local variable's type resolves to `java.util.AbstractMap$SimpleEntry`, and the constructor call keeps its source-form declaring type. The full `NoClasspathTest` class passes (7 tests) with the change.

Fixes #4077

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java
+++ b/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java
@@ -649,6 +649,13 @@ public class JDTTreeBuilderHelper {
 						//keep it explicit
 						return;
 					}
+					if (packageNames != null && packageNames.length != off) {
+						/*
+						 * In no classpath mode, unresolved generic arguments can make JDT treat a declaring type as part of
+						 * the package. The package tokens cannot be aligned reliably, so keep the reference explicit.
+						 */
+						return;
+					}
 					throw new SpoonException("Unexpected QualifiedNameReference tokens " + qualifiedNameReference + " for typeRef: " + originTypeRef);
 				}
 			}
--- a/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java
+++ b/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java
@@ -649,10 +649,11 @@ public class JDTTreeBuilderHelper {
 						//keep it explicit
 						return;
 					}
-					if (packageNames != null) {
+					if (packageNames != null && packageNames.length > off) {
 						/*
-						 * In no classpath mode, unresolved generic arguments can make JDT treat a declaring type as part of
-						 * the package. The package tokens cannot be aligned reliably, so keep the reference explicit.
+						 * In no-classpath mode, unresolved generic type arguments cause JDT to fold the declaring type
+						 * into the package binding's compound name, making it longer than the remaining token count.
+						 * This mismatch is expected; keep the reference explicit rather than treating it as an error.
 						 */
 						return;
 					}
--- a/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java
+++ b/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java
@@ -649,7 +649,7 @@ public class JDTTreeBuilderHelper {
 						//keep it explicit
 						return;
 					}
-					if (packageNames != null && packageNames.length != off) {
+					if (packageNames != null) {
 						/*
 						 * In no classpath mode, unresolved generic arguments can make JDT treat a declaring type as part of
 						 * the package. The package tokens cannot be aligned reliably, so keep the reference explicit.
--- a/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java
+++ b/src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java
@@ -649,6 +649,12 @@ public class JDTTreeBuilderHelper {
 						//keep it explicit
 						return;
 					}
+					if (originTypeRef.getFactory().getEnvironment().getNoClasspath()
+							&& qualifiedNameReference instanceof QualifiedTypeReference qualifiedTypeReference
+							&& (qualifiedTypeReference.resolvedType.tagBits & TagBits.HasUnresolvedTypeVariables) != 0) {
+						// Unresolved generic arguments can leave the enclosing type modeled as part of the package.
+						return;
+					}
 					if (packageNames != null && packageNames.length > off) {
 						/*
 						 * In no-classpath mode, unresolved generic type arguments cause JDT to fold the declaring type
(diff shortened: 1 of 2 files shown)


MartinWitt:
@mvanhorn looks good, can u fix the small qodana issue?

mvanhorn:
Done — 42b8c39 drops the redundant `packageNames.length != off` check Qodana flagged. The earlier `length == off` branch already returns for the equal-length case, so by the time we reach that point `packageNames != null` implies the lengths differ and the condition was always true. Kept the focused NoClasspathTest (7/7), compile, and Checkstyle green locally; Qodana will confirm in CI.

src/main/java/spoon/support/compiler/jdt/JDTTreeBuilderHelper.java
```diff
@@ -649,6 +649,13 @@ private static void handleImplicit(PackageBinding packageBinding, char[][] token
 						//keep it explicit
 						return;
 					}
+					if (packageNames != null) {
```
SirYwell: Is this code only run in no classpath mode?

monperrus:
@mvanhorn I pushed 290b0d8 to your branch to narrow the guard to `packageNames.length > off`. The original `packageNames != null` would have silently swallowed any token-count mismatch in full-classpath mode too, masking real bugs. The specific structural signature of the no-classpath case is that JDT folds the declaring type into the package compound name, making it *longer* than the remaining token offset — so `packageNames.length > off` is the precise condition. Does that match your understanding of the JDT behaviour you observed?

mvanhorn:
Thanks @monperrus — yes, that matches exactly what I observed. In the no-classpath case JDT folds the declaring type into the package compound name, so it ends up longer than the remaining token offset; `packageNames.length > off` is the precise structural signature. Narrowing from `!= null` is a real improvement — the original would've masked genuine token-count mismatches in full-classpath mode. Appreciate the fix.

wang422403:
🙏

mvanhorn:
Ran this under Java 21 locally: after your packageNames.length > off narrowing, testQualifiedGenericTypeWithIncompleteClasspath still errored on 'Unexpected QualifiedNameReference tokens java.util.AbstractMap.SimpleEntry<>'. The remaining case is when JDT folds the declaring type into the package binding because the generic args are unresolved (HasUnresolvedTypeVariables) - added a guard for exactly that ahead of the length check. NoClasspathTest is green now (7/7) and ImportTest/TypeReferenceTest still pass. e562d85

monperrus:
thanks @mvanhorn

PR #6801, opened 2026-07-17
fix: recover anonymous resource problem bindings

martinfrancois:
## Summary

Fix type recovery for Java 10+ `var` resources when an anonymous object does not implement `AutoCloseable`.

Fixes #6800.

## Problem

Consider a Java 10+ `var` resource whose initializer is an anonymous object:

```java
try (var resource = new MissingResource() {
    void call() {}
}) {
    resource.call();
}
```

In no-classpath mode, the allocated class is unresolved and [ECJ](https://github.com/eclipse-jdt/eclipse.jdt.core) cannot prove that the object implements `AutoCloseable`. ECJ returns an `InvalidTypeForAutoManagedResource` problem binding: a placeholder for a type that it could not resolve normally. For this binding, the readable name is the complete allocation expression rather than a valid Java type name. Spoon used that expression as an identifier while creating variable and executable references, which caused `CtReferenceImpl.setSimpleName` to throw `JLSViolation`.

This caused [StoneDetector](https://github.com/StoneDetector/StoneDetector) to fail on the twelve [Elasticsearch](https://github.com/elastic/elasticsearch) files listed below. Each failure has the same malformed allocation-name root cause.

The validated Elasticsearch files are:

- [ClusterApplierServiceWatchdogTests.java](https://github.com/elastic/elasticsearch/blob/1b1acdd6b81464d626ce17c6899ecf6d38f819e3/server/src/test/java/org/elasticsearch/cluster/service/ClusterApplierServiceWatchdogTests.java)
- [TranslogTests.java](https://github.com/elastic/elasticsearch/blob/1b1acdd6b81464d626ce17c6899ecf6d38f819e3/server/src/test/java/org/elasticsearch/index/translog/TranslogTests.java)
- [SecureBytesRefRecyclerTests.java](https://github.com/elastic/elasticsearch/blob/1b1acdd6b81464d626ce17c6899ecf6d38f819e3/x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/support/SecureBytesRefRecyclerTests.java)
- [SharedBlobCacheWarmingServiceTests.java](https://github.com/elastic/elasticsearch/blob/1b1acdd6b81464d626ce17c6899ecf6d38f819e3/x-pack/plugin/stateless/src/test/java/org/elasticsearch/xpack/stateless/cache/SharedBlobCacheWarmingServiceTests.java)
[...]

PR #6803, opened 2026-07-17
fix: terminate wildcard erasure lookup

### LINKED ISSUE #6802: [Bug]: Bounded wildcard erasure recursively overflows
## Describe the bug

Type erasure is the simpler type Spoon uses when it compares generic method or constructor signatures. For example, erasing `? extends E` requires Spoon to derive a comparison type from its upper bound `E`.

In the no-classpath reproduction below, that operation enters this cycle:

1. Spoon tries to find a declaration named `?` for the wildcard.
2. Resolving the bound `E` asks for the containing method declaration.
3. Matching that method against its reference erases the wildcard parameter again.

The same lookup repeats until the JVM throws `StackOverflowError`.

A compatible fix must define wildcard erasure directly: an upper-bounded wildcard must use its bound's erasure, while a lower-bounded wildcard must use `Object`. Executable lookup must not re-enter the same executable reference through any nested or sibling type parameter.

After exact method or constructor lookup succeeds, Spoon must follow the same array, wildcard, declaring-type, and generic-argument path from the reference to the corresponding declaration. A reference created from a lexical type parameter must keep that ownership when it is reparented, cloned, or serialized.

## Source code you are trying to analyze/transform

This is the complete reduced no-classpath reproduction:

```java
import java.util.AbstractSet;
import java.util.Collection;
import java.util.Iterator;
import java.util.Map;
import java.util.Set;
import java.util.function.Function;

abstract class TypeParameterErasureRecursion<W, U> extends AbstractSet<U> {
    private final Set<W> values;

    TypeParameterErasureRecursion(Set<W> values) {
        this.values = values;
    }

    @Override
    public Iterator<U> iterator() {
        return new Iterator<>() {
            private final Iterator<W> iterator = values.iterator();

            @Override
            public boolean hasNext() {
                return iterator.hasNext();
            }

            @Override
            public U next() {
                return unwrap(iterator.next());
            }
        };
    }

    protected abstract U unwrap(W value);

    @Override
    public <A> A[] toArray(A[] array) {
        Object[] result = array;
        for (U value : this) {
            result[0] = value;
        }
        return array;
    }
}

abstract class GenericMap<E> {
    private final Map<String, E> delegate = null;

    Set<Map.Entry<String, E>> entrySet() {
        return new TypeParameterErasureRecursion<>(delegate.entrySet()) {
            @Override
            protected Map.Entry<String, E> unwrap(Map.Entry<String, E> value) {
                return value;
            }
        };
    }
}

class GenericLambda {
    static <T> void mapAll(
            Collection<? extends T> values,
            Function<T, MissingKey> mapper) {
        values.forEach(value -> consume(mapper.apply(value), value));
    }

    static <T> void consume(MissingKey key, T value) {
    }
}
```

The investigation started from a historical StoneDetector stack overflow while traversing [Elasticsearch](https://github.com/elastic/elasticsearch) [AttributeMap.java](https://github.com/elastic/elasticsearch/blob/1b1acdd6b81464d626ce17c6899ecf6d38f819e3/x-pack/plugin/esql/src/main/java/org/elasticsearch/xpack/esql/core/expression/AttributeMap.java). On the audited current Spoon `master` snapshot, that full downstream file also passes without this change, so it is a compatibility replay rather than a current reproduction. The complete reduced source above is the authoritative reproduction for this issue.

## Source code for your Spoon processing

```java
launcher.addInputResource("TypeParameterErasureRecursion.java");
model.getElements(element -> true).forEach(Object::toString);
```

## Actual output

```text
java.lang.StackOverflowError
    at CtTypeParameterReferenceImpl.getDeclaration(...)
    at CtTypeParameterReferenceImpl.getTypeErasure(...)
    at CtTypeImpl.isSameParameter(...)
    at CtTypeImpl.getMethod(...)
    at CtExecutableReferenceImpl.getExecutableDeclaration(...)
    ... repeats ...
```

## Expected output

Model traversal and wildcard erasure terminate without recursively re-entering the same executable lookup.

## Spoon Version

Reproduced on Spoon `master` at `06b187819380700384762416d15db2cc4f112175` with the reduced regression.

## JVM Version

[OpenJDK](https://github.com/openjdk/jdk) 25.0.3

## What operating system are you using?

Linux x86-64

martinfrancois:
## Summary

Prevent recursive type erasure when Spoon resolves a bounded wildcard in no-classpath mode.

Fixes #6802.

## Problem

Type erasure is the simpler type Spoon uses when it compares generic method or constructor signatures. For a parameter such as `? extends E`, Spoon tried to compute that type by finding a declaration named `?`. Resolving the bound `E` then requested the containing method declaration. Matching that method erased the same wildcard parameter again, so the cycle repeated until a `StackOverflowError` occurred.

## Change

- Erase an upper-bounded wildcard to the erasure of its bound.
- Erase a lower-bounded wildcard to `Object`.
- Guard executable matching against re-entering the same executable reference through nested or sibling type parameters.
- Traverse paired return, parameter, declaring, wildcard, array, and generic-argument types with one ordered worklist, preserving the existing first-match order.
- After exact method or constructor lookup succeeds, follow the reference's array, wildcard, declaring-type, and generic-argument path to the corresponding nested declaration.
- Keep the surrounding lexical declaration for references created from a lexical type parameter, including after reparenting, cloning, and model serialization.
- Identify parsed declaration owners by source location rather than mutable method or type names, and retain the owner marker when the model is cloned or serialized.
- Resolve unnamed in-memory source files through Spoon's compilation-unit cache so the same declaration keeps one stable owner.
- Preserve intentional not-found results in declaration lookup helpers while satisfying the repository's Qodana requirements.

## Validation

- Added a reduced no-classpath regression that produces a `StackOverflowError` before the fix and terminates after the fix.
- `mvn -q -Dtest=spoon.test.reference.TypeParameterErasureRecursionTest test` passes all 24 focused tests, including positionless overloads, renamed parsed methods and types, direct-bound reparenting, cloning, serialization, and unnamed in-memory sources.
- `mvn -q -Dtest=spoon.test.generics.GenericsTest,spoon.test.reference.TypeReferenceTest test` passes all 96 tests.
- `mvn -q verify` passes with 16 skipped and no failures or errors.
- As a compatibility replay, [StoneDetector](https://github.com/StoneDetector/StoneDetector) processes the historical [Elasticsearch `AttributeMap.java`](https://github.com/elastic/elasticsearch/blob/1b1acdd6b81464d626ce17c6899ecf6d38f819e3/x-pack/plugin/esql/src/main/java/org/elasticsearch/xpack/esql/core/expression/AttributeMap.java): AST 1/1, CFG 5/5, dominators 5/5, paths 5/5, and no reported errors. That full file also passes on the audited upstream base, so the reduced regression above is the discriminating evidence.
- All 14 GitHub checks pass, including the main Qodana check and the Java 17, 21, and 25 test matrix.

### DIFF
--- a/src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
+++ b/src/main/java/spoon/support/compiler/jdt/ReferenceBuilder.java
@@ -1086,6 +1086,10 @@ public class ReferenceBuilder {
 			return getTypeReferenceOfBoundingType(binding).clone();
 		} else {
 			CtTypeReference<?> ref = this.jdtTreeBuilder.getFactory().Core().createTypeParameterReference();
+			String declarationOwner = typeParameterDeclarationOwner(binding);
+			if (declarationOwner != null) {
+				ref.putMetadata("spoon.typeParameterDeclarationOwner", declarationOwner);
+			}
 			String name = new String(binding.sourceName());
 			if (binding.declaringElement instanceof SyntheticFactoryMethodBinding) {
 				// JDT uses these factory methods for type inference of diamond constructors. In no classpath mode they
@@ -1097,6 +1101,42 @@ public class ReferenceBuilder {
 			ref.setSimpleName(name);
 			return ref;
 		}
+	}
+
+	private static String typeParameterDeclarationOwner(TypeVariableBinding binding) {
+		if (binding.declaringElement instanceof MethodBinding method) {
+			var sourceMethod = method.sourceMethod();
+			if (sourceMethod != null && sourceMethod.compilationResult != null) {
+				return "M|" + normalizedSourceFile(sourceMethod.compilationResult.getFileName())
+						+ "|" + sourceMethod.declarationSourceStart
+						+ "|" + binding.rank;
+			}
+			return "M|" + normalizedTypeName(method.declaringClass)
+					+ "|" + new String(method.selector)
+					+ "|" + method.parameters.length
+					+ "|-1"
+					+ "|" + binding.rank;
+		}
+		if (binding.declaringElement instanceof ReferenceBinding type) {
+			if (type instanceof SourceTypeBinding sourceType
+					&& sourceType.scope != null
+					&& sourceType.scope.referenceContext != null
+					&& sourceType.scope.referenceContext.compilationResult != null) {
+				return "T|" + normalizedSourceFile(sourceType.scope.referenceContext.compilationResult.getFileName())
+						+ "|" + sourceType.scope.referenceContext.declarationSourceStart
+						+ "|" + binding.rank;
+			}
+			return "T|" + normalizedTypeName(type) + "|" + binding.rank;
+		}
+		return null;
+	}
+
+	private static String normalizedSourceFile(char[] fileName) {
+		return new String(fileName).replace('\\', '/');
+	}
+
+	private static String normalizedTypeName(ReferenceBinding type) {
+		return CharOperation.toString(type.compoundName).replace('$', '.');
 	}
 
 	private CtTypeReference<?> getTypeReferenceOfBoundingType(TypeVariableBinding binding) {
--- a/src/main/java/spoon/support/reflect/reference/CtTypeParameterReferenceImpl.java
+++ b/src/main/java/spoon/support/reflect/reference/CtTypeParameterReferenceImpl.java
@@ -9,27 +9,64 @@ package spoon.support.reflect.reference;
 
 import spoon.SpoonException;
 import spoon.reflect.declaration.CtElement;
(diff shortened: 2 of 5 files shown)


martinfrancois:
I revalidated this against current master, `9b74d80e4`. All 24 focused erasure tests pass with this PR applied. The reduced `genericAnonymousIteratorTraversalTerminates` test still throws `StackOverflowError` on master without the fix.

StoneDetector also processes the historical `AttributeMap.java` successfully with this PR applied. That is a compatibility check; the reduced test above demonstrates the remaining upstream bug.

@MartinWitt, following your merge of #6801 in the same reference-building code, would you review the declaration lookup and erasure changes here?

PR #6805, opened 2026-07-17
fix: validate deep binary expressions iteratively

### LINKED ISSUE #6804: [Bug]: Deep binary expressions overflow Spoon visitors
## Describe the bug

[OpenJDK](https://github.com/openjdk/jdk)'s valid [DeepStringConcat.java](https://github.com/openjdk/jdk/blob/be40b6bcdab37368ea3c769e575b36e290d0c6a1/test/langtools/tools/javac/DeepStringConcat.java) contains about 32,000 binary operators in one expression. `AstParentConsistencyChecker` and `ProcessingVisitor` recursively visit one operator per JVM stack frame.

The first overflow occurs while `AstParentConsistencyChecker` checks parent links. If that check is bypassed, `ProcessingVisitor` reaches the same limit while running processors.

A compatible fix may use small explicit stacks for consecutive binary operands in these two concrete visitors. Regardless of implementation, it must preserve Spoon's normal child order, callbacks before and after each child, and subclass override behavior. Malformed-parent diagnostics must not recursively pretty-print the same deep tree.

## Source code you are trying to analyze/transform

[DeepStringConcat.java](https://github.com/openjdk/jdk/blob/be40b6bcdab37368ea3c769e575b36e290d0c6a1/test/langtools/tools/javac/DeepStringConcat.java) from `openjdk/jdk@be40b6bcdab37368ea3c769e575b36e290d0c6a1`.

The file is a valid compiler stress test with a very deep string-concatenation expression.

## Source code for your Spoon processing

Building the model exposes the parent-consistency failure:

```java
launcher.addInputResource("DeepStringConcat.java");
launcher.buildModel();
```

To expose the separate processor traversal failure without first running the parent checker, disable that check and run a processor:

```java
launcher.getEnvironment().disableConsistencyChecks();
launcher.addInputResource("DeepStringConcat.java");
launcher.addProcessor(new AbstractProcessor<CtBinaryOperator<?>>() {
    @Override
    public void process(CtBinaryOperator<?> operator) {
    }
});
launcher.run();
```

## Actual output

```text
java.lang.StackOverflowError
    at spoon.reflect.visitor.CtScanner.visitCtBinaryOperator(...)
    at spoon.support.reflect.code.CtBinaryOperatorImpl.accept(...)
    at spoon.reflect.visitor.AstParentConsistencyChecker.scan(...)
    ... repeats ...
```

With consistency checks disabled, the processor run overflows in `ProcessingVisitor` instead.

## Expected output

Spoon validates and processes the legal deep expression with the normal JVM stack size.

## Spoon Version

Reproduced on Spoon `master` at [`06b187819380700384762416d15db2cc4f112175`](https://github.com/INRIA/spoon/commit/06b187819380700384762416d15db2cc4f112175).

## JVM Version

[OpenJDK](https://github.com/openjdk/jdk) 25.0.3

## What operating system are you using?

Linux x86-64

martinfrancois:
## Summary

Process very deep binary-expression trees without overflowing the JVM stack.

Fixes #6804.

## Problem

[OpenJDK](https://github.com/openjdk/jdk)'s valid [`DeepStringConcat.java`](https://github.com/openjdk/jdk/blob/be40b6bcdab37368ea3c769e575b36e290d0c6a1/test/langtools/tools/javac/DeepStringConcat.java) contains about 32,000 binary operators in one expression. Spoon recursively visited one operator per JVM frame. It first overflowed while checking parent links and then overflowed while running processors.

## Change

- Use an explicit stack when the concrete parent checker visits consecutive binary operands.
- Use the same approach in the concrete processing visitor.
- Preserve Spoon's existing child order and callbacks before and after each child.
- Preserve the normal recursive behavior for subclasses that override scanner hooks.
- Avoid recursive pretty-printing while reporting malformed parent links.
- Restore the parent checker's previous root state after a direct `accept` call.

## Validation

- Added deep-tree tests for parent checking and processor traversal.
- Added tests for child order, synchronous hooks, subclass overrides, malformed-tree diagnostics, and repeated checker use.
- `mvn -q -Dtest=spoon.reflect.visitor.AstParentConsistencyCheckerTest,spoon.support.visitor.ProcessingVisitorTest,spoon.FluentLauncherTest test` passes all 12 tests.
- `mvn -q -Dtest=spoon.test.architecture.SpoonArchitectureEnforcerTest test` passes all 9 tests. The two focused commands therefore run 21 tests in total.
- [StoneDetector](https://github.com/StoneDetector/StoneDetector) processes OpenJDK's [`DeepStringConcat.java`](https://github.com/openjdk/jdk/blob/be40b6bcdab37368ea3c769e575b36e290d0c6a1/test/langtools/tools/javac/DeepStringConcat.java): exit 0, AST 1/1, and no reported errors.
- All 14 GitHub checks pass, including the Java 17, 21, and 25 test matrix, coverage, Javadoc quality, code generation, reproducibility, Qodana, and QDJVM.

### DIFF
--- a/src/main/java/spoon/reflect/visitor/AstParentConsistencyChecker.java
+++ b/src/main/java/spoon/reflect/visitor/AstParentConsistencyChecker.java
@@ -7,12 +7,19 @@
  */
 package spoon.reflect.visitor;
 
+import java.util.ArrayDeque;
+import java.util.Deque;
+
+import spoon.reflect.code.CtBinaryOperator;
 import spoon.reflect.declaration.CtElement;
+import spoon.reflect.declaration.CtNamedElement;
+import spoon.reflect.path.CtRole;
 import spoon.reflect.reference.CtReference;
 
 public class AstParentConsistencyChecker extends CtScanner {
 
 	private CtElement parent;
+
 	@Override
 	public void scan(CtElement element) {
 		// We allow to share references across the AST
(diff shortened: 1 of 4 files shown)


monperrus:
Thanks a lot for the WIP fix. The bug is real.

https://github.com/INRIA/spoon/pull/6882 is an entirely different solution. Does it work for your use case?

martinfrancois:
@monperrus I tested #6882 at `c5888d6` against the original StoneDetector reproducer using OpenJDK `DeepStringConcat.java`.

#6882 prevents the stack overflow, but StoneDetector exits 2 because its processor runs on the new `spoon` thread. StoneDetector records the current source in a plain `ThreadLocal`; the processor therefore receives `null` and throws:

`NullPointerException: Cannot invoke "java.nio.file.Path.getParent()" because "currentFile" is null`

The run reports AST, CFG, dominator-tree, and path counts of 1/1, but the mandatory error file is nonempty. The same replay against #6805 at `d37d694` exits 0 with complete counts and an empty error file.

Therefore #6882 does not work for this use case as written. I would keep #6805 open. The large-stack mechanism can remain complementary if it is opt-in or limited to the CLI, since Spoon cannot safely transfer arbitrary caller `ThreadLocal` state to a new thread.

martinfrancois:
I repeated the comparison against current master, `9b74d80e4`, with each PR applied separately.

This PR passes all 12 focused tests. StoneDetector processes OpenJDK's `DeepStringConcat.java` with exit 0 and an empty error file.

The latest #6882, `cf822ac55`, passes its five deep-expression tests, including the new `InheritableThreadLocal` test. StoneDetector still exits 2 because its processor reads a plain `ThreadLocal` on the new `spoon` thread and receives `null` for `currentFile`.

@monperrus, following your question about #6882, is preserving the caller's processing thread a requirement Spoon wants to retain? That decision determines whether this fix or #6882 fits this use case.

PR #6847, opened 2026-08-12
fix: sniper printer crashes on switch with type pattern

### LINKED ISSUE #6846: [Bug]: Sniper printer throws SpoonException on switch with type-pattern label
### Describe the bug

```java
 spoon.SpoonException: The element of class ...CtCaseImpl does not have CtRole.PATTERN                                                                                                                                            
      at RoleHandlerHelper.getRoleHandler(RoleHandlerHelper.java:54)                                                                                                                                                               
      at ElementSourceFragment.getRoleHandler(ElementSourceFragment.java:271)                                                                                                                                                      
      at CtScanner.visitCtCasePattern(CtScanner.java:1103)   
```

```Java
class SwitchCasePattern {

	String describe(Object value) {
		return switch (value) {
			case Integer i -> "int:" + i;
			case String s -> "str:" + s;
			default -> "other";
		};
	}
}
```

```Java

```

### Actual output

```
empty java file after crash
```

### Expected output

```Java
class SwitchCasePattern {

	String describe(Object value) {
		return switch (value) {
			case Integer i -> "int:" + i;
			case String s -> "str:" + s;
			default -> "other";
		};
	}
}
```

### Spoon Version

[v11.5.0](https://github.com/INRIA/spoon/releases/tag/v11.5.0) also in [v11.5.1-beta-4](https://github.com/INRIA/spoon/releases/tag/v11.5.1-beta-4)

### JVM Version

21

Windows

GooDer:
A `case Type x ->` switch label is modelled by a synthetic CtCasePattern with no source position. While building the source fragment tree the sniper skips that positionless node and offers its child reference to the enclosing CtCase, which does not own CtRole.PATTERN, so RoleHandlerHelper.getRoleHandler threw a SpoonException that aborted printing of the whole compilation unit.

Use getOptionalRoleHandler and skip the connection when the role is absent, so the enclosing element is reprinted from its original source tokens. Adds a sniper regression test that reproduces the crash.

fix #6846

### DIFF
--- a/src/test/java/spoon/test/prettyprinter/TestSniperPrinter.java
+++ b/src/test/java/spoon/test/prettyprinter/TestSniperPrinter.java
@@ -803,11 +803,10 @@ public class TestSniperPrinter {
 	@GitHubIssue(issueNumber = 6846, fixed = true)
 	void testSniperPrintsSwitchWithTypePattern() {
 		// contract: a switch with a type-pattern label (`case Type x ->`) can be sniper printed.
-		// The type pattern is modelled by a synthetic CtCasePattern which has no source position.
-		// While building the source fragment tree the sniper skips that positionless node, so the
-		// nested reference is offered to the enclosing CtCase, which does not own that role. This
-		// used to throw a SpoonException that aborted printing of the whole compilation unit; the
-		// enclosing element must instead be reprinted from its original source tokens.
+		// This requires the CtCasePattern wrapping the type pattern to have a valid source position:
+		// without one the sniper skips the wrapper while building the source fragment tree and offers
+		// the pattern to the enclosing CtCase, which does not own CtRole.PATTERN, throwing a
+		// SpoonException that aborts printing of the whole compilation unit.
 
 		// Noop: printing the unchanged model already forces the sniper to rebuild the fragment tree.
 		Consumer<CtType<?>> noop = type -> {};
--- a/src/test/java/spoon/test/prettyprinter/TestSniperPrinter.java
+++ b/src/test/java/spoon/test/prettyprinter/TestSniperPrinter.java
@@ -800,6 +800,28 @@ public class TestSniperPrinter {
 	}
 
 	@Test
+	@GitHubIssue(issueNumber = 6846, fixed = true)
+	void testSniperPrintsSwitchWithTypePattern() {
+		// contract: a switch with a type-pattern label (`case Type x ->`) can be sniper printed.
+		// The type pattern is modelled by a synthetic CtCasePattern which has no source position.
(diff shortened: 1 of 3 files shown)


SirYwell:
This does not look like an appropriate fix. What is a "synthetic CtCasePattern"? Why doesn't it have a source position? Is the issue fixed if it had a source position?

The fix seems to be more general, are there other problems fixed by it? What are potential problems that arise from this change?

GooDer:
@SirYwell you are right that this fix will not address root cause but mitigate final corruption of class which was touched during analysis. It would probably cause problems, when you also need to do some transformation in this class that get corrupted (not my case, I did transformation to related class and other class was corrupted because of this).

I will push different fix with hopefully correct fix. I am sorry I don't have enough knowledge of internal logic of this library, I just try to support fix for my problem.

If you find it still not correct feel free to cancel this PR and let somebody else to provide fix.

Provided minimal example class from test should be then minimal example for reproduction of this problem.

GooDer:
Hi @SirYwell, how do you see second attemp of the fix?
It would be great if this bug can be fixed. It is preventing us in transforming our project via Spoon and I think that this is great library for our needs :)

src/test/java/spoon/test/prettyprinter/TestSniperPrinter.java
```diff
@@ -799,6 +799,28 @@ void testSniperRespectsSuperWithUnaryOperator() {
...
 
+	@Test
+	@GitHubIssue(issueNumber = 6846, fixed = true)
+	void testSniperPrintsSwitchWithTypePattern() {
+		// contract: a switch with a type-pattern label (`case Type x ->`) can be sniper printed.
+		// The type pattern is modelled by a synthetic CtCasePattern which has no source position.
+		// While building the source fragment tree the sniper skips that positionless node, so the
+		// nested reference is offered to the enclosing CtCase, which does not own that role. This
+		// used to throw a SpoonException that aborted printing of the whole compilation unit; the
+		// enclosing element must instead be reprinted from its original source tokens.
```
SirYwell: This isn't accurate now, I think it's better to write something along the lines "This requires the CtCasePattern to have a valid source position"

SirYwell (changes requested):
Thank you, this version looks good as it tackles the underlying problem. I have one minor change request.

src/test/java/spoon/test/prettyprinter/TestSniperPrinter.java
GooDer: Done

SirYwell (approved):
LGTM

PR #6874, opened 2026-08-31
doc: add missing Javadoc tags to ProcessingManager #3923

AzazelSensei:
Adds the missing `@param` and `@return` tags on `ProcessingManager` so the JavadocMethod checkstyle rules pass for that file.

#3923


monperrus:
thanks @AzazelSensei

PR #6881, opened 2026-09-04
doc: document SpoonObjectTableModel#scanFields #1876

AzazelSensei:
Lowered the architecture-test Javadoc threshold from 32 to 31. That flagged one public method, SpoonObjectTableModel.scanFields, which I documented.

#1876


monperrus:
thanks a lot @AzazelSensei

PR #6884, opened 2026-09-04
fix: treat upper and lower wildcard bounds as unequal

### LINKED ISSUE #6883: [Bug]: CtWildcardReference::equals does not respect isUpper
### Describe the bug

when comparing two wildcards using `equals`, they are considered equal even if one has an upper bound and the other's bound is a lower one.

```Java
irrelevant, see mre below
```

```Java
var factory = new spoon.Launcher().createFactory();
var x = factory.createWildcardReference().setBoundingType(factory.Type().createReference(Object.class)).setUpper(false);
var y = factory.createWildcardReference().setBoundingType(factory.Type().createReference(Object.class)).setUpper(true);
System.out.println(x);
System.out.println(y);
System.out.println(x.equals(y));
System.out.println(y.equals(x));
```

### Actual output

```Java
? super java.lang.Object
? extends java.lang.Object
true
true
```

### Expected output

```Java
? super java.lang.Object
? extends java.lang.Object
false
false
```

### Spoon Version

11.1.0 (i looked through every changelog since then and found none where wildcards are mentioned)

### JVM Version

21.0.10

nixos unstable 26.11.20260823.56c02bc (Zokor)

kalayciburak:
## Summary
`CtWildcardReference.equals` ignored `isUpper`, so `? super T` and `? extends T` compared equal when they shared a bounding type.

`EqualsChecker` now compares `isUpper` like other boolean metamodel fields.

fix #6883

## Test plan
- `mvn -Dtest=EqualTest#testEqualsWildcardRespectsUpperBound test` (RED before the checker change, GREEN after)
- `mvn -Dtest=EqualTest,AssignmentsEqualsTest,GenericsTest#testWildcard,TypeReferenceTest#testClearBoundsForWildcardReference test`
- `mvn spotless:check checkstyle:check`

### DIFF
--- a/src/main/java/spoon/support/visitor/equals/EqualsChecker.java
+++ b/src/main/java/spoon/support/visitor/equals/EqualsChecker.java
@@ -27,6 +27,7 @@ import spoon.reflect.reference.CtArrayTypeReference;
 import spoon.reflect.reference.CtExecutableReference;
 import spoon.reflect.declaration.CtImport;
 import spoon.reflect.reference.CtReference;
+import spoon.reflect.reference.CtWildcardReference;
 import spoon.reflect.visitor.CtInheritanceScanner;
 
 public class EqualsChecker extends CtInheritanceScanner {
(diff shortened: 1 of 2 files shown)


MartinWitt:
@kalayciburak can u fix the javadoc please?

PR #6888, opened 2026-09-05
test: cover CtPackage.setShadow #1818

AzazelSensei:
CtPackage.setShadow was on the Descartes pseudo-tested list. I added a test that stores the shadow flag and checks isShadow, so emptying the method body fails.

#1818


monperrus:
thanks @AzazelSensei

PR #6890, opened 2026-09-12
chore: study bumping the Java baseline from 17 to 21

monperrus:
# Bumps Spoon's Java baseline from 17 to 21

Motivation: Java 21 makes virtual threads final (JEP 444), which would give the Spoon pipeline a heap-grown stack for free.

This a candidate alternative to #6882's sized platform thread for fixing #6804 (stack overflow on legal deeply nested expressions, e.g. OpenJDK's `DeepStringConcat.java` with ~32,000 binary operators). See also @SirYwell's comment on #6882.

## Changes

- `spoon-pom/pom.xml`: `maven.compiler.release` / `testRelease` 17 → 21
- `.github/workflows/tests.yml`: drop Java 17 from the matrix, Windows tests on 21
- `flake.nix`: default dev shell `jdk17` → `jdk21` (the `jdk17` shell definition stays available)

Refs #6804, #6882

🤖 Generated with [agentknit](https://github.com/monperrus/agentknit)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -31,7 +31,7 @@ jobs:
     name: Tests with Java ${{ matrix.java }} on ${{ matrix.os }}
     strategy:
       matrix:
-        java: [17, 21, 25]
+        java: [21, 25]
         os: [ubuntu-latest]
     steps:
       - name: Checkout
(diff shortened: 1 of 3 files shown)


monperrus:
@SirYwell WDYT? that would be good for virtual threads. 21 is already old at internet time.

SirYwell:
I'm not sure if the changes are enough like that (especially tests might also need revisiting), but generally I'm not opposed to upgrading. I assume @MartinWitt would also be happy about it.

